PR #145, opened 2020-08-20
Create Coding Guidelines.md

EHO-makai:
A simple markdown file explaining we will use the cpp core guidelines in Project Reunion.


Coding Guidelines.md
```diff
@@ -0,0 +1,3 @@
+Project Reunion is primarily written in C++.  It will adhere to the "CPP Core Guidelines" at https://github.com/isocpp/CppCoreGuidelines.  Although it is not anticipated that we will require extensions or modifications to these guidelines, we may create a process for such extensions or modifications in the future.
```
jevansaks, line 1: Can we link to the pretty-printed version?

```suggestion
Project Reunion is primarily written in C++.  It will adhere to the [CPP Core Guidelines](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html).  Although it is not anticipated that we will require extensions or modifications to these guidelines, we may create a process for such extensions or modifications in the future.  
```

Coding Guidelines.md
```diff
@@ -0,0 +1,6 @@
+Project Reunion implementations prefer C++ and C++/WinRT implementations.   New code must adhere to the [cppcoreguidelines](https://github.com/isocpp/CppCoreGuidelines)
```
DrusTheAxe, line 1: Please steal content from https://github.com/microsoft/ProjectReunion/compare/user/drustheaxe/styleguide 

We want all that content, but better in a coding guidelines.md than in contributor.md. I'll scrap my branch after you assimiliate its meat :-)

EHO-makai: incorporated the items as discussed offline.  thank you. :-)

Coding Guidelines.md
```diff
@@ -0,0 +1,16 @@
+#### C++
```
jonwis, line 1: Small nit on markdown style.  This is a fourth-level header. Use a single \# to start a section, then multiple \#\# within to make sub-sections. So this would be:

```md
# C++

Content goes here

# Markdown

Content goes here

# Other Languages

Content goes here
```

jonwis: Might start with a "General" section, like:

```md
# General

Project Reunion prefers using industry-standard coding styles, guidelines, and
patterns for any languages used in implementation or testing.
```

jonwis:
Does this filename have a literal space in the name? Maybe instead use `coding-guidelines.md` ?

Coding Guidelines.md
EHO-makai: Missed this one, making this changes to both links.  I thought (mistakenly) that people preferred the "inline" version.

Coding Guidelines.md
EHO-makai: done.

PR #154, opened 2020-08-26
User/eho makai/reunion metapackage

EHO-makai:
Added a ProjectReunionMetaPackage.nuspec file and added build steps to build it and publish it to artifacts/drop.  For now it has a single reference for the Win2d.uwp Nuget package (version 1.25.0).  

Tested by taking the Win2d "simple sample", removing the Win2D reference, downloading the build artifact, setting up a nuget package source to a local directory and then installing and referencing the Microsoft.ProjectReunion.MetaPackage nuget package and building the sample.  

Once we publish the ProjectReunion nuget package (the one with actual contents) we can reference it but publishing has not occurred yet.

### DIFF
--- a/build/NuSpecs/ProjectReunionMetaPackage.nuspec
+++ b/build/NuSpecs/ProjectReunionMetaPackage.nuspec
@@ -1,0 +1,34 @@
+<?xml version="1.0" encoding="utf-8" standalone="yes"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2013/05/nuspec.xsd">
+  <metadata>
+    <id>$ID$</id>
+    <version>0.0.0-SpecifyVersionOnCommandline</version>
+    <title>Microsoft Project Reunion Meta Package</title>
+    <authors>Microsoft</authors>
+    <owners>Microsoft</owners>
+    <description>Project Reunion Meta Package</description>
+    <copyright>© Microsoft Corporation. All rights reserved.</copyright>
+    <tags>Windows Reunion ProjectReunion</tags>
+    <requireLicenseAcceptance>true</requireLicenseAcceptance>
+    <license type="file">license.txt</license>
+    <projectUrl>https://github.com/microsoft/projectreunion</projectUrl>
+    <iconUrl>https://aka.ms/winui_icon</iconUrl>
+  </metadata>
+  <files>
+    <file src="readme.txt"/>
+    <file src="license.txt"/>
+
+    <file target="runtimes" src="$RUNTIMESDIR$\**\*.*"/>
+
+    <!-- Common build logic -->
+    <file target="build\$ID$.targets" src="ProjectReunion-Nuget-Common.targets"/>
+    <file target="buildTransitive\$ID$.targets" src="ProjectReunion-Nuget-Common.targets"/>
+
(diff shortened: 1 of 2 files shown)


build/NuSpecs/ProjectReunionMetaPackage.nuspec
```diff
@@ -0,0 +1,34 @@
...
+    <owners>Microsoft</owners>
+    <description>Project Reunion Meta Package</description>
+    <copyright>© Microsoft Corporation. All rights reserved.</copyright>
+    <tags>Windows Reunion ProjectReunion</tags>
+    <requireLicenseAcceptance>true</requireLicenseAcceptance>
+    <license type="file">license.txt</license>
+    <projectUrl>https://github.com/microsoft/projectreunion</projectUrl>
+    <iconUrl>https://aka.ms/winui_icon</iconUrl>
+    <dependencies> 
+      <!-- List all Project Reunion references here, please maintain in alphabetical order-->
```
DrusTheAxe, line 17: Should be enforced by tooling. Could be a PR-complete gate, could be a build task or nightly task or... Not saying tooling should automagically sort it; possibly, or could possibly merely monitor and flag incorrectness for human follow up

Please create an issue to track the need for tooling to verify correctness

build/NuSpecs/ProjectReunionMetaPackage.nuspec
```diff
@@ -0,0 +1,34 @@
...
+      <dependency id="Win2D.uwp" version="1.25.0"/>
```
DrusTheAxe, line 18: Is this data needed anywhere else? If so, do we generate those from this? Alternatively should those and this be generated from Dependencies.txt etc?

build/NuSpecs/ProjectReunionMetaPackage.nuspec
```diff
@@ -0,0 +1,34 @@
...
+    </dependencies>
+  </metadata>
+  <files>
+    <file src="readme.txt"/>
+    <file src="license.txt"/>
+
+    <file target="runtimes" src="$RUNTIMESDIR$\**\*.*"/>
```
DrusTheAxe, line 25: *.* or just * ?

build/NuSpecs/ProjectReunionMetaPackage.nuspec
EHO-makai: I don't see the value of having another file that just puts the same data into the nuspec file.  Feel free to log an issue about this, though.  Not promising it will get done.

build/NuSpecs/build-nupkg.ps1
```diff
@@ -113,8 +113,33 @@ $CommonNugetArgs = "-properties `"BuildOutput=$BuildOutput``;ID=$nupkgtitle``;RU
...
 $nugetExe = "$scriptDirectory\..\..\tools\NugetWrapper.cmd"
 $NugetCmdLine = "$nugetExe pack ProjectReunion.nuspec $NugetArgs -version $version"
+Write-Host 'Building Project Reunion package'
+Write-Host $NugetCmdLine
+Invoke-Expression $NugetCmdLine
+if ($lastexitcode -ne 0)
+{
+    Write-Host "Nuget returned $lastexitcode"
+    Exit $lastexitcode; 
+}
```
jonwis, line 129: Future: Functionify this.

PR #156, opened 2020-08-27
API Spec for Reunion Deployment Information API

dkbennett:
First cut at proposed API spec for feature proposal: https://github.com/microsoft/ProjectReunion/issues/155

Feedback appreciated, thank you!


specs/projectreuniondeployment/reuniondeployment.md
```diff
@@ -0,0 +1,106 @@
+
+# Background
+
+This is the spec for proposal [Project Reunion Deployment Information API #155](https://github.com/microsoft/ProjectReunion/issues/https://github.com/microsoft/ProjectReunion/issues/155).
+
+Reunion consists of a Main package and framework published by Microsoft while consumers of Reunion will often be non-Microsoft publishers. In order to have Reunion fully present and functional, the Reunion Framework and Main packages must both be deployed, and depending on the operating system, OS updates may also be required.  While the Windows Application Model naturally supports framework dependencies, it does not have a means of declaring a dependency across Main packages of different publishers. This API spec and feature proposal is to provide a means of identifying when the Reunion dependencies are not satisfied, and if that is the case, provide URIs for where the missing dependencies could be satisifed by the user and/or the developer.
```
DrusTheAxe, line 6: Project Reunion. Never say just Reunion

DrusTheAxe, line 6: Please reformat so lines are generally <=100 chars long. See [Markdown](https://github.com/microsoft/ProjectReunion/blob/master/docs/Coding-Guidelines.md#markdown) for more details

I'll hold off commenting on the spec until this to make diffs easier to manage

dkbennett: Fixed everywhere, including the filename.

dkbennett: Fixed for everything except code lines with long names& namespaces that couldn't work out right  and the link.

specs/projectreuniondeployment/projectreuniondeployment.md
```diff
@@ -0,0 +1,138 @@
...
+{
+    /// Represents the current Deployment status of Project Reunion
+    runtimeclass ProjectReunionDeploymentStatus
+    {
+        /// Returns True if Project Reunion is the desired Version and no updates or packages
+        /// are required.
+        Boolean IsOK { get; };
+
+        // Returns True if Windows Updates are required for Project Reunion.
+        Boolean WindowsUpdatesRequired { get; };
```
DrusTheAxe, line 114: NIT: IsWindowsUpdateRequired ?

Q to API Review. What's the common pattern to apply here?

specs/projectreuniondeployment/projectreuniondeployment.md
```diff
@@ -0,0 +1,138 @@
...
+
+        // Returns True if a Package install is required for Project Reunion.
+        Boolean PackageInstallRequired { get; };
```
DrusTheAxe, line 117: IsPackageInstallRequired ?

Q to API Review. What's the common pattern to apply here

specs/projectreuniondeployment/projectreuniondeployment.md
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    /// Used to query deployment information for Project Reunion
+    static runtimeclass ProjectReunionDeployment
+    {
+        /// Returns the current deployment status of Project Reunion and whether it meets the 
+        /// minimum package version specified.
+        static ProjectReunionDeploymentStatus GetStatus(Windows.ApplicationModel.PackageVersion packageVersion);
```
DrusTheAxe, line 125: minVersion

DrusTheAxe, line 125: Do you also need an optional 'architecture' parameter?

Does ProjectReunionFramework.msix come in per-processor architecture flavors and there can be >1 registered to a user? e.g. on an x64 machine can I have ProjectReunionFramework-x64.msix and ProjectReunionFramework-x86.msix?

Will ProjectReunionMain come in multiple flavors e.g. ProjectReunionMainx64 and ProjectReunionMainx86?

If there's multiple possible architecture flavors registered to a user, do you check them all or only the ones I need, e.g. if I'm an x86 app do I care about Fwk-x86 *and* Fwk-x64? About Main-x86 *and* Main-x64?

specs/projectreuniondeployment/projectreuniondeployment.md
```diff
@@ -0,0 +1,138 @@
...
+
+        /// Returns a URI for the direct install link for the latest Project Reunion Package.
+        static Windows.Foundation.Uri GetProjectReunionPackageLink();
```
DrusTheAxe, line 128: What's the Uri point to? URI for the main package e.g. https://.../microsoft.projectreunion.main.msix ?

Need >1? See previous Q re architecture

DrusTheAxe, line 128: Are Main+Framework package versions in lockstep? If not then this should be GetPackageLinks() returning 1+ links? 0+ links?

DrusTheAxe, line 128: GetPackageLink()?

Remove ProjectReunion from the method name?

specs/projectreuniondeployment/projectreuniondeployment.md
```diff
@@ -0,0 +1,138 @@
...
+
+        /// Returns a URI for a user-friendly install instruction for Project Reunion and any
+        /// OS updates that may be required.
+        static Windows.Foundation.Uri GetProjectReunionInstallInformationLink();
```
DrusTheAxe, line 132: This gets 'link for-human-followup vs the other method returns a link for programmatic-followup. Why the dicrepany in names?

```GetProjectReunionPackageLink```
```GetProjectReunionInstallInformationLink```

i.e. Package vs InstallInformation?

CONSIDER: symmetric names e.g. InstallPackage vs InstallInformation. Or Install vs Information? Or Package vs Interactive?

Comment (dkbennett) [reply specs/projectreuniondeployment/projectreuniondeployment.md]: Wasn't intending to have a name discrepancy, but changed "Package" to "InstallPackage" and removed the ProjectReunion from the names since that is in the namespace.

File: specs/projectreuniondeployment/projectreuniondeployment.md (hunk 5 above)
Comment (dkbennett) [reply specs/projectreuniondeployment/projectreuniondeployment.md]: Yes done.

Comment (dkbennett) [reply specs/projectreuniondeployment/projectreuniondeployment.md]: Assuming for now that they are kept updated in lock-step. Regardless, the InstallPackage link is a .appinstaller file. So perhaps this should actually be "AppInstallerLink" ?

Comment (dkbennett) [reply specs/projectreuniondeployment/projectreuniondeployment.md]: The package link should be a .appinstaller package, with an architecture that matches the current system, so no need to provide the architecture. I expect we will have different appinstaller files for each architecture.

File: specs/projectreuniondeployment/projectreuniondeployment.md (hunk 4 above)
Comment (dkbennett) [reply specs/projectreuniondeployment/projectreuniondeployment.md:125]: It will follow normal deployment applicability, which is on x64 machines would require and check both x86 and x64 frameworks are present and of the specified min version, which means an architecture parameter is not required. That also seems incorrect to me for a developer to specify architecture, when we can easily determine that and do the right thing.

File: specs/projectreuniondeployment/projectreuniondeployment.md [hunk 7]
```diff
@@ -0,0 +1,138 @@
...
+# Background
+
+This is the spec for proposal [Project Reunion Deployment Information API #155](https://github.com/microsoft/ProjectReunion/issues/https://github.com/microsoft/ProjectReunion/issues/155).
+
+Project Reunion consists of a Main package and framework published by Microsoft while consumers
+of Project Reunion will often be non-Microsoft publishers. In order to have Project Reunion fully
+present and functional, the Project Reunion Framework and Main packages must both be deployed, and
+depending on the operating system, OS updates may also be required.  While the Windows Application
+Model naturally supports framework dependencies, it does not have a means of declaring a dependency
+across Main packages of different publishers. This API spec and feature proposal is to provide a
```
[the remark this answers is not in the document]
dkbennett: Fixed.

specs/projectreuniondeployment/projectreuniondeployment.md
```diff
@@ -0,0 +1,138 @@
+
+# Background
+
+This is the spec for proposal [Project Reunion Deployment Information API #155](https://github.com/microsoft/ProjectReunion/issues/https://github.com/microsoft/ProjectReunion/issues/155).
+
+Project Reunion consists of a Main package and framework published by Microsoft while consumers
+of Project Reunion will often be non-Microsoft publishers. In order to have Project Reunion fully
+present and functional, the Project Reunion Framework and Main packages must both be deployed, and
```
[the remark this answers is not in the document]
dkbennett: Fixed.

specs/projectreuniondeployment/projectreuniondeployment.md
```diff
@@ -0,0 +1,138 @@
+
+# Background
+
+This is the spec for proposal [Project Reunion Deployment Information API #155](https://github.com/microsoft/ProjectReunion/issues/https://github.com/microsoft/ProjectReunion/issues/155).
+
+Project Reunion consists of a Main package and framework published by Microsoft while consumers
```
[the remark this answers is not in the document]
dkbennett: Fixed.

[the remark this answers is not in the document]
dkbennett: Clarified

specs/projectreuniondeployment/projectreuniondeployment.md
```diff
@@ -0,0 +1,138 @@
...
+of Project Reunion will often be non-Microsoft publishers. In order to have Project Reunion fully
+present and functional, the Project Reunion Framework and Main packages must both be deployed, and
+depending on the operating system, OS updates may also be required.  While the Windows Application
```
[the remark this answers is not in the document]
dkbennett: Restructured sentence.

specs/projectreuniondeployment/projectreuniondeployment.md
```diff
@@ -0,0 +1,138 @@
+
+# Background
+
+This is the spec for proposal [Project Reunion Deployment Information API #155](https://github.com/microsoft/ProjectReunion/issues/https://github.com/microsoft/ProjectReunion/issues/155).
+
+Project Reunion's deployment consists of a Main package and Framework package published by
```
stevenbrix, line 6: It would be worth describing what a "Main package" and "Framework Package" is for folks who aren't familiar with the terms. I don't even know what a Main package is 😊

specs/projectreuniondeployment/projectreuniondeployment.md
```diff
@@ -0,0 +1,138 @@
...
+Still, the developer of programs and app packages can never assume that the version of Project
+Reunion they require is already available, so this feature will provide an API to check that
+availability and provide links for resolving any deployment issues. This API could be expanded
+in the future to wrap deployment behavior for software capable of deploying packages and/or OS
+updates to reduce the end-user friction and minimize potential for developers to lose customers
+due to dropoff during the acquistion of Project Reunion.
+
+# Description
+
+When a Project Reunion-consuming package or app is deployed, it is expected to only have the App
```
stevenbrix, line 33: is this a packaged or unpackaged app, or both?

specs/projectreuniondeployment/projectreuniondeployment.md
```diff
@@ -0,0 +1,138 @@
...
+missing dependencies.
+
+The developer scenarios for this API is to answer the questions:
+1. Does this system have the min version of Project Reunion I require?
+2. What do I do to resolve this if it doesn't?
+
+The API addresses #1 by a simple query of the version they seek and a possible check for OS
+updates. #2 has two possible answers. First is direct resolution by the developer by acquiring and
+installing the packages if possible. Some programs will not have permission or motivation to do
+this, so the second answer is providing a link up to the user where the user can take action to
```
stevenbrix, line 57: By "user" are you referring to someone using Windows, or a developer? Popping up install messages to the user sounds icky. When and why would this happen?

specs/projectreuniondeployment/projectreuniondeployment.md
```diff
@@ -0,0 +1,138 @@
...
+{
+    ...
+}
+```
+
+## GetInstallPackageLink
+
+Provides a direct install link to deploy the latest version of Project Reunion. If no OS Updates
+are required and the program or app has permissions to deploy packages, this can directly resolve
+the missing dependency without burdening the user.
```
stevenbrix, line 84: side note: we should never burden the user

specs/projectreuniondeployment/projectreuniondeployment.md
```diff
@@ -0,0 +1,138 @@
...
+
+```C++
+Windows::Foundation::Uri packageLink = Microsoft::Management::Deployment::ProjectReunionDeployment::GetInstallPackageLink();
+```
+
+## GetInstallInformationLink
+
+Provides a user-friendly link to acquire the latest version of Project Reunion and any OS updates
```
stevenbrix, line 92: Popping up a link to go install some it of software seems ok for developers (although i'm personally always annoyed and frazzled when this happens in VS). It does not seem like something a non-technical user of Windows should ever experience

it would be good to know what scenarios this can happen in so that developers don't shoot themselves in the foot by going down this route

specs/projectreuniondeployment/projectreuniondeployment.md
```diff
@@ -0,0 +1,138 @@
...
+namespace Microsoft.Management.Deployment
+{
+    /// Represents the current Deployment status of Project Reunion
+    runtimeclass ProjectReunionDeploymentStatus
+    {
+        /// Returns True if Project Reunion is the desired Version and no updates or packages
+        /// are required.
+        Boolean IsOK { get; };
+
+        // Returns True if Windows Updates are required for Project Reunion.
```
stevenbrix, line 113: What does this mean to a developer? What do I do if a Windows Update is required?

stevenbrix:
How are developers getting bits on their machine to debug and test apps?

specs/projectreuniondeployment/projectreuniondeployment.md
```diff
@@ -0,0 +1,138 @@
...
+
+This is the means by which an app or program can call to verify Project Reunion is present to a
+minimum version. It should be called once after activation before trying to use Project Reunion
+functionality. This method returns a ProjectReunionStatus object that describes the current
+deployment state of Project Reunion.
+
+```C++
+Windows::ApplicationModel::PackageVersion version{1, 0, 0, 0};
+auto status =
+Microsoft::Management::Deployment::ProjectReunionDeployment::GetStatus(version);
```
dotMorten, line 73: I assume the "ProjectReunionDeployment" class name is temporary working name, and "ProjectReunion" will be dropped from these APIs?

specs/projectreuniondeployment/projectreuniondeployment.md
dotMorten: I agree. Each developer is going to be exposing this information differently. I'd rather this is an OS or installer feature, and giving users a consistent experience and gives a lot of confidence and trust in the process of installing more stuff.

specs/projectreuniondeployment/projectreuniondeployment.md
DrusTheAxe: If it's a URI to *.appinstaller then yes, it should be AppInstallerLink or AppIntallerUri

Or maybe AppInstallerInfo a la [W.AM.Package.GetAppInstallerInfo](https://docs.microsoft.com/uwp/api/windows.applicationmodel.appinstallerinfo?view=winrt-19041)?

stevenbrix:
Last I heard, we were following the WinUI2.x model for shipping, which means that pre-release packages will ship the assemblies app-local, is this still the plan?

stevenbrix:
This spec seems fairly light on "how do I use this" - can we get some more examples please?

specs/projectreuniondeployment/projectreuniondeployment.md
DrusTheAxe: What installer? How does it "do this"?

An app could do none of this in main() and rely on its installer to DoTheRightThing(TM) and main() just assumes the system 'environmental soup' is correct. In which case this sample is what the installer would do.

But...why do you assume there's an installer? 'XCOPY-install' apps don't have an installer. And some devs want their app to 'check their environmental soup' (up front or if/when things go wrong).

NSIS or ClickOnce could do this, if they so choose. But apps can too, and some very much want or need to.

specs/projectreuniondeployment/projectreuniondeployment.md
DrusTheAxe: Is that a typo? Should be "Windows updates"?

specs/projectreuniondeployment/projectreuniondeployment.md
DrusTheAxe: > the experience of informing the user should be a standard experience

Standard how? By new code in Windows, future, present _and past_? By ProjectReunion providing a 'standard UI'...which devs can (and often will, and should) ignore in favor of better integrating into their own user experience?

This seems like a good simple / poor-man's solution. For the 'more integrated' experience @stevenbrix and @dotMorten advocate why wouldn't you use GetInstallPackageLink() to programmatically handle it as you choose, through your own user experience? (Where 'you' = the app itself, the app's installer, an app's troubleshooter tool, etc)

DrusTheAxe:
> Last I heard, we were following the WinUI2.x model for shipping, which means that pre-release packages will ship the assemblies app-local, is this still the plan?

That's under active discussion

There's a strong intent to ship(1) as MSIX Framework packages. Supporting (in the technical as well as policy/sanctioned/official-Support-channels sense) 'app-local' aka the 'redist' model _in addition_ to Framework packages is begin debated. There's also discussions under way re versioning strategy, 'prerelease' and like topics. Important but complex topics so please be patient.

And if you have thoughts on the subject, please share your feedback. Input to this repository is monitored by many directly involved in these (and other) topics. The dialogue(s) here help us understand and shape our efforts.


(1) I assume you mean install/runtime, not dev/build-time. Nuget packages are the plan for the latter, at least for the initial release. Gotta start somewhere :-)

specs/projectreuniondeployment/projectreuniondeployment.md
dkbennett: I agree, though this is a special circumstance due to permissions. The app may not have permissions to install OS updates or other things required by Project Reunion on a downlevel device. This should be an at most 1 time deal for the typical user, and only on downlevel OSes. For future OSes, this can be mitigated by preinstalling Project Reunion, where it will be auto-updated by the store and required OS updates are already present.

specs/projectreuniondeployment/projectreuniondeployment.md
dkbennett: I generally agree with everything here, but there isn't a great solution to this problem. We have downlevel OS updates to deliver, and for packaged apps they will likely not have the capabilities to install other packages. The ideal solution is for the developer to provide the experience in their installer, if possible. If not possible (because they're packaged, for example), then there isn't much else they can do other than direct the user to where they can resolve the problem. For full-trust win32 programs, they could install much of it, and possibly even kick off KBs for the OS updates, but that stuff can change and will be different depending on platform, so its best for us to provide a better experience there. I didn't want to re-write Windows Update here, or re-write their applicability logic for determining which updates are going to be required, that's just going to lead to more problems. 

We have an escape hatch for any developer - link to install project reunion. For more privileged programs and packages and where OS updates are not required, we can get them an appinstaller link so they can kick off the update of the package themselves so the user doesn't have to get involved. But that's really above and beyond expectations for developers. The low barrier to entry and adoption for developers is to redirect to a link we control that can guide them through the process. This should be a 1-time deal, where the first Project Reunion consuming program or package does this, while all subsequent ones benefit from it having already been done before them so they have nothing to do.

specs/projectreuniondeployment/projectreuniondeployment.md
dkbennett: it should be Windows updates (I come from a world where Windows Update is a proper noun).  Anyway, the meaning to a developer is for one who wishes to provide a better user experience. If Windows Updates are required, it is unlikely the developer will be able to resolve that, but if Package updates are required, it is possible the developer could trigger an update using the appinstaller link. It's a crawl -> walk -> run experience. The crawl is the install instructions link. The walk is the developer triggering package updates by themselves for a better user experience. The run is none of this is required at all because it's already handled by the OS (which won't be possible on most downlevel OSes).

specs/projectreuniondeployment/projectreuniondeployment.md
dkbennett: Name is working and open to better suggestions. The reason this exists was defined at the top of the spec, but we may have OS updates required, we may have package updates required. We may not have the package at all. The developer needs to know and should be calling this API to check that all their expected Project Reunion dependencies are met and at a version minimum that they expect, otherwise their app will not function properly. So this API really has two functions:

1) Is Project Reunion the correct version and in a good state for what I need?  They call GetStatus() for this and provide the minimum version of Reunion they tested against.

2) If the status object returned is !IsOK(), what do I do about it? The obvious answer is their program should probably terminate and not try to continue without the dependency being satisfied, but the real question is how do they satisfy the dependency? There are two sub-options here for developers to resolve the problem.

a) They get the install instructions link and shell execute it and inform the user that there is action they need to take. While not great, this should be a 1-time deal for most situations. Sort of like needing to update DirectX or any other redist, but OS updates may be required here.

b) If OS updates are not required, and the are either a non-packaged medium-IL program, or a packaged app with Package Management capability, they can attempt to resolve the problem directly by getting the package (really appinstaller) install link and triggering the package install/update directly themselves.

specs/projectreuniondeployment/projectreuniondeployment.md
```diff
@@ -0,0 +1,138 @@
...
+
+```c#
+namespace Microsoft.Management.Deployment
+{
+    /// Represents the current Deployment status of Project Reunion
+    runtimeclass ProjectReunionDeploymentStatus
+    {
+        /// Returns True if Project Reunion is the desired Version and no updates or packages
+        /// are required.
+        Boolean IsOK { get; };
```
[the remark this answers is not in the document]
jonwis: This is returned from something, so I presume the OK-ness was computed during that operation.

specs/projectreuniondeployment/projectreuniondeployment.md
jonwis: `IsWindowsUpdateRequired` or `RequiresWindowsUpdate` maybe.

specs/projectreuniondeployment/projectreuniondeployment.md
jonwis: Should this be `PackageMissing` ?

Is the intent of the set of bools to be "here are multiple problems you need to resolve"?

specs/projectreuniondeployment/projectreuniondeployment.md
```diff
@@ -0,0 +1,138 @@
...
+
+        // Returns True if Windows Updates are required for Project Reunion.
+        Boolean WindowsUpdatesRequired { get; };
+
+        // Returns True if a Package install is required for Project Reunion.
+        Boolean PackageInstallRequired { get; };
+    }
+
+    /// Used to query deployment information for Project Reunion
+    static runtimeclass ProjectReunionDeployment
```
jonwis, line 121: Consider moving this up into `Microsoft.ProjectReunion` as the type `ProjectReunionPackage` maybe?

jonwis (changes requested):

specs/projectreuniondeployment/projectreuniondeployment.md
```diff
@@ -0,0 +1,138 @@
...
+Project Reunion bootstrapper will try to load the framework and error out with a similar
+instruction on how to resolve the missing dependencies Project Reunion.
+
+Developers must not assume that the Project Reunion main package is present, or that the version
+present is the minimum version required to support the feature set they expect. Instead, this API
+should be used to validate any Project Reunion requirements and direct the user to resolve any
+missing dependencies.
+
+The developer scenarios for this API is to answer the questions:
+1. Does this system have the min version of Project Reunion I require?
```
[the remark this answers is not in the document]
DrusTheAxe: Tip if you use VSCode, install the markdownlist extension and run it occasionally. It flags this sort of thing

specs/projectreuniondeployment/projectreuniondeployment.md
typicalzergling: The intent is to indicate what the problem is so they know whether they can potentially solve it. If it's a package missing or update is required, that is both the same resolution - install the latest version.  If it's a Windows Update that is required, it is unlikely that the developer will be able to resolve that on their own.

Missing is one possibility, but update is another. Of course you could consider not having the correct version also Missing, so I like that change, and it directly relates to the resolution action - it missing, so you need to add it.

Issue #157, opened 2020-08-27
Announcement: Proposed approach for windowing in Project Reunion


rkarman:
# Windowing in Reunion
We are excited to share some news with you today around the future of windowing for Windows developers. We have been listening to your feedback and concerns for this space over the years and we believe the approach we have chosen will address these in a positive way. This area has a lot of history (some of the features you use today, such as sizing and positioning windows, has been around since the first version of Win32) and a lot of investment in the developer community (your multi-windowed apps, your customized title bars, your customized dialogs and tool windows to create unique experiences, etc.) - we want to acknowledge and honor that as we go forward.

This post will outline our approach from a high-level perspective, our guiding principles, and our goals for the windowing space for Project Reunion. We are eager to get your feedback here, and we are open and ready to adjust our plans based on what you tell us. Nothing is set in stone - we are starting this journey here today, together with you.

## A unified windowing space
In the past we have had two very different ways of "doing windowing" for Win32 and UWP - one very powerful but complex to work with even for basic windowing scenarios; one very limited in capabilities but easy to achieve the basics with.

For UWP we have been in a constant state of "catching up" on core functionality, and never being able to. While for Win32 we have been in a state of non-innovation, leaving developers behind because we have focused on bringing new features only to UWP where we can guarantee that guardrails are in place.

We heard you - this situation is not making anyone happy and moving between the two worlds is hard.

With Project Reunion we are taking a bold stance - we want all Windows developers to have the power of Win32 windowing at their disposal, but we also want to provide easy to adopt APIs that can provide consistent experiences across apps, as well as easy to use APIs that lower the bar of entry for new developers.

We also want the windowing model to be fundamentally the same, so that we create a familiar way of working for developers regardless if you chose UWP or Win32 as your application model.

Our approach in order to achieve this? A layered set of APIs.

## Layers of APIs
We acknowledge that if you started with CreateWindowEx and used HWNDs for your app, you want to be able to continue to do so going forward. A re-write of your main window proc is just not going to be feasible. Therefore, we are bringing most of the windowing primitives from USER32 to Reunion. This gives you access to powerful APIs when you need them.

We also acknowledge that some things are really hard to do, or to get right for all situations, with the USER32 APIs, so we bring you AppWindow - a high-level windowing API surface that gives you access to a modern Windowing Model with an easier to use surface for areas where we have gotten a lot of feedback in the past, or where we want to help drive innovation, consistency in experience, or make it easier to adopt new features.

[...]

PR #158, opened 2020-08-28
DynamicDependency - Skeletal outline

DrusTheAxe:
This is not complete yet but it builds cleanly and has the right structure. Sharing as other devs working on similar things coming up right behind me are stumbling over some of the same pains (or don't realize yet how ProjectReunion.sln needs to be refined)

What's new:

\dev
- **Detours** - builds Detours as a static library as [Detours](https://github.com/Microsoft/Detours) provides no NuGet or like reasonably usable prebuilt library and compiling it requires very different settings than the rest of ProjectReunion_DLL. This gives us a 'firewall' so we can build Detours from source but with the same compiler settings as their makefiles do
- **DynamicDependency** - directory containing shared project providing Flat-C API, WinRT API and Detoured package graph APIs
- **Flat-C API** - Header and .def exports per spec. Source is skeletal and mostly not implemented yet; calls return E_NOTIMPL.
- **WinRT API** - IDL per spec. Implemented over Flat-C API.
- **Detoured Package Graph APIs** -- 
- **ProjectReunion_BoostrapDLL** - redist needed by non-packaged apps providing API to find and load Microsoft.ProjectReunion.dll from the ProjectReunion Framework package, so the DynamicDependency API can be called.
- **ProjectReunion_DLL** - References to DynamicDependency and Detours projects. DllMain() to detour package graph APIs and bootstrap access to the DynamicDependency API

\test
- **DynamicDependency** - root of DynamicDependency test code and data
- **data** - root for packages needed by DynamicDependency's tests
- **Framework.Math.Add** - Framework package providing a dll with an export for testing DynamicDependency API
- **Framework.Math.Multiply** - Framework package providing a dll with an export for testing DynamicDependency API
- **DynamicDependency_Test_Win32** - test project to exercise DynamicDependency Flat-C API

Plus some other random things (fixed up ARM[64] project snippets etc)

What's not present:

\dev
- **DynamicDependency Flat-C implementation**
- **DynamicDepedency\pch.h** - should not exist but won't build without. Added this as a temporary stopgap

\test
- **Bootstrap tests**
- **DynamicDependency Flat-C API tests**
- **DynamicDependency WinRT API tests**
- **More test packages**
- **data\makemsix.cmd** - creates .msix but doesn't sign it. Needs to sign. Reviewing microsoft-ui-xaml's scripts to steal shamelessly from :P

These will be addressed in upcoming PRs

Observations:
- **CppShared is too coarse**  One shared project per feature under \dev, but 1 shared project for all features under \test? Test should follow dev's model. As I've done for DynamicDepednency
- **makemsix.cmd is crude** - Low tech but effective. Still, feels like it should be better. We probably need a top level "tools" directory where we can put such things (like windows-ui-xaml has). After moving it for use by others we can refactor it to something better (as opposed to entirely blocking until I created this, crude as it may be)
- **SampleFlatC project should go away** - readme.md should be kept (with some updates) but moved elsewhere (out of the solution) as directions to future devs how to extend the solution
- **SampleWinRT project should go away** - same, readme.md is good (with some tweaks) but should move and rest goes away
- **CppTest_UWP** - delete RSN. Useful for a short while more as a template reference for new projects to spin up doing similar.
- **CppTest_Win32** - delete RSN.  Useful for a short while more as a template reference for new projects to spin up doing similar.
- **CppShared** - delete RSN. Obsolete once CppTest_UWP+Win32 are gone
- **Learn to edit *.vcxproj** - I'm finding it's all too common to need to edit projects, regardless if you create a new one via VS or clone from an existing similar one. Learn how to do both, it's a useful and all too necessary skill
- ** *.vcproj docs could be better** - VS docs are a bit thin and confusing when it comes to this level of hackery. microsoft-ui-xaml project's useful to help reverse engineer how to make VS do what you need but it's still time a lot of surfing the web and piecing together the magic
- **New VC Projects lack ARM|ARM64** - Manual editing *.vcproj required to add these
- **MakeMSIX Via UI or edit .vcproj** - you can add this via the UI but it's a lot easier to directly hack the .vcproj file adding the `<Target>` snippet. Defined generically via macros so simple copy/paste job
- **ProjectReunion_DLL OnePchToRuleThemAll** - this is likely to bloat as the project grows, impacting everyone's inner loop as we get ever more features cohabitating in the same project. Not a blocker yet but an area of concern to keep our eye on


jonwis:
Can we pull in detours as a submodule instead? What were the changes made to support Reunion in Detours?

dev/DynamicDependency/M.AM.DD.CreatePackageDependencyOptions.h
```diff
@@ -0,0 +1,35 @@
...
+
+#include "Microsoft.ApplicationModel.DynamicDependency.CreatePackageDependencyOptions.g.h"
+
+namespace winrt::Microsoft::ApplicationModel::DynamicDependency::implementation
+{
+    struct CreatePackageDependencyOptions : CreatePackageDependencyOptionsT<CreatePackageDependencyOptions>
+    {
+        CreatePackageDependencyOptions() = default;
+
+        Microsoft::ApplicationModel::DynamicDependency::PackageDependencyProcessorArchitectures Architectures();
```
jonwis, line 14: Do you need these prefixes? You're already in the `winrt::Microsoft::ApplicationModel::DynamicDependency` namespace here.

dev/DynamicDependency/M.AM.DD.PackageDependency.cpp
```diff
@@ -0,0 +1,89 @@
...
+    }
+    PackageDependency::PackageDependency(hstring const& id) :
+        m_id(id)
+    {
+    }
+    Microsoft::ApplicationModel::DynamicDependency::PackageDependency PackageDependency::GetFromId(hstring const& id)
+    {
+        return winrt::make<implementation::PackageDependency>(id);
+    }
+    Microsoft::ApplicationModel::DynamicDependency::PackageDependency PackageDependency::Create(hstring const& packageFamilyName, Windows::ApplicationModel::PackageVersion const& minVersion)
```
jonwis, line 31: This lack of newline between function-scope-close and next-line is throwing me off.  I guess at some point we'll figure out how to throw clang-format/clang-tidy at the source (@EHO-makai - something to add to our ES wishlist.)

dev/DynamicDependency/M.AM.DD.PackageDependency.cpp
```diff
@@ -0,0 +1,89 @@
...
+    {
+        const auto mddMinVersion = ::Microsoft::ApplicationModel::DynamicDependency::ToVersion(minVersion);
```
jonwis, line 33: These smell a lot like hungarian; can you just use `minVersion` and `architectures` and `lifetimeKind`, etc?

dev/DynamicDependency/M.AM.DD.PackageDependency.cpp
```diff
@@ -0,0 +1,89 @@
...
+        const auto mddArchitectures = MddPackageDependencyProcessorArchitectures::None;
+        const auto mddLifetimeKind = MddPackageDependencyLifetimeKind::Process;
+        const auto mddOptions = MddCreatePackageDependencyOptions::None;
+        wil::unique_process_heap_string packageDependencyId;
+        THROW_IF_FAILED(MddTryCreatePackageDependency(nullptr, packageFamilyName.c_str(), mddMinVersion, mddArchitectures, mddLifetimeKind, nullptr, mddOptions, wil::out_param(packageDependencyId)));
+        return GetFromId(packageDependencyId.get());
+    }
+    Microsoft::ApplicationModel::DynamicDependency::PackageDependency PackageDependency::Create(hstring const& packageFamilyName, Windows::ApplicationModel::PackageVersion const& minVersion, Microsoft::ApplicationModel::DynamicDependency::CreatePackageDependencyOptions const& options)
+    {
+        const auto mddMinVersion = ::Microsoft::ApplicationModel::DynamicDependency::ToVersion(minVersion);
```
jonwis, line 43: These three should probably call some worker internal method, instead of repeating the same code over and over?

dev/DynamicDependency/M.AM.DD.PackageDependency.cpp
```diff
@@ -0,0 +1,89 @@
...
+        return context;
+    }
+    Microsoft::ApplicationModel::DynamicDependency::PackageDependencyContext PackageDependency::Add(Microsoft::ApplicationModel::DynamicDependency::AddPackageDependencyOptions const& options)
+    {
+        const auto rank = options.Rank();
+        const auto mddOptions = ::Microsoft::ApplicationModel::DynamicDependency::ToAddOptions(options);
+        wil::unique_package_dependency_context packageDependencyContext;
+        THROW_IF_FAILED(MddAddPackageDependency(m_id.c_str(), rank, mddOptions, &packageDependencyContext, nullptr));
+        auto context = winrt::make<implementation::PackageDependencyContext>(packageDependencyContext.get());
+        packageDependencyContext.release();
```
jonwis, line 86: Use `move` here:

```
auto context = winrt::make<implementation::PackageDependencyContext>(std::move(packageDependencyContext));
```

This pattern is a signal you should be moving during ownership transfer:

```c++
auto x = thing_that_creates(context-thing.get());
context-thing.release();
```

dev/ProjectReunion_BootstrapDLL/IDynamicDependencyLifetimeManager.h
```diff
@@ -0,0 +1,171 @@
+
+
+/* this ALWAYS GENERATED file contains the definitions for the interfaces */
```
jonwis, line 3: Why's this getting committed?

dev/ProjectReunion_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -0,0 +1,103 @@
...
+IDynamicDependencyLifetimeManager* g_lifetimeManager = nullptr;
+DLL_DIRECTORY_COOKIE g_dllDirectoryCookie = 0;
+wil::unique_cotaskmem_string g_frameworkPath;
+
+STDAPI MddBootstrapInitialize(
+    const CLSID& appDynamicDependencyLifetimeManager)
+{
+    wil::com_ptr_nothrow<IDynamicDependencyLifetimeManager> lifetimeManager;
+    RETURN_IF_FAILED(CoCreateInstance(appDynamicDependencyLifetimeManager, nullptr, CLSCTX_LOCAL_SERVER, __uuidof(IDynamicDependencyLifetimeManager), reinterpret_cast<void**>(lifetimeManager.addressof())));
+
```
jonwis, line 21: Suggest using WIL's `wil::CoCreateInstance` and changing this (and other exports) to have the "function level try/catch" that lets you use throwing helpers directly.

dev/ProjectReunion_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -0,0 +1,103 @@
...
+#include "IDynamicDependencyLifetimeManager.h"
+
+STDAPI AddFrameworkToPath(PCWSTR  path);
+
+IDynamicDependencyLifetimeManager* g_lifetimeManager = nullptr;
+DLL_DIRECTORY_COOKIE g_dllDirectoryCookie = 0;
+wil::unique_cotaskmem_string g_frameworkPath;
+
+STDAPI MddBootstrapInitialize(
```
jonwis, line 16: Should there be any kind of a lock around this?

dev/ProjectReunion_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -0,0 +1,103 @@
...
+{
+    wil::com_ptr_nothrow<IDynamicDependencyLifetimeManager> lifetimeManager;
+    RETURN_IF_FAILED(CoCreateInstance(appDynamicDependencyLifetimeManager, nullptr, CLSCTX_LOCAL_SERVER, __uuidof(IDynamicDependencyLifetimeManager), reinterpret_cast<void**>(lifetimeManager.addressof())));
+
+    wil::unique_cotaskmem_string path;
+    RETURN_IF_FAILED(lifetimeManager->GetFrameworkPath(path.addressof()));
+
+    RETURN_IF_FAILED(AddFrameworkToPath(path.get()));
+
+    g_lifetimeManager = lifetimeManager.detach();
```
jonwis, line 27: Probably should add a `FAILFAST_IF(g_lifetimeManager)` here to shake out multiple initialization. Or make calling this method clearly idempotent.

dev/ProjectReunion_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -0,0 +1,103 @@
...
+    return S_OK;
+}
+
+STDAPI MddBootstrapShutdown()
+{
+    if (!g_lifetimeManager)
+    {
+        RETURN_IF_FAILED(g_lifetimeManager->Shutdown());
```
jonwis, line 35: Suggest `LOG_IF_FAILED()` here so the remainder of this method is nofail. (Close, release, free, and shutdown methods should generally return `void` as developers don't really have a recourse if it doesn't work.)

dev/ProjectReunion_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -0,0 +1,103 @@
...
+    wil::unique_cotaskmem_string path;
+    RETURN_IF_FAILED(lifetimeManager->GetFrameworkPath(path.addressof()));
+
+    RETURN_IF_FAILED(AddFrameworkToPath(path.get()));
+
+    g_lifetimeManager = lifetimeManager.detach();
+    return S_OK;
+}
+
+STDAPI MddBootstrapShutdown()
```
jonwis, line 31: STDAPI_(void) ?

dev/ProjectReunion_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    return S_OK;
+}
+
+STDAPI AddFrameworkToPath(PCWSTR frameworkPath)
+{
+    // Add the framework to the Loader's DllDirectory list
+    auto dllDirectoryCookie = AddDllDirectory(frameworkPath);
+    RETURN_HR_IF(HRESULT_FROM_WIN32(GetLastError()), dllDirectoryCookie == 0);
```
jonwis, line 69: `RETURN_LAST_ERROR_IF(dllDirectoryCookie)` ?

dev/ProjectReunion_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+    wil::scope_exit([dllDirectoryCookie] {
```
jonwis, line 71: Scope exits need to be assigned. This should be marked [[no_discard]] which would generate an error.

dev/ProjectReunion_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -0,0 +1,103 @@
...
+        }
+    }
+
+    return S_OK;
+}
+
+STDAPI AddFrameworkToPath(PCWSTR frameworkPath)
+{
+    // Add the framework to the Loader's DllDirectory list
+    auto dllDirectoryCookie = AddDllDirectory(frameworkPath);
```
jonwis, line 68: Consider contributing a `wil/win32_helpers.h` method like `wil::unique_dll_add_cookie AddDllDirectory(frameworkPath)` and related methods (ie: `HRESULT AddDllDirectoryNoThrow(PCWSTR path, wil::unique_dll_add_cookie& cookie)`, etc.)

dev/ProjectReunion_DLL/ProjectReunion_DLL.vcxproj
```diff
@@ -167,7 +168,7 @@
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
     <ClCompile>
       <PrecompiledHeader>Use</PrecompiledHeader>
-      <WarningLevel>Level3</WarningLevel>
+      <WarningLevel>Level4</WarningLevel>
```
jonwis, line 171: Thanks!

dev/ProjectReunion_DLL/dllmain.cpp
```diff
@@ -0,0 +1,23 @@
...
+        DisableThreadLibraryCalls(hmodule);
+    case DLL_THREAD_ATTACH:
+    case DLL_THREAD_DETACH:
+    case DLL_PROCESS_DETACH:
+        break;
+    }
+
+    // Give WIL a crack at DLLMain processing
+    // See DLLMain() in wil/result_macros.h for why
+    wil::DLLMain(hmodule, reason, reserved);
```
jonwis, line 20: Should this be `return wil::DLLMain(...)` here? Or does that fail?

dev/Detours/README.md
```diff
@@ -0,0 +1,3 @@
+This is a fork of Detours from https://github.com/microsoft/Detours
+
+Last forked: August'24, 2020 9:19 pm PDT
```
jevansaks, line 3: Can you put the commit you had checked out when you cloned it?

jevansaks:
> Can we pull in detours as a submodule instead? What were the changes made to support Reunion in Detours?

I'd like to avoid submodules, they're kind of a pain. I would have suggested nupkg of source or vcpkg -- looks like @DrusTheAxe had a conversation with detours folks about this (https://github.com/microsoft/Detours/issues/124). Sounds like someone made a vcpkg of detours but it requires that you have vcpkg available. A copy of the source until vcpkg is integrated into the VS install seems ok.

dev/DynamicDependency/M.AM.DD.CreatePackageDependencyOptions.h
DrusTheAxe: Good point. Old habits die hard and I was initially worried about ambiguities, but not a problem now

dev/DynamicDependency/M.AM.DD.PackageDependency.cpp
DrusTheAxe: LOL it throws me off too. But that's how C++/WinRT generates the skeletons so I rolled with it

Sure would be nice if  C++/WinRT generated the 'nicer' layout... Feel free to [file an ask on the C++/WinRT project](https://github.com/Microsoft/cppwinrt/issues)

dev/DynamicDependency/M.AM.DD.PackageDependency.cpp
DrusTheAxe: Ambiguity

PackageDependency PackageDependency::Create(hstring const& packageFamilyName, Windows::ApplicationModel::PackageVersion const& ***minVersion***)

We need 2 variables in the same scope - WinRT + Flat-C datatype
* Windows::ApplicationModel::PackageVersion vs PACKAGE_VERSION
etc

I used the mdd prefix to indicate they're the Flat-C types

I could have skipped the prefix in some cases but then we'd have a mix of no-prefix and prefix so this was clearer

I could have gone with a suffix, e.g. minVersion vs minVersionAsPackageVersion but that'd be confusing since that's the literal WinRT type name, and minVersionAsPACKAGE_VERSION is ugly as sin

jonwis: Namespace-wise, try this:

```c++
namespace wam = winrt::Windows::ApplicationModel;
namespace mdd = winrt::Microsoft::ApplicationModel::DynamicDependency;

wam::PackageVersion foo { ... };
mdd::PackageVersion bar { ... };
```

The `mdd-` variable prefix just seems off.

dev/ProjectReunion_BootstrapDLL/MddBootstrap.cpp
DrusTheAxe: RETURN_LAST_ERROR_IF +1

(dllDirectoryCookie) -1
"Wrong Type: bool/BOOL/BOOLEAN/boolean expected"

RETURN_LAST_ERROR_IF(dllDirectoryCookie == 0) is needed

dev/ProjectReunion_BootstrapDLL/MddBootstrap.cpp
DrusTheAxe: Doh! Thanks!

`This should be marked [[no_discard]]...`
'This' being wil's scope_exit?

dev/ProjectReunion_BootstrapDLL/MddBootstrap.cpp
DrusTheAxe: Yep. I plan to push some changes upstream to WIL once things stabilize here

dev/ProjectReunion_DLL/ProjectReunion_DLL.vcxproj
DrusTheAxe: I /Wall but got oodles of warnings about things we use. I plan to follow up with WIL, Detours, C++/WinRT and Windows teams later. For now, /W4 is as good as we can get. We can selectively enable more if we find some common or egregious issues.

dev/ProjectReunion_DLL/dllmain.cpp
DrusTheAxe: That returns void

wjk:
I have a comment to make on the source code organization discipline of this project. Having one vcxitems file in a subfolder of the dev directory per feature does not scale IMHO, and as #157 shows, this repo is going to get a whole lot bigger given enough time. While I understand the benefits of this approach (fewer files for apps to have to deploy), I would recommend that we use static libraries and the `/WHOLEARCHIVE` and `/OPT:ICF` linker flags instead. Do we really want component X to be able to include private headers from component Y? With vcxitems, that behavior is unavoidable. (Furthermore, I cannot see how the current repository organization in WinUI 3 will possibly survive the vast code dump that is the currently closed-source components of the WinUI NuGet package.) Were any other approaches considered?

DrusTheAxe:
> I have a comment to make on the source code organization discipline of this project... Were any other approaches considered?

A few approaches were considered but this seemed a good start. We're still learning which choices are effective and which seemed like a good idea at the time but... One of the tenets of the Project Reunion effort is to be more flexible than traditional development of Windows, which sometimes means rewriting things (like, say, project structure...). But we need to get that first hand experience to evaluate what's not working and pick a better path.

I share your reservations with the current structure (and I've got the bloody wounds to prove it...). I don't see some of the current lasting long term. I've already made some changes and I expect more to come (by myself and others), but...you gotta start somewhere. In the crawl/walk/run lifecycle we're still in the delivery room.

Your static lib suggestion is interesting. I can see a few other options, though the tradeoffs are a bit fuzzy yet. Maybe I need more bloodletting (or maybe I'm woozy from current blood loss <g>).

I'm aiming to get DynamicDependencies stood up and in master this month. After that, reviewing pain points and proposing changes, including project structure and other technical details, as well as non-technical process issues. I've made a note to include your comment in that list of things to re-evaluate.

The journey of a thousand miles starts with a single step. We're far closer to that 1st step than the 1,000th but we're working on it. Please keep the feedback coming, but also be patient as we work to build a better world. R(ome)union won't be built in a day :-)

    - Howard

P.S. Release configs should already have WholeProgramOptimization enabled (among other things). That should be true regardless of the source structure how we combine the bits into end files.

jonwis:
How much of this is real product work vs. ES work? I'm leery of merging something that looks done (your product work) but isn't. Could some of the ES work be moved out to another branch temporarily while that finishes up?

dev/DynamicDependency/M.AM.DD.CreatePackageDependencyOptions.h
jonwis: See comment elsewhere about namespace aliasing.

dev/ProjectReunion_BootstrapDLL/MddBootstrap.cpp
jonwis: Yes - https://github.com/microsoft/wil/blob/3c00e7f1d8cf9930bbb8e5be3ef0df65c84e8928/include/wil/resource.h#L543 says WI_NODISCARD already. Did you see an error without the assignment with /W4 set?

ProjectReunion.sln
```diff
@@ -40,27 +40,52 @@ Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "CppTest_Win32", "test\CppTe
...
+EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "Detours", "dev\Detours\Detours.vcxproj", "{D6BC25C5-1AA7-4C4A-A02C-B42DEDBFEA33}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "data", "data", "{0C534F12-B076-47E5-A05B-2A711233AC6F}"
+EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "Framework.Math.Add", "test\DynamicDependency\data\Framework.Math.Add\Framework.Math.Add.vcxproj", "{58E95711-A12F-4C0E-A978-C6B4A0842AC8}"
+EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "Framework.Math.Multiply", "test\DynamicDependency\data\Framework.Math.Multiply\Framework.Math.Multiply.vcxproj", "{66D0D8B1-FAF4-4C6A-8303-07F3BA356FE3}"
+EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "Main.Sidecar", "..\..\..\..\howard.000\Source\Repos\ProjectReunion\test\DynamicDependency\data\Main.Sidecar\Main.Sidecar.vcxproj", "{6C10A72D-AF30-4696-9551-2446B689AD92}"
```
jonwis, line 61: Absolute path snuck back in here

dev/Detours/README.md
```diff
@@ -0,0 +1,6 @@
+This is a fork of Detours from https://github.com/microsoft/Detours
+
+Last forked
+-----------
+commit  e9da64c72f293c0223f293f018067fe68fbdb801
```
jonwis, line 5: Per thread with @jevansaks - is there any hope of taking a source package here, rather than dumping the entire thing in place?  Even if we were to produce the package and host it somewhere as an artifact for reference by our build, for instance.

dev/DynamicDependency/PackageGraphNode.h
```diff
@@ -0,0 +1,71 @@
...
+#pragma once
+
+#include <appmodel.h>
+
+#include <MsixDynamicDependency.h>
+
+#include <wil/resource.h>
+#include <wil_msixdynamicdependency.h>
+
+namespace Microsoft
```
jonwis, line 13:
```c++
namespace Microsoft::ApplicationModel::DynamicDependency
{
}
```

maybe?

dev/DynamicDependency/PackageGraphNode.h
```diff
@@ -0,0 +1,71 @@
...
+        {
+            (void)LOG_IF_WIN32_BOOL_FALSE(RemoveDllDirectory(m_addDllDirectoryCookie));
+            m_addDllDirectoryCookie = 0;
+        }
+    }
+
+private:
+    wil::unique_package_info_reference m_packageInfoReference;
+    wil::unique_package_dependency_context m_context;
+    DLL_DIRECTORY_COOKIE m_addDllDirectoryCookie{};
```
jonwis, line 67: ... and in the future when you define a wil::unique_directory_cookie, you can _mostly_ eliminate the hand-authored methods (except for reset) in this type.

dev/ProjectReunion_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -0,0 +1,141 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#include "pch.h"
+
+#include "MddBootstrap.h"
+
+#include "IDynamicDependencyLifetimeManager.h"
+
+HRESULT GetFrameworkPackageInfoForPackage(PCWSTR packageFullName, const PACKAGE_INFO*& frameworkPackageInfo, wil::unique_cotaskmem_ptr<BYTE[]>& packageInfoBuffer);
```
jonwis, line 10: Optional: Consider deleting these and just reorganizing the code blocks so the helpers are defined first and the implementations are listed second.

dev/ProjectReunion_BootstrapDLL/MddBootstrap.cpp
jonwis: Oh also, check out wil's init_once support to make that even magicer.

dev/ProjectReunion_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -0,0 +1,141 @@
...
+        const auto hr = wil::TryGetEnvironmentVariableW(L"PATH", path);
+        if (SUCCEEDED(hr) && path)
+        {
+            const auto frameworkPathLength = wcslen(g_frameworkPath.get());
+            PCWSTR pathWithoutFrameworkPath = g_frameworkPath.get() + frameworkPathLength;
+            if (*pathWithoutFrameworkPath == L';')
+            {
+                ++pathWithoutFrameworkPath;
+            }
+            (void)LOG_IF_WIN32_BOOL_FALSE(SetEnvironmentVariableW(L"PATH", pathWithoutFrameworkPath));
```
jonwis, line 65: Doesn't this assume that only the MDD code is modifying %path% so it's trimming only the start/end? Maybe this should use std::wstring to manipulate the path string, and delete the first instance of the path it found.

Alternatively, I just noticed, does this assume a single target framework package? What if we wanted Reunion and Microsoft.AI to be peer packages? Could this code be reused?

dev/ProjectReunion_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -0,0 +1,141 @@
...
+        }
+    }
+}
+
+HRESULT GetFrameworkPackageInfoForPackage(PCWSTR packageFullName, const PACKAGE_INFO*& frameworkPackageInfo, wil::unique_cotaskmem_ptr<BYTE[]>& packageInfoBuffer)
+{
+    frameworkPackageInfo = nullptr;
+
+    wil::unique_package_info_reference packageInfoReference;
```
jonwis, line 74: This code's a little light on comments explaining what's happening, and there's multiple independent-ish blocks. Either factor into methods that have clear purposes, or put a little one-liner explaining the purpose of the block.

dev/ProjectReunion_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -0,0 +1,141 @@
...
+    else
+    {
+        const auto lastError = GetLastError();
+        RETURN_HR_IF(HRESULT_FROM_WIN32(lastError), lastError != ERROR_ENVVAR_NOT_FOUND);
+        RETURN_IF_WIN32_BOOL_FALSE(SetEnvironmentVariableW(L"PATH", frameworkPath));
+    }
+
+    g_frameworkPath = wistd::move(frameworkPathCopy);
+    g_dllDirectoryCookie = dllDirectoryCookie;
+    dllDirectoryCookie = 0;
```
jonwis, line 139: Use `on_exit.release()` to do this instead of a specific assignment.  Then you can remove the `if (dllDirectoryCookie)` check above.

test/CppShared/TestDynamicDependency.cpp
```diff
@@ -0,0 +1,34 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#include "pch.h"
+
+using namespace Microsoft::VisualStudio::CppUnitTestFramework;
+
+namespace ProjectReunionCppTest
+{
+    TEST_CLASS(DynamicDependenciesTest)
```
jonwis, line 10: ?

test/DynamicDependency/data/Main.Sidecar/winmain.cpp
```diff
@@ -0,0 +1,87 @@
...
+    {
+        return S_OK;
+    }
+
+    HRESULT __stdcall GetPackageFullName(/*[out, retval]*/ PWSTR* packageFullName)
+    {
+        *packageFullName = nullptr;
+
+        WCHAR fullName[PACKAGE_FULL_NAME_MAX_LENGTH + 1]{};
+        UINT32 fullNameLength = _countof(fullName);
```
jonwis, line 35: Guzak tells me this should be ARRAYSIZE as _countof is an unevenly-supported extension.

dev/ProjectReunion_BootstrapDLL/MddBootstrap.cpp
DrusTheAxe: It shows up if I remove the assignment

Warning	C4834	discarding return value of function with 'nodiscard' attribute	ProjectReunion_BootstrapDLL	C:\Users\howard\Source\Repos\ProjectReunion\dev\ProjectReunion_BootstrapDLL\MddBootstrap.cpp	109	

Probably didn't show earlier when we had /W3. All good now (and should stay that way)

ProjectReunion.sln
DrusTheAxe: Fixed. Thanks

dev/Detours/README.md
DrusTheAxe: We can offer a patch to the Detours project to make a nuget package. That'll require a .nuspec, an update to Detours' build pipeline (or creating one if they don't have?) and offering a feed

Worth doing...later. Rolling with this for now to unblock work. We can replace later with a Detours nuget when they create one (or we offer a patch doing it)

Are we tracking these sorts of TODO-vFuture tasks via github or ADO?

dev/DynamicDependency/PackageGraphNode.h
DrusTheAxe: Ooooh! I didn't know that was legal. Excellent! Thanks

dev/DynamicDependency/PackageGraphNode.h
DrusTheAxe: I've got wil_msixdynamicdependency.h as a growing collection of future WIL submissions so I added a wil::unique_dll_directory_cookie there for now and updated to use it

dev/ProjectReunion_BootstrapDLL/MddBootstrap.cpp
DrusTheAxe: `Alternatively` - The intent is to alter path just long enough to LoadLibrary(Microsoft.ProjectReunion.dll) from the PRframework package. That'll initialize the package graph, adding Reunion to it. There's a bit of a chicken-egg problem I haven't entirely settled out which way I want to handle it. I need to get more details working with tests before finalizing. For now this is enough to get a bootstrap UT working, so I can dive into the deeper DynDep waters. I expect to come back here once I get MddAddPackageDependnecy's UT running

dev/ProjectReunion_BootstrapDLL/MddBootstrap.cpp
DrusTheAxe: Solved by WIL-izing the cookie :-)

test/CppShared/TestDynamicDependency.cpp
DrusTheAxe: Leftover from an early start that didn't pan out. Didn't notice the file was still around as I'm living the VS dream and it's not in the VS project view.

Something we should keep in mind for later. Some kind of "xref filesystem vs .vcxproj and flag any files in the former not in the latter" tool at least. Would be useful to help spot garbage accidentally laying around. @EHO-makai do you have a running list of 'build process checks' we want to eventually have? Or a consistent place we can log such things for future action?

test/DynamicDependency/data/Main.Sidecar/winmain.cpp
DrusTheAxe: Makes sense. I regularly see folks do _countof (or other similar non-ARRAYSIZE). I'll point folks toward ARRAYSIZE going forward.

dev/Detours/README.md
DrusTheAxe: @EHO-makai do you know which is preferred github vs ADO Q for TODO-vFuture tasks?

DrusTheAxe:
> How much of this is real product work vs. ES work? I'm leery of merging something that looks done (your product work) but isn't. Could some of the ES work be moved out to another branch temporarily while that finishes up?

All of my work is real product work, as in product code, or test code to verify product code. That happens to include some ES-y things e.g. make new projects for my tests, with the right compiler settings etc. But those are means to my ends, only done out of necessity to finish my work.

Anything that smacks of straight ES I've largely been looking to see the fruits of @EHO-makai 's work :-) We'll eventually have some things we can refine, polish or simplify, and we are cross pollinating as we find problems the other's found solutions for. But so far we're pretty parallel/independent and focused on our lanes.

dev/DynamicDependency/M.AM.DD.CreatePackageDependencyOptions.h
DrusTheAxe: Hmmm. I don't I see a comment elsewhere. Are you suggesting e.g.

namespace Mdd
{
using namespace ::Microsoft::ApplicationModel::DynamicDependency;
}

and then changing
```
::Microsoft::ApplicationModel::DynamicDependency::PackageDependency GetFromId(hstring const& id);
```
to
```
Mdd::PackageDependency GetFromId(hstring const& id);
```

?

DrusTheAxe: Neat trick. Thanks

dev/Detours/README.md
jevansaks: For code that is living on GitHub I'd say we should track it on GitHub. That way there's even a chance an awesome community member might help us out!

dev/DynamicDependency/M.AM.DD.AddPackageDependencyOptions.h
```diff
@@ -0,0 +1,31 @@
...
+    {
+        AddPackageDependencyOptions() = default;
+
+        int32_t Rank();
+        void Rank(int32_t value);
+        bool PrependIfRankCollision();
+        void PrependIfRankCollision(bool value);
+
+    private:
+        int32_t m_rank = MDD_PACKAGE_DEPENDENCY_RANK_DEFAULT;
```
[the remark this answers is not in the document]
DrusTheAxe:
```
ES.23: Prefer the {}-initializer syntax

Reason Prefer {}. The rules for {} initialization are simpler, more general, less ambiguous, and safer than for other forms of initialization.
Use = only when you are sure that there can be no narrowing conversions. For built-in arithmetic types, use = only with auto.

Avoid () initialization, which allows parsing ambiguities.
```

Interesting. I hadn't seen that one before.

dev/DynamicDependency/M.AM.DD.PackageDependency.cpp
```diff
@@ -0,0 +1,105 @@
...
+        return Create(packageFamilyName.c_str(), mddMinVersion, mddArchitectures, mddLifetimeKind, mddOptions);
+    }
+
+    Microsoft::ApplicationModel::DynamicDependency::PackageDependency PackageDependency::CreateForSystem(hstring const& packageFamilyName, Windows::ApplicationModel::PackageVersion const& minVersion, Microsoft::ApplicationModel::DynamicDependency::CreatePackageDependencyOptions const& options)
+    {
+        const auto mddMinVersion = ::Microsoft::ApplicationModel::DynamicDependency::ToVersion(minVersion);
+        const auto mddArchitectures = ::Microsoft::ApplicationModel::DynamicDependency::ToArchitectures(options.Architectures());
+        const auto mddLifetimeKind = ::Microsoft::ApplicationModel::DynamicDependency::ToLifetimeKind(options.LifetimeArtifactKind());
+        auto mddOptions = ::Microsoft::ApplicationModel::DynamicDependency::ToCreateOptions(options, MddCreatePackageDependencyOptions::ScopeIsSystem);
+        wil::unique_process_heap_string packageDependencyId;
```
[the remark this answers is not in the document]
DrusTheAxe: Doh! I really wish VC would warn about such things. Is there a way to tag a class to indicate only firing its ctor+dtor is OK no reason to warn, and anything w/o that marking should generate a warning if not otherwise accessed?

dev/DynamicDependency/MddDetourPackageGraph.cpp
```diff
@@ -0,0 +1,108 @@
...
+    PackagePathType /*packagePathType*/,
+    UINT32* /*bufferLength*/,
+    BYTE* /*buffer*/,
+    UINT32* count)
+{
+    *count = 0;
+
+    auto lock(AcquirePackageGraphLock());
+
+#if 0
```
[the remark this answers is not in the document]
DrusTheAxe: More like temporarily commented out to get overall build clean and work on testing

Anything marked TODO or returning ERROR_NOT_SUPPORTED, E_NOTIMPL or throw hresult_not_implemented() are basically the same thing - "coming soon". Birthing pains :P I've added TODO markers to make this more apparent

dev/DynamicDependency/M.AM.DD.PackageDependency.cpp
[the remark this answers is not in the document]
jevansaks: I'm not sure. I'm not being as stylistically pedantic in this PR (trying to go easy on you :)) as I normally would be, but this is something you should avoid in general following [ES.21 Don't declare a variable until you have a value to initialize it with](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#es22-dont-declare-a-variable-until-you-have-a-value-to-initialize-it-with). I would rather flag this type of thing were it unused or not.

dev/Detours/README.md
DrusTheAxe: Done => https://github.com/microsoft/ProjectReunion/issues/178

test/CppShared/TestDynamicDependency.cpp
DrusTheAxe: Added issue #179 Add 'scan for unreferenced files' tool

dev/ProjectReunion_BootstrapDLL/IDynamicDependencyLifetimeManager.h
DrusTheAxe: It defines the interface that needs to be supported by the PackagedCOM OOP Server 'lifetime manager' in a per-app 'sidecar' helper package. Needed here as bootstrapping needs to CCI that object and use the interface to find the ProjectReunion Framework package so we can load Microsoft.ProjectReunion.dll for non-packaged processes

dev/ProjectReunion_BootstrapDLL/MddBootstrap.cpp
DrusTheAxe: I don't see a need. It's a 1-time call for a non-packaged process, typically in main() so the non-packaged process can access the DynamicDependency API

dev/ProjectReunion_BootstrapDLL/MddBootstrap.cpp
DrusTheAxe: I was going to use WIL but the CoCreateInstanceNoThrow loses error information
```
Note, failures are reported as a null result, the HRESULT is lost
```
and I've been avoiding using exceptions in the bootstrap code

dev/ProjectReunion_BootstrapDLL/MddBootstrap.cpp
DrusTheAxe: It's a 1-time thing for main() so I don't see multiple initialization. I've added FAIL_FAST traps at the start of init as a sanity check in case someone oddly enough calls this twice

init_once is a bit too magical here. Maybe refactor someday but I'd like to keep things a bit simpler for now as we try to get past the crawl stage

dev/DynamicDependency/M.AM.DD.PackageDependency.cpp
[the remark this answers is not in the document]
DrusTheAxe: No worries. Feel free to share. I can always disagree :P

I'd say I agree about avoiding that in general, except I feel more strongly about that practice :-) It's C++ after all not C (and a feature going back decades). I've very fond of just-in-time local declaration

In this case it was a copy/paste leftover from earlier code. Maybe I'll take a look if one of the /Wall warnings would have flagged this and, if so, see how badly things go if I explicitly enable that 1 warning

dev/DynamicDependency/M.AM.DD.CreatePackageDependencyOptions.h
```diff
@@ -0,0 +1,37 @@
...
+        bool VerifyDependencyResolution();
+        void VerifyDependencyResolution(bool value);
+        winrt::PackageDependencyLifetimeArtifactKind LifetimeArtifactKind();
+        void LifetimeArtifactKind(Microsoft::ApplicationModel::DynamicDependency::PackageDependencyLifetimeArtifactKind const& value);
+        hstring LifetimeArtifact();
+        void LifetimeArtifact(hstring const& value);
+
+    private:
+        winrt::PackageDependencyProcessorArchitectures m_architectures = winrt::PackageDependencyProcessorArchitectures::None;
+        bool m_verifyDependencyResolution = false;
```
jevansaks, line 27: Prefer the initializer syntax for in-class initializers (see earlier comment)

DrusTheAxe: Sure. I'll do a pass over the whole thing (while the code's still small). If I miss anything I can take care of it in the next PR (this isn't the end)

dev/DynamicDependency/MddDetourPackageGraph.cpp
```diff
@@ -44,28 +44,33 @@ HRESULT MddDetourPackageGraphInitialize()
     }
 
     // Detour package graph APIs to our implementation
-    DetourRestoreAfterWith();
-    DetourTransactionBegin();
-    DetourUpdateThread(GetCurrentThread());
-    DetourAttach(&(PVOID&)TrueGetCurrentPackageInfo, DynamicGetCurrentPackageInfo);
-    DetourTransactionCommit();
-
+    RETURN_IF_WIN32_BOOL_FALSE(DetourRestoreAfterWith());
```
jevansaks, line 47: Does our code function if these things return false? How do we expect callers to recover / handle a failing hresult versus us doing a failfast and having some semblance of failure context? (I would hope we can lean towards failfast on developer/unexpected error)

DrusTheAxe: All their examples show this. No error handling. Grrrr....

Updating to FailFast if any of this chokes. What they do is pretty unlikely to fail, and if it does we're well into weird territory so not much caller can reasonably do about it. FailFast seems the best fit. We can always change it later if we get feedback needing more nuanced behavior

DrusTheAxe:
PR is complete for review. Several places are marked with //TODO; will be addressed in upcoming PR.

Would like to get signoff and complete this PR in the 'commit early and commit often' approach. Future PRs should be more bite-size given this foundation (easier to review, easier to change, faster code velocity, yadda yadda)

dev/DynamicDependency/MsixDynamicDependency.cpp
```diff
@@ -0,0 +1,79 @@
...
+    MddPackageDependencyProcessorArchitectures /*packageDependencyProcessorArchitectures*/,
+    MddPackageDependencyLifetimeKind /*lifetimeKind*/,
+    PCWSTR /*lifetimeArtifact*/,
+    MddCreatePackageDependencyOptions /*options*/,
+    _Outptr_result_maybenull_ PWSTR* packageDependencyId)
+{
+    *packageDependencyId = nullptr;
+
+    //TODO: Implement MddTryCreatePackageDependency
+    RETURN_HR(E_NOTIMPL);
```
wjk, line 23: Ditto.

DrusTheAxe:
Branch is wonky given the shift to main. I'm going to abandon this PR and make a new one rebased against main

PR #165, opened 2020-09-02
Draft: Toast Notifications for Project Reunion

hulumane:
This spec covers Toasts for project reunion.

### DIFF
--- a/specs/notifications-api/notifications-api.md
+++ b/specs/notifications-api/notifications-api.md
@@ -54,7 +54,7 @@ First, developers would install the Reunion NuGet package.
 
 Then, developers construct a notification as usual, but call the Interop version of ToastNotificationManager to create the toast notifier. With that one change, toasts will work on Win32 without doing anything additional!
 
-```csharp
+```cs
 // Construct the content (same as today)
 ToastContent content = new ToastContentBuilder()
     .AddToastActivationInfo("picOfHappyCanyon", ToastActivationType.Foreground)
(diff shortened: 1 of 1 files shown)


andrewleader:
👍

specs/notifications-api/notifications-api.md
```diff
@@ -0,0 +1,328 @@
...
+1. Declaration for xmlns:desktop
+1. In the IgnorableNamespaces attribute, com and desktop
+1. com:Extension for the COM activator using a new GUID of your choice. Be sure to include the Arguments="-ToastActivated" so that you know your launch was from a toast
+1. desktop:Extension for windows.toastNotificationActivation to declare your toast activator CLSID (the GUID from #4 above).
+
+Then, **in your app's startup code** (WinMain), subscribe to the **OnActivated** event.
+
+```csharp
+
+int main(int argc, char* argv[], char* envp[])
```
wjk, line 109: Is this code C# or C++? That isn't what a main function in C# looks like, and WinMain is used in C++ apps only. If it's actually C#, it should be `Main(string[] args)`. Furthermore,  the `init_apartment()` call is C++-specific; to specify an STA apartment, put `[STAThread]` attribute on the Main method.

wjk (changes requested):

specs/notifications-api/notifications-api.md
hulumane: Ah this is c++ sample code. Not c#. Will change to update that.

hulumane: Good catch. I'll move this to c# syntax

Issue #175, opened 2020-09-04
Proposal: Easing and streamlining named object sharing between packaged and Win32


DefaultRyan:
# Easing and streamlining named object sharing between packaged and Win32

This is a proposal for a flat C API that takes a list of Package Family Names (PFNs) and an access mask, and returns a security descriptor

# Background

Named objects (mutexes, events, waitable timers) created by packaged applications must be ACLed during creation in order to be shared outside the package.

Currently, [our guidance](https://docs.microsoft.com/en-us/windows/uwp/communication/sharing-named-objects) for this scenario is to write a large amount of complex, cumbersome, and largely boilerplate code. ([See here for a taste](https://docs.microsoft.com/en-us/windows/win32/api/securityappcontainer/nf-securityappcontainer-getappcontainernamedobjectpath#examples)).

If we are to meet developers where they are, we need to reduce the friction of Win32 and UWP interop, and make it as easy as possible to share named objects.

# Description

Add a flat C API that takes an array of PFNs and an access mask, and produces the needed `SECURITY_ATTRIBUTES` struct needed by `Create`* APIs would address the most common scenarios and abstract away the complexity of building a security descriptor.

```c
STDAPI GetSecurityDescriptorForPackageFamilyNames(
    DWORD                 cCountOfPackageFamilyNames,
    const LPCWSTR*        pListOfPackageFamilyNames,
    DWORD                 accessMask,
    PSECURITY_DESCRIPTOR* ppSD
)
```

# Open questions

## C# API

Is there any alternative to using PInvoke for this API?

## WinRT API

I can't think of a good way to represent the resulting `PSECURITY_DESCRIPTOR` in the WinRT type system, and I'm not seeing any immediate value to providing an implementation of this in WinRT.

Jaiganeshkumaran:
@DefaultRyan Why not represent PSECURITY_DESCRIPTOR as a WinRT class?

smaillet-ms:
This seems reasonable with a few syntactical updates:

1. Don't use LPCWSTR, just use PCWSTR
2. Don't use the "P" prefixed typedefs unless absolutely necessary, they are confusing with respect to const/volatile qualifiers (e.g. the result isn't intuitively obvious)

I wonder if it's a bit too "simplified" though. Are there other ACE entries one might need or want? Should all packages listed, always get the same rights? What if I wanted to control the rights per package?

smaillet-ms:
[...]

PR #188, opened 2020-09-16
Create spec for sharing security descriptors

DefaultRyan:


DefaultRyan:
This is the spec for proposal #175.

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
```diff
@@ -0,0 +1,49 @@
...
+
+Named objects (mutexes, events, waitable timers) must be ACLed during creation in order to be shared with packaged applications.
+
+Currently, [our guidance](https://docs.microsoft.com/en-us/windows/uwp/communication/sharing-named-objects) for this scenario is to write a large amount of complex, cumbersome, and largely boilerplate code. ([See here for a taste](https://docs.microsoft.com/en-us/windows/win32/api/securityappcontainer/nf-securityappcontainer-getappcontainernamedobjectpath#examples)).
+
+If we are to meet developers where they are, we need to reduce the friction of Win32 and UWP interop, and make it as easy as possible to share named objects.
+
+# Description
+
+Add a flat C API that takes an array of PFNs and access masks, and produces the needed `SECURITY_ATTRIBUTES` object.
```
jonwis, line 15: Can you also add a WinRT wrapper variant that returns the SDDL string form as well? Someone using C# can then use that with things like [System.Security.AccessControl](https://docs.microsoft.com/en-us/dotnet/api/system.security.accesscontrol.objectsecurity.setsecuritydescriptorsddlform?view=dotnet-plat-ext-3.1#System_Security_AccessControl_ObjectSecurity_SetSecurityDescriptorSddlForm_System_String_) or storing?

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
```diff
@@ -0,0 +1,49 @@
...
+    result.create(wil::EventOptions::None, name, pSD.get());
+    return result;
+}
+```
+
+# API Details
+
+```c
+STDAPI GetSecurityDescriptorForPackageFamilyNames(
+    DWORD                 cCountOfPackageFamilyNames,
```
jonwis, line 41: Style - I think we're pushing to use `UINT32` rather than the SDK types.

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
```diff
@@ -0,0 +1,49 @@
...
+    _In_Reads_(cCountOfPackageFamilyNames)
+      const PCWSTR*         pListOfPackageFamilyNames,
+    _In_Reads_(cCountOfPackageFamilyNames)
+      const DWORD*          accessMask,
+    _Outptr_ SECURITY_DESCRIPTOR** ppSD
```
mikebattista, line 46: The type expected by CreateNamedPipe and friends is LPSECURITY_ATTRIBUTES. I think this API should return that type (which has the security descriptor inside of it). The main goal here is to make the simple majority case trivial and not require the developer to learn a complicated security model API or have to create a bunch of nested structs.

jonwis (approved):
Looking pretty good!

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
```diff
@@ -0,0 +1,49 @@
...
+```
+
+# API Details
+
+```c
+STDAPI GetSecurityDescriptorForPackageFamilyNames(
+    DWORD                 cCountOfPackageFamilyNames,
+    _In_Reads_(cCountOfPackageFamilyNames)
+      const PCWSTR*         pListOfPackageFamilyNames,
+    _In_Reads_(cCountOfPackageFamilyNames)
```
mikebattista, line 44: So if I want to grant 10 apps the same level of permissions, I have to create a list with 10 copies of the same permissions?

I think we should definitely support assigning the same permissions to multiple apps without having to list those permissions once for each app. Though also supporting a list of permissions for each app is nice as well to save the developer from having to write the complicated code in that scenario.

To support both, would that be two separate methods?

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
mikebattista: Nice. I like that bridge to support C# through SDDL strings.

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
jonwis: A SECURITY_ATTRIBUTES describes _how_ a security descriptor should be used.  It seems reasonable to add helper/wrapper methods like "CreateSharedEvent" or "SetSharedKeySecurity" or whatever, while retaining this primtive.

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
DefaultRyan: I've added an "API Notes" section that discusses the current API proposal, as well as two alternatives. I chose this one because it carried the smallest downside, in my opinion.

The alternatives either sacrificed valuable functionality (only accept a single permission), introduced complexity to the API (add a single permission, and make the array of permissions optional), or introduced a doc/naming/developer mental model penalty (two nearly identical APIs, one takes a single permission, the other takes an array of permissions). I believe the current proposal sacrifices the least, but I'm willing to listen to counterarguments.

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
DefaultRyan: UINT32 is still an SDK type. Do you mean the C/C++ standard `uint32_t`?

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
mikebattista: The main goal is to make it easy to generate the DACL and pass it to the right Create* function. Is creating the SECURITY_ATTRIBUTES struct trivial once you have the security descriptor? What does that code look like? Presumably you just set the security descriptor field, set inherit to false(?), and what do you set for nLength?

DefaultRyan: I've updated the example code to show this. It's pretty trivial
```c++
    SECURITY_ATTRIBUTES sa;
    sa.nLength = sizeof(SECURITY_ATTRIBUTES);
    sa.lpSecurityDescriptor = pSD.get();
    sa.bInheritHandle = FALSE;
```

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
mikebattista: I definitely don't like 2 separate methods if we can avoid it. Option 2 of having both arguments and using the list if it's provided doesn't seem that complicated to me, as long as that is a relatively standard design pattern for flat C APIs. C# would overload on type for this. Presumably this is the flat C equivalent of that?

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
DrusTheAxe: Yes. All the fashionable developers in Paris and Milan are using uint32_t :-)

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
mikebattista: Ok. I'm ok with that as long as we document this along with the API itself and how to use it.

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
```diff
@@ -0,0 +1,79 @@
...
+    wil::unique_event result;
+    result.create(wil::EventOptions::None, name, &sa);
+    return result;
+}
+```
+
+# API Details
+
+```c
+STDAPI GetSecurityDescriptorForPackageFamilyNames(
```
DrusTheAxe, line 45: ForPackageFamilyNames or ForAppContainerNames?

AppContainers have a Name which is run through an algorithm to produce an AppContainerSID. PackageSID is just an AppContainerSID that's been computed using a PackageFamilyName as the AppContainer Name.

AppContainer Name is a string, w/o constraint I think (no restriction on content or length). PackageFamilyName has defined rules (length and content) so all valid PackageFamilyNames are a subset of all valid AppContainerNames.

Do you intend to error if pListOfPackageFamilyNames contains "Kittens", "Foo$Bar", "LPT1_1234567890abc" or other invalid PackageFamilyNames? If so then your implementation should
```
#include <appmodel.h>

STDAPI GetSecurityDescriptorForPackageFamilyNames(...) noexcept try
{
...
for (DWORD index=0; index < cCountOfPackageFamilyNames; ++index)
{
    auto packageFamilyName = pListOfPackageFamilyNames[index];
    THROW_IF_FAILED(VerifyPackageFamilyName(packageFamilyName));
}
...
}
CATCH_RETURN();
```

Functionally this API could work for AppContainerNames equally well as PackageFamilyNames (other than this parameter validation). Do you want to call it Get...ForAppContainerNames()? Or 2 functions, Get...ForAppContainerNames() doing the work and Get...ForPackageFamilyNames() that does the parameter validation then return Get...ForAppContainerNames()?
This function can operate on AppContainerNames as easily as PackageFamilyNames.

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
```diff
@@ -0,0 +1,79 @@
...
+    DWORD                 cCountOfPackageFamilyNames,
+    _In_Reads_(cCountOfPackageFamilyNames)
+      const PCWSTR*         pListOfPackageFamilyNames,
```
DrusTheAxe, line 48: const?

`const PCWSTR*` or `PCWSTR*` or `const PCWSTR * const` or ?

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
```diff
@@ -0,0 +1,79 @@
...
+    result.create(wil::EventOptions::None, name, &sa);
+    return result;
+}
+```
+
+# API Details
+
+```c
+STDAPI GetSecurityDescriptorForPackageFamilyNames(
+    DWORD                 cCountOfPackageFamilyNames,
```
DrusTheAxe, line 46: NIT: countOfPackageFamilyNames. Or packageFamilyNamesCount

Regardless, no 'c' prefix. Hungarian Notation is soooo last century :P
Per [Coding-Guidelines.md](https://github.com/microsoft/ProjectReunion/blob/master/docs/Coding-Guidelines.md) follow cppcoreguidelines, specifically [NL.5: Avoid encoding type information in names](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rl-name-type)

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
DrusTheAxe, line 48: NIT: listOfPackageFamilyNames. Or just packageFamilyNames

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
```diff
@@ -0,0 +1,79 @@
...
+    _In_Reads_(cCountOfPackageFamilyNames)
+      const PCWSTR*         pListOfPackageFamilyNames,
+    _In_Reads_(cCountOfPackageFamilyNames)
+      const DWORD*          accessMask,
+    _Outptr_ SECURITY_DESCRIPTOR** ppSD
```
DrusTheAxe, line 51: NIT: rename ppSD to securityDescriptor

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
DrusTheAxe: Would this be equivalent
```
SECURITY_ATTRIBUTES sa{ sizeof(sa), pSD.get() };
```
?

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
jonwis: Don't use "const" in that way, it ties the hands of the implementer.  `const PCWSTR*` is sufficient.

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
EHO-makai: Howard, already wrote this, prefer sizeof(sa) over the named type (less error prone).

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
```diff
@@ -0,0 +1,79 @@
...
+Add a flat C API that takes an array of PFNs and access masks, and produces the needed `SECURITY_ATTRIBUTES` object.
+
+# Examples
+
+```c++
+#include <wil/resource.h>
+
+wil::unique_event CreateShareableEvent(PCWSTR name)
+{
+    wil::unique_hlocal_ptr pSD;
```
EHO-makai, line 24: no hungarian

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
```diff
@@ -0,0 +1,79 @@
...
+    PCWSTR packageNames[1] = { L"Contoso.Test.App_12345678" };
+    DWORD accessMasks[1] = { ACCESS_MODIFY_STATE };
+
+    THROW_IF_FAILED(GetSecurityDescriptorForPackageFamilyNames(
+        1, packageNames, accessMasks, &pSD);
+
+    SECURITY_ATTRIBUTES sa;
+    sa.nLength = sizeof(SECURITY_ATTRIBUTES);
```
EHO-makai, line 32: (sorry, duplicate comment), use sizeof(sa)

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
DrusTheAxe: Where do you see PackageFamilyName?
```
USERENVAPI HRESULT DeriveAppContainerSidFromAppContainerName(
  PCWSTR pszAppContainerName,
  PSID   *ppsidAppContainerSid
);
```

Yes, it accepts a PackageFamilyName because an AppContainerName is a string w/o constraint. All PackageFamilyNames are valid AppContainerNames...and so are a lot of values that aren't valid PackageFamilyNames.

GetSecurityDescriptorForPackageFamilyNames(..."Not.A$Package_Family@Name") should fail as that's not a valid package family name.

If the intent is to only support PackageFamilyName values then cool, please ensure the implementation validates its parameters.

If the intent is to support AppContainerNames then the function shouldn't validate parameters per `VerifyPackageFamilyName()` and it should be named *ForAppContainerNames().

mikebattista: What are the scenarios where AppContainerName would be used over the more specific PackageFamilyName?

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
DefaultRyan: Like Jon said.

`PCWSTR*` is bad, because I want to promise not to change the caller-supplied data. Especially since it could be constant.

`const PCWSTR* const` is overkill, because my implmenetation may want to modify the pointer (not the contents). This often happens, for example, when the callee iterates over an array.

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
DrusTheAxe: If someone's using AppContainer without appmodel

AppContainer was invented in Win8 and is a core Windows technology, like CreateProcess() and ReadFile().
The appmodel uses AppContainer and is probably the most popular user, but you can use AppContainer w/o the appmodel. And several folks do.

Anyone who used LowIL would be interested in AppContainer, as a more secure alternative. Edge, Office and other non-packaged apps who historically used LowIL have jumped on AppContainer or been looking how to incorproate and/or switch to it. These folks by definition use AppContainerNames since they're not packaged, but they may find this API helpful just as packaged apps may.

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
DefaultRyan: Sounds like a good idea. I'll work on adding this to the spec, but I'm guessing this is going to be under a new top-level namespace. What do you think that should be?

DrusTheAxe: Microsoft.Security.AccessControl

Quite a few [Windows.Security.* namespaces already exist](https://docs.microsoft.com/uwp/api/?view=winrt-19041) but nothing quite on point as AccessControl

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
mikebattista: Ok. And how do you retrieve your AppContainer name in those scenarios?

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
DefaultRyan: I like it. Not quite as sure if I have time to get to that part right now, or if I need to push that out a couple weeks.

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
DrusTheAxe: No retrieval. If you don't know your AppContainer name how'd you define it in the first place?

To create an AppContainer it needs to be defined to Windows via [CreateAppContainerProfile(PCWSTR pszAppContainerName, ...)](https://docs.microsoft.com/windows/win32/api/userenv/nf-userenv-createappcontainerprofile)

MSIX makes it convenient by always using PackageFamilyName, but if you subtract MSIX from the picture then you had to know your AppContainer name or you couldn't have create the AC profile in the first place.

mikebattista: Ok. I was thinking also how you determine the other app's AppContainer name you're sharing with. PackageFamilyName has other ways like Get-AppxPackage and manifests to query the system for the values. If the apps are coordinating though or built by the same person, then it's just exchanged offline I guess.

I'm ok making this *ForAppContainerNames in this case. I think that's more encompassing and future proof. Then we just doc as part of the API what names are supported and how to get them.

DefaultRyan: I'll update the API name.

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
DefaultRyan: While experimenting with the implementation, it turns out this only works if we also grant access at the user level. This means we also need an access mask for the user in the custom case.

I also want to tweak this, because when we pass two separate arrays, the caller needs to ensure they are both the same length. A more robust design would pass an array of `struct { UINT32 accessMask; PCWSTR name };`.

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
DefaultRyan: Definitely don't have time to do it right now, since we still seem to be working out exactly how the different features will be organized. I've opened #201 to track the WinRT API, and will come back to this once we have a little more infrastructure in place.

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
mikebattista: Could you clarify how this changes the API?

DefaultRyan: The API description and usage example have already been updated. Could you help me understand what you want in addition to that?

specs/GetSecurityDescriptorForAppContainerNames/GetSecurityDescriptorForAppContainerNames.md
```diff
@@ -0,0 +1,86 @@
...
+```c
+struct AppContainerAccess
+{
+    PCWSTR appContainerName;
+    uint32_t accessMask;
+};
+
+STDAPI GetSecurityDescriptorForAppContainerNames(
+    uint32_t countOfAppContainerNames,
+    _In_reads_(countOfPackageFamilyNames)
```
mikebattista, line 54: This needs to be renamed right?

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
mikebattista: I see that now. We never resolved the discussion on one access mask provided/applied to multiple apps (and now also the user), and this design takes us further from that. Are we giving up on simplifying that scenario with the new design?

specs/GetSecurityDescriptorForAppContainerNames/GetSecurityDescriptorForAppContainerNames.md
DefaultRyan: Good catch.

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
DefaultRyan: Yeah, I've abandoned that idea. For the per-app access masks to be optional, the caller needs to pass two arrays (instead of a single array of name/mask pairs).
1. This provides opporunities for callers to get it wrong and pass different-sized arrays.
2. If the array size is determined at runtime, that also means two separate heap allocations.

To me, this outweighs the benefits of having a slightly simpler calling pattern; the sample code hardly suffers from omitting the "simple" flavor.

specs/GetSecurityDescriptorForAppContainerNames/GetSecurityDescriptorForAppContainerNames.md
```diff
@@ -0,0 +1,86 @@
...
+    THROW_IF_FAILED(GetSecurityDescriptorForAppContainerNames(
+        1, access, EVENT_MODIFY_STATE | SYNCHRONIZE, &sd);
+
+    SECURITY_ATTRIBUTES sa;
+    sa.nLength = sizeof(sa);
+    sa.lpSecurityDescriptor = sd.get();
+    sa.bInheritHandle = FALSE;
+
+    wil::unique_event result;
+    result.create(wil::EventOptions::None, name, &sa);
```
mikebattista, line 38: Could you show calling something like CreateEvent directly? Want to make sure there's nothing WIL-biased here.

specs/GetSecurityDescriptorForPackageFamilyNames/GetSecurityDescriptorForPackageFamilyNames.md
mikebattista: Ok. The concern was always granting access to multiple apps, not just one, so the sample code for 1 app wouldn't really suffer in any case. Though I do agree that coupling an AppContainerName to an access mask with a struct has its advantages so am ok with this.

specs/GetSecurityDescriptorForAppContainerNames/GetSecurityDescriptorForAppContainerNames.md
```diff
@@ -0,0 +1,87 @@
...
+    // aka, ::CreateEvent(&sa, FALSE, FALSE, name);
+    return result;
+}
+```
+
+# API Details
+
+```c
+struct AppContainerAccess
```
ptorr-msft, line 47: Name is too generic; there are lots of ways AppContainers are granted "access" to things (e.g. capabilities). Suggest `AppContainerSecurityDescriptorInfo` (it's bad, but can start the conversation :)). Also do we want to tie this to "AppContainer names" or should we use a more generic term? (Today it's only used for stings-that-name-AppContainers, but in theory it can be used for any string).

specs/GetSecurityDescriptorForAppContainerNames/GetSecurityDescriptorForAppContainerNames.md
```diff
@@ -0,0 +1,87 @@
...
+{
+    PCWSTR appContainerName;
+    uint32_t accessMask;
+};
+
+STDAPI GetSecurityDescriptorForAppContainerNames(
+    uint32_t countOfAppContainerNames,
```
ptorr-msft, line 54: Overly long, also doesn't match the other parameter name (would be `countOfappAccess` if using this naming pattern).

specs/GetSecurityDescriptorForAppContainerNames/GetSecurityDescriptorForAppContainerNames.md
```diff
@@ -0,0 +1,87 @@
...
+    _In_reads_(countOfAppContainerNames)
+        const AppContainerAccess* appAccess,
+    uint32_t userAccessMask,
```
ptorr-msft, line 57: Not clear how this is used? Each struct has its own access mask.

specs/GetSecurityDescriptorForAppContainerNames/GetSecurityDescriptorForAppContainerNames.md
```diff
@@ -0,0 +1,87 @@
...
+# API Notes
+
+Some alternative designs were considered.
+- Single method, takes a list of permissions corresponding to the list of PFNs. (Current choice)
+  - Pros
+    - Handles both complicated case and simple case in a single call.
+    - API is straightforward, no extra/optional/multimode parameters.
+    - Only a single method to search/remember/code.
+  - Cons
+    - Simple case is not "optimally" simple - requires filling array with a single value.
```
oldnewthing, line 73: Sorta-simple version:

```cpp
THROW_IF_FAILED(GetSecurityDescriptorForAppContainerNames(
    1, std::array<AppContainerAccess, 1>{ { L"Contoso.Test.App_12345678", EVENT_MODIFY_STATE | SYNCHRONIZE } }.data(),
    EVENT_MODIFY_STATE | SYNCHRONIZE,
   &sd));
```

However you still wouldn't want to do this because it creates a temporary at runtime.

```cpp
```cpp
THROW_IF_FAILED(GetSecurityDescriptorForAppContainerNames(
    1, std::array<AppContainerAccess, 1>{ { L"Contoso.Test.App_12345678", EVENT_MODIFY_STATE | SYNCHRONIZE } }.data(),
    EVENT_MODIFY_STATE | SYNCHRONIZE,
   &sd));
```

This does build a temporary at runtime, though.

```cpp
static constexpr AppContainerAccess access{ L"Contoso.Test.App_12345678", EVENT_MODIFY_STATE | SYNCHRONIZE };

THROW_IF_FAILED(GetSecurityDescriptorForAppContainerNames(
    1, &access,
    EVENT_MODIFY_STATE | SYNCHRONIZE,
   &sd));
```

specs/GetSecurityDescriptorForAppContainerNames/GetSecurityDescriptorForAppContainerNames.md
oldnewthing: Took me a while to figure this out too. This last one is the access mask for medium IL.

specs/GetSecurityDescriptorForAppContainerNames/GetSecurityDescriptorForAppContainerNames.md
```diff
@@ -0,0 +1,87 @@
...
+  - Pros
+    - Handles both complicated case and simple case in a single call.
+    - Only a single method to search/remember/code.
+    - Provides a way to assign a single permission to all PFNs without filling an array.
+  - Cons
+    - Slightly more complicated API. Two parameters are conditionally used for the same purpose.
+- Two methods, one that takes a list of permissions, and one that takes a single permission.
+  - Pros
+    - Each API is straightforward, no extra/optional/multimode parameters.
+    - Dedicated APIs for assigning a single permission vs different permissions.
```
oldnewthing, line 84: I'm okay with adding a second method later if necessary. It would be called `GetSecurityDescriptorForAppContainerName` (singular).

specs/GetSecurityDescriptorForAppContainerNames/GetSecurityDescriptorForAppContainerNames.md
```diff
@@ -0,0 +1,87 @@
...
+
+wil::unique_event CreateShareableEvent(PCWSTR name)
+{
+    wil::unique_hlocal_security_descriptor sd;
+    AppContainerAccess access[1] =
+      {{L"Contoso.Test.App_12345678",
+        EVENT_MODIFY_STATE | SYNCHRONIZE}};
+
+    THROW_IF_FAILED(GetSecurityDescriptorForAppContainerNames(
+        1, access, EVENT_MODIFY_STATE | SYNCHRONIZE, &sd);
```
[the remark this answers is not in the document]
jonwis: SAMPLE: Fix

specs/GetSecurityDescriptorForAppContainerNames/GetSecurityDescriptorForAppContainerNames.md
```diff
@@ -0,0 +1,87 @@
...
+
+    SECURITY_ATTRIBUTES sa;
```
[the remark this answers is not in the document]
jonwis: SAMPLE: Fix

specs/GetSecurityDescriptorForAppContainerNames/GetSecurityDescriptorForAppContainerNames.md
jonwis: RECOMMEND: `AppContainerNameAndAccess`

specs/GetSecurityDescriptorForAppContainerNames/GetSecurityDescriptorForAppContainerNames.md
jonwis: RECOMMEND: `accessRequestCount`, then `accessRequests`

specs/GetSecurityDescriptorForAppContainerNames/GetSecurityDescriptorForAppContainerNames.md
jonwis: DISCUSSION: Sample code on MSDN fetches a PSID and then uses that with this parameter to say "that sid has this access."  The security system effectively makes two passes - one for the low-IL user token, one for the app container identity. It's important to have 

RECOMMEND: Add parameter which is a `PSID principal` and change to `uint32_t principalAccessMask` ... the caller can use one of the WIL helpers (or their own thing) to fetch a `PSID`. Accept `nullptr` to mean "the principal of the current thread/process."

specs/GetSecurityDescriptorForAppContainerNames/GetSecurityDescriptorForAppContainerNames.md
jonwis: DISCUSSION: Neato!

specs/GetSecurityDescriptorForAppContainerNames/GetSecurityDescriptorForAppContainerNames.md
jonwis: DISCUSSION: Agreed, we'll ask the customers

specs/GetSecurityDescriptorForAppContainerNames/GetSecurityDescriptorForAppContainerNames.md
DefaultRyan: Is there any distinction between the principal of the current thread, and the principal of the current process? In other words, can they be different, and if so, which should we use?

jonwis: You should use the one from the current thread, which enables this API to be used correctly on threads that do impersonation (like services and OOP COM and brokers.)

PR #189, opened 2020-09-16
Fix unit tests in reunion repo

axelandrejs:


dev/MRTCore/mrt/Core/unittests/UnitTest.vcxproj
```diff
@@ -135,6 +135,12 @@
       <SubSystem>Windows</SubSystem>
       <AdditionalLibraryDirectories>$(VCInstallDir)UnitTest\lib;%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
     </Link>
+    <PostBuildEvent>
```
huichen123, line 138: nit: is it possible to do it once, instead for each flavor?

dev/MRTCore/mrt/Core/unittests/UnitTest.vcxproj
```diff
@@ -135,6 +135,12 @@
...
+      <Command>
+        copy $(TargetDir)\..\MRM\mrm.dll $(TargetDir)
```
huichen123, line 140: $(TargetDir)..\mrm\mrm.dll?

dev/MRTCore/mrt/Core/unittests/UnitTest.vcxproj
axelandrejs: I tried making it its own property group, but that does not work.

We can coalesce the entire section into one. Most entries are identical, so we could only fork the few that are different. I considered doing that. The problem is that you're fighting against VS when you do so. Once you make changes in the GUI it will once again break it out. We are probably not going to rev this a lot so we can consider.

dev/MRTCore/mrt/Core/unittests/UnitTest.vcxproj
axelandrejs: I prefer this one as the other looks weird:)

huichen123: luckily \\ is tolerated in VS.

huichen123 (approved):

dev/MRTCore/mrt/Core/unittests/UnitTest.vcxproj
axelandrejs: Yep, we actually have that in a bunch of places. If you follow the entire compile spew you see a number of //s

---
In reply to: [489590532](https://github.com/microsoft/ProjectReunion/pull/189#discussion_r489590532) [](ancestors = 489590532)

huichen123: unfortunately project reunion's directory.build.props file sets a bad example.

axelandrejs: On second thought I am going to fix it. Let's not propagate incorrect things.

PR #196, opened 2020-09-17
Changed ResourceMap API to follow WinRT conventions.

ptorr-msft:
Made the `ResourceContext` parameter be the second parameter in the overloads, not the first.

Note: the unit tests run, but for various reasons I can't verify that the two test apps are working correctly. (VS refuses to open the managed project, and the native project needs the NuGet feed to be available and I can't figure out how to build a local version).


huichen123 (approved):

huichen123:
we need to keep github and mscodehub in sync for now

axelandrejs (approved):

PR #197, opened 2020-09-18
add MRTCore to CI pipeline

huichen123:


axelandrejs (approved):

dev/MRTCore/azure-pipelines.yml
```diff
@@ -207,16 +204,17 @@ jobs:
...
   dependsOn:
-    - Build
+    - BuildMRT
   pool:
     vmImage: 'windows-latest'
   
   variables:
     solution: '**/*.sln'
     appxPackageDir: '$(build.artifactStagingDirectory)\AppxPackages\\'
+    codeSign: false
```
EHO-makai, line 217: if I understand this correctly, you are not currently signing and in a subsequent PR we will turn on signing and nuget publish, is that correct?

huichen123: signing is turned on in MSCodeHub. There is no signing in github as no PackgeES support there.

EHO-makai: OK, can you leave a short comment here to that effect?  We'll try to get packageES signing turned on in the near future.

EHO-makai (approved):

dev/MRTCore/azure-pipelines.yml
huichen123: the comment may not be appropriate. we will cleanup the yaml file a bit. In MSCodeHub, we share the same yaml file for CI and release. However, on github, same yaml file is inappropriate. e.g. the PackageES tasks are invalid here.

PR #198, opened 2020-09-18
Add API spec for DWriteCore.

niklasb-ms:


specs/dwritecore/DWriteCore.md
```diff
@@ -0,0 +1,255 @@
...
+create a _restricted factory_ if it does not want the factory or objects created from it to read from or write
+to any cross-process or persistent cache. In addition, a restricted factory's system font collection only includes
+well-known system fonts.
+
+The following example creates a restricted factory.
+
+```c++
+ComPtr<IDWriteFactory7> spFactory;
+HRESULT hr = DWriteCreateFactory(
+    DWRITE_FACTORY_TYPE_RESTRICTED,
```
oldnewthing, line 134: The term "restricted" is ambiguous because it could mean

1. Access to this factory is restricted. Only super-powerful apps get access to it. (Compare "restricted capabilities" which are capabilities that only super-powerful apps get access to.)
2. This factory's actions are restricted. Even super-weak apps can use it.

This is a restriction of type 2. Maybe we can use a less ambiguous term.

The term "isolated" is already being used to mean "no cross-process shared state or persistent cache". This is an extension of "isolated" to limit the font enumeration, so maybe call it "isolated2"?

I think "isolated2" is better than trying to find another adjective. Most of the adjectives are just synonyms of each other, so the distinction is meaningless. Like, what's the difference between "restricted', "isolated", or "constrained"?

specs/dwritecore/DWriteCore.md
```diff
@@ -0,0 +1,255 @@
...
+    __uuidof(IDWriteFactory7),
+    (IUnknown**)&spFactory
+    );
+```
+
+## Drawing glyphs to a system memory bitmap
+
+An application creates a bitmap render target by calling [IDWriteGdiInterop::CreateBitmapRenderTarget](https://docs.microsoft.com/en-us/windows/win32/api/dwrite/nf-dwrite-idwritegdiinterop-createbitmaprendertarget).
+A bitmap render target encapsulates a bitmap in system memory, and eanbles rendering glyphs to the bitmap. On Windows,
```
oldnewthing, line 143:
```suggestion
A bitmap render target encapsulates a bitmap in system memory, and enables rendering glyphs to the bitmap. On Windows,
```

specs/dwritecore/DWriteCore.md
```diff
@@ -0,0 +1,255 @@
...
+the bitmap is actually a GDI DIB selected into a memory HDC, and it is possible to get the bitmap pixels using GDI
+functions. However, DWriteCore introduces a simpler method through the new IDWriteBitmapRenderTarget2 interface.
+The following example shows both the old way and the new way.
```
oldnewthing, line 146: The structure name `DWRITE_BITMAP_DATA_BGRA32` tells me that the pixel format is BGRA32. But what about the other attributes? From reading the documentation for `IDWriteBitmapRenderTarget::GetDC`, I infer that the underlying bitmap data returned by `DWriteBitmapRenderTarget2::GetBitmapData` is

1. A top-down bitmap,
2. with no padding between scan lines.

Should be documented more clearly what format to expect.

specs/dwritecore/DWriteCore.md
```diff
@@ -0,0 +1,255 @@
...
+        if (FAILED(hr))
+        {
+            throw Exception{ hr };
+        }
+    }
+    else
+    {
+        // We're using an older version that doesn't implement IDWriteBitmapRenderTarget2,
+        // so we have to get the bitmap by going through GDI. First get the bitmap handle.
+        HDC hdc = renderTarget_->GetMemoryDC();
```
oldnewthing, line 169:
```suggestion
        HDC hdc = renderTarget->GetMemoryDC();
```

specs/dwritecore/DWriteCore.md
```diff
@@ -0,0 +1,255 @@
...
+
+        // Call a GDI function to fill in the DIBSECTION structure for the bitmap.
+        DIBSECTION dib;
+        if (!GetObjectW(dibHandle, sizeof(dib), &dib))
+        {
+            throw Exception{ HRESULT_FROM_WIN32(GetLastError()) };
+        }
+
+        result.width = dib.dsBm.bmWidth;
+        result.height = dib.dsBm.bmHeight;
```
oldnewthing, line 184: Top-down DIBs report their height as negative. We need to flip the sign before returning it. Otherwise, a 10-pixel tall bitmap is reported as 4 billion+ pixels tall.

```suggestion
        result.height = 0U-dib.dsBm.bmHeight;
```

specs/dwritecore/DWriteCore.md
```diff
@@ -0,0 +1,255 @@
...
+        HDC hdc = renderTarget_->GetMemoryDC();
+        HGDIOBJ dibHandle = GetCurrentObject(hdc, OBJ_BITMAP);
+        if (dibHandle == nullptr)
+        {
+            throw Exception{ HRESULT_FROM_WIN32(GetLastError()) };
+        }
+
+        // Call a GDI function to fill in the DIBSECTION structure for the bitmap.
+        DIBSECTION dib;
+        if (!GetObjectW(dibHandle, sizeof(dib), &dib))
```
oldnewthing, line 178: GetObjectW can return a short buffer if the underlying bitmap is not a `DIBSECTION` but just a plain `BITMAP`.

```suggestion
        if (GetObjectW(dibHandle, sizeof(dib), &dib) != sizeof(dib))
```

specs/dwritecore/DWriteCore.md
```diff
@@ -0,0 +1,255 @@
...
+{
+    DWRITE_BITMAP_DATA_BGRA32 result = {};
+
+    // Query the bitmap render target for the new interface.
+    ComPtr<IDWriteBitmapRenderTarget2> renderTarget2;
+    HRESULT hr = renderTarget->QueryInterface(&renderTarget2);
+    if (SUCCEEDED(hr))
+    {
+        // IDWriteBitmapRenderTarget2 exists, so we can get the bitmap the easy way.
+        hr = renderTarget2->GetBitmapData(OUT &result);
```
oldnewthing, line 159: Please don't abuse SAL annotations like this.

specs/dwritecore/DWriteCore.md
```diff
@@ -0,0 +1,255 @@
...
+## Drawing glyphs to a system memory bitmap
+
+An application creates a bitmap render target by calling [IDWriteGdiInterop::CreateBitmapRenderTarget](https://docs.microsoft.com/en-us/windows/win32/api/dwrite/nf-dwrite-idwritegdiinterop-createbitmaprendertarget).
+A bitmap render target encapsulates a bitmap in system memory, and eanbles rendering glyphs to the bitmap. On Windows,
+the bitmap is actually a GDI DIB selected into a memory HDC, and it is possible to get the bitmap pixels using GDI
+functions. However, DWriteCore introduces a simpler method through the new IDWriteBitmapRenderTarget2 interface.
+The following example shows both the old way and the new way.
+
+```c++
+DWRITE_BITMAP_DATA_BGRA32 TextRenderer::GetBitmapData(IDWriteBitmapRenderTarget* renderTarget)
```
keith-horton, line 149: do we want to express SAL annotations for raw pointers, like specifying this as _In_ ?
presumably they work fine with other compilers?

specs/dwritecore/DWriteCore.md
```diff
@@ -0,0 +1,255 @@
...
+    }
+    else
+    {
+        // We're using an older version that doesn't implement IDWriteBitmapRenderTarget2,
+        // so we have to get the bitmap by going through GDI. First get the bitmap handle.
+        HDC hdc = renderTarget_->GetMemoryDC();
+        HGDIOBJ dibHandle = GetCurrentObject(hdc, OBJ_BITMAP);
+        if (dibHandle == nullptr)
+        {
+            throw Exception{ HRESULT_FROM_WIN32(GetLastError()) };
```
keith-horton, line 173: can we use wil directly from ProjectReunion? would simplify much of this COM and exception work vs. recreating similar types.

specs/dwritecore/DWriteCore.md
keith-horton: we use SAL annotation later (like _Field_size_), so should probably use them consistently.

specs/dwritecore/DWriteCore.md
```diff
@@ -0,0 +1,255 @@
...
+/// Encapsulates a bitmap which can be used for rendering glyphs.
+/// </summary>
+DWRITE_BEGIN_INTERFACE(IDWriteBitmapRenderTarget2, "C553A742-FC01-44DA-A66E-B8B9ED6C3995") : IDWriteBitmapRenderTarget1
+{
+    /// <summary>
+    /// Gets the demensions and a pointer to the system memory bitmap encapsulated by this
+    /// bitmap render target object. The pointer is owned by the render target object, and
+    /// remains valid as long as the object exists.
+    /// </summary>
+    STDMETHOD(GetBitmapData)(_Out_ DWRITE_BITMAP_DATA_BGRA32* bitmapData) PURE;
```
keith-horton, line 252: probably want to avoid "PURE". if we want to express this as a pure-virtual method, should just = 0;
unless there is some precedent for "PURE" that I've never seen :)

specs/dwritecore/DWriteCore.md
```diff
@@ -0,0 +1,255 @@
...
+ - Add a reference to the Project Reunion package.
+ - Include dwrite_core.h instead of dwrite_3.h.
+ - Link to DWriteCore.lib instead of DWrite.lib.
+
+In return, the application gets the benefits of Project Reunion, namely, access to the newest APIs
+and functionality regardless of what version of Windows your customer is running.
+
+# Background
+
+<!-- TEMPLATE
```
jonwis, line 19: Template stuff needs to be deleted. I'll update the template to indicate as such.

specs/dwritecore/DWriteCore.md
jevansaks: Yes, we can use the [public WIL](https://github.com/microsoft/wil)

specs/dwritecore/DWriteCore.md
```diff
@@ -0,0 +1,255 @@
...
+
+## API Versioning
+
+The DirectWrite API has grown over time, and many APIs have multiple versions: IDWriteFactory, 
+IDWriteFactory1, IDWriteFactory2, etc. Prior to Windows 10, each new version got a new header file.
+Starting in TH1, new APIs have been added to dwrite_3.h, but conditionalized based on NTDDI_VERSION
+as follows:
+
+```c++
+#if NTDDI_VERSION >= NTDDI_WIN10_MN
```
jonwis, line 56: These versions aren't meaningful to developers; consider adding a "// Windows 10 release 1910" or whatever

specs/dwritecore/DWriteCore.md
jonwis: Yes, it's open for anybody to use. Similar conversation with the regular inbox API reviews, though; "you need this other thing to use these APIs" is confusing unless we have some blanket statement.

specs/dwritecore/DWriteCore.md
```diff
@@ -0,0 +1,255 @@
...
+for more information.
+
+For an application that already uses DirectWrite, switching to DWriteCore requires minimal changes:
+
+ - Add a reference to the Project Reunion package.
+ - Include dwrite_core.h instead of dwrite_3.h.
+ - Link to DWriteCore.lib instead of DWrite.lib.
+
+In return, the application gets the benefits of Project Reunion, namely, access to the newest APIs
+and functionality regardless of what version of Windows your customer is running.
```
jonwis, line 15: DISCUSSION: There's open issues with how DWriteCore will target other renders, including additional future APIs to target non-GDI things.

jonwis: DISCUSSION: Much of the implementation is about using per-process isolated content, like not accessing system caches or the global list of caches.  A core use is letting an app allow-list fonts. A browser might use that to prevent fingerprinting, or enforcing access control from a very low rights render process vs a display or management process.

specs/dwritecore/DWriteCore.md
jonwis: RECOMMEND: Yes

specs/dwritecore/DWriteCore.md
jonwis: DISCUSSION: Ifdefs are carried from the public headers, which require versioning so apps can restrict API use. DWriteCore lets developers always use the APIs even if they're not on `_MN`.

specs/dwritecore/DWriteCore.md
jonwis: DISCUSSION: This is a new enum, the existing values are "shared" and "isolated" ... shared produces singleton dwrite instances, isolated means you get your very own (but uses cross-process caching.)  This "restricted" value is both isolated (not shared) and limited to in-process content and a subset of available system fonts.  The app is in control of what the instance gets.

In-box DWrite works with the font caching service to let low-privileged apps still get the performance benefits of cross-process sharing.  Some caching goes into %TEMP% which is a per-application cache location.

RECOMMEND: `_ISOLATED2` seems not a bad answer, but needs clear documentation.

jonwis: DISCUSSION: There's a cost for non-caching (memory, cpu, pages.) Have a clear recommendation to use the `_SHARED` model as much as possible for maximal caching.  The `_ISOLATED2` mode is sharper, and puts the onus on the app to use it only when appropriate (like controlling the list of available fonts.)

specs/dwritecore/DWriteCore.md
jonwis: DISCUSSION: It's top-down, 32bpp without scanline padding. (There's a bug in the sample code.)  What if someone wants another format in the future?  We'd create a new API, new input/output to handle it.  For non-breaking changes it could be a `...2` method, for breaking-change boundaries it could be the same method but re-typed so consumers detect the change as a compilation error.

RECOMMEND: Contact the target consumer groups and ask them for feedback.

jonwis: SAMPLE: Can we stuff a composition render target into DWrite to target?

specs/dwritecore/DWriteCore.md
jonwis: SAMPLE: This is sample code so it should be super high quality. Bias in the Win32 samples is to not use SAL. If this was modernized, you'd pass `IDWriteBitmapRenderTarget&` which avoids the pointer-ness. For SAL, `_In_` means "pointer, not optional" here.

RECOMMEND: @oldnewthing  will check the samples guidelines and tell us.

specs/dwritecore/DWriteCore.md
jonwis: RECOMMEND: remove the `OUT` keyword, if you are compelled use `/* out */` as an inline comment.

specs/dwritecore/DWriteCore.md
jonwis: CONSIDER: Use `wil` to simplify windows-specific samples, but may not cross-compile to clang/gcc (@DefaultRyan, @dunhor, fyi).

DISCUSSION: cppwinrt does have support for exceptions with `check_result` but no equivalent to resource managers like `HGDIOBJ`.

specs/dwritecore/DWriteCore.md
```diff
@@ -0,0 +1,255 @@
...
+    /// objects from other factories.
+    /// </summary>
+    DWRITE_FACTORY_TYPE_ISOLATED,
+
+#if DWRITE_CORE
+    /// <summary>
+    /// Objects created from a restricted factory do not use or modify internal state or cached data
+    /// used by other factories. In addition, the system font collection contains only well-known fonts.
+    /// </summary>
+    DWRITE_FACTORY_TYPE_RESTRICTED
```
jonwis, line 224: DISCUSSION: If we add something else to this list, a new factory type? Or do we have to "reunify" first?

CONSIDER: Figure out how to annotate that this is related to the Windows public header so they can align enum literal values.

specs/dwritecore/DWriteCore.md
jonwis: CONSIDER: Remove `PURE` as this is a C++ header already and not using the "regular" "build an interface with macros for C and C++"

niklasb-ms: Agreed it's not really useful for C++, but all the existing DWrite methods use PURE.

specs/dwritecore/DWriteCore.md
niklasb-ms: Regarding WIL or C++/WinRT, I think the raw HGDIOBJ is fine here because there's no resource ownership. An error-checking helper would be nice, but check_result takes an HRESULT, whereas what's needed in this case is a null check followed by GetLastError on failure.

Anyway, this code path in the sample snippet is meant to show how the "old way" (in the absence of the new API) requires jumping through some hoops, so maybe it's OK if the code is slightly more verbose. :-)

specs/dwritecore/DWriteCore.md
niklasb-ms: Renamed and made the recommendation to use _SHARED more prominent in the comments.

specs/dwritecore/DWriteCore.md
niklasb-ms: I added clarification that it's top-down with no scanline padding.

Regarding composition render target, we're still discussing how to support more powerful rendering (like Direct2D offers).

specs/dwritecore/DWriteCore.md
niklasb-ms: Leaving as-is because there's no resource management here (so raw HGDIOBJ works fine) and check_result is not applicable to a null check.

specs/dwritecore/DWriteCore.md
niklasb-ms: I think it is indicated in the template. I just missed this one.

specs/dwritecore/DWriteCore.md
niklasb-ms: I updated the comments to hopefully add more clarity.

specs/dwritecore/DWriteCore.md
niklasb-ms: I added a comment about keeping the enums consistent between future versions of this header in DWriteCore vs. the Windows SDK.

PR #199, opened 2020-09-18
Bring over two changes from mscodehub: adding support for class libs and changing NuGet folder structure per request.

rohanp-msft:


axelandrejs (approved):

dev/MRTCore/TestApps/winui_class_lib/winui_class_lib.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.ApplicationModel.Resources;
+
+// To learn more about WinUI, the WinUI project structure,
+// and more about our project templates, see: http://aka.ms/winui-project-info.
+
+namespace winui_class_lib
```
ptorr-msft, line 14: This is not the naming convention for .NET or WinRT types. It is a C++ naming convention.

I would suggest the namespace be (eg) `SampleLibrary` and the type name be `SampleType` (if indeed the desire is to have a generic sample with no meaningful behaviour other than the "get a string" method.

dev/MRTCore/azure-pipelines.yml
```diff
@@ -168,7 +168,7 @@ jobs:
     inputs:
       SourceFolder: '$(MRTBinariesDirectory)\Release\AnyCPU\Microsoft.ApplicationModel.Resources.Projection'
       Contents: |
-        Microsoft.ApplicationModel.Resources.winmd
+        net5.0\Microsoft.ApplicationModel.Resources.winmd
```
huichen123, line 171: >net5.0 [](start = 8, length = 6)

need to see if this builds

huichen123: nope

---
In reply to: [491236518](https://github.com/microsoft/ProjectReunion/pull/199#discussion_r491236518) [](ancestors = 491236518)

rohanp-msft: On my machine, I see this:

S:\feature_work\reunion_mrt_tooling\MRTCore\mrt\AnyCPU\Debug\Microsoft.ApplicationModel.Resources.Projection\net5.0\Microsoft.ApplicationModel.Resources.winmd

Perhaps the source dir needs to be changed here to
`SourceFolder: '$(MRTBinariesDirectory)\AnyCPU\ReleaseMicrosoft.ApplicationModel.Resources.Projection'`

Also, I wonder why this has worked fine in the CI builds in MSCodeHub...

dev/MRTCore/TestApps/winui_class_lib/winui_class_lib.cs
rohanp-msft: Alright, will fix this in a future, separate change.

rohanp-msft: Will address this in a future, separate change.

dev/MRTCore/azure-pipelines.yml
rohanp-msft: Fixed.

PR #205, opened 2020-09-22
Create initial MRTCore release pipeline file

huichen123:


dev/MRTCore/mrtcorerelease.yml
```diff
@@ -0,0 +1,650 @@
...
+# Build a Universal Windows Platform project using Visual Studio.
+# Add steps that test and distribute an app, save build artifacts, and more:
+# https://aka.ms/yaml
+
+name: $(BuildDefinitionName)-$(date:yyMM).$(date:dd)$(rev:rrr)
+
+#trigger:
+#- master
+
+jobs:
```
EHO-makai, line 11: This file is quite a bit longer than WinUI2's MUX-Release.yml.  Is this duplicating stuff from the CI build?  Can this accept the output of another YML and just do sign and publish?  This is essentially what it looks like the WinUI2 release yml does.

huichen123: we can make improvement to the files later. for now, I'd like to get things started. Without a working pipeline, I'd rather avoid the unnecessary content change for now.

axelandrejs: Agreed, we should get it working ASAP then make a follow-up change to prune if needed.

axelandrejs (approved):

dev/MRTCore/mrtcorerelease.yml
huichen123: @EHO-makai  should i proceed or wait?

PR #206, opened 2020-09-22
Add WinRT API to spec for sharing security descriptors

DefaultRyan:
Add WinRT API to spec for sharing security descriptors


specs/GetSecurityDescriptorForAppContainerNames/GetSecurityDescriptorForAppContainerNames.md
```diff
@@ -61,6 +61,26 @@ STDAPI GetSecurityDescriptorForAppContainerNames(
...
+{
+  struct AppContainerAccess
+  {
+    String appContainerName;
+    UInt32 accessMask;
+  }
+
+  static runtimeclass SecurityDescriptorHelpers
+  {
+    String GetSDDLForAppContainerNames(
```
jonwis, line 76: Does this also need a 'static' here? Or is "class level static" sufficient?

DefaultRyan: Good callout. Apparently, I don't author enough MIDL3 to remember the rules here, but you're correct, I do need `static` on the method.

https://docs.microsoft.com/en-us/uwp/midl-3/intro#classes

mziller (commented):
With the added WinRT APIs, the full doc here still reads as specific to Win32 callers and doesn't describe in much detail usage of the new GetSDDLForAppContainerNames method (e.g. an example use case like shown for Win32)

specs/GetSecurityDescriptorForAppContainerNames/GetSecurityDescriptorForAppContainerNames.md
```diff
@@ -61,6 +61,26 @@ STDAPI GetSecurityDescriptorForAppContainerNames(
 
 If the function succeds, the returned `SECURITY_DESCRIPTOR` must be freed by calling [LocalFree](https://docs.microsoft.com/en-us/windows/desktop/api/winbase/nf-winbase-localfree).
 
+## WinRT
+```c#
+namespace Microsoft.Security.AccessControl
+{
+  struct AppContainerAccess
```
ptorr-msft, line 68: Update name to match native one, if necessary. 
Also, why struct? Win32 structs are extensible, but WinRT structs are not. Should this be a runtime class?

specs/GetSecurityDescriptorForAppContainerNames/GetSecurityDescriptorForAppContainerNames.md
```diff
@@ -61,6 +61,26 @@ STDAPI GetSecurityDescriptorForAppContainerNames(
...
+  {
+    String appContainerName;
+    UInt32 accessMask;
+  }
+
+  static runtimeclass SecurityDescriptorHelpers
+  {
+    static String GetSDDLForAppContainerNames(
```
ptorr-msft, line 76: Strange that this generates the SDDL but the native one generates the actual `SECURITY_DESCRIPTOR`. Is there any reasonable way you can use the SDDL in a projection where you couldn't use the SD? (In other words, does this WinRT API give you a result that effectively requires you to call into Win32 anyway?)

oldnewthing: You could use the SDDL in places like manifests. (There's really no good way to return a `SECURITY_DESCRIPTOR` to WinRT; maybe a `byte[]`?)

oldnewthing, line 76: Presumable `Sddl` per capitalization guidelines.

specs/GetSecurityDescriptorForAppContainerNames/GetSecurityDescriptorForAppContainerNames.md
jonwis: RECOMMEND: Yes, rename to match the C++ one.

RECOMMEND: Switch to runtimeclass here for less breakage

specs/GetSecurityDescriptorForAppContainerNames/GetSecurityDescriptorForAppContainerNames.md
jonwis: CONSIDER: Alternate `byte[] GetSecurityDescriptorBytesFromAppContainerNames` for use with .NET's `RawSecurityDescriptor` constructor. Use customer feedback to know what's important.

jonwis: RECOMMEND: Yes

jonwis:
> With the added WinRT APIs, the full doc here still reads as specific to Win32 callers and doesn't describe in much detail usage of the new GetSDDLForAppContainerNames method (e.g. an example use case like shown for Win32)

RECOMMEND: Show a WinRT sample using the API and populating a RawSecurityDescriptor with it for .NET

jonwis (approved):

PR #209, opened 2020-09-24
Code/test feature for sharing security descriptors

DefaultRyan:
This is the implementation and test code for #188


jonwis:
Can you remove one level of directory structure so it's not "s.ac/s.ac/files" ?

dev/Security.AccessControl/Security.AccessControl/GetSecurityDescriptorForAppContainerNames.cpp
```diff
@@ -0,0 +1,116 @@
...
+            &length
+        ))
+        {
+            auto err = ::GetLastError();
+            if (err != ERROR_INSUFFICIENT_BUFFER)
+            {
+                THROW_WIN32(err);
+            }
+            tg.reset(static_cast<TOKEN_GROUPS*>(::HeapAlloc(::GetProcessHeap(), HEAP_ZERO_MEMORY, length)));
+        }
```
jonwis, line 34: https://github.com/microsoft/wil/blob/master/include/wil/token_helpers.h could acquire this as a helper! (We need to eventually invert the lookup, so you can pass the token info type you want... there are several information classes that share the same output struct.)

dev/Security.AccessControl/Security.AccessControl/GetSecurityDescriptorForAppContainerNames.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+        THROW_IF_WIN32_BOOL_FALSE(::GetTokenInformation(
+            ::GetCurrentProcessToken(),
+            TokenLogonSid,
+            tg.get(),
+            length,
+            &length));
+
+        auto sidLength = ::GetLengthSid(tg->Groups[0].Sid);
+        result.reset(::HeapAlloc(::GetProcessHeap(), 0, sidLength));
```
jonwis, line 44: Instead of allocating a new thing, maybe just return "tg" and let the caller root around in it? This is an internal method so nicety isn't necessarily required.

dev/Security.AccessControl/Security.AccessControl/GetSecurityDescriptorForAppContainerNames.cpp
```diff
@@ -0,0 +1,116 @@
...
+}
+
+STDAPI GetSecurityDescriptorForAppContainerNames(
+    uint32_t accessRequestCount,
+    _In_reads_(accessRequestCount)
+        const AppContainerNameAndAccess* accessRequests,
+    _In_opt_ PSID principal,
+    uint32_t principalAccessMask,
+    _Outptr_ PSECURITY_DESCRIPTOR* securityDescriptor
+)
```
jonwis, line 58: Use a function-level try/catch to reduce one level of indentation.  This should probably be marked noexcept as well.

dev/Security.AccessControl/Security.AccessControl/GetSecurityDescriptorForAppContainerNames.cpp
```diff
@@ -0,0 +1,116 @@
...
+{
+    *securityDescriptor = nullptr;
+    try
+    {
+        vector<EXPLICIT_ACCESS> ea;
+        if (accessRequestCount == UINT32_MAX)
```
jonwis, line 64: Consider removing this "overflow' check; if the reserve fails then eventually the push-backs will fail later and that's OK.

dev/Security.AccessControl/Security.AccessControl/GetSecurityDescriptorForAppContainerNames.cpp
```diff
@@ -0,0 +1,116 @@
...
+            entry.grfAccessPermissions = accessRequests->accessMask;
+            entry.grfInheritance = NO_INHERITANCE;
+            entry.Trustee.TrusteeForm = TRUSTEE_IS_SID;
+            entry.Trustee.TrusteeType = TRUSTEE_IS_USER;
+            entry.Trustee.ptstrName = static_cast<PWSTR>(sid.get());
+            appSids.push_back(std::move(sid));
+            ea.push_back(entry);
+        }
+
+        wil::unique_any<PACL, decltype(&::LocalFree), ::LocalFree> acl;
```
jonwis, line 105: Another thing to wil-ify as unique_hlocal_acl ? (No rush on any "put this in wil" comments.)

dev/Security.AccessControl/Security.AccessControl/GetSecurityDescriptorForAppContainerNames.cpp
```diff
@@ -0,0 +1,116 @@
...
+        THROW_IF_WIN32_ERROR(::SetEntriesInAclW(static_cast<uint32_t>(ea.size()), ea.data(), nullptr, &acl));
+
+        auto result = wil::make_unique_hlocal<SECURITY_DESCRIPTOR>();
+        THROW_IF_WIN32_BOOL_FALSE(::InitializeSecurityDescriptor(result.get(), SECURITY_DESCRIPTOR_REVISION));
+        THROW_IF_WIN32_BOOL_FALSE(::SetSecurityDescriptorDacl(result.get(), TRUE, acl.get(), FALSE));
```
jonwis, line 110: Check out https://docs.microsoft.com/en-us/windows/win32/secauthz/absolute-and-self-relative-security-descriptors - this method needs to hand out a self-relative SD (and be documented as such) so the "binary bytes" thing and freeing the SD works correctly.

jonwis (commented):
Is the WinRT variant coming soon?

dev/Security.AccessControl/Security.AccessControl/GetSecurityDescriptorForAppContainerNames.cpp
DefaultRyan: This function is the implementation of a "Flat C" API. So we have a larger question of whether we are going to mark those APIs as `noexcept` in the headers.

Clearly, that keyword won't work for plain C consumers. It's only available for C++ consumers if they are running C++11 or greater. I'm in favor of this, but not sure if this decision has been made yet.

DefaultRyan: I have moved the API into `extern "C"` which has a similar benefit to `noexcept` when compiling with `/EHsc` (which is the recommended setting).

dev/Security.AccessControl/UWP/AccessControlTest_UWP/Source.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+    std::wstring commandLine = L"AccessControlTest_Helper.exe ";
+    commandLine += familyName.c_str();
+
+    wil::unique_handle child_process;
+    wil::unique_handle child_thread;
+    {
+        STARTUPINFO si{};
+        si.cb = sizeof(si);
+        PROCESS_INFORMATION pi{};
```
jonwis, line 32: See also:

```
    using unique_process_information = unique_struct<PROCESS_INFORMATION, decltype(&details::CloseProcessInformation), details::CloseProcessInformation>;
```

mshaikhcool:
It's sitting  idle for like almost one full year. Windows App SDK still only supports medium IL process. Dynamic Dependencies most likely won't support Elevation by this year. No mention of AppContainer scenarios either.  

@DefaultRyan Are we about to get sort of an update on this anytime sooner ?

DefaultRyan:
This PR got so old that it was easier to create a new one. See #2005

PR #216, opened 2020-09-28
Update copyright text

huichen123:


rohanp-msft:
What about files like Microsoft.ApplicationModel.Resources.References.targets under packaging\?
They currently have:
```
<!--
  Copyright (c) Microsoft Corporation. Licensed under the MIT License
-->
```
I don't know if these files need to have a copyright header. If they do though, updating them also would be good. One or two of these files currently don't have a header...

huichen123:
> What about files like Microsoft.ApplicationModel.Resources.References.targets under packaging?
> They currently have:
> 
> ```
> 
> ```
> 
> I don't know if these files need to have a copyright header. If they do though, updating them also would be good. One or two of these files currently don't have a header...

@EHO-makai @jonwis what's the copyright/license header for .targets/.props files?

axelandrejs (approved):

huichen123:
will get to .targets/.props files later.

PR #221, opened 2020-10-01
MRTCore arm build

huichen123:


dev/MRTCore/SignConfig.xml
```diff
@@ -13,6 +13,10 @@
    <file src="__INPATHROOT__\x86\mrm.dll" signType="AuthenticodeFormer" /> 
    <file src="__INPATHROOT__\x86\Microsoft.ApplicationModel.Resources.dll"  signType="AuthenticodeFormer" /> 
  </job>
+ <job platform="ARM" certSubject="CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US" configuration="Release" dest="__INPATHROOT__\ARM\signed" jobname="MRT Core binary sign (ARM)" approvers="">
```
EHO-makai, line 16: we are going to use the proejct reunion signconfig I'm pretty sure ...

dev/MRTCore/SignConfig.xml
```diff
@@ -13,6 +13,10 @@
...
+   <file src="__INPATHROOT__\ARM\mrm.dll" signType="AuthenticodeFormer" />
```
EHO-makai, line 17: general question - did WinUI3 request ARM?  I was wondering about this a few weeks ago (whether you needed to build ARM).  The Project Reunion sample packaging built ARM.

dev/MRTCore/SignConfig.xml
huichen123: until it's integrated, I'd still update the MRT one

dev/MRTCore/azure-pipelines.yml
```diff
@@ -215,10 +174,10 @@ jobs:
   variables:
     solution: '**/*.sln'
     appxPackageDir: '$(build.artifactStagingDirectory)\AppxPackages\\'
-    codeSign: false
+    codeSign: ${{ eq(variables['System.definitionId'], '57248') }}
```
EHO-makai, line 177: pretty sure we are going to sign everything at the Project Reunion (top) level en masse ...  your script should probably (eventually) just build and copy binaries to an agreed upon aggregation point.  I think Geoffrey is working on this right now.

dev/MRTCore/SignConfig.xml
huichen123: not yet. WinUI3 did request for ARM64. As supporting ARM is also part of priority and Project Reunion builds ARM, I figure MRT needs to build ARM too.

dev/MRTCore/azure-pipelines.yml
huichen123: sure. until then I'd still maintain this one.

huichen123: do we know where is the aggregation point?

getrou: I think the aggregation point will just be some place in the ArtifactStagingDirectory, almost exactly what you're doing now.  I think all we will need to do is change the artifact name from "mrt" to something common that all components can use.

dev/MRTCore/azure-pipelines.yml
```diff
@@ -302,6 +264,15 @@ jobs:
       artifactName: 'signedbinaries_x64'
       downloadPath: '$(Build.ArtifactStagingDirectory)\raw'
 
+  - task: DownloadBuildArtifacts@0
```
getrou, line 267: I don't suppose there's a good way to parameterize these downloads/copies per architecture like you did above?

huichen123: We can still do those in parallel. However, those are very simple tasks of copying files into final places. Doing them on separate jobs and sync/publish/download just adds extra complexity and overhead.

getrou (approved):

PR #230, opened 2020-10-08
User/getrou/converged nuget

getrou:
The new artifact and nuget layout can be seen here:
https://dev.azure.com/ms/ProjectReunion/_build/results?buildId=114478&view=artifacts&type=publishedArtifacts

### DIFF
--- a/build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
+++ b/build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
@@ -3,7 +3,7 @@ parameters:
   jobName: ''
   dependsOn: ''
   buildOutputDir: '$(Build.SourcesDirectory)\BuildOutput'
-  nupkgdir: '$(build.artifactStagingDirectory)'
+  nupkgdir: '$(build.artifactStagingDirectory)\nuget'
 # The "primary" build arch is the one that the nuspec gets its winmd, pri, and other neutral files from
   primaryBuildArch: x86
   buildFlavor: Release
@@ -58,31 +58,53 @@ jobs:
 
   - template: ProjectReunion-InstallNuget-Steps.yml
 
-  - ${{if parameters.signConfig }}:
-    # Re-create framework packages because we made them in the build step against unsigned binaries.
-    - template: ProjectReunion-MakeFrameworkPackages-Steps.yml
+  # UNDONE to move MRTCore into Project Reunion nuget package
+  # - ${{if parameters.signConfig }}:
+  #   # Re-create framework packages because we made them in the build step against unsigned binaries.
+  #   - template: ProjectReunion-MakeFrameworkPackages-Steps.yml
 
-    # Copy the re-made framework packages into the same location in the artifact drop to be republished at the end of the job.
-    - task: CopyFiles@2
-      inputs:
-        targetFolder: '${{ parameters.nupkgdir }}'
-        sourceFolder: '$(Build.SourcesDirectory)\Artifacts\drop'
-        Contents: '**\Microsoft.ProjectReunion*.appx'
+  #   # Copy the re-made framework packages into the same location in the artifact drop to be republished at the end of the job.
+  #   - task: CopyFiles@2
+  #     inputs:
+  #       targetFolder: '${{ parameters.nupkgdir }}'
+  #       sourceFolder: '$(Build.SourcesDirectory)\Artifacts\drop'
+  #       Contents: '**\Microsoft.ProjectReunion*.appx'
 
-  - powershell: |
-      $prereleaseTag = "${{ parameters.prereleaseVersionTag }}"
-      if ("${{ parameters.useReleaseTag}}" -eq [bool]::TrueString) { $prereleaseTag = "" }
+  # - powershell: |
+  #     $prereleaseTag = "${{ parameters.prereleaseVersionTag }}"
+  #     if ("${{ parameters.useReleaseTag}}" -eq [bool]::TrueString) { $prereleaseTag = "" }
 
-      & "$(Build.SourcesDirectory)\build\NuSpecs\build-nupkg.ps1" `
-        -BuildOutput '$(Build.SourcesDirectory)\Artifacts\drop' `
-        -OutputDir '${{ parameters.nupkgdir }}' `
-        -prereleaseversion "$prereleaseTag" `
-        -DateOverride '$(builddate)' `
-        -Subversion '$(buildrevision)' `
-        -BuildArch ${{ parameters.primaryBuildArch }} `
-        -BuildFlavor ${{ parameters.buildFlavor }}
+  #     & "$(Build.SourcesDirectory)\build\NuSpecs\build-nupkg.ps1" `
+  #       -BuildOutput '$(Build.SourcesDirectory)\Artifacts\drop' `
+  #       -OutputDir '${{ parameters.nupkgdir }}' `
+  #       -prereleaseversion "$prereleaseTag" `
+  #       -DateOverride '$(builddate)' `
+  #       -Subversion '$(buildrevision)' `
+  #       -BuildArch ${{ parameters.primaryBuildArch }} `
+  #       -BuildFlavor ${{ parameters.buildFlavor }}
 
-    displayName: 'build-nupkg.ps1'
+  #   displayName: 'build-nupkg.ps1'
+
+  steps:
+  - task: NuGetToolInstaller@1
+    displayName: 'Use NuGet 5.6'
+    inputs:
+      versionSpec: 5.6
+    continueOnError: true
+  - task: 333b11bd-d341-40d9-afcf-b32d5ce6f23b@2
+    inputs:
+      command: 'pack'
+      packagesToPack: '$(Build.SourcesDirectory)\dev\MRTCore\mrt\NuGet\Microsoft.ProjectReunion.nuspec'
+      versioningScheme: 'byPrereleaseNumber'
+      majorVersion: '$(Major)'
+      minorVersion: '$(Minor)'
+      patchVersion: '$(Patch)'
+      basePath: '$(Build.ArtifactStagingDirectory)\drop'
+      packDestination: ${{ parameters.nupkgdir }}
+  - task: PublishBuildArtifacts@1
+    inputs:
+      PathtoPublish: ${{ parameters.nupkgdir }}
+      ArtifactName: 'nuget' 
 
   - ${{if parameters.signConfig }}:
     - task: PkgESCodeSign@10
@@ -98,7 +120,7 @@ jobs:
     displayName: 'Publish artifact: nupkg'
     inputs:
       PathtoPublish: '${{ parameters.nupkgdir }}'
-      artifactName: 'drop'
+      artifactName: 'nuget'
 
 #UNDONE - EHO we need to seed these guid's properly!
 #see, e.g. AzurePipelinesTemplates\ProjectReunion-BuildAndPublishPGONuGet-Job.yml
--- a/build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
+++ b/build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
@@ -128,16 +128,17 @@ jobs:
       PathtoPublish: '${{ parameters.nupkgdir }}'
       artifactName: 'nuget'
       
+  - ${{if parameters.signConfig }}:
     # this mysterious guid fixes the "NuGetCommand@2 is ambiguous" error :-(
-  - task: 333b11bd-d341-40d9-afcf-b32d5ce6f23b@2
-    displayName: 'NuGet push to Project.Reunion.nuget.internal'
-    inputs:
-      command: 'push'
-      packagesToPush: '$(Build.ArtifactStagingDirectory)/*.nupkg;!$(Build.ArtifactStagingDirectory)/*.symbols.nupkg'
-      verbosityPush: 'Detailed' 
-      nuGetFeedType: 'internal'
-      #Note: The project qualifier is always required when using a feed name. Also, do not use organization scoped feeds. 
-      publishVstsFeed: 'ProjectReunion/Project.Reunion.nuget.internal'
+    - task: 333b11bd-d341-40d9-afcf-b32d5ce6f23b@2
+      displayName: 'NuGet push to Project.Reunion.nuget.internal'
+      inputs:
+        command: 'push'
+        packagesToPush: '$(Build.ArtifactStagingDirectory)/nuget/*.nupkg;!$(Build.ArtifactStagingDirectory)/nuget/*.symbols.nupkg'
+        verbosityPush: 'Detailed' 
+        nuGetFeedType: 'internal'
+        #Note: The project qualifier is always required when using a feed name. Also, do not use organization scoped feeds. 
+        publishVstsFeed: 'ProjectReunion/Project.Reunion.nuget.internal'
 
 #UNDONE - EHO we need to seed these guid's properly!
 #see, e.g. AzurePipelinesTemplates\ProjectReunion-BuildAndPublishPGONuGet-Job.yml
--- a/build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
+++ b/build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
@@ -89,11 +89,19 @@ jobs:
   #       -BuildFlavor ${{ parameters.buildFlavor }}
 
   #   displayName: 'build-nupkg.ps1'
+  
+  - task: CopyFiles@2
+    displayName: 'copy license'
+    inputs:
+      SourceFolder: '$(Build.SourcesDirectory)'
+      Contents: |
+        LICENSE
+      TargetFolder: '$(Build.ArtifactStagingDirectory)\drop'
 
   - task: 333b11bd-d341-40d9-afcf-b32d5ce6f23b@2
     inputs:
       command: 'pack'
-      packagesToPack: '$(Build.SourcesDirectory)\dev\MRTCore\mrt\NuGet\Microsoft.ProjectReunion.nuspec'
+      packagesToPack: '$(Build.SourcesDirectory)\build\NuSpecs\Microsoft.ProjectReunion.nuspec'
       versioningScheme: 'byPrereleaseNumber'
       majorVersion: ${{ parameters.major }}
       minorVersion: ${{ parameters.minor }}
--- a/build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
+++ b/build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
@@ -27,7 +27,7 @@ jobs:
       name: Package ES Lab E
   
   # otherwise run on Windows VM
-  ${{if !parameters.signConfig}}:
+  ${{if eq(parameters.signConfig, false)}}:
     pool:
       vmImage: 'windows-2019'
 
--- a/build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
+++ b/build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
@@ -96,8 +96,7 @@ jobs:
       SourceFolder: '$(Build.SourcesDirectory)'
       Contents: |
         LICENSE
-      TargetFolder: '$(Build.ArtifactStagingDirectory)\drop'
-
+      TargetFolder: '$(Build.SourcesDirectory)\Artifacts\drop'
   - task: 333b11bd-d341-40d9-afcf-b32d5ce6f23b@2
     inputs:
       command: 'pack'
--- a/build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
+++ b/build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
@@ -21,13 +21,13 @@ jobs:
     - ${{ parameters.dependsOn }}
   condition: ${{ parameters.condition }}
 
-  # # XES tasks need to be on PackageES machines.
-  # ${{if parameters.signConfig }}:
-  #   pool:
-  #     name: Package ES Lab E
+  # XES tasks need to be on PackageES machines.
+  ${{if parameters.signConfig }}:
+    pool:
+      name: Package ES Lab E
   
-  # # otherwise run on Windows VM
-  # ${{if eq(parameters.signConfig, 'false')}}:
+  # otherwise run on Windows VM
+  ${{if eq(parameters.signConfig, 'false')}}:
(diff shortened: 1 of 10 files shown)


dev/MRTCore/azure-pipelines.yml
```diff
@@ -383,29 +383,15 @@ jobs:
...
         mrm.h
-      TargetFolder: '$(Build.ArtifactStagingDirectory)\mrt\include'
+      TargetFolder: '$(Build.ArtifactStagingDirectory)\drop\include'
 
   - task: CopyFiles@2
     displayName: 'copy license'
     inputs:
       SourceFolder: '$(MRTSourcesDirectory)'
       Contents: |
         LICENSE.txt
```
huichen123, line 393: >LICENSE.txt [](start = 8, length = 11)

we need to deprecate this one with the one under root folder.

dev/MRTCore/azure-pipelines.yml
```diff
@@ -383,29 +383,15 @@ jobs:
...
-      versioningScheme: 'byPrereleaseNumber'
-      majorVersion: '$(Major)'
-      minorVersion: '$(Minor)'
-      patchVersion: '$(Patch)'
-      basePath: '$(Build.ArtifactStagingDirectory)\mrt'
-      packDestination: '$(Build.ArtifactStagingDirectory)\nuget'
+      TargetFolder: '$(Build.ArtifactStagingDirectory)\drop'
 
   - ${{if eq(parameters.doCodeSign, 'true')}}:
     - task: PkgESCodeSign@10
```
huichen123, line 397: this task is no longer needed

dev/MRTCore/azure-pipelines.yml
```diff
@@ -414,12 +400,12 @@ jobs:
         SYSTEM_ACCESSTOKEN: $(System.AccessToken)
       inputs:
         signConfigXml: '$(MRTSourcesDirectory)\SignConfig-nuget.xml'
-        inPathRoot: '$(BUILD.ArtifactStagingDirectory)\nuget'
-        outPathRoot: '$(BUILD.ArtifactStagingDirectory)\nuget'
+        inPathRoot: '$(BUILD.ArtifactStagingDirectory)\drop'
+        outPathRoot: '$(BUILD.ArtifactStagingDirectory)\drop'
   
   - task: PublishBuildArtifacts@1
```
huichen123, line 406: >- task: PublishBuildArtifacts@1 [](start = 2, length = 31)

no longer needed

dev/MRTCore/mrt/NuGet/Microsoft.ProjectReunion.nuspec
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
```
huichen123, line 1: this file should not be under MRTCore folder

dev/MRTCore/mrt/NuGet/Microsoft.ProjectReunion.nuspec
```diff
@@ -0,0 +1,15 @@
...
+<package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
+     <metadata minClientVersion="3.3">
+         <id>Microsoft.ProjectReunion</id>
+         <version>1.0</version>
+         <title>$id$</title>
+         <authors>Microsoft</authors>
+         <description>Resource loader for .PRI files.</description>
```
huichen123, line 8: >Resource loader for .PRI files. [](start = 22, length = 31)

description is not correct now.

dev/MRTCore/mrt/NuGet/Microsoft.ProjectReunion.nuspec
```diff
@@ -0,0 +1,15 @@
...
+         <license type="file">LICENSE.txt</license>
+         <requireLicenseAcceptance>true</requireLicenseAcceptance>
+         <projectUrl>https://aka.ms/MRTCore</projectUrl>
```
huichen123, line 11: >https://aka.ms/MRTCore [](start = 21, length = 22)

also need to point to project reunion

build/ProjectReunion-CI.yml
```diff
@@ -50,8 +55,9 @@ jobs:
 - template: AzurePipelinesTemplates\ProjectReunion-CreateNugetPackage-Job.yml
   parameters:
     jobName: CreateNugetPackage
-    dependsOn: Build
+    dependsOn: PublishMRT
     prereleaseVersionTag: ci
+    condition: Always()
```
huichen123, line 60: >condition: Always() [](start = 4, length = 19)

shouldn't do only when succeeded?

build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
```diff
@@ -53,36 +58,52 @@ jobs:
...
-        -DateOverride '$(builddate)' `
-        -Subversion '$(buildrevision)' `
-        -BuildArch ${{ parameters.primaryBuildArch }} `
-        -BuildFlavor ${{ parameters.buildFlavor }}
-
-    displayName: 'build-nupkg.ps1'
+  - task: 333b11bd-d341-40d9-afcf-b32d5ce6f23b@2
+    inputs:
+      command: 'pack'
+      packagesToPack: '$(Build.SourcesDirectory)\dev\MRTCore\mrt\NuGet\Microsoft.ProjectReunion.nuspec'
```
huichen123, line 96: >MRTCore\mrt\NuGet\Microsoft.ProjectReunion.nuspec [](start = 53, length = 49)

this needs to be outside MRTCore

huichen123:
>         outPathRoot: '${{ parameters.nupkgdir }}\signed'

remove

---
Refers to: build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml:116 in aa04ba8. [](commit_id = aa04ba83695ef955045b453845fe7eaf744c5c8f, deletion_comment = False)

build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
```diff
@@ -53,36 +58,52 @@ jobs:
...
+      versioningScheme: 'byPrereleaseNumber'
+      majorVersion: ${{ parameters.major }}
+      minorVersion: ${{ parameters.minor }}
+      patchVersion: ${{ parameters.patch }}
+      basePath: '$(Build.SourcesDirectory)\Artifacts\drop'
```
huichen123, line 101: >drop [](start = 53, length = 4)

as is we pack way too much stuff. (ok to fix it later)

huichen123:
@EHO-makai , it appears the pipeline is not run. neither for your PR. something is wrong here.

build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
```diff
@@ -53,36 +58,52 @@ jobs:
...
-    - task: CopyFiles@2
-      inputs:
-        targetFolder: '${{ parameters.nupkgdir }}'
-        sourceFolder: '$(Build.SourcesDirectory)\Artifacts\drop'
-        Contents: '**\Microsoft.ProjectReunion*.appx'
-
-  - powershell: |
-      $prereleaseTag = "${{ parameters.prereleaseVersionTag }}"
-      if ("${{ parameters.useReleaseTag}}" -eq [bool]::TrueString) { $prereleaseTag = "" }
+  # UNDONE to move MRTCore into Project Reunion nuget package
```
EHO-makai, line 66: are you going to re-enable this before completing this PR?

build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
EHO-makai: Feels a little weird to have 'Microsoft.ProjectReunion.nuspec' in dev\MRTCore.  I suppose this is just a "phase 1" kind of hting.

build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
EHO-makai: we're supposed to pack the exact same thing we were packing before, this is something I told Alan Wu we would do.

EHO-makai: OTOH, I love that we're using the nuget pack command.  :-)

huichen123: if you open the package, you know what i was saying. :)

build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
```diff
@@ -98,7 +119,7 @@ jobs:
     displayName: 'Publish artifact: nupkg'
     inputs:
       PathtoPublish: '${{ parameters.nupkgdir }}'
-      artifactName: 'drop'
+      artifactName: 'nuget'
```
EHO-makai, line 122: I have a precise directory that I use to do the nuget signing, this will have to be adjusted/re-tested after you completee this PR.

build/ProjectReunion-CI.yml
EHO-makai: Geoffrey, was this the weird "wouldn't run" problem we were looking at before?  Does seem like we should remove always as a condition.

dev/MRTCore/mrt/NuGet/Microsoft.ProjectReunion.nuspec
EHO-makai: I think this is more or less true I wonder when we will have a plan for aggregating multiple nuspecs for the sub modules inside of the Project Reunion repo into one we can feed to the nuget pack command.  I think we don't have this plan yet so it's probably OK that it is here temporarily.

EHO-makai: You two can work out what name this should have.

dev/MRTCore/mrt/NuGet/Microsoft.ProjectReunion.nuspec
EHO-makai: I'll have somebody register aka.ms/projectreunion to point to github/microsoft/projectreunion.

build/ProjectReunion-CI.yml
getrou: Ah, yes.  I still haven't figured out why when this condition runs, succeeded() and succeededOrFailed() both return false.

build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
getrou: No, this removes the creation of the nuget and framework packages for the Project Reunion sample app.  I'm leaving it in as a comment so we have the framework package structure in case we want to use it to build the framework packages for our new MRTCore/ProjectReunion nuget package.

dev/MRTCore/mrt/NuGet/Microsoft.ProjectReunion.nuspec
getrou: I also asked this question, and the answer is that MRTCore doesn't have an identity outside of Project Reunion, so it shouldn't get it's own nuget package.  The assertion is that there won't be anyone using MRTCore without using all of ProjectReunion (aside from WinUI who will do something special).

huichen123: when Reunion grows, say it contains 100 subprojects. It's very unlikely someone will need them all.

dev/MRTCore/azure-pipelines.yml
getrou: If we don't do this, won't we not have access to the signed binaries in the next job?

dev/MRTCore/mrt/NuGet/Microsoft.ProjectReunion.nuspec
getrou: @EHO-makai would you like to take this?

dev/MRTCore/mrt/NuGet/Microsoft.ProjectReunion.nuspec
getrou: Fixed, moved file out of this directory and changed description

build/NuSpecs/Microsoft.ProjectReunion.nuspec
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
+     <metadata minClientVersion="3.3">
+         <id>Microsoft.ProjectReunion</id>
+         <version>1.0</version>
+         <title>$id$</title>
+         <authors>Microsoft</authors>
+         <description>Project Reunion</description>
```
huichen123, line 8: >Project Reunion [](start = 22, length = 15)

doesn't sound very helpful :). we can change later though

dev/MRTCore/azure-pipelines.yml
huichen123: isn't nuget handled by CreateNugetPackage now?

build/NuSpecs/Microsoft.ProjectReunion.nuspec
getrou: Sure.  I just copied what it said in the original ProjectReunion nuspec file.  I'm sure a PM will come up with something :)

build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
huichen123: yes.

dev/MRTCore/azure-pipelines.yml
getrou: Maybe I misunderstand the point of publishing a build artifact.  The point of this is to push the signed binaries to a "common drop point" that other sub-modules (once they exist) will also publish to.  Then in the CreateNugetPackage job, we download the contents of this drop point and wrap them into a nuget package.

So this step is just pushing the binaries to the common drop point.

huichen123: create/sign/publish nuget are all handled in CreateNugetPackage now.

getrou: Sure, that makes sense.  But this "publish" step is just publishing the signed binaries to an intermediate location so they are available to the next job.
This step isn't publishing the final nuget package, it's just publishing some intermediate binaries so that the next job, which is CreateNugetPackage, can use them.

huichen123: sorry, you are right. they are different jobs, thus need publish in between.

build/ProjectReunion-CI.yml
getrou: https://dev.azure.com/ms/ProjectReunion/_build/results?buildId=114742&view=results

Here's what happens when I change it to only run on succeeded().  The same thing happens if I use succeededOrFailed().  I'm not sure why, but Always() is the only condition that is working.

huichen123: why do you put condition inside parameters ? :)

getrou: Since the dependsOn was parameterized, I figured the condition should be as well.

huichen123: oh, you have a parameter whose name collides with a keyword. Not sure whether it's a good practice. Try not to pass condition in, but check succeeded() inside CreateNugetPackage directly and see if it makes any difference. In the end, succeeded() check is default and unnecessary.

getrou: Same result unfortunately.
https://dev.azure.com/ms/ProjectReunion/_build/results?buildId=114749&view=results

huichen123: no clue. I'd say just remove condition as succeeded should be the default behavior.

getrou: I believe that will also result in the job being skipped.  I'm pretty confused about why that is, since the previous job that CreateNugetPacakge depends clearly succeeded.

I'm running it without any condition here:
https://dev.azure.com/ms/ProjectReunion/_build/results?buildId=114751&view=results

getrou: Solution was to make the condition specifically look at the PublishMRT job.  I'm not sure why this is, but it works.

dev/MRTCore/mrt/NuGet/Microsoft.ProjectReunion.nuspec
EHO-makai: I handled this in email, there is a doc (which will be live on github so everyone can see) that explains the packaging plan.  Basically there will be "thin" nuget packages from the various repos that aggregate into Project Reunion and Project Reunion's "metapackage" will reference all of these.  So this repo outputs one "thin" nuget package with metadata and VS targets files, etc. and one metapackage that references everything in Project Reunion.  The actual binaries are aggregated into one framework package that gets used at runtime.

EHO-makai: "thin" == metadata plus binaries (dll's), "fat" == contains dll, Project Reunion contributors will publish thin nuget packages to public nuget.org.  The binaries live in the Project Reunion framework package.

getrou: Awesome, thanks Eho.

dev/MRTCore/mrt/NuGet/Microsoft.ProjectReunion.nuspec
EHO-makai: this was already working, can write:  https://aka.ms/ProjectReunion :-)

build/ProjectReunion-CI.yml
EHO-makai: Is this task going away soon?  I'm wondering if this going to come back ... (no action requested from this comment).

build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
EHO-makai: To be explicit, contact me and let's run the release pipeline together (I have to trigger it, I have the signing authorization).

EHO-makai (approved):
Consider Hui's sign-off mandatory, but here is mine.

dev/MRTCore/mrt/NuGet/Microsoft.ProjectReunion.nuspec
getrou: Yep, done!  This file got moved to build/NuSpecs/Microsoft.ProjectReunion.nuspec, where it points to that link now.

build/ProjectReunion-CI.yml
getrou: Is this job going away soon?  I don't think so.  I think even once we start onboarding more sub-modules, this job will stay mostly as-is.

huichen123:
>         outPathRoot: '${{ parameters.nupkgdir }}\signed'

please address this. or you will 2 packages: one signed, one unsigned.

---
In reply to: [705698105](https://github.com/microsoft/ProjectReunion/pull/230#issuecomment-705698105) [](ancestors = 705698105)

---
Refers to: build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml:116 in aa04ba8. [](commit_id = aa04ba83695ef955045b453845fe7eaf744c5c8f, deletion_comment = False)

huichen123:
>       command: 'push'

this only done when we signed the code

---
Refers to: build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml:135 in 28ae59e. [](commit_id = 28ae59e64ef23f54f0a0659a441700bcbccaa7c7, deletion_comment = False)

build/NuSpecs/Microsoft.ProjectReunion.nuspec
```diff
@@ -0,0 +1,15 @@
...
+         <license type="file">LICENSE</license>
+         <requireLicenseAcceptance>true</requireLicenseAcceptance>
+         <projectUrl>https://aka.ms/projectreunion</projectUrl>
+         <copyright>© Microsoft Corporation. All rights reserved.</copyright>
+         <tags>Microsoft Resources Native</tags>
```
huichen123, line 13: >Microsoft Resources Native [](start = 15, length = 26)

add Reunion as tag?

build/ProjectReunion-CI.yml
```diff
@@ -50,8 +55,9 @@ jobs:
 - template: AzurePipelinesTemplates\ProjectReunion-CreateNugetPackage-Job.yml
   parameters:
     jobName: CreateNugetPackage
-    dependsOn: Build
+    dependsOn: PublishMRT
```
huichen123, line 58: >dependsOn [](start = 4, length = 9)

the dependency in release pipeline also needs change, but that can be in another PR.

huichen123:
in the latest build, there are still binlogs in the package.

huichen123:
also reunion.dll is missing.

build/NuSpecs/Microsoft.ProjectReunion.nuspec
getrou: I'll add the tags from the old nuspec, "Windows Reunion ProjectReunion"

huichen123:
>       publishVstsFeed: $(NuGetFeed)

where is this defined?

---
Refers to: build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml:141 in 97a2e8c. [](commit_id = 97a2e8c818a7bb5e5cb0226cc002deba12184cb2, deletion_comment = False)

getrou:
> also reunion.dll is missing.

I think we want reunion.dll removed.  It's part of our sample app code, and we don't want it in the nuget package.  The goal is that the only thing in the package is the MRTCore stuff for now.

huichen123:
>       command: 'push'

please address this, or the pipeline will fail.

---
In reply to: [706362132](https://github.com/microsoft/ProjectReunion/pull/230#issuecomment-706362132) [](ancestors = 706362132)

---
Refers to: build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml:135 in 28ae59e. [](commit_id = 28ae59e64ef23f54f0a0659a441700bcbccaa7c7, deletion_comment = False)

huichen123:
>       packagesToPush: '$(Build.ArtifactStagingDirectory)/*.nupkg;!$(Build.ArtifactStagingDirectory)/*.symbols.nupkg'

the pattern is incorrect, thus the correct pipeline skips this

---
Refers to: build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml:136 in 60cff54. [](commit_id = 60cff547c08bafddb348b47df2a9d0c37a47bde4, deletion_comment = False)

PR #244, opened 2020-10-21
User/getrou/release pipeline fix

getrou:
Update the release pipeline to not fail now that MRTCore is in the ProjectReunion nuget package.

### DIFF
--- a/build/ProjectReunion-release.yml
+++ b/build/ProjectReunion-release.yml
@@ -72,9 +72,9 @@ jobs:
     doCodeSign: 'true'
 
 - job: SignBinariesAndPublishSymbols
-  dependsOn: Build
+  dependsOn: PublishMRT
   condition:
-    in(dependencies.Build.result, 'Succeeded', 'SucceededWithIssues', 'Skipped')
+    in(dependencies.PublishMRT.result, 'Succeeded', 'SucceededWithIssues', 'Skipped')
   pool:
     name: Package ES Lab E
 
--- a/build/ProjectReunion-release.yml
+++ b/build/ProjectReunion-release.yml
@@ -72,9 +72,15 @@ jobs:
     doCodeSign: 'true'
 
 - job: SignBinariesAndPublishSymbols
-  dependsOn: PublishMRT
-  condition:
-    in(dependencies.PublishMRT.result, 'Succeeded', 'SucceededWithIssues', 'Skipped')
+  dependsOn:
+    - Build
+    - PublishMRT
+  condition: |
+    and
+    (
+      in(dependencies.PublishMRT.result, 'Succeeded', 'SucceededWithIssues', 'Skipped'),
+      in(dependencies.Build.result, 'Succeeded', 'SucceededWithIssues', 'Skipped')
+    )
   pool:
     name: Package ES Lab E
 
--- a/build/SignConfig.xml
+++ b/build/SignConfig.xml
@@ -2,13 +2,13 @@
 <SignConfigXML>
   <job platform="" configuration="" dest="__INPATHROOT__" jobname="dep.controls" approvers="">
     <!-- Overwrite the input files to simplify downstream things like nuget and framework package generation. -->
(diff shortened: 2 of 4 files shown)


build/ProjectReunion-release.yml
```diff
@@ -72,9 +72,9 @@ jobs:
     doCodeSign: 'true'
 
 - job: SignBinariesAndPublishSymbols
-  dependsOn: Build
+  dependsOn: PublishMRT
```
huichen123, line 75: >PublishMRT [](start = 13, length = 10)

should also depend on Build

build/SignConfig.xml
```diff
@@ -2,13 +2,13 @@
...
-    <file src="__INPATHROOT__\Release\arm\Microsoft.ProjectReunion\Microsoft.ProjectReunion.dll" signType="WindowsRT" dest="__INPATHROOT__\Release\arm\Microsoft.ProjectReunion\Microsoft.ProjectReunion.dll" />
-    <file src="__INPATHROOT__\Release\arm\Microsoft.ProjectReunion\Microsoft.ProjectReunion.winmd" signType="WindowsRT" dest="__INPATHROOT__\Release\arm\Microsoft.ProjectReunion\Microsoft.ProjectReunion.winmd" />
-    <file src="__INPATHROOT__\Release\arm64\Microsoft.ProjectReunion\Microsoft.ProjectReunion.dll" signType="WindowsRT" dest="__INPATHROOT__\Release\arm64\Microsoft.ProjectReunion\Microsoft.ProjectReunion.dll" />
-    <file src="__INPATHROOT__\Release\arm64\Microsoft.ProjectReunion\Microsoft.ProjectReunion.winmd" signType="WindowsRT" dest="__INPATHROOT__\Release\arm64\Microsoft.ProjectReunion\Microsoft.ProjectReunion.winmd" />
+    <file src="__INPATHROOT__\runtimes\win10-x86\native\Microsoft.ApplicationModel.Resources.dll" signType="WindowsRT" dest="__INPATHROOT__\runtimes\win10-x86\native\Microsoft.ApplicationModel.Resources.dll" />
```
huichen123, line 5: ><file [](start = 4, length = 5)

compare with Singconfig.xml for MRT, it misses a few files.

EHO-makai: right, looks like this needs mrm.dll at minimum.  Looks good overall, though, thanks.

getrou: We don't need to sign the lib or pdb files correct?

EHO-makai: yes, that is correct, just dll and winmd.

EHO-makai: @huichen123, is mrm.dll supposed to be in this list?

build/ProjectReunion-release.yml
getrou: I believe Build just builds the Project Reunion sample code project that we aren't using anymore.  We aren't signing or publishing any of the output of Build in signconfig.xml I think, so we should be good.

huichen123: yes, it's ok for now, but it won't be always empty. :)

build/SignConfig.xml
huichen123: yes, all dlls need signed. still missing winmd and Microsoft.ApplicationModel.Resources.Projection.dll.

build/ProjectReunion-release.yml
getrou: Okay, that makes sense.  I'll have it depend on both.

build/SignConfig.xml
getrou: It should have all the files now, the winmd and both copies of the projection dlls.

huichen123 (approved):

EHO-makai (approved):
Make sure mrm.dll is in the signconfig.xml before completing the PR, thanks. :-)

PR #252, opened 2020-10-23
User/getrou/fwp separate appx

getrou:
Separates the "fat" nuget package into a thin nuget package and an appx.


tools/MakeAppxHelper.cmd
```diff
@@ -29,15 +29,14 @@ if "%TFS_BUILDCONFIGURATION%" EQU "" (
...
 set BasePackageName=Microsoft.ProjectReunion
 
-echo BUILDOUTPUT_OVERRIDE = %BUILDOUTPUT_OVERRIDE%
-if "%BUILDOUTPUT_OVERRIDE%" == "" (
-	set InputDirectory=%CD%\..\BuildOutput\%TFS_BUILDCONFIGURATION%\%TFS_PLATFORM%\ProjectReunion_dll
-	set OutputDirectory=%CD%\..\BuildOutput\%TFS_BUILDCONFIGURATION%\%TFS_PLATFORM%\FrameworkPackage
-) else (
-	set InputDirectory=%BUILDOUTPUT_OVERRIDE%\ProjectReunion_dll
-	set OutputDirectory=%BUILDOUTPUT_OVERRIDE%\FrameworkPackage
+if "%BUILDOUTPUT_OVERRIDE%" EQU "" (
```
jorvanMS, line 32: Can this happen? How come we don't set it anymore if it does happen?

jorvanMS (changes requested):
Just curious about the one change to "%BUILDOUTPUT_OVERRIDE%" EQU ""

dev/MRTCore/azure-pipelines.yml
```diff
@@ -284,112 +284,121 @@ jobs:
...
+      flattenFolders: true
+
+  - task: CopyFiles@2
+    displayName: 'copy winmd'
+    inputs:
+      SourceFolder: '$(Build.ArtifactStagingDirectory)\mrt_raw\mrtcore_binaries\lib\anycpu'
+      Contents: |
+        Microsoft.ApplicationModel.Resources.winmd
+      TargetFolder: '$(Build.ArtifactStagingDirectory)\fatnuget\lib\uap10.0\'
       flattenFolders: true
```
huichen123, line 308: why do this again?

dev/MRTCore/packaging/native/Microsoft.ApplicationModel.Resources.props
```diff
@@ -5,5 +5,41 @@
...
+      <Publisher>'CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US'</Publisher>
+    </AppxPackageRegistration>
+  </ItemGroup>
+  <ItemGroup>
+    <AppxPackageRegistration Include="$(MSBuildThisFileDirectory)..\tools\AppX\arm64\Release\Microsoft.ProjectReunion.0.1.appx">
+      <Architecture>arm64</Architecture>
+      <Version>$(MicrosoftProjectReunionAppxVersion)</Version>
+      <Publisher>'CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US'</Publisher>
+    </AppxPackageRegistration>
+  </ItemGroup> -->
```
huichen123, line 43: do we need this?

tools/MakeAppxHelper.cmd
getrou: It doesn't happen anywhere in our pipeline.  The only place we were calling this, we set the environment variable immediately beforehand, so I just got rid of the default option to ensure that still happened.

dev/MRTCore/azure-pipelines.yml
getrou: Ah, copy paste error from when I was assembling both the fat nuget and the thin nuget in this file instead of the CreateNuget yaml file.

tools/MakeAppxHelper.cmd
jorvanMS: Just seems like a weird check if we know we are setting it immediately beforehand.  I guess exiting gracefully is a better option than an exception.

jorvanMS (approved):

build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
```diff
@@ -54,8 +58,41 @@ jobs:
...
+      flattenFolders: false
+
+  # Extract framework package from fat nuget package
+  - task: CopyFiles@2
+    displayName: 'Extract Framework Package from Fat Nuget'
+    inputs:
+      SourceFolder: '${{ parameters.fatnupkgdir }}'
+      Contents: |
+        runtimes\**
+        lib\uap10.0\**
```
huichen123, line 92: >uap10.0 [](start = 12, length = 7)

also need net5.0-windows folder

dev/MRTCore/packaging/native/Microsoft.ApplicationModel.Resources.props
getrou: There's an ongoing conversation about how to deal with this.  This AppxPackageRegistration is important for any VS project using this nuget, but the file path points to a local appx on the machine building the VS project, so we can't anticipate what that would be here.  So I left the commented out template here for when we eventually want to use it.

We may end up including the appx in the FWP, in which case we will uncomment this out.  But for now I think we should leave it in.

As an aside, this will really end up going into the ProjectReunion props file, not the MRTCore props file.  But that will come with a separate change once we actually start using the ProjectReunion props file.

build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
getrou: The net5.0 stuff goes into both the AppX and the thin nuget package?  That doesn't seem correct.  I actually didn't even want to put the winmds in here originally, since it's included in the thin nuget package.

build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
```diff
@@ -54,7 +56,12 @@ jobs:
 
   - task: DownloadBuildArtifacts@0 
     inputs: 
-      artifactName: drop
+      artifactName: thin_nuget
```
eliezerpMS: >thin_nuget [](start = 20, length = 10)

Nit: I assume "thin nuget" is a term we came up with internally (nothing comes up on google at least). We should probably find a good place in this repo to put a readme about what this refers too (no need to necessarily do it in this PR though)

eliezerpMS (approved):
:shipit:

build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
getrou: Correct, Eho coined those terms.  I expect before we ship we will want to come up with a nicer term than "thin" and "fat" anyways.  I was going to table that decision until we finalize on whether the AppX is actually going to ship inside the Thin Nuget or separately.

build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
huichen123: I assume you either include both or neither.

---
In reply to: [511201967](https://github.com/microsoft/ProjectReunion/pull/252#discussion_r511201967) [](ancestors = 511201967)

huichen123 (approved):

eliezerpMS (approved):
:shipit:

PR #257, opened 2020-10-27
Fixing release pipeline now that Nuget package is split into "fat" and "thin" nuget packages

getrou:
Summary:
This fixes the signing error in the release pipeline.

Details:
PR https://github.com/microsoft/ProjectReunion/pull/252 split the Nuget package into a "fat" package with all the dlls intact, and a "thin" package with all the dlls removed. This worked fine for the CI pipeline, but broke the release pipeline due to a change in artifact names.  This change ports that artifact name change to the release pipeline.

### DIFF
--- a/build/ProjectReunion-release.yml
+++ b/build/ProjectReunion-release.yml
@@ -96,7 +96,7 @@ jobs:
     condition:
       eq(variables['useBuildOutputFromBuildId'],'')
     inputs:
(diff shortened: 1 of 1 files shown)


getrou:
Folding this check-in into a different PR that it would conflict with.
https://github.com/microsoft/ProjectReunion/pull/258

PR #258, opened 2020-10-27
Rename nuget package to "full"

getrou:
Summary:
Mostly a cosmetic change.  We in are in the process of coming up with real names for the different types of nuget packages, but until then, "full" is a better temporary name than "fat".

This changes the yml files, directory structure, and artifact name to reflect that.


EHO-makai (approved):
I assume you will handle any impact on the release pipeline in that PR (if necessary).

PR #266, opened 2020-10-29
remove SignMRT task

huichen123:
Signing is done in Reunion pipeline


dev/MRTCore/README.md
```diff
@@ -56,7 +56,7 @@ operates as your consent to these practices.
 Microsoft and any contributors grant you a license to the Microsoft documentation and other content
 in this repository under the [Creative Commons Attribution 4.0 International Public License](https://creativecommons.org/licenses/by/4.0/legalcode),
  and grant you a license to any code in the repository under the [MIT License](https://opensource.org/licenses/MIT), see the
-[LICENSE](LICENSE.txt) file.
+[LICENSE](../../LICENSE) file.
```
EHO-makai, line 59: (non-blocking), we need to get the ;licenses straight, Project Reunion PM is driving a conversation with CELA, I will make sure you and Axel and your PM are involved.

EHO-makai (approved):

dev/MRTCore/README.md
huichen123: @EHO-makai I will merge the change. Can you run the release pipeline so that it doesn't break it unless you think this is not a concern?

EHO-makai: I will run the pipeline, I think it will be fine, but better to be sure.

PR #267, opened 2020-10-29
set TargetPlatformMinVersion to 10.0.17134.0

huichen123:
WinUI needs to support RS4

### DIFF
--- a/dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/projection/Microsoft.ApplicationModel.Resources.Projection.csproj
+++ b/dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/projection/Microsoft.ApplicationModel.Resources.Projection.csproj
@@ -2,6 +2,7 @@
   
   <PropertyGroup>
     <TargetFrameworks>net5.0-windows10.0.18362</TargetFrameworks>
+    <TargetPlatformMinVersion>10.0.17134.0</TargetPlatformMinVersion>
     <Platforms>x64;x86</Platforms>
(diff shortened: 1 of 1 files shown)


dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/projection/Microsoft.ApplicationModel.Resources.Projection.csproj
```diff
@@ -2,6 +2,7 @@
   
   <PropertyGroup>
     <TargetFrameworks>net5.0-windows10.0.18362</TargetFrameworks>
```
alwu-msft, line 4: Shouldn't this be `TargetFramework` (no `s`)?

alwu-msft:
You can verify this on your end by opening the interop assembly in a disassembler (e.g. ILSpy) and checking that it has the following pair of asssembly-level attributes:
```csharp
[assembly: TargetPlatform("Windows10.0.18362.0")]
[assembly: SupportedOSPlatform("Windows10.0.17134.0")]
```

alwu-msft (approved):

MikeHillberg:
Whoa, who/what looks at those attributes?

Do we need this for Preview3?

alwu-msft:
It's a new code analyzer in .NET 5 (https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1416). We got tripped up by it ourselves, and setting the TargetPlatformMinVersion on the interop assembly is a change I'm making as part of our final ingestion.

PR #277, opened 2020-11-05
[MRTCore] Call GetPackagingOutputs on referenced UWP projects.

rohanp-msft:
Copy paste of code comment:
The referenced project may not be using MRTCore. In that case, `GetMrtPackagingOutputs` won't be defined. However, if the project is of type UWP, it'll have `GetPackagingOutputs` defined. So, try calling `GetPackagingOutputs`. If it does not exist, we simply no-op - see the `SkipNonexistentTargets` parameter below.

Note: we renamed `GetPackagingOutputs` to `GetMrtPackagingOutputs` in MRTCore because .NET 5 calling `GetPackagingOutputs` resulted in issues. Us calling `GetPackagingOutputs` in a .NET 5 project though should be fine.

How verified:
Resources from the core project of the Adobe app now show up in the top-level resources.pri.

### DIFF
--- a/dev/MRTCore/packaging/Microsoft.ApplicationModel.Resources.PriGen.targets
+++ b/dev/MRTCore/packaging/Microsoft.ApplicationModel.Resources.PriGen.targets
@@ -2104,14 +2104,31 @@
                  and '%(ProjectReferenceWithConfiguration.ReferenceOutputAssembly)' == 'true'"
       SkipNonexistentTargets="true"
       ContinueOnError="$(_ContinueOnError)">
-      <Output TaskParameter="TargetOutputs" ItemName="_PackagingOutputsFromOtherProjects"/>
+      <Output TaskParameter="TargetOutputs" ItemName="_PackagingOutputsFromOtherMrtCoreProjects"/>
+    </MSBuild>
+
+    <!-- The referenced project may not be using MRTCore. In that case, GetMrtPackagingOutputs won't be defined. However, if the
+         project is of type UWP, it'll have GetPackagingOutputs defined. So, try calling GetPackagingOutputs. If it does not exist,
+         we simply no-op - see the SkipNonexistentTargets parameter below. -->
+    <MSBuild
+      Projects="@(ProjectReferenceWithConfiguration)"
+      Targets="GetPackagingOutputs"
+      BuildInParallel="$(BuildInParallel)"
+      Properties="%(ProjectReferenceWithConfiguration.SetConfiguration); %(ProjectReferenceWithConfiguration.SetPlatform)"
+      Condition="'@(ProjectReferenceWithConfiguration)' != ''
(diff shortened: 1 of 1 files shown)


dev/MRTCore/packaging/Microsoft.ApplicationModel.Resources.PriGen.targets
```diff
@@ -2104,14 +2104,31 @@
...
+
+    <!-- The referenced project may not be using MRTCore. In that case, GetMrtPackagingOutputs won't be defined. However, if the
+         project is of type UWP, it'll have GetPackagingOutputs defined. So, try calling GetPackagingOutputs. If it does not exist,
+         we simply no-op - see the SkipNonexistentTargets parameter below. -->
+    <MSBuild
+      Projects="@(ProjectReferenceWithConfiguration)"
+      Targets="GetPackagingOutputs"
+      BuildInParallel="$(BuildInParallel)"
+      Properties="%(ProjectReferenceWithConfiguration.SetConfiguration); %(ProjectReferenceWithConfiguration.SetPlatform)"
+      Condition="'@(ProjectReferenceWithConfiguration)' != ''
```
sbanni, line 2118: You should condition this so it doesn't get run if the Mrt target is run, in the case of UWP + Mrt.Pri nupkg reference

alwu-msft (approved):
I think this looks good. Be sure to implement (and verify) @sbanni 's suggested change.

dev/MRTCore/packaging/Microsoft.ApplicationModel.Resources.PriGen.targets
rohanp-msft: Currently, this combination is not possible. WinUI 3.0, though usable in a UWP, won't import MRTCore. When used directly, MRTCore will noop in a UWP. Supposing this were possible though, yes, both targets will be defined. In that case, there's no harm, is there? We'll just get duplicate entries which will then be de-duped? We should still do this, just trying to understand the situation if we don't.

alwu-msft: My thinking is that we know that the targets are intended for mutually exclusive scenarios so it's safer to enforce that (by using a condition to ensure that the second is run only if the first did not run i.e. yielded no results) rather than have an implicit dependency on some later step. Additionally, running the second target unnecessarily results in extra work that then slows down the build.

rohanp-msft: Talked offline. Decision was made to defer addressing this currently-disabled scenario.

rohanp-msft:
> I think this looks good. Be sure to implement (and verify) @sbanni 's suggested change.

rohanp-msft:
Accidental PR close.... I then re-opened it. Will be merging once the CI pipeline run completes.

PR #282, opened 2020-11-07
User/jorvan/mrtcore helix

jorvanMS:
Summary: this PR runs the MRTCore tests on the helix queues.  It also sets up the helix infrastructure to run multiple work items when we are ready to onboard other TAEF tests.

Details: I removed the sample test project we were running on helix queues. I set up the payload to include the necessary files to allow the MRTCore tests to pass.  The helix project file can have work items added to it to onboard more TAEF tests to run in parallel on the helix queues.  Building the other two tests projects are currently commented out until we can do the work to convert these projects over to TAEF.

Watch out for: I saw a transient failure on running all 268 tests and pass but the task that parses the results was reporting 0 tests run. This was causing the pipeline to fail in the case it should pass.


build/Helix/PrepareHelixPayload.ps1
```diff
@@ -26,10 +26,18 @@ cp build/Helix/ConvertWttLogToXUnit.cs $payloadDir
 
 cp build/Helix/ConvertWttLogToXUnit.ps1 $payloadDir
 
-cp $Platform-$Configuration-taeftest/ProjectReunionHelix.dll $payloadDir
+cp BuildOutput/$Configuration/$Platform/BaseUnitTests/BaseUnitTests.dll $payloadDir
```
EHO-makai, line 29: Do you feel OK about hard-wiring this here? (question only, not blocking comment)

EHO-makai (approved):

build/Helix/PrepareHelixPayload.ps1
jorvanMS: I am okay with it for now. For example, WinUI3 uses another task to gather ALL of the test binaries and bin place them in one place along with the rest of the payload for the helix machine.  We really only have the one test binary right now but this is open to include others.  Regardless of which way we chose to do this we will have to modify some task/powershell script that gathers everything together into one directory before sending it over to the helix machine.  I think this script is self descriptive and easy to understand what it is doing.  It's also simple enough to add to it when we need to add more tests and more stuff will have to be included with the payload.  

I think the downside is if we update the TAEF redist version or move anything around this will break our testing pipeline. But we will be notified about that before a PR is checked in.

huichen123:
what's the benefit using Helix?

build/AzurePipelinesTemplates/ProjectReunion-RunHelixTests-Job.yml
```diff
@@ -59,12 +58,43 @@ jobs:
       feedsToUse: config
       nugetConfigPath: nuget.config
       restoreDirectory: packages
+      
+
+  - task: 333b11bd-d341-40d9-afcf-b32d5ce6f23b@2
+    displayName: 'NuGet restore MRTCore'
+    inputs:
+      command: 'custom'
+      arguments: 'restore dev\MRTCore\mrt\core\src\packages.config -ConfigFile nuget.config -PackagesDirectory dev\MRTCore\mrt\packages'
```
huichen123, line 67: >core\src\packages.config [](start = 42, length = 24)

if your only purpose is to restore TAEF, you picked the wrong one (and made unnecessary update to it)

build/AzurePipelinesTemplates/ProjectReunion-RunHelixTests-Job.yml
```diff
@@ -59,12 +58,43 @@ jobs:
...
+    displayName: Restore NuGet Packages for Taef
+    inputs:
+      restoreSolution: build/Helix/packages.config
+      feedsToUse: config
+      nugetConfigPath: nuget.config
+      restoreDirectory: dev\MRTCore\mrt\packages
+      
   - task: VSBuild@1
-    displayName: Build solution ProjectReunionHelix.sln
+    displayName: Build solution UnitBaseTest
```
huichen123, line 78: >UnitBaseTest [](start = 32, length = 12)

BaseUnitTest

build/AzurePipelinesTemplates/ProjectReunion-RunHelixTests-Job.yml
```diff
@@ -14,15 +14,15 @@ parameters:
   matrix: 
     Release_x86:
       buildPlatform: 'x86'
-      buildConfiguration: 'release'
+      buildConfiguration: 'Release'
       openHelixTargetQueues: 'windows.10.amd64.client19h1.open.xaml%3bwindows.10.amd64.clientrs3.open.xaml'
-      outputDirectory: BuildOutput\Release\x86\ProjectReunionHelix\
+      outputDirectory: BuildOutput\Release\x86\BaseUnitTests\
```
huichen123, line 19: >BaseUnitTests [](start = 47, length = 13)

if we need to do this for every unit tests, it's lots of extra work.

build/Helix/PrepareHelixPayload.ps1
```diff
@@ -26,10 +26,18 @@ cp build/Helix/ConvertWttLogToXUnit.cs $payloadDir
...
 
-cp build/Helix/packages/Taef.Redist.Wlk.10.57.200918006-develop/build/Binaries/$Platform/* $payloadDir
+#cp BuildOutput/$Configuration/$Platform/UnitTest/UnitTest.dll $payloadDir
 
-cp build/Helix/packages/Taef.Redist.Wlk.10.57.200918006-develop/build/Binaries/$Platform/CoreClr/* $payloadDir
+cp dev/MRTCore/mrt/packages/Taef.Redist.Wlk.10.57.200918006-develop/build/Binaries/$Platform/* $payloadDir
```
huichen123, line 33: >Taef.Redist.Wlk.10.57.200918006- [](start = 28, length = 32)

this is fragile

EHO-makai:
> what's the benefit using Helix?

As we move forward into Arcade/Maestro as our ES infrastructure (channels, subscriptions, etc). testing is done in this system using Helix.  So it's the long-term future.  In the immediate term, it's the same system that is used by WinUI, which will be onboarding in early 2021.

build/ProjectReunion-CI.yml
```diff
@@ -37,10 +38,16 @@ jobs:
...
+- template: ..\dev\MRTCore\azure-pipelines.yml
+  parameters:
+    buildJobName: 'BuildMRTCore'
+    
 - template: AzurePipelinesTemplates\ProjectReunion-RunHelixTests-Job.yml
   parameters:
     name: 'RunTestsInHelix'
-    dependsOn: Build
+    dependsOn: 
+      - Build
```
huichen123, line 50: >Build [](start = 8, length = 5)

MRTCore is not part of Build

dev/MRTCore/mrt/Core/src/packages.config
```diff
@@ -1,4 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="Microsoft.Windows.ImplementationLibrary" version="1.0.200519.2" targetFramework="native" />
+  <package id="TAEF.Redist.Wlk" version="10.57.200918006-develop" targetFramework="native" />
+  <package id="Microsoft.Windows.CppWinRT" version="2.0.200729.8" targetFramework="native" />
```
huichen123, line 5: should not change this file.

build/Helix/PrepareHelixPayload.ps1
EHO-makai: Agreed, but at least it is easy to find by whoever is updating the dependency feed ...

huichen123:
we are running the same tests twice now.

build/AzurePipelinesTemplates/ProjectReunion-RunHelixTests-Job.yml
jorvanMS: The purpose is to restore Microsoft.Windows.ImplementationLibrary which is necessary to build the test projects.

build/AzurePipelinesTemplates/ProjectReunion-RunHelixTests-Job.yml
jorvanMS: Not quite sure what you mean here.  For every unit test we will need to build both flavors and run them on the helix queue.

build/Helix/PrepareHelixPayload.ps1
jorvanMS: I agree, I updated my comment above with how WinUI3 handles this and I think it is probably a good solution we can update this going forward instead.

build/ProjectReunion-CI.yml
jorvanMS: Sure, but in the future as we start to add more tests, projects, etc,  I will imagine we will want our tests to wait on the build of the projectreunion.dll

jorvanMS:
Are we not running x86 tests and x64 tests?  I am building both flavors and sending them to run separately on the helix queues.  Is this what we want?

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

dev/MRTCore/mrt/Core/src/packages.config
jorvanMS: I think if we add a nuget feed/config for helix dependencies we can remove this and add that there. This would also clean up the fragile path that includes the TAEF redist version in it.

build/AzurePipelinesTemplates/ProjectReunion-RunHelixTests-Job.yml
huichen123: then you are looking for mrt\mrm\unittests\packages.config

---
In reply to: [520095238](https://github.com/microsoft/ProjectReunion/pull/282#discussion_r520095238) [](ancestors = 520095238)

build/ProjectReunion-CI.yml
huichen123: yes, it should depend on ALL build jobs (which have tests)

---
In reply to: [520097813](https://github.com/microsoft/ProjectReunion/pull/282#discussion_r520097813) [](ancestors = 520097813)

huichen123:
I meant we run the test in both VSTest task and Helix.

---
In reply to: [724260749](https://github.com/microsoft/ProjectReunion/pull/282#issuecomment-724260749) [](ancestors = 724260749)

dev/MRTCore/mrt/Core/src/packages.config
huichen123: you picked wrong packages.config

---
In reply to: [520104696](https://github.com/microsoft/ProjectReunion/pull/282#discussion_r520104696) [](ancestors = 520104696)

build/ProjectReunion-CI.yml
jorvanMS: MRTCore takes ~15 minutes to build and we can build our tests separately from it, they do not depend on that finishing and don't use any of the binaries from that job because we specifically skip over the test binaries.  

I rather get the tests out to the helix queues asap. Realistically we can remove the depends on main build but we will have to put it back eventually.

dev/MRTCore/mrt/Core/src/packages.config
jorvanMS: Thanks I fixed this.

jorvanMS:
Sorry I am not familiar, where is the VSTest task being run?

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

build/ProjectReunion-CI.yml
huichen123: how do you run MRTCore tests without waiting for MRTCore finished to build?

---
In reply to: [520168205](https://github.com/microsoft/ProjectReunion/pull/282#discussion_r520168205) [](ancestors = 520168205)

huichen123:
in mrtcore\azure-pipelines.yml

---
In reply to: [724323425](https://github.com/microsoft/ProjectReunion/pull/282#issuecomment-724323425) [](ancestors = 724323425)

jorvanMS:
So the task “Test MRT” in this pipeline is running the same BaseUnitTest tests through the VSTest task.  Should this be removed if we are moving this to helix? Are we not able to run the other two test projects with the VSTest task? Or is it TAEF only as well.

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

jorvanMS:
We build the BaseUnitTest project and we send that and all of it dependencies over to run on the helix queue.  We don’t use any of the artifacts from the previous jobs.  When I was running the pipeline I saw that Build MRTCore specifically skips over building the BaseUnitTest project.  In the more recent ones (assuming after I did a merge) I am seeing it being built.  Did we recently change this?

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

build/ProjectReunion-CI.yml
jorvanMS: Sorry I am getting lost between the email comments and the ones built in here... I guess they are not compatible.

When I was originally looking at the build MRTCore pipeline I thought I saw some logs that said something like "BaseUnitTests are not built in this configuration" Maybe I was just looking at the wrong flavor? In the test pipeline we build BaseUnitTests ourselves and then send that and their dependencies over to the helix queue.  If that is being built in the MRTCore pipeline we can just grab those artifacts that get published and depend on it finishing instead.

jorvanMS: So looking at this further, it is definitely more efficient to just split these up regardless. It will save on pipeline runtime if we can build MRTCore and the tests in parallel.  That way we can start running the tests on helix as soon as they are built (3 minute build time) instead of waiting for all of MRTCore to build (15 minute build time). 

I will defer to you what you think is best.

jorvanMS:
> So the task “Test MRT” in this pipeline is running the same BaseUnitTest tests through the VSTest task. Should this be removed if we are moving this to helix? Are we not able to run the other two test projects with the VSTest task? Or is it TAEF only as well. 

I think we can just skip the build/test task in the MRTCore pipeline.  We can do it in parallel in the Helix test job and have the pipeline run more efficiently.

PR #284, opened 2020-11-10
Copy the full package target files into the full package

eliezerpMS:
This change copies the full package targets into the full package that will be pushed to the internal repo. The main thing needed here is the references to the project reunion winmds in the targets file.

### DIFF
--- a/build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
+++ b/build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
@@ -100,7 +100,7 @@ jobs:
     displayName: 'Copy Project Reunion full package targets'
     inputs:
       targetType: inline
-      script: >
+      script: |
       $fullpackagePath = "${{ parameters.fullnupkgdir }}"
       $targetsFilePath = "$(Build.SourcesDirectory)\build\NuSpecs"
       Copy-Item -Path "$targetsFilePath\ProjectReunion-Nuget-Native.targets" -Destination "$fullpackagePath\build\native\Microsoft.ProjectReunion.Foundation.targets"
--- a/build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
+++ b/build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
@@ -95,12 +95,16 @@ jobs:
       TargetFolder: '$(Build.ArtifactStagingDirectory)\AppXContents'
       flattenFolders: false
 
-  - powershell: |
+  # Copy the Project Reunion full package specific target files into the full package folder
+  - task: PowerShell@2
+    displayName: 'Copy Project Reunion full package targets'
+    inputs:
+      targetType: inline
+      script: >
       $fullpackagePath = "${{ parameters.fullnupkgdir }}"
       $targetsFilePath = "$(Build.SourcesDirectory)\build\NuSpecs"
       Copy-Item -Path "$targetsFilePath\ProjectReunion-Nuget-Native.targets" -Destination "$fullpackagePath\build\native\Microsoft.ProjectReunion.Foundation.targets"
       Copy-Item -Path "$targetsFilePath\ProjectReunion-Nuget-Common.targets" -Destination "$fullpackagePath\build\Microsoft.ProjectReunion.targets"
-    displayName: 'Copy Project Reunion full package targets'
 
   # debugging - remove or comment out before completing PR      
   - script: |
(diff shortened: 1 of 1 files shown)


build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
```diff
@@ -95,6 +95,13 @@ jobs:
       TargetFolder: '$(Build.ArtifactStagingDirectory)\AppXContents'
       flattenFolders: false
 
+  - powershell: |
```
jorvanMS, line 98: nit:  I have been using this syntax for inline powershell tasks:

  - task: PowerShell@2
    displayName: "displayname here"
    inputs:
      targetType: inline
      script: >
              "your script goes here"

EHO-makai: +1

build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
```diff
@@ -95,6 +95,13 @@ jobs:
...
+      $fullpackagePath = "${{ parameters.fullnupkgdir }}"
+      $targetsFilePath = "$(Build.SourcesDirectory)\build\NuSpecs"
+      Copy-Item -Path "$targetsFilePath\ProjectReunion-Nuget-Native.targets" -Destination "$fullpackagePath\build\native\Microsoft.ProjectReunion.Foundation.targets"
+      Copy-Item -Path "$targetsFilePath\ProjectReunion-Nuget-Common.targets" -Destination "$fullpackagePath\build\Microsoft.ProjectReunion.targets"
```
EHO-makai, line 102: There's a Microsoft.ProjectReunion.Foundation.targets file also, are you purposefully excluding that?  It's a "meta" targets file that rings in MRRCore's targets file.

eliezerpMS: Yep I am purposefully excluding it. The original nuspec used the file like this <file target="build\native\$ID$.targets" src="ProjectReunion-Nuget-Native.targets"/> and so ProjectReunion-Nuget-Native.targets would become Microsoft.ProjectReunion.Foundation.targets.

---
In reply to: [520765659](https://github.com/microsoft/ProjectReunion/pull/284#discussion_r520765659) [](ancestors = 520765659)

build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
eliezerpMS: Yeah I did think this format was a bit odd, especially with the displayName at the end, but that is what the other examples had. I like your format better so I will switch to it.

---
In reply to: [520758307](https://github.com/microsoft/ProjectReunion/pull/284#discussion_r520758307) [](ancestors = 520758307)

jorvanMS (approved):

build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
mikenelte: Instead of doing this in powershell you could use the CopyFiles task, something like:

  - task: CopyFiles@2
    displayName: 'Copy Project Reunion full package targets'
    inputs:
      SourceFolder: '$(Build.SourcesDirectory)\build\NuSpecs'
      Contents: |
        ProjectReunion-Nuget-Native.targets
        ProjectReunion-Nuget-Common.targets
      TargetFolder: '${{ parameters.fullnupkgdir }}\build'
      flattenFolders: false

eliezerpMS: The above is not equivalent. Those files go in different folders and need to be renamed.

---
In reply to: [520918811](https://github.com/microsoft/ProjectReunion/pull/284#discussion_r520918811) [](ancestors = 520918811)

mikenelte: Is there any reason we don't check them into the repo with the names we want them to have?

eliezerpMS: There is already a file with the name Microsoft.ProjectReunion.targets that is being used for a different purpose. We should probably do cleanup and remove it if it is not going to be used but that is out of the scope of this change. This change is mostly just to unblock me.

---
In reply to: [520943020](https://github.com/microsoft/ProjectReunion/pull/284#discussion_r520943020) [](ancestors = 520943020)

mikenelte: ok - that makes sense for this change

PR #285, opened 2020-11-10
ESRP Code Signing

kythant:
Setting up ESRP CodeSigning


jorvanMS (approved):

build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
```diff
@@ -23,16 +22,6 @@ jobs:
     - ${{ parameters.dependsOn }}
   condition: ${{ parameters.condition }}
 
-  # XES tasks need to be on PackageES machines.
```
EHO-makai, line 26: :-)

build/ProjectReunion-release.yml
```diff
@@ -149,14 +142,57 @@ jobs:
...
-      SYSTEM_ACCESSTOKEN: $(system.accesstoken)
+  - task: EsrpCodeSigning@1
     inputs:
-      signConfigXml: '$(Build.SourcesDirectory)\build\SignConfig.xml'
-      inPathRoot: '$(Build.ArtifactStagingDirectory)\fullnuget'
-      outPathRoot: '$(Build.ArtifactStagingDirectory)\fullnuget'
+      ConnectedServiceName: 'Project Reunion ESRP Code Signing Connection'
+      FolderPath: '$(Build.ArtifactStagingDirectory)\fullnuget'
+      Pattern: |
+        lib\uap10.0\Microsoft.ApplicationModel.Resources.winmd
```
EHO-makai, line 150: (non-blocking for now), can we just say *.dll and *.winmd or does this not support wildcards?  If it doesn't support wildcards I'm wondering if we can take the artifact drop output and write a task that spits the dll and winmd names so we don't have to hard-code this.

EHO-makai (approved):

build/ProjectReunion-release.yml
kythant: It does support wild cards but since we were specifying it in build/signconfig.xml, I thought I should specify it here too.

kythant: Another thing is that since the new ESRP task does not support using sign config xml files, there will ne no need for build/SignConfig.xml and build/NugetSignConfig.xml. I will be removing it soon in the next commit.

build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
kythant: I don't think we need this task anymore for ESRP signing?

jorvanMS: I think because we are using ESRP now instead of PackageES codesign we don't need to be on packageES pool.

kythant: Yea I removed that but for some reason the CreateNugetPackage job refuse to run. 
https://dev.azure.com/microsoft/ProjectReunion/_build/results?buildId=28785553&view=logs&j=71b158ba-2b04-598f-7447-b2e90984d12a

I thought maybe the yml file was incorrect so it wouldn't run, so I added it back in to run again.

PR #288, opened 2020-11-12
User/mnelte/cleanup

mikenelte:
Rename Microsoft.ProjectReunion.Foundation to Microsoft.ProjectReunion.Foundation.Full


build/NuSpecs/Microsoft.ProjectReunion.Foundation.nuspec
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
      <metadata minClientVersion="3.3">
-         <id>Microsoft.ProjectReunion.Foundation</id>
+         <id>Microsoft.ProjectReunion.Foundation.Full</id>
```
getrou, line 4: Do we want to make this rename elsewhere as well?  For example, the nuspec/targets/props files?  Totally fine if the answer is no, just wondering.

mikenelte: no - because this is a temporary package.  The final package will be name Microsoft.ProjectReunion.Foundation

eliezerpMS: Maybe a stupid question but why does the Microsoft.ProjectReunion.Foundation "thin" package need to be in the internal feed?

If it does not then why not just name this Microsoft.ProjectReunion.Foundation?

---
In reply to: [522306530](https://github.com/microsoft/ProjectReunion/pull/288#discussion_r522306530) [](ancestors = 522306530)

EHO-makai (approved):
Thank you.

PR #295, opened 2020-11-17
Remove the Project Reunion Foundation CopyLocals for binaries

eliezerpMS:
This is a change to temporarily remove the CopyLocals for the Project Reunion Foundation binaries. This package is only ever used as a framework package and so these do not need to be here for the initial bring up of the pipelines. In the future we will have a more robust programmatic solution.

### DIFF
--- a/build/NuSpecs/ProjectReunion-Nuget-Native.targets
+++ b/build/NuSpecs/ProjectReunion-Nuget-Native.targets
@@ -9,7 +9,7 @@
     <Reference Include="$(MSBuildThisFileDirectory)..\..\lib\uap10.0\Microsoft.ProjectReunion.winmd">
       <Implementation>Microsoft.ProjectReunion.dll</Implementation>
     </Reference>
-	  <!--  TODO: Figure out the right long term approach for VS to not pick these up when using them with a framework package. -->
+	  <!--  Figure out the right long term approach for VS to not pick these up when using them with a framework package. -->
     <!-- <ReferenceCopyLocalPaths Include="$(MSBuildThisFileDirectory)..\..\runtimes\win10-$(Native-Platform)\native\Microsoft.ProjectReunion.dll" /> -->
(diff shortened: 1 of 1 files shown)


build/NuSpecs/ProjectReunion-Nuget-Native.targets
```diff
@@ -9,7 +9,8 @@
     <Reference Include="$(MSBuildThisFileDirectory)..\..\lib\uap10.0\Microsoft.ProjectReunion.winmd">
       <Implementation>Microsoft.ProjectReunion.dll</Implementation>
     </Reference>
-    <ReferenceCopyLocalPaths Include="$(MSBuildThisFileDirectory)..\..\runtimes\win10-$(Native-Platform)\native\Microsoft.ProjectReunion.dll" />
-    <ReferenceCopyLocalPaths Include="$(MSBuildThisFileDirectory)..\..\runtimes\win10-$(Native-Platform)\native\Microsoft.ProjectReunion.pri" />
+    <!--  Figure out the right long term approach for VS to not pick these up when using them with a framework package. -->
+    <!-- <ReferenceCopyLocalPaths Include="$(MSBuildThisFileDirectory)..\..\runtimes\win10-$(Native-Platform)\native\Microsoft.ProjectReunion.dll" /> -->
```
EHO-makai, line 13: Isn't deletion the long-term solution?  These shouldn't be in the targets file?

EHO-makai (approved):

build/NuSpecs/ProjectReunion-Nuget-Native.targets
eliezerpMS: This comment is more a note to myself/other people. In the future if we want to ingest packages that also are meant to be consumed not through a framework package then they will have these types of lines in their targets files. We will want a programmatic solution, whether that is having a script that edits the targets to remove these types of things or some VS props/targets file magic.

PR #317, opened 2020-11-30
Add <disabledPackageSources>

DrusTheAxe:
Add `<disabledPackageSources>`

Per latest guidance "How to Securely Configure Package Source Files" `nuget.config` should have `<packageSources>` and `<disabledPackageSources>` and start both with `<clear />`. Current `nuget.config` had no `<disabledPackageSources>`.

### DIFF
--- a/nuget.config
+++ b/nuget.config
@@ -1,5 +1,9 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
+  <disabledPackageSources>
+    <clear />
+  </disabledPackageSources>
+
   <packageSources>
     <clear />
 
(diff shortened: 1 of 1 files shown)


nuget.config
```diff
@@ -10,4 +10,8 @@
 
     <add key="localpackages" value="localpackages" />
   </packageSources>
-</configuration>
\ No newline at end of file
+
+  <disabledPackageSources>
+    <clear />
```
EHO-makai, line 15: what does <clear /> do?  I was reading up on disabledPackgeSources, primary reference here:
https://docs.microsoft.com/en-us/nuget/reference/nuget-config-file#disabledpackagesources
and it refers to leaving the "list" empty.

EHO-makai: Also, I read some documentation that says this should go first and that is basically where it is in the WinUI3 nuget.config.

DrusTheAxe: Moved the disabledPackageSources block above the packageSources block as suggested offline

EHO-makai (approved):

PR #328, opened 2020-12-05
FAQ: Update "What's a UWP" section

Felix-Dev:
Reading through the [What's a UWP](https://github.com/microsoft/ProjectReunion/blob/main/docs/faq.md#whats-a-uwp) section in the FAQ, I always felt a fundamental aspect of being a UWP app was missing - namely the fact that UWP is currently the only supported platform for Xbox, HoloLens, and Surface Hub applications. It's possible this might change in the future but it's [far from guaranteed](https://github.com/microsoft/ProjectReunion/issues/215#issuecomment-702243440) and in case it does happen it will likely be years from now. As such, I think it is worthwile to point out this characteristic of a UWP app compared to a Win32 app for us Windows developers.

#### Additional note
Indirectly related to this PR, perhaps the FAQ could also be extended with the question "Will Project Reunion enable my Win32 app to run on all Windows 10 devices including Xbox, HoloLens and Surface Hub?"? We have seen this question come up a few times already in one form or another (see [here](https://github.com/microsoft/ProjectReunion/issues/215#issuecomment-702044380) and [here](https://github.com/microsoft/ProjectReunion/issues/301) for example) so I think it makes sense to list this question *directly* in the FAQ, even though it already [appears to have been answered](https://github.com/microsoft/ProjectReunion/blob/main/docs/faq.md#is-reunion-a-new-app-model) at least indirectly.


docs/faq.md
```diff
@@ -115,7 +115,8 @@ We're looking to support:
 Referenced throughout this repo, the term "UWP" references a
 [Universal Windows Platform App](https://docs.microsoft.com/en-us/windows/uwp/get-started/) -
 an app running inside an app container, at a lower privilege level, with
-package identity, delivered via an MSIX. This app's lifecycle from activation
+package identity, delivered via an MSIX. It is the only app type supported on Windows devices 
+such as Xbox, HoloLens and Surface Hub (Windows 10 Team). This app's lifecycle from activation
```
Felix-Dev, line 119: Two remarks:
* I did not list Windows 10 IoT Core here since to my understanding that product is [currently in the process of being superceded by Windows 10 IoT Enterprise](https://blogs.windows.com/windowsdeveloper/2020/05/26/build-your-iot-devices-with-windows-for-iot-a-comprehensive-platform-for-every-device-developer/) - which supports Win32 apps. I am also not yet listing Windows 10X here since I don't yet know about the exact development details Windows 10X will arrive with.
* Windows 10 Team now supports apps like the new Edge browser based on Chromium or Microsoft Teams (see [here](https://docs.microsoft.com/en-us/surface-hub/surface-hub-2020-update-whats-new#additional-features)) which to my understanding aren't UWP apps even on the Surface Hub so that wording might be a bit imprecise ("only app type supported"). That said, at least commonly [UWP apps are the only ones being supported](https://techcommunity.microsoft.com/t5/surface-it-pro-blog/new-surface-hub-os-update-released-for-public-preview/bc-p/1552149/highlight/true#M313) on Windows 10 Team so the wording might be fine after all.

hansmbakker:
For raspberry pi-like devices there seems no path forward to Windows IoT (Enterprise) given by Microsoft so please do include Windows IoT Core

Felix-Dev:
@hansmbakker Added Windows 10 IoT Core to the list.

asklar:
Fyi I wrote something similar/complementary for react native for Windows apps (which are UWPs)
https://microsoft.github.io/react-native-windows/docs/next/native-code

docs/faq.md
jonwis: Please move the "It is the only type..." to a separate paragraph. This paragraph should describe precisely what a UWP _is_ and put the _where it's useful_ separately.

docs/faq.md
```diff
@@ -115,9 +115,10 @@ We're looking to support:
 Referenced throughout this repo, the term "UWP" references a
 [Universal Windows Platform App](https://docs.microsoft.com/en-us/windows/uwp/get-started/) -
 an app running inside an app container, at a lower privilege level, with
-package identity, delivered via an MSIX. This app's lifecycle from activation
-through termination is controlled by the platform for best power and resource
-usage. Its access to the user's resources are gated by a Request model so the
+package identity, delivered via an MSIX. It is the only app type supported on Windows devices 
+such as Xbox, HoloLens, Surface Hub (Windows 10 Team) and devices running Windows 10 IoT Core.
```
jonwis, line 119: While this sentence is correct, please move it from here to a separate paragraph so the flow of "what defines the app type" isn't mixed with "where might the app type be useful."

jonwis (changes requested):

docs/faq.md
riverar: Nit: Surface Hub can be configured to run alternate Windows 10 SKUs that *do* allow for other app types.

I think the fact these restrictions are arbitrary and could be changed at any time, is a good reason it's not appropriate for inclusion in the document. It's more documentation debt.

ecovio1:
@Felix-Dev  
well, two types of apps will be here , **(1.)native modern windows apps(mix of both win32+winRT apis),,, (2.)PWA/Web apps.**

there's no place for a third.
**UWP itself will be redundant basically, once Windows App Container technology comes to all win32 apps.** 

therefore, I humbly suggest you to drop the PR altogether. thanks.

BenJKuhn:
Closing old or abandoned pull requests. If you would like to pick this up again, address feedback, and push it over the finish line go ahead and reactivate.

PR #333, opened 2020-12-08
Start Running Passing Win32 AppLifeCycle Tests in Foundation Pipeline

kythant:
Start Running Passing Win32 AppLifeCycle Tests in Foundation Pipeline

### DIFF
--- a/test/AppLifecycle/FunctionalTests.cpp
+++ b/test/AppLifecycle/FunctionalTests.cpp
@@ -89,38 +89,38 @@ namespace ProjectReunionCppTest
             VERIFY_IS_NULL(AppLifecycle::GetActivatedEventArgs());
         }
 
-        TEST_METHOD(GetActivatedEventArgsIsNotNull)
-        {
-            BEGIN_TEST_METHOD_PROPERTIES()
-                // Run this test for both PackagedWin32 and Win32.
-                TEST_METHOD_PROPERTY(L"RunAs", L"{UAP,InteractiveUser}")
+        // TEST_METHOD(GetActivatedEventArgsIsNotNull)
+        // {
+        //     BEGIN_TEST_METHOD_PROPERTIES()
+        //         // Run this test for both PackagedWin32 and Win32.
+        //         TEST_METHOD_PROPERTY(L"RunAs", L"{UAP,InteractiveUser}")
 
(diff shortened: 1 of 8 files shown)


test/AppLifecycle/FunctionalTests.cpp
```diff
@@ -60,8 +60,8 @@ namespace ProjectReunionCppTest
             {
                 DeleteContentFile(c_testDataFileName_Packaged);
```
aeloros, line 61: nit: This can be commented out as well if you want.

EHO-makai (commented):
Looks fine to me, Scott has the controlling approval.

test/AppLifecycle/FunctionalTests.cpp
```diff
@@ -89,38 +89,38 @@ namespace ProjectReunionCppTest
...
-        }
+        //     VERIFY_IS_NOT_NULL(AppLifecycle::GetActivatedEventArgs());
+        // }
 
-        TEST_METHOD(GetActivatedEventArgsForLaunch)
-        {
-            BEGIN_TEST_METHOD_PROPERTIES()
-                // Run this test for both PackagedWin32 and Win32.
-                TEST_METHOD_PROPERTY(L"RunAs", L"{UAP,InteractiveUser}")
+        // TEST_METHOD(GetActivatedEventArgsForLaunch)
```
aeloros, line 106: I don't think we should be commenting out this valid test.

aeloros (changes requested):

test/AppLifecycle/FunctionalTests.cpp
kythant: "// Run this test for both PackagedWin32 and Win32."

I just noticed this comment. These were one of the tests that was failing on the pipeline so I am guess maybe it failed for the packaged version. I will try disabling just the packaged portion.

aeloros:
Try changing the property instead.  You would change it to be the below.  This would change it to only run for Win32.  That’s a good point.  We may have issues in the pipeline that stop TAEF from using RunAs:UAP that need to be fixed (elevated privs?).


TEST_METHOD_PROPERTY(L"RunAs", L"InteractiveUser")

aeloros (approved):

PR #351, opened 2020-12-28
Fix broken npos check

nickrandolph:
Because 4 was being added to the result of rfind, the subsequent check for npos will always fail. This also means there is a risk later on of argsStart being > argsEnd. By adjusting the logic to add 4 after the npos check, the npos check will now work and the subsequent check for argsStart >argsEnd is no longer required.


dev/AppLifecycle/AppLifecycle.cpp
```diff
@@ -27,11 +29,6 @@ namespace winrt::Microsoft::ProjectReunion::implementation
             return {L"", L""};
         }
 
-        if (argsStart > argsEnd)
```
aeloros, line 30: Add this back in.

dev/AppLifecycle/AppLifecycle.cpp
```diff
@@ -14,11 +14,13 @@ namespace winrt::Microsoft::ProjectReunion::implementation
 {
     std::tuple<std::wstring, std::wstring> ParseCommandLine(std::wstring commandLine)
     {
-        auto argsStart = commandLine.rfind(L"----") + 4;
+        auto argsStart = commandLine.rfind(L"----");
         if (argsStart == std::wstring::npos)
```
aeloros, line 18: Thanks!

aeloros (changes requested):

riverar:
@nickrandolph Hey Nick! Can you push a new update to close out this PR?

aeloros:
I've moved the npos change to this branch, which will be going into main as soon as a dependent change is merged.

https://github.com/microsoft/ProjectReunion/blob/user/aeloros/startuptasks/dev/AppLifecycle/AppLifecycle.cpp

PR #357, opened 2021-01-04
remove VCRTForwarders from microsoft.applicationmodel.resources.dll

huichen123:


jonwis:
Should this be applied to all the built DLLs? @EHO-makai  or @BenJKuhn ?

huichen123:
> Should this be applied to all the built DLLs? @EHO-makai or @BenJKuhn ?

MRTCore builds 3 dlls: mrm.dll, m.a.r.dll and m.a.r.projection.dll. I did "link /dump /imports" on all 3 dlls. mrm.dll doesn't have dependency on vcruntime. m.a.r.dll is fixed with this change and switched from vcruntime140*_app.dll to vcruntime140*.dll. m.a.r.projection.dll is already dependent on vcruntime140*.dll.

jonwis:
I meant "all the other DLLs that Project Reunion produces" - is everybody using the right VCRT?

huichen123:
according to the email thread, the only other one is Microsoft.Web.WebView2.Core.dll.

---
In reply to: [754221380](https://github.com/microsoft/ProjectReunion/pull/357#issuecomment-754221380) [](ancestors = 754221380)

axelandrejs (approved):

stevenbrix:
> Should this be applied to all the built DLLs? @EHO-makai or @BenJKuhn ?

@jonwis, I do think all Project Reunion dll's should have `DesktopCompatible` set in the project file. The WinUI3 project templates for C++ components have this property set as well.

stevenbrix (approved):

jevansaks:
How do we ensure that a packaged app that consumes reunion depends on Microsoft.VCLibs.140.00.UWPDesktop instead of Microsoft.VCLibs.140.00? I would have expected some other targets file change updated with that package name.

jonwis:
@BenJKuhn or @EHO-makai - how do you want to handle "make this change across the Project Reunion codebase" style issues?

huichen123:
according to the task, WinUI is going to change Microsoft.WinUI.Appx.targets, and set <SDKReference Include="Microsoft.VCLibs.Desktop, Version=14.0" />

---
In reply to: [754272500](https://github.com/microsoft/ProjectReunion/pull/357#issuecomment-754272500) [](ancestors = 754272500)

PR #365, opened 2021-01-10
Delete sample apps since sample apps are to be in a separate repo.

rohanp-msft:


axelandrejs (approved):

huichen123:
we may need those later for some E2E testing.

PR #386, opened 2021-01-14
Copy winrt generated files for Microsoft.ApplicationModel.Resources into NuGet package.

rohanp-msft:
Copy winrt generated files for Microsoft.ApplicationModel.Resources into NuGet package.

Needed for C++/WinRT in Win32 apps.


dev/MRTCore/azure-pipelines.yml
```diff
@@ -376,6 +376,14 @@ jobs:
...
         mrm.h
       TargetFolder: '$(Build.ArtifactStagingDirectory)\fullnuget\include'
+
+  - task: CopyFiles@2
+    displayName: 'copy winrt generated files'
+    inputs:
+      SourceFolder: '$(MRTSourcesDirectory)\mrt\Microsoft.ApplicationModel.Resources\src\Generated Files'
+      Contents: |
+        *
+      TargetFolder: '$(Build.ArtifactStagingDirectory)\fullnuget\include'
```
huichen123, line 386: why?

axelandrejs (approved):

dev/MRTCore/azure-pipelines.yml
rohanp-msft: Needed for C++/WinRT in Win32 apps.

huichen123: doesn't feel right. winmd should be the only thing needed

rohanp-msft: "To successfully reference a C++ Windows Runtime component from a C++ app, you need to use C++/WinRT to generate projection header files of your component. You can then include these header files in your app code to call your component." https://blogs.windows.com/windowsdeveloper/2019/04/30/enhancing-non-packaged-desktop-apps-using-windows-runtime-components/

Sample app: https://github.com/microsoft/RegFree_WinRT/tree/master/Cpp

huichen123: that means whoever consumes the component (i.e. app developer) needs to generate the headers, not we need to provide them in nuget.

dev/MRTCore/packaging/native/Microsoft.ApplicationModel.Resources.WinRt.props
```diff
@@ -3,6 +3,9 @@
 -->
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
+  <!-- An empty '$(ApplicationType)' corresponds to a C++, Desktop app. -->
```
huichen123, line 6: >Desktop [](start = 59, length = 7)

do we really want to use "Desktop"? I hope we don't type app type to platform.

dev/MRTCore/azure-pipelines.yml
rohanp-msft: we no longer add the generated files to the nuget package. they are generated from the winmd on app build in the app folder.

dev/MRTCore/packaging/native/Microsoft.ApplicationModel.Resources.WinRt.props
rohanp-msft: using win32 as agreed upon offline

dev/MRTCore/azure-pipelines.yml
```diff
@@ -251,14 +251,26 @@ jobs:
...
-    displayName: 'copy winmd'
+    displayName: 'copy winmd into uap10.0 folder'
     inputs:
       SourceFolder: '$(Build.ArtifactStagingDirectory)\mrt_raw\mrtcore_binaries\lib\anycpu'
       Contents: |
         Microsoft.ApplicationModel.Resources.winmd
       TargetFolder: '$(Build.ArtifactStagingDirectory)\fullnuget\lib\uap10.0\'
       flattenFolders: true
 
+  # Needed so that the NuGet package can be installed in a win32 app. It avoids this error: "You are trying to
```
rohanp-msft, line 262: I understand that this winmd is not needed other than to avoid this error... Ideally, we can avoid this error without adding a copy of the winmd, but I don't know how to do that / if that is possible. The error is explicit...

huichen123 (approved):
:shipit:

huichen123 (changes requested):
the nuspec changes are probably incorrect. transport package is internal. why cannot i find EHO or Michael Nelte any more?

PR #394, opened 2021-01-22
Changes to generate a Microsoft.ProjectReunion.Foundation NuGet package that supports MRTCore on win32.

rohanp-msft:
Corresponding ADO change: https://microsoft.visualstudio.com/ProjectReunion/_git/ProjectReunion/pullrequest/5623504


wjk:
Question about this PR. I note that the MSBuild code in the MRT targets files currently present in the repo are designed to work in the context of a UWP app’s build system, where there are several PRI files, separated by qualifier, and where the PRI files are managed entirely by MSBuild.

However, my use-case for MRTCore is quite different. I create one PRI file per csproj, containing the resources needed by that assembly, with all resources all all qualifiers embedded inside that single file. I do this quite easily by manually calling `makepri.exe` with a custom `priconfig.xml` file, and by storing all the input files (resw, resjson, embedded binary, etc) in one subdirectory. Now that I understand how `priconfig.xml` files work, it’s very easy. However, since the MRTCore DLLs are not yet available on NuGet, I must vendor their sources into all of my solutions myself, and it’s a big chunk of code to copy around.

Would it be possible to be able to use an MRTCore NuGet package without enabling the UWP-oriented automatic management of these files as shown in this PR? Thanks!

wjk:
I am noting that I have since found a way to merge my manually constructed PRI files (as detailed in my post) above into the MSBuild-generated PRI files that this PR is concerned with. My issue from three days ago is now resolved. Thanks again!

rohanp-msft:
@huichen123 The Microsoft.ApplicationModel.Resources.Win32AppPropertySheet.props file isn't needed. There was a bug in Microsoft.ApplicationModel.Resources.WinRt.props: the winmd path was not right. Fixing that removed the need for Microsoft.ApplicationModel.Resources.Win32AppPropertySheet.props. I don't know how things worked before despite the winmd path being incorrect, but anyway, it's fixed now.

dev/MRTCore/azure-pipelines.yml
```diff
@@ -251,14 +251,26 @@ jobs:
...
-    displayName: 'copy winmd'
+    displayName: 'copy winmd into uap10.0 folder'
     inputs:
       SourceFolder: '$(Build.ArtifactStagingDirectory)\mrt_raw\mrtcore_binaries\lib\anycpu'
       Contents: |
         Microsoft.ApplicationModel.Resources.winmd
       TargetFolder: '$(Build.ArtifactStagingDirectory)\fullnuget\lib\uap10.0\'
       flattenFolders: true
 
+  # Needed so that the NuGet package can be installed in a win32 app. It avoids this error: "You are trying to
```
rohanp-msft, line 262: @huichen123 I know this isn't ideal... if you have a better idea, I'm all ears.

rohanp-msft:
> I am noting that I have since found a way to merge my manually constructed PRI files (as detailed in my post) above into the MSBuild-generated PRI files that this PR is concerned with. My issue from three days ago is now resolved. Thanks again!

Hello William, I am glad that your problem is resolved. To answer your earlier question, we currently don't have plans on enabling something different from automatic management of PRI generation (that is UWP-oriented, as you correctly pointed out). The reason for that is that, so far, we've not received enough feedback for us to think that's worthwhile. Note that the MRTCore DLLs in the NuGet package aren't related to PRI generation. They are for the MRTCore runtime, i.e., they support the various APIs you can call in your app code. If you have any other questions or comments, feel free to let us know!

axelandrejs (approved):

dev/MRTCore/packaging/native/Microsoft.ApplicationModel.Resources.WinRt.props
```diff
@@ -15,7 +15,7 @@
   <!-- OutputType Exe corresponds to C++. -->
   <ItemGroup Condition="'$(OutputType)' == 'Exe'">
     <Reference Include="Microsoft.ApplicationModel.Resources.winmd">
-      <HintPath>$(MSBuildThisFileDirectory)..\..\lib\uap10.0\Microsoft.ApplicationModel.Resources\Microsoft.ApplicationModel.Resources.winmd</HintPath>
```
huichen123, line 18: if you take a look at Microsoft.ProjectReunion.Foundation package, this is actually the current path. The reason to add an extra layer of folder is to avoid conflict with the file coming from WinUI.

dev/MRTCore/azure-pipelines.yml
```diff
@@ -250,13 +250,31 @@ jobs:
 
 # START of copying files into their final package location.
 
+  # Place Microsoft.ApplicationModel.Resources.winmd in the Microsoft.ApplicationModel.Resources\
```
rohanp-msft, line 253: @huichen123 the corresponding ADO change: https://microsoft.visualstudio.com/ProjectReunion/_git/ProjectReunion/pullrequest/5623504

huichen123 (approved):

eliezerpMS (approved):

PR #397, opened 2021-01-25
Add TryGetValue

huichen123:


dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/src/Microsoft.ApplicationModel.Resources.idl
```diff
@@ -54,6 +54,10 @@ namespace Microsoft.ApplicationModel.Resources
         IKeyValuePair<String, ResourceCandidate> GetValueByIndex(UInt32 index);
         [method_name("GetValueByIndexWithContext")]
         IKeyValuePair<String, ResourceCandidate> GetValueByIndex(UInt32 index, ResourceContext context);
+
+        ResourceCandidate TryGetValue(String resource);
```
jonwis, line 58: Should these go into an `MrtContract, 1.1` since it's changing the public surface area in a nonbreaking fashion?

jonwis (approved):

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/src/Microsoft.ApplicationModel.Resources.idl
huichen123: I thought it was ok to keep the contract due to not having official release. If it's not ok, I can add a new contact. What do you think?

alwu-msft, line 58: Should we instead make this `bool TryGetValue(String resource, out ResourceCandidate)` in order to fit the general pattern for TryGet* APIs (e.g. https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.trygetvalue?view=net-5.0)?

alwu-msft (changes requested):

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/src/Microsoft.ApplicationModel.Resources.idl
huichen123: According to https://osgwiki.com/wiki/WinRT_API_Design_Patterns#Try_Pattern, we may return null or false. @ptorr-msft , any opinion?

axelandrejs (approved):

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/src/Microsoft.ApplicationModel.Resources.idl
jonwis: @EHO-makai - what do _you_ think? It's probably OK to leave this at the 1.0 version for now, given our current conversations about release vehicles and versions.

EHO-makai: That's probably what I would do.  We should probably take advantage of the relaxed rules around breaking changes for 0.* release.  It would be a bit odd to arrive at Reunion 1.0 and have MrtContract 1.6.  :-)

jonwis: What @huichen123 has is the preferred form for methods that return reference types.  "bool+out" pairs are for value types.  This change avoids a behavioral-breaking change while adding the "don't throw if not found" behavior.

jonwis (approved):

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/src/Microsoft.ApplicationModel.Resources.idl
EHO-makai: The way I read the guidance, you should be using the Boolean pattern, as suggested.  But I'm not expert on this pattern so I defer to @jonwis.  (I would do the bool TryGetValue form, it seems to match most of the existing examples).

EHO-makai: Jon just approved the PR, so I guess he has spoken. :-)

alwu-msft: Cool, I learned something new today! :)

alwu-msft (approved):

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/src/Microsoft.ApplicationModel.Resources.idl
DrusTheAxe: > take advantage of the relaxed rules around breaking changes for 0.* release

Link to doc?

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/src/ResourceMap.cpp
```diff
@@ -40,15 +40,25 @@ Resources::ResourceMap ResourceMap::GetSubtree(hstring const& reference)
     return winrt::make<ResourceMap>(m_resourceManager, m_resourceManagerHandle, subtree);
 }
 
-Resources::ResourceCandidate ResourceMap::GetValueImpl(const Resources::ResourceContext* context, hstring const& resource)
+Resources::ResourceCandidate ResourceMap::GetValueImpl(const Resources::ResourceContext* context, hstring const& resource, bool notFoundOk)
 {
```
DrusTheAxe, line 44: `notFoundOK`

Odd name (negative names lead to double-negative code and harder to grok and mainain)

Suggestion: `notFoundIsOK` or `treatNotFoundAsOK`

DrusTheAxe (approved):

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/src/ResourceMap.cpp
ptorr-msft: `allowMissing` is another, or to flip the meaning, `treatMissingAsError`.

Issue #443, opened 2021-02-12
Project Reunion Installer


dkbennett:
## Summary
Standalone installer for deploying Project Reunion frameworks and main packages. Intended for use by win32 setups and custom installers to easily and consistently deploy Project Reunion dependencies.

## Rationale
This is a gap in Project Reunion deployment for win32 software. MSIX packages can include Project Reunion frameworks as dependencies for easy deployment, but win32 setups wishing to utilize Project Reunion features still need to ensure the Project Reunion packages are deployed and available. We expect multiple Project Reunion packages to be required for install with the Dynamic Dependencies feature (and likely others to follow in the future), so deploying these multiple packages is a point of friction for win32 developers wishing to use Project Reunion.

Reasons to consider this feature:
* Reduce win32 developer friction in deploying Project Reunion frameworks and main packages by having a simple .exe/.dll that can deploy the reunion packages for their software.
* Provide an example installer as a sample way to add Project Reunion packages, which developers can incorporate into their own code if they desire more complex installation.
* Input for build pipeline to produce installers for inner loop testing of private  and release builds of Project Reunion packages.

dkbennett:
This was complete a while ago :)

PR #444, opened 2021-02-12
Initial creation of the standalone Project Reunion Installer

dkbennett:
See issue #443.

This PR introduces the starting project for a standalone Project Reunion installer. It produces an executable with embedded packages and when run attempts to install and provision those packages. Provisioning is best-effort since users may not be admin and that is not an actionable error by developers and does not prevent the packages from being deployed for the current user.

It builds using the 0.1 released framework packages, which are not included in this PR, and provides an override mechanism for providing packages at build time (see project_reuion_definitions.h for details).

This is intended to be consumed by build pipelines to produce installers for just-built packages.

More changes are expected based on feedback, and tests are planned for a later PR to build and deploy test packages to verify the installer.


dev/Installer/src/ProjectReunionInstall.vcxproj
```diff
@@ -0,0 +1,157 @@
...
+      <Platform>Win32</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Debug|x64">
+      <Configuration>Debug</Configuration>
+      <Platform>x64</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Release|x64">
+      <Configuration>Release</Configuration>
+      <Platform>x64</Platform>
+    </ProjectConfiguration>
```
riverar, line 33: [External]
Missing ARM64 platform.

dev/Installer/src/ProjectReunionInstall.vcxproj
```diff
@@ -0,0 +1,157 @@
...
+    <CppWinRTOptimized>true</CppWinRTOptimized>
+    <CppWinRTRootNamespaceAutoMerge>true</CppWinRTRootNamespaceAutoMerge>
+    <CppWinRTGenerateWindowsMetadata>true</CppWinRTGenerateWindowsMetadata>
+    <MinimalCoreWin>true</MinimalCoreWin>
+    <VCProjectVersion>15.0</VCProjectVersion>
+    <ProjectGuid>{e6e59b30-9f55-4550-aa73-3b3b3dc89872}</ProjectGuid>
+    <Keyword>Win32Proj</Keyword>
+    <RootNamespace>ProjectReunionInstall</RootNamespace>
+    <WindowsTargetPlatformVersion Condition=" '$(WindowsTargetPlatformVersion)' == '' ">10.0.18362.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformMinVersion>10.0.17134.0</WindowsTargetPlatformMinVersion>
```
riverar, line 14: [External]
MinVersion is now `10.0.17763.0` per https://github.com/microsoft/ProjectReunion/discussions/413

dev/Installer/src/help.cpp
```diff
@@ -0,0 +1,18 @@
...
+    std::wcout << L"Installs the Microsoft Project Reunion framework." << std::endl;
+
+    // Get the program name for the module, since argv[0] is unreliable if launched from CreateProcess.
+    wchar_t programName[MAX_PATH];
+    if (::GetModuleFileNameW(GetModuleHandle(NULL), programName, MAX_PATH))
+    {
+        printf("%ls\n", programName);
+    }
+
+    printf("Usage:\n");
```
riverar, line 15: [External]
Probably want to excise all printf usage and ansi strings in the project, at some point.

dev/Installer/src/install.cpp
```diff
@@ -0,0 +1,172 @@
...
+    {
+        return true;
+    }
+
+    // On x64 systems, x86 architecture is also applicable
+    if (systemArch == ProcessorArchitecture::X64 && arch == ProcessorArchitecture::X86)
+    {
+        return true;
+    }
+
```
riverar, line 54: Missing ARM64. Also be careful, `GetNativeSystemInfo` will return `PROCESSOR_ARCHITECTURE_INTEL` on that hardware with native executables.

dev/Installer/src/install.cpp
```diff
@@ -0,0 +1,172 @@
...
+    // Skip non-applicable architectures.
+    if (!is_architecture_applicable(properties->architecture))
+    {
+        return S_OK;
+    }
+
+
+    wchar_t package_filename[MAX_PATH];
+    RETURN_LAST_ERROR_IF(0 == GetTempFileName(std::filesystem::temp_directory_path().c_str(), L"PRP", 0u, package_filename));
```
riverar, line 125: [External]
I would perhaps refactor at some point to utilize a uuid-based approach and one that leverages CreateFile's `FILE_FLAG_DELETE_ON_CLOSE` to make cleanup guaranteed. Would also make changes to perform the appropriate measurements to eliminate MAX_PATH.

dev/Installer/src/install.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+#include "pch.h"
+#include "packages.h"
+
+struct PackageProperties
+{
+    wil::unique_cotaskmem_string full_name;
+    wil::unique_cotaskmem_string family_name;
+    winrt::Windows::System::ProcessorArchitecture architecture = winrt::Windows::System::ProcessorArchitecture::Unknown;
+    UINT64 version = 0u;
```
riverar, line 10: [External]
Nit: Can we use `std::uint64_t` here?

dev/Installer/src/main.cpp
```diff
@@ -0,0 +1,58 @@
...
+int wmain(int argc, wchar_t *argv[])
+{
+    init_apartment();
+
+    bool quiet = false;
+
+    for (int i = 1; i < argc; ++i)
+    {
+        auto arg = std::wstring_view(argv[i]);
+        if ((arg == L"-q") || (arg == L"-quiet"))
```
riverar, line 34: [External]
tui apps generally use long form parameters prefaced with two hyphens (e.g. https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html)

dev/Installer/src/ProjectReunionInstall.vcxproj
```diff
@@ -0,0 +1,157 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project DefaultTargets="Build" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
dhoehna, line 2: This might be a problem in the future.  Is there a way to define the package location in a MACRO instead of using ../?  I know that packages can be a pain because people might store their packages in a different location.

dev/Installer/src/install.cpp
```diff
@@ -0,0 +1,172 @@
...
+#include "packages.h"
+#include "install.h"
+
+using namespace winrt;
+using namespace Windows::ApplicationModel;
+using namespace Windows::Foundation;
+using namespace Windows::Management::Deployment;
+using namespace Windows::System;
+
+winrt::hresult add_package(Uri packageUri)
```
dhoehna, line 11: NIT: THese methods use _'s to separate words.  We should come to a conclusion on what format our method names should be.

dev/Installer/src/ProjectReunionInstall.vcxproj
```diff
@@ -0,0 +1,157 @@
...
+    <ClCompile Include="help.cpp" />
+    <ClCompile Include="install.cpp" />
+    <ClCompile Include="main.cpp" />
+    <ClCompile Include="pch.cpp">
+      <PrecompiledHeader>Create</PrecompiledHeader>
+    </ClCompile>
+  </ItemGroup>
+  <ItemGroup>
+    <None Include="packages.config" />
+    <None Include="Packages\x64\Microsoft.ProjectReunion.0.1.appx" />
```
jonwis, line 130: The extension for Windows app packages is `.msix` - please fix *all* instances, in code or in solution configuration.

dev/Installer/src/help.cpp
```diff
@@ -0,0 +1,18 @@
+#include "pch.h"
+#include "help.h"
+
+void show_help()
+{
+    std::wcout << L"Installs the Microsoft Project Reunion framework." << std::endl;
```
jonwis, line 6: These should likely be localized.

dev/Installer/src/help.cpp
jonwis: Move everything to `std::wcout`

dev/Installer/src/install.cpp
jonwis: http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#nl8-use-a-consistent-naming-style

@DefaultRyan or @kennykerr probably have casing opinions.  My guess is they'd want `snake_case` but I'm an old fogey who's partial to `PascalCasing`.

dev/Installer/src/install.cpp
```diff
@@ -0,0 +1,172 @@
...
+{
+    PackageManager packageManager;
+    auto deploymentOperation{ packageManager.AddPackageAsync(packageUri, nullptr, DeploymentOptions::None) };
+    deploymentOperation.get();
+    if (deploymentOperation.Status() != AsyncStatus::Completed)
+    {
+        return deploymentOperation.ErrorCode();
+    }
```
jonwis, line 19: Hey @DrusTheAxe, won't this `.get()` still will throw because a failed deployment operation emits a failed `HRESULT` from the `IAsyncOperationWithProgress<T,P>::GetResult`? So this unfortunately needs a try/catch, if that's the case.

dev/Installer/src/install.cpp
```diff
@@ -0,0 +1,172 @@
...
+{
+    HMODULE const hmodule = GetModuleHandle(NULL);
+    HRSRC hrsrc = ::FindResource(hmodule, resourceName.c_str(), resourceType.c_str());
+    THROW_LAST_ERROR_IF_NULL(hrsrc);
+    HGLOBAL hresource = LoadResource(hmodule, hrsrc);
+    THROW_LAST_ERROR_IF_NULL(hresource);
+    const BYTE* data = reinterpret_cast<BYTE*>(::LockResource(hresource));
+    THROW_LAST_ERROR_IF_NULL(data);
+    DWORD size = ::SizeofResource(hmodule, hrsrc);
+    wil::com_ptr<IStream> stream = ::SHCreateMemStream(data, size);
```
jonwis, line 68: Sadly, this causes a leak - `SHCreateMemStream` returns a stream with a +1 refcount, and the ctor here bumps it with another ref.  Maybe write a function like this:

```c++
wil::com_ptr<IStream> create_mem_stream(const void* data, size_t size)
{
    wil::com_ptr<IStream> retval;
    retval.attach(::SHCreateMemStream(data, size));
    return retval;
}
```

dev/Installer/src/install.cpp
```diff
@@ -0,0 +1,172 @@
...
+    wil::com_ptr<IStream> outstream;
+    try
+    {
+        outstream = open_temp_file(package_filename);
+    }
+    catch (wil::ResultException ex)
+    {
+        return ex.GetErrorCode();
+    }
+
```
jonwis, line 143: I keep seeing these try/catch blocks. Shouldn't all this be exception-based? Just let things throw and catch them at the higher layer. Only add try/catch if you think you can recover from the problems here and change course, or retry, or whatever.

dev/Installer/src/install.cpp
```diff
@@ -0,0 +1,172 @@
...
+    // Provisioning is expected to fail if the program is not run elevated or the user is not admin.
+    hresult provision_result = provision_package(properties->family_name.get());
+    if (!quiet)
+    {
+        printf("Provisioning result: 0x%08x\n", provision_result.value);
+    }
+    LOG_IF_FAILED((HRESULT)provision_result);
+
+    // Cleanup the temp file
+    LOG_IF_WIN32_BOOL_FALSE(::DeleteFile(package_filename));
```
jonwis, line 169: Add a `wil::scope_exit` to do this, or use @riverar 's temp-file-delete-on-close model.

dev/Installer/src/install.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+#include "pch.h"
+#include "packages.h"
+
+struct PackageProperties
```
jonwis, line 5: This type is mixing casing models; one thing we have in the OS repo is "types are PascalCased, members are camelCased."

dev/Installer/src/packages.config
```diff
@@ -0,0 +1,5 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="Microsoft.Windows.CppWinRT" version="2.0.201217.4" targetFramework="native" />
```
jonwis, line 3: Hey @EHO-makai  or @BenJKuhn - is there a way to have this solution share wil/cppwinrt packages with the other Project Reunion solutions?

dev/Installer/src/main.cpp
```diff
@@ -0,0 +1,58 @@
...
+            return 0;
+        }
+        else if ((arg == L"-a") || (arg == L"-add"))
+        {
+            show_help();
+            return 0;
+        }
+        else
+        {
+            printf("Unknown argument: %ws\n", arg.data());
```
jonwis, line 50: Note for the future - `std::wstring_view` is not guaranteed to be null terminated. But when you move to `std::wcout` (or here, `std::wcerr`) the `operator<<` takes care of that for you.

jonwis (changes requested):
Overall a good start. Are there tests or simulations to verify behavior in the CI/CD pipeline?

dev/Installer/src/help.cpp
kennykerr: https://kennykerr.ca/2014/12/08/in-defense-of-printf/ 😉

dev/Installer/src/install.cpp
riverar: @jonwis I'm in your camp as well, to align with .NET. Are there coding guidelines coming soon?

kennykerr: I don't have an opinion. For reference, The C++/WinRT library distinguishes between projected types and library types by using snake_case for the latter, but other projects/libraries are free to do otherwise.

dev/Installer/src/ProjectReunionInstall.vcxproj.filters
```diff
@@ -0,0 +1,62 @@
...
+      <Filter>Source Files</Filter>
+    </ClCompile>
+    <ClCompile Include="install.cpp">
+      <Filter>Source Files</Filter>
+    </ClCompile>
+  </ItemGroup>
+  <ItemGroup>
+    <None Include="PropertySheet.props" />
+    <None Include="packages.config" />
+    <None Include="Packages\x86\Microsoft.ProjectReunion.0.1.appx" />
```
EHO-makai, line 54: adjusting this is going to be a per-release hand-edit?

dev/Installer/src/ProjectReunionInstall.vcxproj
EHO-makai: The place for that change is probably not here.  I queried a bunch of people over our usage of .appx and got a <shrug> but I apparently didn't poll you at the time. :-)  Can you file an issue on this?  It has engineering system costs so it isn't a trivial change and shouldn't be dirven by David's PR (IMO).

dev/Installer/src/install.cpp
dhoehna: m_ PascalCase for member variables
CONSTANTS_THAT_HAVE_SPACES
PascalCase for methods and classes
camelCase for local variables

dev/Installer/src/packages.config
EHO-makai: I think this involves having a nuget.config at the top level of the repo that solutions share.  I haven't actually done this before so someone will have to prototype the change.  As always, we should log an issue and triage.  I think it is beneficial to make the version of cppwinrt common, at the risk of causing a compat problem between solutions in the project (one solution prefers a different cppwinrt version due to a subtle difference).  But I believe a solution with such an issue can override this with its own nuget.config at its own solution "level".  @BenJKuhn should probably own this issue / scheduling the work for this change.

dev/Installer/src/install.cpp
EHO-makai: I note that this "trap for the unwary" is not noted on the documentation for SHCreateMemStream.

dev/Installer/src/install.cpp
jonwis: @dhoehna yes, except for the first one - `m_camelCase` is a scoped variable, so it gets `m_` and `camelCase`.

dev/Installer/src/ProjectReunionInstall.vcxproj
EHO-makai: > You can fix this after we finish fixing this issue:  https://github.com/microsoft/ProjectReunion/issues/447

DrusTheAxe (changes requested):

dev/Installer/src/install.cpp
riverar: @DrusTheAxe arm64 images return PROCESSOR_ARCHITECTURE_INTEL. (Yeah, I know. That platform is slightly cursed.)

dev/Installer/src/help.cpp
DrusTheAxe:
```
int x = 123456;
std::cout << std::hex << x;
```
vs
```
int x = 123456;
printf("%x", x);
```
Printf easily wins on clarity and maintainability even in simple cases let alone real world uses. The perf benefits are just the cherry on top :-)

dev/Installer/src/help.cpp
```diff
@@ -0,0 +1,18 @@
...
+
+    // Get the program name for the module, since argv[0] is unreliable if launched from CreateProcess.
+    wchar_t programName[MAX_PATH];
```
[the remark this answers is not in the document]
riverar: No version of Windows ships with longPathAware enabled at this time. So including this fusion manifest flag just means you now have to support users that manually flip this switch. Noble but you have to decide if it's something you _really want to support_ at this stage.

dev/Installer/src/packages.config
jonwis: Filed #448

dev/Installer/src/help.cpp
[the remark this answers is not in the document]
DrusTheAxe: There's the world that is, and the world we desire it to be. We must live in the one as we strive to build the other.

I see no downside to support longPathAware here for the installer.

Windows doesn't enable longPathAware by default because it's not longPathAware-clean yet. It would be unwise to enable it by default until that's addressed. But Project Reunion's installer is a smaller problem space than Windows ;-) so much more tractable to enable.

dev/Installer/src/ProjectReunionInstall.vcxproj
dkbennett: > The extension for Windows app packages is `.msix` - please fix _all_ instances, in code or in solution configuration.

That's the exact filename in the nuget distribution for released version 0.1, so that's what I have as the default file extension for the frameworks.  Though that said, it doesn't matter what the extension is for deployment to consume it (and an extension isn't specified when copying from memory to temp file), so there isn't need to have any specific extension.

However, the entries here in the vcxproj are actually an artifact from an earlier version of the installer and can be removed enitrely, so I'll do that. The ".appx" specification in the project_reunion_definitions.h file I will leave alone for now until the frameworks in the nuget are changed. If we have these frameworks on a release share in .msix form I can update the paths to point there by default.

dev/Installer/src/ProjectReunionInstall.vcxproj
dkbennett: I'll update to add definitions for the ARM packages and inclusion.

dev/Installer/src/ProjectReunionInstall.vcxproj
dkbennett: I don't know which one I should use here. Based on @DrusTheAxe 's comment, should model after the Framework.Math example?

dev/Installer/src/install.cpp
dkbennett: > Missing ARM64. Also be careful, `GetNativeSystemInfo` will return `PROCESSOR_ARCHITECTURE_INTEL` on that hardware with native executables.

What is the correct way to test that the native system is in fact arm64 and not x86?

dev/Installer/src/install.cpp
dkbennett: I tried using FILE_FLAG_DELETE_ON_CLOSE initially and ran into problems with the deployment API not using it. I suspect the issue is that for FILE_FLAG_DELETE_ON_CLOSE to work properly, all openers must open with FILE_FLAG_SHARE_DELETE, and I don't think the deployment pipeline is doing that.

I order to keep the installer simple and reliable, I'm going to avoid using this flag and simply delete the file on scope exit. It's still being created in a temporary directory so if for some reason scope exit doesn't properly delete the file it can still be reclaimed when cleaning up temporary files.

dev/Installer/src/install.cpp
dkbennett: My intent here was to try to comply with the https://github.com/isocpp/CppCoreGuidelines by using PascalCase for types and preferring snake_case everywhere else (per [NL.10: Prefer underscore_style names](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#nl10-prefer-underscore_style-names)

However, NL 10 also states that it is only if I 'have a choice' - since this was a new project, I assumed there was a choice here and preferred underscore except for Type names, but from feedback it appears what is desired is Windows-style guidelines which @dhoehna describes above.

I'm perfectly OK with this. I personally find snake_case difficult to read and awkward, and I prefer the Windows-style. Since MRT Core has uses Windows style and has gone through review I will assume this is an acceptable naming convention and convert all naming in this project over to Windows-style and ignore NL 10 in the CppCoreGuidelines.

I've created issue #452 to get some clarity on what is preferred for this project with regards to NL 10.

dev/Installer/src/help.cpp
dkbennett: I've created #456 to address this along with the printf vs wcout which seems related there. I believe this should be addressed in a subsequent update to the project.

dev/Installer/src/install.cpp
riverar: I'd replace `GetNativesystemInfo` usage here with [`IsWow64Process2`](https://docs.microsoft.com/en-us/windows/win32/api/wow64apiset/nf-wow64apiset-iswow64process2). Then you can inspect `pNativeMachine`, which returns the correct `IMAGE_FILE_MACHINE_ARM64`.

dev/Installer/src/install.cpp
dkbennett: I'm trying to have the coordinating install function able to report errors where applicable, but I will rework this to bubble up most of these except where particularly relevant (such as catching provisioning failures, which are non-fatal and expected).

dev/Installer/src/install.cpp
dkbennett: > Overall a good start. Are there tests or simulations to verify behavior in the CI/CD pipeline?

Tests are coming in an update a little later this week. Testing has been tricky because the installer is intended to be built with real packages and we don't want to be messing with the configuration of reunion on the dev machine when doing inner loop testing of the installer. So to work around that and the plan is to build 3 test packages (one for each framework architecture) and the default project will target including those test packages when no override header is present. The tests for the will the run the installer and verify the test packages are installed and then remove the test packages. So if you download the project and build the installer, you'll get an installer that installs just-built test packages that don't do anything to the system. That will allow inner loop testing of the installer and for tests to validate that the installer functions as expected. 

Then to produce real packages or packages for a targeted build, the override header will be included to target different packages. 

So pipeline is expected to be:
1) build the installer (using test packages)
2) test the installer
3) generate override header
4) generate the build's real reunion packages
5) build the installer using override header to target the real packages
6) sign the installer

jonwis: Seems like this tool could be used to bundle any set of MSIXs in any order though, right?  That is, the installer could find a little file within itself with a list of package descriptions (not baked into code), run the mini-applicability on the data in that file, and call the package manager APIs as necessary? Then you should be able to test that in isolation with completely bogus test packages now, no additional signing required.

@EHO-makai or @BenJKuhn, do we have a git large file storage location available to us somewhere? Committing large binaries to the repo is definitely not good, but somewhere we need packages available so everybody who's running the test suite locally can get them.

dev/Installer/src/ProjectReunionInstall.vcxproj
dkbennett: Fixed in latest push to match Framework.Math example.

dev/Installer/src/help.cpp
dkbennett: I moved it all over to std::wcout and std::wcerr, though I am sympathetic to the plight of printf and its simplicity and readability.

dev/Installer/src/install.cpp
dkbennett: I added Arm64 and what should be Arm64 detection, but I noticed that visual studio's drop-down menu architecture configuration did not detect _M_ARM64 and instead appears to be _M_IX86. This is a more complicated issue to resolve since it has to do with both packaging the correct frameworks in the Arm64 version as well as runtime detection to solve it, so I created a new issue to track it - #479

I suspect resolving that is going to be a non-trivial amount of work and should not block this initial installer deployment.

dev/Installer/src/install.h
dkbennett: Fixed

dev/Installer/src/install.cpp
dkbennett: I aligned all around the standard Windows style in the latest push.

dev/Installer/src/ProjectReunionInstall.vcxproj
dkbennett: .msix is used everywhere in this project now.

dev/Installer/src/install.cpp
dkbennett: Wrapped both deployment operations in try-catch.

dev/Installer/src/install.h
dkbennett: Fixed to match Windows style guidelines.

dev/Installer/src/main.cpp
dkbennett: Fixed by moving to wcout.

dev/Installer/src/ProjectReunionInstall.vcxproj.filters
dkbennett: Test packages are built as part of the default project now if no override is specified, so no packages are actually checked in for this project to successfully build, deploy, and be tested. 

We may need to add additional support as more packages come along but its setup to be able to handle that scaling.

dev/Installer/src/ProjectReunionInstall.vcxproj.filters
```diff
@@ -0,0 +1,62 @@
...
+    <None Include="Packages\x64\Microsoft.ProjectReunion.0.1.appx" />
```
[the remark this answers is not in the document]
dkbennett: No more included packages, and arm64 has been added.

dev/Installer/src/install.cpp
```diff
@@ -0,0 +1,172 @@
...
+    try
+    {
+        outstream = open_temp_file(package_filename);
+    }
+    catch (wil::ResultException ex)
+    {
+        return ex.GetErrorCode();
+    }
+ 
+    ULARGE_INTEGER streamSize = { 0 };
```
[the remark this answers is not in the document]
dkbennett: Fixed the nit.  The GetTempFileName() call creates a unique file because that is specified in the arguments and then returns the name of that unique file. Unless that method is lying and not actually creating a unique filename I think this API should be sufficient for these purposes.

dev/Installer/src/main.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include "packages.h"
+
+using namespace winrt;
+using namespace Windows::ApplicationModel;
+using namespace Windows::Foundation;
+using namespace Windows::System;
+
+int deploy_packages(bool quiet)
+{
+    for (const auto& package : c_packages)
```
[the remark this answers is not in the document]
dkbennett: Moved it all to namespace and made explicit reference.

dev/Installer/src/install.cpp
```diff
@@ -0,0 +1,172 @@
...
+}
+
+wil::com_ptr<IStream> open_temp_file(_In_ PCWSTR path)
+{
+    wil::com_ptr<IStream> outstream;
+    THROW_IF_FAILED(SHCreateStreamOnFileEx(path, STGM_WRITE | STGM_READ | STGM_SHARE_DENY_WRITE | STGM_CREATE, FILE_ATTRIBUTE_NORMAL, TRUE, nullptr, wil::out_param(outstream)));
+    return outstream;
+}
+
+winrt::hresult deploy_package_from_resource(ResourcePackageInfo resource, bool quiet) noexcept
```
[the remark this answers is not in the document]
dkbennett: I was aiming to be consistent with using the same type for hresult throughout.

dev/Installer/src/install.cpp
```diff
@@ -0,0 +1,172 @@
...
+    THROW_IF_FAILED(id->GetPackageFamilyName(&properties->family_name));
+    APPX_PACKAGE_ARCHITECTURE arch;
+    THROW_IF_FAILED(id->GetArchitecture(&arch));
+    properties->architecture = static_cast<ProcessorArchitecture>(arch);
+    THROW_IF_FAILED(id->GetVersion(&properties->version));
+
+    return properties;
+}
+
+wil::com_ptr<IStream> open_temp_file(_In_ PCWSTR path)
```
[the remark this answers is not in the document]
dkbennett: Fixed

dev/Installer/src/install.cpp
```diff
@@ -0,0 +1,172 @@
...
+    SYSTEM_INFO systemInfo;
+    GetNativeSystemInfo(&systemInfo);
+    ProcessorArchitecture systemArch = static_cast<ProcessorArchitecture>(systemInfo.wProcessorArchitecture);
+
+    if (arch == systemArch)
+    {
+        return true;
+    }
+
+    // On x64 systems, x86 architecture is also applicable
```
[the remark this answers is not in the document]
dkbennett: Yes, added arm64 detection where both are applicable.

dev/Installer/src/install.cpp
```diff
@@ -0,0 +1,172 @@
...
+using namespace Windows::System;
+
+winrt::hresult add_package(Uri packageUri)
+{
+    PackageManager packageManager;
+    auto deploymentOperation{ packageManager.AddPackageAsync(packageUri, nullptr, DeploymentOptions::None) };
+    deploymentOperation.get();
+    if (deploymentOperation.Status() != AsyncStatus::Completed)
+    {
+        return deploymentOperation.ErrorCode();
```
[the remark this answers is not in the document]
dkbennett: Is this a broader project rhetorical question or would you like me to change to to ExtendedErrorCode?

dev/Installer/src/help.h
```diff
@@ -0,0 +1,4 @@
+#pragma once
+#include "pch.h"
+
+void show_help();
```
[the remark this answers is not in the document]
dkbennett: fixed

dev/Installer/src/help.cpp
```diff
@@ -0,0 +1,18 @@
...
+    if (::GetModuleFileNameW(GetModuleHandle(NULL), programName, MAX_PATH))
+    {
+        printf("%ls\n", programName);
+    }
+
+    printf("Usage:\n");
+    printf("  [-quiet] : Suppress all console output.\n");
```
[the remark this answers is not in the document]
dkbennett: Fixed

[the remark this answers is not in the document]
dkbennett: Fixed

dev/Installer/src/help.cpp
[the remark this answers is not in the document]
dkbennett: Fixed to use the wil helper.

dev/Installer/src/help.cpp
```diff
@@ -0,0 +1,18 @@
+#include "pch.h"
+#include "help.h"
+
+void show_help()
+{
+    std::wcout << L"Installs the Microsoft Project Reunion framework." << std::endl;
+
+    // Get the program name for the module, since argv[0] is unreliable if launched from CreateProcess.
+    wchar_t programName[MAX_PATH];
+    if (::GetModuleFileNameW(GetModuleHandle(NULL), programName, MAX_PATH))
```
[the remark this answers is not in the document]
dkbennett: Fixed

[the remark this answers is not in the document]
dkbennett: Done

dev/Installer/src/ProjectReunionInstall.vcxproj
```diff
@@ -0,0 +1,157 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project DefaultTargets="Build" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="..\packages\Microsoft.Windows.CppWinRT.2.0.201217.4\build\native\Microsoft.Windows.CppWinRT.props" Condition="Exists('..\packages\Microsoft.Windows.CppWinRT.2.0.201217.4\build\native\Microsoft.Windows.CppWinRT.props')" />
+  <PropertyGroup Label="Globals">
+    <CppWinRTOptimized>true</CppWinRTOptimized>
+    <CppWinRTRootNamespaceAutoMerge>true</CppWinRTRootNamespaceAutoMerge>
+    <CppWinRTGenerateWindowsMetadata>true</CppWinRTGenerateWindowsMetadata>
+    <MinimalCoreWin>true</MinimalCoreWin>
+    <VCProjectVersion>15.0</VCProjectVersion>
```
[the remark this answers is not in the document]
dkbennett: fixed

dev/Installer/src/install.cpp
dkbennett: Fixed to use scope_exit. Currently the installer project looks for a specific header with different package paths for identifying which packages to include, otherwise it includes some test-signed dummy packages it builds. The installer tests assume the dummy packages and verify their installation and do cleanup of those packages.

It can be used to bundle any set of MSIXes in any order, though there is no dependency mappings since this is intended to explicitly install frameworks in addition to main packages. For dependencies the order would potentially matter, but shouldn't be an issue if frameworks are registered first. If we want to move to only installing the main package and then adding other dependency packages it would change the install logic a bit. Dependency support could be added, though I don't think it is required at this time. I could add a script step that reads a data file and creates the header. That's essentially what the build pipeline needs to do - whether it's creating a data file or a header it still has to indicate the location of the packages to install. 

I'd want to carefully think about how to structure that process if it is to do some code generation during build, but I think what it has now is sufficient and I can update it as we discover pipeline issues or needs.

It doesn't need any packages in order to test the installer functionality. It will create dummy packages to verify deployment and the installer works as expected.

dev/Installer/src/install.cpp
dkbennett: Updated to have everything throw except the highest layer and only meaningful function, "DeployPackages" which deploys all of the packages. Any failures in there are exceptional, save the provisioning which is best-effort.

dev/Installer/src/install.cpp
```diff
@@ -0,0 +1,194 @@
...
+using namespace Windows::System;
+
+namespace ProjectReunionInstaller {
+
+    winrt::hresult AddPackage(const Uri& packageUri)
+    {
+        PackageManager packageManager;
+        const auto deploymentOperation{ packageManager.AddPackageAsync(packageUri, nullptr, DeploymentOptions::None) };
+        deploymentOperation.get();
+        if (deploymentOperation.Status() != AsyncStatus::Completed)
```
jonwis, line 18: I _think_ this `if` block is no longer necessary - the deployment operation if it fails will cause a throw during the 'get' so this `Status` should always be `Completed` at this point.  Maybe ask Carl?

(Applies to all similar blocks.)

dev/Installer/src/install.cpp
```diff
@@ -206,19 +173,22 @@ namespace ProjectReunionInstaller {
         }
         LOG_IF_FAILED(static_cast<HRESULT>(hrProvisionResult));
 
-        return S_OK;
+        return;
```
DrusTheAxe, line 176: NIT: Remove.

It's the last line in a void function so return; is redundant.

dev/Installer/src/install.h
```diff
@@ -9,8 +9,8 @@ namespace ProjectReunionInstaller {
         wil::unique_cotaskmem_string fullName;
         wil::unique_cotaskmem_string familyName;
         winrt::Windows::System::ProcessorArchitecture architecture = winrt::Windows::System::ProcessorArchitecture::Unknown;
-        std::uint64_t version = 0u;
+        std::uint64_t version{ 0u };
```
DrusTheAxe, line 12: NIT: `std::uint64_t version{};`

No need to explicitly define the value as {} gets you the default for the type, which is 0 for integers.

dev/Installer/src/install.h
```diff
@@ -9,8 +9,8 @@ namespace ProjectReunionInstaller {
         wil::unique_cotaskmem_string fullName;
         wil::unique_cotaskmem_string familyName;
         winrt::Windows::System::ProcessorArchitecture architecture = winrt::Windows::System::ProcessorArchitecture::Unknown;
```
DrusTheAxe, line 11: NIT: `...architecture{ winrt::Windows::System::ProcessorArchitecture::Unknown };`

dev/Installer/src/install.cpp
dkbennett: It is necessary because the operation doesn't throw when it fails, it returns a failure status. This is observed in the Provisioning operation, which regularly fails with access denied whenever called by a non-admin, nor not supported when called on a framework. These failures don't result in an exception, just a failed operation.

installer/dev/install.cpp
```diff
@@ -0,0 +1,194 @@
...
+    {
+        wil::com_ptr<IStream> outstream;
+        THROW_IF_FAILED(SHCreateStreamOnFileEx(path, STGM_WRITE | STGM_READ | STGM_SHARE_DENY_WRITE | STGM_CREATE, FILE_ATTRIBUTE_NORMAL, TRUE, nullptr, wil::out_param(outstream)));
+        return outstream;
+    }
+
+    void DeployPackageFromResource(const ProjectReunionInstaller::ResourcePackageInfo& resource, const bool quiet)
+    {
+        // Get package properties by loading the resource as a stream and reading the manifest.
+        wil::com_ptr<IStream> packageStream;
```
jonwis, line 124: Nit/bikeshed - you can do single-line init here for both:

```c++
auto packageStream = GetResourceStream(...);
auto packageProperties = GetPackagePropertiesFromStream(packageStream);
```

installer/dev/install.cpp
```diff
@@ -0,0 +1,194 @@
...
+        {
+            std::wcout << "Package Full Name: " << packageProperties->fullName.get() << std::endl;
+            std::wcout << "Temp package path: " << packageFilename << std::endl;
+        }
+
+        // Write the package to a temp file. The PackageManager APIs require a Uri.
+        wil::com_ptr<IStream> outStream{ OpenFileStream(packageFilename) };
+        ULARGE_INTEGER streamSize{};
+        THROW_IF_FAILED(::IStream_Size(packageStream.get(), &streamSize));
+        THROW_IF_FAILED(packageStream.get()->CopyTo(outStream.get(), streamSize, nullptr, nullptr));
```
jonwis, line 155: Nit: `.get()` not required; `wil::com_ptr::operator->` does the right thing.

installer/dev/install.cpp
```diff
@@ -0,0 +1,194 @@
...
+        if (!quiet)
+        {
+            std::wcout << "Provisioning result : 0x" << std::hex << hrProvisionResult.value << std::endl;
+        }
+        LOG_IF_FAILED(static_cast<HRESULT>(hrProvisionResult));
+
+        return;
+    }
+
+    HRESULT DeployPackages(const bool quiet) noexcept
```
jonwis, line 179: Nit: Use a function-block try:

```c++
HRESULT DeployPackages(bool quiet) noexcept try
{
    for (const auto& package : c_packages)
    {
        DeployPackageFromResource(package, quiet);
    }
}
CATCH_RETURN();
```

installer/dev/project_reunion_definitions.h
```diff
@@ -0,0 +1,64 @@
...
+#define PR_FRAMEWORK_X64_RCTYPE PACKAGE
+#define PR_FRAMEWORK_X64_PATH "..\\test\\testpackages\\framework_x64.msix"
+
+// arm64 Framework
+#define PR_FRAMEWORK_ARM64_ID L"PR_FWPACKAGE_ARM64"
+#define PR_FRAMEWORK_ARM64_RCID PR_FWPACKAGE_ARM64
+#define PR_FRAMEWORK_ARM64_TYPE L"PACKAGE"
+#define PR_FRAMEWORK_ARM64_RCTYPE PACKAGE
+#define PR_FRAMEWORK_ARM64_PATH "..\\test\\testpackages\\framework_arm64.msix"
+
```
jonwis, line 41: Consider removing the "ARM" targets from the project & solution since we're not supporting them, eventually.

installer/dev/project_reunion_definitions.h
```diff
@@ -0,0 +1,64 @@
...
+// Each package has a package definition listed below.
+// Names for macros are intentionally kept short due to resource compiler length limitations.
+
+// x86 Framework
+#define PR_FRAMEWORK_X86_ID L"PR_FWPACKAGE_X86"
+#define PR_FRAMEWORK_X86_RCID PR_FWPACKAGE_X86
+#define PR_FRAMEWORK_X86_TYPE L"PACKAGE"
+#define PR_FRAMEWORK_X86_RCTYPE PACKAGE
+#define PR_FRAMEWORK_X86_PATH "..\\test\\testpackages\\framework_x86.msix"
```
jonwis, line 26: Feels like this data should be some kind of an input parameter.  This mechanism - packing up MSIXes into an SFX that can install them - isn't specific to Project Reunion.

In some future update, read these from a .json file so it's not hard-coded.

installer/test/CreateInstallerTestPackages/CreateInstallerTestPackages.vcxproj
```diff
@@ -0,0 +1,165 @@
...
+    <NMakeOutput>$(OutputPath)Microsoft.ProjectReunion.Test.InstallerPackages</NMakeOutput>
+    <NMakeCleanCommandLine>nmake /f makefile.mak ProjectDir=$(ProjectDir) SolutionDir=$(SolutionDir) OutDir=$(OutDir) TargetName=$(TargetName) clean</NMakeCleanCommandLine>
+    <NMakeReBuildCommandLine>nmake /f makefile.mak ProjectDir=$(ProjectDir) SolutionDir=$(SolutionDir) OutDir=$(OutDir) TargetName=$(TargetName)  rebuild</NMakeReBuildCommandLine>
+    <NMakePreprocessorDefinitions>WIN32;NDEBUG;$(NMakePreprocessorDefinitions)</NMakePreprocessorDefinitions>
+    <OutDir>$(OutDir)</OutDir>
+    <IntDir>$(IntDir)</IntDir>
+    <NMakeIncludeSearchPath>$(SolutionDir)tools</NMakeIncludeSearchPath>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
+    <NMakeBuildCommandLine>nmake /C /f makefile.mak ProjectDir=$(ProjectDir) SolutionDir=$(SolutionDir) OutDir=$(OutDir) TargetName=$(TargetName) </NMakeBuildCommandLine>
```
jonwis, line 139: nmake? msbuild is likely able to do anything you needed to do with nmake, but the syntax is clumsier. Revisit this at some point - we should reduce the number of unique build mechanisms being used (cc @EHO-makai )

jonwis:
Can we avoid committing in the .msix files?  Git doesn't like binaries.

jonwis (approved):
Nothing major to change, thanks for all the tests. If you can, please _avoid_ checking in the `.msix` test files. Maybe add a simple WAP packaging project to generate a dummy/empty MSIX during the build and use that?

installer/dev/project_reunion_definitions.h
DrusTheAxe: These are the defaults, for inner-loop testing. There's an override mechanism to use alternate data - as the pipeline does/will to produce the actual signed/release/shipping exe, and others can use to rebuild with their own packages (for those so motivated)

DrusTheAxe (approved):
LGTM

installer/dev/project_reunion_definitions.h
dkbennett: ARM64 also? I thought it was just ARM that was not supported.

installer/dev/project_reunion_definitions.h
dkbennett: Agree, I want to move this over to a build script that generates the .rc and code from a template to include packages defined in a .json file so it can more easily include any package sets, but that ended up being a much larger work item than I had time allotted.  Created #520 to track.

installer/test/CreateInstallerTestPackages/CreateInstallerTestPackages.vcxproj
dkbennett: > Can we avoid committing in the .msix files? Git doesn't like binaries.

The broader issue is now referenced in #521, but here's how I'm addressing it here:

I originally was using the MSTest cert that was checked into /build to build these on-the-fly (which was expired and was using Howard's updated copy locally, but as that is not checked in and may not, F5 for the installer would be broken if I didnt include the actually valid signed packages. I wanted to have a reliable build of the installer with tests that anyone could include and simply F5, including those without access to the MSTest cert to verify local builds of the installer. These packages also don't change unless adding more tests and have no executable code, so including them does speed up build times, so I changed it to be a separate solution to build the test packages that can then be added as test collateral when needed.

I could update this to get the cert from Azure every time and build the packages with every installer build, but that would add significantly to build time and break F5 by requiring the developer to run a pipeline. And it coudl also cause the solution to not work for folks without access to the MSTest cert. The current solution should work for everyone to build and validate the installer with a simple F5 and then have it include their own packages with an override header without need of pipeline access or execution. I've added #521 discuss this problem. When there's a good solution here I'll be happy to make the changes to not include the .msix, but as it currently stands this appears to be the best way to have an accessible and efficient inner loop for the installer.

installer/dev/install.cpp
dkbennett: Fixed this and other similar nits.

PR #454, opened 2021-02-18
Startup activations, AppLifecycle updates, and activations.

aeloros:
Startup activations are now supported with this change.  All activations now are passed to the app using a Uri scheme that encodes the argument data as key/value pairs.  A new internal to reunion way of activating the app through registering for a special protocol was added.  These two things should facilitate adding new event arg types in an easier manner.

Updated the API surface to match the latest changes from @andreww-msft's spec.

Fixed a couple bugs in existing tests, and added a primitive startup activation test.


dev/ProjectReunion_DLL/COMExports.cpp
```diff
@@ -0,0 +1,105 @@
...
+extern "C" HRESULT __stdcall DllRegisterServer() noexcept try
+{
+    RegisterCoClass(__uuidof(EncodedLaunchExecuteCommandFactory));
+    return S_OK;
+}
+CATCH_RETURN()
+
+extern "C" HRESULT __stdcall DllUnregisterServer() noexcept try
+{
+    std::wstring scheme = L"ms-launch";
```
aeloros, line 45: update to use const

dev/ProjectReunion_DLL/COMExports.cpp
```diff
@@ -0,0 +1,105 @@
...
+        }
+    } while (actual_size + 1 > path_size);
+
+    path.resize(actual_size);
+    return path;
+}
+
+std::wstring GetCoClassPath(const GUID& iid)
+{
+    std::wstring path{ LR"(SOFTWARE\Classes\CLSID\{????????-????-????-????-????????????})" };
```
aeloros, line 80: Update to use const

dev/ProjectReunion_DLL/COMExports.cpp
```diff
@@ -0,0 +1,105 @@
...
+    ::StringFromGUID2(iid, path.data() + 23, 39);
+    return path;
+}
+
+void RegisterCoClass(const GUID& iid)
+{
+    std::wstring key_path{ GetCoClassPath(iid) };
+    key_path += LR"(\InProcServer32)";
```
aeloros, line 88: update to use const

test/TestApps/AppLifecycleTestApp/main.cpp
```diff
@@ -65,20 +65,37 @@ int main()
...
+
+                // Signal event that file was registered.
+                SignalPhase(c_testStartupPhaseEventName);
+                succeeded = true;
+            }
+            else if (argument.compare(L"UnregisterStartup") == 0)
+            {
+                ActivationRegistrationManager::UnregisterForStartupActivation(L"this_is_a_test");
+
+                // Signal event that file was unregistered.
```
DrusTheAxe, line 95: >// Signal event that file was unregistered. [](start = 16, length = 43)

Is this comment significant? Seems redundant with next line

DrusTheAxe:
> static const int c_phaseTimeout = (30 * 1000); // 30 seconds

constexpr ?

---
Refers to: test/inc/TestDef.h:5 in 70878ee. [](commit_id = 70878eeb30a7b81e625e3a8551a399a77e3e820e, deletion_comment = False)

DrusTheAxe:
> static const std::wstring c_testProtocolScheme = L"reuniontestprotocol";

__declspec(selectany) or constexpr be better, e.g. avoid linker errors if 2 files #include this one

---
Refers to: test/inc/TestDef.h:8 in 70878ee. [](commit_id = 70878eeb30a7b81e625e3a8551a399a77e3e820e, deletion_comment = False)

DrusTheAxe:
> static const std::wstring c_testProtocolScheme = L"reuniontestprotocol";

This fires constructors are module load time. Does std::wstring_view suffer the same problem? If not that would be better; if it does then PCWSTR would be better

---
Refers to: test/inc/TestDef.h:8 in 70878ee. [](commit_id = 70878eeb30a7b81e625e3a8551a399a77e3e820e, deletion_comment = False)

dev/AppLifecycle/ActivationRegistrationManager.h
```diff
@@ -6,18 +6,37 @@
 
 namespace winrt::Microsoft::ProjectReunion::implementation
 {
+    // Registration constant values.
+    static const std::wstring c_argumentPrefix = L"----";
```
DrusTheAxe, line 10: >std::wstring [](start = 17, length = 12)

See other comment re std::wstring for constants vs (maybe) std::wstring_view or PCWSTR

DrusTheAxe, line 10: >c_argumentPrefix = L"----"; [](start = 30, length = 27)

...c_argumentPrefix{ L"-----" };

ES.23: Prefer the {}-initializer syntax
[https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#es23-prefer-the--initializer-syntax](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#es23-prefer-the--initializer-syntax)

Same comment applies throughout the (not just this file) i.e. never use = for initialization, always use {}

dev/AppLifecycle/ActivationRegistrationManager.cpp
```diff
@@ -8,47 +8,89 @@
 #include "ProtocolActivatedEventArgs.h"
 #include "FileActivatedEventArgs.h"
 #include "Association.h"
+#include "ExtensionContract.h"
+#include "EncodedLaunchExecuteCommand.h"
+
+extern "C" HRESULT __stdcall DllRegisterServer() noexcept;
```
DrusTheAxe, line 14: >DllRegisterServer [](start = 29, length = 17)

Why?

Is this pre-DynamicDependencies workaround?

dev/AppLifecycle/ActivationRegistrationManager.cpp
```diff
@@ -8,47 +8,89 @@
...
 
 namespace winrt::Microsoft::ProjectReunion::implementation
 {
+    std::wstring GenerateEncodedLaunchUri(std::wstring const& appUserModelId, std::wstring const& contractId)
```
DrusTheAxe, line 18: >GenerateEncodedLaunchUri [](start = 17, length = 24)

This assumes valid inputs e.g. can't be "It's not a valid AUMID & other <badness> that's not (properly) eñcod=d!" but I don't see where that's prevented e.g. RegisterForFileTypeActivation accepts strings with no validation which are then encoded as-is in the URI. Somewhere in that chain the parameters should be (1) validated and (2) URI escaped.

dev/AppLifecycle/ActivationRegistrationManager.cpp
```diff
@@ -8,47 +8,89 @@
...
+    }
+
+    std::wstring GenerateCommandLine(std::wstring const& modulePath)
+    {
+        std::wstring command = modulePath;
+
+        if (modulePath.empty())
+        {
+            command = GetModulePath();
+        }
```
DrusTheAxe, line 30: Suggestion: `std::wstring command{ modulePath.empty() ? GetModulePath() : modulePath };`

dev/AppLifecycle/ActivationRegistrationManager.cpp
```diff
@@ -8,47 +8,89 @@
...
+
+        command += L" " + c_argumentPrefix + c_protocolArgumentString + c_argumentSuffix;
```
DrusTheAxe, line 32: > command += L" " + c_argumentPrefix + c_protocolArgumentString + c_argumentSuffix; [](start = 7, length = 82)

This causes multiple reallocations and copies. Is there a more efficient option e.g. 

```
auto command{ std::wstring::concat(modulePath.empty() ? GetModulePath() : moduelPath, L" ", c_argumentPrefix, c_protocolArgumentString, c_argumentSuffix) };
```

dev/AppLifecycle/ActivationRegistrationManager.cpp
```diff
@@ -8,47 +8,89 @@
...
+#include "ExtensionContract.h"
+#include "EncodedLaunchExecuteCommand.h"
+
+extern "C" HRESULT __stdcall DllRegisterServer() noexcept;
 
 namespace winrt::Microsoft::ProjectReunion::implementation
 {
+    std::wstring GenerateEncodedLaunchUri(std::wstring const& appUserModelId, std::wstring const& contractId)
+    {
+        return c_encodedLaunchSchemeName + L"://" + appUserModelId + L"?ContractId=" + contractId;
```
DrusTheAxe, line 20: >L"?ContractId=" [](start = 69, length = 15)

Magic constant

riverar:
Are paths prefaced with `\\?\`, for longPath and other scenarios, supported? (e.g. `\\?\C:\Windows\notepad.exe`)

dev/AppLifecycle/ActivationRegistrationManager.cpp
```diff
@@ -8,47 +8,89 @@
...
         {
             throw hresult_illegal_method_call();
         }
 
-        auto appId = ComputeAppId();
+        auto appId = ComputeAppId(exePath.c_str());
         auto type = AssociationType::File;
         auto progId = ComputeProgId(appId, type);
 
         if (supportedFileTypes.size())
```
DrusTheAxe, line 49: > if (supportedFileTypes.size()) [](start = 7, length = 31)

If the array's empty what happens?

dev/AppLifecycle/ActivationRegistrationManager.cpp
```diff
@@ -8,47 +8,89 @@
...
         {
-            RegisterFileExtension(extension.c_str());
+            RegisterFileExtension(fileType.c_str());
             
             for (auto verb : supportedVerbs)
             {
-                auto command = GetModulePath();
-                command += L" ----" + c_fileArgumentString + L":" + verb.c_str() + L",%1";
+                auto command = GenerateCommandLine(exePath.c_str()) + GenerateEncodedLaunchUri(L"App",
+                    c_fileContractId) + L"&Verb=" + verb.c_str() + L"&File=%1";
```
DrusTheAxe, line 62: >verb [](start = 52, length = 4)

Any restrictions on allowed values? Required to be ascii? Can contain punctuation % & \ / : yadda yadda or is that invalid? I suspect implementation dependent errors can be returned e.g. if verb = "Cookies&Cream" or "A\B".

Bad input should result in E_INVALIDARG

dev/AppLifecycle/ActivationRegistrationManager.cpp
```diff
@@ -58,39 +100,75 @@ namespace winrt::Microsoft::ProjectReunion::implementation
...
+    {
+        if (HasIdentity())
+        {
+            throw hresult_illegal_method_call();
+        }
 
-        RegisterProgId(progId.c_str(), L"", applicationDisplayName.c_str(), logo.c_str());
+        // Example: HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Run
+        wil::unique_hkey key;
+        THROW_IF_WIN32_ERROR(::RegOpenKeyEx(HKEY_CURRENT_USER, c_runKeyPath.c_str(), 0, KEY_WRITE,
```
DrusTheAxe, line 137: >KEY_WRITE [](start = 88, length = 9)

Any concerns about WOW? e.g. x86 process calls this on an x64 machine.

DrusTheAxe:
>                 0, nullptr, 0, KEY_WRITE, nullptr, openWithKey.put(), nullptr));

Any concerns about WOW? e.g. should this be `KEY_WRITE | KEY_WOW64_64KEY`?

---
Refers to: dev/AppLifecycle/Association.cpp:423 in 70878ee. [](commit_id = 70878eeb30a7b81e625e3a8551a399a77e3e820e, deletion_comment = False)

dev/AppLifecycle/AppLifecycle.cpp
```diff
@@ -38,7 +49,33 @@ namespace winrt::Microsoft::ProjectReunion::implementation
...
+    bool IsEncodedLaunch(IProtocolActivatedEventArgs const& args)
+    {
+        return (args.Uri().SchemeName() == c_encodedLaunchSchemeName);
+    }
+
+    IActivatedEventArgs GetEncodedLaunchActivatedEventArgs(IProtocolActivatedEventArgs const& args)
+    {
+        for (auto const& pair : args.Uri().QueryParsed())
+        {
+            if (pair.Name() == L"ContractId")
```
DrusTheAxe, line 61: >ContractId [](start = 33, length = 10)

Is this case-sensitive? Is "contractid" equally valid?

The RFC for URIs says it's a per-protocol decision
[https://stackoverflow.com/questions/24699643/are-query-string-keys-case-sensitive](https://stackoverflow.com/questions/24699643/are-query-string-keys-case-sensitive)

dev/AppLifecycle/Association.cpp
```diff
@@ -54,13 +54,19 @@ namespace winrt::Microsoft::ProjectReunion::implementation
         return path;
     }
 
-    std::wstring ComputeAppId()
+    std::wstring ComputeAppId(const std::wstring& customSeed)
     {
         // PRF = Project Reunion Framework.  A simple prefix to help identify ProgIds we compute here.
-        // AppId = Prefix + Hash(modulePath)
```
DrusTheAxe, line 60: >Hash(modulePath) [](start = 28, length = 16)

If I move my exe it's a totally different value?
If I unzip some app to C:\Foo and register things, then move it to C:\Bar I need to register to pick up the changed path. Is this documented?

Also, I now have old data in the registry pointing to C:\Foo with no way for it to be removed?

Is there a way to avoid such WinRot? Can you hash just the filename of the exe? Some input provided by the app?
Is a Cleanup() API to find old orphaned data and delete them possible? Or is that ambiguous with valid Win32 practices?

DrusTheAxe:
>         std::wstring result{ c_progIdPrefix };

auto result{ std::to_wstring(hash) };

---
Refers to: dev/AppLifecycle/Association.cpp:74 in 70878ee. [](commit_id = 70878eeb30a7b81e625e3a8551a399a77e3e820e, deletion_comment = False)

dev/AppLifecycle/ActivationRegistrationManager.cpp
aeloros: This has nothing to do with DynamicDependencies.  This is purely because it is DllRegisterServer which is special and there is no reason to name a differently named function just to wrap this one.

dev/AppLifecycle/ActivationRegistrationManager.cpp
aeloros: Yes, that is done as part of the Uri class.

dev/AppLifecycle/EncodedLaunchExecuteCommand.cpp
```diff
@@ -0,0 +1,100 @@
...
+        if (SUCCEEDED(bindResult))
+        {
+            com_ptr<::IAssocHandler> assocHandler;
+            while (S_OK == enumHandlers->Next(1, assocHandler.put(), nullptr))
+            {
+                // Check if the assocHandler has an AppUserModelId that matches our target.
+                auto objWithAppId = assocHandler.try_as<::IObjectWithAppUserModelID>();
+                wil::unique_cotaskmem_string appUserModelId;
+                if (objWithAppId && SUCCEEDED(objWithAppId->GetAppID(&appUserModelId)) &&
+                    appUserModelId.get() == targetAppUserModelId)
```
DrusTheAxe, line 78: >appUserModelId.get() == targetAppUserModelId [](start = 20, length = 44)

`(CompareStringOrdinal(appUserModelId.get(), -1, targetAppUserModelId.c_str(), -1, TRUE) == CSTR_EQUAL)`

AUMID is case-insensitive

dev/AppLifecycle/EncodedLaunchExecuteCommand.cpp
```diff
@@ -0,0 +1,100 @@
...
+        THROW_IF_FAILED(m_selection->GetItemAt(0, item.put()));
+
+        LPWSTR buffer = nullptr;
+        THROW_IF_FAILED(item->GetDisplayName(SIGDN_URL, &buffer));
+        std::wstring_view rawUri{ buffer };
+
+        Uri uri{ rawUri };
+        auto host = uri.Host();
+
+        auto separator = rawUri.find(L":");
```
DrusTheAxe, line 27: >(L":") [](start = 36, length = 6)

NIT: ...find(L':');`

Slightly more efficient

dev/AppLifecycle/EncodedLaunchExecuteCommand.cpp
```diff
@@ -0,0 +1,100 @@
...
+        if (separator == std::wstring::npos)
+        {
+            return E_INVALIDARG;
+        }
```
DrusTheAxe, line 31: `RETURN_HR_IF(E_INVALIDARG, separator == std::wstring::npos)`

Better logging on failure

dev/AppLifecycle/EncodedLaunchExecuteCommand.cpp
```diff
@@ -0,0 +1,100 @@
...
+using namespace winrt::Windows::Foundation;
+
+namespace winrt::Microsoft::ProjectReunion::implementation
+{
+    IFACEMETHODIMP EncodedLaunchExecuteCommand::Execute() noexcept try
+    {
+        THROW_HR_IF_NULL(E_INVALIDARG, m_selection);
+
+        com_ptr<IShellItem> item;
+        THROW_IF_FAILED(m_selection->GetItemAt(0, item.put()));
```
DrusTheAxe, line 18: >THROW [](start = 8, length = 5)

Use RETURN_IF_FAILED inside this function (rather than THROW_IF_FAILED) since it returns HRESULT? RETURN_... is equivalent but more efficient

dev/AppLifecycle/EncodedLaunchExecuteCommand.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+        com_ptr<::IUnknown> site;
+        THROW_IF_FAILED(GetSite(IID_PPV_ARGS(&site)));
+        executionInfo.hInstApp = reinterpret_cast<HINSTANCE>(site.get());
+
+        auto succeeded = ShellExecuteEx(&executionInfo);
+        if (!succeeded)
+        {
+            auto resultCode = reinterpret_cast<int64_t>(executionInfo.hInstApp);
+            return static_cast<int32_t>(resultCode);
```
DrusTheAxe, line 52: >int32_t [](start = 31, length = 7)

Since the function returns HRESULT should this be `static_cast<HRESULT>(...)`?

dev/AppLifecycle/Association.cpp
aeloros: This is a known limitation that we're living with.  The scenarios where this go wrong are:

1. App moves legitimately as part of some app supported scenario.  The app is responsible for correcting it's associations anyway (remove re-add later).
2. App is moved by user even though it doesn't support it.  Nothing we can do here, as the EXE path would be wrong anyway and nothing would run.
3. The fact that the App can specify the seed means it can recreate the required information for later removal using the API.

There is no solution for this type of WinRot that can be solved in this component as it's an open playground in the registry.  Ultimately the plan is to move to a more stable seed, ideally something like the publisher from a signature or the like.  For now this is what we've decided to do.

dev/AppLifecycle/EncodedLaunchExecuteCommand.cpp
```diff
@@ -0,0 +1,100 @@
...
+        executionInfo.lpFile = encodedLaunchUri.c_str();
+
+        com_ptr<::IUnknown> site;
+        THROW_IF_FAILED(GetSite(IID_PPV_ARGS(&site)));
+        executionInfo.hInstApp = reinterpret_cast<HINSTANCE>(site.get());
+
+        auto succeeded = ShellExecuteEx(&executionInfo);
+        if (!succeeded)
+        {
+            auto resultCode = reinterpret_cast<int64_t>(executionInfo.hInstApp);
```
DrusTheAxe, line 51: >int64_t [](start = 47, length = 7)

Is hInstApp 64bit on WIN32 platforms e.g. x86?

Shouldn't this be `reinterpret_cast<std::uintptr_t>(executionInfo.hInstApp);`

[https://stackoverflow.com/questions/153065/converting-a-pointer-into-an-integer#26586211](https://stackoverflow.com/questions/153065/converting-a-pointer-into-an-integer#26586211)

dev/AppLifecycle/ExtensionContract.h
```diff
@@ -0,0 +1,32 @@
...
+
+#include "LaunchActivatedEventArgs.h"
+#include "ProtocolActivatedEventArgs.h"
+#include "FileActivatedEventArgs.h"
+#include "StartupActivatedEventArgs.h"
+
+namespace winrt::Microsoft::ProjectReunion::implementation
+{
+    // Extension contract ids.
+    static const std::wstring c_launchContractId = L"Windows.Launch";
```
DrusTheAxe, line 13: > static const std::wstring [](start = 3, length = 26)

constexpr? PCWSTR?

see previous comment

dev/ProjectReunion_DLL/COMExports.cpp
```diff
@@ -0,0 +1,105 @@
...
+#include "pch.h"
+#include "EncodedLaunchExecuteCommand.h"
+#include "Association.h"
+
+using namespace winrt::Microsoft::ProjectReunion::implementation;
+
+std::wstring GetDllModulePath();
+std::wstring GetCoClassPath(const GUID& iid);
+void RegisterCoClass(const GUID& iid);
+void UnregisterCoClass(const GUID& iid);
```
DrusTheAxe, line 13: Why?

Is this pre-DynamicDependencies temporary workaround or something more lasting?

dev/ProjectReunion_DLL/COMExports.cpp
```diff
@@ -0,0 +1,105 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#include "pch.h"
+#include "EncodedLaunchExecuteCommand.h"
+#include "Association.h"
+
+using namespace winrt::Microsoft::ProjectReunion::implementation;
+
+std::wstring GetDllModulePath();
```
DrusTheAxe, line 10: >GetDllModulePath [](start = 13, length = 16)

Why? Is this only used for the following functions or something else?

dev/ProjectReunion_DLL/COMExports.cpp
aeloros: You're going to have to articulate your question differently.

dev/ProjectReunion_DLL/COMExports.cpp
aeloros: You're going to have to articulate your question differently.

dev/ProjectReunion_DLL/pch.h
```diff
@@ -9,6 +13,7 @@
 #include <wil/token_helpers.h>
 #include <wil/resource.h>
 #include <wil/result_macros.h>
+#include <shellapi.h>
 #include <appmodel.h>
```
DrusTheAxe, line 17: shellapi.h + appmodel.h should come before WIL headers

Good header order is more generic to specific e.g.
* Windows
* Standard C++
* WIL
* ProjectReunion

ProjectReunion.sln
```diff
@@ -115,8 +115,8 @@ Global
 		{C62688A1-16A0-4729-B6ED-842F4FAA29F3}.Release|x64.Build.0 = Release|x64
 		{C62688A1-16A0-4729-B6ED-842F4FAA29F3}.Release|x86.ActiveCfg = Release|Win32
 		{C62688A1-16A0-4729-B6ED-842F4FAA29F3}.Release|x86.Build.0 = Release|Win32
-		{7C502995-59C3-483B-86BA-815985353633}.Debug_test|Any CPU.ActiveCfg = Release|x64
```
DrusTheAxe, line 118: >Debug_test [](start = 41, length = 10)

What's the Debug_test Configuration used for? Do any *.vcxproj reference it? Artifact of initial solution creation or is there a need?

Any reason we can't consolidate on Debug?

dev/AppLifecycle/ActivationRegistrationManager.cpp
aeloros: Ah, I think I see what you might be thinking.

Yes, this code is for a world where we don't have a framework package that we can load yet.  Until your changes make it into main, we are still pre-DynamicDependencies and thus need this.  I won't be removing it until your changes are in and my tests succeed.  So yeah, once I can actually register COM objects to be hosted inside the framework package this code must remain.

DrusTheAxe: I suspected. Or at least, I hoped so :-)

Just checking that there's not some longer-term reason we need it once DynDep, Framework package and other foundation's available

ProjectReunion.sln
aeloros: The SLN file is generated, and these changes are from VS.  Specifically me fixing one of the flavors.  It sounds like your question should be for Elliot and/or Ben and unrelated to my changes.

test/TestApps/AppLifecycleTestApp/main.cpp
aeloros: Since the event for each test is used multiple times for different reasons, the comment is to explain the current reason.

dev/AppLifecycle/Association.cpp
DrusTheAxe: >There is no solution for this type of WinRot that can be solved in this component

One option is a "LifetimeArtifact" style solution, similar to what @jonwis had suggested for a similar problem for Dynamic Dependencies. To handle when a definition's created and the software goes away without deleting the definition (dev forgets, app's deleted rather than 'properly' uninstalled, if an app moves it needs to register but not worry about deregistering previous definition, etc).

For instance, to detect WinRot when using a protocol

1. RegisterForProtocolActivation(...LifetimeArtifact="c:\foo\bar.exe")
2. When writing the activation info also RegWrite("LifetimeArtifact", REG_SZ, "c:\foo\bar.exe")
3. When the protocol is activated and the definition is retrieved from the registry, also retrieve LifetimeArtifact and if (!fileexists(filename)) then consider the definition 'expired' -- Unregister it (best-effort) and act as if the protocol registration didn't exist

Another option:
1. RegisterForProtocolActivation(...LifetimeArtifact="c:\foo\bar.exe")
2. When writing the activation info also write info to a central location (HKCU\SOFTWARE\Something, the ProjectReunionMain package's ApplicationData, etc)
3. Have a recurring job (e.g. BackgroundTask) enumerate this list, verify the LifetimeArtifact exists and if not call Unregister

DynamicDependencies supports this sort of garbage collection. MddTryCreatePackageDependency takes a filename or regkey. On use (MddAddDynamicDependency) it checks if a definition's 'expired', and an upcoming addition will regularly check for and remove obsolete DynamicDependencyLifetimeManager packages.

There may be other options. Bottom lie is to tie the definition to some durable artifact and when that goes away, the definition goes away. For Protocol/etc activation the executable can often be a good LifetimeArtifact as it's typically the target for a filetype or protocol activation. After all if the target exe goes away the definition likely isn't needed anymore either.

Question is how important is it to prevent WinRot.

IMO it's not significant for this PR _other than_ its affect on the API.

dev/ProjectReunion_DLL/COMExports.cpp
DrusTheAxe: Well that's a weird comment. Sorry, I don't know how this was created. Must have fat fingered the keyboard

dev/ProjectReunion_DLL/COMExports.cpp
DrusTheAxe: Let me rephase: Do you intend these functions to exist long-term and if so, why? Or is this a short-term workaround to enable your code to run and test until the Dynamic Depenendencies PR's checked in?

I can see why they're needed before DynDep. I don't see any comment about this being a temporary workaround for DynDep's absence and I'm not aware of a reason they'd be needed longer-term. So I was wondering if I'm missing something.

dev/AppLifecycle/Association.cpp
```diff
@@ -54,13 +54,19 @@ namespace winrt::Microsoft::ProjectReunion::implementation
         return path;
     }
 
-    std::wstring ComputeAppId()
+    std::wstring ComputeAppId(const std::wstring& customSeed)
     {
         // PRF = Project Reunion Framework.  A simple prefix to help identify ProgIds we compute here.
```
DrusTheAxe, line 59: SUGGESTION: Change PRFblah to Microsoft.AppLifecycle.blah

A few reasons:
* PRF is a bit cryptic and ambiguous
* Last year the point was made several times not to put "ProjectReunion" into code; I don't recall data constracts being mentioned though they'd seem to meet the spirit of the guidance.
* ProgIDs have large length limits (255?)

DrusTheAxe (changes requested):

dev/AppLifecycle/ActivationRegistrationManager.cpp
aeloros: URI escaping is handled in the Uri class as part of creating the arguments.  Encoding it early doesn't gain any validation as there aren't any wrong data options that cause validation to fail.  Everything is encoded with escape sequences regardless of the data.

The best that could be done would be to ensure the combination of everything is still a valid Uri before progressing, which is what I'll probably add for that here.  All that ensures is that the code in this component is correct given the two inputs which of course is good.

dev/AppLifecycle/ActivationRegistrationManager.cpp
aeloros: Not really as this is a process command line with arguments.  It is parsed by ShellExecute and a set of platform APIs before execution.  We will not be providing any preventative validation unless we find evidence of it being a common error to get it wrong, and why.

dev/AppLifecycle/ActivationRegistrationManager.cpp
aeloros: No, this key is not virtualized by WOW.

dev/AppLifecycle/AppLifecycle.cpp
aeloros: Nope, and I have had plans on changing this and one or two other places.  It's just going to be more resilient if we ignore case here.  The initial writing of this needs touching up.

dev/AppLifecycle/EncodedLaunchExecuteCommand.cpp
aeloros: yeah, this is the other place I think I meant in my comment about contractId.  I think all of the compares so far in the component need to be switched.

dev/ProjectReunion_DLL/COMExports.cpp
aeloros: I think we went over this in another comment.  Yes, this is temporary until we have a different solution.

dev/AppLifecycle/Association.cpp
aeloros: I'm not opposed to finding a different prefix that makes this more readable and/or more compliant with any thinking around using ProjectReunion in code.  The binary is named this currently, and I know that is staying while we will be changing the namespace.  If this specific instance is not required to change due to whatever ruleset governs that, then we only have readability to go off of for reasons to change this.  

That being said, using the namespace isn't a good option either.

Also, ProgIds do not have a large limit.  They are max of 39.  https://docs.microsoft.com/en-us/windows/win32/com/-progid--key

aeloros:
> Are paths prefaced with `\\?\`, for longPath and other scenarios, supported? (e.g. `\\?\C:\Windows\notepad.exe`) I haven't run the scenario through all this code yet, so just throwing this out there early.

Is this in reference to some piece of code in the PR?  Could you provide more context please?

dev/ProjectReunion_DLL/COMExports.cpp
aeloros: Actually I'm going to not touch the CoClass stuff for consts since it will be removed once I consume the DynDep changes.

dev/AppLifecycle/ActivationRegistrationManager.cpp
aeloros: Small performance cost for readability and ease of use.

dev/AppLifecycle/EncodedLaunchExecuteCommand.cpp
aeloros: This was a cut-n-paste from a totally different usage.  It's definitely wrong, but not for the reasons you think.  I'll update.  This member variable is special in that it's specifically a re-used compat value for old Win16 days.  So it's been documented to be treated a very specific way.  In our current usage this is never a pointer, but int64 is still the wrong size.

DrusTheAxe (approved):

DrusTheAxe:
> Are paths prefaced with `\\?\`, for longPath and other scenarios, supported? (e.g. `\\?\C:\Windows\notepad.exe`)

`\\?\` is extended path notation, supporting up to ~32K but it's an NT path notation and not entirely handled the same as a Win32 path, so one should only use it when necessary. There's also complications handling share vs non-share notations

```
(Non-extended) drive/path notation is x:\path...
Extended path notation for drive/path values is \\?\x:\path...

(Non-extended) UNC notation is \\server\share\...
Extended path notation for UNC is \\?\UNC\server\share\...

Yes, syntax for UNC with extended path notation differs from drive/path -- \\?\ vs \\?\UNC\ prefix
```

We shouldn't be explicitly converting paths into extended path notation, but we shouldn't reject such paths either. As much of Project Reunion is APIs we can (and should) leave it up to callers to hand APIs the appropriate data.

Long-term enabling long path support is the right answer -- non-extended (aka 'normal' :P) notation but paths up to 32K instead of MAX_PATH. One day... ;-)

riverar:
@DrusTheAxe Yep, our understanding of extended paths is aligned.

I was just asking because [we landed on Reunion supporting `longPathAware`](https://github.com/microsoft/ProjectReunion/pull/444#discussion_r575626186) via fusion manifest elsewhere. I know extended paths aren't needed in that scenario but it was unclear if we were going "whole hog" here in terms of paths. Am in agreement with the stated goals 👍

andrewchoi5 (approved):

PR #459, opened 2021-02-19
Stop building ARM for ProjectReunion Foundation

kythant:
There was a decision to not support ARM and some of the packages we are using will not have an ARM version. 

Don't want to do have to do something like
#ifndef arm32
in our code.

### DIFF
--- a/build/ProjectReunion-BuildFoundation.yml
+++ b/build/ProjectReunion-BuildFoundation.yml
@@ -29,10 +29,6 @@ jobs:
         buildConfiguration: 'Release'
         normalizedConfiguration: 'fre'
         PGOBuildMode: 'Optimize'
-      Release_Arm:
-        buildPlatform: 'arm'
-        buildConfiguration: 'Release'
-        normalizedConfiguration: 'fre'
       Release_Arm64:
         buildPlatform: 'arm64'
         buildConfiguration: 'Release'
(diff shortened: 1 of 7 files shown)


huichen123:
please search "buildPlatform". there are other places where arm is built

build/ProjectReunion-BuildFoundation.yml
```diff
@@ -29,10 +29,6 @@ jobs:
         buildConfiguration: 'Release'
         normalizedConfiguration: 'fre'
         PGOBuildMode: 'Optimize'
-      Release_Arm:
-        buildPlatform: 'arm'
-        buildConfiguration: 'Release'
-        normalizedConfiguration: 'fre'
       Release_Arm64:
```
DrusTheAxe, line 32: Why do Release_x86 + Release_x64 have PGOBuildMode:Optimize and Release_Arm64 doesn't?

DrusTheAxe (approved):

BenJKuhn (approved):

EHO-makai (approved):
I assume we will follow up with the other places that attempt to build or package up arm32.

build/ProjectReunion-BuildFoundation.yml
kythant: Sounds like optimize is good to have for a release build. Should I add it to Arm64?

huichen123:
please also search "arm".

build/ProjectReunion-BuildFoundation.yml
BenJKuhn: Good question to save as an issue / bug. I wouldn't get hung up on it in this review.

huichen123 (approved):

PR #460, opened 2021-02-19
add appx manifest to foundation transport package

erik0686:
Adding an Appx Manifest that was previously being hardcoded into the PR Framework Package. Now it will be obtained from the Project Reunion Foundation Transport package.

### DIFF
--- a/build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
+++ b/build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
@@ -47,9 +47,9 @@ jobs:
         LICENSE
       TargetFolder: '${{ parameters.fullnupkgdir }}'
 
-  # Copy the Project Reunion full package specific target files into the full package folder
+  # Copy the Project Reunion full package specific target files and AppxManifest into the full package folder
   - task: PowerShell@2
-    displayName: 'Copy Project Reunion full package targets'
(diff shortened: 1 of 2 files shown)


build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
```diff
@@ -57,7 +57,8 @@ jobs:
         $fullpackagePath = "${{ parameters.fullnupkgdir }}"
         Copy-Item -Path "$targetsFilePath\ProjectReunion-Nuget-Native.targets" -Destination "$fullpackagePath\build\native\Microsoft.ProjectReunion.Foundation.targets"
         Copy-Item -Path "$targetsFilePath\Microsoft.ProjectReunion.Foundation.props" -Destination "$fullpackagePath\build\native\Microsoft.ProjectReunion.Foundation.props"
-
+        Copy-Item -Path "$targetsFilePath\AppxManifest.xml" -Destination "$fullpackagePath\AppxManifest.xml"
```
Scottj1s, line 60: instead of piggybacking on this task that's specific to targets, either:
- update the comment and displayName
- create a separate task, similar to 'copy license to full nuget'

Scottj1s (approved):
looks good - maybe just factor out the copy to prevent rot

Scottj1s (approved):

Scottj1s (approved):

PR #472, opened 2021-02-23
Use Framework UDK to get default resource path

huichen123:


dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/src/Microsoft.ApplicationModel.Resources.vcxproj
```diff
@@ -188,15 +180,24 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </ProjectReference>
   </ItemGroup>
+  <PropertyGroup>
+    <Ixp-Arch Condition="'$(Platform)' == 'AnyCPU'">x86</Ixp-Arch>
```
axelandrejs, line 184: Can we call this something more generic than IXP?

huichen123: We don't want to accidentally collide with a built in name. do you have a recommendation?

axelandrejs: Hmmm....are we the first to consume this kind of package? What about Udk-Arch?

axelandrejs (approved):

PR #477, opened 2021-02-24
Make MRTCore work in framework package

huichen123:
I will stamp ProjectReunionFrameworkPackage in aggregation pipeline.

### DIFF
--- a/dev/MRTCore/packaging/native/Microsoft.ApplicationModel.Resources.WinRt.props
+++ b/dev/MRTCore/packaging/native/Microsoft.ApplicationModel.Resources.WinRt.props
@@ -16,7 +16,7 @@
   <ItemGroup Condition="'$(OutputType)' == 'Exe'">
     <Reference Include="Microsoft.ApplicationModel.Resources.winmd">
       <HintPath>$(MSBuildThisFileDirectory)..\..\lib\uap10.0\Microsoft.ApplicationModel.Resources\Microsoft.ApplicationModel.Resources.winmd</HintPath>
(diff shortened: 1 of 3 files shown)


axelandrejs (approved):

rohanp-msft (approved):

dev/MRTCore/packaging/native/Microsoft.ApplicationModel.Resources.WinRt.props
```diff
@@ -16,7 +16,7 @@
   <ItemGroup Condition="'$(OutputType)' == 'Exe'">
     <Reference Include="Microsoft.ApplicationModel.Resources.winmd">
```
rohanp-msft, line 17: ><Reference Include="Microsoft.ApplicationModel.Resources.winmd"> [](start = 4, length = 64)

What about this? Is adding this reference OK?

huichen123: we need to have this reference. or we will get build break.

rohanp-msft: won't the winmd ref be added by the nuget package install? just like the DLL refs, etc.?

huichen123: you added that. So I guess it's a question for you? :)

rohanp-msft: currently, when the nuget package is installed, it does add the winmd ref

Scottj1s:
thanks Hui!

PR #478, opened 2021-02-24
Remove pfx and setup Azure KeyVault to pull a .cer file #DRAFT

kythant:
Remove pfx and cer file and setup Azure KeyVault to pull a .cer file


riverar:
Recommend instead using [azuresigntool](https://github.com/vcsjones/AzureSignTool) and skipping all the file-based operations completely. (Example in our production workflow https://github.com/File-New-Project/EarTrumpet/blob/master/.github/workflows/main.yml#L258-L259).

DrusTheAxe (changes requested):
MSTest.cer/pfx is a test cert **required** to for inner-loop development. DynamicDependencies tests require signed packages, as will everyone else using DynamicDependencies to test not-packaged scenarios.

MSTest.cer/pfx shouldn't be used to sign real packages and you may even have an alternative solution for tests executed during the CI/Pipeline but how do you expect inner-loop testing to happen?

BenJKuhn:
Closing old or abandoned pull requests. If you would like to pick this up again, address feedback, and push it over the finish line go ahead and reactivate.

PR #480, opened 2021-02-25
Fix flakey builds on nuget restore

kythant:
Attempt to fix the flakey builds caused by the solution using restore with a private nuget feed. 401 errors appears inconsistantly. The two combination that seems to work are:
Renewing the PAT on the service connection with slightly different permissions. 
Calling Nuget Authenticate on the feed.


huichen123 (approved):

kythant:
image

PR #494, opened 2021-02-26
Add PR dependencies feed for Taef Nuget

kythant:
Add PR dependencies feed for Taef Nuget


build/Helix/NuGet.config
```diff
@@ -19,7 +19,8 @@
     <clear />
     <add key="NuGet official package source" value="https://nuget.org/api/v3/" />
     <add key="dotnetfeed" value="https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json" />
-    <add key="WindowsES-External" value="https://microsoft.pkgs.visualstudio.com/_packaging/WindowsES-External/nuget/v3/index.json" />	
+    <add key="WindowsES-External" value="https://microsoft.pkgs.visualstudio.com/_packaging/WindowsES-External/nuget/v3/index.json" />
```
huichen123, line 22: do we still need this? also we can remove those authenticate steps

kythant: Yea we don't

huichen123 (approved):

riverar (changes requested):
This still fails in the public build. CppTest still has the bad reference.

From the referenced issue:
https://github.com/microsoft/ProjectReunion/blob/93e786dc7461e41ddd4afc2476a4d1cf765a6bdd/test/CppTest/CppTest.vcxproj#L342

kythant:
> This still fails in the public build. CppTest still has the bad reference.
> 
> From the referenced issue:
> https://github.com/microsoft/ProjectReunion/blob/93e786dc7461e41ddd4afc2476a4d1cf765a6bdd/test/CppTest/CppTest.vcxproj#L342

Are you absolutely sure? Nothing has changed on that end. THe same version was uploaded.

riverar:
@kythant The line highlighted performs the Error action if the `..\..\packages\Microsoft.Taef.10.58.210222006-develop\build\Microsoft.Taef.targets` file does not exist. Because I don't have access to that package, this will always fail when compiling CppTest.

This should be easy to repro on an external machine. Happy to provide RDP to a test machine if you need one for testing this PR.

BenJKuhn (approved):

riverar:
Looks like the Taef package is now available; build passes. 👍

kythant:
> @kythant The line highlighted performs the Error action if the `..\..\packages\Microsoft.Taef.10.58.210222006-develop\build\Microsoft.Taef.targets` file does not exist. Because I don't have access to that package, this will always fail when compiling CppTest.
> 
> This should be easy to repro on an external machine. Happy to provide RDP to a test machine if you need one for testing this PR.

Did you restore the package? I've added it to PR-dependencies so you can pull the package from it.

YexuanXiao:
It looks like the package has been removed from the source at the moment, so it's still not possible to build the project.

PR #502, opened 2021-03-01
Add Background Activation stubs

danielayala94:
Added Background Activation class declarations to the IDL and created the header/cpp files for them.


riverar:
@danielayala94 So, err, is this now dead or?

PR #504, opened 2021-03-01
Adding code and tests for the GetFor family of methods

dhoehna:
This PR includes implementation and tests for the methods

* GetForProcess
* GetForUser
* GetForMachine


test/Microsoft.Process.Environment.Tests/EnvironmentManagerCentennialTests.cpp
```diff
@@ -0,0 +1,26 @@
+﻿#include <pch.h>
+#include <EnvironmentManagerCentennialTests.h>
+
+using namespace winrt::Microsoft::ProjectReunion;
+
+namespace ProjectReunionCppTest
+{
+    void EnvironmentManagerCentennialTests::CentennialTestGetForProcess()
+    {
+        EnvironmentManager environmentManager = EnvironmentManager::GetForProcess();
```
DrusTheAxe, line 10: NIT: `EnvironmentManager environmentManager{ EnvironmentManager::GetForProcess() };`

See https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#es23-prefer-the--initializer-syntax

test/Microsoft.Process.Environment.Tests/EnvironmentManagerCentennialTests.cpp
```diff
@@ -0,0 +1,26 @@
...
+        VERIFY_IS_NOT_NULL(environmentManager);
```
DrusTheAxe, line 11: Can you do test more than just you got anything? Verify you got the right type of environment manager?

test/Microsoft.Process.Environment.Tests/EnvironmentManagerWin32Tests.cpp
```diff
@@ -0,0 +1,32 @@
+﻿#include <pch.h>
+#include <EnvironmentManagerWin32Tests.h>
+
+using namespace winrt::Microsoft::ProjectReunion;
+
+namespace ProjectReunionCppTest
+{
+    bool EnvironmentManagerWin32Tests::TestSetup()
+    {
+        winrt::init_apartment(winrt::apartment_type::single_threaded);
```
DrusTheAxe, line 10: It's TAEF so you can do this via `TEST_CLASS_PROPERTY(L"ThreadingModel", L"STA")` instead of code inside the function

And why STA? Are your tests doing GUI or something that can't do MTA?

And of course...you don't do anything else in TestSetup so why have it? Or is this a placeholder for uipcoming additions?

test/Microsoft.Process.Environment.Tests/EnvironmentManagerWin32Tests.h
```diff
@@ -0,0 +1,15 @@
+﻿#pragma once
+
+namespace ProjectReunionCppTest
+{
+    class EnvironmentManagerWin32Tests {
+        BEGIN_TEST_CLASS(EnvironmentManagerWin32Tests)
+            TEST_CLASS_PROPERTY(L"ActivationContext", L"CppTest.dll.manifest")
```
DrusTheAxe, line 7: Ooooh. That's cute. I haven't see `ActivationContext` before. What's that do?

test/Microsoft.Process.Environment.Tests/EnvironmentManagerWin32Tests.cpp
dhoehna: That is an awesome idea.  Lemme change it.

To answer your question of "Why" it is because this is the example I saw for reg-free WinRT.  Even though I saw examples of setting the Threading model in the TEST_CLASS_PROPERTY I didn't connect the two.

And no, this is not a place holder for something in the future.

test/Microsoft.Process.Environment.Tests/EnvironmentManagerWin32Tests.h
dhoehna: Activation Context is for reg-free winrt.  This SxS manifest tells cppTest what dll to use and the class to use from the dll.

Once Dynamic Dependencies comes out I can switch this all out.

dhoehna: One consequence (Maybe) of adding this dependency is that ProjectReunion_DLL will build anytime cpptest is being built.

test/Microsoft.Process.Environment.Tests/EnvironmentManagerCentennialTests.cpp
dhoehna: That is a good idea.  That would require me adding the Scope enum to the IDL so I can expose it to the testing code.

Is it okay to alter the API space after the initial review from my first PR?

DrusTheAxe: I'd think so in this case. But check with your API rep. Maybe just needs a spec update and a simple (offline/email) API Review of the change. OR updated spec and your API rep brings the change to API Review and lets you know if there's any problem.

jonwis: Instead, write a test that does something with the process-scoped env manager that shouldn't be visible to another process.  Like:

1. Get the env manager here
2. Set a property on it
3. Read the real registry and verify that the set did not go through to the persistent hive
4. Create a child process from here
5. Verify that the set property from 2 does show up in the new child process

jonwis: Also feel free to _not_ have the "verify this is not null." If it ever does become null, the follow-on "set" calls will access-violate and tell you the GetForProcess didn't return something.

test/Microsoft.Process.Environment.Tests/EnvironmentManagerWin32Tests.h
DrusTheAxe: :-(

Don't add to CppTest. That really should be renamed AppLifecycleTest.

"Product" code cohabitates in ProjectReunion_DLL but for tests isolation is waaaay better. Adding multiple features into 1 test DLL creates unnecessary tying and friction. This is worse than product code as tests often do a lot more and, to be blunt, weird things. Making all that cohabitate nicely is harder than doing it for product code. It's also degrades the inner-loop - last thing I want to deal with is some having to rebuilding EVERY test when I'm iterating in 1 specific area.

Or to put it simply: Politicians take make small problems and roll them up into one big problem. Developers take one big problem and break it down into many small problems. Be a developer :P

test/CppTest/CppTest.vcxproj
```diff
@@ -324,10 +327,21 @@
       <Project>{b73ad907-6164-4294-88fb-f3c9c10da1f1}</Project>
       <Private>true</Private>
     </ProjectReference>
-    <ProjectReference Include="..\TestApps\AppLifecycleTestApp\AppLifecycleTestApp.vcxproj">	
-      <Project>{7c502995-59c3-483b-86ba-815985353633}</Project>	
+    <ProjectReference Include="..\TestApps\AppLifecycleTestApp\AppLifecycleTestApp.vcxproj">
```
DrusTheAxe, line 330: Why? I don't understand the reason for this change

test/Microsoft.Process.Environment.Tests/Microsoft.Process.Environment.Tests.vcxitems
```diff
@@ -0,0 +1,34 @@
...
+    <ClCompile Include="$(MSBuildThisFileDirectory)EnvironmentManagerUWPTests.cpp" />
+    <ClCompile Include="$(MSBuildThisFileDirectory)EnvironmentManagerWin32Tests.cpp" />
+  </ItemGroup>
+  <ItemGroup>
+    <None Include="$(MSBuildThisFileDirectory)PropertySheet.props" />
+  </ItemGroup>
+  <ItemGroup>
+    <Xml Include="$(MSBuildThisFileDirectory)AppxManifest.pkg.xml" />
+    <Xml Include="$(MSBuildThisFileDirectory)CentennialAppxManifest.pkg.xml" />
+    <Xml Include="$(MSBuildThisFileDirectory)CppTest.dll.manifest" />
```
DrusTheAxe, line 32: Why? Are you adding EnvironmentManager tests into CppTest? Please don't; make your own test project(s).

See previous comment

dhoehna: I'll answer this question for posterity and to maybe figure out a better way to make a test class.

The reason why I added my tests to CppTest with a shared project is because that is what I saw AppLifecycleTest do.

Should I create my own dll for testing?

test/CppTest/CppTest.vcxproj
dhoehna: I don't know either.  I bet I mis-clicked and added the shared project by mistake.

test/Microsoft.Process.Environment.Tests/EnvironmentManagerCentennialTests.cpp
dhoehna: > Instead, write a test that does something with the process-scoped env manager that shouldn't be visible to another process. Like:
> 
> 1. Get the env manager here
> 2. Set a property on it
> 3. Read the real registry and verify that the set did not go through to the persistent hive
> 4. Create a child process from here
> 5. Verify that the set property from 2 does show up in the new child process

I'm confused.  I haven't implemented SetEnvironmentVariable yet, so I can't call it.  Additionally, there are no properties to set.  If I did set a property, why would it go to the registry?

dhoehna: > Also feel free to _not_ have the "verify this is not null." If it ever does become null, the follow-on "set" calls will access-violate and tell you the GetForProcess didn't return something.

Isn't that the point of the GetFor[Process\User\Machine] calls, to get a non-null object?

test/Microsoft.Process.Environment.Tests/Microsoft.Process.Environment.Tests.vcxitems
DrusTheAxe: Yes. Create your own project creating a DLL for your tests. Or multiple projects if necessary (DynamicDependencies has 7 test projects)

Add a feature subdir under test like under dev and put your test project in there. Or make subdirs under there if need be. I recommend the latter as it's easier if you later need to add a 2nd project e.g.

```
test\
    EnvironmentManager\
        EnvironmentManagerTest\
            EnvironmentManagerTest.vcxproj
            ...
```
or
```
test\
    EnvironmentManager\
        kitties\
            ...
        puppies\
            ...
```

Use your judgement. Main point is keep your tests in your own test project, separate from other features. TestExplorer makes it easy to run tests across features if/when needed but not drag down day-to-day inner-loop.

test/Microsoft.Process.Environment.Tests/EnvironmentManagerCentennialTests.cpp
dhoehna: ping @jonwis

jonwis: I see. This PR only sets up the scaffolding for the code, and the methods that "do the things" are still coming.

I guess for now, I'd add modify the test to actually detect the "not implemented" state, like:

```c++
VERIFY_THROWS(manager.SetEnvironmentVariable(), winrt::hresult_not_implemented);
```

jonwis (approved):

test/Microsoft.Process.Environment.Tests/EnvironmentManagerWin32Tests.h
dhoehna: @DrusTheAxe how does this look?

PR #505, opened 2021-03-01
Added HttpRequest to TestAppServer

pmpurifoy:
Added HttpRequest to TestAppServer when provided a ChannelUri

### DIFF
--- a/test/TestApps/PushNotificationsWin32App/main.cpp
+++ b/test/TestApps/PushNotificationsWin32App/main.cpp
@@ -1,12 +1,70 @@
 ﻿#include "pch.h"
 #include <iostream>
+#include <sstream>
+#include <winrt/Windows.Foundation.h>
+#include <winrt/Windows.Security.Cryptography.h>
+#include <winrt/Windows.Storage.Streams.h>
+#include <winrt/Windows.Web.Http.Headers.h>
+
 using namespace winrt;
+using namespace winrt::Windows::Storage;
+using namespace winrt::Windows::Storage::Streams;
 using namespace winrt::Windows::Foundation;
+using namespace winrt::Windows::ApplicationModel::Activation;
 using namespace winrt::Microsoft::ProjectReunion;
+using namespace Windows::Web::Http;
+
+std::wstring RequestChannel()
+{
+    return // Insert ChannelUri here.;
+}
+
+std::wstring BuildNotificationPayload()
+{
+    std::wstring channel = RequestChannel();
+    std::wstring channelUri = L"\"ChannelUri\":\"" + channel + L"\"";
+    std::wstring x_wns_type = L"\"X_WNS_Type\": \"wns/raw\"";
+    std::wstring contentType = L"\"Content_Type\": \"application/octet-stream\"";
+    std::wstring payload = L"\"Payload\": \"<toast></toast>\"";
+    std::wstring delay = L"\"Delay\": \"false\"";
+    return { L"{" + channelUri + L"," + x_wns_type + L"," + contentType + L"," + payload + L"," + delay + L"}" };
+}
+
+void SendRequestToServer()
+{
+    HttpResponseMessage httpResponseMessage;
+    std::wstring httpResponseBody;
+    try {
+        // Construct the HttpClient and Uri. This endpoint is for test purposes only.
+        HttpClient httpClient;
+        Uri requestUri{ L"https://notificationtestappserver.azurewebsites.net/api/PostPushNotification?" };
+
+        // Construct the JSON to post.
+        HttpStringContent jsonContent(
+            BuildNotificationPayload(),
+            UnicodeEncoding::Utf8,
+            L"application/json");
+
+        // Post the JSON, and wait for a response.
+        httpResponseMessage = httpClient.PostAsync(
+            requestUri,
+            jsonContent).get();
+
+        // Make sure the post succeeded, and write out the response.
+        httpResponseMessage.EnsureSuccessStatusCode();
+        httpResponseBody = httpResponseMessage.Content().ReadAsStringAsync().get();
+        std::wcout << httpResponseBody.c_str() << std::endl;
+    }
+    catch (winrt::hresult_error const& ex)
+    {
+        std::wcout << ex.message().c_str();
+    }
+}
 
 int main()
 {
-    // Add your code here.
-    std::cout << "Hello World!\n";
+    bool succeeded = true;
+    SendRequestToServer();
+    std::getchar();
     return 0;
 }


pmpurifoy:
When I make the HttpRequest, I get this error 'Class not registered. The process will receive the payload from the TestAppServer and exit with return code 0, but in the console output I get this error.

danielayala94:
> When I make the HttpRequest, I get this error 'Class not registered'. The process receives the payload from the TestAppServer and exit with return code 0, but in the console output I get this error.

@pmpurifoy so are we successful receiving data from the test app server? Then the "class not registered" sounds like mere noise?

test/TestApps/PushNotificationsWin32App/main.cpp
```diff
@@ -1,12 +1,70 @@
...
+    }
+}
 
 int main()
 {
-    // Add your code here.
-    std::cout << "Hello World!\n";
+    bool succeeded = true;
+    SendRequestToServer();
+    std::getchar();
```
danielayala94, line 68: I guess this is to have the time to verify the server response?

test/TestApps/PushNotificationsWin32App/main.cpp
```diff
@@ -1,12 +1,70 @@
...
+            L"application/json");
+
+        // Post the JSON, and wait for a response.
+        httpResponseMessage = httpClient.PostAsync(
+            requestUri,
+            jsonContent).get();
+
+        // Make sure the post succeeded, and write out the response.
+        httpResponseMessage.EnsureSuccessStatusCode();
+        httpResponseBody = httpResponseMessage.Content().ReadAsStringAsync().get();
```
danielayala94, line 55: How a success response looks like? And a failure response?

test/TestApps/PushNotificationsWin32App/main.cpp
```diff
@@ -1,12 +1,70 @@
...
+    {
+        std::wcout << ex.message().c_str();
+    }
+}
 
 int main()
 {
-    // Add your code here.
-    std::cout << "Hello World!\n";
+    bool succeeded = true;
```
danielayala94, line 66: Not needed

test/TestApps/PushNotificationsWin32App/main.cpp
```diff
@@ -1,12 +1,70 @@
...
+            UnicodeEncoding::Utf8,
+            L"application/json");
+
+        // Post the JSON, and wait for a response.
+        httpResponseMessage = httpClient.PostAsync(
+            requestUri,
+            jsonContent).get();
+
+        // Make sure the post succeeded, and write out the response.
+        httpResponseMessage.EnsureSuccessStatusCode();
```
hulumane, line 54: How would the caller of this process know if the request to server has failed? Does this throw an exception?

test/TestApps/PushNotificationsWin32App/main.cpp
```diff
@@ -1,12 +1,70 @@
...
+        httpResponseBody = httpResponseMessage.Content().ReadAsStringAsync().get();
+        std::wcout << httpResponseBody.c_str() << std::endl;
+    }
+    catch (winrt::hresult_error const& ex)
+    {
+        std::wcout << ex.message().c_str();
```
hulumane, line 60: I believe this method should throw and the caller should be handling all exceptions here.

test/TestApps/PushNotificationsWin32App/main.cpp
hulumane: This is a bit strange. Why are we waiting for input here?

test/TestApps/PushNotificationsWin32App/main.cpp
```diff
@@ -1,12 +1,70 @@
...
+using namespace winrt::Windows::Storage;
+using namespace winrt::Windows::Storage::Streams;
 using namespace winrt::Windows::Foundation;
+using namespace winrt::Windows::ApplicationModel::Activation;
 using namespace winrt::Microsoft::ProjectReunion;
+using namespace Windows::Web::Http;
+
+std::wstring RequestChannel()
+{
+    return // Insert ChannelUri here.;
```
hulumane, line 19: Not sure what this method is for

test/TestApps/PushNotificationsWin32App/main.cpp
```diff
@@ -1,12 +1,70 @@
...
+}
 
 int main()
 {
-    // Add your code here.
-    std::cout << "Hello World!\n";
+    bool succeeded = true;
+    SendRequestToServer();
+    std::getchar();
     return 0;
```
hulumane, line 69: IMHO, main should return a failure if the Send operation fails. Otherwise it should return success.

test/TestApps/PushNotificationsWin32App/main.cpp
danielayala94: I think they just want to delay the program exit to see the how the http response output looks like.

We can have a sleep timer there, or say "press a button to exit", but let's not just check in the getchar() trick 😁.

pmpurifoy: When running the application with 'Local Windows Debugger', once the process has completed main(), it displays the error code and waits for user input to close the window. If we were to run the application as packaged and installed on our local machine and run it as a program, it will complete the main(), then kill itself but we aren't able to see any output. std::getchar() allows us to see what succeeded/failed.

hulumane: So couple of things here:
1) The function should wait till the http response comes back and log that either in command line or UI depending on how this process is called.
2) Main should return per-determined error codes so that the caller knows that we failed or succeeded.

Can we achieve 1) and 2)?

hulumane: Again it comes down to how this tool is going to be used for automation scenarios. If it is infact being used for automation, I would get rid of the std::getchar() since it doesnt; really do much (outside of debuggability) and that problem can be solved by the solutions above.

test/TestApps/PushNotificationsWin32App/main.cpp
pmpurifoy: It's basically a method stub for Sharath to fill with the ProjectReunion component. Currently I am requesting a channel on our NotificationsTestApp and hardcoded that channel in the console app.

hulumane: Ah. Instead of doing that, can you simply call into our Public WinRT APIs here and return the string?

test/TestApps/PushNotificationsWin32App/main.cpp
pmpurifoy: It throws an exception and displays error request message received from TestAppServer. But as of right now, there is no BadRequestObjectResult being returned from the server, so for local testing I added it.

test/TestApps/PushNotificationsWin32App/main.cpp
pmpurifoy: Success: Whatever string is in the the OkObjectResult from the TestppServer. (Currently there is no string, so it returns nothing but I added string for local testing).
Failure: There is no failure ObjectResult being returned from TestAppServer, so I added a BadRequestObjectResult in TestAppServer for when the request returns a non-success error code.

pmpurifoy:
This function should be in the UI app. This is the console app. Will address comments in PR for the UI app.

PR #513, opened 2021-03-03
TAEF Framework for packaged win32 app

pmpurifoy:
Built TAEF framework for testing packaged application. Launches the MSIX app with a protocol and the processes communicate via events to determine success/failure.


sharath2727:
@pmpurifoy - can you please create a child branch from WNP_Reunion for functional tests here and check this in there. As discussed offline, WNP_Reunion is only for components. Thanks.

PR #522, opened 2021-03-04
Prepare MRTCore binaries for scanning tools.

rohanp-msft:
**How verified**
Built the three projects affected.


jonwis:
Are the other DLLs in here (the Project Reunion main one) already building this way?

dev/MRTCore/mrt/Core/src/MRM.vcxproj
```diff
@@ -104,7 +104,7 @@
            supported according to the CRT maintainer. Dynamic linking against the CRT makes the binaries a bit smaller
            than they would otherwise be if the CRT, runtime, and STL were all statically linked in. -->
       <IgnoreSpecificDefaultLibraries>%(IgnoreSpecificDefaultLibraries);libucrtd.lib</IgnoreSpecificDefaultLibraries>
-      <AdditionalOptions>%(AdditionalOptions) /defaultlib:ucrtd.lib</AdditionalOptions>
+      <AdditionalOptions>%(AdditionalOptions) /defaultlib:ucrtd.lib /debug /debugtype:cv,fixup /opt:ref /opt:icf /incremental:no</AdditionalOptions>
```
huichen123, line 107: Is APIScan only done on debug build? what if release build does something different (although it's not recommended)?

rohanp-msft:
> Are the other DLLs in here (the Project Reunion main one) already building this way?

I don't know. I tackled just the MRTCore ones. Would offer to do the main one too but I have another blocking Reunion bug on me.

riverar:
Can you elaborate a bit on what this is? Externals don't have access to the linked resource.

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/src/Microsoft.ApplicationModel.Resources.vcxproj
```diff
@@ -109,7 +109,7 @@
       <WarningLevel>Level4</WarningLevel>
       <AdditionalOptions>%(AdditionalOptions) /bigobj</AdditionalOptions>
       <!--Temporarily disable cppwinrt heap enforcement to work around xaml compiler generated std::shared_ptr use -->
-      <AdditionalOptions Condition="'$(CppWinRTHeapEnforcement)'==''">/DWINRT_NO_MAKE_DETECTION %(AdditionalOptions)</AdditionalOptions>
+      <AdditionalOptions Condition="'$(CppWinRTHeapEnforcement)'==''">/DWINRT_NO_MAKE_DETECTION /Zi /Gf /Gy %(AdditionalOptions)</AdditionalOptions>
```
jonwis, line 112: Split this into its own line - otherwise this only takes effect when the cppwinrtheapenforcement setting is enabled.

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/src/Microsoft.ApplicationModel.Resources.vcxproj
```diff
@@ -126,6 +126,9 @@
     <ClCompile>
       <PreprocessorDefinitions>_DEBUG;%(PreprocessorDefinitions)</PreprocessorDefinitions>
     </ClCompile>
+    <Link>
+      <AdditionalOptions Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">/debug /debugtype:cv,fixup /opt:ref /opt:icf /incremental:no %(AdditionalOptions)</AdditionalOptions>
```
jonwis, line 130: Are there individual linker/build settings to enable these, rather than modifying the options settings directly?

jonwis: Turning on no-increment-link and ICF is going to make the developer inner loop more costly.  Can we limit the impact only to release bits? While I agree with Hui that there's danger in letting debug & release have different behaviors, the scans should only be on the code we're releasing.

jonwis (changes requested):
Fix the cppwinrtheap one, please; and check with @BenJKuhn whether this change can be limited to release bits so debug builds can retain incrementality.

BenJKuhn:
> Can you elaborate a bit on what this is? Externals don't have access to the linked resource.

The most I can say is that tools are non-public and validate specific Microsoft policies & guidelines that don't have much meaning outside of our organization. They inspect compiled code, and require certain linker options to be able to process the binaries correctly, which is what this change is about.

Thanks,

Ben

riverar:
@BenJKuhn That's frustrating, given this impacts inner-loop development for external folks (us) and undermines Project Reunion commitments about working in the open. 😐

DrusTheAxe:
> Are the other DLLs in here (the Project Reunion main one) already building this way?

PLEASE hold off touching the ProjectReunion.sln projects until my PR is checked in (RSN). Merge conflicts from hell we don't need, plus it adds the bootstrapper dll/project which should get the same treatment as the Microsoft.ProjectReunion.dll/project

dev/MRTCore/mrt/Core/src/MRM.vcxproj
rohanp-msft: the /debug switch here doesn't refer to the build configuration
it refers to whether a pdb should be generated
for the question of whether we want the binaries to be ready for apiscan for one or both build configurations, jon thought perhaps we can do just release (so that debug builds retain incrementality) and asked me to check with ben
iteration 2 does that: it preps the binaries for apiscan only for the release configuration
i'll check with ben if that's ok

dev/MRTCore/mrt/Core/src/MRM.vcxproj
```diff
@@ -124,6 +126,10 @@
            than they would otherwise be if the CRT, runtime, and STL were all statically linked in. -->
       <IgnoreSpecificDefaultLibraries>%(IgnoreSpecificDefaultLibraries);libucrt.lib</IgnoreSpecificDefaultLibraries>
       <AdditionalOptions>%(AdditionalOptions) /defaultlib:ucrt.lib</AdditionalOptions>
+      <!-- Configure the release build binary to be as required by internal API scanning tools.
+           Note: /opt:icf and /opt:ref achieve the same result as setting EnableCOMDATFolding and OptimizeReferences to
+           true, which is done below for all build configurations, so we exclude them here. -->
+      <AdditionalOptions>%(AdditionalOptions) /debug /debugtype:cv,fixup /incremental:no</AdditionalOptions>
```
rohanp-msft, line 132: Pending: per John's comment, I'll replace all direct option setting in this change with individual linker/build settings (like we have for `EnableCOMDATFolding` and `OptimizeReferences` currently).

rohanp-msft: addressed

rohanp-msft:
> > Are the other DLLs in here (the Project Reunion main one) already building this way?
> 
> @jonwis I don't know... I tackled just the MRTCore ones. Would offer to do the main one too but I have a blocking Reunion bug on me.

I'd like to keep this MRTCore focused. I can do the other binaries in the repo in a separate change.

huichen123:
we should handle all the projects the same way. Define something like APIScanCompileOptions/APIScanLinkOptions (name TBD) everyone can use, plus make them overridable.

DrusTheAxe:
> > Are the other DLLs in here (the Project Reunion main one) already building this way?
> 
> @jonwis I don't know... I tackled just the MRTCore ones. Would offer to do the main one too but I have a blocking Reunion bug on me.

No. Only MRT uses the UCRT. Everyone *should*, hence Issue [#573 Update ProjectReunion to use UCRT](https://github.com/microsoft/ProjectReunion/issues/573)

BenJKuhn (approved):

PR #530, opened 2021-03-06
address build/restore issues causing vsix pipeline failures

Scottj1s:
Main project file would occasionally fail to restore, VsToolsPath would occasionally not be set, causing VSIX targets to not execute.  Fix was to add fallback, as template project files do.

Template project files would fail to build, requiring 'win' RuntimeIdentifier.  Fix for this was to add it to directory.build.props, and not the template projects, as that breaks template instantation.


BenJKuhn (approved):

Scottj1s:
Thanks!

PR #548, opened 2021-03-10
cherry pick: always use internal feed

Scottj1s:


riverar:
PR title was misleading, seems like you're removing the internal feed (which is great). Is that intentional?

jonwis (approved):

BenJKuhn (approved):

PR #549, opened 2021-03-10
Update ProjectCapability statically for now.

DefaultRyan:
Still working out how to do this in the pipeline in such a way that it "sticks" for consumers. So we'll just update this statically at the moment.

VersionSpecific = ProjectReunion.Foundation.0.1.0-prerelease

Also changing the VersionGeneral tag to the more specific ProjectReunion.Foundation.


BenJKuhn:
Please move this to the release branch, and use 0.5.0. Since we won't be spinning 0.1.0 again, there's no point in putting in telemetry for the old version. In terms of auto-updating, you can edit XML easily using inline powershell scripts in the YAML file. I can point you to examples if you need it.

DefaultRyan:
> Please move this to the release branch, and use 0.5.0. Since we won't be spinning 0.1.0 again, there's no point in putting in telemetry for the old version. In terms of auto-updating, you can edit XML easily using inline powershell scripts in the YAML file. I can point you to examples if you need it.

I'll create a new PR for the release branch, rather than trying to merge it over there. I'm assuming we'll still want to merge this back to main at some point.

For auto-updating, I'll take you up on the offer for samples, but there also needs to be a little more plumbing done in the pipelines, because last I checked, the version 0.1.0 was being set "pretty far down" the stack and needs to be moved up so more consumers (like this task) can get it.

PR #550, opened 2021-03-10
Update ProjectCapability statically for now.

DefaultRyan:
Still plumbing in how to do this in the pipeline in such a way that it "sticks" for consumers. So we'll just update this statically at the moment. 

VersionSpecific = ProjectReunion.Foundation.0.5.0-prerelease

Also changing the VersionGeneral tag to the more specific ProjectReunion.Foundation.


BenJKuhn:
Closing this out since we're investigating how to integrate version data into the capability.

PR #554, opened 2021-03-11
MRTCore will also lookup parent folder

huichen123:
When file path is not provided (default or only filename), MRTCore will also lookup current module path as well as the parent folder for the resource file. This is to accommodate Visual Studio which usually creates EXE in a subfolder with project name unless the project has TargetPlatformIdentifier set to UAP.

### DIFF
--- a/dev/MRTCore/mrt/Core/src/MRM.cpp
+++ b/dev/MRTCore/mrt/Core/src/MRM.cpp
@@ -853,7 +853,12 @@ STDAPI_(void) MrmFreeResource(_In_opt_ void* resource)
     return;
 }
 
-// Append filename to current module path
+// Append filename to current module path. If the file doesn't exist, it will
+// append the filename to parent path. If none exists, file in current module
+// path will be returned.
+// Visual Studio usually builds the executable to a subfolder with project name 
+// unless TargetPlatformIdentifier is UAP. To accommodate this behavior, we will
+// try to search resources.pri from both module path and parent path.
 STDAPI MrmGetFilePathFromName(_In_ PCWSTR filename, _Outptr_ PWSTR* filePath)
 {
     *filePath = nullptr;
@@ -883,10 +888,12 @@ STDAPI MrmGetFilePathFromName(_In_ PCWSTR filename, _Outptr_ PWSTR* filePath)
     // Remove module file name
     PWSTR pointerToPath = pathAllocated ? pathAllocated.get() : path;
     PWSTR lastSlash = nullptr;
+    PWSTR secondToLastSlash = nullptr;
     while (*pointerToPath)
     {
         if (*pointerToPath == L'\\')
         {
+            secondToLastSlash = lastSlash;
             lastSlash = pointerToPath;
         }
         pointerToPath++;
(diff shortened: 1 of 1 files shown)


axelandrejs (approved):

dev/MRTCore/mrt/Core/src/MRM.cpp
```diff
@@ -899,26 +906,53 @@ STDAPI MrmGetFilePathFromName(_In_ PCWSTR filename, _Outptr_ PWSTR* filePath)
...
 
-    RETURN_IF_FAILED(StringCchCopyW(outputPath.get(), length, pointerToPath));
-    RETURN_IF_FAILED(StringCchCatW(outputPath.get(), length, filename));
+        *(secondToLastSlash + 1) = 0;
+    }
 
-    *filePath = outputPath.release();
+    *filePath = finalPath.release();
 
     return S_OK;
```
rohanp-msft, line 957: Could we have the fallback impl be in a separate function that we call if filePath is empty after the existing code runs?

huichen123: not quite sure what you mean. but this has least code churn.

rohanp-msft: The existing code looks in the current module path, right? We're adding looking in the parent folder too. I'm saying having the addition, i.e., looking in the parent folder, be in a separate function. That way if we want to remove it in the future, for example, it is much easier. It also makes the code more self-explanatory.

---
In reply to: [591992825](https://github.com/microsoft/ProjectReunion/pull/554#discussion_r591992825) [](ancestors = 591992825)

dev/MRTCore/mrt/Core/src/MRM.cpp
```diff
@@ -899,26 +906,53 @@ STDAPI MrmGetFilePathFromName(_In_ PCWSTR filename, _Outptr_ PWSTR* filePath)
...
+            // The file exists. Done.
+            finalPath.swap(outputPath);
+            break;
+        }
 
-    PWSTR rawOutputPath = reinterpret_cast<PWSTR>(MrmAllocateBuffer(size));
-    RETURN_IF_NULL_ALLOC(rawOutputPath);
+        if (i == 0)
+        {
+            // If none of the file exists, will return the file in current path
```
rohanp-msft, line 943: >will return the file in current path [](start = 43, length = 36)

But if the file exists in the current path, we won't even get here. We will break above.

dev/MRTCore/mrt/Core/src/MRM.cpp
huichen123: then the second function will be similar to the first one. People will ask why you cannot abstract the logic. Plus, a few more places need to be updated. Similar logic of lookup/check existence/lookup again will also need to apply to multiple places. If you want to prepare for possible future removal, this is probably the easiest.

dev/MRTCore/mrt/Core/src/MRM.cpp
huichen123: this is when the file doesn't exist in either location, we will return the file path in current folder.

rohanp-msft: but if we're here, the file doesn't exist in the current folder, no?

---
In reply to: [591996835](https://github.com/microsoft/ProjectReunion/pull/554#discussion_r591996835) [](ancestors = 591996835)

dev/MRTCore/mrt/Core/src/MRM.cpp
```diff
@@ -899,26 +906,53 @@ STDAPI MrmGetFilePathFromName(_In_ PCWSTR filename, _Outptr_ PWSTR* filePath)
...
 
-    RETURN_IF_FAILED(DWordAdd(length, lengthOfName, &length));
-    RETURN_IF_FAILED(DWordAdd(length, 1, &length));
-    RETURN_IF_FAILED(DWordMult(length, sizeof(wchar_t), &size));
+        std::unique_ptr<wchar_t, decltype(&MrmFreeResource)> outputPath(rawOutputPath, MrmFreeResource);
+
+        RETURN_IF_FAILED(StringCchCopyW(outputPath.get(), length, pointerToPath));
+        RETURN_IF_FAILED(StringCchCatW(outputPath.get(), length, filename));
+
+        if (GetFileAttributes(outputPath.get()) != INVALID_FILE_ATTRIBUTES)
```
riverar, line 934: This will also succeed if a directory, not a file, with the same name exists. Might want to explicitly test for and exclude items with `FILE_ATTRIBUTE_DIRECTORY`.

dev/MRTCore/mrt/Core/src/MRM.cpp
rohanp-msft: ok, this is fine with me

---
In reply to: [591996471](https://github.com/microsoft/ProjectReunion/pull/554#discussion_r591996471) [](ancestors = 591996471)

dev/MRTCore/mrt/Core/src/MRM.cpp
huichen123: correct. this function returns a filename. If the file doesn't exist, it will still return you a name. There is no point to fail.

rohanp-msft (approved):
:shipit:

riverar (commented):
That path manipulation code is a bit scary at glance, would like to see another pass at some point replace it with `std::experimental::filesystem::path`. But sounds like minimal churn is a constraint.

dev/MRTCore/mrt/Core/src/MRM.cpp
huichen123: Yes. There is little time before release. Also this is no throwing code, and this doesn't use std. 
I was debating whether exclude directory or make code simple. The chance you have  a folder called resources.pri is extremely slim. I can make that check if you insist.

riverar: Ha! I'm an external so can't insist anything. Agree the risk is low, we can create an issue to track it and move forward.

PR #555, opened 2021-03-11
User/modanish/power management

huzaifa-d:
Adding the open source implementation of Power Management

### DIFF
--- a/dev/PowerNotifications/PowerNotifications.cpp
+++ b/dev/PowerNotifications/PowerNotifications.cpp
@@ -8,494 +8,233 @@
 
 namespace winrt::Microsoft::ProjectReunion::implementation
 {
-    
-    PowerManager::eventType PowerManager::m_energySaverStatusChangedEvent;
-    PowerManager::eventType PowerManager::m_batteryStatusChangedEvent;
-    PowerManager::eventType PowerManager::m_powerSupplyStatusChangedEvent;
-    PowerManager::eventType PowerManager::m_remainingChargePercentChangedEvent;
-    PowerManager::eventType PowerManager::m_remainingDischargeTimeChangedEvent;
+    auto factory = make_self<factory_implementation::PowerManager>();
 
-    PowerManager::eventType PowerManager::m_powerSourceStatusChangedEvent;
-    PowerManager::eventType PowerManager::m_displayStatusChangedEvent;
-    PowerManager::eventType PowerManager::m_systemIdleStatusChangedEvent;
-    PowerManager::eventType PowerManager::m_powerSchemePersonalityChangedEvent;
-    PowerManager::eventType PowerManager::m_userPresenceStatusChangedEvent;
-    PowerManager::eventType PowerManager::m_systemAwayModeStatusChangedEvent;
-
-    EnergySaverStatusRegistration      PowerManager::m_energySaverStatusHandle;
-    CompositeBatteryStatusRegistration PowerManager::m_batteryStatusHandle;
-    DischargeTimeRegistration          PowerManager::m_dischargeTimeHandle;
-    PowerConditionRegistration         PowerManager::m_powerConditionHandle;
-    DisplayStatusRegistration          PowerManager::m_displayStatusHandle;
-    SystemIdleStatusRegistration       PowerManager::m_systemIdleStatusHandle;
-    PowerSchemePersonalityRegistration PowerManager::m_powerSchemePersonalityHandle;
-    UserPresenceStatusRegistration     PowerManager::m_userPresenceStatusHandle;
-    SystemAwayModeStatusRegistration   PowerManager::m_systemAwayModeStatusHandle;
-
-    std::mutex PowerManager::m_mutex;
-
-    int   PowerManager::m_batteryChargePercent, PowerManager::m_oldBatteryChargePercent;
-    DWORD PowerManager::m_cachedDisplayStatus;
-    DWORD PowerManager::m_cachedUserPresenceStatus;
-    DWORD PowerManager::m_cachedSystemAwayModeStatus;
-    DWORD PowerManager::m_cachedPowerCondition;
-    GUID  PowerManager::m_cachedPowerSchemePersonality;
-    ULONGLONG               PowerManager::m_cachedDischargeTime;
-    CompositeBatteryStatus *PowerManager::m_cachedCompositeBatteryStatus;
-    ::EnergySaverStatus PowerManager::m_cachedEnergySaverStatus;
-    winrt::Microsoft::ProjectReunion::BatteryStatus     PowerManager::m_batteryStatus;
-    winrt::Microsoft::ProjectReunion::BatteryStatus     PowerManager::m_oldBatteryStatus = winrt::Microsoft::ProjectReunion::BatteryStatus::Uninitialized;
-    winrt::Microsoft::ProjectReunion::PowerSupplyStatus PowerManager::m_powerSupplyStatus;
-    winrt::Microsoft::ProjectReunion::PowerSupplyStatus PowerManager::m_oldPowerSupplyStatus = winrt::Microsoft::ProjectReunion::PowerSupplyStatus::Uninitialized;
-
-
-    bool PowerManager::NotAlreadyRegisteredForEvents(PowerManager::eventType eventObj)
+    // EnergySaverStatus Functions
+    EventType& EnergySaverStatus_Event()
     {
-        return !eventObj.get();
+        return factory->m_EnergySaverStatusChangedEvent;
     }
 
-    PowerManager::eventType PowerManager::GetEventObj(PowerFunction const& fn)
+    void EnergySaverStatus_Register()
     {
-        switch (fn) {
-        case EnergySaverStatusFn:      return PowerManager::m_energySaverStatusChangedEvent;
-        case BatteryStatusFn:          return PowerManager::m_batteryStatusChangedEvent;
-        case PowerSupplyStatusFn:      return PowerManager::m_powerSupplyStatusChangedEvent;
-        case RemainingChargePercentFn: return PowerManager::m_remainingChargePercentChangedEvent;
-        case RemainingDischargeTimeFn: return PowerManager::m_remainingDischargeTimeChangedEvent;
-        case PowerSourceStatusFn:      return PowerManager::m_powerSourceStatusChangedEvent;
-        case DisplayStatusFn:          return PowerManager::m_displayStatusChangedEvent;
-        case SystemIdleStatusFn:       return PowerManager::m_systemIdleStatusChangedEvent;
-        case PowerSchemePersonalityFn: return PowerManager::m_powerSchemePersonalityChangedEvent;
-        case UserPresenceStatusFn:     return PowerManager::m_userPresenceStatusChangedEvent;
-        case SystemAwayModeStatusFn:   return PowerManager::m_systemAwayModeStatusChangedEvent;
-        default: throw E_NOTIMPL;
-        }
+        check_hresult(RegisterEnergySaverStatusChangedListener(&PowerManager::EnergySaverStatusChanged_Callback, &factory->m_EnergySaverStatusHandle));
     }
 
-    EventToken PowerManager::AddCallback(PowerFunction const& fn, PowerEventHandle const& handler)
+    void EnergySaverStatus_Unregister()
     {
-        auto eventObj = GetEventObj(fn);
-        std::scoped_lock<std::mutex> lock(m_mutex);
-
-        //Check prior registrations
-        if (NotAlreadyRegisteredForEvents(eventObj))
-        {
-            RegisterListener(fn);
-        }
-        eventObj.get().add(handler);
+        check_hresult(UnregisterEnergySaverStatusChangedListener(factory->m_EnergySaverStatusHandle));
     }
 
-    void PowerManager::RegisterListener(PowerFunction const& fn)
+    void EnergySaverStatus_Update()
     {
-        switch (fn) {        
-            case BatteryStatusFn:          
-            case PowerSupplyStatusFn:      
-            case RemainingChargePercentFn: check_hresult(RegisterCompositeBatteryStatusChangedListener(CompositeBatteryStatusChanged_Callback, &m_batteryStatusHandle)); break;
-            case EnergySaverStatusFn:      check_hresult(RegisterEnergySaverStatusChangedListener(EnergySaverStatusChanged_Callback, &m_energySaverStatusHandle)); break;
-            case RemainingDischargeTimeFn: check_hresult(RegisterDischargeTimeChangedListener(RemainingDischargeTimeChanged_Callback, &m_dischargeTimeHandle));    break;
-            case PowerSourceStatusFn:      check_hresult(RegisterPowerConditionChangedListener(PowerSourceStatusChanged_Callback, &m_powerConditionHandle));       break;
-            case DisplayStatusFn:          check_hresult(RegisterDisplayStatusChangedListener(DisplayStatusChanged_Callback, &m_displayStatusHandle));             break;
-            case SystemIdleStatusFn:       check_hresult(RegisterSystemIdleStatusChangedListener(SystemIdleStatusChanged_Callback, &m_systemIdleStatusHandle));    break;
-            case PowerSchemePersonalityFn: check_hresult(RegisterPowerSchemePersonalityChangedListener(PowerSchemePersonalityChanged_Callback, &m_powerSchemePersonalityHandle)); break;
-            case UserPresenceStatusFn:     check_hresult(RegisterUserPresenceStatusChangedListener(UserPresenceStatusChanged_Callback, &m_userPresenceStatusHandle));             break;
-            case SystemAwayModeStatusFn:   check_hresult(RegisterSystemAwayModeStatusChangedListener(SystemAwayModeStatusChanged_Callback, &m_systemAwayModeStatusHandle));       break;
-            default: throw E_NOTIMPL;
-        }
+        check_hresult(GetEnergySaverStatus(&factory->m_cachedEnergySaverStatus));
     }
 
-    void PowerManager::UnregisterListener(PowerFunction const& fn)
+    // BatteryStatus Functions
+    EventType& BatteryStatus_Event()
     {
-        switch (fn) {        
-            case BatteryStatusFn:          
-            case PowerSupplyStatusFn:      
-            case RemainingChargePercentFn: check_hresult(UnregisterCompositeBatteryStatusChangedListener(m_batteryStatusHandle)); break;
-            case EnergySaverStatusFn:      check_hresult(UnregisterEnergySaverStatusChangedListener(m_energySaverStatusHandle));  break;
-            case RemainingDischargeTimeFn: check_hresult(UnregisterDischargeTimeChangedListener(m_dischargeTimeHandle));          break;
-            case PowerSourceStatusFn:      check_hresult(UnregisterPowerConditionChangedListener(m_powerConditionHandle));        break;
-            case DisplayStatusFn:          check_hresult(UnregisterDisplayStatusChangedListener(m_displayStatusHandle));          break;
-            case SystemIdleStatusFn:       check_hresult(UnregisterSystemIdleStatusChangedListener(m_systemIdleStatusHandle));    break;
-            case PowerSchemePersonalityFn: check_hresult(UnregisterPowerSchemePersonalityChangedListener(m_powerSchemePersonalityHandle)); break;
-            case UserPresenceStatusFn:     check_hresult(UnregisterUserPresenceStatusChangedListener(m_userPresenceStatusHandle));         break;
-            case SystemAwayModeStatusFn:   check_hresult(UnregisterSystemAwayModeStatusChangedListener(m_systemAwayModeStatusHandle));     break;
-            default: throw E_NOTIMPL;
-        }
+        return factory->m_batteryStatusChangedEvent;
     }
 
-    void PowerManager::RemoveCallback(PowerFunction const& fn, EventToken const& token)
+    void BatteryStatus_Register()
     {
-        auto eventObj = GetEventObj(fn);
-        eventObj.get().remove(token);
-        std::scoped_lock<std::mutex> lock(m_mutex);
-        //Check prior registrations
-        if (NotAlreadyRegisteredForEvents(eventObj))
-        {
-            UnregisterListener(fn);
-        }
+        check_hresult(RegisterCompositeBatteryStatusChangedListener(&PowerManager::CompositeBatteryStatusChanged_Callback, &factory->m_batteryStatusHandle));
     }
 
-    void PowerManager::FireEvent(PowerFunction const& fn)
+    void BatteryStatus_Unregister()
     {
-        GetEventObj(fn).get()(nullptr, nullptr);
+        check_hresult(UnregisterCompositeBatteryStatusChangedListener(factory->m_batteryStatusHandle));
     }
 
-    //Checks if an event is already registered. If none are, then gets the status and caches it
-    void PowerManager::CheckRegistrationAndOrUpdateValue(PowerFunction const& fn)
+    void BatteryStatus_Update()
     {
-        auto eventObj = GetEventObj(fn);
-        if (NotAlreadyRegisteredForEvents(eventObj))
-        {
-            switch (fn) {
-                case EnergySaverStatusFn:
-                    check_hresult(GetEnergySaverStatus(&m_cachedEnergySaverStatus));
-                    break;
-
-                case BatteryStatusFn:
-                    check_hresult(GetCompositeBatteryStatus(&m_cachedCompositeBatteryStatus));
-                    ProcessCompositeBatteryStatus(*m_cachedCompositeBatteryStatus);
-                    break;
-
-                case PowerSupplyStatusFn:
-                    break;
-
-                case RemainingChargePercentFn:
-                    break;
-
-                case RemainingDischargeTimeFn:
-                    check_hresult(GetDischargeTime(&m_cachedDischargeTime));
-                    break;
-
-                case PowerSourceStatusFn:
-                    check_hresult(GetPowerCondition(&m_cachedPowerCondition));
-                    break;
-
-                case DisplayStatusFn:
-                    check_hresult(GetDisplayStatus(&m_cachedDisplayStatus));
-                    break;
-
-                case SystemIdleStatusFn:
-                    break;
-
-                case PowerSchemePersonalityFn:
-                    check_hresult(GetPowerSchemePersonality(&m_cachedPowerSchemePersonality));
-                    break;
-
-                case UserPresenceStatusFn:
-                    check_hresult(GetUserPresenceStatus(&m_cachedUserPresenceStatus));
-                    break;
-
-                case SystemAwayModeStatusFn:
-                    check_hresult(GetSystemAwayModeStatus(&m_cachedSystemAwayModeStatus));
-                    break;
-
-                default: throw E_NOTIMPL;
-            }
-        }
-
+        check_hresult(GetCompositeBatteryStatus(&factory->m_cachedCompositeBatteryStatus));
+        factory->ProcessCompositeBatteryStatus(*factory->m_cachedCompositeBatteryStatus);
     }
-
-    EnergySaverStatus PowerManager::EnergySaverStatus()
+   
+    // PowerSupplyStatus Functions
+    EventType& PowerSupplyStatus_Event()
     {
-        CheckRegistrationAndOrUpdateValue(EnergySaverStatusFn);
-        return static_cast<winrt::Microsoft::ProjectReunion::EnergySaverStatus>(m_cachedEnergySaverStatus);
+        return factory->m_powerSupplyStatusChangedEvent;
     }
 
-    EventToken PowerManager::EnergySaverStatusChanged(PowerEventHandle const& handler)
+    void PowerSupplyStatus_Register()
     {
-        return AddCallback(EnergySaverStatusFn, handler);
+        BatteryStatus_Register();
     }
 
-    void PowerManager::EnergySaverStatusChanged(EventToken const& token)
+    void PowerSupplyStatus_Unregister()
     {
-        RemoveCallback(EnergySaverStatusFn, token);
+        BatteryStatus_Unregister();
     }
 
-    void PowerManager::EnergySaverStatusChanged_Callback(::EnergySaverStatus energySaverStatus)
+    void PowerSupplyStatus_Update()
     {
-        m_cachedEnergySaverStatus = energySaverStatus;
-        FireEvent(EnergySaverStatusFn);
+        BatteryStatus_Update();
     }
-
-    void PowerManager::ProcessCompositeBatteryStatus(CompositeBatteryStatus const& compositeBatteryStatus)
+   
+    // RemainingChargePercent Functions
+    EventType& RemainingChargePercent_Event()
     {
-        //Calculate battery charge
-        auto fullChargedCapacity = compositeBatteryStatus.Information.FullChargedCapacity;
-        auto remainingCapacity = compositeBatteryStatus.Status.Capacity;
-        auto newRemainingChargePercent = static_cast<int>((remainingCapacity * 200) / fullChargedCapacity);
-        newRemainingChargePercent += 1;
-        newRemainingChargePercent /= 2;
-        m_batteryChargePercent = newRemainingChargePercent;
-
-        auto powerState = compositeBatteryStatus.Status.PowerState;
-
-        //Set battery status
-        if (compositeBatteryStatus.ActiveBatteryCount == 0)
-        {
-            m_batteryStatus = BatteryStatus::NotPresent;
-        }
-        else if ((powerState & BATTERY_DISCHARGING) != FALSE)
-        {
-            m_batteryStatus = BatteryStatus::Discharging;
-        }
-        else if ((powerState & BATTERY_CHARGING) != FALSE)
-        {
-            m_batteryStatus = BatteryStatus::Charging;
-        }
-        else
-        {
-            m_batteryStatus = BatteryStatus::Idle;
-        }
-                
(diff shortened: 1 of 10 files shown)


dev/PowerNotifications/PowerNotifications.idl
```diff
@@ -0,0 +1,115 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+// REVIEW: We're not doing doing any versioning within IDL/WinMDs, right?
+
+namespace Microsoft.ProjectReunion
```
jonwis, line 6: We've recently settled on some namespacing rules (doc coming soon) so this should be under `Microsoft.Windows.System.Power` (since it's a snowclone of the existing `Windows.System.Power` namespaces.

dev/PowerNotifications/PowerNotifications.idl
```diff
@@ -0,0 +1,115 @@
...
+{
+    // Enums duplicated from IPowerManagerStatics
+    enum EnergySaverStatus
```
jonwis, line 9: Looks like we still need to do a review of the power-state stuff in @andreww-msft 's [branch](https://github.com/microsoft/ProjectReunion/compare/user/andreww-msft#diff-150d556fe5d95f35b9cbada87b6ad1a00bdd78f539634970cc6b38feedcd5a67). I'll prioritize that for the coming week.  This IDL needs some help before we merge it in.  Some initial things here you can get started on...

dev/PowerNotifications/PowerNotifications.idl
```diff
@@ -0,0 +1,115 @@
...
+    {
+        Disabled = 0,
+        Off,
+        On
+    };
+
+    enum BatteryStatus
+    {
+        Uninitialized = -1,
```
jonwis, line 18: Don't define an `Uninitialized` value, and WinRT enums really don't like `-1` as a value.

docs/faq.md
```diff
@@ -122,5 +122,67 @@ user is in control. Windows and views created by the app are positioned by the
 OS shell. Its UX is either drawn using XAML or Composition objects or a
 framework that produces XAML or Composition objects.
 
-## I don't see my question here! [Create an issue to ask a question or start a discussion](https://github.com/microsoft/ProjectReunion/issues/new/choose).
+## How will I consume Project Reunion components? How will they be packaged?
+
```
jonwis, line 126: What're these edits from?

dev/PowerNotifications/PowerNotifications.cpp
```diff
@@ -0,0 +1,501 @@
...
+namespace winrt::Microsoft::ProjectReunion::implementation
+{
+    
+    PowerManager::eventType PowerManager::m_energySaverStatusChangedEvent;
+    PowerManager::eventType PowerManager::m_batteryStatusChangedEvent;
+    PowerManager::eventType PowerManager::m_powerSupplyStatusChangedEvent;
+    PowerManager::eventType PowerManager::m_remainingChargePercentChangedEvent;
+    PowerManager::eventType PowerManager::m_remainingDischargeTimeChangedEvent;
+
+    PowerManager::eventType PowerManager::m_powerSourceStatusChangedEvent;
```
jonwis, line 18: Hey @kennykerr - this PowerManager type is mostly static members. What's the C++/WinRT idiomatic way to implement it so we don't have to declare the static-members and then put their definitions here?

dev/PowerNotifications/PowerNotifications.cpp
```diff
@@ -0,0 +1,501 @@
...
+    PowerManager::eventType PowerManager::m_displayStatusChangedEvent;
+    PowerManager::eventType PowerManager::m_systemIdleStatusChangedEvent;
+    PowerManager::eventType PowerManager::m_powerSchemePersonalityChangedEvent;
+    PowerManager::eventType PowerManager::m_userPresenceStatusChangedEvent;
+    PowerManager::eventType PowerManager::m_systemAwayModeStatusChangedEvent;
+
+    EnergySaverStatusRegistration      PowerManager::m_energySaverStatusHandle;
```
jonwis, line 25: While the CppCoreGuidelines don't take a position on this, please don't use "vertical alignment" - it makes maintaining the codebase much harder over time.  @chrisguzak was looking into adding a .editorrc to get consistency for new code.

dev/PowerNotifications/PowerNotifications.cpp
```diff
@@ -0,0 +1,501 @@
...
+    PowerConditionRegistration         PowerManager::m_powerConditionHandle;
+    DisplayStatusRegistration          PowerManager::m_displayStatusHandle;
+    SystemIdleStatusRegistration       PowerManager::m_systemIdleStatusHandle;
+    PowerSchemePersonalityRegistration PowerManager::m_powerSchemePersonalityHandle;
+    UserPresenceStatusRegistration     PowerManager::m_userPresenceStatusHandle;
+    SystemAwayModeStatusRegistration   PowerManager::m_systemAwayModeStatusHandle;
+
+    std::mutex PowerManager::m_mutex;
+
+    int   PowerManager::m_batteryChargePercent, PowerManager::m_oldBatteryChargePercent;
```
jonwis, line 37: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#es10-declare-one-name-only-per-declaration - split this across two lines.

dev/PowerNotifications/PowerNotifications.cpp
```diff
@@ -0,0 +1,501 @@
...
+    DWORD PowerManager::m_cachedDisplayStatus;
+    DWORD PowerManager::m_cachedUserPresenceStatus;
+    DWORD PowerManager::m_cachedSystemAwayModeStatus;
+    DWORD PowerManager::m_cachedPowerCondition;
+    GUID  PowerManager::m_cachedPowerSchemePersonality;
+    ULONGLONG               PowerManager::m_cachedDischargeTime;
+    CompositeBatteryStatus *PowerManager::m_cachedCompositeBatteryStatus;
+    ::EnergySaverStatus PowerManager::m_cachedEnergySaverStatus;
+    winrt::Microsoft::ProjectReunion::BatteryStatus     PowerManager::m_batteryStatus;
+    winrt::Microsoft::ProjectReunion::BatteryStatus     PowerManager::m_oldBatteryStatus = winrt::Microsoft::ProjectReunion::BatteryStatus::Uninitialized;
```
jonwis, line 47: Use `{ ... }` style braced-init.

dev/PowerNotifications/PowerNotifications.cpp
```diff
@@ -0,0 +1,501 @@
...
+        case PowerSupplyStatusFn:      return PowerManager::m_powerSupplyStatusChangedEvent;
+        case RemainingChargePercentFn: return PowerManager::m_remainingChargePercentChangedEvent;
+        case RemainingDischargeTimeFn: return PowerManager::m_remainingDischargeTimeChangedEvent;
+        case PowerSourceStatusFn:      return PowerManager::m_powerSourceStatusChangedEvent;
+        case DisplayStatusFn:          return PowerManager::m_displayStatusChangedEvent;
+        case SystemIdleStatusFn:       return PowerManager::m_systemIdleStatusChangedEvent;
+        case PowerSchemePersonalityFn: return PowerManager::m_powerSchemePersonalityChangedEvent;
+        case UserPresenceStatusFn:     return PowerManager::m_userPresenceStatusChangedEvent;
+        case SystemAwayModeStatusFn:   return PowerManager::m_systemAwayModeStatusChangedEvent;
+        default: throw E_NOTIMPL;
```
jonwis, line 71: Don't throw raw HRESULTs. Use `throw winrt::hresult_not_implemented();` instead. Applies everywhere.

jonwis: (This is also developer/implementer error, so feel free to use WIL's FAIL_FAST_* family of macros.)

dev/PowerNotifications/PowerNotifications.cpp
```diff
@@ -0,0 +1,501 @@
...
+        }
+    }
+
+    EventToken PowerManager::AddCallback(PowerFunction const& fn, PowerEventHandle const& handler)
```
jonwis, line 75: Does this compile? You probably need to `return` something.

dev/PowerNotifications/PowerNotifications.cpp
```diff
@@ -0,0 +1,501 @@
...
+        }
+        eventObj.get().add(handler);
+    }
+
+    void PowerManager::RegisterListener(PowerFunction const& fn)
+    {
+        switch (fn) {        
+            case BatteryStatusFn:          
+            case PowerSupplyStatusFn:      
+            case RemainingChargePercentFn: check_hresult(RegisterCompositeBatteryStatusChangedListener(CompositeBatteryStatusChanged_Callback, &m_batteryStatusHandle)); break;
```
jonwis, line 93: Style note - split this across lines; this line is too long:

```c++
case RemainingChangePercentFn:
    check_hresult(...);
    break;
```

dev/PowerNotifications/PowerNotifications.cpp
```diff
@@ -0,0 +1,501 @@
...
+        if (NotAlreadyRegisteredForEvents(eventObj))
+        {
+            RegisterListener(fn);
+        }
+        eventObj.get().add(handler);
+    }
+
+    void PowerManager::RegisterListener(PowerFunction const& fn)
+    {
+        switch (fn) {
```
jonwis, line 90: Inconsistent indentation - in some places your "case" is indented, in others it's not.

dev/PowerNotifications/PowerNotifications.cpp
```diff
@@ -0,0 +1,501 @@
...
+        }
+    }
+
+    void PowerManager::RemoveCallback(PowerFunction const& fn, EventToken const& token)
+    {
+        auto eventObj = GetEventObj(fn);
+        eventObj.get().remove(token);
+        std::scoped_lock<std::mutex> lock(m_mutex);
+        //Check prior registrations
+        if (NotAlreadyRegisteredForEvents(eventObj))
```
jonwis, line 130: This condition seems backwards? If the event has already been registered, then we should unregister.

dev/PowerNotifications/PowerNotifications.cpp
```diff
@@ -0,0 +1,501 @@
...
+        m_cachedEnergySaverStatus = energySaverStatus;
+        FireEvent(EnergySaverStatusFn);
+    }
+
+    void PowerManager::ProcessCompositeBatteryStatus(CompositeBatteryStatus const& compositeBatteryStatus)
+    {
+        //Calculate battery charge
+        auto fullChargedCapacity = compositeBatteryStatus.Information.FullChargedCapacity;
+        auto remainingCapacity = compositeBatteryStatus.Status.Capacity;
+        auto newRemainingChargePercent = static_cast<int>((remainingCapacity * 200) / fullChargedCapacity);
```
jonwis, line 223: This math needs a little comment as to what it's doing and why.

dev/PowerNotifications/PowerNotifications.cpp
```diff
@@ -0,0 +1,501 @@
...
+        m_batteryChargePercent = newRemainingChargePercent;
+
+        auto powerState = compositeBatteryStatus.Status.PowerState;
+
+        //Set battery status
+        if (compositeBatteryStatus.ActiveBatteryCount == 0)
+        {
+            m_batteryStatus = BatteryStatus::NotPresent;
+        }
+        else if ((powerState & BATTERY_DISCHARGING) != FALSE)
```
jonwis, line 235: Try using the WIL helpers `WI_IsFlagSet` / `WI_IsFlagClear` here; I'm having problems parsing the booleans.

dev/PowerNotifications/PowerNotifications.idl
```diff
@@ -0,0 +1,115 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+// REVIEW: We're not doing doing any versioning within IDL/WinMDs, right?
```
jonwis, line 4: We are, and @MikeHillberg / @jevansaks  are building guidance for it. For now, you can leave this as is and we'll apply versioning markup prior to release.

dev/PowerNotifications/PowerNotificationsPal.h
```diff
@@ -0,0 +1,78 @@
...
+{
+    BATTERY_INFORMATION Information;
+    BATTERY_STATUS Status;
+    ULONG ActiveBatteryCount;
+};
+
+DECLARE_HANDLE(CompositeBatteryStatusRegistration);
+HRESULT GetCompositeBatteryStatus(CompositeBatteryStatus** compositeBatteryStatusOut);
+typedef void (*OnCompositeBatteryStatusChanged)(CompositeBatteryStatus* compositeBatteryStatus);
+HRESULT RegisterCompositeBatteryStatusChangedListener(OnCompositeBatteryStatusChanged listener, CompositeBatteryStatusRegistration* registration);
```
jonwis, line 18: Where are these methods defined or implemented?  My guess is this is trying out the closed-open boundary system that @EHO-makai and @BenJKuhn are working on.  We should probably discuss how the headers for those components are generated and transferred.

jonwis (changes requested):
This PR is a good start but there's definitely things to address.  Was the intent to make a pull into the main Reunion repo, or into your fork's "main" instead?

dev/PowerNotifications/PowerNotifications.cpp
kennykerr: @jonwis those definitions should not be necessary. Note that static events should be implemented on the factory object with the static_lifetime marker. [Here's an example](https://github.com/microsoft/cppwinrt/blob/198a8270d16cead25edec4a63d8792b267539fd5/test/old_tests/Component/Static.h). I thought I had an article explaining this but I can't find it - @oldnewthing did you ever write an article about static events and factories?

kennykerr: Basically, you need to pin the factory to ensure that its lifetime is stable as event handlers are added and removed. That's what `static_lifetime` is for. And you can do it all in a header.

kennykerr: I see we have docs for `static_lifetime` - perhaps [this doc](https://docs.microsoft.com/en-us/uwp/cpp-ref-for-winrt/static-lifetime) just needs to be updated to show an example of a static event [like this](https://github.com/microsoft/cppwinrt/blob/198a8270d16cead25edec4a63d8792b267539fd5/test/old_tests/Component/Static.h). @stevewhims is that something you can do?

dev/PowerNotifications/PowerNotifications.h
```diff
@@ -0,0 +1,158 @@
...
+
+        static winrt::Microsoft::ProjectReunion::SystemIdleStatus SystemIdleStatus();
+        static EventToken SystemIdleStatusChanged(PowerEventHandle const&);
+        static void SystemIdleStatusChanged(EventToken const&);
+        static void SystemIdleStatusChanged_Callback();
+
+        static winrt::Microsoft::ProjectReunion::PowerSchemePersonality PowerSchemePersonality();
+        static EventToken PowerSchemePersonalityChanged(PowerEventHandle const&);
+        static void PowerSchemePersonalityChanged(EventToken const&);
+        static void PowerSchemePersonalityChanged_Callback(GUID);
```
hulumane, line 71: Do we know if there are any events here to differentiate state changes from Disconnected Standby and Connected Standby? That would be useful for scenarios like socket rundown for persistent connection scenarios. Today we kind of struggle with that and completely rely on NLM to tell us that we were disconnected when going into low power.

dev/PowerNotifications/PowerNotifications.cpp
```diff
@@ -0,0 +1,501 @@
...
+
+    void PowerManager::EnergySaverStatusChanged_Callback(::EnergySaverStatus energySaverStatus)
+    {
+        m_cachedEnergySaverStatus = energySaverStatus;
+        FireEvent(EnergySaverStatusFn);
+    }
+
+    void PowerManager::ProcessCompositeBatteryStatus(CompositeBatteryStatus const& compositeBatteryStatus)
+    {
+        //Calculate battery charge
```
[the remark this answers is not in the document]
jonwis: @chrisguzak offered to produce a `.clang-format` file for us.  Sadly the CppCoreGuidelines are mum on the subject of whitespacing, other than "pick a consistent style."  I agree that space-after `//` is consistent with the rest of the current codebase.

dev/PowerNotifications/PowerNotifications.idl
theShaan: What would be a better approach then?

dev/PowerNotifications/PowerNotifications.cpp
theShaan: I'm curious : how does it make it harder? It's just to make it look better

dev/PowerNotifications/PowerNotifications.idl
jonwis: Just remove this line entirely and set the first (`NotPresent`) value to `0` and use that as the "unspecified" value.

dev/PowerNotifications/PowerNotifications.cpp
theShaan: I clearly missed out adding the return in the last line, but it surprisingly does compile.

dev/PowerNotifications/PowerNotifications.cpp
jonwis: The problem is maintenance.  When you add a new member here of a type whose name is _longer_ you have to go shift the layout here again, which makes reviewing code changes harder.  Please eliminate these alignment spaces.

dev/PowerNotifications/PowerNotificationsPal.h
theShaan: Yes, that is correct but as far as I know @brialmsft is working on it.

dev/PowerNotifications/PowerNotifications.h
theShaan: @brialmsft Are we supporting these events?

dev/PowerNotifications/PowerNotificationsPal.h
brialmsft: This is effectively a rough draft of the API surface; this header will ultimately be ingested from the closed source NuGet package. We're working on the open-source and closed-source pieces in parallel, since the open-source piece is independently testable with mocks.
With that said, I'm not sure it's appropriate to complete a pull into main while the underlying closed-source code is stubbed out.

dev/PowerNotifications/PowerNotifications.h
brialmsft: I don't believe any of the events we're planning to implement currently cover that, but we're not experts in the power space. If you think there's good reason to include those events, it may be possible to include them in future versions.

dev/PowerNotifications/PowerNotifications.h
```diff
@@ -0,0 +1,158 @@
...
+        int   oldBatteryChargePercent, batteryChargePercent;
+        winrt::Microsoft::ProjectReunion::BatteryStatus     oldBatteryStatus, batteryStatus;
+        winrt::Microsoft::ProjectReunion::PowerSupplyStatus oldPowerSupplyStatus, powerSupplyStatus;
+    };
+
+    struct PowerManager
+    {
+        PowerManager() = default;
+
+        using eventType = wil::object_without_destructor_on_shutdown<winrt::event<PowerEventHandle>>;
```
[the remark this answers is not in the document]
theShaan: Should we just use a normal event?
@brialmsft any suggestions?

dev/PowerNotifications/PowerNotifications.cpp
riverar: Non-exhaustive list of things to consider:
* the code is shared, but your understanding of what's prettier is subjective, so there's a mismatch right out of the gate here
* it increases the work needed to modify this code; folks have to now make a decision on every edit--change my variable names so the length matches, screw up the alignment and move on, or spend the extra few minutes realigning everything so it looks prettier.
* potentially impacts readability in several ways, such as requiring additional horizontal eyeball travel time, fighting the brain that naturally wants to treat the two  terms as unrelated, creating longer horizontal scrollbars when zoomed in, etc.

dev/PowerNotifications/PowerNotifications.h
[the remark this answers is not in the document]
brialmsft: Yeah, this carried over from my first draft of the C++/WinRT implementation. I know how to set this up properly in WRL, but all of the documentation for C++/WinRT I could find didn't point toward a robust way to implement static events. Should have followed up instead of leaving it as-is.

If we don't leak, you still have COM unload issues (which I'm all too familiar with...): this DLL unloads, the event helper objects destruct, and they call Release on the registered delegates which were implemented in other potentially-already-unloaded DLLs (most notably dangerous in ExitProcess and final CoUninitialize scenarios). I think leaking is *less* risky than the naive option of letting the helper objects destruct: if we leak them, then we never signal the events, so nobody should be making unsafe calls into the delegates via our leaked pointers. Which is why I went with leak-on-unload at first.

With that said, the information you and @kennykerr provided should help us do this the right way, where we have a persistent factory object to hang semantically static state off of, and don't need to use the DLL's global/static state.

dev/PowerNotifications/PowerNotifications.cpp
stevewhims: Yes, I've added a second example to the topic: https://docs.microsoft.com/uwp/cpp-ref-for-winrt/static-lifetime.

kennykerr: Thanks!

dev/PowerNotifications/PowerNotifications.h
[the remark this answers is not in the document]
oldnewthing: The point of the static lifetime store is that the objects there are run down *before* COM starts tearing down DLLs. So when the event starts releasing handlers, all the handlers are still good (DLL unloading hasn't started yet).

dev/PowerNotifications/PowerNotifications.idl
oldnewthing: You can still privately use `-1` to mean Uninitialized, but it doesn't need to be part of the public API surface.

dev/PowerNotifications/PowerNotifications.cpp
oldnewthing: Allow me to speak from experience. I became the owner of a rather large file, 7000 LOC. The previous owner enjoyed lining things up vertically in this cute way. Whenever I make a change to this file, Visual Studio reformats it and destroys the vertical alignment, and my pull request says that I touched nearly half of the lines in the file. I have to go back and carefully record the lines that I intended to change, revert the file, check it back out, then carefully reapply the changes **via Notepad**, because I trust only Notepad to be able to modify a file without secretly reformatting it.

I do not enjoy working on that file.

jonwis: (GH needs the "caring" reaction.)

riverar: @oldnewthing I can't reproduce that behavior, but I generally agree with the story's message here.

dev/PowerNotifications/PowerNotifications.idl
hulumane, line 9: Is this the same as Battery Saver or something else?

dev/PowerNotifications/PowerNotifications.idl
```diff
@@ -0,0 +1,115 @@
...
+        Present = 0,
+        Absent
+    };
+
+    enum SystemAwayModeStatus
+    {
+        Entering = 0,
+        Exiting
+    };
+
```
hulumane, line 72: I was looking at power states here: https://docs.microsoft.com/en-us/windows/win32/api/winbase/ns-winbase-system_power_status
I noticed that the battery charge status is missing in these enums. Should that be added too?

dev/PowerNotifications/PowerNotifications.h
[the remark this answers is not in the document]
brialmsft: Right. I was talking specifically about the naive implementation without static lifetime, where we have no guarantee of singleton-instancing on the factory object, and thus have no place to store WinRT-static state other than the DLL's static/global storage. That's the scenario that has COM unload concerns if you don't leak the event helper objects.

I agree that using static lifetime to cache the factory and provide singleton instancing is much cleaner.

dev/PowerNotifications/PowerNotifications.h
hulumane: I know BASE was trying to solve a very similar problem for one of their APIs recently. CS is technically simply LowPower and the platform exposes APIs for it. Infact the BatterySaver event fires when the device goes into this state. DS was a little weird: We have to use a Network Disconnected event to figure this one out in addition to the lower power state changes. It would be awesome to combine the 2 and have some predictable way in the future to inform app developers.

dev/PowerNotifications/PowerNotifications.idl
huzaifa-d: Working on this one. Leading to compiler errors in the auto-generated files.

docs/faq.md
huzaifa-d: I am not sure, but will be moving to a different PR ultimately, so this shouldn't be an issue

dev/PowerNotifications/PowerNotifications.cpp
```diff
@@ -0,0 +1,501 @@
...
+        auto fullChargedCapacity = compositeBatteryStatus.Information.FullChargedCapacity;
```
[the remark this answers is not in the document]
huzaifa-d: Handled

dev/PowerNotifications/PowerNotifications.h
huzaifa-d: We can aim this for a future release as suggested.

dev/PowerNotifications/PowerNotifications.cpp
```diff
@@ -0,0 +1,501 @@
...
+
+    bool PowerManager::NotAlreadyRegisteredForEvents(PowerManager::eventType eventObj)
+    {
+        return !eventObj.get();
+    }
+
+    PowerManager::eventType PowerManager::GetEventObj(PowerFunction const& fn)
+    {
+        switch (fn) {
+        case EnergySaverStatusFn:      return PowerManager::m_energySaverStatusChangedEvent;
```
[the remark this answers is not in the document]
huzaifa-d: Refactored everything

dev/PowerNotifications/PowerNotifications.idl
huzaifa-d: No, this is EnergySaverStatus as mentioned in https://docs.microsoft.com/en-us/uwp/api/Windows.System.Power.PowerManager?view=winrt-19041

dev/PowerNotifications/PowerNotifications.idl
huzaifa-d: We are only supporting the ones noted in the IDL according to the [spec](https://microsoft.sharepoint.com/:w:/r/teams/specstore/_layouts/15/doc2.aspx?sourcedoc=%7BAEB926B3-B313-42DC-AFF6-0EF55394C24F%7D&file=API%20Spec%20-%20AppLifecycle%20Power%20Notifications.docx&action=default&mobileredirect=true&cid=ed1d6af1-7a2c-4b89-b30b-6a9913417558)

huzaifa-d:
Will be moving to a topic branch, but have pushed changes here in the meantime since we already had useful comments.

dev/PowerNotifications/PowerNotifications.h
```diff
@@ -0,0 +1,709 @@
...
+{
+
+     struct PowerManager
+     {
+        PowerManager() = default;
+
+        //Get function forwards
+        static winrt::Microsoft::ProjectReunion::EnergySaverStatus EnergySaverStatus()
+        {
+            return make_self<factory_implementation::PowerManager>()->EnergySaverStatus();
```
jonwis, line 610: This doesn't seem right - every method call here creates a new instance of PowerManager, calls one method, then discards it.

dev/PowerNotifications/PowerNotifications.h
```diff
@@ -0,0 +1,709 @@
...
+        int m_batteryChargePercent;
+        int m_oldBatteryChargePercent;
+        DWORD m_cachedDisplayStatus;
+        DWORD m_cachedUserPresenceStatus;
+        DWORD m_cachedSystemAwayModeStatus;
+        DWORD m_cachedPowerSourceStatus;
+        GUID  m_cachedPowerSchemePersonality;
+        ULONGLONG m_cachedDischargeTime;
+        ::EnergySaverStatus m_cachedEnergySaverStatus;
+        CompositeBatteryStatus* m_cachedCompositeBatteryStatus;
```
oldnewthing, line 103: Who cleans up this pointer at destruction?

dev/PowerNotifications/PowerNotifications.h
```diff
@@ -0,0 +1,709 @@
...
+            {
+                m_powerSupplyStatus = PowerSupplyStatus::Inadequate;
+            }
+            else
+            {
+                m_powerSupplyStatus = PowerSupplyStatus::Adequate;
+            }
+        }
+
+        void FireCorrespondingCompositeBatteryEvent()
```
oldnewthing, line 334: There is no locking here, so bad things happen if two callbacks occur at the same time (say because the batter status and energy saver status both changed at the same time).

jonwis (changes requested):
The static methods in the powermanager type shouldn't be creating fresh instances for each method. I'll see if I can push a sample of what I mean to a new branch you can look at.

dev/PowerNotifications/PowerNotifications.cpp
```diff
@@ -0,0 +1,240 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#include <pch.h>
+#include <PowerNotifications.h>
+#include <PowerNotificationsPal.h>
+#include <PowerManager.g.cpp>
+
+namespace winrt::Microsoft::ProjectReunion::implementation
+{
+    auto factory = make_self<factory_implementation::PowerManager>();
```
brialmsft, line 11: Making the factory as part of a static init could have issues. This make_self would run under the loader lock as part of loading the DLL. Some of the stuff under make_factory, like activating CoreApplication to cache the factory in the static lifetime store (and potentially initializing COM via CoIncrementMTAUsage if it's not already initialized) may not get along well with the loader lock.

dev/PowerNotifications/PowerNotifications.idl
```diff
@@ -0,0 +1,113 @@
...
+    };
+    
+    enum DisplayStatus
+    {
+        Off = 0,
+        On,
+        Dimmed
+    };
+    
+    enum SystemIdleStatus
```
brialmsft, line 46: As we recently discussed, SystemIdleStatus isn't persistently available, we just get a dataless event pulse when the system is non-idle. So this will unfortunately need to break from the existing pattern, and can't be queried.

If you'd prefer to make this change after the initial PR, that should be fine. Just want to call out that adjustments will need to be made.

dev/PowerNotifications/PowerNotifications.cpp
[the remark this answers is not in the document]
kennykerr: Yes, there's no need to store this in a static and as others have noted its very bad to do so.

dev/PowerNotifications/PowerNotifications.h
```diff
@@ -0,0 +1,709 @@
...
+        UserPresenceStatusRegistration m_userPresenceStatusHandle;
+        SystemAwayModeStatusRegistration m_systemAwayModeStatusHandle;
+
+        PowerFunctionDetails energySaverStatusFunc{
+            &winrt::Microsoft::ProjectReunion::implementation::EnergySaverStatus_Event,
+            &winrt::Microsoft::ProjectReunion::implementation::EnergySaverStatus_Register,
+            &winrt::Microsoft::ProjectReunion::implementation::EnergySaverStatus_Unregister,
+            &winrt::Microsoft::ProjectReunion::implementation::EnergySaverStatus_Update };
+
+        PowerFunctionDetails compostieBatteryStatusFunc{
```
brialmsft, line 137: "compostie

dev/PowerNotifications/PowerNotifications.h
oldnewthing: It works because `factory_implementation::PowerManager` is marked `static_lifetime`, so each `make_self` just returns the same instance back.

oldnewthing (commented):
While writing up the PR, I realized that this class is a mess because it's six different classes all glued together, so the member names are all unwieldy so they don't step on each other, the knowledge of each of the six "mini-classes" is scattered all over the place. I'm refactoring it and it's a lot cleaner now. PR coming.

dev/PowerNotifications/PowerNotifications.cpp
```diff
@@ -0,0 +1,240 @@
...
+   
+    // PowerSupplyStatus Functions
+    EventType& PowerSupplyStatus_Event()
+    {
+        return factory->m_powerSupplyStatusChangedEvent;
+    }
+
+    void PowerSupplyStatus_Register()
+    {
+        BatteryStatus_Register();
```
oldnewthing, line 64: This means that if an app registers both for battery status and power supply status, we double-register for composite battery changes (and leak a registration). And then when they unregister from one, we unregister the composite battery change even though the other event is still registered by the app. And then when the app does the second unregister, we unregister an invalid handle.

Multiple events fired from a single registration makes this model a little trickier to manage since the relationship between events and registrations is not 1-1. (The original code had this same problem.)

BenJKuhn:
@huzaifa-d, this PR hasn't been changed in some time. Can you either drive it to completion or close it out?

huzaifa-d:
@oldnewthing you had mentioned "PR coming" in your last comment. Did you have a chance?

huzaifa-d:
Closing PR as it will be tracked on an updated one.

PR #577, opened 2021-03-12
API design for power state, activation, and instancing

andreww-msft:
Merge AppLifecycle API specs to main.


aeloros (approved):
Looks good :)

riverar:
Nit: Can we get a better PR title next time? 😋

jonwis:
We should run this through the API review group; I'll set up time for Tuesday 3/16 1-2pm Seattle time. And if I can figure it out, I'll drop a link for folks to join a Teams call.

specs/AppLifecycle/Activation/AppLifecycle Activation.md
```diff
@@ -0,0 +1,434 @@
...
...IActivatedEventArgs](https://docs.microsoft.com/en-us/uwp/api/Windows.ApplicationModel.Activation.IActivatedEventArgs), which exposes a Kind property. Apps can get an IActivatedEventArgs reference, examine the Kind and then cast the reference to the specific concrete class type that implements IActivatedEventArgs. This enables generic functions such as GetActivatedEventArgs and the Activated event handler, which provide the app with an IActivatedEventArgs that the app can then convert to a specific concrete type. However, IActivatedEventArgs also defines other properties that are only useful in a UWP context. In addition, we need a mechanism that supports both platform ActivationKinds and also Reunion-only ExtendedActivationKinds and their corresponding concrete class types. <br />The new ActivationArguments type serves this purpose: the app can examine the Kind, and then cast the Data property to the specific concrete type (which can be either a platform IActivatedEventArgs concrete class, or a Reunion class).<br />**Kind** is the enum value that indicates the kind of activation this object represents.<br />**Data** can be cast to a specific concrete class corresponding to Kind. |
```
hulumane, line 340: So if we need to add a new PushActivatedKind which isn't part of Windows.Foundation today, we should be able to technically throw this into the IInspectable ref over here correct and add a new ActivationKindExtension?
@danielayala94  FYI

specs/AppLifecycle/Activation/AppLifecycle Activation.md
```diff
@@ -0,0 +1,434 @@
...
...nion is maintaining of all running instances. <br />**RedirectTo** enables an instance of the app to redirect the current activation request to another instance. This is very similar to the existing platform [AppInstance.RedirectActivationTo](https://docs.microsoft.com/en-us/uwp/api/windows.applicationmodel.appinstance.redirectactivationto) method, except that the Reunion implementation allows the app to pass an ActivationArguments payload, thus opening the scope to allow the app to modify or replace the activation arguments that the target instance will receive.<br />**GetActivatedEventArgs** provides similar behavior as the existing platform [AppInstance.GetActivatedEventArgs](https://docs.microsoft.com/en-us/uwp/api/windows.applicationmodel.appinstance.getactivatedeventargs), except that it returns a new ActivationArguments object for the current activation instead of an IActivatedEventArgs. <br />**Key** an app-defined string value that can be used to identify an instance for redirection purposes.<br />**IsCurrent** indicates whether this instance object represents the current instance of the app or a different instance.<br />**ProcessId** the process ID of the instance.<br/> |
```
hulumane, line 341: What is the ActivatedEvent specifically for? I think its missing a description in the second column of the table

specs/AppLifecycle/Activation/AppLifecycle Activation.md
aeloros: @hulumane Take a look at my instancing topic branch for a better idea of where the implementation is going at least.  It should match this, but has some examples.

ChrisGuzak (commented):
"protocol" is a misnomer that has persisted based on its historic use. "uri scheme" is the correct terminology. it is hard to resist "protocol" given how widely it is used in our documentation and APIs, but I think this should be called out at some point "by protocol we mean uri scheme".

specs/AppLifecycle/Activation/AppLifecycle Activation.md
```diff
@@ -0,0 +1,506 @@
...
+}
+```
+
+**RegisterForFileTypeActivation** registers the app for activation based on file types (extensions.)
+
+-   `supportedFileTypes` - one or more supported filetypes, specified by the file extension
+    including the leading ".", eg ".docx"
+-   `logo` - used by the Shell and Settings for this filetype. For packaged apps, a package-relative
+    path to an image file. For unpackaged, a literal filepath to a binary file (DLL, EXE) plus a
+    resource index. The caller can pass either.
```
oldnewthing, line 465: Or resource ID (as a negative number)?

What about absolute path to image file?

(Repeat for other logo parameters.)

specs/AppLifecycle/Activation/AppLifecycle Activation.md
```diff
@@ -0,0 +1,506 @@
...
+-   `displayName` - used by the Shell and Settings for this filetype.
+-   `supportedVerbs` - zero or more app-defined verbs. Each verb is added to the File Explorer
+    context menu when a registered file is right-clicked; and the selected verb is passed to the app
+    as the `FileActivatedEventArgs::Verb` property.
```
oldnewthing, line 469: What's the localization story? Can I pass a resource indirect string for automatic localization based on user preferred language?

For Win32 we generally distinguish between the display name of the verb (localized, shown to the user) and the programmatic name of the verb (not localized, passed to the app). This avoids apps having to do locale-sensitive parsing code when parsing their verbs. Particularly troublesome when words collide between languages.

```cpp
auto verb = args.Verb();
if (verb == L"spend" /* English */ || verb == L"bezahl" /* German */)
{
   DoSpending(); // German "spend" accidentally ends up doing this instead of DoDonate
}
else if (verb == L"donate" /* English */ || verb == L"spend" /* German */)
{
  DoDonate();
}
```

specs/AppLifecycle/Activation/AppLifecycle Activation.md
```diff
@@ -0,0 +1,506 @@
...
+```
+
+**RegisterForFileTypeActivation** registers the app for activation based on file types (extensions.)
+
+-   `supportedFileTypes` - one or more supported filetypes, specified by the file extension
+    including the leading ".", eg ".docx"
+-   `logo` - used by the Shell and Settings for this filetype. For packaged apps, a package-relative
+    path to an image file. For unpackaged, a literal filepath to a binary file (DLL, EXE) plus a
+    resource index. The caller can pass either.
+-   `displayName` - used by the Shell and Settings for this filetype.
```
oldnewthing, line 466: Can I use a resource indirect string here for automatic localization based on user preferred language?

(Repeat for other displayName parameters.)

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
```diff
@@ -0,0 +1,422 @@
...
+-   **Reunion**: Any instance can be a redirection target, whether or not it has a registered key.
+
+### Post-Redirection Behavior
+
+Redirection is a terminal operation. After calling the method to redirect, there is nothing the app
+can do - regardless of whether the redirect succeeded or failed. In calling redirect, the app is
+saying that it cannot/will not handle this activation request - so even if the redirection fails,
+there's nothing this instance can or will do anyway. The platform will explicitly terminate an
+instance that has chosen to redirect. If redirection fails, the activation request fails. The UX is
+the same as an activation failure today.
```
oldnewthing, line 73: Can the redirected-to app perform another redirect to pass the request further? Or is it required to handle it directly?

E.g., "The previous instance redirected to me because I had an existing Contoso document open, but due to a race condition, by the time I received the activation, I'm no longer viewing a Contoso document, but I know someone else who is, so I want to re-redirect to them."

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
```diff
@@ -0,0 +1,422 @@
...
+    _In_ LPWSTR lpCmdLine, _In_ int nCmdShow)
+{
+    // First, we'll get our rich activation event args.
+    ActivationArguments activationArgs =
+        AppInstance::GetCurrent().GetActivatedEventArgs();
+
+    // An app might want to set itself up for possible redirection in
+    // the case where it opens files - for example, to prevent multiple
+    // instances from working on the same file.
+    ExtendedActivationKind kind = activationArgs.Kind;
```
oldnewthing, line 125:
```suggestion
    ExtendedActivationKind kind = activationArgs.Kind();
```

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
```diff
@@ -0,0 +1,422 @@
...
+    if (kind == ExtendedActivationKind::File)
+    {
+        auto fileArgs = activationArgs.Data.as<FileActivatedEventArgs>();
+        IStorageItem file = fileArgs.Files().GetAt(0);
+
+        // Let's try to register this instance for this file.
+        AppInstance instance =
+            AppInstance::FindOrRegisterForKey(file.Name());
+        if (instance.IsCurrent)
```
oldnewthing, line 134:
```suggestion
        if (instance.IsCurrent())
```

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
```diff
@@ -0,0 +1,422 @@
...
+        {
+            // If we successfully registered this instance, we can now just
+            // go ahead and do normal initialization.
+            RegisterClassAndStartMessagePump(hInstance, nCmdShow);
+        }
+        else
+        {
+            // Some other instance has already registered for this file,
+            // so we'll redirect this activation to that instance instead.
+            instance.RedirectTo(activationArgs);
```
oldnewthing, line 144: How is the terminal-ness of RedirectTo manifested? Does the function never return? Would this bypass destructors?

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
```diff
@@ -0,0 +1,422 @@
...
+    }
+    else
+    {
+        // For other activation kinds, we'll trawl all instances to see if
+        // any are suitable for redirecting this request. First, get a list
+        // of all running instances of this app.
+        auto instances = AppInstance::GetInstances();
+
+        // In the simple case, we'll redirect to any other instance.
+        AppInstance instance = instances.GetAt(0);
```
oldnewthing, line 192: I assume GetInstances() always returns at least 1 instance (the currently running instance).

Is there the possibility that GetAt(0) returns the currently running instance? It won't be an "other" instance.

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
```diff
@@ -0,0 +1,422 @@
...
+
+        // If the app re-registers re-usable instances, we can filter for these instead.
+        bool isFound = false;
+        for (AppInstance instance : instances)
+        {
+            if (instance.Key.c_str() == "REUSABLE")
```
oldnewthing, line 198: Code as written doesn't compile and even after adding necessary syntax, it would always be false because the `c_str()` will never match the embedded literal.

```suggestion
            if (instance.Key() == L"REUSABLE")
```

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
```diff
@@ -0,0 +1,422 @@
...
+            {
+                isFound = true;
+                instance.RedirectTo(activationArgs);
+                break;
```
oldnewthing, line 202: This code seems to think that RedirectTo can return. Can it?

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
```diff
@@ -0,0 +1,422 @@
...
+            }
+        }
+        if (!isFound)
+        {
+            // We'll register this as a reusable instance, and then
+            // go ahead and do normal initialization.
+            winrt::hstring szKey = L"REUSABLE";
+            AppInstance::FindOrRegisterForKey(szKey);
```
oldnewthing, line 210: Presumably if another app raced and registered first, then this FindOrRegister is effectively a nop.

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
```diff
@@ -0,0 +1,422 @@
...
+wishes to do based on those arguments.
+
+```c++
+int APIENTRY wWinMain(
+    _In_ HINSTANCE hInstance, _In_opt_ HINSTANCE hPrevInstance,
+    _In_ LPWSTR lpCmdLine, _In_ int nCmdShow)
+{
+    // First, hook up the Activated event, to allow for this instance of the app
+    // getting reactivated as a result of multi-instance redirection.
+    AppInstance::Activated([](ActivationArguments args)
```
oldnewthing, line 234: We seem to be inconsistent about which things are members of the singleton AppInstance object and which are statics. Should pick a pattern and apply consistently.

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
```diff
@@ -0,0 +1,422 @@
...
+
+    // For example, we might want to redirect protocol activations.
+    if (kind == ExtendedActivationKind::Protocol)
+    {
+        auto protocolArgs = args.Data().as<ProtocolActivatedEventArgs>();
+        Uri uri = protocolArgs.Uri();
+
+        // We'll try to find the instance that handles protocol activations.
+        // If there isn't one, then this instance will take over that duty.
+        auto instance = AppInstance::FindOrRegisterForKey(uri.AbsoluteUri().c_str());
```
oldnewthing, line 286:
```suggestion
        auto instance = AppInstance::FindOrRegisterForKey(uri.AbsoluteUri());
```

No need to convert from `hstring` to `wchar_t const*` then back to `hstring`.

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
oldnewthing, line 234:
```suggestion
    AppInstance::Activated([](ActivationArguments const& args)
```

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
```diff
@@ -0,0 +1,422 @@
...
+to include:
+
+```idl
+namespace Microsoft.Windows.AppLifecycle
+{
+    runtimeclass AppInstance
+    {
+        // Existing activation functionality like GetCurrent and GetActivatedEventArgs
+
+        static event EventHandler<ActivationArguments> Activated;
```
oldnewthing, line 365:
```suggestion
        static event EventHandler<AppActivationArguments> Activated;
```
?

or
```
        static event EventHandler<AppInstanceActivatedEventArgs> Activated;
```
?

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
```diff
@@ -0,0 +1,422 @@
...
+target for redirection.
+
+Following the previous example of an app that supports filetype activation, and redirects to the
+instance that already has the selected file open - when that file is closed, the app could
+unregister itself.
+
+```c++
+void CALLBACK OnFileClosed(HWND hDlg, UINT message, WPARAM wParam, LPARAM lParam)
+{
+    AppInstance::GetCurrent().UnregisterKey(g_fileName);
```
oldnewthing, line 326: What happens if the app exits without unregistering the key, and as it's on its way out, somebody tries to redirect to it? Does the redirection silently disappear?

How long should an app linger after unregistering the key to make sure that any pending redirections have been drained?

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
```diff
@@ -0,0 +1,422 @@
...
+    runtimeclass AppInstance
+    {
+        // Existing activation functionality like GetCurrent and GetActivatedEventArgs
+
+        static event EventHandler<ActivationArguments> Activated;
+
+        static Windows.Foundation.IVector<AppInstance> GetInstances();
+
+        static AppInstance FindOrRegisterForKey(String key);
+        void UnregisterKey(String key);
```
oldnewthing, line 370: Should we call this UnregisterRedirectionKey so it's clearer what the key is for? (Similar the other key-named methods.)

We used the shorter name in UWP because AppInstance was just for app activation redirection. But in Reunion, AppInstance is used for a lot of stuff beyond just redirection, so this would help clarify the purpose of the key for people who wander into AppInstance for non-redirection-related reasons.

riverar:
@jonwis Friendly reminder re: teams URL.

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
```diff
@@ -0,0 +1,422 @@
...
+
+        void RedirectTo(ActivationArguments args);
```
oldnewthing, line 372: I think the method name `RedirectActivationTo` is clearer about what is being redirected, especially since AppInstance is now being used for lots more than just activation redirection.

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
```diff
@@ -0,0 +1,422 @@
...
+
+        String Key;
```
oldnewthing, line 374: Why is this key mutable? What happens if I set it directly rather than using FindOrRegisterForKey? What if I set it to a value that duplicates another instsance?

specs/AppLifecycle/StateNotifications/.gitkeep
```diff
@@ -0,0 +1 @@
+
```
oldnewthing, line 1: This file is unnecessary because the directory is nonempty.

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
```diff
@@ -0,0 +1,389 @@
...
+notification it wants to receive, and when the event is raised the system invokes that callback. The
+app could start listening for the notifications at any time; it would most often do this early in
+its life, perhaps in WinMain - but the AppLifecycle mechanism does not impose any constraints on
+this, and the app is free to subscribe or unsubscribe at any time.
+
+```c++
+int APIENTRY wWinMain(
+    _In_ HINSTANCE hInstance, _In_opt_ HINSTANCE hPrevInstance,
+    _In_ LPWSTR lpCmdLine, _In_ int nCmdShow)
+{
```
oldnewthing, line 73: Do we need to CoInitialize? Or can these methods operate without COM?

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
```diff
@@ -0,0 +1,422 @@
...
+activation is redirected to this instance. The AppInstance class will expose an Activated event, for
+which the app can register. In its Activated event handler, the app would examine the type of
+activation, retrieve the rich ActivationArguments object, and then do whatever app-specific work it
+wishes to do based on those arguments.
+
+```c++
+int APIENTRY wWinMain(
+    _In_ HINSTANCE hInstance, _In_opt_ HINSTANCE hPrevInstance,
+    _In_ LPWSTR lpCmdLine, _In_ int nCmdShow)
+{
```
oldnewthing, line 231: Do we need to CoInitialize? Or can these methods operate without COM?

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
```diff
@@ -0,0 +1,422 @@
...
+
+If it doesn't find another instance with the file open, it registers itself as the instance that has
+this file open, and continues with normal app initialization (creating its main window, starting its
+message pump, and so on).
+
+```c++
+int APIENTRY wWinMain(
+    _In_ HINSTANCE hInstance, _In_opt_ HINSTANCE hPrevInstance,
+    _In_ LPWSTR lpCmdLine, _In_ int nCmdShow)
+{
```
oldnewthing, line 117: Do we need to CoInitialize? Or can these methods operate without COM?

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
```diff
@@ -0,0 +1,389 @@
...
+    while (GetMessage(&msg, nullptr, 0, 0))
+    {
+        TranslateMessage(&msg);
+        DispatchMessage(&msg);
+    }
+    return (int)msg.wParam;
+}
+```
+
+In the app-defined callback, the app can do whatever it likes. Typically, it would fetch the current
```
oldnewthing, line 106: Presumably the callback is invoked from an arbitrary thread (properly marshaled, to respect the agility of the delegate), and apps need to apply appropriate synchronization as required. Should document this.

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
```diff
@@ -0,0 +1,389 @@
...
+this, and the app is free to subscribe or unsubscribe at any time.
+
+```c++
+int APIENTRY wWinMain(
+    _In_ HINSTANCE hInstance, _In_opt_ HINSTANCE hPrevInstance,
+    _In_ LPWSTR lpCmdLine, _In_ int nCmdShow)
+{
+    // Optionally, register callbacks for power/battery state changes.
+    PowerManager::BatteryStatusChanged([](auto&&...)
+        { OnBatteryStateChange(); });
```
oldnewthing, line 76:
```suggestion
        { OnBatteryStatusChanged(); });
```

Helper method name should align with event name, to reduce confusion when reading the sample.

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
```diff
@@ -0,0 +1,389 @@
...
+status of the particular feature (that is, the battery status, power supply status, energy saver
+status, and so on), and use that information to form its decision. For example, if battery is at a
+low power state and discharging, the app might choose to defer any non-critical background work, in
+the interests of being a good power/battery citizen.
+
+```c++
+void OnBatteryStateChange()
+{
+    BatteryStatus batteryState = PowerManager::BatteryStatus();
+    int remainingCharge = PowerManager::RemainingChargePercent();
```
oldnewthing, line 116: What does this return if there is no battery?

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
```diff
@@ -0,0 +1,389 @@
...
+```c++
+int APIENTRY wWinMain(
+    _In_ HINSTANCE hInstance, _In_opt_ HINSTANCE hPrevInstance,
+    _In_ LPWSTR lpCmdLine, _In_ int nCmdShow)
+{
+    // Optionally, register callbacks for power/battery state changes.
+    PowerManager::BatteryStatusChanged([](auto&&...)
+        { OnBatteryStateChange(); });
+    PowerManager::PowerSupplyStatusChanged([](auto&&...)
+        { OnPowerStateChange(); });
```
oldnewthing, line 78:
```suggestion
        { OnPowerSupplyStatusChanged(); });
```

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
```diff
@@ -0,0 +1,389 @@
...
+    PowerManager::EnergySaverStatusChanged([](auto&&...)
+        { OnEnergySaverStateChange(); });
```
oldnewthing, line 80:
```suggestion
        { OnEnergySaverStateChanged(); });
```
etc for the remaining events.

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
```diff
@@ -0,0 +1,389 @@
...
+```
+
+### New API combines multiple old APIs
+
+The additional enums, properties and events brought forward for the PowerSettingRegisterNotification
+functionality should be completely seamless and indistinguishable from the existing PowerManager
+members, such that the app can consume these in exactly the same way.
+
+```c++
+PowerManager::ConsoleStatusChanged([](auto&&...)
```
oldnewthing, line 169: Should this be named DisplayStatusChanged, since that's the property you query from the PowerManager when it is raised?

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
```diff
@@ -0,0 +1,389 @@
...
+    }
+
+    // Clone of Windows.System.Power.PowerSupplyStatus
+    enum PowerSupplyStatus {
+        NotPresent,
+        Inadequate,
+        Adequate
+    }
+
+    // Status of the device's power source
```
oldnewthing, line 278:
```suggestion
    // Kind of the device's power source
```

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
```diff
@@ -0,0 +1,389 @@
...
+
+    // User presence indicator - whether the user is present or absent
+    // (active or idle)
+    enum UserPresenceStatus {
+        Present,
+        Absent
+    }
+
+    // Indicates whether the system is entering or exiting "away" mode
+    enum SystemAwayModeStatus {
```
oldnewthing, line 313: Entering/Exiting is a transition, so not meaningful for a status query. (Unless the system is always "entering" and "exiting" away mode and never quite gets there or leaves?) Should be something like

```csharp
enum SystemAwayMode
{
    NotAway,
    Away,
}
```

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
```diff
@@ -0,0 +1,389 @@
...
+
+        // gets the device's energy-saver status
+        static EnergySaverStatus EnergySaverStatus { get; };
+
+        // gets the device's power supply status
+        static PowerSupplyStatus PowerSupplyStatus { get; };
+
+        // gets the total percentage of charge remaining from all batteries connected to
+        // the device
+        static Int32 RemainingChargePercent { get; };
```
oldnewthing, line 332: What if there is no battery?

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
```diff
@@ -0,0 +1,389 @@
...
+
+        // gets the total runtime remaining from all batteries connected to the device
+        static TimeSpan RemainingDischargeTime { get; };
```
oldnewthing, line 335: What if there is no battery?

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
jonwis: SAMPLE: Fix

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
jonwis: DISCUSSION: Yes, per above comment - we'll document how this is different from the UWP activation model.

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
jonwis: DISCUSSION: Yes, the underlying code ensures that someone wins and is the target for the key.

RECOMMEND: Add comments about what this sample is trying to do.

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
jonwis: RECOMMEND: Add the roinit helper from C++/WinRT, though it auto-inits

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
jonwis: RECOMMEND: Typo from the prior version of the spec - the Activated event goes on an instance, which you get from `::GetCurrent()`

jonwis: RECOMMEND: Fix

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
jonwis: RECOMMEND: Fix

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
jonwis: DISCUSSION: The lifetime of the registration will be pulled out on process termination, but an app needs to proactively remove registrations during shutdown. The UWP activation management correctly determines whether the target instance had terminated, and spin up a new instance to handle the lingering redirection.

CONSIDER: Figure out a way to signal back to the original redirector. Might imply a change to the `Redirect...` call, including some form of `Async` behavior so apps don't have to block.  There's a race between apps who want to redirect to another app and the target deciding to terminate.

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
jonwis: RECOMMEND: `AppActivationEventArgs` should have been the name.

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
jonwis: DISCUSSION: This key gets fancier over time - for now, we're using the key for redirection. For the future, we might add a way to use the key to find IPC ports.

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
jonwis: RECOMMEND: Use `RedirectActivationTo`

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
jonwis: RECOMMEND: Fix, typo during conversion; should be read-only

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
```diff
@@ -0,0 +1,389 @@
...
+This spec describes new APIs that address these issues by incorporating a comprehensive set of the
+[Windows.System.Power.PowerManager](https://docs.microsoft.com/en-us/uwp/api/Windows.System.Power.PowerManager)
+WinRT API in the Reunion AppLifecycle component, and include additional low-level notifications
+based on the Win32
+[PowerSettingRegisterNotification](https://docs.microsoft.com/en-us/windows/win32/api/powersetting/nf-powersetting-powersettingregisternotification)
+and matching
+[PowerSettingUnregisterNotification](https://docs.microsoft.com/en-us/windows/win32/api/powersetting/nf-powersetting-powersettingunregisternotification)
+APIs.
+
+<br>
```
jonwis, line 32: RECOMMEND: Remove HTML junk

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
jonwis: RECOMMEND: See previous answers

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
jonwis: RECOMMEND: Fix to have the names match the events

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
jonwis: RECOMMEND: Same

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
jonwis: RECOMMEND: Same

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
jonwis: RECOMMEND: Document that these are regular WinRT event delegate invocations. See also https://github.com/microsoft/xlang/issues/673

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
jonwis: RECOMMEND: Add documentation into the IDL/docs. The `state` has an enum that says `NotPresent`.  The remaining charge percent should probably be "all" when "not present" (desktop-class machine.)

SAMPLE: This sample does not run power-intensive work when on AC, for example.

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
jonwis: RECOMMEND: Change to `DisplayStatusChanged`

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
```diff
@@ -0,0 +1,389 @@
...
+### New API combines multiple old APIs
+
+The additional enums, properties and events brought forward for the PowerSettingRegisterNotification
+functionality should be completely seamless and indistinguishable from the existing PowerManager
+members, such that the app can consume these in exactly the same way.
+
+```c++
+PowerManager::ConsoleStatusChanged([](auto&&...)
+{
+    DisplayStatus displayState = PowerManager::DisplayStatus();
```
jonwis, line 171: RECOMMEND: s/displayState/status/

jonwis: DISCUSSION: "State" is how things are when you check them.  "Status" is the result of an operation of some kind.

RECOMMEND: Normalize on `-State` since these are properties, not operations

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
jonwis: RECOMMEND: Fix

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
```diff
@@ -0,0 +1,389 @@
...
+    // State of the primary display
+    enum DisplayState {
+        Off,
+        On,
+        Dimmed
+    }
+
+    // Indicates whether the system is busy or idle.
+    enum SystemBusyState {
+        Idle,
```
jonwis, line 294: DISCUSSION: There's a single pulse, "going from idle to non-idle" which means there's no way to ask the system "are you idle," and there's no way to know that the system is going from non-idle to idle.  This event is the "hey you should go do a bunch of work."  There's no matching property of `SystemBusyState` to query.

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
jonwis: DISCUSSION: Many of these names were straight conversions from the GUID forms of the power state identifiers. 

RECOMMEND: Use the `SystemAwayMode` listed.

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
jonwis: DISCUSSION: See above; need a resolution on behavior here.

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
jonwis: DISCUSSION: See above

specs/AppLifecycle/Activation/AppLifecycle Activation.md
```diff
@@ -0,0 +1,506 @@
...
+This spec addresses the rich activation APIs in the component.
+
+## Background
+
+Traditional Win32 apps expect to get their arguments passed into WinMain in the form of a string
+array. They can also call the Win32
+[GetCommandLineW](https://docs.microsoft.com/en-us/windows/win32/api/processenv/nf-processenv-getcommandlinew)
+API. Windows Forms apps expect to call
+[Environment.GetCommandLineArgs](https://docs.microsoft.com/en-us/dotnet/api/system.environment.getcommandlineargs)
+for the same purpose. This is true regardless of the "activation kind" - in fact, for unpackaged
```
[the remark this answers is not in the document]
riverar: Sparse apps can leverage identity from a package but are otherwise unpackaged. What about those?

specs/AppLifecycle/Activation/AppLifecycle Activation.md
```diff
@@ -0,0 +1,506 @@
...
+-   Rich activation objects
+-   Selective single/multi-instancing
+-   System state and power notifications.
+
+This spec addresses the rich activation APIs in the component.
+
+## Background
+
+Traditional Win32 apps expect to get their arguments passed into WinMain in the form of a string
+array. They can also call the Win32
```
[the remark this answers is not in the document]
andreww-msft: Fixed.

specs/AppLifecycle/Activation/AppLifecycle Activation.md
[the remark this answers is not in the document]
andreww-msft: Fixed.

specs/AppLifecycle/Activation/AppLifecycle Activation.md
```diff
@@ -0,0 +1,506 @@
...
+registration, others do not. The target list for v1 is as follows:
+
+| Activation kind | Description                                                                                                                                      |
+| --------------- | ------------------------------------------------------------------------------------------------------------------------------------------------ |
+| Launch          | Launch from a Start tile/shortcut, from a command window, or programmatically via ShellExecute/CreateProcess.                                    |
+| File            | The app is activated when the user double-clicks a registered file, or when a caller calls ShellExecute or LaunchFileAsync on a registered file. |
+| Protocol        | Activated via ShellExecute or LaunchUriAsync, or by specifying a protocol string on a command-line.                                              |
+| StartupTask     | The app starts on user log-in - either registered in the registry, or via shortcuts in a well-known startup folder.                              |
+
+<br/>
```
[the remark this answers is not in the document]
andreww-msft: Fixed.

specs/AppLifecycle/Activation/AppLifecycle Activation.md
```diff
@@ -0,0 +1,506 @@
...
+can call to register its interest in rich activations. The app is not required to have an MSIX
+manifest, nor an MSIX package. Using the API approach rather than the manifest approach has the
+added benefit that it also addresses a related customer request for dynamic activation registrations
+even for UWP apps.
+
+Currently all registered activation points are managed per-user. If your app is installed for
+multiple users via an MSI or elevated installer your app will need to re-register launch points for
+every user. For v1, the platform will support only per-user registration.
+
+ <br>
```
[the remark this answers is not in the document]
andreww-msft: Fixed.

specs/AppLifecycle/Activation/AppLifecycle Activation.md
```diff
@@ -0,0 +1,506 @@
...
+### Register for rich activation
+
+Providing APIs for activation registration means that an app can register at any time. The most
+common case would be very early in the app's main function. The app can register for any of the
+supported activation kinds. For most activation kinds, the app can register multiple options (for
+example, multiple filetypes or dataformats, or verbs) - and could therefore call the RegisterXXX
+function multiple times, if required. For example, an app could register for ".jpg" and ".bmp"
+filetype activations at one point, and then later also register for ".png".
+
+Each registered filetype must be unique for the app. To allow for the case where the app determines
```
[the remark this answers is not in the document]
andreww-msft: Fixed.

specs/AppLifecycle/Activation/AppLifecycle Activation.md
```diff
@@ -0,0 +1,506 @@
...
+the combination of options at runtime (and potentially changes the combinations at a later time),
+where an app registers the same filetype multiple times, the latest registration is honored and
+overwrites any previous registrations.
+
+In the existing platform, UWP apps can only register one startup activation extension. However, for
+packaged desktop apps, multiple startup extensions are permitted, and each one can specify a
```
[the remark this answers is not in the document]
andreww-msft: Clarified.

specs/AppLifecycle/Activation/AppLifecycle Activation.md
```diff
@@ -0,0 +1,506 @@
...
+    // Register one or more supported image filetypes,
+    // an icon (specified by binary file path plus resource index),
+    // a display name to use in Shell and Settings,
+    // zero or more verbs for the File Explorer context menu,
+    // and the path to the EXE to register for activation.
+    std::wstring imageFileTypes[3] = { L".jpg", L".png", L".bmp" };
+    std::wstring verbs[2] = { L"view", L"edit" };
+    ActivationRegistrationManager::RegisterForFileTypeActivation(
+        imageFileTypes,
+        L"C:\\Program Files\\Contoso\\MyResources.dll, 123",
```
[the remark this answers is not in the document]
andreww-msft: Fixed.

specs/AppLifecycle/Activation/AppLifecycle Activation.md
```diff
@@ -0,0 +1,506 @@
...
+        L"Contoso File Types",
```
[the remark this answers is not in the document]
andreww-msft: Added a note.

specs/AppLifecycle/Activation/AppLifecycle Activation.md
```diff
@@ -0,0 +1,506 @@
...
+    // Register some URI schemes for protocol activation,
+    // specifying the scheme name, icon, display name and EXE path.
+    ActivationRegistrationManager::RegisterForProtocolActivation(
+        L"foo",
+        L"C:\\Program Files\\Contoso\\MyResources.dll, 45",
+        L"Contoso Foo Protocol",
+        L"C:\\Program Files\\Contoso\\MyApp.exe");
+
+    // Register a protocol, specifying the scheme name,
+    // UWP-style image file path, display name, and defaulting
```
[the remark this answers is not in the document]
andreww-msft: Fixed.

specs/AppLifecycle/Activation/AppLifecycle Activation.md
```diff
@@ -0,0 +1,506 @@
...
+originally passed in when registering.
+
+```c++
+void UnregisterForActivation()
+{
+    // Unregister one or more registered filetypes.
+    std::wstring imageFileTypes[3] = { L".jpg", L".png", L".bmp" };
+    ActivationRegistrationManager::UnregisterForFileTypeActivation(
+        imageFileTypes,
+        L"C:\\Program Files\\Contoso\\MyApp.exe");
```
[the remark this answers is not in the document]
andreww-msft: Added a note.

specs/AppLifecycle/Activation/AppLifecycle Activation.md
```diff
@@ -0,0 +1,506 @@
...
+        L"AnotherTaskId",
+        L"");
+}
+```
+
+## API Details
+
+### AppInstance
+
+This type represents an activaiton of an application in Project Reunion, providing methods to find
```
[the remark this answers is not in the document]
andreww-msft: Fixed.

specs/AppLifecycle/Activation/AppLifecycle Activation.md
```diff
@@ -0,0 +1,506 @@
...
+
+**GetActivatedEventArgs** provides similar behavior as the existing platform
+[AppInstance.GetActivatedEventArgs](https://docs.microsoft.com/en-us/uwp/api/windows.applicationmodel.appinstance.getactivatedeventargs),
+except that it returns a new ActivationArguments object for the current activation instead of an
+IActivatedEventArgs.
+
+
+### ExtendedActivationKind
+
+namespace Microsoft.Windows.AppLifecycle { enum ExtendedActivationKind { ExtensionBase = 4096 } }
```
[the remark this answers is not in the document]
andreww-msft: Fixed.

specs/AppLifecycle/Activation/AppLifecycle Activation.md
```diff
@@ -0,0 +1,506 @@
...
+
+This enum is based off the platform
+[ActivationKind](https://docs.microsoft.com/en-us/uwp/api/Windows.ApplicationModel.Activation.ActivationKind).
+All platform ActivationKind values are cloned to this class. Then, we set a high-value
+**ExtensionBase** value which is well above the existing highest ActivationKind value (1026), to
+allow for new values to be added to the platform. Going forward, any new platform values will also
+be added to the new enum, plus any new values that are only defined in the new enum.
```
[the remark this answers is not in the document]
andreww-msft: Clarified.

specs/AppLifecycle/Activation/AppLifecycle Activation.md
```diff
@@ -0,0 +1,506 @@
...
+
+Similarly, if the app registers multiple times for startup, this is also idempotent (a single
+registration, no error, any new values supplied overwrite any old values).
+
+```idl
+namespace Microsoft.Windows.AppLifecycle
+{
+    static runtimeclass ActivationRegistrationManager
+    {
+        static void RegisterForFileTypeActivation(String[] supportedFileTypes, String logo,
```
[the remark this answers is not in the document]
andreww-msft: Samples do compile.

specs/AppLifecycle/Activation/AppLifecycle Activation.md
andreww-msft: Added a note.

specs/AppLifecycle/Activation/AppLifecycle Activation.md
andreww-msft: No, added a note above.

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
andreww-msft: Expanded description.

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
andreww-msft: Fixed.

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
andreww-msft: Fixed.

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
andreww-msft: Fixed.

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
andreww-msft: Added further description: in Reunion, redirection is not terminal.

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
andreww-msft: Fixed.

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
andreww-msft: Fixed.

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
andreww-msft: Fixed.

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
andreww-msft: Fixed.

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
andreww-msft: Fixed.

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
andreww-msft: Added further description.

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
andreww-msft: Fixed.

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
andreww-msft: Fixed.

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
andreww-msft: Fixed.

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
andreww-msft: Added further description.

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
andreww-msft: Fixed.

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
andreww-msft: Added further description in pre-amble above.

specs/AppLifecycle/Activation/AppLifecycle Activation.md
```diff
@@ -0,0 +1,506 @@
...
+This type represents an activaiton of an application in Project Reunion, providing methods to find
+other instances of itself, redirect activations across instances, and handle incoming non-launch
+activations.
+
+```idl
+namespace Microsoft.Windows.AppLifecycle
+{
+    runtimeclass AppInstance
+    {
+        static AppInstance GetCurrent();
```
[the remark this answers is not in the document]
andreww-msft: Properties should not throw exceptions, and we could throw in this workflow, hence the method.

specs/AppLifecycle/Activation/AppLifecycle Activation.md
```diff
@@ -0,0 +1,506 @@
...
+
+        ActivationArguments GetActivatedEventArgs();
```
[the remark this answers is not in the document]
andreww-msft: Yes, following precedent.

specs/AppLifecycle/Activation/AppLifecycle Activation.md
```diff
@@ -0,0 +1,506 @@
...
+
+
+### ExtendedActivationKind
+
+namespace Microsoft.Windows.AppLifecycle { enum ExtendedActivationKind { ExtensionBase = 4096 } }
+
+This enum is based off the platform
+[ActivationKind](https://docs.microsoft.com/en-us/uwp/api/Windows.ApplicationModel.Activation.ActivationKind).
+All platform ActivationKind values are cloned to this class. Then, we set a high-value
+**ExtensionBase** value which is well above the existing highest ActivationKind value (1026), to
```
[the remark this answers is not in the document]
andreww-msft: It doesn't. We'll document only, not have a specific enum value for this.

specs/AppLifecycle/Activation/AppLifecycle Activation.md
andreww-msft: Localization is an exercise for vNext, we cannot fit it into v1.

specs/AppLifecycle/Instancing/AppLifecycle SingleMulti-Instancing.md
andreww-msft: Added further description.

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
andreww-msft: Fixed.

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
andreww-msft: Fixed.

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
andreww-msft: Fixed.

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
andreww-msft: Fixed.

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
andreww-msft: Fixed.

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
andreww-msft: Added further description, plus the link.

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
andreww-msft: Fixed.

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
andreww-msft: Fixed (but see later, all "status" changed to "state".

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
andreww-msft: Fixed.

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
andreww-msft: Fixed.

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
andreww-msft: Fixed.

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
andreww-msft: Clarified: if Battery::NotPresent, then RemainingChargePercent==100.

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
andreww-msft: Fixed.

specs/AppLifecycle/StateNotifications/AppLifecycle StateNotifications.md
andreww-msft: Fixed.

PR #579, opened 2021-03-12
Implementation and tests for GetEnvironmentVariables

dhoehna:
PR for implementing and testing the functionality to get all environment variables for process, user, and machine scope.

Getting the environment variables for process scope has a different API set as compared to user and machine.  Because of that, there is a check to see what the scope is.


dev/Microsoft.Process.Environment/EnvironmentManager.cpp
```diff
@@ -31,7 +31,18 @@ namespace winrt::Microsoft::ProjectReunion::implementation
 
     Windows::Foundation::Collections::IMapView<hstring, hstring> EnvironmentManager::GetEnvironmentVariables()
     {
-        throw hresult_not_implemented();
+        Windows::Foundation::Collections::StringMap environmentVariables;
+
+        if (m_Scope == Scope::Process)
+        {
+            environmentVariables = GetProcessEnvironmentVariables();
```
shrbinic, line 38: Test comment

dev/Microsoft.Process.Environment/EnvironmentManager.cpp
```diff
@@ -63,4 +74,125 @@ namespace winrt::Microsoft::ProjectReunion::implementation
...
     }
+
+    StringMap EnvironmentManager::GetProcessEnvironmentVariables()
+    {
+        //Get the pointer to the process block
+        LPWSTR environmentVariablesString = GetEnvironmentStrings();
+        THROW_HR_IF_NULL(E_POINTER, environmentVariablesString);
+
+        // parse the block into a map
+        LPTSTR variable;
```
shrbinic, line 85: Why not a LPWSTR here?  I think LPTSTR is an old thing we're not supposed to use anymore

dev/Microsoft.Process.Environment/EnvironmentManager.cpp
```diff
@@ -63,4 +74,125 @@ namespace winrt::Microsoft::ProjectReunion::implementation
     {
         throw hresult_not_implemented();
     }
+
+    StringMap EnvironmentManager::GetProcessEnvironmentVariables()
+    {
+        //Get the pointer to the process block
+        LPWSTR environmentVariablesString = GetEnvironmentStrings();
```
shrbinic, line 81: I believe it's preferable to use PWSTR instead of LPWSTR.  They are the same but LPWSTR is obsolete.

DrusTheAxe: Yes. Avoid L-prefixes. They're from days of yore when Win16 and near+far pointers roamed the earth... ;-)

Very passe'. Current coding standards say not to use them anymore

DrusTheAxe, line 81: `PWSTR environmentVariablesString{ GetEnvironmentStrings() };`

dev/Microsoft.Process.Environment/EnvironmentManager.cpp
DrusTheAxe, line 85: `PWSTR variable{};`

Always initialize variables

dev/Microsoft.Process.Environment/EnvironmentManager.cpp
```diff
@@ -63,4 +74,125 @@ namespace winrt::Microsoft::ProjectReunion::implementation
...
+        THROW_HR_IF_NULL(E_POINTER, environmentVariablesString);
+
+        // parse the block into a map
+        LPTSTR variable;
+        variable = (LPTSTR)environmentVariablesString;
```
DrusTheAxe, line 86: Why the cast?

1. Use C++ new-style cast notation (const_cast etc). Don't use C style casts.
2. Cast shouldn't be necessary

DrusTheAxe, line 86: Don't declare a variable and then (separately) immediately assign to it. Merge lines 85+86

```
auto variable{ environmentVariablesString };
```

dev/Microsoft.Process.Environment/EnvironmentManager.cpp
```diff
@@ -63,4 +74,125 @@ namespace winrt::Microsoft::ProjectReunion::implementation
...
+
+        StringMap environmentVariables;
+        while (*variable)
+        {
+            std::wstring environmentVariable(variable);
+            size_t locationOfNull = environmentVariable.find_last_of(L'=');
```
DrusTheAxe, line 92: NIT: `...locationOfNull{ ... };`

DrusTheAxe, line 92: NIT: `const size_t...`

DrusTheAxe, line 92: Why find_last_of rather than find_first_of?

dev/Microsoft.Process.Environment/EnvironmentManager.cpp
```diff
@@ -63,4 +74,125 @@ namespace winrt::Microsoft::ProjectReunion::implementation
...
+
+            environmentVariables.Insert(environmentVariable.substr(0, locationOfNull), environmentVariable.substr(locationOfNull + 1));
+
+            variable += lstrlen(variable) + 1;
```
DrusTheAxe, line 96: wcslen

dev/Microsoft.Process.Environment/EnvironmentManager.cpp
DrusTheAxe, line 92: NIT: Naming's a bit odd.

SUGGESTION: `offsetToNull` or even just `offset`

dev/Microsoft.Process.Environment/EnvironmentManager.cpp
```diff
@@ -63,4 +74,125 @@ namespace winrt::Microsoft::ProjectReunion::implementation
...
+        THROW_HR_IF_NULL(E_POINTER, environmentVariablesString);
+
+        // parse the block into a map
+        LPTSTR variable;
+        variable = (LPTSTR)environmentVariablesString;
+
+        StringMap environmentVariables;
+        while (*variable)
+        {
+            std::wstring environmentVariable(variable);
```
DrusTheAxe, line 91: You're just looking to enumerate the strings in the block and split that at the '='. No need to make a string object to do the work

```
for (auto variable=environmentVariablesString; *variable; variable += wcslen(variable) + 1)
{
    auto delimiter{ wcschr(variable, L'=') };
    FAIL_FAST_HR_IF_NULL(E_UNEXPECTED, delimiter);
    std::wstring variableName(variable, 0, delimiter - variable);
    auto variableValue{ delimiter + 1 };
    environmentVariables.Insert(variableName, variableValue);
}
```

dev/Microsoft.Process.Environment/EnvironmentManager.cpp
```diff
@@ -63,4 +74,125 @@ namespace winrt::Microsoft::ProjectReunion::implementation
...
+            , nullptr // lpcchClass
+            , nullptr // lpReserved
+            , nullptr // lpcSubKeys
+            , nullptr // lpcbMaxSubKeyLen
+            , nullptr // lpcbMaxCLassLen
+            , &numberOfValues //lpcValues
+            , &sizeOfLongestNameInCharacters
+            , &sizeOfLongestValueInCharacters
+            , nullptr // lpcmSecurityDescriptor
+            , nullptr)); // lpftLastWriteTime
```
DrusTheAxe, line 133: Ugh. Lots of noise here

Also, registry APIs return LONG i.e. Win32 error (not HRESULT). Needs `THROW_IF_WIN32_ERROR`

```
THROW_IF_WIN32_ERROR(RegQueryInfoKeyW(environmentVariablesHKey.get(), nullptr, nullptr, nullptr, nullptr, nullptr, nullptr,
                                      &numberOfValues, &sizeOfLongestNameInCharacters, &sizeOfLongestValueInCharacters, nullptr, nullptr));
```

dev/Microsoft.Process.Environment/EnvironmentManager.cpp
```diff
@@ -63,4 +74,125 @@ namespace winrt::Microsoft::ProjectReunion::implementation
...
+
+        // +1 for null character
+        const DWORD NAME_LENGTH = sizeOfLongestNameInCharacters + 1;
```
DrusTheAxe, line 136: `c_nameLength`

ALL_CAPS is for preprocessor symbols.

dev/Microsoft.Process.Environment/EnvironmentManager.cpp
```diff
@@ -63,4 +74,125 @@ namespace winrt::Microsoft::ProjectReunion::implementation
...
+        const DWORD VALUE_SIZE_IN_BYTES = sizeOfLongestValueInCharacters * sizeof(WCHAR);
```
DrusTheAxe, line 137: `c_valueSizeInBytes`

DrusTheAxe, line 137: DWORD * sizeof() == DWORD * size_t == size_t

Working on x86? You won't see these things. One of the reasons I recommend primarily working on x64 builds and let the compiler spot these (rather than the build pipeline :-)

```
const DWORD c_valueSizeInBytes{ static_cast<DWORD>(sizeOfLongestValueInCharacters * sizeof(WCHAR)) };
```

dev/Microsoft.Process.Environment/EnvironmentManager.cpp
```diff
@@ -63,4 +74,125 @@ namespace winrt::Microsoft::ProjectReunion::implementation
...
+            DWORD nameLength = NAME_LENGTH;
+            DWORD valueSize = VALUE_SIZE_IN_BYTES;
+            LSTATUS enumerationStatus = RegEnumValueW(environmentVariablesHKey.get()
+                , valueIndex
+                , environmentVariableName.get()
+                , &nameLength
+                , nullptr
+                , nullptr
+                , environmentVariableValue.get()
+                , &valueSize);
```
DrusTheAxe, line 153: Why
do
you
hate
widescreen
monitors
so
?
<g>

dev/Microsoft.Process.Environment/EnvironmentManager.cpp
```diff
@@ -63,4 +74,125 @@ namespace winrt::Microsoft::ProjectReunion::implementation
...
+            , nullptr)); // lpftLastWriteTime
+
+        // +1 for null character
+        const DWORD NAME_LENGTH = sizeOfLongestNameInCharacters + 1;
+        const DWORD VALUE_SIZE_IN_BYTES = sizeOfLongestValueInCharacters * sizeof(WCHAR);
+
+        std::unique_ptr<wchar_t[]> environmentVariableName(new wchar_t[NAME_LENGTH]);
+        std::unique_ptr<BYTE[]> environmentVariableValue(new BYTE[VALUE_SIZE_IN_BYTES]);
+
+        for (int valueIndex = 0; valueIndex < numberOfValues; valueIndex++)
```
DrusTheAxe, line 142: `DWORD`

you later compare with numberOfValues -- signed/unsigned comparison

Do you have warning level = 4? (/W4). It usually spots these sorts of things. If you don't you should (VS template defaults = /W3. Why oh why in this day and age...)

dev/Microsoft.Process.Environment/EnvironmentManager.cpp
```diff
@@ -63,4 +74,125 @@ namespace winrt::Microsoft::ProjectReunion::implementation
...
+            // If there was an error getting the value
+            if (enumerationStatus != ERROR_SUCCESS && enumerationStatus != ERROR_NO_MORE_ITEMS)
+            {
+                THROW_HR(HRESULT_FROM_WIN32(enumerationStatus));
+            }
+
+            environmentVariables.Insert(environmentVariableName.get(), reinterpret_cast<LPWSTR>(environmentVariableValue.get()));
+
+            environmentVariableName.reset(new wchar_t[NAME_LENGTH]);
+            environmentVariableValue.reset(new BYTE[VALUE_SIZE_IN_BYTES]);
```
DrusTheAxe, line 170: You determine the longest lengths up front, allocate memory for them and ask the registry API to fill your buffers. So why throw that memory away and allocate a new same sized buffer again? Just "zero out their length"

```
*environmentVariableName.get() = L'\0';
*environmentVariableValue.get() = L'\0';
```

Though even that's questionable. Why bother? Doesn't RegEnumValueW not null-terminate the resulting strings? (and if not you've got a bigger problem not handed above...)

dev/Microsoft.Process.Environment/EnvironmentManager.cpp
```diff
@@ -63,4 +74,125 @@ namespace winrt::Microsoft::ProjectReunion::implementation
...
+        }
+
+        return environmentVariables;
+    }
+
+    wil::unique_hkey EnvironmentManager::GetRegHKeyForEVUserAndMachineScope(bool needsWriteAccess)
+    {
+        assert(m_Scope != Scope::Process);
```
DrusTheAxe, line 178: Don't bother with assert. Same lesson as Windows CHK builds -- unless you routinely use them you get no value (except surprises when they break). And we don't use them.

Alternatives:
* Do nothing
* Error handling e.g. RETURN_HR_IF(E_INVALIDARG, ...)
* FailFast e.g. FAIL_FAST_HR_IF(E_INVALIDARG, ...)

dev/Microsoft.Process.Environment/EnvironmentManager.h
```diff
@@ -29,6 +29,13 @@ namespace winrt::Microsoft::ProjectReunion::implementation
 
     private:
         Scope m_Scope;
```
DrusTheAxe, line 31: NIT: `Scope m_scope{};` (or some initial value)

dev/Microsoft.Process.Environment/EnvironmentManager.h
```diff
@@ -29,6 +29,13 @@ namespace winrt::Microsoft::ProjectReunion::implementation
...
+
+        std::wstring USER_EV_REG_LOCATION = L"Environment";
```
DrusTheAxe, line 33: Does this need to be a wstring? If this is just a constant just make it PCWSTR

or maybe `inline constexpr std::wstring_view blah{ L"Environment" };. There was a recent discussion of how to most efficiently declare null-terminated string constants (wstring_view isn't guaranteed to be null-terminated e.g. `std::wstring_view(std::wstring(L"Foo"), 0, 1);` but in this case it is). There were a dozen+ notations discussed and 2-3 were correct and most efficient, but I don't recall which was best.

@jonwis @oldnewthing do you recall? What's the recommendation for now (until the coding guidelines are updated)?

dev/Microsoft.Process.Environment/EnvironmentManager.h
```diff
@@ -29,6 +29,13 @@ namespace winrt::Microsoft::ProjectReunion::implementation
...
+        std::wstring MACHINE_EV_REG_LOCATION = L"SYSTEM\\CurrentControlSet\\Control\\Session Manager\\Environment";
+
+        StringMap GetProcessEnvironmentVariables();
```
DrusTheAxe, line 36: Does StringMap support move semantics?

DrusTheAxe, line 36: `StringMap Get...() const;` ?

dev/Microsoft.Process.Environment/EnvironmentManager.h
```diff
@@ -29,6 +29,13 @@ namespace winrt::Microsoft::ProjectReunion::implementation
...
+        StringMap GetUserOrMachineEnvironmentVariables();
```
DrusTheAxe, line 37: `StringMap Get...() const;` ?

test/EnvironmentManagerTests/EnvironmentManagerWin32Tests.h
```diff
@@ -11,5 +11,8 @@ namespace ProjectReunionEnvironmentManagerTests
         TEST_METHOD(TestGetForProcess);
         TEST_METHOD(TestGetForUser);
         TEST_METHOD(TestGetForMachine);
+        TEST_METHOD(TestGetEnvironmentVariablesForProcess);
+        TEST_METHOD(TestGetEnvironmentVariablesForUser);
+        TEST_METHOD(TestGetEnvironmentVariablesForMachine);
```
DrusTheAxe, line 16: Where's all the fun RunAs permutations

RunAs:RestrictedUser = MediumIL
RunAs:ElevatedUserOrSystem = Elevated

Dupe all these to cover both those RunAs cases?

test/EnvironmentManagerTests/Helper.h
```diff
@@ -0,0 +1,167 @@
+﻿#pragma once
```
DrusTheAxe, line 1: Missing standard `// Copyright...`

Same applies throughout

test/EnvironmentManagerTests/Helper.h
```diff
@@ -0,0 +1,167 @@
...
+#include "pch.h"
+
+using namespace winrt::Microsoft::ProjectReunion; // for EnvironmentManager
```
DrusTheAxe, line 4: Please don't specify `// for blah`. Well intentioned but they age badly.

test/EnvironmentManagerTests/Helper.h
```diff
@@ -0,0 +1,167 @@
...
+using namespace winrt::Windows::Foundation::Collections; // for IMapView
+typedef IMapView<winrt::hstring, winrt::hstring> EnvironmentVariables;
+
+inline const wchar_t* USER_EV_REG_LOCATION = L"Environment";
```
DrusTheAxe, line 8: inline constexpr ?

DrusTheAxe, line 8: Why `const wchar_t*`? Why not PCWSTR?

The latter includes a SAL annotation `_Null_terminated_`

test/EnvironmentManagerTests/Helper.h
```diff
@@ -0,0 +1,167 @@
...
+inline const wchar_t* MACHINE_EV_REG_LOCATION = L"SYSTEM\\CurrentControlSet\\Control\\Session Manager\\Environment";
+
+inline EnvironmentVariables GetEnvironmentVariablesFromRegistry(HKEY hKey)
```
DrusTheAxe, line 11: Lots of comments but same as mentioned above. I'll wait for an update before commenting here

test/EnvironmentManagerTests/Helper.h
DrusTheAxe, line 1: 'Helper' is so unhelpfully <g> generic.

Please choose a better name

And if there's multiple things going on here split into multiple functionally-oriented files. Better that than one large mosh pit of a dumping ground

test/EnvironmentManagerTests/Helper.h
```diff
@@ -0,0 +1,167 @@
...
+
+        variable += lstrlen(variable) + 1;
+    }
+
+    VERIFY_WIN32_BOOL_SUCCEEDED(FreeEnvironmentStrings(environmentVariablesBlock));
+
+    return environmentVariables.GetView();
+}
+
+/// <summary>
```
DrusTheAxe, line 112: Blech. XML doccomments.

I'm partial to [Doxygen](https://www.doxygen.nl/index.html] with Javadoc-style notation. Cheatsheet for 99%+ of all your doccomment needs:

```
/// First sentence is your brief description just like Javadoc. You can
/// have more verbiage too for a full description if one sentence isn't
/// enough.
/// @param paramname description of parameter
/// @return description of return value
/// @exception MyExceptionObject description of the exception that could be thrown
/// @see xref to one or more class, function, method, variable, file or URL
/// @note message of interest
/// @warning more concerning message of interest
/// @todo description of something to do
and 3+ tilde (~) to start a code block (great for samples). Ends with another line with 3+ tilde (~)
```

Equivalent information but a lot less noise. Plus reading the comment in source becomes reasonably legible in its own right

test/EnvironmentManagerTests/Helper.h
```diff
@@ -0,0 +1,167 @@
...
+        return;
+    }
+
+    auto winRTIterator = fromWinRT.First();
+    do
+    {
+        // Make sure the same key exists in both collections
+        auto winRTKey = winRTIterator.Current().Key();
+
+        std::wstringstream logMessage;
```
DrusTheAxe, line 147: sprintf is your friend :-)

test/EnvironmentManagerTests/Helper.h
```diff
@@ -0,0 +1,167 @@
...
+
+        WEX::Logging::Log::Comment(logMessage.str().c_str());
+
+        VERIFY_IS_TRUE(fromWin32.HasKey(winRTKey));
+
+        // Make sure they have the same value
+        auto winRTValue = winRTIterator.Current().Value();
+
+        std::wstringstream valueMessage;
+        valueMessage << L"Value from test: ";
```
DrusTheAxe, line 159: sprintf FTW!

Not `std::format` until C++20 so [wil::str_printf()](https://github.com/microsoft/wil/wiki/String-helpers#wilstr_printf) is your friend

DrusTheAxe (changes requested):

dev/Microsoft.Process.Environment/EnvironmentManager.cpp
shrbinic, line 85: Also should name this variable something more descriptive than variable :)  head, or stringSearcher? or something...

dev/Microsoft.Process.Environment/EnvironmentManager.cpp
dhoehna: Thank you for the history lesson.  Didn't know those were a thing.  The L prefix has been removed and I'll work on getting all the L's removed.

dev/Microsoft.Process.Environment/EnvironmentManager.cpp
dhoehna: I think I saw an old example somewhere and used it.

dev/Microsoft.Process.Environment/EnvironmentManager.cpp
dhoehna: Correct.  All I'm doing is splitting up the environment block into pairs.

dev/Microsoft.Process.Environment/EnvironmentManager.cpp
dhoehna: Changed.

dev/Microsoft.Process.Environment/EnvironmentManager.cpp
dhoehna: I thought all caps for any constants?

dev/Microsoft.Process.Environment/EnvironmentManager.cpp
dhoehna: Thank you for catching that.  This project has taught me to carefully read what is returned from a windows API.

dev/Microsoft.Process.Environment/EnvironmentManager.cpp
dhoehna: I get it.  You're referring to the fact that I'm multiplying a DWORD and size_t together and it makes a size_t.  Gotta cast that to a DWORD.

dev/Microsoft.Process.Environment/EnvironmentManager.cpp
dhoehna: I was on W3.  Turned on W4.  I'll keep it on.

dev/Microsoft.Process.Environment/EnvironmentManager.cpp
dhoehna: image

I have a widescreen monitor.  It is easier to see.

dev/Microsoft.Process.Environment/EnvironmentManager.cpp
dhoehna: RegEnumValueW returns the strings with a null character.

A pointer to a buffer that receives the name of the value as a null-terminated string.

dhoehna: What is your recommendation?  I like the L"\0" idea.

>and if not you've got a bigger problem not handed above..

Are you referring to the reinterpret_cast of the BYTE array to a LPWSTR? (Going to change it to a PWSTR)

dhoehna: I believe I fixed the issue.  I set the BYTE array at position valueSize to L'\0'

dev/Microsoft.Process.Environment/EnvironmentManager.h
dhoehna: const works.  I can add it to my functions.

dhoehna: I haven't dealt with move before.  How can I tell?

dev/Microsoft.Process.Environment/EnvironmentManager.cpp
dhoehna: Because the first four Environment Variables of this block are
*  =::=::\
*  =C:=C:\Users\dahoehna
* =D:=D:\git\microsoft\ProjectReunion\BuildOutput\Debug\x64\EnvironmentManagerTests
* =ExitCode=00000000

dhoehna: According to GetEnvironmentStrings there should only be 1 "=" and that delimits the value and name.

DrusTheAxe: What are those?

Are those defined by CMD.EXE?

>According to GetEnvironmentStrings there should only be 1 "=" and that delimits the value and name.
"should"? (if I could do the Spock one eyebrow quirked thing that's what I'd be doing right now)

Don't your examples above contradict that statement?

For more fun try this at a command line

```
C:>set X="1=1!=2==1+1"

C:>echo %x%
"1=1!=2==1+1"
```

Searching for 'last' is a bad idea

dev/Microsoft.Process.Environment/EnvironmentManager.cpp
DrusTheAxe: Nope. `PREPROCESSOR_CONSTANTS_SHOULD_BE_ALL_UPCASE` but for C++ constants use usual camelCase like for variables but with a `c_` prefix

test/EnvironmentManagerTests/Helper.h
dhoehna: Because I didn't know PSWSTR was a thing.  I'll be changing them now.

test/EnvironmentManagerTests/Helper.h
dhoehna: I see.  Yeah, the XML styled comments came from my C# days.

test/EnvironmentManagerTests/Helper.h
dhoehna: OMG.  Thank you.  I didn't know this was a thing.

dev/Microsoft.Process.Environment/EnvironmentManager.cpp
dhoehna: > Are those defined by CMD.EXE? 
I have no idea where those are coming from.

> Don't your examples above contradict that statement?
It does.

I'll get it changed.

test/EnvironmentManagerTests/EnvironmentManagerWin32Tests.h
dhoehna: Something like this?

test/EnvironmentManagerTests/Helper.h
dhoehna: Changed to a constexpr and renamed the helper file.

test/EnvironmentManagerTests/EnvironmentManagerCentennialTests.h
```diff
@@ -7,7 +7,7 @@ namespace ProjectReunionEnvironmentManagerTests
 {
     class EnvironmentManagerCentennialTests {
         BEGIN_TEST_CLASS(EnvironmentManagerCentennialTests)
-            TEST_CLASS_PROPERTY(L"RunAs", L"UAP")
+            TEST_CLASS_PROPERTY(L"RunAs", L"{UAP,Restricted,Elevated,InteractiveUser}")
```
DrusTheAxe, line 10: That's new to me. What's it do - run the test once for each context?

DrusTheAxe (approved):

test/EnvironmentManagerTests/EnvironmentManagerCentennialTests.h
dhoehna: Close.  This will make one instance of the test class for each entry called metadata[1-4]

For example, Centennial tests have
ProjectReunionEnvironmentManagerTests::EnvironmentManagerCentennialTests#metadataSet[0-3]

Issue #148, opened 2020-08-21
Enhance MSIX to support Semantic Versioning per http://semver.org


DrusTheAxe:
# Proposal: MSIX support for Semantic Versioning per http://semver.org

## Summary
MSIX package identity supports versioning via `major.minor.build.revision`. This differs from [Semantic Versioning](http://semver.opg) in several ways, in syntax and function. It would be helpful to extend MSIX to support semver's syntax and semantics, both to expand MSIX' functionality and to reduce the impedance mismatch with software using semver.

## Rationale
Semantic Versioning is increasingly popular for software projects. MSIX constrains how semver can be used with packaged software, sometimes severely (especially when bringing non-packaged software using semver to MSIX).

## Scope
| Capability | Priority |
| :---------- | :------- |
| Major version changes have semantic meaning (breaking change) | Should |
| Version fields (especially major.minor.build) support for values >65535 | Should |
| Support 'prerelease' syntax and semantics | Should |
| Support 'buildinfo' metadata | Could |
| Support existing MSIX packages and tooling | Must |

## Important Notes
MSIX package identity has a version of major.minor.build.revision
SemVer has a version of major.minor.revision[-prerelease][+build]

These differ in syntactical and semantic ways.

### Major version
Major in semver is a number and it has functional meaning e.g. major=3 is a breaking change vs major=2. It's significant.

Major in MSIX is just a number. You accomplish the functional meaning by putting the major value in your package name, e.g. KittyVideos version 4.5.6 in semver can be expressed in MSIX as package name = "KittyVideos4" and version = 4.5.6.0. That makes for a different package family from KittyVideos3 and KittyVideos5. That's how framework packages have done it to date e.g. package full name Microsoft.NET.Native.Runtime.2.2_2.2.28604.0_x64__8wekyb3d8bbwe broken down into the 5-tuple:
- name = Microsoft.NET.Native.Runtime.2.2
- version = 2.2.28604.0
- architecture = x64
- resource id = (empty)
- publisher id = 8wekyb3d8bbwe
This gives consumers fine-grained control over taking a breaking change.

However, MSIX doesn't currently provide a way to say "Version >=1.2.3.4 thru 3.* are all equally good to me" as you'd need to say <PackageDependency ((Name="KittyVideos1" MinVersion="1.2.3.4") OR (Name="KittyVideos2" MinVersion="0.0.0.0") OR (Name="KittyVideos3" MinVersion="0.0.0.0"))> (or something equivalent).

[...]

PR #597, opened 2021-03-17
Use custom versioning scheme

kythant:
Use custom versioning scheme

### DIFF
--- a/build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
+++ b/build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
@@ -10,7 +10,15 @@ parameters:
   useReleaseTag: 'false'
   prereleaseVersionTag: 'prerelease'
   condition: ''
-
+variables:
+  # Define the components of the version number. SemVer rules should be used to determine when to update.
+  versionDate: $[format('{0:yyyyMMdd}', pipeline.startTime)]
+  versionCounter: $[counter(variables['versionDate'], 0)]
+  major: '0'
+  minor: '5'
+  patch: '0'
+  versionTag: 'CI'
+  versionFinal: $[format('{0}.{1}.{2}-{3}.{4}.{5}', variables['major'], variables['minor'], variables['patch'], variables['versionTag'], variables['versionDate'], variables['versionCounter'])]
 jobs:
 - job: ${{ parameters.jobName }}
   dependsOn:
@@ -84,9 +92,8 @@ jobs:
     inputs:
       targetType: 'inline'
       script: |
-        $version = '$(versionFinal)'
         [xml]$publicNuspec = Get-Content -Path $(Build.SourcesDirectory)\build\NuSpecs\Microsoft.ProjectReunion.Foundation.nuspec
-        $publicNuspec.package.metadata.version = $version
+        $publicNuspec.package.metadata.version = '$(versionFinal)'
         Set-Content -Value $publicNuspec.OuterXml $(Build.SourcesDirectory)\build\NuSpecs\Microsoft.ProjectReunion.Foundation.nuspec
 
   - task: 333b11bd-d341-40d9-afcf-b32d5ce6f23b@2
--- a/build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
+++ b/build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
@@ -17,7 +17,7 @@ parameters:
   minor: '5'
   patch: '0'
   versionTag: 'CI'
-  versionFinal: $[format('{0}.{1}.{2}-{3}.{4}.{5}', ${{ parameters.major }}, ${{ parameters.minor }}, ${{ parameters.patch }}, ${{ parameters.versionTag }}, ${{ parameters.versionDate }}, ${{ parameters.versionCounter }})]
+
 jobs:
 - job: ${{ parameters.jobName }}
   dependsOn:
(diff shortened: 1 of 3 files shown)


riverar (approved):
Intent looks OK. Perhaps in the future, you could wire up something like [GitVersion](https://gitversion.net) that'll do all this work for you, using tags/branches as the single source of truth.

BenJKuhn:
> Intent looks OK. Perhaps in the future, you could wire up something like [GitVersion](https://gitversion.net) that'll do all this work for you, using tags/branches as the single source of truth.

Interesting suggestion, we may need to look into something like that at some point. Thanks for the pointer.

build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
```diff
@@ -82,18 +85,25 @@ jobs:
   #       -BuildFlavor ${{ parameters.buildFlavor }}
 
   #   displayName: 'build-nupkg.ps1'
+  - task: PowerShell@2
+    displayName: UpdateFoundationNuSpecVersion
+    inputs:
+      targetType: 'inline'
+      script: |
+        $versionFinal = '${{ parameters.major }}.${{ parameters.minor }}.${{ parameters.patch }}-${{ parameters.versionTag }}.${{ parameters.versionDate }}.${{ parameters.versionCounter }}'
```
BenJKuhn, line 93: do you enver need to worry abou thte "tag/date/counter" empty case (e.g. "release" versioning scheme?)

BenJKuhn (approved):

build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
kythant: I was just planning on setting it to “release” or “official” if that’s okay :)

riverar (changes requested):
No, you cannot keep the prerelease identifier there. `1.0.0-release` would be seen as a prerelease, with the next expected major release to be `1.0.0`.

kythant:
@riverar which prerelease identifier are you talking about. I believe I don't have any?

riverar:
$versionFinal = '${{ parameters.major }}.${{ parameters.minor }}.${{ parameters.patch }}-${{ parameters.versionTag }}.${{ parameters.versionDate }}.${{ parameters.versionCounter }}'

@kythant During review, you commented that you'd set `versionTag` to a value of `release`. This is actually called a pre-release identifier and must follow the rules above. 1.0, for example, cannot be provided as 1.0-release.

kythant:
@riverar 
Gotcha. So for the official release, the tag has to be empty?

riverar:
@kythant Yup, if we're abiding by semver rules. Otherwise downstream clients that are ignoring prelease bits will overlook something labeled `1.0.0-release` or `1.0.0-official`.

build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
```diff
@@ -82,18 +85,28 @@ jobs:
...
+  - task: PowerShell@2
+    displayName: UpdateFoundationNuSpecVersion
+    inputs:
+      targetType: 'inline'
+      script: |
+        $versionFinal = '${{ parameters.major }}.${{ parameters.minor }}.${{ parameters.patch }}.${{ parameters.versionDate }}.${{ parameters.versionCounter }}'
+        if(-not [String]::IsNullOrEmpty('${{ parameters.versionTag }}'))
+        {
+          $versionFinal = '${{ parameters.major }}.${{ parameters.minor }}.${{ parameters.patch }}-${{ parameters.versionTag }}.${{ parameters.versionDate }}.${{ parameters.versionCounter }}'
+        }
```
riverar, line 97: So let's walk through this together with some test values. May want to have https://semver.org/ open as well.

Let's pretend we're releasing 1.0 to the world today, and your internal counter is zeroed.

$versionFinal becomes `1.0.0.20210318.0`. SemVer 2.0 states the version number should be in MAJOR.MINOR.PATCH form. So this is invalid.

It appears you're trying to add _build metadata_ here, so let's look at the spec:
> Build metadata MAY be denoted by appending a plus sign and a series of dot separated identifiers immediately following the patch or pre-release version.

Cool. Some examples:
* 1.0.0 (major release, no metadata, no pre-release identifiers)
* 1.0.0+0000.0 (major release with build metadata that does not influence decision making)
* 1.0.0-preview+20210318.1 (pre-release with build metadata [...])
* 1.0.0-preview+20210318.102 (pre-release with build metadata [...])

riverar: (cont)

Be careful though. It may be tempting to represent everything as _build metadata_ but lets look at the spec again.
> Build metadata MUST be ignored when determining version precedence. Thus two versions that differ only in the build metadata, have the same precedence.

That means these are ranked the same (i.e. the client running .1 may not see .102 as different and not pull it down):
* 1.0.0-CI+20210318.1 (pre-release with build metadata [...])
* 1.0.0-CI+20210318.102 (pre-release with build metadata [...])

So you're going to need a hybrid approach here of pre-release identifiers and build metadata.

DrusTheAxe:
Where are the Version Requirements / Goals defined, and how those map to our various tech (SemVer, NuGet, VISX, MSIX, MS Store, ...)?

These all support an overlapping set of versioning behavior but none of them are quite the _same_ behavior. For instance, how does 1.0.0-preview+20210318.102 map to MSIX? Not in any "intuitively obvious to the most casual of observers" way. VISX and others have similar overlapping-but-different impedance mismatch issues.

It's hard to say if a solution looks good if one doesn't know the target.

P.S. Are you familiar with [Enhance MSIX to support Semantic Versioning per http://semver.org #148](https://github.com/microsoft/ProjectReunion/issues/148)?

mrlacey:
Echoing what @DrusTheAxe says above, where are the plans and details for versioning of Project Reunion (Also asked as part of  #582)

BenJKuhn:
@kythant, I think this is abandoned & no longer needed at this point based on other changes you've been making, but feel free to re-open if this change (or an updated version of it) is still required.

PR #600, opened 2021-03-18
Enable running BinSkim on Foundation

DefaultRyan:
This change adds a build task to enable [BinSkim](https://github.com/Microsoft/binskim) (a security-oriented analyzer) on our build pipeline, as well as fixes a few project settings to address binskim errors:

- Enable [Control Flow Guard](https://docs.microsoft.com/en-us/cpp/build/reference/guard-enable-control-flow-guard?view=msvc-160)
- Enable [Buffer Security Check](https://docs.microsoft.com/en-us/cpp/build/reference/gs-buffer-security-check?view=msvc-160) for Detours project
- Upgrade [source file hash algorithm in debug info](https://docs.microsoft.com/en-us/cpp/build/reference/zh?view=msvc-160) from MD5 to SHA-256.


DefaultRyan:
We'll start off by enabling running BinSkim in each feeder's repo locally. Once those all pass, we'll start running BinSkim in the aggregation pipeline.

dev/ProjectReunion_DLL/ProjectReunion_DLL.vcxproj
```diff
@@ -258,6 +253,7 @@
       <ConformanceMode>true</ConformanceMode>
       <PrecompiledHeaderFile>pch.h</PrecompiledHeaderFile>
       <AdditionalOptions>/ZH:SHA_256 %(AdditionalOptions)</AdditionalOptions>
+      <ControlFlowGuard>Guard</ControlFlowGuard>
```
jonwis, line 256: The other thread about linker configuration settled on a "we should do this for the build pipeline but not developer desktop runs."  Does CFG add cost or build time, or prevent incremental linking?

DefaultRyan: CFG doesn't affect incremental linking. It is, however, incompatible with /Zi (but not /ZI) which is why I didn't turn it on for Debug builds.

I built with CFG on and off several times with Visual Studio's build timing enabled, and I couldn't discern a meaningful difference. CFG adds a couple intructions around indirect calls (like virtual dispatch, function pointers), so it added 5 kB to Microsoft.ProjectReunion.dll (385 kB vs 380 kB). /ZH:SHA_256 may have added a **tiny** amount of compile time, but it was so low as to be nearly indistinguishable from random noise (~100ms out of ~16 seconds).

The impact on inner-loop dev experience seems to be somewhere between "negligible" and "tiny", so it's worth keeping the pipeline and local build settings in sync.

DrusTheAxe (changes requested):
You missed dev\ProjectReunion_BootstrapDLL\ProjectReunion_BootstrapDLL.vcxproj

dev/Detours/Detours.vcxproj
```diff
@@ -250,7 +250,7 @@
       <FunctionLevelLinking>true</FunctionLevelLinking>
       <RemoveUnreferencedCodeData>false</RemoveUnreferencedCodeData>
       <OmitDefaultLibName>true</OmitDefaultLibName>
-      <BufferSecurityCheck>false</BufferSecurityCheck>
+      <AdditionalOptions>/ZH:SHA_256 %(AdditionalOptions)</AdditionalOptions>
```
DrusTheAxe, line 253: Can these changes follow the same advice as Jon gave [here](https://github.com/microsoft/ProjectReunion/pull/553/files#r593628407)?

>Please change this to use a model like:
>
>```
><AdditionalOptions>$(AdditionalPipelineLinkerOptions) %(AdditionalOptions)</AdditionalOptions>
>```
>
>Where that "additional pipeline" value is set by the ADO build pipeline as necessary. This lets us configure the linker commands during official build runs but not impact the individual developer's output.

Expect developers to add more *.vcxproj. Any common properties we want set should be handled in a similar way to better scale as the work (and #devs) grows.

DrusTheAxe (changes requested):

dev/Detours/Detours.vcxproj
BenJKuhn: binskim is supposed to validate what we ship, which should be the default configuration for the build. This is different from some of the needs we have for one-off customized scanning builds.

DrusTheAxe: @BenJKuhn I'm confused. Or maybe I need more coffee

Are you agreeing this should be handled via `$(blah)` like @jonwis advocated?
Or no `/ZH:blah` etc here is the better answer?

riverar: I think in this case, you want to push the better hashing practices here onto all devs (instead of injecting during CI time). This also makes it one baby step easier to check off some hashing algorithm use compliance checkboxes. (We deal with this stuff in government sector.)

DefaultRyan: Agreed. We should strive to keep build configs identical between inner loop and pipeline builds as much as possible. Provided we don't impose an undue burden on inner loop developers. I don't think any of the options here impose such a burden.

Your comment about adding more projects and developers, though, makes me think that we should consider collecting some of these build configurations into a props/targets file. Of course, then we're paying a concept tax by moving some of these build definitions out of the local vcxproj files, so that devs have to go hunting to figure out why/where certain build options have been set. (Also, I'm not sure how we set debug vs release options in a props/targets file.)

DefaultRyan:
> You missed dev\ProjectReunion_BootstrapDLL\ProjectReunion_BootstrapDLL.vcxproj

Is that a binary we ship? I don't see it in the published artifcats in the pipeline build. I'm not figuring out how that binary fits into the overall picture.

DrusTheAxe:
> Is that a binary we ship? I don't see it in the published artifcats in the pipeline build

For 0.5? No. But then again Microsoft.ProjectReunion.dll isn't a feature for 0.5.0 either yet we're shipping it (though arguably we shouldn't have). Same goes for Detours.

For 0.8? Absolutely. Nothing under /dev today is fake or dead code. `DIR /s dev\*.vcxproj` and there's 3 sets of projects

1. dev\VSIX
2. dev\MRTCore
3. dev\*   <= i.e. everything else in ProjectReunion

```
dev\Detours\Detours.vcxproj
dev\DynamicDependencyDataStore\DynamicDependency.DataStore\DynamicDependency.DataStore.vcxproj
dev\DynamicDependencyDataStore\DynamicDependency.DataStore.ProxyStub\DynamicDependency.DataStore.ProxyStub.vcxproj
dev\DynamicDependencyLifetimeManager\DynamicDependencyLifetimeManager\DynamicDependencyLifetimeManager.vcxproj
dev\DynamicDependencyLifetimeManager\DynamicDependencyLifetimeManager.ProxyStub\DynamicDependencyLifetimeManager.ProxyStub.vcxproj
dev\DynamicDependencyLifetimeManager\IDynamicDependencyLifetimeManager\IDynamicDependencyLifetimeManager.vcxproj
dev\ProjectReunion_BootstrapDLL\ProjectReunion_BootstrapDLL.vcxproj
dev\ProjectReunion_DLL\ProjectReunion_DLL.vcxproj
```

8 projects that produce artifacts shipping in 0.8. Different types of projects -- static lib (Detours), dll (ProxyStubs, Bootstrap, Reunion), exe (DataStore, LifetimeManager) -- but all C++. BinSkim, /ZH and other changes should be equally applied to all.

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/src/Microsoft.ApplicationModel.Resources.vcxproj
```diff
@@ -125,11 +126,17 @@
   <ItemDefinitionGroup Condition="'$(Configuration)'=='Debug'">
     <ClCompile>
       <PreprocessorDefinitions>_DEBUG;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <ControlFlowGuard Condition="'$(Configuration)|$(Platform)'=='Debug|ARM64'">false</ControlFlowGuard>
```
jonwis, line 129: Shouldn't these be covered by the `.props` file from above?

dev/MRTCore/mrt/mrm/mrmmin/mrmmin.vcxproj
```diff
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="..\..\..\..\ProjectReunion.Build.Cpp.props" />
```
jonwis, line 3: Is there an msbuild property that represents "the solution root" so you could say `$(that)\dev\ProjectReunion.Build.Cpp.props` ?

jonwis (approved):

BenJKuhn (approved):

dev/MRTCore/mrt/mrm/mrmmin/mrmmin.vcxproj
DrusTheAxe: `$(SolutionDir)`

[Common macros for MSBuild commands and properties](https://docs.microsoft.com/cpp/build/reference/common-macros-for-build-commands-and-properties?view=msvc-160) is your friend :-)

riverar: You don't typically want to use solution relative properties because solutions can be created externally and owned by the user. Your projects may, for example, be added to a user's grand build solution that lives outside expected paths and now these paths will break.

PR #615, opened 2021-03-19
[Prototype] Implement CreateChannelAsync to request for channelURI from WNS

sharath2727:
Issue:
As part of the Reunion effort, introduce retry channel request logic within the CreateChannelAsync API to be able to handle benign channel request failures on behalf of the developer.

Description:
As of today, we believe when we fail to fetch channel requests due to errors like ERROR_TIMEOUT etc.., we are throwing exceptions right away and propagating back to the user. These are benign errors and gives us the flexibility to retry channel requests and avoid throwing exceptions for such cases. So for reunion implement a retry logic to request channels on behalf of app developers when we see mentioned errors. This will reduce the number of exceptions we propagate back to the app developers and improves the success rate for channel requests for the caller.


dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -0,0 +1,141 @@
...
+        // Allow to register the progress and complete handler
+        co_await resume_background();
+
+        auto progress{ co_await winrt::get_progress_token() };
+
+        bool remoteIdPresent = false;
+        {
+            auto lock = s_lock.lock();
+            auto it = std::find(s_remoteIdList.begin(), s_remoteIdList.end(), remoteId);
+            remoteIdPresent = (it != s_remoteIdList.end()) ? true : false;
```
David-Quero, line 70: You don't need "? true : false", the comparison already returns a boolean.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -0,0 +1,141 @@
...
+        }
+
+        winrt::Microsoft::ProjectReunion::PushNotificationCreateChannelResult channelResult{ nullptr };
+
+        if (remoteIdPresent)
+        {
+            channelResult = make<Microsoft::ProjectReunion::implementation::PushNotificationCreateChannelResult>(
+                nullptr, WPN_E_OUTSTANDING_CHANNEL_REQUEST, PushNotificationChannelStatus::CompletedFailure);
```
David-Quero, line 78: Is WPN_E_OUTSTANDING_CHANNEL_REQUEST the right hresult? It seems like this condition can happen if an application that already has a valid channel request a channel again so there is not an outstanding request but a channel already.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -0,0 +1,141 @@
...
+        }
+
+        {
+            auto lock = s_lock.lock();
+            s_remoteIdList.push_back(remoteId);
+        }
+
+        uint8_t retryCount = 0;
+        Microsoft::ProjectReunion::PushNotificationCreateChannelStatus
+            channelStatus = { E_PENDING /* ExtendedError */, PushNotificationChannelStatus::InProgress /* Status */, retryCount /* RetryCount */ };
```
David-Quero, line 89: Remove verbose comments.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -0,0 +1,141 @@
...
+        static std::vector<winrt::guid> s_remoteIdList;
+        static wil::critical_section s_lock;
+
+        // NOTE: API supports channel requests only for packaged applications
+        IsPackagedProcess();
+
+        check_pointer(&remoteId);
+
+        // Allow to register the progress and complete handler
+        co_await resume_background();
```
David-Quero, line 62: Where is resume_background defined? What does it does?

sharath2727: https://docs.microsoft.com/en-us/uwp/cpp-ref-for-winrt/resume-background - it does two things here.
1. It returns control to the caller allowing the caller to register progress and completion handler
2. It continues with the execution of the API immediately on a thread pool thread.

dev/PushNotifications/PushNotificationCreateChannelResult.cpp
```diff
@@ -0,0 +1,25 @@
+﻿#include "pch.h"
+#include "PushNotificationCreateChannelResult.h"
+#include "PushNotificationCreateChannelResult.g.cpp"
+
+namespace winrt::Microsoft::ProjectReunion::implementation
+{
+    PushNotificationCreateChannelResult::PushNotificationCreateChannelResult(Microsoft::ProjectReunion::PushNotificationChannel const& channel, winrt::hresult const& extendedError, Microsoft::ProjectReunion::PushNotificationChannelStatus const& status)
```
hulumane, line 7: If this fails, would the channel be nullptr in the projection? Have we tested that case? I worry because this is a pass by ref and doesn't seem like it handles null channels.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    inline bool IsPackagedProcess()
+    {
+        uint32_t packageFullNameLength = 0;
+        const auto rc{ GetCurrentPackageFullName(&packageFullNameLength, nullptr) };
+
+        if (rc != ERROR_INSUFFICIENT_BUFFER)
+        {
+            winrt::throw_hresult(rc);
```
hulumane, line 46: This doesnt look right. The method says it returns true or false. But here it throws an exception. Are we saying that we get InsufficientBuffer if the process is Packaged? I'm thinking this API needs to be changed.

hulumane: The API definition should be as follows: returns true if packaged, else false. Only throw for critical errors.

sharath2727: GetCurrentPackageFullName(&packageFullNameLength, nullptr) - the API here takes a nullptr as the buffer parameter. So we see ERROR_INSUFFICIENT_BUFFER it means that the process is packaged but we couldn't place the packagefullname as there is no buffer pointer passed to it. For all other errors where we either couldn't validate it to be a packaged process or we have seen infra problems so we throw the hresult and return the control back. Thanks.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+    Windows::Foundation::IAsyncOperationWithProgress<Microsoft::ProjectReunion::PushNotificationCreateChannelResult, Microsoft::ProjectReunion::PushNotificationCreateChannelStatus> PushNotificationManager::CreateChannelAsync(winrt::guid remoteId)
+    {
+        static std::vector<winrt::guid> s_remoteIdList;
+        static wil::critical_section s_lock;
+
+        // NOTE: API supports channel requests only for packaged applications
+        IsPackagedProcess();
+
+        check_pointer(&remoteId);
```
hulumane, line 59: I would flip this to be the very first check.

hulumane: Also GUID is not a pointer? So not sure how this would work. I think you meant to check for GUID_NULL?

hulumane: Good unit test should catch all these special cases

dev/PushNotifications/PushNotificationManager.cpp
hulumane: I'm not sure how this is possible. Then the check below would always be true because this API would throw when the app isn't packaged.

dev/PushNotifications/PushNotificationManager.cpp
sharath2727: @hulumane if it is not a packaged application why would we want to verify the validity of remoteId. In fact we should not be concerned about it right.

hulumane: Arguments always need to be validated in the beginning of the function or API call. 

RemoteId will be used for all Cobalt+ builds by calling into remote internal APIs irrespective of packaged or not. If the API does not exist in the SKU, we use the logic that we have below.

dev/PushNotifications/PushNotificationManager.cpp
hulumane: Yes. Its redundant

dev/PushNotifications/PushNotificationManager.cpp
David-Quero: PushNotificationChannelStatus::InProgress is already more descriptive than "Status"

dev/PushNotifications/PushNotificationManager.cpp
danielayala94: I just read the [doc about GetCurrentPackageFullName()](https://docs.microsoft.com/en-us/windows/win32/api/appmodel/nf-appmodel-getcurrentpackagefullname). The idea of this logic is fine, I'd just change the last return statement to simply `return true`.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -0,0 +1,141 @@
...
+            return true;
+        default:
+            return false;
+        }
+    }
+
+    inline bool IsPackagedProcess()
+    {
+        uint32_t packageFullNameLength = 0;
+        const auto rc{ GetCurrentPackageFullName(&packageFullNameLength, nullptr) };
```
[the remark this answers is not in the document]
sharath2727: Sure. Addressed.

dev/PushNotifications/PushNotificationManager.cpp
sharath2727: Sure I will take the comments off then.

dev/PushNotifications/PushNotificationCreateChannelResult.cpp
sharath2727: I need to understand this question better. I think what you are suggesting is in cases like:

1. if we fail to allocate ProjectReunion channel where we are creating it in the PushNotificationManager.cpp.
2. if we fail to fetch the PushNotification channel from across the ABI due to some exception

Then in both the cases, we catch the exception to evaluate if we can retry this, if we fail to retry then we prepare channelResult with CompletedFailure status and pass nullptr.

channelResult = winrt::make<winrt::Microsoft::ProjectReunion::implementation::PushNotificationCreateChannelResult>(nullptr, ex.code(), PushNotificationChannelStatus::CompletedFailure);
 
So yes, the projection will be nullptr and the user should only access the channel projection after checking the status. Below is an example, 

auto result = sender.GetResults();
                if (result.Status() == PushNotificationChannelStatus::CompletedSuccess)
                {
                    auto channelUri = result.Channel().Uri();
                    auto channelExpiry = result.Channel().ExpirationTime();

                    //sendRequestToServer(channelUri);
                    std::cout << "channelUri: " << winrt::to_string(channelUri.ToString()) << std::endl;
                    // Persist the channelUri and Expiry in the App Service
                }
                else if (result.Status() == PushNotificationChannelStatus::CompletedFailure)
                {
                       std::cout << "Channel is nullptr" << std::endl;
                 }

dev/PushNotifications/PushNotificationManager.cpp
sharath2727: Based on the discussion with the team, I have addressed it this way:

IsPackagedProcess() API will:
1.  return true is packaged,
2.  return false if unpackaged - caller of the above API will throw if needed
3.  will throw for all other critical errors.

Thanks.

dev/PushNotifications/PushNotificationChannel.h
```diff
@@ -0,0 +1,24 @@
+﻿#pragma once
+#include "PushNotificationChannel.g.h"
+
+namespace winrt::Microsoft::ProjectReunion::implementation
+{
+    struct PushNotificationChannel : PushNotificationChannelT<PushNotificationChannel>
+    {
+        PushNotificationChannel() = default;
+
+        PushNotificationChannel(Windows::Networking::PushNotifications::PushNotificationChannel const& channel);
```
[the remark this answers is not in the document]
sharath2727: we are taking the reference to the object that was allocated on the heap in PushNotificationManager.cpp. Any particular concerns?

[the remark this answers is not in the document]
David-Quero: Why isn't & enough?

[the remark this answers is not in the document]
sharath2727: It should be enough according to me :). It is auto generated david when we introduce the .idl. So I let it be

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -0,0 +1,182 @@
...
+            {
+                s_remoteIdInProgress = (!remoteIdInProgress);
+            }
+            else
+            {
+                remoteIdInProgress = true;
+            }
+        }
+
+        winrt::Microsoft::Windows::PushNotifications::PushNotificationCreateChannelResult channelResult{ nullptr };
```
David-Quero, line 91: Is there a way to make a nullptr constructor with a no throw guarantee?

dev/PushNotifications/PushNotifications.idl
```diff
@@ -1,38 +1,63 @@
...
 
-namespace Microsoft.ProjectReunion
+namespace Microsoft.Windows.PushNotifications
 {
-    enum ChannelStatus
+    enum PushNotificationChannelStatus
     {
         InProgress, // The request is in progress and there is no retry operation
-        InProgressRetry, // The request is in progress and is in a backoff retry state
+        InProgressRetry, // The request is in progress and is in a backoff retry state. Check ExtendedError for HRESULT for retryable error.
```
David-Quero, line 9: What do you thing of adding another state: WaitingForRetry?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -0,0 +1,182 @@
...
+        else if(status == PushNotificationChannelStatus::CompletedFailure)
+        {
+            channelResult = winrt::make<winrt::Microsoft::Windows::PushNotifications::implementation::PushNotificationCreateChannelResult>(
+                nullptr,
+                channelRequestResult,
+                status);
+        }
+
+        co_return channelResult;
+    }
```
David-Quero, line 181: I assume the progress token is automatically updated when you return from the function, is that correct?

dev/PushNotifications/PushNotifications.idl
sharath2727: DId not get the context here? Can you summarize here please?

dev/PushNotifications/PushNotificationManager.cpp
sharath2727: whenever you call progress(); function the progress status is updated. When you return from the function depending on the scenario we are at progress API may or may not be called. Thanks.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -0,0 +1,182 @@
...
+        {
+            throw_hresult(E_NOTIMPL);
+        }
+
+        bool remoteIdInProgress = false;
+        {
+            auto lock = s_lock.lock();
+            if (s_remoteIdInProgress == false)
+            {
+                s_remoteIdInProgress = (!remoteIdInProgress);
```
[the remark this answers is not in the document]
sharath2727: It is a point I wanted to bring forward on purpose that s_remoteIdInProgress and remoteIdInProgress are coupled. I would prefer to keep this as it is. We can discuss further offline if necessary. Thanks.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+        UINT32 packageFullNameLength = static_cast<UINT32>(ARRAYSIZE(packageFullName));
+
+        const auto packagedProcessHResult = ::GetCurrentPackageFullName(&packageFullNameLength, packageFullName);
+
+        if (packagedProcessHResult == APPMODEL_ERROR_NO_PACKAGE)
+        {
+            return false;
+        }
+        else if(packagedProcessHResult != S_OK)
```
hulumane, line 52: if (FAILED(packagedProcessHResult))

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    winrt::Windows::Foundation::IAsyncOperationWithProgress<Microsoft::Windows::PushNotifications::PushNotificationCreateChannelResult, Microsoft::Windows::PushNotifications::PushNotificationCreateChannelStatus> PushNotificationManager::CreateChannelAsync(winrt::guid remoteId)
+    {
+        static bool s_remoteIdInProgress;
+        static wil::critical_section s_lock;
+
+        if (remoteId == winrt::guid())
+        {
+            throw_hresult(E_INVALIDARG);
```
hulumane, line 67: is there a throw_hresult_if macro? If so, you can eliminate the if check here.

dev/PushNotifications/PushNotificationManager.cpp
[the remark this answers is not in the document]
hulumane: Agreed. This is very confusing. The global and locals are being mixed here.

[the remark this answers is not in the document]
hulumane: And if it is true, we should just return in this lock rather than use that local. Local isn'treally needed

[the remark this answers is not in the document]
sharath2727: Sure.

dev/PushNotifications/PushNotificationManager.cpp
sharath2727: there is nothing being allocated here for a throw to arise. It is just setting a variable to nullptr. that is it. Memory for it is being allocated at a later point. Thanks.

Issue #567, opened 2021-03-11
Dynamic Dependencies doesn't support Elevation


DrusTheAxe:
Dynamic Dependencies relies on PackagedCOM, but PackagedCOM types are not visible to elevated processes. Thus the Dynamic Depednencies and Bootstrapper APIs fail when called by an elevated process.

Elevated processes need to be supported.

DynDep uses of PackagedCOM:

1. **DataStore** - DynamicDependencies  API uses a PackagedCOM OOP Server to store persisted package dependency definitions in Project Reunion's Main (PRmain) package's `ApplicationData` if `MddCreatePackageDependencyOptions::ScopeIsSystem` is not specified (definitions are directly written to a location under `Program Data` if `ScopeIsSystem` is specified, and nothing's persisted if `MddPackageDependencyLifetimeKind::Process` is specified).
2. **LifetimeManager** - the LifetimeManager is a PackagedCOM OOP Server to indicate to the Deployment pipeline a specific framework package is in use (and thus do not remove it).

`MddCore::DataStore::GetDataStorePathForUser()` can be changed to get the path to PRmain's `ApplicationData` via `Windows.ApplicationModel.ApplicationDataManager` rather than asking the PackagedCOM if the process is elevated (or possibly in all cases).

The LifetimeManager isn't so easily solved. DynDep needs a process running with PRddlm's identity from the call to `MddBootstrapInitialize` until `MddBootstrapShutdown` (or process death). PackagedCOM provides a convenient solution to reliable lifetime management e.g. what if someone doesn't call `MddBootstrapShutdown`. An alternative mechanism is needed.

One option is a defining the DDLM process as an application with full-trust RuntimeBehavior=packagedClassicApp. ActivateApplication would be needed with a parameter (e.g. process id) so the process can `WaitOnSingleObject(OpenProcess(ProcessIdToHandle(args.pid)), INFINITE)` to detect if the `MddBootstrapInitialize` caller terminates without calling `MddBootstrapShutdown` and self-terminate.

ProcessId isn't a great answer as Windows can reuse them e.g. caller terminates and a new process created between `CreateProcess(ddlm)` and `OpenProcess(ProcesIdToHandle(args.pid))`. A better answer is a uniquely named IPC mechanism e.g. `CreateNamedPipe(name); CreateProcess(ddlm.exe, cmdline=name);`.

[...]

PR #617, opened 2021-03-19
Add temporary trap to block calling DynamicDependencies (or Bootstrap) APIs if Elevated

DrusTheAxe:
Adding a trap if DynDep|Bootstrap APIs are called from an Elevated process to prevent accidental misuse, false bug reports and bloodshed and tears all around.

This block will be removed when we address Issue #567 https://github.com/microsoft/ProjectReunion/issues/567

### DIFF
--- a/dev/DynamicDependency/MsixDynamicDependency.cpp
+++ b/dev/DynamicDependency/MsixDynamicDependency.cpp
@@ -23,10 +23,14 @@ bool IsStaticPackageGraphEmpty()
 // Temporary check to prevent accidental misuse and false bug reports until we address Issue #567 https://github.com/microsoft/ProjectReunion/issues/567
 bool IsElevated(HANDLE token = nullptr)
 {
-    wistd::unique_ptr<TOKEN_MANDATORY_LABEL> tokenMandatoryLabel;
-    FAIL_FAST_IF_FAILED(wil::get_token_information_nothrow(tokenMandatoryLabel, !token ? GetCurrentThreadEffectiveToken() : token));
+    wistd::unique_ptr<TOKEN_MANDATORY_LABEL> tokenMandatoryLabel{ wil::get_token_information_failfast<TOKEN_MANDATORY_LABEL>(!token ? GetCurrentThreadEffectiveToken() : token) };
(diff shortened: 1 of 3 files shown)


riverar (approved):

dev/DynamicDependency/MsixDynamicDependency.cpp
```diff
@@ -31,6 +40,9 @@ STDAPI MddTryCreatePackageDependency(
     MddCreatePackageDependencyOptions options,
     _Outptr_result_maybenull_ PWSTR* packageDependencyId) noexcept try
 {
+    // Dynamic Dependencies doesn't support elevation
+    FAIL_FAST_HR_IF_MSG(HRESULT_FROM_WIN32(ERROR_NOT_SUPPORTED), MddCore::IsElevated() || MddCore::IsElevated(GetCurrentProcessToken()), "DynamicDependencies doesn't support elevation. See Issue #567 https://github.com/microsoft/ProjectReunion/issues/567");
```
jonwis, line 44: Consider collapsing this into a helper method like `FailFastIfElevated()`  Copy-pasta usually suggests a method.

jonwis (approved):

dev/DynamicDependency/MsixDynamicDependency.cpp
```diff
@@ -19,6 +19,15 @@ bool IsStaticPackageGraphEmpty()
     (void) LOG_HR_IF_MSG(HRESULT_FROM_WIN32(rc), (rc != APPMODEL_ERROR_NO_PACKAGE) && (rc != ERROR_INSUFFICIENT_BUFFER), "GetCurrentStaticPackageInfo rc=%d", rc);
     return rc == APPMODEL_ERROR_NO_PACKAGE;
 }
+
+// Temporary check to prevent accidental misuse and false bug reports until we address Issue #567 https://github.com/microsoft/ProjectReunion/issues/567
+bool IsElevated(HANDLE token = nullptr)
+{
+    wistd::unique_ptr<TOKEN_MANDATORY_LABEL> tokenMandatoryLabel;
+    FAIL_FAST_IF_FAILED(wil::get_token_information_nothrow(tokenMandatoryLabel, !token ? GetCurrentThreadEffectiveToken() : token));
```
jonwis, line 27: Oh also, use `wil::get_token_information_failfast` and let it deal with it.

DrusTheAxe: Forgot about that one. Thanks!

Issue #559, opened 2021-03-11
Rewrite DynamicDependencies tests using TAEF


DrusTheAxe:
DynamicDependencies' tests are written using VS' CppUnitTestsFramework. That provides a subset of TAEF's functionality (no surprise as it was inspired by and derivative of TAEF) and has other limitations, e.g. doesn't support ARM64.

DrusTheAxe:
Completed

PR #620, opened 2021-03-19
DynamicDependencies: Only Detour functions in not-packaged processes

DrusTheAxe:
DynamicDependencies only supports not-packaged processes but the Detours hooks were always wired up. Fixed to skip skip that in a packaged process so Microsoft.ProjectReunion.dll can be used in packaged processes for other (not DynDep) features.

Also beefed up parameter validation in the Detour'd GetCurrentPackageInfo[2] to match the OS APIs' parameter validation.

Expanded test cases.

NOTE: Tests verifying package'd process behavior is blocked awaiting [Rewrite DynamicDependencies tests using TAEF #559](https://github.com/microsoft/ProjectReunion/issues/559)

### DIFF
--- a/dev/DynamicDependency/PackageGraphManager.cpp
+++ b/dev/DynamicDependency/PackageGraphManager.cpp
@@ -65,6 +65,12 @@ HRESULT MddCore::PackageGraphManager::GetCurrentPackageInfo2(
     UINT32* count,
     GetCurrentPackageInfo2Function getCurrentPackageInfo2) noexcept try
 {
+    // Check parameter(s) per GetCurrentPackageInfo2 compatibility
+    //   * bufferLength is required
+    //   * buffer is required if bufferLength >0
+    //   * count is required if buffer is not nullptr
+    RETURN_HR_IF(E_INVALIDARG, (bufferLength == nullptr) || (*bufferLength > 0 && buffer == nullptr) || ((count == nullptr) && (buffer != nullptr)));
+
     if (count)
     {
         *count = 0;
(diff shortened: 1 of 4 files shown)


dev/ProjectReunion_DLL/dllmain.cpp
```diff
@@ -8,8 +8,22 @@
 
 #include <../Detours/detours.h>
 
+static bool IsPackagedProcess()
```
aeloros, line 11: Isn't this like the 4th or 5th copy of this we've made?

dev/DynamicDependency/PackageGraphManager.cpp
```diff
@@ -65,6 +65,12 @@ HRESULT MddCore::PackageGraphManager::GetCurrentPackageInfo2(
     UINT32* count,
     GetCurrentPackageInfo2Function getCurrentPackageInfo2) noexcept try
 {
+    // Check parameter(s) per GetCurrentPackageInfo2 compatibility
+    //   * bufferLength is required
+    //   * buffer is required if bufferLength >0
+    //   * count is required if buffer is not nullptr
+    RETURN_HR_IF(E_INVALIDARG, (bufferLength == nullptr) || (*bufferLength > 0 && buffer == nullptr) || ((count == nullptr) && (buffer != nullptr)));
```
aeloros, line 72: Nit: Horizontal scrollbar when diffing :(  Maybe wrap it?

aeloros (approved):

dev/DynamicDependency/PackageGraphManager.cpp
DrusTheAxe: The maxlen=100 rule is for *.md (specs), not code.

dev/ProjectReunion_DLL/dllmain.cpp
DrusTheAxe: Lol 2nd or maybe 3rd. I'm planning to centralize this (and a few other things) but waiting on a couple of naming decisions first.

There's a laundry list of refactoring I've got my eye on. Planning to put some time in after I finish some critical path work.

PR #622, opened 2021-03-20
Remove unwanted code from project templates

wjk:
Inspired by https://github.com/microsoft/ProjectReunion/discussions/574.

### DIFF
--- a/dev/VSIX/ProjectTemplates/Desktop/CppWinRT/PackagedApp/BlankApp/MainWindow.h
+++ b/dev/VSIX/ProjectTemplates/Desktop/CppWinRT/PackagedApp/BlankApp/MainWindow.h
@@ -12,11 +12,6 @@ namespace winrt::$safeprojectname$::implementation
     struct MainWindow : MainWindowT<MainWindow>
     {
         MainWindow();
-
-        int32_t MyProperty();
-        void MyProperty(int32_t value);
-
-        void myButton_Click(Windows::Foundation::IInspectable const& sender, Microsoft::UI::Xaml::RoutedEventArgs const& args);
     };
 }
 
--- a/dev/VSIX/ProjectTemplates/UWP/CppWinRT/BlankApp/MainPage.h
+++ b/dev/VSIX/ProjectTemplates/UWP/CppWinRT/BlankApp/MainPage.h
@@ -7,11 +7,6 @@ namespace winrt::$safeprojectname$::implementation
     struct MainPage : MainPageT<MainPage>
     {
         MainPage();
-
-        int32_t MyProperty();
-        void MyProperty(int32_t value);
-
-        void myButton_Click(Windows::Foundation::IInspectable const& sender, Microsoft::UI::Xaml::RoutedEventArgs const& args);
     };
 }
 
--- a/dev/VSIX/ProjectTemplates/Desktop/CSharp/PackagedApp/BlankApp/MainWindow.xaml
+++ b/dev/VSIX/ProjectTemplates/Desktop/CSharp/PackagedApp/BlankApp/MainWindow.xaml
@@ -7,7 +7,7 @@
     xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
     mc:Ignorable="d">
 
-    <StackPanel Orientation="Horizontal" HorizontalAlignment="Center" VerticalAlignment="Center">
-        <Button x:Name="myButton" Click="myButton_Click">Click Me</Button>
-    </StackPanel>
+    <Grid>
+
+    </Grid>
 </Window>
--- a/dev/VSIX/ProjectTemplates/Desktop/CSharp/PackagedApp/BlankApp/MainWindow.xaml.cs
+++ b/dev/VSIX/ProjectTemplates/Desktop/CSharp/PackagedApp/BlankApp/MainWindow.xaml.cs
@@ -27,10 +27,5 @@ namespace $safeprojectname$
         {
             this.InitializeComponent();
         }
-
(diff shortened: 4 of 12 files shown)


dotMorten:
Love it, except I don't get the point of the button. I'd prefer that to be removed too. It's throw away code. Don't add throw-away to a starter template.

wjk:
The check failures regarding Helix appear to be because I’m a contributor outside of Microsoft, and as such do not have permission to kick off Helix runs (whatever they are), but the build tried anyway. We should take a look at this sometime.

mrlacey:
I see the arguments for making the changes, but I'd want to know the reasons behind adding that code originally before debating whether to remove it.

riverar:
Agreed, feels like the changes here are purely subjective. At the very least, this should go through the proper workflow for public discourse. We've skipped that step here.

dev/VSIX/ProjectTemplates/Desktop/CppWinRT/PackagedApp/BlankApp/MainWindow.h
```diff
@@ -15,8 +15,6 @@ namespace winrt::$safeprojectname$::implementation
 
         int32_t MyProperty();
         void MyProperty(int32_t value);
```
Jaiganeshkumaran, line 17: I think these properties should be removed. They are useless and everytime you create a new window, you're going to just get rid of it so no use.

dev/VSIX/ProjectTemplates/UWP/CppWinRT/BlankApp/MainPage.h
```diff
@@ -10,8 +10,6 @@ namespace winrt::$safeprojectname$::implementation
 
         int32_t MyProperty();
         void MyProperty(int32_t value);
```
Jaiganeshkumaran, line 12: These properties should also be removed

Jaiganeshkumaran (changes requested):

Jaiganeshkumaran:
The C++ version of item templates and project templates seem to nest the source file as top-level if it doesn't have a XAML file rather than the IDL or header file. This is inconsistent with current templates included in the C++/WinRT VSIX and others built-in to Visual Studio.

BenJKuhn:
@MikeHillberg, can you review this PR or find someone to help drive it to closure?

dotMorten:
Any update on this? The current template is driving me bonkers :-D

aepot:
Current templates confused me. Why I must clean-up  xaml and code-behind before everything?

I'm using MVVM+IoC patterns which means no code-behind is needed, except ViewModel property. StackPanel instead of Grid looks suspicious too. First thought was "no Grid support?".

As SDK user I like the changes and suggest to merge the PR.

Jaiganeshkumaran:
Templates now seem to include Microsoft copyright and MIT License notice? Any reason why this was done. It makes sense for generated files to be copyrighted, but templates shouldn't be (obviously you'll add your own code after creating an instance of the template and that's not copyrighted by Microsoft!).

bpulliam:
Closing PRs that appear to have been abandoned.

riverar:
Bob's closure of this PR is part of an slow moving effort to clean up the repository. This PR skipped the proper workflow (at the time) for public discourse/changes, has minor and subjective changes in it, and has had no real engagement since 2021. Seems like this fits the bill for closure to me.

If the author feels strongly about this change, I recommend opening an issue to discuss and get sign off before making template changes.

bpulliam:
@riverar is right, but I also should add more context.
1) This original intent of this PR is to change the templates in a pretty fundamental way - empty content vs single button content. This should be resolved in a discussion first, and then we can look at the PR.  I do see a significant number of bug reports that use the template to attach some functionality to the button, and those would have to create content instead of having other remove it.  These templates were based on C++ winrt templates so there is some history there.  Not saying we can't change, but we should be deliberate about it.
2) This also had become a catch-all for template issues.  We have addressed some of those in #3324 and #3493 already.
3) You are absolutely correct that this conversation should have happened a year ago.

I am happy to have the conversation about this, or any other issue we resolve or close, as the goal is to bring additional clarity on those issues.

JaiganeshKumaran:
> I do see a significant number of bug reports that use the template to attach some functionality to the button, and those would have to create content instead of having other remove it. These templates were based on C++ winrt templates so there is some history there. Not saying we can't change, but we should be deliberate about it.

Apart from bug reports, you are mostly going to delete them anyway. It's only wasting time for you. I did open a pull request to remove them from the C++/WinRT (UWP) templates but it was not merged due to not having enough people to main the VSIX. Windows App SDK, however, is well maintained and this change should be merged. Maybe as an example, the stack panel and button can be included in MainPage.xaml created with the project template, but not in item templates for page or user control.

riverar:
@JaiganeshKumaran "Well maintained" [spits out coffee].

JaiganeshKumaran:
@riverar  Sure is more well-maintained than C++/WinRT, which is done for now.

dotMorten:
> These templates were based on C++ winrt templates so there is some history there.

History? None of the WPF, UWP or WinForms templates ship with anything in them. They are blank and ready to throw content on. It's been like this for many many years.

But yes we should have an issue to discuss this in I guess. I was merely pointing out the messaging of having a PR discussed, then [assigned to a Microsoft team member](https://github.com/microsoft/WindowsAppSDK/pull/622#issuecomment-871112147) to follow it through, only to be closed years later after no one actually following through on it, and no explanation why it was closed besides "stale". It's a really bad messaging to send to community members who spent time trying to improve the SDK, and there's nothing the community could have done here to prevented it from going stale.

PR #632, opened 2021-03-23
Adding tests

huzaifa-d:


dev/PowerNotifications/ClosedSourceMock.cpp
```diff
@@ -0,0 +1,271 @@
...
+unordered_map<index, OnDischargeTimeChanged> onDischargeTimeChanged_callbacks;
+unordered_map<index, OnEnergySaverStatusChanged> onEnergySaverStatusChanged_callbacks;
+unordered_map<index, OnPowerConditionChanged> onPowerConditionChanged_callbacks;
+unordered_map<index, OnDisplayStatusChanged> onDisplayStatusChanged_callbacks;
+unordered_map<index, OnSystemIdleStatusChanged> onSystemIdleStatusChanged_callbacks;
+unordered_map<index, OnPowerSchemePersonalityChanged> onPowerSchemePersonalityChanged_callbacks;
+unordered_map<index, OnUserPresenceStatusChanged> onUserPresenceStatusChanged_callbacks;
+unordered_map<index, OnSystemAwayModeStatusChanged> onSystemAwayModeStatusChanged_callbacks;
+
+HRESULT GetCompositeBatteryStatus(CompositeBatteryStatus** compositeBatteryStatusOut)
```
brialmsft, line 23: The argument type here should probably have been CompositeBatteryStatus* instead. For a fixed-size struct outparam, we shouldn't need to have the implementation heap-allocate it; we should be able to have the caller allocate it and the implementation fill it. This pattern with CoTaskMemAlloc is more useful for dynamically-sized or opaque outparams, where the caller can't appropriately initialize an empty outparam for the implementation to fill.

dev/PowerNotifications/ClosedSourceMock.cpp
```diff
@@ -0,0 +1,271 @@
...
+
+HRESULT RegisterCompositeBatteryStatusChangedListener(OnCompositeBatteryStatusChanged listener, CompositeBatteryStatusRegistration* registration)
+{
+    *registration = reinterpret_cast<CompositeBatteryStatusRegistration>(g_idx++);
+    onCompositeBatteryStatusChanged_callbacks[g_idx] = listener;
+    std::thread thread([]() {
+        std::this_thread::sleep_for(std::chrono::milliseconds(50));
+        for (const auto& [key, callbackFn] : onCompositeBatteryStatusChanged_callbacks)
+        {
+            auto status = wil::make_unique_cotaskmem<CompositeBatteryStatus>();
```
brialmsft, line 44: With this and the status.release() in the call, you're transferring ownership of the CompositeBatteryStatus to callbackFn. That's not the conventional lifetime semantics for in-params. Normally an in-param is only valid for the duration of the call, and the callee must copy it if they want to persist it; they don't receive ownership, and aren't responsible for freeing the parameter. As-is, if the callee doesn't call CoTaskMemFree at some point (potentially after the call returns), this allocation will leak.

I'd recommend stack-allocating the CompositeBatteryStatus unless we've got a really good reason to break convention.

dev/PowerNotifications/ClosedSourceMock.cpp
```diff
@@ -0,0 +1,271 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
```
brialmsft, line 3: It's worth putting a comment here calling this out as a *temporary* test strategy, which is only in place until the actual cross-binary ABI is available for mocking. If this file ever ends up in the main branch, something has gone wrong...

nuget.config
```diff
@@ -4,6 +4,7 @@
     <clear />
     <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
     <add key="ProjectReunion-Dependencies" value="https://pkgs.dev.azure.com/ms/ProjectReunion/_packaging/ProjectReunion-Dependencies/nuget/v3/index.json" />
+    <add key="WindowsES-External" value="https://microsoft.pkgs.visualstudio.com/_packaging/WindowsES-External/nuget/v3/index.json" />
```
brialmsft, line 7: I'm a bit fuzzy on the ES details, but I think any new NuGet feeds should be configured in the ProjectReunion-Dependencies pipeline. You should follow up with Reunion ES folks though.

test/PowerNotificationsTests/framework.h
```diff
@@ -0,0 +1,7 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+#pragma once
+
+#define WIN32_LEAN_AND_MEAN             // Exclude rarely-used stuff from Windows headers
```
brialmsft, line 5: Is there a reason to have this separate from pch.h? It's not a problem, explicitly. Just not sure what the benefit of splitting off this header is.

nuget.config
brialmsft: Also, IIRC you were explicitly asked not to use WindowsES-External.

test/PowerNotificationsTests/PowerNotificationsTests.vcxproj
```diff
@@ -91,12 +91,13 @@
       <ConformanceMode>true</ConformanceMode>
       <PrecompiledHeader>Use</PrecompiledHeader>
       <PrecompiledHeaderFile>pch.h</PrecompiledHeaderFile>
-      <AdditionalIncludeDirectories>$(OutDir)\..\ProjectReunion_DLL</AdditionalIncludeDirectories>
+      <AdditionalIncludeDirectories>$(ProjectDir)..\..\packages\Microsoft.Taef.10.58.210222006-develop\build\Include;$(OutDir)\..\ProjectReunion_DLL</AdditionalIncludeDirectories>
```
brialmsft, line 94: This is fine for now, but isn't ideal (as I'm sure you're aware, since this issue is presumably the reason you were using the other NuGet feed for TAEF). Having the package version number copied around everywhere is obnoxious and adds friction for whenever we need to update the package version.

I believe that packages\Microsoft.Taef.Redist.Wlk.10.52.200221001\build\Taef.Redist.Wlk.targets is _supposed_ to auto-inject the appropriate properties for consuming TAEF into projects that consume the NuGet package. I'm not 100% sure why this doesn't work; possibly just because the it's supposed to be named **Microsoft**.Taef.Redist.Wlk.targets to match the NuGet package name.
IMO, we should follow up separately with Reunion ES and TAEF folks and get this fixed properly. The best option would be to fix the NuGet package. Barring that, we might still consider having a centralized .props file tracking the project's TAEF version (assuming we want to standardize which TAEF version is used; I guess that's non-essential, since all of the test binaries build independently of each other).

dev/PowerNotifications/PowerNotifications.h
```diff
@@ -666,7 +666,7 @@ namespace winrt::Microsoft::ProjectReunion::implementation
             return make_self<factory_implementation::PowerManager>()->EnergySaverStatusChanged_Callback(energySaverStatus);
         }
 
-        static void CompositeBatteryStatusChanged_Callback(CompositeBatteryStatus* compositeBatteryStatus)
+        static void CompositeBatteryStatusChanged_Callback(CompositeBatteryStatus compositeBatteryStatus)
```
brialmsft, line 669: Just because we're changing the type to CompositeBatteryStatus* for outparams doesn't mean it needs to be CompositeBatteryStatus for in-params. If you do that, you're copying the struct for the function calls. I'd say go with either CompositeBatteryStatus* or CompositeBatteryStatus& to avoid unnecessary copying.

Also: we should make a point of putting const on all of the places where in-params are being passed by reference/pointer.

test/PowerNotificationsTests/PowerNotificationsTests.vcxproj
huzaifa-d: Excellent point. I will follow up with them.

dev/PowerNotifications/PowerNotificationsPal.h
```diff
@@ -14,7 +14,7 @@ struct CompositeBatteryStatus
 
 DECLARE_HANDLE(CompositeBatteryStatusRegistration);
 HRESULT GetCompositeBatteryStatus(CompositeBatteryStatus* compositeBatteryStatusOut);
-typedef void (*OnCompositeBatteryStatusChanged)(CompositeBatteryStatus compositeBatteryStatus);
+typedef void (*OnCompositeBatteryStatusChanged)(CompositeBatteryStatus &compositeBatteryStatus);
```
brialmsft: Mixed feelings about this one. This is describing the public surface of the closed-source ABI, and we've historically made the OS's public Win32 APIs C-compatible. Not sure if we intend to follow the same principle for the Reunion closed-source ABI; if so, this would need to be a pointer and not a reference. @jonwis, thoughts?

Considering it non-blocking for this PR. If we do need to make this header/ABI C-compatible, we can clean up in a subsequent PR.

brialmsft (approved):

PR #642, opened 2021-03-24
Adding the ability to get an environment variable and set an environment variable

dhoehna:
This change includes the implementation and testing of
1. GetEnvironmentVariable
2. SetEnvironmentVariable
3. ChangeTracking

The Change Tracker allows a centennial app to track any environment variable changes in the user and machine scope.  These changes will be removed when the app is uninstalled.

A non-Centennial app can still use this API to set an environment variable however.


dhoehna:
Because of timing with other work this PR is going to be pulled into EVAndPath now.  However, please add all comments and I will address them in another PR.

PR #645, opened 2021-03-25
Optimize the size of the VSIX icon

mrlacey:
Use an image in the VSIX manifest of optimized size.

Previously using an image that is 200x200 pixels but the image is only ever rendered at 90x90, 48x48, and 64x64 pixels.

Moving to a smaller image to avoid unnecessary scaling and unnecessary size being added to the VSIX file.

I know it's only a small saving in one file (5.42Kb) but it's good to show attention to detail from the start of a project.
Also, even this small saving amounts to over half a gigabyte in saved download size if this package is downloaded 100,000 times. (I expect you hope it will eventually be downloaded many more.)


riverar:
This asset is also used as the VSIX `PreviewImage`, which requires this size https://docs.microsoft.com/en-us/visualstudio/extensibility/vsix-manifest-designer?view=vs-2019#uielement-list:~:text=The%20preview%20image%20must%20be%20200x200%20pixels

sylveon:
This PR also ignores high DPI scenarios, which would lead to blurry icons

BenJKuhn:
Now that the project has a real name, Windows App SDK, we'll be updating the logos to new assets at some point in the near future. We appreciate the suggestions for optimizing the images though.

Ben

PR #654, opened 2021-03-28
User/benkuhn/add vsix patch option

BenJKuhn:
Provide option for force a particular VSIX version during runs of the VSIX build pipeline. Default for local builds or pipeline builds will still use a version adapted from Project Reunion. I tried a few different ways of putting this into the YAML, and while I'm sure it's possible, in the end I found it a lot simpler to bake this into the project file instead, just passing the option through to msbuild.


BenJKuhn:
Note, this will be a situation of "review after commit" for logistical reasons, but @Scottj1s, I'd still greatly appreciate if you review this at your earliest convenience.

PR #682, opened 2021-04-01
Update AppLifecycle SingleMulti-Instancing.md

andreww-msft:
Removed the string parameter from UnregisterKey, since an instance can only register one key, the parameter is redundant.


jonwis:
Only supporting a single key for the process seems odd. The samples showed the "find the instance editing this document" example.  So an instance can only really be editing one document at a time? How would an instance register that it's the host for multiple documents?

The `FindOrRegisterForKey` doesn't make it super clear that it can only be called once.

andreww-msft:
An instance can register whatever string it likes for its key. If it has multiple files open, it can compose a string that includes all file names. Registering a composite string avoids the need for multiple keys, and leaves it up to the app to compose/parse its keys however it likes.

riverar:
> Removed the string parameter from UnregisterKey, since an instance can only register one key, the parameter is redundant.

Hm given this new information, does this API surface still make sense?

```diff
runtimeclass AppInstance
{
    static event EventHandler<AppActivationArguments> Activated;
    static Windows.Foundation.IVector<AppInstance> GetInstances();
    static AppInstance FindOrRegisterForKey(String key);
    void UnregisterKey();
   void RedirectActivationTo(AppActivationArguments args);

    String Key { get; };
    bool IsCurrent { get; };
    UInt32 ProcessId { get; };
}
```
I'm now wondering:
* Why expose a key property at all? Or the others?
* Why can I create multiple instances of AppInstance?

andreww-msft:
@jonwis We can consider adding the ability for an instance to register multiple keys in a later release, if it turns out a composite key isn't enough for some reason.
@riverar I'm not quite sure what you're getting at. The key is so that the app can stash some information about each instance - it's only relevant to the app itself, composed by the app, parsed by the app. It can put whatever it likes in that string - and interpret it in any way it likes. As to why multiple instances - there will be one AppInstance object for each running instance of the app. The AppInstance is the API that represents the app instance. The app itself does not create AppInstance objects, instead the system hands them out when the app requests them.

aeloros:
> > Removed the string parameter from UnregisterKey, since an instance can only register one key, the parameter is redundant.
> 
> Hm given this new information, does this API surface still make sense?
> 
> ```diff
> runtimeclass AppInstance
> {
>     static event EventHandler<AppActivationArguments> Activated;
>     static Windows.Foundation.IVector<AppInstance> GetInstances();
>     static AppInstance FindOrRegisterForKey(String key);
>     void UnregisterKey();
>    void RedirectActivationTo(AppActivationArguments args);
> 
>     String Key { get; };
>     bool IsCurrent { get; };
>     UInt32 ProcessId { get; };
> }
> ```
> 
> I'm now wondering:
> 
> * Why expose a key property at all? Or the others?

The key is a friendly identify defined by the app developer.  It could be used to say ‘this instance has a role’, and automatically encapsulates atomicity on assigning that role.  The synchronization is done for the app developer, as the guarantee is that only one instance can sign up for the key.  This could be used to implement single instanced behavior, coordinator/worker behavior, task based behavior, or any other design that requires coordination between instances on job assignment.

> * Why can I create multiple instances of AppInstance?

You aren't creating them at all.  You're accessing an object that represents one.  You create multiple instances of AppInstance by creating multiple instances of the app (CreateProcessEx()).

aeloros:
> Only supporting a single key for the process seems odd. The samples showed the "find the instance editing this document" example. So an instance can only really be editing one document at a time? How would an instance register that it's the host for multiple documents?
> 
> The `FindOrRegisterForKey` doesn't make it super clear that it can only be called once.

How is that not super clear?  I think the 'FindOrRegister' pattern of naming makes it clear there are only two paths of success for the API call.  Either find an existing registered instance, or register the current instance.

I agree it would be nice to have multiple key support, and we talked about it.  We felt it was a bit of a stretch for V1, and since the existing platform APIs showed precedence we considered it an improvement on the design.  It just felt like it was acceptable to make that improvement a V2 feature.

aeloros (approved):

BenJKuhn:
It looks like this largely came in on another pull request. The remaining change is just a minor wording tweak. Cancelling this request.

PR #683, opened 2021-04-02
Fix image not being displayed no main readme

mrlacey:
I don't know why, but the locale specific redirect wasn't working on the first image so specifying the `en-us` version as that's better than the image not displaying at all.

This is how it looked before:

partial screenshot showing image missing

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 | Delight users with modern UI | Access new platform features | Backwards compatible | 
 |:--:|:--:|:--:|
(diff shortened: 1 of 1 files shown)


riverar:
Not sure this is going to fix it.
image

I suspect there's something up with GitHub as these URLs aren't actually in the resulting DOM at render time.
```html
<img src="https://camo.githubusercontent.com/132193443d96bcf0d0f499f4eb5e72eb35e652809c50255ada34769f55e9e6be/68747470733a2f2f646f63732e6d6963726f736f66742e636f6d2f6d656469612f696c6c757374726174696f6e732f62697a74616c6b2d6765742d737461727465642d6765742d737461727465642e737667" width="250" alt="Modern navigation" data-canonical-src="https://docs.microsoft.com/media/illustrations/biztalk-get-started-get-started.svg" style="max-width:100%;">
```

mrlacey:
Based on the observations of @riverar and the fact I no longer see a broken image, the issue seems to be an issue with how GitHub handles the redirected svg images. I'll close this and open another PR with a better fix.

Issue #448, opened 2021-02-13
Ensure consistent project and nuget versions across repo


jonwis:
> Should be 16.0?
> MRT's *.vxcproj say 15.0 but everything else in Reunion says 16.0

_Originally posted by @DrusTheAxe in https://github.com/microsoft/ProjectReunion/pull/444#discussion_r575502195_

The ES should have some sort of check that ensures we're being consistent in the tools we're using.

DrusTheAxe:
See Incorrect WindowsTargetPlatform[Min]Version in *.vcxproj #445 for more details

EHO-makai:
@huichen123, do you have an issue to address the VS version in your project file?

EHO-makai:
Assigning to @BenJKuhn to get into the ES backlog.

huichen123:
It should be ok to update MRTCore VS version and WindowsTargetPlatform[Min]Version.

BenJKuhn:
There are a number of different "version" aspects that this could cover. Most version considerations have been either consolidated, simplified away, or moved to management by Maestro at this point, so I'm closing out this issue.

PR #689, opened 2021-04-03
cherry pick: Move MrtCore up to cswinrt 1.2 to keep in sync with WinUI 

Scottj1s:


mrlacey:
Is there a process in place to automate the synchronization of dependencies like this?
If not, there should be. 😉

jonwis:
We have #448 that needs fixing as well.

jonwis (approved):

PR #696, opened 2021-04-05
Clean up of Microsoft.ApplicationModel.Resources.PriGen.targets (round 2)

rohanp-msft:
Remove some more unused Appx* properties and items in Microsoft.ApplicationModel.Resources.PriGen.targets.

Microsoft.ApplicationModel.Resources.PriGen.targets was created from Microsoft.AppxPackage.targets where a bunch of appx generation MSBuild script code was removed (Microsoft.AppxPackage.targets does more than just PRI generation, which is all that we want Microsoft.ApplicationModel.Resources.PriGen.targets to do, as the name suggests). At that point though, not all non-PRI gen related MSBuild script code was removed. I'll now try to remove it over several rounds. This is round 2.

Round 1: https://github.com/microsoft/ProjectReunion/pull/664

**How verified**
Used the MRTCore CPP sample app. Got rid of build outputs. Replaced the prigen targets file in the nuget package install with the one in this change. Built and ran the app.

### DIFF
--- a/dev/MRTCore/packaging/Microsoft.ApplicationModel.Resources.PriGen.targets
+++ b/dev/MRTCore/packaging/Microsoft.ApplicationModel.Resources.PriGen.targets
@@ -1,13 +1,6 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003" ToolsVersion="15.0">
 
-  <!-- ========================================================================== -->
-  <!-- Note: To test whether schema changes are OK, copy changed XSD file to      -->
-  <!-- %ProgramFiles(x86)%\Microsoft Visual Studio 15.0\Xml\Schemas\1033\MSBuild, -->
-  <!-- then open this file in Visual Studio.                                      -->
-  <!-- It should not display any schema warnings in error list window.            -->
-  <!-- ========================================================================== -->
-
   <!-- With .NET 5, the propery TargetPlatformIdentifier is not UAP but rather windows.
        Likewise, for other console apps, TargetPlatformIdentifier is windows.
        This file expects UAP for all .NET versions, so adjust. -->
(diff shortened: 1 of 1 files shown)


axelandrejs (approved):

dev/MRTCore/packaging/Microsoft.ApplicationModel.Resources.PriGen.targets
```diff
@@ -1435,44 +1245,6 @@
...
-
-  <PropertyGroup>
-    <_CreateAppxBundlePlatformSpecificArtifactsDependsOn>
-      $(Before_CreateAppxBundlePlatformSpecificArtifacts)
-
-      _GetPackageProperties;
-
-      _GetDefaultResourceLanguage;
-
-      _CreatePriConfigXmlForSplitting;
```
huichen123, line 1450: >_CreatePriConfigXmlForSplitting; [](start = 6, length = 32)

are these tasks still called?

rohanp-msft: some may not be, i'll remove them in round 3

huichen123: for those still need to be run, are they still running when you remove this?

rohanp-msft: yup, e.g., _GetDefaultResourceLanguage

huichen123: any task in the list with PRI needs to run, e.g. _ExpandMainPriFile?

rohanp-msft: I suspect that any target in this list that's not referenced by another target in this file can go.

PR #704, opened 2021-04-06
Clean up of Microsoft.ApplicationModel.Resources.PriGen.targets (round 3)

rohanp-msft:
Remove some unused targets and associated properties, as well as two unused items.

Microsoft.ApplicationModel.Resources.PriGen.targets was created from Microsoft.AppxPackage.targets where a bunch of appx generation MSBuild script code was removed (Microsoft.AppxPackage.targets does more than just PRI generation, which is all that we want Microsoft.ApplicationModel.Resources.PriGen.targets to do, as the name suggests). At that point though, not all non-PRI gen related MSBuild script code was removed. I'll now try to remove it over several rounds. This is round 3.

With regards to this specific change:
These targets are APPX specific. They were run before _CreateAppxBundlePlatformSpecificArtifacts which was removed in version 0 of Microsoft.ApplicationModel.Resources.PriGen.targets.

Round 1: #664
Round 2: #696

**How verified**
Used the MRTCore CPP sample app. Got rid of build outputs. Replaced the prigen targets file in the nuget package install with the one in this change. Built and ran the app.


dev/MRTCore/packaging/Microsoft.ApplicationModel.Resources.PriGen.targets
```diff
@@ -1245,243 +1158,6 @@
     </PropertyGroup>
   </Target>
 
-  <!-- Create upload pri config -->
```
axelandrejs, line 1248: My favorite kind of change!

axelandrejs (approved):

PR #706, opened 2021-04-07
Switch to Open Helix Queues (#665)

huichen123:
cherry-pick https://github.com/microsoft/ProjectReunion/pull/665

### DIFF
--- a/build/Helix/RunTestsInHelix.proj
+++ b/build/Helix/RunTestsInHelix.proj
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <HelixSource>pr</HelixSource>
     <HelixType>DevTestSuite</HelixType>
-    <HelixTargetQueues>Windows.10.Amd64.ClientRS5.Xaml;</HelixTargetQueues>
+    <HelixTargetQueues>Windows.10.Amd64.ClientRS5.Open.Xaml;</HelixTargetQueues>
     <EnableXUnitReporter>true</EnableXUnitReporter>
     <EnableAzurePipelinesReporter>true</EnableAzurePipelinesReporter>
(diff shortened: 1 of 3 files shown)


build/Helix/RunTestsInHelix.proj
```diff
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <HelixSource>pr</HelixSource>
     <HelixType>DevTestSuite</HelixType>
-    <HelixTargetQueues>Windows.10.Amd64.ClientRS5.Xaml;</HelixTargetQueues>
+    <HelixTargetQueues>Windows.10.Amd64.ClientRS5.Open.Xaml;</HelixTargetQueues>
```
jonwis, line 5: Are there other queues we can run across? Like 19h1, Vb, 20h1, etc?

Should we also be targeting x86?

huichen123: @kmahone ?

kmahone: Yes, there are other queues available. from RS5 to 20H2. This change was just to get the tests working again after switching to Open queues. We should merge this to get things unblocked and create a new PR after we decide on the new test matrix.

For WinUI2, we spread our build flavors (x86chk, x64fre, etc.) across the different OS versions. That way we avoid excessively blowing up the matrix. See 
[MUX-RunHelixTests-Job.yml from WinUI2]([https://github.com/microsoft/microsoft-ui-xaml/blob/master/build/AzurePipelinesTemplates/MUX-RunHelixTests-Job.yml#L15). We want want to do something similar here.

kmahone (approved):

PR #723, opened 2021-04-13
Remove other version references in props and targets

kythant:
Remove other version references in props and targets

Continuation of work from
https://github.com/microsoft/ProjectReunion/pull/702


erik0686 (approved):

erik0686:
This is a cool improvement :) Thanks

PR #727, opened 2021-04-14
Rewrite DynamicDependencies tests using TAEF

DrusTheAxe:
Issue [#559 Rewrite DynamicDependencies tests using TAEF](https://github.com/microsoft/ProjectReunion/issues/559)

Change the Dynamic Dependencies tests from VSTest to TAEF

Some tests were explicitly tagged with IsolationMode:Method to en masse execution produces clean results (thank you TAEF!). There's some potential optimizations in further refinements but considering the tests take ~40s on an 3rd generation i7 it's good for today.

I was also able to get rid of the ugly undo-STA-so-we-can-get-an-MTA hack by replacing VSTest with TAEF and its support for declarative COM initialization (thanks TAEF!)

Also: updated C++/WinRT, TAEF and WIL package dependencies across all projects to the latest revs.
And now all reference the _same_ versions, for Solution-wide consistency.


test/DynamicDependency/Test_WinRT/pch.h
```diff
@@ -32,10 +32,8 @@
 #include <MddBootstrap.h>
 #include <MddBootstrapTest.h>
 
-#include "CppUnitTest.h"
-#include "CppUnitTestX.h"
+#include <WexTestClass.h>
```
aeloros, line 35: nit: Why is the above MsixDynamicDepency in the middle of the winrt headers instead of grouped outside it? ;P

aeloros (approved):

PR #744, opened 2021-04-19
Move MrtCore up to cswinrt 1.2.3 to keep in sync with WinUI

j0shuams:


dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/projection/Microsoft.ApplicationModel.Resources.Projection.csproj
```diff
@@ -8,7 +8,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Windows.CsWinRT" Version="1.2.1" />
+    <PackageReference Include="Microsoft.Windows.CsWinRT" Version="1.2.3" />
```
huichen123, line 11: does this require sdk version change?

Scottj1s:
not sure this update is necessary.  Reunion 0.5.5 contained WinUI built with 1.2.2, and MRTCore built with 1.2.1.  Along with 1.2.3, winrt.runtime.dll has not changed assembly file version across any of these (1.2.0.0), so from that perspective, it should be fine.

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/projection/Microsoft.ApplicationModel.Resources.Projection.csproj
Scottj1s: it does not require an  SDK later than the current requirement of framework *.16 (which will become .NET 5.0.6) - no change to winrt.runtime.dll version, 1.2.0.0

PR #747, opened 2021-04-21
Publish symbols to both public and internal servers

DefaultRyan:
This moves the private unstripped symbols to the MSFT internal server, and produces stripped symbols to publish to the public symbol servers.


riverar:
Doesn't this also strip private symbols from open source artifacts? Ideally full symbol information would be available for open bits.

DefaultRyan:
> Doesn't this also strip private symbols from open source artifacts? Ideally full symbol information would be available for open bits.

That's a fair point. Perhaps stripping the symbols does more harm than good in this case. @jonwis, any thoughts?

DrusTheAxe:
> That's a fair point. Perhaps stripping the symbols does more harm than good in this case.

It does. Please don't strip symbols from open source code. It degrades fidelity and productivity for no upside

jonwis:
> That's a fair point. Perhaps stripping the symbols does more harm than good in this case. @jonwis, any thoughts?

Agreed. Don't strip from OSS components.

From our closed-source helpers I'm reasonably certain we should _not_ strip either. Nothing in there will be "secret sauce" - it's wrappers and helpers around internals that need a public surface.  (Ideally we have a mode of stripping that retains good method names for the top layer.)

build/ProjectReunion-BuildFoundation.yml
```diff
@@ -232,7 +232,20 @@ jobs:
 #      dir /s $(Build.ArtifactStagingDirectory)\symbols
 
   - task: PublishSymbols@2
-    displayName: 'Publish symbols (without source indexing)'
+    displayName: 'Publish private symbols to internal server (without source indexing)'
+    inputs:
+      searchPattern: '$(Build.ArtifactStagingDirectory)\symbols\**\*.pdb'
+      symbolServerType: 'TeamServices'
+# this ADO task does not support indexing of github sources currently :-(
```
riverar, line 239: Nit: Indent and capitalize

build/ProjectReunion-BuildFoundation.yml
```diff
@@ -232,7 +232,20 @@ jobs:
...
-    displayName: 'Publish symbols (without source indexing)'
+    displayName: 'Publish private symbols to internal server (without source indexing)'
+    inputs:
+      searchPattern: '$(Build.ArtifactStagingDirectory)\symbols\**\*.pdb'
+      symbolServerType: 'TeamServices'
+# this ADO task does not support indexing of github sources currently :-(      
+      indexSources: false 
+      detailedLog: true
+    # There is a bug which causes this task to fail if LIB includes an inaccessible path (even though it does not depend on it).
+    # To work around this issue, we just force LIB to be any dir that we know exists.
```
riverar, line 243: Nit: Is there a bug number tracking this? I'd add it if so. That would give folks a method of quickly checking up on the bug to answer the question "can I delete these lines now?"

build/StripSymbols.ps1
```diff
@@ -0,0 +1,80 @@
...
+    {
+        return Get-ItemProperty -Path $windowsSDKRegPath | Select-Object -ExpandProperty $windowsDebuggingToolsRegRootKey
+    }
+    catch
+    {
+        return $null
+    }
+}
+
+function Strip-PrivateInformation(
```
riverar, line 28: Nit: `Strip` isn't an [approved PowerShell verb](https://docs.microsoft.com/en-us/powershell/scripting/developer/cmdlet/approved-verbs-for-windows-powershell-commands).
```suggestion
function Remove-PrivateSymbolInformation(
```

riverar (approved):
Just several non-blocking nits. Thank you very much for accommodating the public symbol changes.

build/ProjectReunion-BuildFoundation.yml
DefaultRyan: Not that I'm aware of. This comment is actually just copy pasted from the lines above, and I'm not the original author of the other one. 🤷‍♂️

PR #754, opened 2021-04-22
Lay groundwork for build/dyndep alignment

DrusTheAxe:
Add ProjectReunionBuildPipeline=1 when pipeline build is run (ProjectReunionBuildPipeline=0 for local development)

Add build step to generate overrides for DynamicDependencies (e.g. unique CLSID for each pipeline build of the DDLM)

Update DynamicDependencies to support overrides during pipeline builds (and use local constants for local development)

Deleted unused file buildall.props

Remaining work:
* Update DynamicDependency*.msix projects to modify appxmanifest.xml to match generated overrides (needed for tests)
* Update aggregator pipeline to match (needed for releases :P)


dev/DynamicDependency/MddLifetimeManagement.cpp
```diff
@@ -101,81 +101,90 @@ STDAPI MddLifetimeManagementGC() noexcept try
     winrt::Windows::Management::Deployment::PackageManager packageManager;
     winrt::hstring currentUser;
 
-    const UINT32 c_maxMajorVersion{ 9 };
+    const UINT16 c_majorMinorVersions[][3]{ // [n] = { Major, minMinor, maxMinor }
+        { 0, 8, 9 },
+        { 1, 0, 2 }
```
BenJKuhn, line 106: how does this not become stale? Does this need to change over time? If so, can this be generated into a .h in a build script or similar? Will may have thoughts about how to tie this to other build metadata for valid version numbers.

BenJKuhn, line 106: is the removal process going to break older versions? E.g. when 1.3 releases, it will remove 1.2, but this package doesn't know to look for 1.3. Or am I misunderstanding this relationship?

dev/DynamicDependencyDataStore/DynamicDependency.DataStore/winmain.cpp
```diff
@@ -17,7 +19,7 @@ using namespace Microsoft::WRL;
 // Implement the DataStore as a classic COM Out-of-Proc server, via WRL
 // See https://docs.microsoft.com/en-us/cpp/cppcx/wrl/how-to-create-a-classic-com-component-using-wrl?redirectedfrom=MSDN&view=vs-2019 for more details
 
-struct __declspec(uuid("D1AD16C7-EC59-4765-BF95-9A243EB00507")) DynamicDependencyDataStoreImpl WrlFinal : RuntimeClass<RuntimeClassFlags<ClassicCom>, IDynamicDependencyDataStore>
+struct __declspec(uuid(PR_DYNDEP_DATASTORE_CLSID_STRING)) DynamicDependencyDataStoreImpl WrlFinal : RuntimeClass<RuntimeClassFlags<ClassicCom>, IDynamicDependencyDataStore>
```
BenJKuhn, line 22: don't need a separate macro, can stringify PR_DYNDEP_DATASTORE_CLSID_UUID with a '#' prefix

BenJKuhn (commented):
I'll leave this for @kythant to approve. Left a few comments that require some clarification, and maybe further documentation if I misunderstood the behavior or intent.

dev/DynamicDependency/MddLifetimeManagement.cpp
DrusTheAxe: It's the DDLM cleanup support. Yes, it can go stale if we have 0.10 or 1.11 but it's also cross-version but we have no plans for 0.9+ and by the time we get to 1.11 or 2.1 I expect other changes in this space.

Making this work for current-version (1.0 implementation only cleans up 1.0.*, 1.1 only cleans up 1.1.*, etc) creates additional problems, e.g. we can't have 1 scheduled job doing the work, we need N (1 for each major.minor installed on the machine). Right now the API must be explicitly called. We don't have a recurring 'maintenance task' to tap this yet. @dhoehna and I've discussed that sort of maintenance-GC (he has similar needs for EnvironmentManager), plus there's the activity topic.

Changing this to only look for version.current poses additional complications for build (we lack some of that cross-pipeline coordination today) and testing (tests would need similar build-time not-constant support).

Given how the dyndep-build-time-overrides is shaping up a simple solution is starting to materialize. I'll make a note to review this for 1.0.

DrusTheAxe: Nope. By definition we will only ever remove _unnecessary_ DDLMs, meaning if a DDLM is the only one to satisfy a need then it stays.

DDLM packages are named with their exact version, e.g. **Microsoft.ProjectReunion.DDLM-0.8.0.0-x64**_<version>_x64__8wekyb3d8bbwe. If the only 0.8 x64 DDLM registered is 0.8.0.0 then it stays. If you've got 0.8.0.1, 0.8.0.2 and 0.8.0.3 then .1 + .2 can be removed (if they're not in use) as .3 will meet all callers' needs.

That's why the logic here enumerates all the DDLMs with the same Major.Minor+Architecture and doesn't mess with the highest version

BenJKuhn (approved):

PR #761, opened 2021-04-24
Add property to toggle creation of preview VSIX 

alwu-msft:
This change adds the MSBuild property `EnableExperimentalVSIXFeatures` which allows for enabling/disabling inclusion of preview/experimental features in the VSIX. Currently, setting the property to true (whether via MSBuild or via a parameter for the Create-VSIX pipeline) does the following:
* Appends `(Preview)` to the VSIX display name
* Appends `.Preview` to the VSIX Id
* Includes the UWP project templates and adds the `UWP` tag to the VSIX
* Marks the VSIX as `Preview`
* Prepends `[Experimental]` to every item and project template

The preview VSIX can be installed alongside the release VSIX.


dev/VSIX/Directory.Build.targets
```diff
@@ -1,12 +1,14 @@
 <?xml version="1.0" encoding="utf-8"?>
```
Scottj1s, line 1: something to consider: one thing I didn't get around to when porting from winui, was to remove the need to revert direct edits to source files by first copying them to a staging folder, as cppwinrt does:
[https://github.com/microsoft/cppwinrt/blob/342edc5290086334366881faf8c4564bba00625b/vsix/vsix.csproj#L153](https://github.com/microsoft/cppwinrt/blob/342edc5290086334366881faf8c4564bba00625b/vsix/vsix.csproj#L153)

Scottj1s (approved):
thanks!

dev/VSIX/Directory.Build.targets
alwu-msft: Yes, that's an excellent suggestion; it'd likely also resolve the issue I've run into sometimes wherein VS decides it doesn't need to run the revert target, forcing me to manually rebuild. I'll make a note to implement this next time I'm in this space.

PR #772, opened 2021-04-27
AppInstancing feature

aeloros:
This change is the initial implementation of the AppInstancing AppLifecycle feature.  It also contains changes to the API namespace and re-enables the AppLifecycle Desktop Bridge tests.


dev/AppLifecycle/ActivatedEventArgsBase.h
```diff
@@ -2,7 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 #pragma once
 
-namespace winrt::Microsoft::ApplicationModel::Activation::implementation
+namespace winrt::Microsoft::Windows::AppLifecycle::implementation
```
EHO-makai, line 5: I must be behind the times.  All our API's will now be Microsoft::Windows::something? Solves some issues with us (Windows) owning "Microsoft" top level namespace node. (Microsoft::UI is kind of presumptuous).

dev/AppLifecycle/ActivationRegistrationManager.cpp
```diff
@@ -11,14 +11,9 @@
...
 {
-    std::wstring GenerateEncodedLaunchUri(std::wstring const& appUserModelId, std::wstring const& contractId)
-    {
-        Windows::Foundation::Uri uri{ wil::str_printf<std::wstring>(L"%s://%s?%s=%s",
-            c_encodedLaunchSchemeName, appUserModelId.c_str(), c_contractIdKeyName, contractId.c_str()).c_str() };
-        return uri.AbsoluteUri().c_str();
-    }
+    using namespace winrt::Windows::Foundation;
 
     std::wstring GenerateCommandLine(std::wstring const& modulePath)
```
EHO-makai, line 18: This code does two levels of substitution with printf specifiers and is really hard to reason about.  Can you put a short example comment of what this does?

dev/AppLifecycle/ActivationRegistrationManager.cpp
```diff
@@ -48,9 +43,8 @@ namespace winrt::Microsoft::ApplicationModel::Activation::implementation
             
             for (auto verb : supportedVerbs)
             {
-                auto command = GenerateCommandLine(exePath.c_str()) + GenerateEncodedLaunchUri(L"App",
-                    c_fileContractId) + L"&Verb=" + verb.c_str() + L"&File=%1";
-
+                auto args = winrt::make<FileActivatedEventArgs>(verb.c_str(), L"%1", true).as<IInternalValueMarshalable>();
```
EHO-makai, line 46: did you create IInternalValueMarshalable?  Where is that specified?

dev/AppLifecycle/ActivationRegistrationManager.cpp.orig
```diff
@@ -0,0 +1,181 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
EHO-makai, line 1: what is an orig file?  Can you delete this?

dev/AppLifecycle/ActivationRegistrationManager.h
```diff
@@ -2,18 +2,15 @@
...
+#include <Microsoft.Windows.AppLifecycle.ActivationRegistrationManager.g.h>
 
-namespace winrt::Microsoft::ApplicationModel::Activation::implementation
+namespace winrt::Microsoft::Windows::AppLifecycle::implementation
 {
     // Registration constant values.
     static LPCWSTR c_argumentPrefix{ L"----" };
     static LPCWSTR c_argumentSuffix{ L":" };
     static LPCWSTR c_protocolArgumentString{ L"ms-protocol" };
-    static LPCWSTR c_launchSchemeName{ L"ms-launch" };
```
EHO-makai, line 13: why are these deleted?

dev/AppLifecycle/AppInstance.cpp
```diff
@@ -0,0 +1,415 @@
...
+#include <Microsoft.Windows.AppLifecycle.AppInstance.g.cpp>
+
+#include "ActivationRegistrationManager.h"
+#include "LaunchActivatedEventArgs.h"
+#include "ProtocolActivatedEventArgs.h"
+#include "FileActivatedEventArgs.h"
+#include "Association.h"
+#include "ExtensionContract.h"
+
+using namespace winrt;
```
EHO-makai, line 15: isn't this asking for trouble with respect to calling to the ABI instead of the projection?

dev/AppLifecycle/AppInstance.cpp
```diff
@@ -0,0 +1,415 @@
...
+
+    std::tuple<std::wstring, std::wstring> ParseCommandLine(std::wstring commandLine)
+    {
+        auto argsStart = commandLine.rfind(c_argumentPrefix);
+        if (argsStart == std::wstring::npos)
+        {
+            return { L"", L"" };
+        }
+
+        argsStart += 4;
```
EHO-makai, line 33: comment here (why 4?)

dev/AppLifecycle/AppInstance.cpp
```diff
@@ -0,0 +1,415 @@
...
+        return { kind, data };
+    }
+
+    AppInstance::AppInstance(uint32_t processId)
+    {
+        m_processId = processId;
+        m_isCurrent = (GetCurrentProcessId() == processId);
+
+        m_moduleName = ComputeAppId();
+        m_processName = wil::str_printf<std::wstring>(L"%s_%d", m_moduleName.c_str(), processId);
```
EHO-makai, line 75: Is this scheme potentially compromised by pid recycling?  (unlikely but wondering if you considered this)

dev/AppLifecycle/AppLifecycle.idl
```diff
@@ -1,11 +1,79 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
-namespace Microsoft.ApplicationModel.Activation
+namespace Microsoft.Windows.AppLifecycle
 {
-    static runtimeclass AppLifecycle
+    enum ExtendedActivationKind
```
EHO-makai, line 6: what was the "source data" for this list?

dev/AppLifecycle/AppInstance.cpp
```diff
@@ -0,0 +1,415 @@
...
+
+        m_instances.Init(m_moduleName + L"_Module");
+
+        // Wire up the Activated event.
+        std::wstring eventName = m_processName + c_activatedEventNameSuffix;
+        m_innerActivated.create(wil::EventOptions::ManualReset, eventName.c_str());
+
+        // This mutex should always be created first by the process it's associated with.
+        std::wstring mutexName = m_processName + L"_Mutex";
+        m_dataMutex.create(mutexName.c_str(), CREATE_MUTEX_INITIAL_OWNER, MUTEX_ALL_ACCESS);
```
EHO-makai, line 85: have you accounted for potential named mutex squatting?

aeloros:
> did you create IInternalValueMarshalable?  Where is that specified?  

Yep, ValueMarshaling.h below.

aeloros:
> isn't this asking for trouble with respect to calling to the ABI instead of the projection?   

I think that problem mostly exists in a WRL heavy code base. The few places the ABI is referenced here, I'm pretty sure it's being handled by the C++/WinRT interface projections.

dev/AppLifecycle/AppInstance.cpp
```diff
@@ -0,0 +1,415 @@
...
+        {
+            releaseOnExit = m_dataMutex.ReleaseMutex_scope_exit();
+
+            auto onInnerActivated = [weak_this]
+            {
+                auto strong_this{ weak_this.get() };
+                if (strong_this)
+                {
+                    strong_this->ProcessRedirectionRequests();
+                }
```
EHO-makai, line 105: if the weakRef does not resolve, do nothing is OK?

dev/AppLifecycle/AppInstance.cpp
```diff
@@ -0,0 +1,415 @@
...
+            }
+        }
+    }
+
+    IAsyncAction AppInstance::QueueRequest(AppLifecycle::AppActivationArguments args)
+    {
+        auto strongThis{ get_strong() };
+
+        // Push this work onto a background thread.
+        co_await resume_background();
```
EHO-makai, line 194: this is a threadpool thread, nothing below apears to require COM initialization, is that correct?

dev/AppLifecycle/ExtensionContract.h
```diff
@@ -2,31 +2,52 @@
...
-namespace winrt::Microsoft::ApplicationModel::Activation::implementation
+namespace winrt::Microsoft::Windows::AppLifecycle::implementation
 {
-    // Extension contract ids.
-    static LPCWSTR c_launchContractId = L"Windows.Launch";
-    static LPCWSTR c_fileContractId = L"Windows.File";
-    static LPCWSTR c_protocolContractId = L"Windows.Protocol";
-    static LPCWSTR c_startupTaskContractId = L"Windows.StartupTask";
-
     // This array holds the mapping between a class factory and it's extension contract Id.
```
EHO-makai, line 13: where did you move these constants?

dev/AppLifecycle/ValueMarshaling.h
```diff
@@ -0,0 +1,27 @@
...
+    // This interface used for the internal value marshaling mechanism.  Implementing this interface means
+    // that both Serialize and Deserialize be implemented.  Deserialize is a static and can't be expressed
+    // here.  Deserialize is required also by the encoded launch mechanism, and is how ActivatedEventArgs
+    // are created regardless of marshaling.  The below method is only used for marshaling.
+    MIDL_INTERFACE("1E875DD8-3E09-465D-8684-E76EF58145C5") IInternalValueMarshalable : IInspectable
+    {
+        virtual winrt::Windows::Foundation::Uri Serialize() PURE;
+    };
+
+    inline std::wstring GenerateEncodedLaunchUri(std::wstring const& appUserModelId, std::wstring const& contractId)
```
EHO-makai, line 20: an example (comment) of what this looks like "in real life" is priceless. :-)
L"%s://%s?%s=%s"

dev/AppLifecycle/RedirectionRequest.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+namespace winrt::Microsoft::Windows::AppLifecycle::implementation
+{
+    void RedirectionRequest::Open(std::wstring name)
+    {
+        m_data.Open(name);
+        m_name = name;
+    }
+
+    void RedirectionRequest::MarshalArguments(Microsoft::Windows::AppLifecycle::AppActivationArguments const& args)
```
EHO-makai, line 17: Please have someone on comd alias/team review this file.

dev/AppLifecycle/SharedMemory.h
```diff
@@ -0,0 +1,96 @@
...
+    {
+        m_name.erase();
+        m_view.reset();
+        m_file.reset();
+    }
+
+protected:
+    bool OpenInternal(size_t size)
+    {
+        m_file = wil::unique_handle(CreateFileMapping(INVALID_HANDLE_VALUE, nullptr, PAGE_READWRITE, 0, static_cast<DWORD>(size), m_name.c_str()));
```
EHO-makai, line 83: another opportunity for name squatting attacks

dev/AppLifecycle/SharedProcessList.h
```diff
@@ -0,0 +1,65 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+#pragma once
+#include "SharedMemory.h"
+#include "Association.h"
+
+namespace winrt::Microsoft::Windows::AppLifecycle::implementation
+{
+
+#define MAX_INSTANCE_COUNT 512
+    typedef DWORD SharedProcessListData[MAX_INSTANCE_COUNT];
```
EHO-makai, line 11: this constant governs how many simultaneous (temporary) instances of a single instance app we can arbitrate amongst?

test/TestApps/AppLifecycleTestApp/SingleInstanceTest.cpp
```diff
@@ -0,0 +1,61 @@
...
+    if (!keyInstance)
+    {
+        return false;
+    }
+
+    if (keyInstance.IsCurrent())
+    {
+        auto token = keyInstance.Activated(auto_revoke, [&keyInstance](const auto& sender, const AppActivationArguments& args) { OnActivated(sender, args); });
+
+        // Launch a second instance and test that it redirects to us.
```
EHO-makai, line 46: I wasn't sure if the MAX_INSTANCE_COUNT in the shared process list was trying to govern the number of simultaneous instances you could have redirect to the "one true instance".  Do you think there is any value in stressing this a little bit by launching a few more than 1 in rapid succession to see if the system holds up under a moderate amount of multi-process contention?

EHO-makai (commented):
One request and one last comment for this review round.
Request - can you have someone from your team review?
Comment - I think I remember reading somewhere that using named objects (mutex, sharedmemory) was not the most robust way to do this.  I think the suggestion was to use a file in the user profile directory.  You might search msdn and see if you think this is a more robust mechanism for this. 
It's really great that we are writing this code for everyone because it's not too bad to write a "mostly OK" version of this, but it's quite difficult to write a really robust version of this.

dev/AppLifecycle/ExtensionContract.h
EHO-makai: nevermind, found them in the individual files where they apply.

aeloros:
> Please have someone on comd alias/team review this file.  

Yup, I added Brian Alker for that.

dev/AppLifecycle/RedirectionRequest.cpp
```diff
@@ -0,0 +1,112 @@
...
+        if (supportInternalValueMarshaling)
+        {
+            THROW_IF_FAILED(StringCchCopy(reinterpret_cast<LPWSTR>(streamStart), (uri.length() + 1), uri.c_str()));
+        }
+        else
+        {
+            // Use COM stream marshaling.
+            com_ptr<IStream> stream;
+            THROW_IF_FAILED(CreateStreamOnHGlobal(nullptr, TRUE, stream.put()));
+            THROW_IF_FAILED(CoMarshalInterface(stream.get(), uuidofArgs, unk.get(), MSHCTX_LOCAL, nullptr, MSHLFLAGS_NORMAL));
```
brialmsft, line 58: Sorry for the basic question, but this is a big PR and as I understand I'm mostly being asked to evaluate the COM details. What's the context of this marshaling? Who's marshaling, and who's unmarshaling?

The reason this might be significant is that COM uses the marshaling stream to provide marshaling code with the broader context of why the marshaling is happening. If an object is being marshaled as part of a call, for example, we use an IStream object that knows about the call, and then marshaling code is able to QI the IStream for various interfaces to interrogate that context. We need that context to do stuff like avoiding premature stub rundown for calls that get delayed by one of the involved processes suspending. Or to apply security policies like "don't use untrusted unmarshalers for marshaling data streams that originated in appcontainer" (see COMGLB_UNMARSHALING_POLICY_HYBRID).

Direct CoMarshalInterface with a standard IStream implementation doesn't have access to any of that, so it's generally a red flag. If you're marshaling an arbitrary IUnknown (i.e. not custom-marshaled) cross-process for some arbitrary pair of processes (i.e. it's possible that one of the involved processes is suspendable), then the premature stub rundown issue in particular becomes problematic. This generally causes problems as follows:
* One of the processes suspends while this marshaling data is in flight.
* Marshaling created a stub for the object. If a stub spends too long without any connected proxies, we tear it down. This rundown is intended to avoid leaking the references the stub holds on the marshaled object.
* The suspended process resumes, and successfully unmarshals a proxy from the marshaling data. But when you make a call on the proxy, you get RPC_E_DISCONNECTED because the stub is gone.

dev/AppLifecycle/AppInstance.cpp
```diff
@@ -0,0 +1,421 @@
...
+                wil::unique_handle process(::OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION, FALSE, pid));
+                if (process != nullptr)
+                {
+                    instances.Append(make<AppInstance>(pid));
+                }
+                else
+                {
+                    // Remove orphan.
+                    auto releaseOnExit = s_current->m_dataMutex.acquire();
+                    for (uint32_t index = 0; index < s_current->m_instances.Size(); index++)
```
dabhattimsft, line 285: SharedProcessList::Remove already seems to be checking whether pid is present in the list or not. Maybe this loop can be avoided?

dabhattimsft (approved):

dev/AppLifecycle/ActivatedEventArgsBase.h
jonwis: That - `Microsoft.Windows.*` is correct, and our rule has been to only put things into third-level namespaces like this.

dev/AppLifecycle/AppActivationArguments.h
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+#pragma once
+#include "Microsoft.Windows.AppLifecycle.AppActivationArguments.g.h"
+
+namespace winrt::Microsoft::Windows::AppLifecycle::implementation
+{
+    struct AppActivationArguments : AppActivationArgumentsT<AppActivationArguments>
+    {
+        AppActivationArguments(winrt::Windows::ApplicationModel::Activation::IActivatedEventArgs platformArgs)
```
jonwis, line 10: Take C++/WinRT types by `const&` so `IActivatedEventArgs const& platformArgs`

dev/AppLifecycle/AppActivationArguments.h
```diff
@@ -0,0 +1,31 @@
...
+        {
+            m_kind = kind;
+            m_data = data;
+        }
+
+        Microsoft::Windows::AppLifecycle::ExtendedActivationKind Kind() { return m_kind; }
+        winrt::Windows::Foundation::IInspectable Data() { return m_data; }
+
+    private:
+        AppActivationArguments() = default;
```
jonwis, line 26: You can use `= delete` and get a much cooler compiler error and remove the "private" part since nobody should be calling this at all.

dev/AppLifecycle/ExtensionContract.h
```diff
@@ -2,31 +2,52 @@
...
+        return CompareStringOrdinal(uri.SchemeName().c_str(), -1, c_encodedLaunchSchemeName, -1, TRUE) == CSTR_EQUAL;
+    }
+
+    inline std::tuple<ExtendedActivationKind, winrt::Windows::Foundation::IInspectable> DecodeActivatedEventArgs(winrt::Windows::Foundation::Uri const& uri)
+    {
+        for (auto const& pair : uri.QueryParsed())
+        {
+            if (CompareStringOrdinal(pair.Name().c_str(), -1, c_contractIdKeyName, -1, TRUE) == CSTR_EQUAL)
+            {
+                for (int index = 0; index < _countof(c_extensionMap); index++)
```
jonwis, line 39: If you want to get fancy, you can use range-based for:

```c++
for (const auto& extension : c_extensionMap) {
    /* use "extension" instead of "c_extensionMap[index]" in the body */
}
```

dev/AppLifecycle/RedirectionRequest.cpp
```diff
@@ -0,0 +1,112 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+#include <pch.h>
+#include "ValueMarshaling.h"
+#include "RedirectionRequest.h"
+#include "ExtensionContract.h"
+#include "Association.h"
+
+namespace winrt::Microsoft::Windows::AppLifecycle::implementation
+{
+    void RedirectionRequest::Open(std::wstring name)
```
jonwis, line 11: Consider taking `std::wstring const&` here

dev/AppLifecycle/RedirectionRequest.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+            ULONG bytesWritten = 0;
+            THROW_IF_FAILED(stream->Write(streamStart, streamSize, &bytesWritten));
+
+            const LARGE_INTEGER headOffset{};
+            THROW_IF_FAILED(stream->Seek(headOffset, STREAM_SEEK_SET, nullptr));
+
+            com_ptr<::IUnknown> unk;
+            THROW_IF_FAILED(CoUnmarshalInterface(stream.get(), guid_of<AppActivationArguments>(), unk.put_void()));
+            return unk.as<winrt::Microsoft::Windows::AppLifecycle::AppActivationArguments>();
```
jonwis, line 109: I remember there being a way to get a "stream on blob of memory without copies" but I can't remember the types for it. So you'd basically do like "memstream str(streamStart, streamSize)" and then pass that right to the CoUnmarshal.

dev/AppLifecycle/RedirectionRequestQueue.h
```diff
@@ -0,0 +1,138 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+#pragma once
+#include "SharedMemory.h"
+#include "RedirectionRequest.h"
+#include <guiddef.h>
+
+namespace winrt::Microsoft::Windows::AppLifecycle::implementation
+{
+    class RedirectionRequestQueue
```
jonwis, line 10: Is there any locking required to read/modify this data?

jonwis (commented):
First round of comments - looks generally OK

dev/AppLifecycle/RedirectionRequest.cpp
brialmsft: We had an offline discussion about these concerns. The activated process should already be protected from suspension long enough to unmarshal these arguments. The activating process could take out a suspension deferral on itself if needed. If neither process involved in the marshaling suspends (or otherwise takes long enough to unmarshal that rundown kicks in), then you shouldn't see premature rundowns.
One other thing you could do is marshal with MSHLFLAGS_TABLESTRONG. This prevents the object from running down at all, which means that if marshal data gets lost in transit (e.g. the client dies before it touches it), and nobody calls CoReleaseMarshalData, then the server will just leak the object. It also allows the same unmarshal data to be unmarshaled arbitrarily many times, which means that you _always_ need to call CoReleaseMarshalData to free the marshal data when you're done with it. Under normal marshaling you can just discard the stream after you call CoUnmarshalInterface, but TABLESTRONG requires the extra cleanup step to avoid leaking.
It's up to you whether you'd prefer to go with explicit suspension avoidance (may lead to premature rundowns if you're wrong and one of the processes does manage to suspend at an inopportune moment) or disabling rundown (may cause permanent leaks in the activating process if the activated process crashes during startup; activating process could potentially do extra work to detect and clean up after this).

In terms of the security policy I mentioned, you do have the option of having your unmarshaling stream object implement IMarshalingStream, and responding to CO_MARSHALING_SOURCE_IS_APP_CONTAINER queries (i.e. checking if the marshaling process is appcontainer). That way we can consider the unmarshaling data less trusted and avoid feeding it to custom unmarshalers that may be exploitable. This security policy and rundown are the most significant things driven off of the marshaling stream.

Overall, it's difficult to wholeheartedly endorse any raw use of CoMarshalInterface with a "dumb" marshaling stream just because it gives us fewer tools to work with to ensure the marshaling is done in a way appropriate to the context. Our preference would be that you marshal this data in a COM call. I'm not going to block the PR over it, since I don't think any of these concerns are fatal, but be warned that you're going off of our most-supported path. There may be dragons.

dev/AppLifecycle/RedirectionRequest.cpp
```diff
@@ -0,0 +1,112 @@
...
+            {
+                stream->Seek(headOffset, STREAM_SEEK_SET, nullptr);
+                CoReleaseMarshalData(stream.get());
+            });
+
+            THROW_IF_FAILED(stream->Seek(headOffset, STREAM_SEEK_SET, nullptr));
+
+            STATSTG stats{};
+            THROW_IF_FAILED(stream->Stat(&stats, STATFLAG_NONAME));
+            THROW_HR_IF(E_FAIL, static_cast<ULONG>(stats.cbSize.QuadPart) > streamSize);
```
brialmsft, line 71: streamSize isn't quite right, because it's sizeof(bool) +\<marshalData size\>, but streamStart is already advanced past the bool. I think a buffer overrun is possible if CoGetMarshalSizeMax underestimated by a small amount.

brialmsft (changes requested):

brialmsft (approved):

test/inc/TestDef.h
```diff
@@ -14,3 +15,34 @@ static const std::wstring c_testFileExtension_Packaged = L".reuniontestfile-pack
...
+
+static const std::wstring c_testInstanceRedirectedPhaseEventName = L"ReunionTestInstanceRedirectedPhaseEventName";
+
+#ifndef WIDEN2
+#define WIDEN2(x) L ## x
+#endif
+
+#ifndef WIDEN
+#define WIDEN(x) WIDEN2(x)
+#endif
```
DrusTheAxe, line 27: Does STL have something for this?

#x is accomplished via
```
// Rely on _STRINGIZE(x) in yvals_core.h
#ifndef _STRINGIZE
#define _STRINGIZEX(x)  #x
#define _STRINGIZE(x)   _STRINGIZEX(x)
#endif
```

Is there something equivalent already present in std headers to do this?

test/inc/TestDef.h
```diff
@@ -14,3 +15,34 @@ static const std::wstring c_testFileExtension_Packaged = L".reuniontestfile-pack
...
+#if defined(_WIN32)
+#define __WFUNCTION__ WIDEN(__FUNCTION__)
+#else
+#define __WFUNCTION__ L""
+#endif
+#endif
+
+inline std::wstring StripNS(PCWSTR symbol)
+{
+    std::wstring result = symbol;
```
DrusTheAxe, line 43: >std::wstring result = symbol; [](http://example.com/codeflow?start=4&length=29)

NIT: {}

std::wstring result{ symbol };

DrusTheAxe, line 43: >result [](http://example.com/codeflow?start=17&length=6)

Name's a bit bland

SUGGESTION: `symbolAsString`

test/inc/TestDef.h
```diff
@@ -14,3 +15,34 @@ static const std::wstring c_testFileExtension_Packaged = L".reuniontestfile-pack
...
+    return result.substr(result.rfind(L":") + 1);
```
DrusTheAxe, line 44: >L":") [](http://example.com/codeflow?start=38&length=5)

NIT: `L':'`

DrusTheAxe, line 44: >result.rfind(L":") + 1 [](http://example.com/codeflow?start=25&length=22)

If rfind doesn't find a match it returns `string::npos` == `size_t(~0)` == 0xFFFFFFF....

+1 turns that into 0

substr(0) keeps the whole string

Cryptic way to make "not found" an expensive NOP

CONSIDER: bit more straightforward logic

dev/AppLifecycle/AppActivationArguments.h
DrusTheAxe: Are `const Foo&` and `Foo const&` equivalent?

dev/AppLifecycle/ActivationRegistrationManager.h
```diff
@@ -2,18 +2,15 @@
...
-    static LPCWSTR c_encodedLaunchSchemeName{ L"ms-encodedlaunch" };
     static LPCWSTR c_runKeyPath{ LR"(Software\Microsoft\Windows\CurrentVersion\Run\)" };
```
DrusTheAxe, line 13: >LPCWSTR [](http://example.com/codeflow?start=11&length=7)

PCWSTR

same comment applies throughout (global search/replace LPCWSTR->PCWSTR ?)

dev/AppLifecycle/ActivationRegistrationManager.cpp
```diff
@@ -11,19 +11,15 @@
...
-    }
+    using namespace winrt::Windows::Foundation;
 
     std::wstring GenerateCommandLine(std::wstring const& modulePath)
     {
         std::wstring exePath{ modulePath.empty() ? GetModulePath() : modulePath };
 
+        // Example: C:\some\path\App.exe ----ms-protocol:
         return wil::str_printf<std::wstring>(L"%s %s%s%s", exePath.c_str(), c_argumentPrefix,
             c_protocolArgumentString, c_argumentSuffix);
```
DrusTheAxe, line 24: Complicated way to say exePath + "----msprotocol:"

I know all the cool kids frown upon #define but that would simplify to exePath.c_str() + C_ARGPREFIX C_ARGSTRING C_ARGSUFFIX

Perhaps 1 typedef constant (PCWSTR) with Prefix+Arg+Suffix to make the preprocessor haters happy and still be efficient and clear?

dev/AppLifecycle/ActivationRegistrationManager.cpp
```diff
@@ -48,9 +44,8 @@ namespace winrt::Microsoft::ApplicationModel::Activation::implementation
             
             for (auto verb : supportedVerbs)
             {
-                auto command = GenerateCommandLine(exePath.c_str()) + GenerateEncodedLaunchUri(L"App",
-                    c_fileContractId) + L"&Verb=" + verb.c_str() + L"&File=%1";
-
+                auto args = winrt::make<FileActivatedEventArgs>(verb.c_str(), L"%1", true).as<IInternalValueMarshalable>();
```
DrusTheAxe, line 47: >L"%1" [](http://example.com/codeflow?start=78&length=5)

Magic constant

CONSIDER: make a symbolic constant

DrusTheAxe, line 47: > auto args = [](http://example.com/codeflow?start=15&length=12)

NIT: auto args { .. };

dev/AppLifecycle/AppLifecycle.idl
```diff
@@ -1,11 +1,79 @@
...
+        StartupTask,
+        CommandLineLaunch,
+        BarcodeScannerProvider,
+        PrintSupportJobUI,
+        PrintSupportSettingsUI,
+        PhoneCallActivation,
+        VpnForeground,
+
+        // NOTE: Values below 5000 are designated for the platform.  The above list is kept in sync with
+        // Windows.ApplicationModel.Activation.ActivationKind.
```
DrusTheAxe, line 54: >// Windows.ApplicationModel.Activation.ActivationKind. [](http://example.com/codeflow?start=8&length=54)

This list is intended to be updated regularly (whenever Windows.ApplicationModel.Activation.ActivationKind is updated and published in a new SDK)?

dev/AppLifecycle/AppLifecycle.idl
```diff
@@ -1,11 +1,79 @@
...
+        GameUIProvider,
+        StartupTask,
+        CommandLineLaunch,
+        BarcodeScannerProvider,
+        PrintSupportJobUI,
+        PrintSupportSettingsUI,
+        PhoneCallActivation,
+        VpnForeground,
+
+        // NOTE: Values below 5000 are designated for the platform.  The above list is kept in sync with
```
DrusTheAxe, line 53: > // NOTE: Values below 5000 are designated for the platform.  The above list is kept in sync with [](http://example.com/codeflow?start=7&length=97)

Does the OS promise to only use values <5000?

(Can we look forward to such a comment in an OS header sometime soon?)

dev/AppLifecycle/EncodedLaunchExecuteCommand.cpp
```diff
@@ -2,13 +2,14 @@
...
 #include "pch.h"
+#include "ExtensionContract.h"
 #include "EncodedLaunchExecuteCommand.h"
 #include "ActivationRegistrationManager.h"
 
 using namespace winrt;
 using namespace winrt::Windows::Foundation;
 
-namespace winrt::Microsoft::ApplicationModel::Activation::implementation
+namespace winrt::Microsoft::Windows::AppLifecycle::implementation
```
DrusTheAxe, line 12: >Microsoft::Windows::AppLifecycle [](http://example.com/codeflow?start=17&length=32)

You'll also need to update your namespace at...

build\AzurePipelinesTemplates\ProjectReunion-CreateNugetPackage-Job.yml
build\CopyFilesToStagingDir.ps1
build\NuSpecs\AppxManifest.xml
build\NuSpecs\Intellisense\Microsoft.ApplicationModel.Activation.xml
build\NuSpecs\ProjectReunion-Nuget-Native.WinRt.props
test\DynamicDependency\data\Microsoft.ProjectReunion.Framework\appxmanifest.xml

Until those are corrected some of those will fail the pipeline fatally, and some will fail the pipeline nonfatally - as in, an error in a powershell script will be ignored and PS moves on, resulting in the pipeline end status says OK/green, but it's actually incomplete and anyone using Reunion NuGet and MSIX will get build + runtime errors (yeah, known issue. Being worked on)

FYI regex I used to find all references to the namespace

Microsoft..?ApplicationModel..?Activation

as a case-insenstive regex across all files in ProjectReunion

(and the internal build pipeline, which found no hits. As it shouldn't; that's all data driven by what comes out of the ProjectReunion repo and build pipeline)

DrusTheAxe: BTW
```
DIR *activation* /s/b
```
only turns up 1 hit build\NuSpecs\Intellisense\Microsoft.ApplicationModel.Activation.xml (mentioned above). You'll need to rename that file when you rename the content of the rest to point to the right new filename

dev/AppLifecycle/AppInstance.cpp
DrusTheAxe: `argsStart += wcslen(c_argumentPrefix);`

Of course if c_argumentPrefix was a #define that could just be `argsStart += (ARRAYSIZE(c_argumentPrefix) - 1);`. He who liveth by the string sword, dieth by the string sword...

dev/AppLifecycle/AppInstance.cpp
```diff
@@ -0,0 +1,415 @@
...
+        // We explicitly use find_first_of here, so that the resulting data may contain : as a valid character.
+        auto argsEnd = commandLine.find_first_of(L":", argsStart);
+        if (argsEnd == std::wstring::npos)
+        {
+            return { L"", L"" };
+        }
+
+        if (argsStart > argsEnd)
+        {
+            throw std::overflow_error("commandLine");
```
DrusTheAxe, line 45: >overflow_error [](http://example.com/codeflow?start=23&length=14)

Kind of a weird exception for this error. Is there a better option?

dev/AppLifecycle/AppInstance.cpp
```diff
@@ -0,0 +1,415 @@
...
+
+    std::tuple<ExtendedActivationKind, winrt::Windows::Foundation::IInspectable> GetEncodedLaunchActivatedEventArgs(IProtocolActivatedEventArgs const& args)
+    {
+        auto kind = ExtendedActivationKind::Protocol;
+        winrt::Windows::Foundation::IInspectable data;
+
+        std::tie(kind, data) = DecodeActivatedEventArgs(args.Uri());
+
+        // Let the caller args pass through if nothing was determined here.
+        if (data == nullptr)
```
DrusTheAxe, line 62: >data == nullptr [](http://example.com/codeflow?start=12&length=15)

NIT: `if (!data)`

dev/AppLifecycle/AppInstance.cpp
```diff
@@ -0,0 +1,415 @@
...
+
+        m_instances.Init(m_moduleName + L"_Module");
+
+        // Wire up the Activated event.
+        std::wstring eventName = m_processName + c_activatedEventNameSuffix;
+        m_innerActivated.create(wil::EventOptions::ManualReset, eventName.c_str());
+
+        // This mutex should always be created first by the process it's associated with.
+        std::wstring mutexName = m_processName + L"_Mutex";
+        m_dataMutex.create(mutexName.c_str(), CREATE_MUTEX_INITIAL_OWNER, MUTEX_ALL_ACCESS);
```
DrusTheAxe, line 86: > m_dataMutex.create(mutexName.c_str(), CREATE_MUTEX_INITIAL_OWNER, MUTEX_ALL_ACCESS); [](http://example.com/codeflow?start=7&length=85)

Is this API expected to be used in a process running in an AppContainer? Will this be accessed from outside the AppContainer?

dev/AppLifecycle/AppInstance.cpp
```diff
@@ -0,0 +1,415 @@
...
+        auto alreadyExists = (GetLastError() == ERROR_ALREADY_EXISTS);
```
DrusTheAxe, line 87: >(GetLastError() == ERROR_ALREADY_EXISTS); [](http://example.com/codeflow?start=29&length=41)

Is there a WIL mutex.create() that does this? An optional out parameter or something?

dev/AppLifecycle/AppInstance.cpp
```diff
@@ -0,0 +1,415 @@
...
+
+        // This means the AppInstance object was constructed for a process that
+        // does not exist.  This is typical for when we found an orphan in m_instances.
+        auto orphan = (!alreadyExists && !m_isCurrent);
+        assert(!orphan);
```
DrusTheAxe, line 92: >assert [](http://example.com/codeflow?start=8&length=6)

assert is not recommended. Only works for 'debug'/CHK binaries, which are rare birds (nigh extinct) these days. Alternatives:
1. Do nothing
2. Error handling (e.g. THROW_HR_IF)
3. Fatality (e.g. FAIL_FAST_HR_IF(...))

dev/AppLifecycle/AppInstance.cpp
```diff
@@ -0,0 +1,415 @@
...
+                }
+            };
+
+            m_terminationWatcher.create(m_instanceHandle.get(), onInstanceTerminated);
+        }
+
+        m_redirectionArgs.Init(m_processName + L"_RedirectionQueue");
+    }
+
+    void AppInstance::OnInstanceTerminated()
```
DrusTheAxe, line 144: >AppInstance::OnInstanceTerminated() [](http://example.com/codeflow?start=9&length=35)

I see _reactive_ behavior, removing PIDs when the API is called. Is there a _proactive_ monitoring for process death to cleanup on process termination? That would reduce the window for PID recycling to grief you

dev/AppLifecycle/AppInstance.cpp
```diff
@@ -0,0 +1,415 @@
...
+    {
+        auto releaseOnExit = m_dataMutex.acquire();
+        m_redirectionArgs.Enqueue(id);
+    }
+
+    void AppInstance::ProcessRedirectionRequests()
+    {
+        m_innerActivated.ResetEvent();
+
+        GUID id;
```
DrusTheAxe, line 167: >GUID id; [](http://example.com/codeflow?start=8&length=8)

NT: `GUID id{};`

dev/AppLifecycle/AppInstance.cpp
```diff
@@ -0,0 +1,415 @@
...
+        while ((id = DequeueRedirectionRequestId()) != GUID_NULL)
```
DrusTheAxe, line 168: >GUID_NULL [](http://example.com/codeflow?start=55&length=9)

Is this or `GUID{}` the recommended way to checkfor GUID=00000...

dev/AppLifecycle/SharedMemory.h
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    void Resize(size_t size)
+    {
+        auto name = m_name;
+        Reset();
+        Open(name, size);
+    }
+
+    size_t Size()
```
DrusTheAxe, line 58: >size_t Size() [](http://example.com/codeflow?start=4&length=13)

NIT: const

dev/AppLifecycle/SharedMemory.h
```diff
@@ -0,0 +1,96 @@
...
+
+    T* Get()
+    {
+        return &m_view.get()->data;
+    }
+
+    T* operator->()
+    {
+        return Get();
+    }
```
DrusTheAxe, line 71: NIT: overload this with const variants?

dev/AppLifecycle/SharedProcessList.h
```diff
@@ -0,0 +1,65 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+#pragma once
+#include "SharedMemory.h"
+#include "Association.h"
+
+namespace winrt::Microsoft::Windows::AppLifecycle::implementation
+{
+
+#define MAX_INSTANCE_COUNT 512
```
DrusTheAxe, line 10: >#define MAX_INSTANCE_COUNT [](http://example.com/codeflow?start=0&length=26)

NIT: `const auto c_maxInstanceCount{ 512 };`

dev/AppLifecycle/SharedProcessList.h
```diff
@@ -0,0 +1,65 @@
...
+            for (DWORD index = 0; index < m_view.size(); index++)
+            {
+                if (m_view[index] == processId)
+                {
+                    m_view[index] = 0;
+                    return;
+                }
+            }
+
+            return;
```
DrusTheAxe, line 48: >         return; [](http://example.com/codeflow?start=3&length=16)

NIT: remove

dev/AppLifecycle/SharedProcessList.h
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        DWORD Size()
```
DrusTheAxe, line 51: const

dev/AppLifecycle/SharedProcessList.h
```diff
@@ -0,0 +1,65 @@
...
+        {
+            return m_view.size();
+        }
+
+        DWORD operator[](int index)
```
DrusTheAxe, line 56: const

test/TestApps/AppLifecycleTestApp/Helpers.cpp
```diff
@@ -0,0 +1,80 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+#include "pch.h"
+#include "Helpers.h"
```
DrusTheAxe, line 4: >Helpers [](http://example.com/codeflow?start=10&length=7)

Not fond of the name :-(

Personal nit. Helper, Util and Temp are usually too blandly generic and makes me sad.

test/TestApps/AppLifecycleTestApp/Helpers.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include <TestDef.h>
+
+const wchar_t* g_bootStrapDllName = L"Microsoft.ProjectReunion.Bootstrap.dll";
+
+typedef HRESULT (*BootStrapTestInit)(PCWSTR prefix, PCWSTR publisherId);
+typedef HRESULT (*BootStrapInit)(const UINT32 majorMinorVersion, const PACKAGE_VERSION minVersion);
```
DrusTheAxe, line 10: >rVersion, co [](http://example.com/codeflow?start=55&length=12)

MddBootstrapInitialize() recently changed to add a 3rd parmeter -- majorminor, PCWSTR versionTag, minversion

Needed to support Reunion's versioning. Pass null or "" for 'no tag' and you're good

Alas, it's a C API so I couldn't overload it to accept 2 + 3 parameters and make versionTag optional

test/TestApps/AppLifecycleTestApp/Helpers.cpp
```diff
@@ -0,0 +1,80 @@
...
+typedef void (*BootStrapShutdown)();
+
+bool IsPackagedProcess()
+{
+    UINT32 n{};
+    return ::GetCurrentPackageFullName(&n, nullptr) == ERROR_INSUFFICIENT_BUFFER;;
+}
```
DrusTheAxe, line 17: Dupe(ish) of IsPackagedProcess() in test\inc\ProjectReunion.Test.AppModel.h

Remove, #include <ProjectReunion.Test.AppModel.h> and add namespace prefix to callers (or "using Test::AppModel;" ?)

test/TestApps/AppLifecycleTestApp/Helpers.cpp
```diff
@@ -0,0 +1,80 @@
...
+        return;
+    }
+
+    wil::unique_hmodule bootStrapDll(LoadLibraryEx(g_bootStrapDllName, NULL, 0));
+    if (!bootStrapDll)
+    {
+        return;
+    }
+
+    BootStrapShutdown mddShutdown = reinterpret_cast<BootStrapShutdown>(GetProcAddress(bootStrapDll.get(), "MddBootstrapShutdown"));
```
DrusTheAxe, line 74: >= [](http://example.com/codeflow?start=34&length=1)

NIT: {}

DrusTheAxe (approved):
:shipit:

test/TestApps/AppLifecycleTestApp/SingleInstanceTest.cpp
DrusTheAxe: Good idea to add a test that activates max+1 and verifies it fails as expected

PR #792, opened 2021-05-03
This is a sample solution to test if we can log telemetry from OSS. T…

ZenBird:
…his solution will be built using the build pipeline. The build pipeline will replace the reference to WIL nuget package with the WIL internal nuget package.


riverar:
> [...] will replace the reference to WIL nuget package with the WIL internal nuget package.
Hm. I am concerned by this language. Will the community be able to compile main with these _internal nuget_ references in place?

jonwis:
We should not be committing another copy of WIL in this repo like this.  There's got to be another way to accomplish this test.  Maybe make a private fork, or work in your branch and ask the build system to run the pipelines using the `/azp run` switch.  @BenJKuhn or @kythant maybe can help you out here.

Please do *not* merge this as-is.

(Also note that `dev/` is where product code goes; if you'd like to add `test/` or `sample/` that's OK, but note that we have a samples repo already.)

@riverar - No, this should not restrict the repo to only building with the internal version of WIL.  I suspect that meant "use an internal build-pipe configuration when building the released bits" ... the ADO pipelines used to build our released software have additional configuration knobs applied.  Thanks for helping keep our "working in the open with our customers" promise!

ProjectReunion.sln
```diff
@@ -94,6 +94,8 @@ Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "DynamicDependencyLifetimeMa
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Common", "Common", "{A657A14D-910D-43F3-9994-DAB0E22C4530}"
 EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "TelemetryTest", "dev\TelemetryTest\TelemetryTest.vcxproj", "{EAB089AB-2F5E-4BB8-9ECB-E90E2A2C12F7}"
```
jonwis, line 97: See comments on the PR. This is the wrong location to put a test executable.

dev/TelemetryTest/wil/telemetry/MicrosoftTelemetry.h
```diff
@@ -0,0 +1,238 @@
+﻿#pragma once
+/* ++
+
+Copyright (c) Microsoft Corporation.  All rights reserved.
+
+Module Name:
+
+    MicrosoftTelemetry.h
```
jonwis, line 8: This should *definitely not* be committed here. Please remove, and then we'll need to rotate the telemetry identifiers when you bring it back in a different form.

jonwis (changes requested):

PR #796, opened 2021-05-04
Target MRTCore projection dll to any CPU

huichen123:


jonwis:
Does this support apps that _do_ want to target a specific CPU?

huichen123:
> Does this support apps that _do_ want to target a specific CPU?

The app using the projection dll can still target a specific CPU.

axelandrejs (approved):

PR #798, opened 2021-05-04
Update MddBootstrapInitialize docs

DrusTheAxe:


specs/dynamicdependencies/DynamicDependencies.md
```diff
@@ -1263,8 +1263,12 @@ This header contains the Bootstrap API
 /// for use by the current process. If multiple packages meet the criteria the best
 /// candidate is selected.
 ///
+/// @param majorMinorVersion major and minor version of Project Reunion's framework package, encoded as `0xMMMMNNNN` where M=Major, N=Minor (e.g. 1.2 == 0x00010002).
+/// @param versionTag version tag (if any), e.g. "prerelease".
 /// @param minVersion the minimum version to use
 STDAPI MddBootstrapInitialize(
+    const UINT32 majorMinorVersion,
```
jonwis, line 1270: Do not put const in method declarations like this.  Same goes for the `minVersion` below; take `PCPACKAGE_VERSION` maybe if it's a large structure.

specs/dynamicdependencies/DynamicDependencies.md
```diff
@@ -1263,8 +1263,12 @@ This header contains the Bootstrap API
 /// for use by the current process. If multiple packages meet the criteria the best
 /// candidate is selected.
 ///
+/// @param majorMinorVersion major and minor version of Project Reunion's framework package, encoded as `0xMMMMNNNN` where M=Major, N=Minor (e.g. 1.2 == 0x00010002).
+/// @param versionTag version tag (if any), e.g. "prerelease".
```
jonwis, line 1267: So this should be marked `_In_opt_`?

jonwis (changes requested):

specs/dynamicdependencies/DynamicDependencies.md
DrusTheAxe: It's not, just sizeof(INT64)

specs/dynamicdependencies/DynamicDependencies.md
DrusTheAxe: No need. That's redundant with PCWSTR -- pointers can be null and PCWSTR data can't be changed so effectively _In_opt_

jonwis (approved):

PR #800, opened 2021-05-05
Fix the versioning on Foundation nuspec

kythant:
Fix the versioning on Foundation nuspec.
This was becoming too confusing and uninformative.


DrusTheAxe (approved):
Snicker

build/ProjectReunion-Versions.yml
```diff
@@ -0,0 +1,12 @@
+variables:
+  # Version Control
+  major: "0"
+  minor: "5"
+  patch: "0"
+  branch: 'main'
```
huichen123, line 6: the branch specific stuff needs to be in a separate file.

kythant: I don't think I agree with that. This is all part of the version name so we should just keep it here instead of in a separate place

huichen123: if we change the format, we have to manually port the change to different branches (i.e. cannot cherrypick). Not a big deal though as this is a small file.

kythant: Ah I see what you meaannn

build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
```diff
@@ -7,12 +7,8 @@ parameters:
 # The "primary" build arch is the one that the nuspec gets its winmd, pri, and other neutral files from
   primaryBuildArch: x86
```
DrusTheAxe, line 8: Is x86 the touchstone we want to use?
Wondering if x64 wouldn't make a better choice

Just a thought

build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
```diff
@@ -7,12 +7,8 @@ parameters:
...
   buildFlavor: Release
-  useReleaseTag: 'false'
   prereleaseVersionTag: 'prerelease'
```
DrusTheAxe, line 10: Why's major/minor/patch removed but tag still present?

Or that too is going just not there yet?

build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
```diff
@@ -87,15 +82,23 @@ jobs:
...
   - task: 333b11bd-d341-40d9-afcf-b32d5ce6f23b@2
     displayName: 'Pack Full Nuget Package'
     inputs:
       command: 'pack'
       packagesToPack: '$(Build.SourcesDirectory)\build\NuSpecs\Microsoft.ProjectReunion.Foundation.nuspec'
-      versioningScheme: 'byPrereleaseNumber'
-      majorVersion: ${{ parameters.major }}
-      minorVersion: ${{ parameters.minor }}
-      patchVersion: ${{ parameters.patch }}
+      versioningScheme: 'off'
```
DrusTheAxe, line 101: What's this do? (sounds like a proper noun fronting a bunch of machinery)

build/ProjectReunion-BuildFoundation.yml
```diff
@@ -361,7 +361,7 @@ jobs:
     dependsOn: SignBinariesAndPublishSymbols
     signConfigPattern: 'Microsoft.ProjectReunion*.nupkg'
     useReleaseTag: '$(ProjectReunionFinalRelease)'
-    prereleaseVersionTag: prerelease
+    prereleaseVersionTag: nightly
```
DrusTheAxe, line 364: VersionTag should be alphabetically sortable -- A<B<C<D...

As this is (the?) earliest in the lifecycle it would preclude naming any tab starting with a-m
Maybe call this "daily" (or "alpha"?)

build/ProjectReunion-CI.yml
```diff
@@ -138,6 +137,9 @@ jobs:
   condition: in(dependencies.BuildMRT.result, 'Succeeded')
   pool:
     vmImage: 'windows-latest'
+  variables:
+    minimumExpectedTestsExecutedCount: 500  # Sanity check for minimum expected tests to be reported
```
DrusTheAxe, line 141: 500? We have that many that routinely run? (or more actually, 500 is the "minimum expected")

build/ProjectReunion-Versions.yml
DrusTheAxe: Sticking the version into its own specific file is ideal. Easiest to manage, as it's purely data (separate from tasks/jobs/script/etc which are 'code')

build/ProjectReunion-BuildFoundation.yml
riverar: ⚠ There's an on-going proposal/discussion around naming, if you push this into main, you're telling the world the official experimental builds are `-nightly` and will have locked us into that naming. @andrewleader are you in sync with @kythant here? This does not seem right.

riverar (changes requested):

build/ProjectReunion-BuildFoundation.yml
andrewleader: Thanks Rafael, syncing up with Kyaw now to learn more.

BenJKuhn: There's some pressure within the team to get nightly builds running, not just CI to help with integration, testing, and workflow issues. This is independent of the discussion around and release channels, and further churn in package names should probably be expected at this point in the project lifecycle. This name avoid conflation with release channels discussion.

build/ProjectReunion-Versions.yml
kythant: I can move it to a specific yml file like this one. This yml file is meant to be purely data.
I can totally separate the variables other than major, minor, and patch to another file called ProjectReunion-BuildData.yml or something of sort.

build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
kythant: This is so the task doesn't use it's own versioning scheme.

build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
kythant: The tag is present as a parameter because we have two different entries to the Foundation build. One internally and one in the 'ms' org azdo.

build/ProjectReunion-CI.yml
kythant: This variable is not used it seems.  It is in the commented out code below so i'll just remove it.

build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
kythant: This parameter is not being used either

build/ProjectReunion-BuildFoundation.yml
riverar: Thanks @BenJKuhn. My concern was the merge target (main) here. But perhaps my understanding of the Reunion workflow/branching is incorrect. (It certainly appear to align with typical open source workflows.)

build/ProjectReunion-Versions.yml
```diff
@@ -1,12 +1,6 @@
 variables:
   # Version Control
   major: "0"
-  minor: "5"
+  minor: "6"
```
DrusTheAxe, line 4: 8?

kythant: Just testing :)

DrusTheAxe (approved):

PR #809, opened 2021-05-07
Fixes for Preview VSIX

alwu-msft:
During testing, we discovered some issues with the preview VSIX:
* Missing project templates if both GA and Preview VSIX were installed side-by-side
Resolved by appending `.Preview` to TemplateID of all templates in preview VSIX
* Nuget restore error when creating new C++ projects
Resolved by ensuring that the Nuget template wizard data was pointing at the correct extension ID to use as a source repository
* Nuget compatibility error from .wapproj when using VS 16.10 Preview to build a Desktop WinUI app
The `AssetTargetFallback` property needs to be defined after the definition of the `TargetPlatformVersion` property as the former has a dependency on the latter.

Additionally, we decided to not set the `Preview` flag in the Preview VSIX's manifest as the title already declares it to be a preview.

### DIFF
--- a/dev/VSIX/Directory.Build.targets
+++ b/dev/VSIX/Directory.Build.targets
@@ -52,10 +52,32 @@
             <Output TaskParameter="Result" PropertyName="_tempVsTemplateDisplayName" />
         </XmlPeek>
 
+        <XmlPeek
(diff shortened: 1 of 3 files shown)


Scottj1s (approved):
thanks!

PR #814, opened 2021-05-11
MRTCore will also search for [modulename].pri for unpackaged app

huichen123:
For unpackaged app, the tool will build the resource file as [modulename].pri. It's not straightforward to update build tool. Change the runtime behavior to also look up [modulename].pri for unpackaged app.
MrtCoreUnpackagedTest only runs in command line, as the test adapter launches te.processhost.exe from nuget cache folder, thus we cannot copy manifest and pri file there.


dev/MRTCore/mrt/Core/src/MRM.cpp
```diff
@@ -899,24 +903,38 @@ STDAPI MrmGetFilePathFromName(_In_ PCWSTR filename, _Outptr_ PWSTR* filePath)
         pointerToPath++;
     }
 
+    pointerToPath = pathAllocated ? pathAllocated.get() : path;
```
rohanp-msft, line 906: >pointerToPath = pathAllocated ? pathAllocated.get() : path; [](http://example.com/codeflow?start=4&length=59)

Comment that this resets `pointerToPath` would be helpful.

rohanp-msft:
>     while ((length == size) && (lastError == ERROR_INSUFFICIENT_BUFFER))

It would be helpful for the reader if you sprinkled comments in the body. For example, a comment for this while block will help.

---
Refers to: dev/MRTCore/mrt/Core/src/MRM.cpp:878 in 18d0d84. [](commit_id = 18d0d845eb60b659dbc90810ba44c258a9a72826, deletion_comment = False)

dev/MRTCore/mrt/Core/src/MRM.cpp
```diff
@@ -942,14 +962,25 @@ STDAPI MrmGetFilePathFromName(_In_ PCWSTR filename, _Outptr_ PWSTR* filePath)
         {
             // If none of the file exists, will return the file in current path
```
rohanp-msft, line 963: >If none of the file exists [](http://example.com/codeflow?start=15&length=26)

Using "If the pending second iteration doesn't find the file either" instead of this would make it easier to understand this comment.

dev/MRTCore/mrt/Core/src/MRM.cpp
```diff
@@ -942,14 +962,25 @@ STDAPI MrmGetFilePathFromName(_In_ PCWSTR filename, _Outptr_ PWSTR* filePath)
...
+            // Prep for second iteration
+            if (filename == nullptr || *filename == L'\0')
+            {
+                // Change to [modulename].pri
+                RETURN_IF_FAILED(PathCchRenameExtension(moduleFilename, ARRAYSIZE(moduleFilename), L"pri"));
+                filenameToUse = moduleFilename;
+            }
+            else
+            {
+                // move to parent folder
```
rohanp-msft, line 975: >move [](http://example.com/codeflow?start=19&length=4)

Change the path

dev/MRTCore/mrt/Core/src/MRM.cpp
```diff
@@ -25,6 +25,7 @@ typedef struct
 
 constexpr wchar_t ResourceUriPrefix[] = L"ms-resource://";
 constexpr int ResourceUriPrefixLength = ARRAYSIZE(ResourceUriPrefix) - 1;
+constexpr wchar_t c_defaultResourceFilename[] = L"resources.pri";
```
rohanp-msft, line 28: >c_defaultResourceFilename [](http://example.com/codeflow?start=18&length=25)

`c_defaultResourceIndexFilename` or `c_defaultPriFilename` would be better IMO since resource file can mean resw files, image files, etc.

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/src/Helper.cpp
```diff
@@ -46,16 +62,25 @@ HRESULT GetDefaultPriFileForCurrentPackage(winrt::hstring& filePath)
...
+
     filePath = resourceFile;
 
     return S_OK;
 }
 
 HRESULT GetDefaultPriFile(winrt::hstring& filePath)
 {
-    if (FAILED(GetDefaultPriFileForCurrentPackage(filePath)))
+    HRESULT hr = GetDefaultPriFileForCurrentPackage(filePath);
```
rohanp-msft, line 78: >GetDefaultPriFileForCurrentPackage [](http://example.com/codeflow?start=17&length=34)

If the app is packaged, will this function try to get the path of resources.pri? Adding that as a comment would be helpful. From the name "ResourceManager_GetDefaultResourcePathForPackageFullName", I can't tell.

Also, it would be good to have a comment saying why we fallback even in the packaged case. I mean, the name of the first function implies it handles the packaged case, yet we call the second function sometimes with `isPackaged = true`.

rohanp-msft:
>     TEST_METHOD(InvalidPriName)

Is <module_name>.pri always considered valid?

---
Refers to: dev/MRTCore/mrt/Core/unittests/MrmTests.cpp:513 in 18d0d84. [](commit_id = 18d0d845eb60b659dbc90810ba44c258a9a72826, deletion_comment = False)

dev/MRTCore/mrt/Core/src/MRM.cpp
```diff
@@ -899,24 +903,38 @@ STDAPI MrmGetFilePathFromName(_In_ PCWSTR filename, _Outptr_ PWSTR* filePath)
         pointerToPath++;
     }
 
+    pointerToPath = pathAllocated ? pathAllocated.get() : path;
+
+    wchar_t moduleFilename[MAX_PATH];
+    RETURN_IF_FAILED(StringCchCopy(moduleFilename, ARRAYSIZE(moduleFilename), lastSlash != nullptr ? lastSlash + 1 : pointerToPath));
```
jefgen, line 909: nit: The rest of this function seems to use `StringCchCopyW` directly (instead of `StringCchCopy`).

dev/MRTCore/mrt/Core/src/MRM.cpp
huichen123: i don't think we want comment that just repeats what the code is doing.

dev/MRTCore/mrt/Core/unittests/MrmTests.cpp
```diff
@@ -560,6 +560,23 @@ TEST_CLASS(BasicTest)
...
 
+    TEST_METHOD(GetFilePath)
+    {
+        wchar_t* path;
+        Assert::AreEqual(MrmGetFilePathFromName(L"resources.pri", &path), S_OK);
+        Assert::IsNotNull(wcsstr(path, L"resources.pri"));
+        MrmFreeResource(path);
+
+        Assert::AreEqual(MrmGetFilePathFromName(L"something.pri", &path), S_OK);
+        // Even if the file doesn't exist, we will still return a path. This is for those don't use PRI for resource.
```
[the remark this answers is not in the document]
huichen123: an existing app may want to migrate to switch to WinUI3, but not ready to switch the resource management yet. They can use ResourceNotFound event to hook up their existing resource management.

huichen123:
>     TEST_METHOD(InvalidPriName)

The name should really be NonexistPriName

---
In reply to: [837786596](https://github.com/microsoft/ProjectReunion/pull/814#issuecomment-837786596) [](http://example.com/codeflow?ancestors=837786596,837786596)

---
Refers to: dev/MRTCore/mrt/Core/unittests/MrmTests.cpp:513 in 18d0d84. [](commit_id = 18d0d845eb60b659dbc90810ba44c258a9a72826, deletion_comment = False)

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/UnpackagedTests/TestClass.cs
```diff
@@ -0,0 +1,173 @@
...
+        public void DefaultResourceManagerWithExePri()
+        {
+            File.Copy(Path.Combine(m_assemblyFolder, "resources.pri.standalone"), Path.Combine(m_assemblyFolder, "te.processhost.pri"));
+
+            var resourceManager = new ResourceManager();
+            var resourceMap = resourceManager.MainResourceMap;
+            var map = resourceMap.GetSubtree("resources");
+            Verify.AreNotEqual(map.ResourceCount, 0u);
+            var resource = map.GetValue("IDS_MANIFEST_MUSIC_APP_NAME").ValueAsString;
+            Verify.AreEqual(resource, "Groove Music");
```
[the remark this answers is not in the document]
huichen123: it's still short enough. I'd prefer a test is self contained.

rohanp-msft:
One question: in the wapproj case, will resources.pri exist alongside <csproj/vcxproj name>.pri? (resources.pri is a merge of <csproj/vcxproj name>.pri and other stuff) *Assuming* it does, if there is an issue generating resources.pri, our code will now succeed (it'll fall back to <csproj/vcxproj name>.pri). Don't we want it to fail instead?

dev/MRTCore/mrt/Core/src/MRM.cpp
rohanp-msft: ok

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/UnpackagedTests/TestClass.cs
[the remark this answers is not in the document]
rohanp-msft: ok

dev/MRTCore/mrt/Core/unittests/MrmTests.cpp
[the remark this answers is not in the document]
rohanp-msft: Adding that or part of that as a comment would be good, IMO.

huichen123:
1. if merge fails, the build would break
2. for packaged app, we will not use [project name].pri.

---
In reply to: [840038734](https://github.com/microsoft/ProjectReunion/pull/814#issuecomment-840038734) [](http://example.com/codeflow?ancestors=840038734,840038734)

rohanp-msft:
Regarding #2 above, with wapproj-less packaged apps, we do want to use <project name>.pri, don't we? There won't be a resources.pri in that case, no?

---
In reply to: [840038734](https://github.com/microsoft/ProjectReunion/pull/814#issuecomment-840038734) [](http://example.com/codeflow?ancestors=840038734)

build/build-mrt.yml
```diff
@@ -145,6 +145,7 @@ steps:
       !**\*TestAdapter.dll
       !**\TE.*.dll
       !**\obj\**
+      !**\MrtCoreUnpackagedTests.dll
```
axelandrejs, line 148: Very nice!

dev/MRTCore/mrt/Core/src/MRM.cpp
```diff
@@ -853,18 +854,21 @@ STDAPI_(void) MrmFreeResource(_In_opt_ void* resource)
...
+// path. If the file doesn't exist, search [modulename].pri instead.
+// For unpackaged app, the built resource name is [modulename].pri. We still search resources.pri
+// because that is a supported scenario for inbox MRT (Xaml islands).
+STDAPI MrmGetFilePathFromName(_In_opt_ PCWSTR filename, _Outptr_ PWSTR* filePath)
 {
     *filePath = nullptr;
 
-    RETURN_HR_IF(E_INVALIDARG, filename == nullptr || *filename == L'\0');
-
     wchar_t path[MAX_PATH];
```
axelandrejs, line 872: Maybe not as part of this PR, but we should remove usages of MAX_PATH.

axelandrejs (approved):

huichen123:
for any packaged app (wap or not), the resource file needs to be resources.pri. If it's not, we need to make that happen, or the app may have issue with system MRT.
I think 2 is correct.

---
In reply to: [840116910](https://github.com/microsoft/ProjectReunion/pull/814#issuecomment-840116910) [](http://example.com/codeflow?ancestors=840116910,840038734)

dev/MRTCore/mrt/Core/src/MRM.cpp
```diff
@@ -899,24 +903,38 @@ STDAPI MrmGetFilePathFromName(_In_ PCWSTR filename, _Outptr_ PWSTR* filePath)
...
+    }
+
     // Will build the path at most twice.
-    // First time using current path. If not exist, do another time with parent path.
+    // If filename is provided:
+    //   - search under exe path
+    //   - if not exist, search parent path
+    // If filename is not provided:
+    //   - search under exe path with default name (resources.pri)
+    //   - if not exist, search under same path with [modulename].pri
```
alwu-msft, line 930: Why wouldn't we want to probe the parent directory if relying upon default filenames?

dev/MRTCore/mrt/Core/unittests/MrmTests.cpp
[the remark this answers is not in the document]
alwu-msft: Agreed. The comment on 963-964 of MRM.cpp was not enlightening because it didn't explain _why_ that was the desired behavior.

dev/MRTCore/mrt/Core/src/MRM.cpp
huichen123: i don't think we have a case where the pri file is created as [project name].pri and in the parent directory of exe.

rohanp-msft (approved):
:shipit:

dev/MRTCore/mrt/Core/src/MRM.cpp
alwu-msft: But `resources.pri` could be in the parent directory. I find it odd that a developer relying on the default behavior of MRT Core (by not specifying a filename) won't be able to handle one of the possible default behaviors of the rest of the stack (.wapproj putting the `resources.pri` in the parent directory). Am I misunderstanding how that scenario is intended to play out?

huichen123: for packaged app, "resources.pri" is passed into this API, and parent folder will be searched. The detail is encapsulated from the developer using ResourceManager.

alwu-msft: Got it, thanks for the explanation!

PR #821, opened 2021-05-14
[Draft] Add PushNotifications unit tests

pmpurifoy:


test/TestApps/PushNotificationsTestApp/main.cpp
```diff
@@ -0,0 +1,388 @@
...
+            AsyncStatus const /* asyncStatus */)
+        {
+            auto result = sender.GetResults();
+            if (result.Status() == PushNotificationChannelStatus::CompletedSuccess)
+            {
+                auto channelUri = result.Channel().Uri();
+                auto channelExpiry = result.Channel().ExpirationTime();
+            }
+            else if (result.Status() == PushNotificationChannelStatus::CompletedFailure)
+            {
```
danielayala94, line 95: What happens if by some reason there is no internet connection and the channel request ends up in a completed failure? Looks like we still return S_OK when it shouldn't right?

test/TestApps/PushNotificationsTestApp/main.cpp
```diff
@@ -0,0 +1,388 @@
...
+    channelOperation.Completed(
+        [&channelEvent](
+            IAsyncOperationWithProgress<PushNotificationCreateChannelResult, PushNotificationCreateChannelStatus> const& sender,
+            AsyncStatus const /* asyncStatus */)
+        {
+            auto result = sender.GetResults();
+            if (result.Status() == PushNotificationChannelStatus::CompletedSuccess)
+            {
+                auto channelUri = result.Channel().Uri();
+                auto channelExpiry = result.Channel().ExpirationTime();
```
danielayala94, line 92: Should we at least print these values? Otherwise this unit test it feels like a black box that returns S_OK but we don't know if got a channel. See my related comment on the failure scenario.

test/TestApps/PushNotificationsTestApp/main.cpp
```diff
@@ -0,0 +1,388 @@
...
+
+    // The maximum amount of time it takes for channel request to be obtained - 16mins
+    if (WAIT_OBJECT_0 != WaitForSingleObject(channelEvent.get(), 960000 /* milliseconds */))
+    {
+        channelOperation.Cancel();
+        hr = ERROR_TIMEOUT;
+    }
+    else
+    {
+        channelOperation.Close(); // Do not call getresults after this
```
danielayala94, line 109: What happens if Close() is not called?

test/TestApps/PushNotificationsTestApp/main.cpp
```diff
@@ -0,0 +1,388 @@
...
+    auto hr = S_OK;
+
+    auto channelOperation1 = PushNotificationManager::CreateChannelAsync(remoteId1);
+    auto channelOperation2 = PushNotificationManager::CreateChannelAsync(remoteId1);
+
+    channelOperation2.Completed(
+        [&channelEvent, &hr](
+            IAsyncOperationWithProgress<PushNotificationCreateChannelResult, PushNotificationCreateChannelStatus> const& sender,
+            AsyncStatus const /* asyncStatus */)
+        {
```
danielayala94, line 127: How are we sure that we the HRESULT from the second channel request is going to be WPN_E_OUTSTANDING_CHANNEL_REQUEST? Is it because the first channel request will take long enough?

danielayala94: What if the first channel request returns a cached channel (thinking on a request that is processed really quick)? How are we preventing this scenario in the tests?

test/TestApps/PushNotificationsTestApp/main.cpp
```diff
@@ -0,0 +1,388 @@
...
+    }
+    else
+    {
+        channelOperation.Close(); // Do not call getresults after this
+    }
+
+    return (hr == S_OK);
+}
+
+bool MultipleChannelRequestUsingSameRemoteId()
```
danielayala94, line 115: Is the purpose of the test to just verify that the second channel request has to return WPN_E_OUTSTANDING_CHANNEL_REQUEST, right?

test/TestApps/PushNotificationsTestApp/main.cpp
```diff
@@ -0,0 +1,388 @@
...
+            PushNotificationRegistrationKind::PushTrigger | PushNotificationRegistrationKind::ComActivator,
+            winrt::guid("00000000-0000-0000-0000-000000000001"));
+
+        auto token = PushNotificationManager::RegisterActivator(info);
+        if (token.Cookie() == 0 || token.TaskRegistration() == nullptr)
+        {
+            return false;
+        }
+
+        PushNotificationManager::UnregisterActivator(token, PushNotificationRegistrationKind::ComActivator);
```
danielayala94, line 304: If we are registering a task, we should unregister it. Right now this is only cleaning the COM registration.

test/TestApps/PushNotificationsTestApp/main.cpp
```diff
@@ -258,19 +257,26 @@ bool ThreeChannelRequestUsingSameRemoteId()
...
         PushNotificationActivationInfo info(
             PushNotificationRegistrationKind::PushTrigger | PushNotificationRegistrationKind::ComActivator,
-            winrt::guid("00000000-0000-0000-0000-000000000001"));
+            c_fakeComServerId);
 
         auto token = PushNotificationManager::RegisterActivator(info);
-        if (token.Cookie() == 0 || token.TaskRegistration() == nullptr)
+        if (token.Cookie() == 0)
+        {
+            return false;
```
hulumane, line 272: Why would this return 0?

test/TestApps/PushNotificationsTestApp/main.cpp
```diff
@@ -258,19 +257,26 @@ bool ThreeChannelRequestUsingSameRemoteId()
...
+        }
+
+        if (token.TaskRegistration() == nullptr)
```
hulumane, line 275: !token.TaskREgistration()

test/TestApps/PushNotificationsTestApp/main.cpp
```diff
@@ -280,13 +286,14 @@ bool RegisterActivator()
...
 {
+    PushNotificationManager::UnregisterActivator(token, PushNotificationRegistrationKind::PushTrigger | PushNotificationRegistrationKind::ComActivator);
     try
     {
         PushNotificationActivationInfo info(
             PushNotificationRegistrationKind::PushTrigger | PushNotificationRegistrationKind::ComActivator,
-            winrt::guid("00000000-0000-0000-0000-000000000001"));
+            c_fakeComServerId);
 
         auto token = PushNotificationManager::RegisterActivator(info);
```
hulumane, line 298: Bit confused as to why we need a RegisterActivator in an UnREgisterActivator test

test/PushNotificationTests/APITests.cpp
```diff
@@ -0,0 +1,184 @@
...
+            WaitForEvent(event, m_failed);
+        }
+
+        TEST_METHOD(BackgroundActivation)
+        {
+            wil::unique_event event = CreateTestEvent(c_testProtocolScheme_Packaged);
+            auto LocalBackgroundTask = winrt::create_instance<winrt::Windows::ApplicationModel::Background::IBackgroundTask>(c_comServerId, CLSCTX_ALL);
+            auto mockBackgroundTaskInstance = winrt::make<MockBackgroundTaskInstance>();
+            LocalBackgroundTask.Run(mockBackgroundTaskInstance);
+            WaitForEvent(event, m_failed);
```
hulumane, line 161: When does this failed event fire? Why is it passed and what is it's purpose?

hulumane: I guess protocolevent is fired for success and failedevent is fired for failure? And this method basically verifies failure or success?

test/PushNotificationTests/APITests.cpp
```diff
@@ -0,0 +1,184 @@
...
+        }
+
+        TEST_METHOD(RegisterActivator)
```
hulumane, line 164: I think we should combine the REgistActivator and UnregisterActivator test. It makes sense to combine the 2 because Unregister takes the token from Register.

test/PushNotificationTests/APITests.cpp
```diff
@@ -0,0 +1,184 @@
...
+            VERIFY_IS_TRUE(launchResult);
+
+            WaitForEvent(event, m_failed);
+        }
+
+        TEST_METHOD(BackgroundActivation)
+        {
+            wil::unique_event event = CreateTestEvent(c_testProtocolScheme_Packaged);
+            auto LocalBackgroundTask = winrt::create_instance<winrt::Windows::ApplicationModel::Background::IBackgroundTask>(c_comServerId, CLSCTX_ALL);
+            auto mockBackgroundTaskInstance = winrt::make<MockBackgroundTaskInstance>();
```
hulumane, line 159: Can you also please verify that the payload that you pass into the Mock here is the same payload that is extracted when we launch the app?

test/TestApps/PushNotificationsTestApp/main.cpp
sharath2727: Yes

test/TestApps/PushNotificationsTestApp/main.cpp
sharath2727: Let me take you to the design back again:

The sequence of events happen in the order as soon as you call CreateChannelAsync:

1. Guid null check
2. check for a remoteId that is currently being serviced, if none, then we set the s_remoteIdInProgress = true; for the first request.
3. co_await - Resume_background - Allows the second request to come in
4. while the first one is being serviced, the second one is checking for remoteIdInProgress, as it is true it immediately returns WPN_E_OUTSTANDING_CHANNEL_REQUEST. The second one would be pretty quick operation in comparison to the first one as the first one goes across the COM no matter if the channel is cached. So far I have had success in this test case passing 100% of the time.

test/TestApps/PushNotificationsTestApp/main.cpp
sharath2727: nothing is going to happen. Just a good practice and that is why I added an extra comment since I am calling close on what not to do after that.

test/TestApps/PushNotificationsTestApp/main.cpp
sharath2727: I need to check with Paul to see how we can print on the taef test command line screen. Let me know if you are aware of any such APIs that would assist here. Then I can add the print for these values. Can I fix this going ahead probably in the next round?

test/TestApps/PushNotificationsTestApp/main.cpp
sharath2727: Ah that is bug will fix it thanks.

test/PushNotificationTests/APITests.cpp
```diff
@@ -0,0 +1,184 @@
...
+            LocalBackgroundTask.Run(mockBackgroundTaskInstance);
```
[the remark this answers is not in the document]
sharath2727: @pmpurifoy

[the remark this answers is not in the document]
pmpurifoy: Yup! create_instance is the winrt version of CoCreateInstance. Calling Run(MockBackgroundTaskInstance) will trigger the app through background activation. No protocol is needed to launch the app.

test/PushNotificationTests/APITests.cpp
pmpurifoy: Yes. WaitForEvent will wait for the named event and if it has not either heard back success/failure for 10 seconds, it will fail.

test/PushNotificationTests/APITests.cpp
pmpurifoy: Yes. This is already being checked in the ExtendedActivationKind::Push block. I compare the the payload from PushNotificationReceivedEventArgs with the c_rawNotificationPayload listed in TestDef.h

test/TestApps/PushNotificationsTestApp/main.cpp
pmpurifoy: I thought that CoRegisterClassObject in PushNotificationManager would return 0 in the case of failing to register but we already have a THROW_IF_FAILED wrapping that function, so it looks like we don't to check for Cookie being 0 here.

test/TestApps/PushNotificationsTestApp/main.cpp
hulumane: I think it's important to Close the channel after we are done. Otherwise, subsequent tests will use the cached channel in the platform

test/TestApps/PushNotificationsTestApp/main.cpp
danielayala94: Yes. Let's follow up later. Will resolve for now.

test/TestApps/PushNotificationsTestApp/main.cpp
sharath2727: Done

danielayala94 (approved):

hulumane (approved):

PR #827, opened 2021-05-17
Update Nuspecs for unpackaged consumption

DrusTheAxe:
Correct some Nuspec files for VS C++ consumption for unpackaged apps

Correct Nuget content determined via VS simple C++ project with a dependency on Project Reunion followed by repeated hacking the NuGet's native build files until the desired outcome, then porting the local changes to the checked in Nuspecs. Fun fun.

### DIFF
--- a/build/NuSpecs/ProjectReunion-Nuget-Native.C.props
+++ b/build/NuSpecs/ProjectReunion-Nuget-Native.C.props
@@ -6,10 +6,6 @@
     <_ProjectReunionFoundationPlatform Condition="'$(Platform)' != 'Win32'">$(Platform)</_ProjectReunionFoundationPlatform>
   </PropertyGroup>
 
-  <!--
-    Some of the manual reference adding below may not be necessary due to the files in question being in default locations, but there's no harm in it.
-  -->
-
   <ItemDefinitionGroup>
     <ClCompile>
       <AdditionalIncludeDirectories>$(MSBuildThisFileDirectory)..\..\include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
@@ -20,10 +16,21 @@
     <Link>
       <AdditionalDependencies>
         $(MSBuildThisFileDirectory)..\..\lib\win10-$(_ProjectReunionFoundationPlatform)\Microsoft.ProjectReunion.lib;
+        %(AdditionalDependencies);
+      </AdditionalDependencies>
+    </Link>
+  </ItemDefinitionGroup>
+
+  <ItemDefinitionGroup Condition="'$(AppxPackage)' != 'true'">
+    <Link>
+      <AdditionalDependencies>
         $(MSBuildThisFileDirectory)..\..\lib\win10-$(_ProjectReunionFoundationPlatform)\Microsoft.ProjectReunion.Bootstrap.lib;
         %(AdditionalDependencies);
       </AdditionalDependencies>
     </Link>
+    <PostBuildEvent>
+      <Command>xcopy /y "$(MSBuildThisFileDirectory)..\..\runtimes\lib\native\$(_ProjectReunionFoundationPlatform)\Microsoft.ProjectReunion.Bootstrap.dll" "$(OutDir)"</Command>
+    </PostBuildEvent>
   </ItemDefinitionGroup>
 
 </Project>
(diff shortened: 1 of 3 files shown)


build/NuSpecs/ProjectReunion-Nuget-Native.C.props
```diff
@@ -20,10 +16,21 @@
...
+
+  <ItemDefinitionGroup Condition="'$(AppxPackage)' != 'true'">
+    <Link>
+      <AdditionalDependencies>
         $(MSBuildThisFileDirectory)..\..\lib\win10-$(_ProjectReunionFoundationPlatform)\Microsoft.ProjectReunion.Bootstrap.lib;
         %(AdditionalDependencies);
       </AdditionalDependencies>
     </Link>
+    <PostBuildEvent>
+      <Command>xcopy /y "$(MSBuildThisFileDirectory)..\..\runtimes\lib\native\$(_ProjectReunionFoundationPlatform)\Microsoft.ProjectReunion.Bootstrap.dll" "$(OutDir)"</Command>
```
aeloros, line 32: Is this inside the $(OutDir)/buildoutput?  We really should avoid this type of relative reference.  Relative file paths are terrible for refactoring type of changes which are typical this early in a repo's life.  Please file an issue to fix this later.  I don't want to block this change for such a thing.

build/NuSpecs/ProjectReunion-Nuget-Native.WinRt.props
```diff
@@ -6,20 +6,20 @@
     <_ProjectReunionFoundationPlatform Condition="'$(Platform)' != 'Win32'">$(Platform)</_ProjectReunionFoundationPlatform>
```
aeloros, line 6: I'm pretty sure these are unneeded (I know you didn't add them).  I believe the proper one to use is $(PlatformTarget).

build/NuSpecs/ProjectReunion-Nuget-Native.WinRt.props
```diff
@@ -6,20 +6,20 @@
...
+  <ItemGroup Condition="'$(AppxPackage)' != 'true'">
     <Reference Include="Microsoft.ApplicationModel.DynamicDependency.winmd">
-      <HintPath>$(MSBuildThisFileDirectory)..\..\lib\uap10.0\Microsoft.ApplicationModel.DynamicDependency.winmd</HintPath>
+      <HintPath>$(MSBuildThisFileDirectory)..\..\lib\native\Microsoft.ApplicationModel.DynamicDependency.winmd</HintPath>
       <Implementation Condition="'$(ProjectReunionFrameworkPackage)' != 'true'">$(MSBuildThisFileDirectory)..\..\runtimes\win10-$(_ProjectReunionFoundationPlatform)\native\Microsoft.ApplicationModel.DynamicDependency.dll</Implementation>
       <IsWinMDFile>true</IsWinMDFile>
     </Reference>
   </ItemGroup>
 
+  <ItemGroup>
```
aeloros, line 17: In general, we need to probably take a look at grooming this to be a LOT easier.  Adding things to the builds shouldn't require our developers to make this type of extensive changes.  Obviously it's just because of the infancy of the project, but I would still like to just call it out.

aeloros (approved):

build/NuSpecs/ProjectReunion-Nuget-Native.C.props
DrusTheAxe: It's self-relative.

MSBuildThisFileDirectory is literally this file - ProjectReunion-Nuget-Native.C.props. The relative path'ing is to back us up past build/Nuspecs so we can drill down into runtimes\...

If you know a $(TheRootOfThePackageContainingThisFile) instead that'd be better but when in Rome...

build/NuSpecs/ProjectReunion-Nuget-Native.WinRt.props
DrusTheAxe: This is to deal with VS' penchant for using "Win32" in project configs rather than "x86"
I think you're right, PlatformTarget would do the equivalent, but best left for an all up pass over the Nuspecs (that'll impact not-native files as well)

build/NuSpecs/ProjectReunion-Nuget-Native.WinRt.props
DrusTheAxe: This was based on previous work (MRT) but yeah, I'm unsure how much of that was necessary or just what-seemed-like-a-good-idea-at-the-time. We'll want to take an overall deepthink across all the Nuspecs\* tree. If we're going to dynamically generate content for appxmanifest.xml we should be dynamically generating a bunch of content here too.

Something for a later day

PR #831, opened 2021-05-18
Fix DynamicDependencies generated build overrides

DrusTheAxe:
Build\override support wasn't being picked up correctly, causing an inconsistency between the DDLM registering one CLSID and the Bootstrap API looking for another. This appears to correct the problem, but need to turn a build to verify E2E

FYI ProjectReunion.props exists but appears dead - not referenced by anything I can find, including source tree searches and detailed log reviews via MSBuild Structured Log Viewer. Unless it's actually used by some arcane machinery I can't find it should be deleted in a future PR


ProjectReunion.props
```diff
@@ -11,13 +11,24 @@
     <!-- If not set, default it to no . -->
     <ProjectReunionBuildPipeline>0</ProjectReunionBuildPipeline>
   </PropertyGroup>
+  <Target Name="DirectoryBuildPropsInfo">
+    <Message Importance="High" Text="ProjectReunion.props detects ProjectReunionBuildPipeline=$(ProjectReunionBuildPipeline)"/>
+  </Target>
 
   <!-- Conditional behavior for build pipeline vs local development -->
```
DrusTheAxe, line 18: What is this file used for? It doesn't seem to be used by local VS development forcing me to add to Directory.Build.props but obviously used in pipeline builds else it bad code caused errors. The comment on line 34 seems to reinforce this.

Yes the override info's duplicated right now. There was more in this file I wasn't sure of why or if it's safe for local development so worst case we've got some duplicate information we can refactor later. If this file's purely for pipeline builds then it should be renamed something more specific.

Regardless, the changes unblock critical path work. This cleanup can be done at a later date

kythant (approved):

ProjectReunion.props
kythant: @mikenelte This would be a good question for Michael who was in the original V team to bring up most of this. I've never touch this file before.

PR #835, opened 2021-05-18
Convert MRT Core UnitTests from MSTest to TAEF

jefgen:
This PR converts the MRT Core Unit Tests from the MSTest framework to TAEF.
This will enable running the Unit Tests inside Helix in the future in another PR.

There are 4 different test suites for MRT Core currently:
- BaseUnitTests 
  - Already using TAEF.
  - This already runs in Helix but the test results aren't reported in ADO.
- UnitTests
  - This PR converts this test from MSTest to TAEF, and runs this test on the build machine.
- ManagedTests
  - Uses MSTest/vstest.console.exe to run.
  - Will investigate converting this to TAEF in a future PR.
- UnpackagedTests
  - Already using TAEF.
  - Doesn't yet run in the CI pipeline. (I plan to try and enable this in a future PR).

#### Notes on the changes:
I had to use `wcscmp` for comparing strings in the tests, as otherwise the `VERIFY_` macro would get confused and try to compare the value of the pointer instead of the string itself. Another possible option that I tried was to wrap each pointer with `std::wstring` (which is what some other tests in the repo do). However, it seemed a bit wasteful to construct a temporary string if we can just call `wcscmp`.

~~We can't invoke the converted TAEF UnitTest via the `VSTest` Azure DevOps Task in the pipeline, and need to invoke `te.exe` directly. The UnitTest expects to be invoked from the same directory as the test binaries, such that `resources.pri` is alongside the `UnitTest.dll` file. If it is invoked from another directory, then the tests will fail.~~
~~This means that we can't invoke the test using the ADO VSTest task (which would use the TAEF VS Test adapter) as that would have a different current directory. Unfortunately, there is no way (from what I can tell) to change the current working directory with the VSTest task.~~
~~The VSTest task *always* uses this as  the current working directory: `$(System.DefaultWorkingDirectory)`
I looked into the code for `VSTest@2`, and it always sets the working directory before executing:
https://github.com/microsoft/azure-pipelines-tasks/blob/master/Tasks/VsTestV2/vstest.ts#L221
Sadly there is no option to change this.
There is also no way to use a custom "runsettings" file to change the working directory either.~~

While working on this, I noticed that the WinUI/MUX Helix conversion scripts that are used for converting the TAEF WTL output to xUnit don't seem to actually be working. Note that if the test fails then the build pipeline will still fail, so we aren't losing any test coverage. However, this does mean that we lack the nice/pretty reporting features that the ADO UI provides. (But you can still see the test output in the pipeline results). The converted xUnit files seem okay on the surface, so I'll need to investigate more as to why this isn't working.

### DIFF
--- a/build/build-mrt.yml
+++ b/build/build-mrt.yml
@@ -249,3 +249,4 @@ steps:
   inputs:
     PathtoPublish: '$(Build.ArtifactStagingDirectory)\mrt_test'
     ArtifactName: 'mrtcore_test'
+  condition: or(eq(variables['BuildPlatform'], 'x64'), eq(variables['BuildPlatform'], 'x86'))
--- a/build/build-mrt.yml
+++ b/build/build-mrt.yml
@@ -151,9 +151,7 @@ steps:
   inputs:
     targetType: filePath
     filePath: build\Helix\ConvertWttLogToXUnit.ps1
-    arguments: -WttInputPath $LogPath -WttSingleRerunInputPath 'unused.wtl' -WttMultipleRerunInputPath 'unused2.wtl' -XUnitOutputPath 'onBuildMachineResults.xml' -TestNamePrefix '$(BuildConfiguration).$(BuildPlatform)'
-  env:
-    LogPath: ${{ parameters.MRTBinariesDirectory }}\$(buildConfiguration)\$(buildPlatform)\UnitTest\MrtUnitTest.wtl
+    arguments: -WttInputPath '${{ parameters.MRTBinariesDirectory }}\$(buildConfiguration)\$(buildPlatform)\UnitTest\MrtUnitTest.wtl' -WttSingleRerunInputPath 'unused.wtl' -WttMultipleRerunInputPath 'unused2.wtl' -XUnitOutputPath 'onBuildMachineResults.xml' -TestNamePrefix '$(BuildConfiguration).$(BuildPlatform)'
   condition: or(eq(variables['BuildPlatform'], 'x64'), eq(variables['BuildPlatform'], 'x86'))
 
 - task: PublishTestResults@2
--- a/build/build-mrt.yml
+++ b/build/build-mrt.yml
@@ -135,53 +135,53 @@ steps:
     platform: '$(buildPlatform)'
     configuration: '$(buildConfiguration)'
 
-- task: CmdLine@2
-  displayName: 'test MRT - UnitTest'
-  inputs:
-    script: te.exe UnitTest.dll /enablewttlogging /appendwttlogging /logFile:%LogPath%
-    workingDirectory: '${{ parameters.MRTBinariesDirectory }}\$(buildConfiguration)\$(buildPlatform)\UnitTest'
-  env:
-    LogPath: ${{ parameters.MRTBinariesDirectory }}\$(buildConfiguration)\$(buildPlatform)\UnitTest\MrtUnitTest.wtl
-  condition: or(eq(variables.buildPlatform, 'x64'), eq(variables.buildPlatform, 'x86'))
+# - task: CmdLine@2
+#   displayName: 'test MRT - UnitTest'
+#   inputs:
+#     script: te.exe UnitTest.dll /enablewttlogging /appendwttlogging /logFile:%LogPath%
+#     workingDirectory: '${{ parameters.MRTBinariesDirectory }}\$(buildConfiguration)\$(buildPlatform)\UnitTest'
+#   env:
+#     LogPath: ${{ parameters.MRTBinariesDirectory }}\$(buildConfiguration)\$(buildPlatform)\UnitTest\MrtUnitTest.wtl
+#   condition: or(eq(variables.buildPlatform, 'x64'), eq(variables.buildPlatform, 'x86'))
 
-- task: PowerShell@2
-  displayName: 'Convert Test Logs from WTL to xUnit format'
-  inputs:
-    targetType: filePath
-    filePath: build\Helix\ConvertWttLogToXUnit.ps1
-    arguments: -WttInputPath '$env:LogPath\MrtUnitTest.wtl' -XUnitOutputPath '$env:LogPath\onBuildMachineResults.xml' -TestNamePrefix '$(BuildConfiguration).$(BuildPlatform)' -HelixWorkItem 0
-  env:
-    LogPath: ${{ parameters.MRTBinariesDirectory }}\$(buildConfiguration)\$(buildPlatform)\UnitTest
-  condition: or(eq(variables['BuildPlatform'], 'x64'), eq(variables['BuildPlatform'], 'x86'))
+# - task: PowerShell@2
+#   displayName: 'Convert Test Logs from WTL to xUnit format'
+#   inputs:
+#     targetType: filePath
+#     filePath: build\Helix\ConvertWttLogToXUnit.ps1
+#     arguments: -WttInputPath '$env:LogPath\MrtUnitTest.wtl' -XUnitOutputPath '$env:LogPath\onBuildMachineResults.xml' -TestNamePrefix '$(BuildConfiguration).$(BuildPlatform)' -HelixWorkItem 0
+#   env:
+#     LogPath: ${{ parameters.MRTBinariesDirectory }}\$(buildConfiguration)\$(buildPlatform)\UnitTest
+#   condition: or(eq(variables['BuildPlatform'], 'x64'), eq(variables['BuildPlatform'], 'x86'))
 
-- task: PublishTestResults@2
-  displayName: 'Upload converted test logs'
-  inputs:
-    testResultsFormat: 'xUnit'
-    testResultsFiles: '**/onBuildMachineResults.xml' 
-    searchFolder: '${{ parameters.MRTBinariesDirectory }}\$(buildConfiguration)\$(buildPlatform)'
-    testRunTitle: 'test MRT - On Build Machine Tests'
-    buildPlatform: $(BuildPlatform)
-    buildConfiguration: $(BuildConfiguration)
-  condition: or(eq(variables['BuildPlatform'], 'x64'), eq(variables['BuildPlatform'], 'x86'))
+# - task: PublishTestResults@2
+#   displayName: 'Upload converted test logs'
+#   inputs:
+#     testResultsFormat: 'xUnit'
+#     testResultsFiles: '**/onBuildMachineResults.xml' 
+#     searchFolder: '${{ parameters.MRTBinariesDirectory }}\$(buildConfiguration)\$(buildPlatform)'
+#     testRunTitle: 'test MRT - On Build Machine Tests'
+#     buildPlatform: $(BuildPlatform)
+#     buildConfiguration: $(BuildConfiguration)
+#   condition: or(eq(variables['BuildPlatform'], 'x64'), eq(variables['BuildPlatform'], 'x86'))
 
-- task: CopyFiles@2
-  displayName: 'test MRT - Copy test logs for Artifacts'
-  inputs:
-    Contents: |
-     **/*.wtl
-     **/*onBuildMachineResults.xml
-    TargetFolder: '$(Build.ArtifactStagingDirectory)\mrt_test\$(buildConfiguration)\$(buildPlatform)'
-    OverWrite: true
-    flattenFolders: true
-  condition: or(eq(variables['BuildPlatform'], 'x64'), eq(variables['BuildPlatform'], 'x86'))
+# - task: CopyFiles@2
+#   displayName: 'test MRT - Copy test logs for Artifacts'
+#   inputs:
+#     Contents: |
+#      **/*.wtl
+#      **/*onBuildMachineResults.xml
+#     TargetFolder: '$(Build.ArtifactStagingDirectory)\mrt_test\$(buildConfiguration)\$(buildPlatform)'
+#     OverWrite: true
+#     flattenFolders: true
+#   condition: or(eq(variables['BuildPlatform'], 'x64'), eq(variables['BuildPlatform'], 'x86'))
 
-- task: PublishBuildArtifacts@1
-  displayName: 'Publish Artifact: MRT Test results'
-  inputs:
-    PathtoPublish: '$(Build.ArtifactStagingDirectory)\mrt_test'
-    ArtifactName: 'mrtcore_test'
-  condition: or(eq(variables['BuildPlatform'], 'x64'), eq(variables['BuildPlatform'], 'x86'))
+# - task: PublishBuildArtifacts@1
+#   displayName: 'Publish Artifact: MRT Test results'
+#   inputs:
+#     PathtoPublish: '$(Build.ArtifactStagingDirectory)\mrt_test'
+#     ArtifactName: 'mrtcore_test'
+#   condition: or(eq(variables['BuildPlatform'], 'x64'), eq(variables['BuildPlatform'], 'x86'))
 
 - task: VSTest@2
   displayName: 'test MRT'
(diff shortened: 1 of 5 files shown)


huichen123:
"such that resources.pri is alongside the UnitTest.dll file"

The resources.pri should be alongside with unittest.dll. Or do you mean alongside with te.exe?

jefgen:
> "such that resources.pri is alongside the UnitTest.dll file"
> 
> The resources.pri should be alongside with unittest.dll. Or do you mean alongside with te.exe?

Sorry, I think I wasn't really clear. I can try to explain again.

Right, the `resources.pri` is indeed alongside the `unittest.dll` in the output `UnitTest` folder.
What I mean to say is that if you try to run the test from a different working directory than the UnitTest folder, then it won't find the `resources.pri` file. (It expects to find it in the current working directory).

For example, if you build MRT Core and then do this:
```
> cd ProjectReunion\BuildOutput\Release\x64
> UnitTest\te.exe UnitTest\UnitTest.dll
```
The test fails as it looks for `resources.pri` under `ProjectReunion\BuildOutput\Release\x64` instead of `ProjectReunion\BuildOutput\Release\x64\UnitTest`.

When the VSTest task runs in the pipeline the working directory is set to `$(System.DefaultWorkingDirectory)`, which is always the source checkout folder (ie: top of the repo), so it fails.

jefgen:
> "such that resources.pri is alongside the UnitTest.dll file"
> 
> The resources.pri should be alongside with unittest.dll. Or do you mean alongside with te.exe?

Sorry, I think I wasn't really clear. I can try to explain again.

Right, the `resources.pri` is indeed alongside the `unittest.dll` in the output `UnitTest` folder.
What I mean to say is that if you try to run the test from a different working directory than the UnitTest folder, then it won't find the `resources.pri` file. (It expects to find it in the current working directory).

For example, if you build MRT Core and then do this:
```
>cd ProjectReunion\BuildOutput\Release\x64
UnitTest\te.exe UnitTest\UnitTest.dll
```
The test fails as it looks for `resources.pri` under `ProjectReunion\BuildOutput\Release\x64` instead of `ProjectReunion\BuildOutput\Release\x64\UnitTest`.

When the VSTest task runs in the pipeline the working directory is set to `$(System.DefaultWorkingDirectory)`, which is always the source checkout folder (ie: top of the repo), so it fails.

huichen123:
can you add a setup in the test to updat the working directory?

---
In reply to: [843624704](https://github.com/microsoft/ProjectReunion/pull/835#issuecomment-843624704) [](http://example.com/codeflow?ancestors=843624704,843624704)

jefgen:
> can you add a setup in the test to updat the working directory?

I actually did think about doing that -- but I didn't know if you'd prefer doing that versus running the test directly from the command line. (I thought it might be better to make as few changes to the test as possible, but I can try doing that if you like).

huichen123:
1. fewer exceptions is preferred
2. otherwise we still have problem with test explorer

---
In reply to: [843631910](https://github.com/microsoft/ProjectReunion/pull/835#issuecomment-843631910) [](http://example.com/codeflow?ancestors=843631910,843631910)

dev/MRTCore/mrt/Core/unittests/UnitTest.vcxproj
```diff
@@ -346,7 +347,17 @@
...
   </ItemGroup>
+  <ItemGroup>
+    <None Include="packages.config" />
+  </ItemGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
-  <ImportGroup Label="ExtensionTargets">
+    <ImportGroup Label="ExtensionTargets">
+    <Import Project="..\..\packages\Microsoft.Taef.10.58.210222006-develop\build\Microsoft.Taef.targets" Condition="Exists('..\..\packages\Microsoft.Taef.10.58.210222006-develop\build\Microsoft.Taef.targets')" />
   </ImportGroup>
+  <Target Name="EnsureNuGetPackageBuildImports" BeforeTargets="PrepareForBuild">
```
rohanp-msft, line 357: >EnsureNuGetPackageBuildImports [](http://example.com/codeflow?start=16&length=30)

Just curious Jeff, did you add this to help the dev in case the package was absent? Instead of them getting a possibly mysterious error or something?

build/build-mrt.yml
```diff
@@ -135,6 +135,56 @@ steps:
...
+    arguments: -WttInputPath '${{ parameters.MRTBinariesDirectory }}\$(buildConfiguration)\$(buildPlatform)\UnitTest\MrtUnitTest.wtl' -XUnitOutputPath '${{ parameters.MRTBinariesDirectory }}\$(buildConfiguration)\$(buildPlatform)\UnitTest\onBuildMachineResults.xml' -TestNamePrefix '$(BuildConfiguration).$(BuildPlatform)' -HelixWorkItem 0
+  condition: or(eq(variables['BuildPlatform'], 'x64'), eq(variables['BuildPlatform'], 'x86'))
+
+- task: PublishTestResults@2
+  displayName: 'Upload converted test logs'
+  inputs:
+    testResultsFormat: 'xUnit' # Options: JUnit, NUnit, VSTest, xUnit, cTest
+    testResultsFiles: '**/onBuildMachineResults.xml' 
+    searchFolder: '${{ parameters.MRTBinariesDirectory }}\$(buildConfiguration)\$(buildPlatform)'
+    #mergeTestResults: false # Optional
```
rohanp-msft, line 162: Can we remove the commented lines in this task?

build/build-mrt.yml
```diff
@@ -135,6 +135,56 @@ steps:
     platform: '$(buildPlatform)'
     configuration: '$(buildConfiguration)'
 
+- task: CmdLine@2
+  displayName: 'test MRT - UnitTest'
+  inputs:
+    script: |
```
rohanp-msft, line 141: nitty nit: why the pipe + line break?

build/build-mrt.yml
```diff
@@ -135,6 +135,56 @@ steps:
...
+  env:
+    LogPath: ${{ parameters.MRTBinariesDirectory }}\$(buildConfiguration)\$(buildPlatform)\UnitTest\MrtUnitTest.wtl
+  condition: or(eq(variables.buildPlatform, 'x64'), eq(variables.buildPlatform, 'x86'))
+
+- task: PowerShell@2
+  displayName: 'Convert Test Logs from WTL to xUnit format'
+  inputs:
+    targetType: filePath
+    filePath: build\Helix\ConvertWttLogToXUnit.ps1
+    arguments: -WttInputPath '${{ parameters.MRTBinariesDirectory }}\$(buildConfiguration)\$(buildPlatform)\UnitTest\MrtUnitTest.wtl' -XUnitOutputPath '${{ parameters.MRTBinariesDirectory }}\$(buildConfiguration)\$(buildPlatform)\UnitTest\onBuildMachineResults.xml' -TestNamePrefix '$(BuildConfiguration).$(BuildPlatform)' -HelixWorkItem 0
```
rohanp-msft, line 153: Would be good to break up the very long lines like this so it's easier to read.

build/build-mrt.yml
```diff
@@ -135,6 +135,56 @@ steps:
...
+
+- task: PublishTestResults@2
+  displayName: 'Upload converted test logs'
+  inputs:
+    testResultsFormat: 'xUnit' # Options: JUnit, NUnit, VSTest, xUnit, cTest
+    testResultsFiles: '**/onBuildMachineResults.xml' 
+    searchFolder: '${{ parameters.MRTBinariesDirectory }}\$(buildConfiguration)\$(buildPlatform)'
+    #mergeTestResults: false # Optional
+    #failTaskOnFailedTests: false # Optional
+    testRunTitle: 'test MRT - On Build Machine Tests' # Optional
```
rohanp-msft, line 164: Is this a title of a previous test run? I don't see it above, so I guess not. Is this a title of this publish test run results task? If it is, would be good to give it a more generic name maybe? I say that because it seems as though this task grabs more than just the onBuildMachineResults.xml file to upload (it also grabs the xUnit file).

jefgen:
> 2. otherwise we still have problem with test explorer

Good point about the test explorer in VS. I'll try updating the test to change the current directory...

---
In reply to: [843649239](https://github.com/microsoft/ProjectReunion/pull/835#issuecomment-843649239) [](http://example.com/codeflow?ancestors=843649239,843631910)

build/build-mrt.yml
jefgen: sure, will do.

dev/MRTCore/mrt/Core/unittests/UnitTest.vcxproj
jefgen: I actually copied this from the TAEF sample test app -- but you're quite right, it gives a more user-friendly error message if the Microsoft.Taef nuget package isn't present. :)

build/build-mrt.yml
jefgen: I found it easier to read/see the actual command that gets executed, but I can make it all one line if you like.

build/build-mrt.yml
jefgen: I found it easier to read/see the actual command that gets executed, but I can make it all one line if you like.

rohanp-msft: No, it's fine. It's not important of course. And there's no right or wrong here.

build/build-mrt.yml
jefgen: Sure, I'll see if I can use a variable to make it smaller.

build/build-mrt.yml
jefgen: Actually, I was able to get the test to run via the ADO VSTest task, which means that I don't need to do any of this manual processing of the logs -- as the TAEF adapter handles this automatically when invoked via the ADO task.
I'll push an update out which removes this part.

jefgen:
**Update:**
I was able to get the converted TAEF UnitTest to run via the ADO `VSTest` task, which means that I no longer need to do any of the manual processing of the logs, since the TAEF adapter handles this automatically when invoked via the ADO task.

The other nice thing is that the converted test also works with the test explorer in VS:
image

So the changes in the PR are much simpler now.

rohanp-msft (approved):
Great change, thanks Jeff!

dev/MRTCore/mrt/Core/unittests/MrmTests.cpp
```diff
@@ -343,58 +370,58 @@ TEST_CLASS(BasicTest)
...
 
         MrmMapHandle childChildResourceMap;
-        Assert::AreEqual(MrmGetChildResourceMap(resourceManager, childResourceMap, L"Resources", &childChildResourceMap), S_OK);
+        VERIFY_ARE_EQUAL(MrmGetChildResourceMap(resourceManager, childResourceMap, L"Resources", &childChildResourceMap), S_OK);
 
         UINT32 count;
-        Assert::AreEqual(MrmGetResourceCount(resourceManager, childChildResourceMap, &count), S_OK);
-        Assert::AreEqual<UINT32>(count, 78);
+        VERIFY_ARE_EQUAL(MrmGetResourceCount(resourceManager, childChildResourceMap, &count), S_OK);
+        VERIFY_ARE_EQUAL(count, static_cast<UINT32>(78));
```
axelandrejs, line 384: NIT: You should be able to avoid the cast by doing 78u.

(due to the way the verify macros use templates the implicit conversion of a constant to an unsigned integer is lost; but you can fix that by explicitly marking the constant as unsigned)

dev/MRTCore/mrt/Core/unittests/MrmTests.cpp
```diff
@@ -343,58 +370,58 @@ TEST_CLASS(BasicTest)
...
+                VERIFY_ARE_EQUAL(wcscmp(resourceName, L"AutomationNameAlphaSlider"), 0);
+                VERIFY_ARE_EQUAL(wcscmp(resourceString, L"Opacity"), 0);
                 break;
             }
 
             case 1:
             {
-                Assert::AreEqual(resourceName, L"AutomationNameAlphaTextBox");
-                Assert::AreEqual(resourceString, L"Opacity");
+                VERIFY_ARE_EQUAL(wcscmp(resourceName, L"AutomationNameAlphaTextBox"), 0);
```
axelandrejs, line 409: We have this often enough that a helper macro might be beneficial. That would also have the ability to print the expected versus actual string, which the current macro would not do.

dev/MRTCore/mrt/Core/unittests/MrmTests.cpp
```diff
@@ -405,46 +432,46 @@ TEST_CLASS(BasicTest)
...
 
         // Enumerate under Files/Controls
-        Assert::AreEqual(MrmGetChildResourceMap(resourceManager, nullptr, L"Files", &childResourceMap), S_OK);
-        Assert::AreEqual(MrmGetChildResourceMap(resourceManager, childResourceMap, L"Controls", &childChildResourceMap), S_OK);
-        Assert::AreEqual(MrmGetResourceCount(resourceManager, childChildResourceMap, &count), S_OK);
-        Assert::AreEqual<UINT32>(count, 28);
+        VERIFY_ARE_EQUAL(MrmGetChildResourceMap(resourceManager, nullptr, L"Files", &childResourceMap), S_OK);
+        VERIFY_ARE_EQUAL(MrmGetChildResourceMap(resourceManager, childResourceMap, L"Controls", &childChildResourceMap), S_OK);
+        VERIFY_ARE_EQUAL(MrmGetResourceCount(resourceManager, childChildResourceMap, &count), S_OK);
+        VERIFY_ARE_EQUAL(count, static_cast<UINT32>(28));
```
axelandrejs, line 441: Same as above.

dev/MRTCore/mrt/Core/unittests/MrmTests.cpp
```diff
@@ -563,35 +590,37 @@ TEST_CLASS(BasicTest)
...
+        VERIFY_ARE_EQUAL(MrmGetFilePathFromName(nullptr, &path), S_OK);
+        VERIFY_IS_NOT_NULL(wcsstr(path, L"resources.pri"));
         MrmFreeResource(path);
     }
 
 private:
     void VerifyQualifierValue(UINT32 qualifierCount, PWSTR* qualifierNames, PWSTR* qualifierValues, PCWSTR name, PCWSTR expectedValue)
     {
-        Assert::IsTrue(qualifierCount > 0);
+        VERIFY_IS_TRUE(qualifierCount > 0);
```
axelandrejs, line 610: VERIFY_IS_GREATER_THAN

(this has the benefit of displaying the value on failure)

dev/MRTCore/mrt/Core/unittests/MrmTests.cpp
```diff
@@ -563,35 +590,37 @@ TEST_CLASS(BasicTest)
...
         bool found = false;
         for (UINT32 i = 0; i < qualifierCount; i++)
         {
             if (0 == _wcsicmp(name, qualifierNames[i]))
             {
-                Assert::IsTrue(0 == _wcsicmp(expectedValue, qualifierValues[i]));
+                VERIFY_IS_TRUE(0 == _wcsicmp(expectedValue, qualifierValues[i]));
```
axelandrejs, line 616: VERIFY_ARE_EQUAL

huichen123 (approved):

dev/MRTCore/mrt/Core/unittests/MrmTests.cpp
jefgen: Thanks! Good idea to use 78u instead of a cast.

dev/MRTCore/mrt/Core/unittests/MrmTests.cpp
jefgen: Good idea! I'll add a helper for comparing strings.

dev/MRTCore/mrt/Core/unittests/MrmTests.cpp
jefgen: Good point

dev/MRTCore/mrt/Core/unittests/MrmTests.cpp
jefgen: I'll replace this with the string helper.

axelandrejs (approved):

PR #770, opened 2021-04-27
Clean up of Microsoft.ApplicationModel.Resources.PriGen.targets (round 4)

rohanp-msft:
Remove more unused MSBuild code.

Microsoft.ApplicationModel.Resources.PriGen.targets was created from Microsoft.AppxPackage.targets where a bunch of appx generation MSBuild script code was removed (Microsoft.AppxPackage.targets does more than just PRI generation, which is all that we want Microsoft.ApplicationModel.Resources.PriGen.targets to do, as the name suggests). At that point though, not all non-PRI gen related MSBuild script code was removed. I'll now try to remove it over several rounds. This is round 4.

Round 1: #664
Round 2: #696
Round 3: #704 

How verified
Used the MRTCore C# sample app and lib. Got rid of build outputs. Replaced the prigen targets file in the nuget package install with the one in this change. Built and ran the app.

PR #839, opened 2021-05-19
Clean up of Microsoft.ApplicationModel.Resources.PriGen.targets (round 5)

rohanp-msft:
Round 5 of removal of unused MSBuild code in the PRI gen targets file.

Microsoft.ApplicationModel.Resources.PriGen.targets was created from Microsoft.AppxPackage.targets where a bunch of appx generation MSBuild script code was removed (Microsoft.AppxPackage.targets does more than just PRI generation, which is all that we want Microsoft.ApplicationModel.Resources.PriGen.targets to do, as the name suggests). At that point though, not all non-PRI gen related MSBuild script code was removed. I'll now try to remove it over several rounds. This is round 5.

Round 1: #664
Round 2: #696
Round 3: #704
Round 4: #770 

**How verified**
Tested using C++ MRT Core sample app.


DrusTheAxe:
Is this obsolete?

rohanp-msft:
> Is this obsolete?

@DrusTheAxe This isn't needed, but is definitely good to do. I'll close this PR, but just keep my branch alive. I'll try to return to this and get it done for v1.1.

PR #843, opened 2021-05-19
Onboard Foundation to use Maestro 

kythant:
Start publishing asset information to Maestro
Trigger the subscription on the internal pipeline but not the CI pipeline
Added eng/common folder which will be shared across every repo that uses Maestro


build/AzurePipelinesTemplates/ProjectReunion-CreateNugetPackage-Job.yml
```diff
@@ -150,6 +151,13 @@ jobs:
         #Note: The project qualifier is always required when using a feed name. Also, do not use organization scoped feeds.
         publishVstsFeed: 'ProjectReunion/Project.Reunion.nuget.internal'
 
+  - ${{ if eq(parameters.publishToMaestro, 'true') }}:
+    - template: ..\..\eng\common\Maestro-PublishBuildToMaestro.yml
+      parameters:
+        IsGitHubRepo: true
+        AssetName: 'Microsoft.ProjectReunion.Foundation.TransportPackage'
+        AssetVersion: '$(version).${{ parameters.prereleaseVersionTag }}'
```
huichen123, line 159: >${{ parameters.prereleaseVersionTag }} [](http://example.com/codeflow?start=34&length=38)

do we get a "." at end if prereleaseVersionTag is missing?

kythant: The precondition is that there will always be a prereleaseVersionTag :)

kythant: Refactring this because '$(version).${{ parameters.prereleaseVersionTag }}' is repeated else where

huichen123: even for a stable release build?

eng/common/MaestroHelpers.ps1
```diff
@@ -0,0 +1,9 @@
+# Pram: CollectionUri must be in the form below
```
BenJKuhn, line 1: Param

BenJKuhn (approved):

PR #845, opened 2021-05-19
DynamicDependencies bug in AppxManifest'd winrt inprocserver handling

DrusTheAxe:
2 bugs in AppxManifest-based UndockedRegFreeWinRT support we unnoticed due to a gap in test coverage

* 2+ `<Extension Category='windows.activatableClass.inProcessServer'>` in a manifest
* 2+ inprocservers with the same Path

Code updated to handle these cases as expected.

The PRfwk test package (`test\DynamicDependency\data\Microsoft.ProjectReunion.Framework`) was updated to ensure coverage of these scenarios

Added new WinRT_RoGetActivationFactory_* test cases to verify the fix. Tests pass now as expected

### DIFF
--- a/dev/DynamicDependency/WinRTPackage.cpp
+++ b/dev/DynamicDependency/WinRTPackage.cpp
@@ -180,17 +180,27 @@ void MddCore::WinRTPackage::ParseAppxManifest_InProcessServer(
                     }
 
                     // Next!
-                    hr =xmlReader->MoveToNextAttribute();
+                    hr = xmlReader->MoveToNextAttribute();
                 }
                 THROW_HR_IF_MSG(APPX_E_INVALID_MANIFEST, activatableClassId.empty() || (threadingModel == MddCore::WinRT::ThreadingModel::Unknown), "Error 0x%X parsing <ActivatableClass> in %ls", hr, filename.c_str());
                 winrtInProcModule.AddInprocServer(activatableClassId, threadingModel);
+            }
+        }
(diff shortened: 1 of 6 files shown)


axelandrejs (approved):

dev/DynamicDependency/WinRTPackage.cpp
```diff
@@ -180,17 +180,27 @@ void MddCore::WinRTPackage::ParseAppxManifest_InProcessServer(
...
+                // Found </InProcessServer> so we're done
+                break;
+            }
+        }
     }
     THROW_HR_IF_MSG(APPX_E_INVALID_MANIFEST, winrtInProcModule.Path().empty() || winrtInProcModule.InprocServers().empty(), "Error 0x%X parsing <ActivatableClass> in %ls", APPX_E_INVALID_MANIFEST, filename.c_str());
 
     // Gotcha!
-    m_inprocModules.push_back(std::move(winrtInProcModule));
+    AddInprocModule(winrtInProcModule);
```
jonwis, line 203: You should probably still `move` here and use `&&` syntax on the method, since it might move-from `winrtInProcModule` leaves this local here in the moved-from state.

jonwis (approved):

dev/DynamicDependency/WinRTPackage.cpp
DrusTheAxe: Good idea. Looking to cook this quickly for others' testing so I'll add do that to a future PR.

PR #846, opened 2021-05-19
Add framework deployment for DDLM, update output text

dkbennett:
Updated help and output text and added deployment behavior option for main packages that need to be deployed like frameworks. DDLM package definitions updated to have framework-like deployment.

Updated tests and verified installer tests are successful.


installer/dev/install.cpp
```diff
@@ -78,7 +78,7 @@ namespace ProjectReunionInstaller {
 
         // It is assumed that all available architectures for non-framework packages are present,
         // so only the same-architecture or neutral will be matched for non-frameworks.
-        if (!packageProperties->isFramework)
+        if (!packageProperties->isFramework && !(deploymentBehavior == DeploymentBehavior::Framework))
```
DrusTheAxe, line 81: NIT: `(deploymentBehavior != DeploymentBehavior::Framework)`

installer/dev/main.cpp
```diff
@@ -31,6 +31,19 @@ int wmain(int argc, wchar_t *argv[])
         }
     }
 
-    return ProjectReunionInstaller::DeployPackages(quiet);
+    const HRESULT deployPackagesResult = ProjectReunionInstaller::DeployPackages(quiet);
```
DrusTheAxe, line 34: NIT: `const HRESULT deployPackagesResult{ ProjectReunionInstaller::DeployPackages(quiet) };`

installer/dev/main.cpp
```diff
@@ -31,6 +31,19 @@ int wmain(int argc, wchar_t *argv[])
...
+    if (!quiet)
+    {
+        if (SUCCEEDED(deployPackagesResult))
+        {
+            std::cout << "All packages were installed successfully." << std::endl;
+        }
+        else
+        {
+            std::wcerr << "One or more packages failed to install. Result: 0x" << std::hex << deployPackagesResult << std::endl;
```
DrusTheAxe, line 43: NIT: cout vs wcerr. Probably best to pick 1 and be consistent cout+cerr or (better) wcout+wcerr.

For a future cleanup PR

DrusTheAxe, line 43: Does `deployPackagesResult` include the package(s) that failed? If not that would make for a nifty future enhancement so the error could report '%d %s failed to install (...packages...)", count, pluralize(count, "Package", "Packages")) or some such

installer/dev/packages.h
```diff
@@ -11,10 +11,17 @@ using namespace Windows::System;
 
 namespace ProjectReunionInstaller {
 
+    enum class DeploymentBehavior
+    {
+        Default,        // Package will be deployed based on its properties alone.
+        Framework       // Package will be treated as a framework, even if it is not.
```
DrusTheAxe, line 17: NIT: `Framework,`

Last symbol in the enum can have a trailing comma (not an error, for 5-10 years now) and this way future additions make for simpler diffs (this line won't change because you don't need to add a comma)

installer/dev/packages.h
```diff
@@ -11,10 +11,17 @@ using namespace Windows::System;
...
+    };
+
     struct ResourcePackageInfo
     {
         std::wstring id;
         std::wstring resourceType;
+        DeploymentBehavior deploymentBehavior;
```
DrusTheAxe, line 24: Needs initialization

`DeploymentBehavior deploymentBehavior{};`

or

`auto deploymentBehavior{ DeploymentBehavior::Default };`

DrusTheAxe (changes requested):

DrusTheAxe:
LGTM modulo the comments I left. Minor or future things so I'm signing off to unblock completion when you're ready

DrusTheAxe (approved):

installer/dev/main.cpp
dkbennett: I want the DeployPackages to return a result object that contains detailed status so we can report those details, which packages failed, what was the failure for each package, etc.  For now since it fails the entire operation the first time a fatal error is encountered it returns that error. Each package is announced in output so it is fairly obvious which package failed currently, unless the failure occurred in obtaining the package information. For that the result object could at least include what the resource identifier is. But all of that was more churn than I wanted for this PR.

PR #850, opened 2021-05-21
cherry pick change 849

huichen123:
cherry-pick https://github.com/microsoft/ProjectReunion/pull/849.


huichen123:
i will abort this and create a new one as the history is not correct.

PR #851, opened 2021-05-21
cherry pick change (#849)

huichen123:
cherry-pick #849.


huichen123:
abandon. already in preview2.

PR #881, opened 2021-05-27
Enable MRT Core Unpackaged Tests to work in Visual Studio Test Explorer

jefgen:
This change enables the MRT Core Unpackaged Tests to work in the Visual Studio Test Explorer.

The Visual Studio Test Explorer by default looks for test adapters under the "`packages`" folder and will use those ahead of (ie: ignore) ones in the `BuildOutput` folder. This means that when `TE.exe` is invoked by the Visual Studio Test Explorer, it is run from the path:

```
ProjectReunion\dev\MRTCore\mrt\packages\Microsoft.Taef.10.58.210222006-develop\build\Binaries\x64
```

instead of from the `BuildOutput` folder. This causes the test to fail, as it looks in the current directory for the `app.manifest` and also for the `Microsoft.ApplicationModel.Resources.*` files.

This change adds a `.runsettings` [configuration file](https://docs.microsoft.com/en-us/visualstudio/test/configure-unit-tests-by-using-a-dot-runsettings-file?view=vs-2019), which tells the Visual Studio Test Explorer `VSTest.console.exe` wrapper process to invoke/use the `TE.exe` from the `BuildOutput` folder instead of the one from the "packages" folder.

Unfortunately, the ".runsettings" file can only deal with full absolute paths, and it has no support for variables, so I had to use (a perhaps overly clever) approach of using inline PowerShell to update the path as part of the PostBuild step.

With this change, all of the MRT Core tests now work in the VS Test Explorer. 🙂

image


dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/UnpackagedTests/UnpackagedTests.csproj
```diff
@@ -63,6 +63,9 @@
     <None Include="resources.pri.standalone">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
+    <None Include=".runsettings">
```
huichen123, line 66: can we give it a name? Starting with a dot makes it a generated file.

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/UnpackagedTests/UnpackagedTests.csproj
```diff
@@ -78,10 +81,15 @@
     </PropertyGroup>
     <Error Condition="!Exists('..\..\packages\Microsoft.Taef.10.58.210222006-develop\build\Microsoft.Taef.targets')" Text="$([System.String]::Format('$(ErrorText)', '..\..\packages\Microsoft.Taef.10.58.210222006-develop\build\Microsoft.Taef.targets'))" />
   </Target>
+  <!-- The powershell command below updates the TestAdaptersPaths path in the runsettings file, in order to allow the test to run in VS Test Explorer. -->
   <PropertyGroup>
-    <PostBuildEvent>copy /Y "$(OutputPath)..\Microsoft.ApplicationModel.Resources\Microsoft.Internal.FrameworkUdk.dll"  "$(OutputPath)Microsoft.Internal.FrameworkUdk.dll" 
+    <PostBuildEvent>copy /Y "$(OutputPath)..\Microsoft.ApplicationModel.Resources\Microsoft.Internal.FrameworkUdk.dll"  "$(OutputPath)Microsoft.Internal.FrameworkUdk.dll"
+    powershell -command "$file = gi $(OutputPath)\.runsettings; $x = [xml](gc $file); $x.RunSettings.RunConfiguration.TestAdaptersPaths = '$(OutputPath)'; $x.Save($file.Fullname)"
```
huichen123, line 87: "\\" not needed

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/UnpackagedTests/UnpackagedTests.csproj
```diff
@@ -78,10 +81,15 @@
...
     </PostBuildEvent>
   </PropertyGroup>
+  <PropertyGroup>
+    <RunSettingsFilePath>$(OutputPath)\.runsettings</RunSettingsFilePath>
```
huichen123, line 91: "\\" not needed

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/UnpackagedTests/UnpackagedTests.csproj
jefgen: Sure, how about "`UnpackagedTests.runsettings`".

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/UnpackagedTests/UnpackagedTests.csproj
jefgen: good point, thanks

huichen123 (approved):

PR #882, opened 2021-05-27
Adding Suspend/Resume notification code()

huzaifa-d:

### DIFF
--- a/dev/PowerNotifications/PowerNotifications.h
+++ b/dev/PowerNotifications/PowerNotifications.h
@@ -608,10 +608,10 @@ namespace winrt::Microsoft::ProjectReunion::factory_implementation
         //SystemSuspend Functions
         winrt::Microsoft::ProjectReunion::SystemSuspendStatus SystemSuspendStatus()
         {
-            // Review: Agree with this "only callable during SystemSuspendStatusChanged callback" enforcement?
+            // Review: Agree with this "only callable after subscription callback" enforcement?
             if (m_systemSuspendStatus == SystemSuspendStatus::Uninitialized)
             {
-                throw winrt::hresult_error(E_FAIL, L"There are no active subscriptions on SystemSuspendStatus");
+                throw winrt::hresult_error(E_FAIL, L"API only callable after a SystemSuspendStatusChanged callback");
             }
             return static_cast<winrt::Microsoft::ProjectReunion::SystemSuspendStatus>(m_systemSuspendStatus);
         }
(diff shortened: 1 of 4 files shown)


dev/PowerNotifications/PowerNotifications.h
```diff
@@ -592,6 +605,41 @@ namespace winrt::Microsoft::ProjectReunion::factory_implementation
             RaiseEvent(systemAwayModeStatusFunc);
         }
 
+        //SystemSuspend Functions
+        winrt::Microsoft::ProjectReunion::SystemSuspendStatus SystemSuspendStatus()
+        {
+            return static_cast<winrt::Microsoft::ProjectReunion::SystemSuspendStatus>(m_systemSuspendStatus);
```
brialmsft, line 611: What happens if somebody calls this before anyone has registered a listener and had it fire? Isn't this undefined?
Is the expectation that this is only legal to call from inside of a callback? If so, that's different then other propgets on the same class, which seems potentially confusing.

huzaifa-d: Added uninitialized enum value to handle that. Will be finalized by Andrew/Jon as discussed.

dev/PowerNotifications/PowerNotifications.h
```diff
@@ -608,6 +608,11 @@ namespace winrt::Microsoft::ProjectReunion::factory_implementation
         //SystemSuspend Functions
         winrt::Microsoft::ProjectReunion::SystemSuspendStatus SystemSuspendStatus()
         {
+Addi            // Review: Agree with this usage-only-after-subscription enforcement?
```
brialmsft, line 611: Unintended text "Addi"?

dev/PowerNotifications/PowerNotifications.h
```diff
@@ -608,6 +608,11 @@ namespace winrt::Microsoft::ProjectReunion::factory_implementation
...
+            if (m_systemSuspendStatus == SystemSuspendStatus::Uninitialized)
+            {
+                throw winrt::hresult_error(E_FAIL, L"There are no active subscriptions on SystemSuspendStatus");
```
brialmsft, line 614: I think "Only callable during SystemSuspendStatusChanged callback" is more accurate.

dev/PowerNotifications/PowerNotifications.h
huzaifa-d: Yes. Fixed.

dev/PowerNotifications/PowerNotifications.h
brialmsft: I was referring to the error message on line 614. Even if you have an active subscription, you're blocked from calling this method from outside of a callback, right? That's why I'm recommending that the message be clarified.

brialmsft (approved):

PR #888, opened 2021-05-30
Use the Universal C Runtime

DrusTheAxe:
Update Microsoft.ProjectReunion.dll, Microsoft.ProjectReunion.Bootstrap.dll, DynamicDependency DataStore and LifetimeManager to use the [Universal CRT](https://docs.microsoft.com/cpp/porting/upgrade-your-code-to-the-universal-crt?view=msvc-160) (like MRTCore's been doing) and VS' static CRT. This removes the dependency on the VS CRT redistributable at negible cost (see below).

Static analysis before/after using `LINK /DUMP /IMPORTS`, linker options `/MAP /VERBOSE` and `SizeBench` confirms the trivial size delta.

Disk size of x64 Binaries

|File | Size Before | Size After | Delta|
|---|---:|---:|---:|
|DynamicDependency.DataStore.exe                |  58,368 |  81,408 |  23,040 |
|DynamicDependency.DataStore.ProxyStub.dll      |  13,312 |  24,064 |  10,752 |
|DynamicDependencyLifetimeManager.exe           |  41,984 |  64,000 |  22,016 |
|DynamicDependencyLifetimeManager.ProxyStub.dll |  13,312 |  24,064 |  10,752 |
|Microsoft.ProjectReunion.dll                   | 516,096 | 554,496 |  38,400 |
|Microsoft.ProjectReunion.Bootstrap.dll         |  80,896 | 104,448 |  23,552 |
| Total                                          | 723,968 | 852,480 | 128,512 |

SizeBench delta for Microsoft.ProjectReunion.dll

| Name | Size on disk Diff | Size in memory Diff | Size in memory (including section padding) Diff |
|---|---:|---:|---:|
| .text | 23552 | 23637 | 24576 |
| .rdata | 11264 | 11432 | 12288 |
| .pdata | 1536 | 1644 | 0 |
| .data | 1024 | 2816 | 4096 |
| .reloc | 512 | 532 | 0 |
| _RDATA | 512 | 244 | 4096 |
| .detourc | 0 | 0 | 0 |
| .detourd | 0 | 0 | 0 |
| .rsrc | 0 | 0 | 0 |

SizeBench delta for Microsoft.ProjectReunion.Bootstrap.dll

| Name | Size on disk Diff | Size in memory Diff | Size in memory (including section padding) Diff |
|---|---:|---:|---:|
| .text | 15872 | 16176 | 16384 |
| .rdata | 5120 | 5272 | 4096 |
| .pdata | 1024 | 996 | 4096 |
| .data | 512 | 480 | 0 |
| .reloc | 512 | 256 | 0 |
| _RDATA | 512 | 252 | 4096 |
| .didat | 0 | 0 | 0 |
| .rsrc | 0 | 0 | 0 |

The `.vcxproj` changes are based the implementation in `MRM.vcxproj` and `BaseUnitTests.vcxproj`

1. Add UseUniCrt=true to the 'Globals' PropertyGroup
2. Added RuntimeLibrary=MultiThreadedDebug (Debug) or MultiThreaded (Release)
3. Add IgnoreSpecificDefaultLibraries=libucrtd.lib (Debug) or libucrt.lib (Release)
4. Add AdditionalOptions /defaultlib:ucrtd.lib (Debug) or /defaultlib:ucrt.lib (Release)

Snippets below:

#1 UseUnicrt
```xml
  <PropertyGroup Label="Globals">
    <UseUnicrt>true</UseUnicrt>
```

#2-4 ClCompile and Link options

```xml
  <ItemDefinitionGroup Condition="'$(Configuration)'=='Debug'">
      <!-- We use MultiThreadedDebug, rather than MultiThreadedDebugDLL, to avoid DLL dependencies on VCRUNTIME140d.dll and MSVCP140d.dll. -->
      <RuntimeLibrary>MultiThreadedDebug</RuntimeLibrary>
    </ClCompile>
    <Link>
      <!-- Link statically against the runtime and STL, but link dynamically against the CRT by ignoring the static CRT
           lib and instead linking against the Universal CRT DLL import library. This "hybrid" linking mechanism is
           supported according to the CRT maintainer. Dynamic linking against the CRT makes the binaries a bit smaller
           than they would otherwise be if the CRT, runtime, and STL were all statically linked in. -->
      <IgnoreSpecificDefaultLibraries>%(IgnoreSpecificDefaultLibraries);libucrtd.lib</IgnoreSpecificDefaultLibraries>
      <AdditionalOptions>%(AdditionalOptions) /defaultlib:ucrtd.lib</AdditionalOptions>
    </Link>
  </ItemDefinitionGroup>
  <ItemDefinitionGroup Condition="'$(Configuration)'=='Release'">
    <ClCompile>
      <!-- We use MultiThreaded, rather than MultiThreadedDLL, to avoid DLL dependencies on VCRUNTIME140.dll and MSVCP140.dll. -->
      <RuntimeLibrary>MultiThreaded</RuntimeLibrary>
    </ClCompile>
    <Link>
      <!-- Link statically against the runtime and STL, but link dynamically against the CRT by ignoring the static CRT
           lib and instead linking against the Universal CRT DLL import library. This "hybrid" linking mechanism is
           supported according to the CRT maintainer. Dynamic linking against the CRT makes the binaries a bit smaller
           than they would otherwise be if the CRT, runtime, and STL were all statically linked in. -->
      <IgnoreSpecificDefaultLibraries>%(IgnoreSpecificDefaultLibraries);libucrt.lib</IgnoreSpecificDefaultLibraries>
      <AdditionalOptions>%(AdditionalOptions) /defaultlib:ucrt.lib</AdditionalOptions>
    </Link>
  </ItemDefinitionGroup>
```

This really should go into a common `UniversalCrt.props` like other common properties (no reason all projects shouldn't use this) along with stripping unwanted options and/or verifying no inconsistencies but those build infrastructure refinements are left as an exercise for a future PR. First step is to use UCRT for product.

### DIFF
--- a/dev/DynamicDependencyLifetimeManager/DynamicDependencyLifetimeManager.ProxyStub/DynamicDependencyLifetimeManager.ProxyStub.vcxproj
+++ b/dev/DynamicDependencyLifetimeManager/DynamicDependencyLifetimeManager.ProxyStub/DynamicDependencyLifetimeManager.ProxyStub.vcxproj
@@ -42,6 +42,7 @@
     <RootNamespace>DynamicDependencyLifetimeManager.ProxyStub</RootNamespace>
     <WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>
     <ProjectName>DynamicDependencyLifetimeManager.ProxyStub</ProjectName>
+    <UseUnicrt>true</UseUnicrt>
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Configuration">
@@ -159,6 +160,36 @@
     <LinkIncremental>false</LinkIncremental>
     <CustomBuildAfterTargets>PostBuildEvent</CustomBuildAfterTargets>
   </PropertyGroup>
+
+  <ItemDefinitionGroup Condition="'$(Configuration)'=='Debug'">
+    <ClCompile>
+      <!-- We use MultiThreadedDebug, rather than MultiThreadedDebugDLL, to avoid DLL dependencies on VCRUNTIME140d.dll and MSVCP140d.dll. -->
+      <RuntimeLibrary>MultiThreadedDebug</RuntimeLibrary>
+    </ClCompile>
+    <Link>
+      <!-- Link statically against the runtime and STL, but link dynamically against the CRT by ignoring the static CRT
+           lib and instead linking against the Universal CRT DLL import library. This "hybrid" linking mechanism is
+           supported according to the CRT maintainer. Dynamic linking against the CRT makes the binaries a bit smaller
+           than they would otherwise be if the CRT, runtime, and STL were all statically linked in. -->
+      <IgnoreSpecificDefaultLibraries>%(IgnoreSpecificDefaultLibraries);libucrtd.lib</IgnoreSpecificDefaultLibraries>
+      <AdditionalOptions>%(AdditionalOptions) /defaultlib:ucrtd.lib</AdditionalOptions>
+    </Link>
+  </ItemDefinitionGroup>
+  <ItemDefinitionGroup Condition="'$(Configuration)'=='Release'">
+    <ClCompile>
+      <!-- We use MultiThreaded, rather than MultiThreadedDLL, to avoid DLL dependencies on VCRUNTIME140.dll and MSVCP140.dll. -->
+      <RuntimeLibrary>MultiThreaded</RuntimeLibrary>
+    </ClCompile>
+    <Link>
+      <!-- Link statically against the runtime and STL, but link dynamically against the CRT by ignoring the static CRT
+           lib and instead linking against the Universal CRT DLL import library. This "hybrid" linking mechanism is
+           supported according to the CRT maintainer. Dynamic linking against the CRT makes the binaries a bit smaller
+           than they would otherwise be if the CRT, runtime, and STL were all statically linked in. -->
+      <IgnoreSpecificDefaultLibraries>%(IgnoreSpecificDefaultLibraries);libucrt.lib</IgnoreSpecificDefaultLibraries>
+      <AdditionalOptions>%(AdditionalOptions) /defaultlib:ucrt.lib</AdditionalOptions>
+    </Link>
+  </ItemDefinitionGroup>
+
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
     <ClCompile>
       <WarningLevel>Level4</WarningLevel>
--- a/dev/ProjectReunion_DLL/ProjectReunion_DLL.vcxproj
+++ b/dev/ProjectReunion_DLL/ProjectReunion_DLL.vcxproj
@@ -42,6 +42,7 @@
     <Keyword>Win32Proj</Keyword>
     <RootNamespace>Microsoft.ProjectReunion</RootNamespace>
     <WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>
+    <UseUnicrt>true</UseUnicrt>
     <CppWinRTModernIDL>true</CppWinRTModernIDL>
     <CppWinRTFastAbi>true</CppWinRTFastAbi>
     <CppWinRTOptimized>true</CppWinRTOptimized>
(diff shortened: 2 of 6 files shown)


riverar:
The cost is security, is this really what we want to do?

riverar:
To other reviewers: If you're like me, you keep falling for this trick. No security issue here, see https://github.com/microsoft/ProjectReunion/issues/573#issuecomment-797277512 for more context. 👍

dev/DynamicDependencyLifetimeManager/DynamicDependencyLifetimeManager.ProxyStub/DynamicDependencyLifetimeManager.ProxyStub.vcxproj
```diff
@@ -159,6 +160,36 @@
     <LinkIncremental>false</LinkIncremental>
     <CustomBuildAfterTargets>PostBuildEvent</CustomBuildAfterTargets>
   </PropertyGroup>
+
+  <ItemDefinitionGroup Condition="'$(Configuration)'=='Debug'">
+    <ClCompile>
+      <!-- We use MultiThreadedDebug, rather than MultiThreadedDebugDLL, to avoid DLL dependencies on VCRUNTIME140d.dll and MSVCP140d.dll. -->
```
jonwis, line 166: Not sure this comment is required (applies everywhere)

dev/DynamicDependencyLifetimeManager/DynamicDependencyLifetimeManager.ProxyStub/DynamicDependencyLifetimeManager.ProxyStub.vcxproj
```diff
@@ -159,6 +160,36 @@
...
+      <RuntimeLibrary>MultiThreadedDebug</RuntimeLibrary>
+    </ClCompile>
+    <Link>
+      <!-- Link statically against the runtime and STL, but link dynamically against the CRT by ignoring the static CRT
```
jonwis, line 170: Same comment here as well. This is an interesting point that could be put up in a documentation .md file under docs/ maybe.

jonwis (approved):

dev/ProjectReunion_DLL/ProjectReunion_DLL.vcxproj
```diff
@@ -167,6 +168,36 @@
     <LinkIncremental>false</LinkIncremental>
     <TargetName>Microsoft.ProjectReunion</TargetName>
   </PropertyGroup>
+
+  <ItemDefinitionGroup Condition="'$(Configuration)'=='Debug'">
```
Scottj1s, line 172: consider gathering all this logic into a root directory.build.props file to ensure/enforce that all current and future reunion projects use hybrid CRT linkage.

Scottj1s:
I am in favor of this strategy, and love the analysis above which everyone has been asking for.  As you say, the tax is negligible and a fair trade-off for eliminating CRT redist woes for the end developer.  That said, the competition for this is the VCLibs.UWPDesktop converged framework package and dynamic dependency deployment (yes?).  Should pros/cons of that option be considered and included?

DrusTheAxe:
> VCLibs.UWPDesktop converged framework package and dynamic dependency deployment (yes?). Should pros/cons of that option be considered and included?

No. They're not really equivalent if you consider process startup and loader dependencies. 

VCLibs framework package is cool for packaged processes as they're **born** with the manifested dependency in their loader search order. Not so for unpackaged apps. For non-C/C++ modules it's no big deal since they have no imports to the CRT so e.g. C# foo.exe can run, use DynDep to load VCLibs.UWPDesktop than use CRT APIs. For C/C++ (and Rust? Others?) modules they'd with any IAT entries for the CRT would have to delayload the CRT. Possible but irksome so the added friction is unappealing.

The small tax for this hybrid approach is a more compelling answer, at least for Project Reunion.

dev/ProjectReunion_DLL/ProjectReunion_DLL.vcxproj
DrusTheAxe: I'll do that in a future PR. I didn't modify the test projects, or even verify if they would just pick this up or need updates.

Will refactor in a follow up PR.

dev/DynamicDependencyLifetimeManager/DynamicDependencyLifetimeManager.ProxyStub/DynamicDependencyLifetimeManager.ProxyStub.vcxproj
DrusTheAxe: Yep. I'll do that when I refactor these out to a common .props file in a future PR (as mentioned below)

dev/DynamicDependencyLifetimeManager/DynamicDependencyLifetimeManager.ProxyStub/DynamicDependencyLifetimeManager.ProxyStub.vcxproj
DrusTheAxe: See other comment

jaykrell:
I'm copying this, for a few places we were thinking of using static C runtime.
But I'm not satisfied.

I want vcredist to be easier to use and then always dynamically link.
The tax shown is too high, as well as the risk of shared state being duplicated.

The main but not only places in question are MSI custom actions.
It should be really easy to have a multi-dll custom action, and use the loose files next to the custom action in a temporary directory (i.e.  xcopy install). And run vcredist. And reboot at the end, maybe. But really, the reboot thing is dumb. You can move files out of the way that are in use. So do a quick rename, copy back, copy over. With risk of breaking concurrent process launch.

Alternatively, run vcredist early, then reboot, then continue.

PR #890, opened 2021-05-31
Replace PNG assets with lossless optimized variants

riverar:
Replaces loose PNG assets with optimized variants sans embedded metadata. This process is lossless/does not affect image quality in any way.

Saved bytes: 291.04 KiB


riverar:
Went stale and assets changed.

Issue #893, opened 2021-05-31
AnyCPU support


dotMorten:
**Describe the bug**
I'm trying to use MRTCore in a WPF application. I reference ProjectReunion package, and compile and I'm presented with this error:

> 1>C:\Users\mn\.nuget\packages\microsoft.projectreunion.winui\0.8.0-preview\buildTransitive\Microsoft.WinUI.targets(15,5): error : This version of WinUI does not support AnyCPU. Either set Platform or PlatformTarget to one of the following: x86, x64, or arm64.

I'm in no way interested in using WinUI here, but the current nuget dependency forces me to get WinUI included, including all the types etc. IMHO this is backwards, and instead `Microsoft.ProjectReunion.WinUI` you be referencing `Microsoft.ProjectReunion`. The error above is extremely confusing as I am not interested in or explicitly referencing the WinUI _extension_.

Sure I can avoid using AnyCPU, but that's not so much the point (and I just followed the sample app config which has AnyCPU as well: https://github.com/microsoft/Project-Reunion-Samples/blob/main/MrtCore/winforms_unpackaged_app/winforms_unpackaged_app.csproj). Also why can't you do AnyCPU? Other .NET3+ apps can handle deploying multiple architecture runtimes in an AnyCPU app.

**Steps to reproduce the bug**
Create a WPF Application and add reunion:

```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>WinExe</OutputType>

    <TargetFramework>net5.0-windows10.0.19041.0</TargetFramework>
    <UseWPF>true</UseWPF>
    <Platforms>AnyCPU;x86;x64</Platforms>
    <RuntimeIdentifiers>win10-x86;win10-x64</RuntimeIdentifiers>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Microsoft.ProjectReunion" Version="0.8.0-preview" />
  </ItemGroup>
</Project>

```
Compile for AnyCPU

I get the app packages the same runtime dependencies, but it creates confusing unrelated errors, and exposes a set of APIs that I never intended to reference. I'm only interested in MRTCore which is packaged with the "base" reunion package, yet I'm getting DWrite, InteractiveExperiences and WinUI throw in as well.

**Expected behavior**
WinUI, DWrite and InteractiveExperiences package is not included. Intellisense doesn't autocomplete those APIs either.

**Version Info**

NuGet package version: 0.8.0-preview

[...]

PR #897, opened 2021-06-01
Add UseWinUI property

huichen123:
We will use this property to identify WinUI project.


dotMorten:
I'm curious what this property does, since there are no ms build changes?
Would have been cleaner that the WinUI package set hit for you if it was needed (but that would require nuget dependency hierarchy to be more logically ordered, so you don't get the WinUI package transitively like it is now - see #893)

huichen123:
> I'm curious what this property does, since there are no ms build changes?
> Would have been cleaner that the WinUI package set this for you if it was needed (but that would require nuget dependency hierarchy to be more logically ordered, so you don't get the WinUI package transitively like it is now - see #893)

The property will be used to exclude WinUI stuff from those not interesting in WinUI. The build change is targeted for 1.0. Getting this property in first so that 0.8 can be upgraded to 1.0 (i am not saying there will be no other breaking change though).

alwu-msft (approved):
:shipit:

PR #908, opened 2021-06-03
Replace custom path handling in MRT

huichen123:
also removed custom build configurations in mrt build task.

### DIFF
--- a/dev/MRTCore/mrt/Core/src/MRM.cpp
+++ b/dev/MRTCore/mrt/Core/src/MRM.cpp
@@ -1,8 +1,10 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 #include <Windows.h>
 #include <Pathcch.h>
+#include "wil/win32_helpers.h"
+#include "wil/filesystem.h"
 #include "mrm/BaseInternal.h"
 #include "mrm/common/platform.h"
 #include "mrm/readers/MrmReaders.h"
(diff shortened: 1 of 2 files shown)


dev/MRTCore/mrt/Core/src/MRM.cpp
```diff
@@ -874,49 +876,20 @@ STDAPI MrmGetFilePathFromName(_In_opt_ PCWSTR filename, _Outptr_ PWSTR* filePath
 {
     *filePath = nullptr;
 
-    wchar_t path[MAX_PATH];
-    DWORD size = ARRAYSIZE(path);
-    DWORD length = GetModuleFileName(nullptr, path, size);
-    DWORD lastError = GetLastError();
-    std::unique_ptr<wchar_t[]> pathAllocated;
+    wil::unique_cotaskmem_string path;
+    RETURN_IF_FAILED((wil::GetModuleFileNameW<wil::unique_cotaskmem_string, 256>(nullptr, path)));
```
jonwis, line 880: You can leave off the `, 256` parameter and then remove the additional `( )` on this.

dev/MRTCore/mrt/Core/src/MRM.cpp
```diff
@@ -874,49 +876,20 @@ STDAPI MrmGetFilePathFromName(_In_opt_ PCWSTR filename, _Outptr_ PWSTR* filePath
...
-    while (*pointerToPath)
-    {
-        if (*pointerToPath == L'\\')
-        {
-            secondToLastSlash = lastSlash;
-            lastSlash = pointerToPath;
-        }
-        pointerToPath++;
-    }
+    PCWSTR name = wil::find_last_path_segment(path.get());
```
jonwis, line 882: Should this use the `std::filesystem` APIs instead?

jonwis (approved):

axelandrejs (approved):

dev/MRTCore/mrt/Core/src/MRM.cpp
huichen123: 256 is what i want to override. the default is 128. Many times, the path created by the reunion template is fairly long. Thus i want to bump up the default.

dev/MRTCore/mrt/Core/src/MRM.cpp
huichen123: The file contains no-throwing code only. with std, i am going to worry about whether it throws, and std::filesystem API does throw.

dev/MRTCore/mrt/Core/src/MRM.cpp
rohanp-msft, line 880: Isn't it better to use MAX_PATH instead of 256 here? MAX_PATH may be greater than 256 after a user modification, no?

dev/MRTCore/mrt/Core/src/MRM.cpp
```diff
@@ -874,49 +876,20 @@ STDAPI MrmGetFilePathFromName(_In_opt_ PCWSTR filename, _Outptr_ PWSTR* filePath
...
+    wchar_t moduleFilename[MAX_PATH];
```
rohanp-msft, line 883: 883-884 can go into the "// Prep for the second iteration" if block where if name is empty/null, we just break. We can avoid the unnecessary setting of module name to path and then a second iteration.

rohanp-msft (approved):

dev/MRTCore/mrt/Core/src/MRM.cpp
huichen123: 256 is just twice of the default (128). Some people may just don't like MAX_PATH. I doubt anyone is going to modify MAX_PATH. If it does, that's just one more reason to not use MAX_PATH.

dev/MRTCore/mrt/Core/src/MRM.cpp
huichen123: at that time, path is already modified, and lost the name.

dev/MRTCore/mrt/Core/src/MRM.cpp
riverar: You're trading codebase readability for a perceived micro-optimization. Other folks coming across this code are going to scratch their heads as to why the initial buffer size was not used here. Something to consider.

jonwis: It's better to use _no_ number here. WIL always produces a heap-allocated string from this method, and the buffer parameter is just a seed to know where to start growing the buffer.  Just don't specify, let WIL do its thing, then there's no Mysterious Constant hanging out here.  (Mysterious Constants in library implementations are fine.)

PR #910, opened 2021-06-04
Remove Config=Debug_test and Platform=ARM

DrusTheAxe:
Remove 'dead code' from Solution and Project files

* Config=Debug_test -- unused
* Platform=ARM -- ARM32 isn't supported per [Final decision: No ARM32 support (only ARM64) #406](https://github.com/microsoft/ProjectReunion/discussions/406)


huichen123:
there are also Debug_test in Directory.Build.props.

huichen123 (approved):
:shipit:

PR #925, opened 2021-06-10
set class library platforms

huichen123:
Without setting the platforms, the project is created default targeting to AnyCPU, and then you get a build error because WinUI doesn't support AnyCPU.


alwu-msft:
@huichen123 I thought we agreed in shiproom earlier this week to simply document this in the release notes? See: https://microsoft.visualstudio.com/OS/_workitems/edit/33652722.

huichen123:
this was late for 0.8-rc. For 0.8 GA, the WinUI will allow AnyCPU.

PR #928, opened 2021-06-11
0.8: Remove experimental bits for 0.8 release

DrusTheAxe:


build/NuSpecs/AppxManifest.xml
```diff
@@ -3,7 +3,7 @@
   <Identity Name="Microsoft.ProjectReunion.0.5" ProcessorArchitecture="x64" Publisher="CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US" Version="0.52102.3000.0" />
```
huichen123, line 3: >5 [](http://example.com/codeflow?start=45&length=1)

if you change below, it's probably better to keep this consistent.

huichen123:
>   </PropertyGroup>

not needed

---
Refers to: build/NuSpecs/ProjectReunion-Nuget-Native.C.props:7 in 809b65b. [](commit_id = 809b65b982132456421f924ae7248b17a915d2df, deletion_comment = False)

build/NuSpecs/ProjectReunion-Nuget-Native.targets
```diff
@@ -9,18 +9,4 @@
     <Native-Platform Condition="'$(Platform)' != 'Win32'">$(Platform)</Native-Platform>
   </PropertyGroup>
```
huichen123, line 10: are these needed?

build/NuSpecs/ProjectReunion-Nuget-Native.WinRt.props
```diff
@@ -6,20 +6,4 @@
     <_ProjectReunionFoundationPlatform Condition="'$(Platform)' != 'Win32'">$(Platform)</_ProjectReunionFoundationPlatform>
   </PropertyGroup>
```
huichen123, line 7: not needed

build/CopyFilesToStagingDir.ps1
```diff
@@ -79,27 +69,11 @@ if($PublishAppxFiles)
...
-# Libraries (*.lib)
-PublishFile $FullBuildOutput\ProjectReunion_BootstrapDLL\Microsoft.ProjectReunion.Bootstrap.lib $NugetDir\lib\win10-$Platform
-PublishFile $FullBuildOutput\ProjectReunion_DLL\Microsoft.ProjectReunion.lib $NugetDir\lib\win10-$Platform
-#
-# MSIX Framework package - DLLs
-PublishFile $FullBuildOutput\ProjectReunion_DLL\Microsoft.ProjectReunion.dll $NugetDir\runtimes\win10-$Platform\native
-PublishFile $FullBuildOutput\ProjectReunion_DLL\Microsoft.ProjectReunion.pdb $NugetDir\runtimes\win10-$Platform\native
-#
 # MSIX Main package
 PublishFile $FullBuildOutput\DynamicDependency.DataStore\DynamicDependency.DataStore.exe $NugetDir\runtimes\win10-$Platform\native
```
huichen123, line 78: >DynamicDependency [](http://example.com/codeflow?start=57&length=17)

so bootstrap is experimental, but dd is not?

huichen123:
if this is for 0.8, should we do it in 0.8 branch instead?

axelandrejs:
> if this is for 0.8, should we do it in 0.8 branch instead?

Normally we go main->release, but here I agree that this makes sense to do directly in the release branch. We want to keep the experimental APIs in main.

build/CopyFilesToStagingDir.ps1
DrusTheAxe: DynDep (which includes bootstrap) is experimental, but for the purpose of expediting the release the DataStore (PRmain) and DDLM (PRddlm) are untouched. The build will create but then we'll ignore those artifacts

Unless @BenJKuhn tells me it's easier to remove them here AND update the aggregator pipeline to not fail on the absence than to manually remove them from the final result

build/NuSpecs/ProjectReunion-Nuget-Native.WinRt.props
DrusTheAxe: But not harmful either. I was focused on deleting what's experimental but leaving anything inert as-is

@BenJKuhn which way do you want to go?

DrusTheAxe:
> > if this is for 0.8, should we do it in 0.8 branch instead?
> 
> Normally we go main->release, but here I agree that this makes sense to do directly in the release branch. We want to keep the experimental APIs in main.

Doh! That's the intent. Thanks for spotting that. Let me see where to alter the PR...

build/NuSpecs/ProjectReunion-Nuget-Native.targets
DrusTheAxe: See next comment?

DrusTheAxe:
> Doh! That's the intent. Thanks for spotting that. Let me see where to alter the PR...

I don't see how I can retarget the PR to the release branch instead of main. @BenJKuhn  @axelandrejs @huichen123 @kythant any idea short of abandoning this and making a new PR?

Directory.Build.props
```diff
@@ -6,11 +6,8 @@
 <Project InitialTargets="DirectoryBuildPropsInfo" ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup Condition="$(BuildingWithBuildExe) != 'true'">
     <IsTDPConfiguration>false</IsTDPConfiguration>
-    <IsTDPConfiguration Condition="'$(Configuration)' == 'Debug_test'">true</IsTDPConfiguration>
```
huichen123, line 9: ><IsTDPConfiguration Condition="'$(Configuration)' == 'Debug_test'">true</IsTDPConfiguration> [](http://example.com/codeflow?start=4&length=92)

now we have unrelated changes

DrusTheAxe: rebased to release/reunion-0.8

huichen123: maybe better just start fresh

DrusTheAxe: I'd rather not. It takes ~40 minutes to turn a build, plus however long we're queued. This change has no impact on the resulting Transport Package, nor even the build pipeline or inner-loop as we don't use Debug_test configs (hence the reason they've been stripped from main)

huichen123: i doubt we want to include these changes to release branch.

DrusTheAxe:
Bad rebase! Bad! No soup for you!

Replacing with a new PR to avoid pulling in unwanted changes from main

PR #929, opened 2021-06-12
Amending PowerMode Functions

huzaifa-d:
- Amending PowerMode Functions to use 
[PowerRegisterForEffectivePowerModeNotifications()](https://docs.microsoft.com/en-us/windows/win32/api/powersetting/nf-powersetting-powerregisterforeffectivepowermodenotifications)
- Removed SystemAwayStatus code

### DIFF
--- a/dev/PowerNotifications/PowerNotifications.cpp
+++ b/dev/PowerNotifications/PowerNotifications.cpp
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 #include <pch.h>
+#include <future>
 #include <PowerNotifications.h>
 #include <PowerNotificationsPal.h>
 #include <PowerManager.g.cpp>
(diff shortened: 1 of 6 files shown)


dev/PowerNotifications/PowerNotifications.cpp
```diff
@@ -197,32 +198,64 @@ namespace winrt::Microsoft::ProjectReunion::implementation
...
+        ULONG version = IsEffectivePowerModeV2Supported() ? EFFECTIVE_POWER_MODE_V2 : EFFECTIVE_POWER_MODE_V1;
+        promiseObj = std::promise<EFFECTIVE_POWER_MODE>();
+        futureObj = promiseObj.get_future();
+
+        check_hresult(PowerRegisterForEffectivePowerModeNotifications(
+            version, [](EFFECTIVE_POWER_MODE mode, PVOID)
+            {
+                promiseObj.set_value(mode);
+            }, NULL, &handle));
+        auto& mode = make_self<factory_implementation::PowerManager>()->m_cachedPowerMode;
```
jonwis, line 254: This is confusing compared to:

```c++
make_self<factory_implementation::PowerManager>()->m_cachedPowerMode = futureObj.get();
```

dev/PowerNotifications/PowerNotifications.cpp
```diff
@@ -197,32 +198,64 @@ namespace winrt::Microsoft::ProjectReunion::implementation
...
     }
 
-    void PowerSchemePersonality_Update()
+    void EffectivePowerMode_Update()
     {
-        check_hresult(GetPowerSchemePersonality(
-            &make_self<factory_implementation::PowerManager>()->m_cachedPowerSchemePersonality));
+        // Effectively the Get() for PowerMode
+        // Needs to get a temporary subscription to get most recent value
+        static std::promise<EFFECTIVE_POWER_MODE> promiseObj;
```
jonwis, line 242: This mechanism is confusing; what's this method trying to do? Having static promises & futures doesn't seem right.

brialmsft, line 242: Forcing a notification roundtrip every time doesn't seem ideal. Assuming that every time somebody gets a PowerModeChanged event, they immediately check what the new mode is, that means we need two notifications.

In theory, can't we say "somebody has a listener registered" plus "that listener has ever fired" equals "m_cachedPowerMode is already up-to-date"? Then we'd only need to do waiting here if a) no listener is registered, or b) somebody registered a listener and then called Get without waiting for that listener to fire.

dev/PowerNotifications/PowerNotifications.cpp
huzaifa-d: Changed

dev/PowerNotifications/PowerNotifications.cpp
huzaifa-d: Added a comment clarifying.

dev/PowerNotifications/PowerNotifications.cpp
```diff
@@ -197,32 +198,64 @@ namespace winrt::Microsoft::ProjectReunion::implementation
         check_hresult(UnregisterSystemIdleStatusChangedListener(
             make_self<factory_implementation::PowerManager>()->m_systemIdleStatusHandle));
     }
-   
-    // PowerSchemePersonality Functions
-    EventType& PowerSchemePersonality_Event()
+
+    bool IsEffectivePowerModeV2Supported()
```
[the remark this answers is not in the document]
huzaifa-d: I went ahead with that because checking if V2 is supported is what we were doing. But I guess since we don't know if new versions are going to be added later, we can incorporate this. Changed.

dev/PowerNotifications/PowerNotifications.cpp
huzaifa-d: We already have that. A subscription will fire an initial callback with the latest value. And if a get is called then, that latest value is returned.

brialmsft: Right, I forgot that this only gets called if there isn't an active subscription.

brialmsft (approved):

dev/PowerNotifications/PowerNotifications.h
```diff
@@ -649,14 +604,16 @@ namespace winrt::Microsoft::ProjectReunion::factory_implementation
         }
 
     };
-
 };
 
 namespace winrt::Microsoft::ProjectReunion::implementation
 {
+    // EffectivePowerMode variables
+    static ULONG g_powerModeVersion = NULL;
```
jonwis, line 612: What locking mode is required/desired here? Should both of these be `std::atomic<T>` so they can be safely updated/read? Or should there be a `std::mutex` that can be locked while mutating/reading them?

jonwis, line 612: Use `0` instead of `NULL` here, this is a ULONG not a pointer.

dev/PowerNotifications/PowerNotifications.cpp
```diff
@@ -197,32 +198,65 @@ namespace winrt::Microsoft::ProjectReunion::implementation
...
     {
-        check_hresult(GetPowerSchemePersonality(
-            &make_self<factory_implementation::PowerManager>()->m_cachedPowerSchemePersonality));
+        // Effectively the Get() for PowerMode
+        // Needs to get a temporary subscription to get most recent value
+        // Also, we need a static so that it can be accessed in the lambda below
+        // since references or [&] doesn't work
+        static std::promise<EFFECTIVE_POWER_MODE> promiseObj;
+        static std::future<EFFECTIVE_POWER_MODE> futureObj;
+        PVOID handle;
```
jonwis, line 246: Please add `using unique_power_notification = wil::unique_any<PVOID, decltype(&PowerUnregisterFromEffectivePowerModeNotifications), PowerUnregisterFromEffectivePowerModeNotifications>;` and then use it to manage the lifecycle of these things everywhere.

dev/PowerNotifications/PowerNotifications.cpp
```diff
@@ -197,32 +198,65 @@ namespace winrt::Microsoft::ProjectReunion::implementation
...
+        ULONG version = GetEffectivePowerModeVersion();
+        promiseObj = std::promise<EFFECTIVE_POWER_MODE>();
+        futureObj = promiseObj.get_future();
+
+        check_hresult(PowerRegisterForEffectivePowerModeNotifications(
```
jonwis, line 251: Don't smuggle out the context in a promise, define a little callback structure with a slim event, like this:

```c++
struct notify_callback {
    EFFECTIVE_POWER_MODE mode;
    wil::slim_event done;
} context;
auto handler = [](EFFECTIVE_POWER_MODE mode, PVOID rawContext) {
    auto context = reinterpret_cast<notify_callback*>(rawContext);
    context->mode = mode;
    context->done.SetEvent();
};
check_hresult(PowerRegisterForEffectivePowerModeNotifications(version, handler, &context, &handle));
context.done.wait();
check_hresult(PowerUnregisterFromEffectivePowerModeNotifications(handle));
```

jonwis (changes requested):

dev/PowerNotifications/PowerNotifications.h
huzaifa-d: I don't think any locking is desired; but have still changed it to atomic. There is an unlikely scenario that two threads will check for the version, which will result in checking the version again instead of using this cached value. Is there any other scenario that I might have missed?

huzaifa-d: Of course I would use a `nullptr` instead if it was a pointer. But I prefer using NULL. Is there a guidance against it?

dev/PowerNotifications/PowerNotifications.cpp
huzaifa-d: Great idea; changed. Just out of curiosity though, what's wrong with futures/promises?

jonwis (approved):

arvindn5 (approved):

dev/PowerNotifications/PowerNotifications.h
jonwis: Yes, because `NULL` is for pointers.  If you want an "unknown yet" model use `std::optional<ULONG>` instead. Don't use `NULL` here. If you mean zero, say literal zero.

PR #932, opened 2021-06-14
Strip experimental bits and references from 0.8 release

DrusTheAxe:


axelandrejs (approved):

build/NuSpecs/build-nupkg.ps1
```diff
@@ -108,11 +108,8 @@ function ConfigureNugetCommandLine {
 
 ConfigureNugetCommandLine("Microsoft.ProjectReunion")
 
-Copy-IntoNewDirectory -IfExists $BuildOutput\$BuildFlavor\x86\Microsoft.ProjectReunion\Microsoft.ProjectReunion.dll "$runtimesDir\win10-x86\native"
 Copy-IntoNewDirectory -IfExists $BuildOutput\$BuildFlavor\x86\Microsoft.ProjectReunion\Microsoft.ProjectReunion.pri "$runtimesDir\win10-x86\native"
```
huichen123, line 111: >Microsoft.ProjectReunion.pri [](http://example.com/codeflow?start=87&length=28)

what is this?

build/CopyFilesToStagingDir.ps1
```diff
@@ -35,93 +35,9 @@ function PublishFile {
 
 PublishFile $OverrideDir\DynamicDependency-Override.json $FullPublishDir\
```
huichen123, line 36: this?

build/NuSpecs/build-nupkg.ps1
DrusTheAxe: Got me. It was there. PRI file for PRfwk? Do we need it? If we remove it will PRfwk contain a resources.pri?

build/CopyFilesToStagingDir.ps1
DrusTheAxe: I don't think so, at least, not directly. It's not failing in "Copy files to staging dir" but in "Publish artifact: Full Nuget (Reunion DLLs)" (2 steps after the copy). I suspect dropped all the file copies to a dir so we don't have the dir now

Yeah. That sounds like it

```
steps:
  - task: powershell@2
    displayName: 'Copy files to staging dir'
    inputs:
      targetType: filePath
      filePath: build\CopyFilesToStagingDir.ps1
      arguments: ...-NugetDir '$(Build.ArtifactStagingDirectory)\FullNuget'...

Maybe we can just mkdir and make him happy. Let's see

build/NuSpecs/build-nupkg.ps1
huichen123: The only PRI in Reunion I know of comes from WinUI.

DrusTheAxe: How peculiar. But it's optional (-IfExists) so no harm if it's missing for now

If it's unnecessary we should drop it in the main branch. Will follow up

PR #936, opened 2021-06-17
Update VSIX to work with Dev17

DefaultRyan:
This change updates the VSIX to target the next version of VS. Fortunately, this VSIX contains only templates and no code, so it can target both Dev16 and Dev17 fairly easily.

The prerequisite on `Microsoft.VisualStudio.Component.CoreEditor` was causing issues in Dev17, so I removed it. The other two prerequisites I simply extended the version range.

I tested installing it into a private build of Dev17, and verified:
- Project and item templates still function as expected.
- Was able to create a new WinUI3 project, build, and run it successfully.


dev/VSIX/Directory.Build.props
```diff
@@ -7,7 +7,7 @@
             https://pkgs.dev.azure.com/microsoft/ProjectReunion/_packaging/Project.Reunion.nuget.internal/nuget/v3/index.json
```
BenJKuhn, line 7: Consider updating restore sources so that the build pipeline uses the ADO internal feed, but public builds use the version from nuget.org.

dev/VSIX/Extension/ProjectReunion.Extension.csproj
```diff
@@ -56,11 +56,11 @@
     <Reference Include="System" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.VisualStudio.SDK" Version="16.0.206" ExcludeAssets="runtime">
+    <PackageReference Include="Microsoft.VisualStudio.SDK" Version="17.0.0-previews-1-31325-273" ExcludeAssets="runtime">
```
BenJKuhn, line 59: Can you open a bug to track updating this to release version when available?

dev/VSIX/Extension/source.extension.vsixmanifest
```diff
@@ -14,17 +14,31 @@
...
+    </InstallationTarget>
+    <InstallationTarget Id="Microsoft.VisualStudio.Enterprise" Version="[17.0, 18.0)">
+      <ProductArchitecture>amd64</ProductArchitecture>
+    </InstallationTarget>
   </Installation>
   <Dependencies>
     <Dependency Id="Microsoft.Framework.NDP" DisplayName="Microsoft .NET Framework" d:Source="Manual" Version="[4.5,)" />
   </Dependencies>
   <Prerequisites>
-    <Prerequisite Id="Microsoft.VisualStudio.Component.CoreEditor" Version="[16.0,17.0)" DisplayName="Visual Studio core editor" />
```
BenJKuhn, line 25: Did this get renamed? Follow up and check on why this is needed, and if there's something this should be replaced with.

Fabi:
To run it from dev17 doesn't it also require an update to the Single-project MSIX Packaging Tools extension? Only using the ProjectReunion extension will result in a debug target error.

DefaultRyan:
> To run it from dev17 doesn't it also require an update to the Single-project MSIX Packaging Tools extension? Only using the ProjectReunion extension will result in a debug target error.

@Fabi could you elaborate on how you got the error you're referring to? I haven't seen it yet.

I tested this workflow successfully:
- Install this updated VSIX into Dev17
- Create a new project, using template "Blank App, Packaged with WAP (WinUI 3 in Desktop)
- Build project, and run with F5

I'm not what I would consider a VSIX expert, so if you have some insights I'll gladly listen. :)

wjk:
@DefaultRyan The error @Fabi was referring to only occurs if you copy the MSIX manifest and assets out of the wapproj into the application project, add `<EnablePreviewMsixTooling>true</EnablePreviewMsixTooling>` to the csproj file, and then try to run the csproj. While this isn’t the default MSIX packaging behavior (yet!), it is far superior to using wapproj.

DefaultRyan:
@Fabi and @wjk thanks. Somebody on our end is already aware of this and has the work scheduled. It's just out of scope for this particular PR.

dev/VSIX/Extension/source.extension.vsixmanifest
DefaultRyan: I had a weird install of VS 2022 Preview that wasn't resolving this Prerequisite. Did some reinstalling/updating and things work again, so I was able to restore this prerequisite.

dev/VSIX/Directory.Build.props
DefaultRyan: I think this might already be happening as a fallback mechanism - the packages that were restored when building locally appeared to come from nuget.org.

I'm not exactly sure why that would have happened, though. And the package restore versions should have a better story that's a little more connected to the real-life versions, and aren't this hard-coded 0.5.7 business. I've opened a task to track this and try to make it better.

dev/VSIX/Extension/ProjectReunion.Extension.csproj
DefaultRyan: Done.

dev/VSIX/Directory.Build.props
alwu-msft: Nuget.org is an upstream feed for our internal Nuget feed.

BenJKuhn (approved):

sylveon:
@wjk do you know if this option works for C++ projects?

wjk:
@sylveon I see no reason why it wouldn’t, although the IDE integration might be a bit screwy, since I don’t know if the VS bits have been tested with vcxproj as well as with csproj.

`Microsoft.ProjectReunion.WinUI.nupkg` contains props/targets that enable use in C++ projects. Inspection reveals that these targets include Microsoft.Build.Msix.targets if `EnableMsixPreviewTooling` is set to true; there is even a bit of C++-specific MSIX glue. Hope this helps!

PR #938, opened 2021-06-17
Remove build output modifying behavioral overrides from the command-line.

aeloros:
This change moves the needle of having the local dev inner loop output to be closer to the nightly build output.  Before these changes directory/file layout was different between the two.  This left it up to the contributing developer to reference more than one path when consuming the outputs of another project (A consumes B).


DrusTheAxe (approved):
Very nice

huichen123:
which project has the different output folder? It's the same for MRTCore. maybe it's the project setting?

Directory.Build.props
```diff
@@ -4,29 +4,28 @@
...
   <!-- Compilation differs for the build pipeline vs local development -->
   <PropertyGroup Condition="$(ProjectReunionBuildPipeline) != '1'">
     <!-- If not set, default it to no . -->
     <ProjectReunionBuildPipeline>0</ProjectReunionBuildPipeline>
   </PropertyGroup>
-  <Target Name="DirectoryBuildPropsInfo" Condition="$(ProjectReunionBuildPipeline) == '1'">
-    <Message Importance="High" Text="Directory.Build.props detects ProjectReunionBuildPipeline=$(ProjectReunionBuildPipeline)"/>
+
+  <Target Name="DirectoryBuildPropsInfo">
+    <Message Condition="$(ProjectReunionBuildPipeline) == '1'" Importance="High" Text="Directory.Build.props detects ProjectReunionBuildPipeline=$(ProjectReunionBuildPipeline)"/>
```
huichen123, line 28: >$(ProjectReunionBuildPipeline) [](http://example.com/codeflow?start=145&length=30)

print "ProjectReunionBuildPipeline=1" doesn't make too much sense to most people. Maybe a more human readable "detect running in build pipeline"?

build/AfterCpp.targets
```diff
@@ -0,0 +1,19 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <!-- <PostBuildDependsOn>$(PostBuildDependsOn);VerboseOutputLayoutInfo</PostBuildDependsOn> -->
+  </PropertyGroup>
+
+  <!-- Display the entire build output directory structure. -->
```
huichen123, line 8: > [](http://example.com/codeflow?start=2&length=61)

does this also run on local build?

Directory.Build.props
```diff
@@ -4,29 +4,28 @@
...
+  <Import Project="$(MSBuildThisFileDirectory)build\EnsureOutputLayout.props" Condition="$(EnsureOutputLayoutPropsIncluded) == ''" />
 
-    <OutputPath>$(OutDir)\$(MSBuildProjectName)\</OutputPath>
+  <PropertyGroup>    
+    <IsTDPConfiguration>false</IsTDPConfiguration>
 
     <GeneratedFilesDir>$(BaseIntermediateOutputPath)Generated Files\</GeneratedFilesDir>
     <GenerateProjectSpecificOutputFolder>True</GenerateProjectSpecificOutputFolder>
+
+    <CustomBeforeMicrosoftCommonTargets>$(BuildRoot)BeforeCommon.targets</CustomBeforeMicrosoftCommonTargets>
```
huichen123, line 17: ><CustomBeforeMicrosoftCommonTargets>$(BuildRoot)BeforeCommon.targets</CustomBeforeMicrosoftCommonTargets> [](http://example.com/codeflow?start=4&length=105)

unless this is specified in command line, it's probably useless to define here (already too late). I think we can remove this alltogether.

build/EnsureOutputLayout.props
```diff
@@ -0,0 +1,33 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <!-- Due to import ordering inconsistencies between different project types, we import this file
+         from multiple places.  We use this property to denote that it has been imported, to avoid
+         multiple import warnings. -->
+    <EnsureOutputLayoutPropsIncluded>true</EnsureOutputLayoutPropsIncluded>
```
huichen123, line 8: >EnsureOutputLayoutPropsIncluded [](http://example.com/codeflow?start=5&length=31)

i assume you planned to set based on this condition?

dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/UnpackagedTests/UnpackagedTests.csproj
```diff
@@ -83,7 +83,7 @@
   </Target>
   <!-- The powershell command below updates the TestAdaptersPaths path in the runsettings file, in order to allow the test to run in VS Test Explorer. -->
   <PropertyGroup>
-    <PostBuildEvent>copy /Y "$(OutputPath)..\Microsoft.Windows.ApplicationModel.Resources\Microsoft.Internal.FrameworkUdk.dll"  "$(OutputPath)Microsoft.Internal.FrameworkUdk.dll"
```
huichen123, line 86: >.. [](http://example.com/codeflow?start=42&length=2)

the path changed?

Directory.Build.props
DrusTheAxe: ProjectReunionBuildPipeline is an optional build parameter, and this block is a build debugging tool. I'd rather see "foo=bar" with the actual value than "foo is in use" or some such English prose.

This was also done to avoid assuming there'd be no other values in the future. =2, =foo, whatever. Depending how the build pipeline evolves, but not precluding it, and not making it more difficult than necessary if/when we choose to go there.

DrusTheAxe, line 28: Why move the condition from the Target parent to just the Message child?

huichen123: i'd agree if it always printed $(ProjectReunionBuildPipeline). However, here it only prints 1.

DrusTheAxe: I originally had it w/o the condition, which is fine for pipeline builds but it also caused spew in inner-loop builds The condition was put in (on the parent node) to avoid that noise

And it always prints $(ProjectReunionBuildPipeline), it just so happens that only has a value of 1 today. As I said, left open for other uses as the build pipelines future was very uncertain (and is still fuzzy)

PR #956, opened 2021-06-21
Microsoft.ApplicationModel.Resources.References.targets should not reference ProjectReunionFrameworkPackage

Scottj1s:
ProjectReunionFrameworkPackage should be used to distinguish between Project Reunion Framework Package mode (the default)  and InApp mode, and not used for transport package inner loop testing.  

Microsoft.ApplicationModel.Resources.References.targets currently enables local binplacing of MRTCore assemblies if ProjectReunionFrameworkPackage is 'true'.  The InApp support temporarily worked around this by defining _AddMrtCoreAssembliesToReferenceCopyLocalPaths to 'false' to prevent build breaks.  This condition will no longer work with recent InApp-related PRs.  MRTCore should instead use a distinct property for inner loop testing, as WinUI does (ProjectReunionWinUI), e.g.:

<_AddMrtCoreAssembliesToReferenceCopyLocalPaths Condition="'$(_AddMrtCoreAssembliesToReferenceCopyLocalPaths)'=='' AND '$(ProjectReunionMRTCore)' != 'false'">false</_AddMrtCoreAssembliesToReferenceCopyLocalPaths>


huichen123:
> 

we may keep this for potentially using Foundation.Transport package for inner loop testing.

---
Refers to: dev/MRTCore/packaging/Microsoft.ApplicationModel.Resources.References.targets:24 in 015e804. [](commit_id = 015e804028c6a42846308211b30f58e3f2c7aeb5, deletion_comment = True)

dev/MRTCore/packaging/Microsoft.ApplicationModel.Resources.References.targets
```diff
@@ -1,23 +0,0 @@
...
-<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-
-  <PropertyGroup>
-    <_MrtCoreRuntimeIdentifier Condition="'$(Platform)' == 'Win32'">x86</_MrtCoreRuntimeIdentifier>
-    <_MrtCoreRuntimeIdentifier Condition="'$(Platform)' != 'Win32'">$(Platform)</_MrtCoreRuntimeIdentifier>
-    <_MrtCoreDllDir>$(MSBuildThisFileDirectory)..\runtimes\win10-$(_MrtCoreRuntimeIdentifier)\native\</_MrtCoreDllDir>
-
-    <_AddMrtCoreAssembliesToReferenceCopyLocalPaths Condition="'$(_AddMrtCoreAssembliesToReferenceCopyLocalPaths)'=='' AND '$(Platform)' == 'AnyCPU'">false</_AddMrtCoreAssembliesToReferenceCopyLocalPaths>
-    <_AddMrtCoreAssembliesToReferenceCopyLocalPaths Condition="'$(_AddMrtCoreAssembliesToReferenceCopyLocalPaths)'=='' AND '$(ProjectReunionFrameworkPackage)' == 'true'">false</_AddMrtCoreAssembliesToReferenceCopyLocalPaths>
-    <_AddMrtCoreAssembliesToReferenceCopyLocalPaths Condition="'$(_AddMrtCoreAssembliesToReferenceCopyLocalPaths)'=='' AND '$(MSBuildProjectExtension)' == '.wapproj'">true</_AddMrtCoreAssembliesToReferenceCopyLocalPaths>
```
Scottj1s, line 13: @huichen123 is there a better condition to use than '$(ProjectReunionFrameworkPackage)' == 'true' ?  I'd like to keep ProjectReunionFrameworkPackage normally undefined, and only explicitly set to 'false' for InApp mode.  But I don't want that condition used here, or it will collide with the InApp mode support.

huichen123: if we want that property to be implicit, we have to introduce other property explicitly set for inapp mode.

Scottj1s: It sounds like you need another property in any case, to distinguish a third condition (MRT inner loop testing) from Project Reunion deployment, whether in Framework Package or InApp mode.  For the related WinUI PR, I used the ProjectReunionWinUI property for that purpose.  Should I just create something like that here?  ProjectReunionMRT?

dev/MRTCore/packaging/Microsoft.ApplicationModel.Resources.References.targets
```diff
@@ -9,7 +9,7 @@
     <_MrtCoreDllDir>$(MSBuildThisFileDirectory)..\runtimes\win10-$(_MrtCoreRuntimeIdentifier)\native\</_MrtCoreDllDir>
 
     <_AddMrtCoreAssembliesToReferenceCopyLocalPaths Condition="'$(_AddMrtCoreAssembliesToReferenceCopyLocalPaths)'=='' AND '$(Platform)' == 'AnyCPU'">false</_AddMrtCoreAssembliesToReferenceCopyLocalPaths>
-    <_AddMrtCoreAssembliesToReferenceCopyLocalPaths Condition="'$(_AddMrtCoreAssembliesToReferenceCopyLocalPaths)'=='' AND '$(ProjectReunionFrameworkPackage)' == 'true'">false</_AddMrtCoreAssembliesToReferenceCopyLocalPaths>
+    <_AddMrtCoreAssembliesToReferenceCopyLocalPaths Condition="'$(_AddMrtCoreAssembliesToReferenceCopyLocalPaths)'=='' AND '$(ProjectReunionMRTCore)' != 'false'">false</_AddMrtCoreAssembliesToReferenceCopyLocalPaths>
```
huichen123, line 12: >$(ProjectReunionMRTCore)' != 'false'" [](http://example.com/codeflow?start=124&length=37)

logically this would be "'$(ProjectReunionInApp)' == 'true'", even in reality when "'$(ProjectReunionMRTCore) == 'false'", $(ProjectReunionInApp) would always be true. OTOH, it's not clear to me what value it provides to distinguish the deployment type (reunion or standalone).

Scottj1s: ProjectReunionInApp no longer exists, and we should avoid reinventing it for just this special purpose. We can call the property whatever we like, as long as it's distinct from ProjectReunionFrameworkPackage, which can be true or false but only with Reunion deployment.  For standalone (this scenario) we need a third state (property).

huichen123: that's a made-up prop. As you are adding another implicit prop, I am wondering what's the collision you mentioned for "ProjectReunionFrameworkPackage != false" here?

Scottj1s: The collision that existed previously, which InApp support had to work around, was this:  
With ProjectReunionFrameworkPackage set to 'false', the condition here would fail (comparing to 'true') and fall through to binplacing dlls, which collides with InApp support.  InApp worked around that by setting _AddMrtCoreAssembliesToReferenceCopyLocalPaths to 'false' explicitly.  

To clean this up and remove the workaround, we need a distinct check for your inner loop tests.  ProjectReunionFrameworkPackage is no longer set to 'true', so even in Framework Package mode, this check will fail and attempt to binplace.  And in InApp mode, when it's set explicitly to 'false', it will also attempt to binplace.  In both cases, the behavior is wrong.  When in Reunion deployment, we should never binplace.  

So a separate property is needed.

huichen123: Thanks for the explanation! We may make it consistent (with WinUI), and create a separate props file to set it to true (which also provides a convenient place for comment). Also the name would be ProjectReunionFoundation as this is not MRT specific.

Scottj1s: sounds good - I'll make it happen

huichen123 (approved):
:shipit:

kythant (approved):

PR #978, opened 2021-06-24
Enable running 2 of the MRT Core test suites in Helix using the WinUI.Helix package

jefgen:
~~*[Note: This is currently a DRAFT change. Pending the WinUI.Helix Nuget package changes]*~~

This change enables running two of the MRT Core test suites in Helix in the feeder repo using the WinUI.Helix Nuget package.

The CI pipeline now runs the "UnitTest" and "BaseUnitTest" tests in Helix on RS5. This provides additional test coverage on a down-level version of Windows, as before the tests only ran on "`windows-latest`" (which is Server 2019).
(Note: Previously the "BaseUnitTests" were running in Helix. However the test results weren't being reported as part of the pipeline Test results.)

With this change, the test results from the Helix runs are now reported as part of the "Test" tab for the pipeline:

image

Expanding the results for the Helix queue shows the results from the two MRT Core tests:

image

The tests are prefixed with the feature area (MrtCore), test binary name, build arch, and build type.
(This is to allow adding additional tests in the future.)

This change uses the "`Microsoft.Internal.WinUI.Helix`" Nuget package, and removes the old/out-of-date checked-in copies of the Helix scripts from the repo. I've pushed this Nuget package to the "`ProjectReunion-Dependencies`" Nuget feed, so that other feeder repos (including this one) can use it.

I also pushed the "Microsoft.DotNet.Helix.Sdk" package to the "`ProjectReunion-Dependencies`" Nuget feed, as that is used by the `RunTestsInHelix.proj` file.

~~I also added `dotnet-eng` to the top-level `nuget.config` file, as this is needed for the "Microsoft.DotNet.Helix.Sdk".~~
~~Note: I hade to make a copy of the WinUI.Helix Nuget pipeline scripts in order to make changes to support running without a Helix Access token. I have a [pull-request](https://microsoft.visualstudio.com/WinUI/_git/microsoft-ui-xaml-lift/pullrequest/6189183) in the lifted WinUI XAML repo to update the scripts there. Once that is completed and the Nuget is updated, then the files can be removed from this PR.~~

This change is based on the [work](https://microsoft.visualstudio.com/ProjectReunion/_git/ProjectReunionInternal/pullrequest/6160055) done by Keith for enabling running test automation in the PR internal repo. (Thanks Keith!)

It uses a similar Project Structure, and adds the following two projects under the `test` folder:
- **ProjectReunion.Helix.TestCommon** includes the scaffolding code for setup and to run the tests in Helix.
- **ProjectReunion.Helix.Test.NetCore** is a dummy project to copy the .NET Core runtime DLLs to the build output in a way that satisfies TAEF.

Aside: This change also fixes a few minor nits found in other files while working on the changes.


build/ProjectReunion-BuildFoundation.yml
```diff
@@ -391,14 +391,6 @@ jobs:
 #    dependsOn: CreateNugetPackage
 #    pkgArtifactPath: '$(artifactDownloadPath)\fullnuget\FrameworkPackage'
```
kmahone, line 392: Delete.

kmahone:
> #    artifactName: 'FrameworkPkgTestsDrop'

Delete

---
Refers to: build/ProjectReunion-BuildFoundation.yml:406 in 4303c99. [](commit_id = 4303c995021158892ff066659e8375f76106afc2, deletion_comment = False)

kmahone:
> 

Delete.

---
Refers to: build/ProjectReunion-CI.yml:188 in 4303c99. [](commit_id = 4303c995021158892ff066659e8375f76106afc2, deletion_comment = False)

kmahone:
> #- template: AzurePipelinesTemplates\ProjectReunion-NugetReleaseTest-Job.yml

And delete the files that it is referring to.

---
Refers to: build/ProjectReunion-CI.yml:169 in 4303c99. [](commit_id = 4303c995021158892ff066659e8375f76106afc2, deletion_comment = False)

build/AzurePipelinesTemplates/ProjectReunion-RunHelixTests-Job.yml
```diff
@@ -1,28 +1,22 @@
...
       buildConfiguration: 'Release'
-      outputDirectory: BuildOutput\Release\x86\BaseUnitTests\
     Release_x64:
       buildPlatform: 'x64'
       buildConfiguration: 'Release'
-      outputDirectory: BuildOutput\Release\x64\BaseUnitTests\
+  
+  # Both Azure DevOps and Helix don't have support for arm64.
+  conditionForRunTests: and(succeeded(), ne(variables['buildPlatform'], 'arm64'))
+  conditionForProcessTestResults: and(succeededOrFailed(), ne(variables['buildPlatform'], 'arm64'))
```
kmahone, line 16: I don't think you need this.

I added this in the other repo because a I wanted to build the tests and run the tests in the same job, but for arm64 I only wanted to build. It was an ugly hack to get it to work in a single Job (which is good when there is latency on the build machine queue).

I don't think that is necessary for this case, since nothing that interesting is being built in this Job - correct?

build/AzurePipelinesTemplates/ProjectReunion-RunHelixTests-Job.yml
```diff
@@ -32,99 +26,188 @@ jobs:
...
-    helixCommonArgs: '/binaryLogger:$(Build.SourcesDirectory)/${{parameters.name}}.$(buildPlatform).$(buildConfiguration).binlog /p:HelixBuild=$(Build.BuildId).$(buildPlatform).$(buildConfiguration) /p:Platform=$(buildPlatform) /p:Configuration=$(buildConfiguration) /p:HelixType=${{parameters.helixType}} /p:TestSuite=${{parameters.testSuite}} /p:ProjFilesPath=$(Build.ArtifactStagingDirectory) /p:rerunPassesRequiredToAvoidFailure=${{parameters.rerunPassesRequiredToAvoidFailure}}'
+    # This should be kept in sync with the value in these files:
+    #   build\Helix\packages.config
+    #   test\ProjectReunion.Helix.TestCommon\ProjectReunion.Helix.TestCommon.csproj
+    winUIHelixVer: 0.5.0
+    winUIHelixPipelineScripts: build\Helix\packages\Microsoft.Internal.WinUI.Helix.$(winUIHelixVer)\scripts\pipeline
+
+    # TODO: I had to make changes on top of the WinUI Helix Nuget to fix things for the PR feeder repo.
+    # Once the nuget is updated with the modifications then these files can be removed from the repo.
+    modifiedWinUIHelixPipelineScripts: build\Helix\pipeline
```
kmahone, line 37: Let's do that before merging this PR.

jefgen: Yes, that's why this PR is a draft. :)

build/AzurePipelinesTemplates/ProjectReunion-RunHelixTests-Job.yml
jefgen: Right, good point.
 For arm64 it would only be building the two projects
"ProjectReunion.Helix.TestCommon" and "ProjectReunion.Helix.Test.NetCore" -- which are pretty much useless as they don't actually "test" anything.

I'll remove this. Thanks!

build/ProjectReunion-BuildFoundation.yml
jefgen: Good point.
It looks like the solution doesn't exist either (`test\ProjectReunion-ReleaseTest`). I'll remove this and the related files.
Thanks!

jefgen:
> #    artifactName: 'FrameworkPkgTestsDrop'

Just to be (perhaps pedantically) clear, you mean the whole block and not just the line for `artifactName`, right?

(ie: I think we should also remove the file for `ProjectReunion-WACKTests-Job.yml`)

---
In reply to: [867963032](https://github.com/microsoft/WindowsAppSDK/pull/978#issuecomment-867963032) [](http://example.com/codeflow?ancestors=867963032,867963032)

---
Refers to: build/ProjectReunion-BuildFoundation.yml:406 in 4303c99. [](commit_id = 4303c995021158892ff066659e8375f76106afc2, deletion_comment = False)

kmahone:
> #    artifactName: 'FrameworkPkgTestsDrop'

Yes. There are no WACK tests in this repo. 

---
In reply to: [867977402](https://github.com/microsoft/WindowsAppSDK/pull/978#issuecomment-867977402) [](http://example.com/codeflow?ancestors=867977402,867963032)

---
Refers to: build/ProjectReunion-BuildFoundation.yml:406 in 4303c99. [](commit_id = 4303c995021158892ff066659e8375f76106afc2, deletion_comment = False)

nuget.config
```diff
@@ -4,6 +4,7 @@
     <clear />
     <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
     <add key="ProjectReunion-Dependencies" value="https://pkgs.dev.azure.com/ms/ProjectReunion/_packaging/ProjectReunion-Dependencies/nuget/v3/index.json" />
+    <add key="dotnet-eng" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json" />
```
huichen123, line 7: ><add key="dotnet-eng" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json" /> [](http://example.com/codeflow?start=4&length=115)

we are not supposed to add multiple sources

build/AzurePipelinesTemplates/ProjectReunion-BuildHelixTestProject-Steps.yml
```diff
@@ -0,0 +1,33 @@
+parameters:
```
huichen123, line 1: this appears a generic build job instead of a helix test specific.

nuget.config
jefgen: Hmm, in that case I'll probably have to push the Helix SDK to the ProjectReunion-Dependencies feed.

huichen123: yeah, it appears what you are trying to add is internal feed, and we cannot use internal feed.

build/AzurePipelinesTemplates/ProjectReunion-BuildHelixTestProject-Steps.yml
jefgen: Good point. It's probably better to rename this to something else... will update.

nuget.config
jefgen: Well the `dotnet-eng` feed is a public feed from the .NET Core team.
You can browse the feed here (without signing in):
https://dev.azure.com/dnceng/public/_packaging?_a=feed&feed=dotnet-eng

It's also publicly documented here:
https://github.com/dotnet/arcade/blob/main/src/Microsoft.DotNet.Helix/Sdk/Readme.md

But I think we can only have one source though, so might need to publish it to `ProjectReunion-Dependencies` feed anyways.

huichen123: oh, i thought it was for "Microsoft.Internal.WinUI.Helix", and assumed it's internal based on the name. 
Which feed does "Microsoft.Internal.WinUI.Helix" come from?

build/build-mrt.yml
```diff
@@ -93,12 +90,14 @@ steps:
...
 
+# TODO: Is this actually used? (ie: What installs it?)
 - task: powershell@2
-  displayName: 'Download .net runtime Windows installer'
+  displayName: 'Download .NET Runtime Windows installer'
   inputs:
     targetType: filePath
     workingDirectory: ${{ parameters.MRTSourcesDirectory }}\build
     filePath: ${{ parameters.MRTSourcesDirectory }}\build\DownloadDotNetRuntimeInstaller.ps1
+    arguments: -Platform $(buildPlatform) -OutputDirectory $(Build.SourcesDirectory)\BuildOutput\$(buildPlatform)\$(buildConfiguration)\TestDependencies
```
huichen123, line 100: >arguments: -Platform $(buildPlatform) -OutputDirectory $(Build.SourcesDirectory)\BuildOutput\$(buildPlatform)\$(buildConfiguration)\TestDependencies [](http://example.com/codeflow?start=4&length=148)

what is this for?

nuget.config
jefgen: I put it in the PR description :)
> This change uses the "Microsoft.Internal.WinUI.Helix" Nuget package, and removes the old/out-of-date checked-in copies of the Helix scripts from the repo. I've pushed this Nuget package to the "ProjectReunion-Dependencies" Nuget feed, so that other feeder repos (including this one) can use it.

@kmahone -- I think Hui has a good point that the word "internal" in the name sort of implies that it might not be public.
FWIW, the public WinUI repo is already using this Nuget package though. 
See `packages.config` here: https://github.com/microsoft/microsoft-ui-xaml/blob/main/build/Helix/packages.config#L8

Perhaps we could consider renaming it in the future maybe?

build/build-mrt.yml
```diff
@@ -122,6 +121,30 @@ steps:
     PathtoPublish: $(Build.SourcesDirectory)/mrtcore.$(buildPlatform).$(buildConfiguration).binlog
     artifactName: binlogs
 
+# Copy and Publish the test binaries so we can run them in Helix later.
```
huichen123, line 124: do this after VSTest?

build/build-mrt.yml
jefgen: I updated the DownloadDotNetRuntimeInstaller.ps1 script to take explicit paramters instead of inferring things from environment variables, so that I could share the script with the Helix job.

build/build-mrt.yml
jefgen: sure, good idea. probably want to make sure the tests pass before creating any artifacts.

axelandrejs (approved):

nuget.config
jefgen: Update: I pushed the "Microsoft.DotNet.Helix.Sdk" Nuget package to the "ProjectReunion-Dependencies" feed, so we don't need to add another feed to the `nuget.config`.

huichen123: It appears you just changed the version of "Microsoft.DotNet.Helix.Sdk". why was the need for a new feed (or adding it to a feed)?

jefgen: Right, I did bump the version for the "Microsoft.DotNet.Helix.Sdk". (I changed it to match the version used by the WinUI repo.)

The need for the new feed (or adding it to a feed) is coming from this step: "`NuGet restore Helix/RunTestsInHelix.proj`", in the file `build\AzurePipelinesTemplates\ProjectReunion-RunHelixTests-Job.yml`.

This does a Nuget restore on the .proj file, which pulls in the "Microsoft.DotNet.Helix.Sdk" package.
Before there was no Nuget restore being done -- so it wasn't being pulled in.

jefgen: Actually, from testing, it doesn't seem like doing the Nuget restore for the proj is *strictly necessary*.
Though I'm somewhat inclined to keep it, so that we're consistent with how it's being used in the PR-Internal repo and the WinUI repo.

build/AzurePipelinesTemplates/ProjectReunion-RunHelixTests-Job.yml
```diff
@@ -32,99 +22,173 @@ jobs:
...
-    helixCommonArgs: '/binaryLogger:$(Build.SourcesDirectory)/${{parameters.name}}.$(buildPlatform).$(buildConfiguration).binlog /p:HelixBuild=$(Build.BuildId).$(buildPlatform).$(buildConfiguration) /p:Platform=$(buildPlatform) /p:Configuration=$(buildConfiguration) /p:HelixType=${{parameters.helixType}} /p:TestSuite=${{parameters.testSuite}} /p:ProjFilesPath=$(Build.ArtifactStagingDirectory) /p:rerunPassesRequiredToAvoidFailure=${{parameters.rerunPassesRequiredToAvoidFailure}}'
+    # This should be kept in sync with the value in these files:
+    #   build\Helix\packages.config
+    #   test\ProjectReunion.Helix.TestCommon\ProjectReunion.Helix.TestCommon.csproj
+    winUIHelixVer: 0.5.2
+    winUIHelixPipelineScripts: build\Helix\packages\Microsoft.Internal.WinUI.Helix.$(winUIHelixVer)\scripts\pipeline
+
+    # The target queues to run the tests on.
+    # Currently runs on queues owned by the WinUI/XAML team. This should run on queues hosted by ProjectReunion in the future.
+    helixTargetQueues: 'Windows.10.Amd64.ClientRS5.Open.Xaml'
```
kmahone, line 33: Can you set things to run on non-Open queues when running on an internal Pipeline. See [here](https://github.com/microsoft/microsoft-ui-xaml/blob/090fd4d11be744473324056d8d00dce797cdda6c/build/AzurePipelinesTemplates/MUX-RunHelixTests-Job.yml#L147) for an example.

jefgen: Thanks for the example!
Are there closed (non-Open) Helix queues that we could (or should) use for ProjectReunion?
ie: Did you want me to update this to use non-Open queues? (I'm not sure if this was an ask, or if you were pointing out how to use both types of queues...)

kmahone: When running outside of dev.azure.com/ms, use the closed queues, which have the same name, just without the .Open part of the name.

jefgen: Thanks! Will update in a moment.

jefgen:
Note: I rebased onto main in order to resolve merge conflicts, then force-pushed the branch.

jefgen:
Note: There is one check that keeps showing as "waiting" on this PR ([ProcessTestResults](https://github.com/microsoft/WindowsAppSDK/blob/1714557cad5e740c0153e451fbf0311c8e5bdfc1/build/AzurePipelinesTemplates/ProjectReunion-ProcessTestResults-Job.yml#L10)). However, it is actually removed in this change, so it shouldn't be shown at all. I think GitHub/Azure is getting confused somehow.

axelandrejs (approved):

build/AzurePipelinesTemplates/ProjectReunion-RunHelixTests-Job.yml
jorvanMS: This looks good, do we want to target more versions than just RS5?

jorvanMS (approved):

erik0686 (approved):

build/AzurePipelinesTemplates/ProjectReunion-RunHelixTests-Job.yml
jefgen: Good point. I thought we could start with RS5 for now (since that is the lowest version supported), and add more later based on need. I also didn't want to slow down the PR builds too much. Perhaps we can try experimenting with adding other versions and see what the impact is to the time for builds.

jefgen:
Note: I had to push another (small) commit to handle the changes from https://github.com/microsoft/WindowsAppSDK/commit/6404a51fd3ae106b237f2855c3d0f50cb36dfa32 which recently got merged to `main`.

PR #985, opened 2021-06-25
Update README.md

LeftTwixWand:
Updated path for Bugs and Feature proposal badges according to the new repository name


LeftTwixWand:
The issue has already fixed

PR #1016, opened 2021-07-02
Add 19h1 to Helix queues, plus minor pipeline clean up.

jefgen:
This change adds 19H1 to list of OS versions that the Helix tests run on. It also cleans up some of the pipeline.

(This also renames the `displayName: 'Discover tests'` in the pipeline to be more descriptive. Thanks to Alex for pointing this out.)

Note: Since the Helix tests run in parallel, this doesn't add any additional time to the overall build. (It still takes ~35-38 mins)


build/AzurePipelinesTemplates/ProjectReunion-RunHelixTests-Job.yml
```diff
@@ -30,8 +30,9 @@ jobs:
 
     # The target queues to run the tests on.
     # Currently runs on queues owned by the WinUI/XAML team. This should run on queues hosted by ProjectReunion in the future.
-    helixTargetQueuesOpen: 'Windows.10.Amd64.ClientRS5.Open.Xaml'
-    helixTargetQueuesClosed: 'Windows.10.Amd64.ClientRS5.Xaml'
+    # Note: %3b is the escape sequence for ';' which is used as the delimiter
```
rohanp-msft, line 33: >escape sequence for ';' [](http://example.com/codeflow?start=23&length=23)

nit: is %3b not the encoding for ; rather than an escape sequence?

rohanp-msft (approved):
:shipit:

huichen123 (approved):

build/AzurePipelinesTemplates/ProjectReunion-RunHelixTests-Job.yml
jefgen: Hmm, well I guess I've seen it referred to both ways... (Some [sites](https://www.urlencoder.io/learn/) use "URL encoding", [others](
https://www.werockyourweb.com/url-escape-characters/) use "URL escaping").

For me, I like to think of it like this:
U+003B is the encoding for the semicolon character.
The % that is used for URLs is to "escape" the character, so the `%3b` is the "escape sequence" for the semicolon.

But maybe that is just me... :)

rohanp-msft: Got it. Thanks for explaining, Jeff.

jefgen: No problem! Thank you very kindly for taking a look at the PR! :)

PR #1018, opened 2021-07-02
Add resource project for WindowsAppSDK telemetry

ZenBird:
This project add a resource dll for telemetry resources

    * Adding a script to auto generate the rc file for the resource dll that will contain the values needed to log PartB telemetry.
    * The .rc file for the resource project is generated from the resource.templ file checked in with the project.
    * The script must be executed before running the build step on this solution.
    * Adding ReunionTelemetry.h that will reference this resource dll to construct the PartB structure.


ZenBird:
@MythiliM is added to the review. #Closed

ProjectReunion.sln
```diff
@@ -120,6 +120,8 @@ Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "ManualTestApp", "test\TestA
 EndProject
 Project("{C7167F0D-BC9F-4E6E-AFE1-012C56B48DB5}") = "AppLifecycleTestPackage", "test\TestApps\AppLifecycleTestPackage\AppLifecycleTestPackage.wapproj", "{A3FBA80D-5B35-471F-9A45-DB4B29E195B9}"
 EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "ProjectReunionTelemetryResource", "dev\ProjectReunionTelemetryResource\ProjectReunionTelemetryResource.vcxproj", "{0F2BDB3F-B785-423E-B09F-F395754ADD6B}"
+EndProject
```
ZenBird, line 124: Revert. This project should not be part of ReunionCore. This is part of a separate solution that will be invoked in the aggregation step.

dev/Common/ReunionTelemetry.h
```diff
@@ -0,0 +1,55 @@
...
+#define _REUNION_EXPERIMENTATION_FLAG_RESOURCEID    10002
+
+static std::wstring s_ReunionPackageVer = L"";
+static std::wstring s_ExperimentationLevel = L"not_initialized";
+
+wchar_t* REUNION_PACKAGE_VER()
+{
+   if(s_ReunionPackageVer.empty())
+   {
+       HMODULE resourceDll = LoadLibraryW(L"ProjectReunionTelemetryResource.dll");
```
jefgen, line 20: Does this code need to handle errors? (Should it check if the return is `NULL`? )

dev/Common/ReunionTelemetry.h
```diff
@@ -0,0 +1,55 @@
...
+   }
+
+   return s_ReunionPackageVer.c_str();
+}
+
+wchar_t* REUNION_EXPERIMENTATION_LEVEL()
+{
+    if (s_ExperimentationLevel.compare(L"not_initialized") == 0)
+    {
+        HMODULE resourceDll = LoadLibraryW(L"ProjectReunionTelemetryResource.dll");
```
jefgen, line 34: Same question here about checking for errors.

dev/Common/ReunionTelemetry.h
```diff
@@ -0,0 +1,55 @@
...
+#define __REUNIONTELEMETRY_INCLUDED
+
+#ifdef __WIL_TRACELOGGING_H_INCLUDED
+#error "WIL Tracelogging.h must not be explicitly included when including this file"
+#endif
+
+#define _REUNION_PACKAGE_VER_RESOURCEID             10001
+#define _REUNION_EXPERIMENTATION_FLAG_RESOURCEID    10002
+
+static std::wstring s_ReunionPackageVer = L"";
```
jefgen, line 13: I vaguely wonder about having a static variable like this in header that will presumably be used in many .cpp files.
Won't this mean that each translation unit (.cpp file) gets its own copy of the variable? (ie: bloat).

dev/Common/ReunionTelemetry.h
ZenBird: If the process is unable to load the resource dll, we must have a way of signaling the configuration error.

dev/Common/ReunionTelemetry.h
ZenBird: Is there an alternative?

jonwis: Yes; define the methods here in the header, and put the implementations into a .cpp file that compiles into a static lib.

Use a function local static:

```c++
inline std::wstring load_resource_string(const char* dllName, uint32_t id) {
    wil::unique_hmodule module{::LoadLibraryA(dllName)};
    if (module) {
        wchar_t buffer[256];
        LoadStringW(module, id, buffer, ARRAYSIZE(buffer));
        return buffer;
    } else {
        return { };
    }
} 

inline const wchar_t* WINAPPSDK_PACKAGE_VER()
{
    static std::wstring packageVer = load_resource_string("ProjectReunionTelemetryResource.dll", _REUNION_PACKAGE_VER_RESOURCEID);
    return packageVer.c_str();
}
```

The runtime will ensure that static initializer is called exactly once and then the result is reused to be returned to future callers.

dev/Common/ReunionTelemetry.h
jonwis: Why? If it can't load the resource DLL it should act as if it wasn't there and the package version and configuration settings aren't present.  That is, "missing the DLL means not configured."

dev/Common/ReunionTelemetry.h
```diff
@@ -0,0 +1,63 @@
...
+    return s_ExperimentationLevel.c_str();
+}
+
+
+#define REUNION_SDK_VER REUNION_PACKAGE_VER()
+#define REUNION_EXPERIMENTATION_LEVEL REUNION_EXPERIMENTATION_LEVEL()
+
+#define _GENERIC_PARTB_FIELDS_ENABLED \
+        TraceLoggingStruct(3, "PartB_Reunion_API"), \
+        TraceLoggingString(REUNION_SDK_VER, "Reunion package version"), \
```
jonwis, line 58: This project is now called the "Windows App SDK".  Please replace "Reunion" with "WinAppSDK" everywhere (including in the file names, project names, etc.)

dev/ProjectReunionTelemetryResource/ProjectReunionTelemetryResource.sln
```diff
@@ -0,0 +1,31 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.31410.357
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "ProjectReunionTelemetryResource", "ProjectReunionTelemetryResource\ProjectReunionTelemetryResource.vcxproj", "{D69DC3ED-C753-455D-A3D1-D0B1329E76CE}"
+EndProject
+Global
+	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|x64 = Debug|x64
```
jonwis, line 10: Please add ARM64 targets as well.

jonwis (changes requested):

dev/Common/WindowsAppSDKTelemetry.h
```diff
@@ -0,0 +1,24 @@
...
+
+#ifdef __WIL_TRACELOGGING_H_INCLUDED
+#error "WIL Tracelogging.h must not be explicitly included when including this file"
+#endif
+
+char* WINDOWSAPPSDK_PACKAGE_VER();
+char* WINDOWSAPPSDK_EXPERIMENTATION_LEVEL();
+
+#define _GENERIC_PARTB_FIELDS_ENABLED \
+        TraceLoggingStruct(4, "PartB_WINDOWSAPPSDK_API"), \
```
MythiliM, line 17: Can this be COMMON_WINDOWSAPPSDK_PARAMS?

dev/Common/WindowsAppSDKTelemetry.h
```diff
@@ -0,0 +1,24 @@
...
+        TraceLoggingString(WINDOWSAPPSDK_PACKAGE_VER(), "Windows App SDK package version"), \
```
MythiliM, line 18: Please don't use spaces for your column names, retrieving that in Scope, MCNative might become interesting later.

MythiliM: Please rename the column name from "Windows App SDK package version" to SDKVersion

dev/Common/WindowsAppSDKTelemetry.h
```diff
@@ -0,0 +1,24 @@
...
+        TraceLoggingString(WINDOWSAPPSDK_EXPERIMENTATION_LEVEL(), "Windows App SDK experimentation level"), \
```
MythiliM, line 19: Rename column name to SDKExperimentationLevel

PR #1026, opened 2021-07-06
Commonize Hybrid CRT support

DrusTheAxe:
Move hybrid CRT support out of individual project files and up to the project level as a whole

This is a build generalization of [Use the Universal C Runtime #888](https://github.com/microsoft/WindowsAppSDK/pull/888). It changes no product binaries, just test binaries that weren't previously using the hybrid CRT technique.

Verified by comparing file sizes and import (LINK /DUMP /IMPORTS) of binaries' previously using hybrid CRT (before vs after). Also reviewed app binaries' (product+test) imports to verify no VS CRT imports.

NOTE: This PR intentionally doesn't change the installer as its build structure's a bit different and needs additional changes. That'll come in a following PR


riverar:
Wow, it looks like VC16 is now explicitly blocking the hybrid CRT scenario? 👀

DrusTheAxe:
> Wow, it looks like VC16 is now explicitly blocking the hybrid CRT scenario? 👀 Or were those MRT build failures always there?

That's a surprise. MRT always built before, all this did was move the definition to an earlier point in the build flow. And MRT failed but ProjectReunion.sln is built before MRT and no problems there? How peculiar. Spelunking...

Scottj1s:
> > Wow, it looks like VC16 is now explicitly blocking the hybrid CRT scenario? 👀 Or were those MRT build failures always there?
> 
> That's a surprise. MRT always built before, all this did was move the definition to an earlier point in the build flow. And MRT failed but ProjectReunion.sln is built before MRT and no problems there? How peculiar. Spelunking...

dev\MRTCore\mrt\Microsoft.Windows.ApplicationModel.Resources\src\Microsoft.Windows.ApplicationModel.Resources.vcxproj is building for store deployment, which prohibits static CRT.  Should be able to simply remove these:
    <AppContainerApplication>true</AppContainerApplication>    
    <ApplicationType>Windows Store</ApplicationType>
    <ApplicationTypeRevision>10.0</ApplicationTypeRevision>

riverar:
@Scottj1s Good catch!

DrusTheAxe:
@huichen123 does this make any sense to you? OK it can't find a DLL but which one? I suspect it can't find `Microsoft.Internal.FrameworkUdk.dll` but I don't understand why this would be popping up now and not months ago? Was there a recent MRT change adding a dependent on the framework? But why show up here and not elsewhere before now? I'm very puzzled

@huichen123 this fails running ManagedTest. How do you manually run that if you build everythin in MrtCore.sln?

@alexlamtest is there a known issue in the Helix test environment?

```
2021-08-10T05:03:51.6127288Z Successfully launched process Id '5360'.
2021-08-10T05:03:53.7457614Z Closing app with package full name '0ef6c1bb-5f8e-4e8d-8db2-d5c115a1b166_1.0.0.0_x64__8wekyb3d8bbwe'.
2021-08-10T05:03:54.0638022Z   Failed GetStringTest [196 ms]
2021-08-10T05:03:54.0653326Z   Error Message:
2021-08-10T05:03:54.0654658Z    Test method ManagedTest.ResourceLoaderTest.GetStringTest threw exception: 
2021-08-10T05:03:54.0656774Z System.IO.FileNotFoundException: The specified module could not be found. (Exception from HRESULT: 0x8007007E)
2021-08-10T05:03:54.0658240Z   Stack Trace:
2021-08-10T05:03:54.0659366Z       at System.StubHelpers.StubHelpers.GetWinRTFactoryObject(IntPtr pCPCMD)
2021-08-10T05:03:54.0660461Z    at Microsoft.ApplicationModel.Resources.ResourceLoader..ctor(String fileName)
2021-08-10T05:03:54.0663480Z    at ManagedTest.ResourceLoaderTest.GetStringTest()
2021-08-10T05:03:54.0676677Z   Failed GetStringTest_NonDefaultNamespace [2 ms]
```

DrusTheAxe:
This failure is a known issue http://task.ms/34642605 :-(

Following up with @kythant

huichen123:
> This failure is a known issue http://task.ms/34642605 :-(
> 
> Following up with @kythant

your failure is not a known issue. Locally I run the test with test explorer in VS.

DrusTheAxe:
> your failure is not a known issue. Locally I run the test with test explorer in VS.

Indeed. It appears to be something specific to the machine pool / environment used to run Helix (which runs the test). Following up with @kythant

huichen123 (changes requested):
Please do not disable the test.

EHO-makai (commented):
Did the build issues get resolved?  What test is failing?  I couldn't quite follow the details from the conversation.  

Overall, it looks fine but it looks like there are some tail end issues blocking completion?

DrusTheAxe:
> Did the build issues get resolved? 

If you mean the test issue, no. Under discussion

> What test is failing?

MrtCoreManagedTest

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/unittests/MrtCoreManagedTest.csproj
```diff
@@ -153,6 +154,9 @@
     </PackageReference>
   </ItemGroup>
   <ItemGroup>
+    <None Include="MrtCoreManagedTests.runsettings">
```
huichen123, line 157: ><None Include="MrtCoreManagedTests.runsettings"> [](http://example.com/codeflow?start=4&length=48)

what makes this a new requirement?

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/UnpackagedTests/MrtCoreUnpackagedTests.csproj
```diff
@@ -84,17 +84,18 @@
   <!-- The powershell command below updates the TestAdaptersPaths path in the runsettings file, in order to allow the test to run in VS Test Explorer. -->
   <PropertyGroup>
     <PostBuildEvent>copy /Y "$(OutputPath)..\Microsoft.ApplicationModel.Resources\Microsoft.Internal.FrameworkUdk.dll"  "$(OutputPath)Microsoft.Internal.FrameworkUdk.dll"
+    copy /Y "$(OutputPath)..\MRM\MRM.dll"  "$(OutputPath)MRM.dll"
```
huichen123, line 87: >copy /Y "$(OutputPath)..\MRM\MRM.dll"  "$(OutputPath)MRM.dll" [](http://example.com/codeflow?start=4&length=61)

is this required? I see it in the output folder.

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/unittests/MrtCoreManagedTest.csproj
DrusTheAxe: Same as for MrtCoreUnpackagedTests. Removing that file and the powershell postbuildevent to update its path results in "Test run aborted"

MSTest doesn't run the test without it.

huichen123: strange.

huichen123 (approved):
:shipit:

Directory.Build.props
```diff
@@ -44,4 +44,7 @@
       <AdditionalIncludeDirectories>$(SolutionDir)\build\override;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
     </Midl>
   </ItemDefinitionGroup>
+
+  <!-- Build with the hybrid CRT (Universal CRT + Static VS CRT (for what little the Universal CRT doesn't cover) -->
+  <Import Project="$(MSBuildThisFileDirectory)HybridCRT.props" />
```
huichen123, line 49: >HybridCRT [](http://example.com/codeflow?start=46&length=9)

there is some insistency with EnsureOutputLayout.props. That one is under build folder.

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/UnpackagedTests/MrtCoreUnpackagedTests.csproj
DrusTheAxe: > > copy /Y "$(OutputPath)..\MRM\MRM.dll"  "$(OutputPath)MRM.dll"
> 
> is this required? I see it in the output folder.

Yes. Those copy commands can probably be replaced with ReferenceCopyLocalPaths like MrtCoreManagedTest but in the spirit of surgically unblocking I've tried to avoid making larger changes. The copy command this way doesn't work for the managed (packaged) test because it only copies into $(OutDir) but MSTest actually creates $(OutDir)AppX with copies from OutDir **of what it thinks belongs**.

You could change MrtCoreUnpackagedTest to use ReferenceCopyLocalPaths but it's worth looking into removing the need for the runsettings file. Perhaps with more love in its [AssemblyInitialize] (copy files, call AddDllDirectory, other). Something to consider in the future

Directory.Build.props
DrusTheAxe: That's the only one under build, all other *.props are in the root. And it's unclear if "build" is for all builds or pipeline builds, so better to stick with the crowd and have someone and have someone (sooner or later) take a holistic look

PR #1031, opened 2021-07-06
Switch to using private hosted pool for Project Reunion Foundation

kythant:
Switch to using private hosted pool for Project Reunion Foundation


build/ProjectReunion-BuildFoundation.yml
```diff
@@ -13,8 +13,7 @@ parameters:
 
 jobs:
 - job: CredScan
-  pool:
-    vmImage: 'windows-2019'
+  pool: $(ProjectReunionBuildPool)
```
huichen123, line 16: >ProjectReunionBuildPool [](http://example.com/codeflow?start=10&length=23)

do we need to have a default value in yml?

kythant: I don't think we need to have one.

huichen123:
>     name: Package ES Lab E

are they going to decommission those Package ES lab very soon?

---
Refers to: build/ProjectReunion-BuildFoundation.yml:187 in 992ffc4. [](commit_id = 992ffc42184f41947a5b3e9a7d79ae8d2a852bb2, deletion_comment = False)

huichen123 (approved):
:shipit:

PR #1035, opened 2021-07-08
Adding PowerNotifcation APIs

huzaifa-d:
Adding the PowerNotification APIs as mentioned in this [spec](https://github.com/microsoft/WindowsAppSDK/blob/main/specs/AppLifecycle/StateNotifications/AppLifecycle%20StateNotifications.md)


test/PowerNotifications/PowerTests.cpp
```diff
@@ -0,0 +1,205 @@
...
+#include "winrt/Microsoft.ProjectReunion.h"
+
+using PowerEventHandle =
+winrt::Windows::Foundation::EventHandler<winrt::Windows::Foundation::IInspectable>;
+
+using namespace winrt::Microsoft::ProjectReunion;
+
+namespace ProjectReunionPowerTests
+{
+    constexpr auto TIMEOUT = 2000;
```
DrusTheAxe, line 14: Why 2000?

What's the expected average and worst case timeouts you'll encounter?

test/PowerNotifications/PowerTests.cpp
```diff
@@ -0,0 +1,205 @@
...
+        TEST_METHOD(GetRemainingChargePercent)
+        {
+            auto stat = PowerManager::RemainingChargePercent();
+            VERIFY_ARE_EQUAL(stat, 100);
+        }
+
+        TEST_METHOD(CompositeBatteryStatusCallback)
+        {
+            // This test can only be run on systems/VM with a battery
+            // The values will need to be changed to reflect real-time values
```
DrusTheAxe, line 44: `//TODO:...` ?

test/PowerNotifications/PowerTests.cpp
```diff
@@ -0,0 +1,205 @@
...
+            auto stat = PowerManager::RemainingDischargeTime();
+            VERIFY_ARE_EQUAL(stat.count(), -10000000);
+        }
+
+        TEST_METHOD(RemainingDischargeTimeCallback)
+        {
+            static wil::unique_handle event(CreateEvent(nullptr, false, false, nullptr));
+            THROW_LAST_ERROR_IF_NULL(event.get());
+
+            auto stat = winrt::Windows::Foundation::TimeSpan(std::chrono::seconds(1));
```
DrusTheAxe, line 80: `...TimeSpan(1s);`

std::chrono defines a bunch of funky operators so you can declare simple literals, e.g. `operator ""ms(unsigned long long ms)` is why `10ms` works. You can see the full set of chrono literal constant suffixes @ https://en.cppreference.com/w/cpp/chrono/duration

test/PowerNotifications/PowerTests.dll.manifest
```diff
@@ -0,0 +1,12 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<assembly manifestVersion="1.0" xmlns="urn:schemas-microsoft-com:asm.v1">
+    <assemblyIdentity version="1.0.0.0" name="MyApplication.app"/>
+
+    <file name="Microsoft.ProjectReunion.dll">
```
DrusTheAxe, line 5: You'll want to delete this.

If you use the Bootstrap API you can test use your WinRT objects from the test WASfwk package like the standard MSIX distribution model. See email for more details.

Also, it looks like this Fusion manifest is solely to use RegFreeWinRT to get at your WinRT API. That won't work on RS5 (w/o Reunion) so your test will fail in some of the Helix configurations (we run tests across all supported OSes, which today means RS5+). The ::TP::Bootstrap solution works on all supported platforms.

DrusTheAxe (changes requested):

test/PowerNotifications/PowerTests.cpp
dhoehna, line 14: c_Timeout

test/PowerNotifications/PowerTests.cpp
```diff
@@ -0,0 +1,205 @@
...
+using namespace winrt::Microsoft::ProjectReunion;
+
+namespace ProjectReunionPowerTests
+{
+    constexpr auto TIMEOUT = 2000;
+    class PowerTests
+    {
+    public:
+        BEGIN_TEST_CLASS(PowerTests)
+            TEST_CLASS_PROPERTY(L"ActivationContext", L"PowerTests.dll.manifest")
```
dhoehna, line 19: @DrusTheAxe DO these tests need to run for different IL and AC levels like Environmentmanager?

dev/PowerNotifications/PowerNotifications.cpp
```diff
@@ -0,0 +1,316 @@
...
+
+    // EnergySaverStatus Functions
+    EventType& EnergySaverStatus_Event()
+    {
+        return make_self<factory_implementation::PowerManager>()->m_energySaverStatusChangedEvent;
+    }
+
+    void EnergySaverStatus_Register()
+    {
+        check_hresult(PowerNotifications_RegisterEnergySaverStatusChangedListener(
```
dhoehna, line 22: what does check_hresult do?  Isn't that legacy CppWinRT stuff?  THROW_IF_FAILED would be better here and more expressive.

dev/PowerNotifications/PowerNotifications.cpp
```diff
@@ -0,0 +1,316 @@
...
+    void BatteryStatus_Unregister()
+    {
+        check_hresult(PowerNotifications_UnregisterCompositeBatteryStatusChangedListener(
+            make_self<factory_implementation::PowerManager>()->m_batteryStatusHandle));
+    }
+
+    void BatteryStatus_Update()
+    {
+        auto status = &make_self<factory_implementation::PowerManager>()->m_cachedCompositeBatteryStatus;
+        if (PowerNotifications_GetCompositeBatteryStatus(status) == S_OK)
```
dhoehna, line 61: No update if there is a failure?  Silently let it pass?

dev/PowerNotifications/PowerNotifications.cpp
```diff
@@ -0,0 +1,316 @@
...
+    }
+
+    void DisplayStatus_Update()
+    {
+        check_hresult(PowerNotifications_GetDisplayStatus(
+            &make_self<factory_implementation::PowerManager>()->m_cachedDisplayStatus));
+    }
+
+    // SystemIdleStatus Functions
+    EventType& SystemIdleStatus_Event()
```
dhoehna, line 185: I see many comments that say what the family of functions do.  Instead of using comments, put the implementations into different files.

dev/PowerNotifications/PowerNotifications.cpp
```diff
@@ -0,0 +1,316 @@
...
+            version, [](EFFECTIVE_POWER_MODE mode, PVOID)
+            {
+                PowerManager::EffectivePowerModeChanged_Callback(mode);
+            }, nullptr, &g_powerModeHandle));
+    }
+
+    void EffectivePowerMode_Unregister()
+    {
+        check_hresult(PowerUnregisterFromEffectivePowerModeNotifications(g_powerModeHandle));
+        g_powerModeHandle = NULL;
```
dhoehna, line 236: I'm getting code smell from the global.  Is a global truly needed?

dev/PowerNotifications/PowerNotifications.cpp
```diff
@@ -0,0 +1,316 @@
...
+            &make_self<factory_implementation::PowerManager>()->m_cachedUserPresenceStatus));
+    }
+
+    // SystemSuspendStatus Functions
+    EventType& SystemSuspendStatus_Event()
+    {
+        return make_self<factory_implementation::PowerManager>()->m_systemSuspendStatusChangedEvent;
+    }
+
+    ULONG SuspendResumeCallback(PVOID, ULONG powerEvent, PVOID)
```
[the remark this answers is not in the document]
dhoehna: OMG.  Is that the reason why  people do that?

test/PowerNotifications/PowerTests.cpp
DrusTheAxe: Possibly

General rule of thumb, you want RunAs:RestrictedUser as that ensures the test runs as a 'normal' user (not elevated etc) regardless who invoked TE.EXE. If you have concerns about normal vs elevated invocation (or other user contexts) then you'll want to specify multiple RunAs, or multiple tests with different fixtures if need be.

I wouldn't say it's _required_ for tests to run at varying IL/AC. That's something each feature team needs to judge. But if there are differences or notable scenarios then yes, it'd be wise to do so.

Also remember, this is Unit and Functional Tests. You also have opportunity to run Integration Tests at the top of the stack using the end NuGet+MSIXs. Some tests may be better suited for one vs the other. Ideally you do as much as you can here  so problems are detected early but YMMV.

dev/PowerNotifications/PowerNotifications.cpp
```diff
@@ -0,0 +1,316 @@
...
+#include <PowerManager.g.cpp>
+#include <powrprof.h>
+
+namespace winrt::Microsoft::ProjectReunion::implementation
+{
+
+    // EnergySaverStatus Functions
+    EventType& EnergySaverStatus_Event()
+    {
+        return make_self<factory_implementation::PowerManager>()->m_energySaverStatusChangedEvent;
```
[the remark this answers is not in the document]
huzaifa-d: No, I need the reference. It is used later in the code. And since it's all static the same instance is always returned.

dev/PowerNotifications/PowerNotifications.cpp
```diff
@@ -0,0 +1,316 @@
...
+#include <PowerNotifications.h>
+#include <frameworkudk\PowerNotificationsPal.h>
+#include <PowerManager.g.cpp>
+#include <powrprof.h>
+
+namespace winrt::Microsoft::ProjectReunion::implementation
+{
+
+    // EnergySaverStatus Functions
+    EventType& EnergySaverStatus_Event()
```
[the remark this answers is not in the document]
huzaifa-d: I can't use a const due to mismatches in later code.

Also is there an advantage of using a getter?

dev/PowerNotifications/PowerNotifications.cpp
```diff
@@ -0,0 +1,316 @@
...
+    {
+        check_hresult(PowerNotifications_UnregisterSystemIdleStatusChangedListener(
+            make_self<factory_implementation::PowerManager>()->m_systemIdleStatusHandle));
+    }
+
+    ULONG GetEffectivePowerModeVersion()
+    {
+        // We check for the version supported by checking if the feature is supported
+        // Using NULL as an indicator for uninitialized cache value
+        if (g_powerModeVersion == NULL)
```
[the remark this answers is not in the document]
huzaifa-d: I have explicitly set is as NULL and this is why I kept this check this way. 
Is there a guidance for this BTW?

dev/PowerNotifications/PowerNotifications.cpp
```diff
@@ -0,0 +1,316 @@
...
+        {
+            PVOID handle;
```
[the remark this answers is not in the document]
huzaifa-d: Good to know. :)

dev/PowerNotifications/PowerNotifications.cpp
```diff
@@ -0,0 +1,316 @@
...
+    {
+        // Effectively the Get() for PowerMode
+        // Needs to get a temporary subscription to get most recent value
+
+        struct notify_callback {
+            EFFECTIVE_POWER_MODE mode;
+            wil::slim_event done;
+        } context;
+
+        PVOID handle;
```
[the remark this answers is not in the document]
huzaifa-d: The name is such on purpose because it is only used in the next 3-4 lines and the context is simple.
Any guidance for this?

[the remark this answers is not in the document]
huzaifa-d: It's actually used as soon as the other variables are declared.

dev/PowerNotifications/PowerNotifications.idl
```diff
@@ -0,0 +1,109 @@
...
+        Dimmed
+    };
+
+    enum EffectivePowerMode
+    {
+        BatterySaver,
+        BetterBattery,
+        Balanced,
+        HighPerformance,
+        MaxPerformance, // v1 last supported
```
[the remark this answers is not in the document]
huzaifa-d: Good idea

test/PowerNotifications/PowerTests.cpp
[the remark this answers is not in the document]
huzaifa-d: Changed. But chrono would need another level of conversion.

test/PowerNotifications/PowerTests.cpp
```diff
@@ -0,0 +1,205 @@
...
+            TEST_CLASS_PROPERTY(L"ThreadingModel", L"MTA")
+            END_TEST_CLASS()
```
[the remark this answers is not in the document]
huzaifa-d: Yeah. Mixed multiple commits. Fixed.

test/PowerNotifications/PowerTests.cpp
huzaifa-d: Not really. The comment is just to guide for systems that can run it.

test/PowerNotifications/PowerTests.cpp
huzaifa-d: TIL. Changed.

test/PowerNotifications/PowerTests.dll.manifest
huzaifa-d: Working on this.

dev/PowerNotifications/PowerNotifications.cpp
huzaifa-d: Not sure about legacy but THROW_IF_FAILED  is better; agreed.

dev/PowerNotifications/PowerNotifications.cpp
huzaifa-d: Yes, this is in cases where there is no battery on the system.

dev/PowerNotifications/PowerNotifications.cpp
huzaifa-d: Changed to member.

dev/PowerNotifications/PowerNotifications.cpp
huzaifa-d: Putting in different files would break the header which is actually a problem when you have a factory_implementation. It results in more issues than is worthwhile to fix; have tried it. :(

test/PowerNotifications/PowerTests.cpp
huzaifa-d: TBH I just choose a value where I was sure that it will succeed. Informal testing shows they take <500 ms and the Power team also didn't have data on this. Do you have any suggestions?

huzaifa-d:
I wanted to push after the renaming to AppSDK but wasn't sure how long that would take and wanted this to be updated from my side.

dev/PowerNotifications/PowerNotifications.h
```diff
@@ -0,0 +1,699 @@
...
+                m_systemSuspendStatus = SystemSuspendStatus::Uninitialized;
+            }
+
+        };
+    }
+
+    namespace implementation
+    {
+        // EffectivePowerMode variables
+        static std::atomic<ULONG> g_powerModeVersion = NULL;
```
DrusTheAxe, line 598: Is `= NULL` necessary? Can this just be `static std::atomic<ULONG> g_powerModeVersion;` ?

dev/PowerNotifications/PowerNotifications.h
```diff
@@ -0,0 +1,699 @@
...
+        static PVOID g_powerModeHandle;
```
DrusTheAxe, line 599: `static void* g_powerModeHandle{};`

dev/PowerNotifications/PowerNotifications.cpp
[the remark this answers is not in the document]
DrusTheAxe: It's preferable. For one, it's the common modern C++ style notation. For another, it's less puzzling. This method ends with `return g_powerModeVersion;` but the return type is `ULONG` so `NULL` was weird. I had to hunt up the declaration and see the attribute's a `std::atomic<ULONG>` to make sense. So using `!foo` would have been clearer / less puzzling.

Plus explicitly initializing it as NULL seems unnecessary. See other comment.

dev/PowerNotifications/PowerNotifications.cpp
DrusTheAxe: Does `PowerNotifications_GetCompositeBatteryStatus()` return any SUCCEEDED() HRESULT other than S_OK?

Some old APIs can (e.g. MSXML.loadXML() can return S_OK or S_FALSE as different flavors of success), but that's been out of vogue for a long time now. As long as `PowerNotifications_GetCompositeBatteryStatus()` doesn't return varying values you need to distinguish you should do `if (SUCCEEDED(PowerNotifications_GetCompositeBatteryStatus(status)))`

(Any code explicitly checking ==/!=S_OK immediately flags suspicion as possibly flawed)

dev/PowerNotifications/PowerNotifications.cpp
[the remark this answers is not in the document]
DrusTheAxe: I thought this was a method of the object. If it's a free floating function then ignore my comment it's not applicable.

If it's a method of an object then declaring it as a getter gives more natural property-like syntax (reasonable for simple property-like things). If it's a free-floating function then making the method const is N/A since methods of objects can be const but non-object functions can't.

test/PowerNotifications/PowerTests.cpp
dhoehna: NIT: c_TimeoutInMs

dhoehna: Or, change this to a std::chrono::duration.  Makes it more explicit with what you mean.

dev/PowerNotifications/PowerNotifications.h
huzaifa-d: Fixed.

test/PowerNotifications/PowerTests.cpp
huzaifa-d: The variable name has been changed. I didn't use chrono as it would need another level of conversion and the variable name makes it clear. 
Will take suggestions on a possible better time out value though. :)

dev/PowerNotifications/PowerNotifications.h
huzaifa-d: Changed.

dev/PowerNotifications/PowerNotifications.cpp
huzaifa-d: Changed

dhoehna: This is still being compared to S_OK

huzaifa-d: Github is confusing; you are referring to old code.

And if there was a battery a filled compositebattery struct would be returned and we would calculate the rest based on the values in it.

DrusTheAxe (approved):
Please flip warnings to level4 (in this PR or a follow up) and correct anything it flags

huzaifa-d:
The actual PR has only Level4. Did I miss something?
https://github.com/microsoft/WindowsAppSDK/pull/1142

Thanks,
H.

DrusTheAxe:
My comment predates the Level4 change (or maybe I was looking at an older commit in Codeflow when I made that comment?)

But all good now :-)

huzaifa-d:
Your comment was on an old PR actually. I had to make a new one after the AppSDK rename. I should have clarified earlier. Apologies.

Thanks,
H.

DrusTheAxe (approved):
Some minor comments but otherwise looks good

DrusTheAxe:
Is this PR current or obsolete?

huzaifa-d:
It is old. Apologies for the confusion. Closing.

PR #1060, opened 2021-07-14
fix path for unpackagedtests

huichen123:
this is a regression from https://github.com/microsoft/WindowsAppSDK/pull/938.


rohanp-msft (approved):
:shipit:

axelandrejs (approved):

PR #1064, opened 2021-07-15
Add unpackaged support to PushNotifications spec

pmpurifoy:


specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -196,8 +198,9 @@ Process A (Registration of the Push Trigger only):
 ```cpp
 int main()
 {
+    PushNotificationRegistrationOptions pushOptions = PushNotificationManager::GetPushActivatorKind(false);
```
sharath2727, line 201: GetPushActivatorOptions(false);??

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -212,27 +215,27 @@ Process B (Register the inproc COM server and handle the background activation):
 ```cpp
 int main()
 {
+    PushNotificationRegistrationOptions pushOptions = PushNotificationManager::GetPushActivatorKind(true);
```
sharath2727, line 218: GetPushActivatorOptions(true);??

specs/PushNotifications/PushNotifications-spec.md
pmpurifoy: This API call will be the equivalent of IsSupported, which will return flags for the developer to call into RegisterActivator. Packaged apps should set this isInProcActivation parameter to true if the calling process will handle background activation.

pmpurifoy:
Because packaged applications will use ComActivator or ProtocolActivator depending on the build (builds earlier than 19041), they should check the flag that was used to register and unregister accordingly.
- ComActivator = Pass the token and ComActivator flag
- ProtocolActivator = Pass nullptr token and ProtocolActivator flag if they want to remove notifications from the application completely

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -185,7 +187,7 @@ int main()
     }
 
     // Unregisters the inproc COM Activator before exiting
-    PushNotificationManager::UnregisterActivator(token, PushNotificationRegistrationKind::ComActivator);
+    PushNotificationManager::UnregisterActivator(token, pushOptions);
```
danielayala94, line 190: If you pass the flags as is and it is ProtocolActivator, we prevent the app to get notifications, which we don't want.

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -88,34 +88,36 @@ Power Management compliance.
 ```cpp
 int main()
 {
+    PushNotificationRegistrationOptions pushOptions = PushNotificationManager::GetPushActivatorOptions(true);
```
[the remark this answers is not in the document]
danielayala94: This way, there is no need to look for GetPushActivatorOptions for Process B. We have the confidence to pass a COM flag on Process B.

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -258,7 +261,135 @@ sealed partial class App : Application
...
+        auto result = channelOperation.GetResults();
+        if (result.Status() == PushNotificationChannelStatus::CompletedSuccess)
+        {
+            result.Channel().PushReceived(pushToken);
+        }
+    }
+
+    // To unregister Push Notifications
+    PushNotificationManager::UnregisterActivator(nullptr, pushOptions);
+    MddBootstrapShutdown();
```
danielayala94, line 389: What is the purpose of this call?

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -258,7 +261,135 @@ sealed partial class App : Application
...
+        // Unregister the Push event for Foreground before exiting
+        auto result = channelOperation.GetResults();
+        if (result.Status() == PushNotificationChannelStatus::CompletedSuccess)
+        {
+            result.Channel().PushReceived(pushToken);
+        }
+    }
+
+    // To unregister Push Notifications
+    PushNotificationManager::UnregisterActivator(nullptr, pushOptions);
```
danielayala94, line 388: Only if the developer doesn't want to receive notifications anymore, they should call this.

If the developer is expecting to launch an app after receiving a notification, this call should be removed.

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -459,12 +592,21 @@ namespace Microsoft.Windows.PushNotifications
         // Unregister any activator if present using a token and registrationKind
         // 1) If kind = PushTrigger is specified, the trigger itself will be removed
         // 2) If kind = ComActivator is specified, the Project Reunion Background Task component will no longer act as an InProc COM Server
-        static void UnregisterActivator(PushNotificationRegistrationToken token, PushNotificationRegistrationKind kind);
+        // 3) If kind = ProtocolActivator is specified, the application will be unregistered from the PushNotificationsLongRunningTask
+        static void UnregisterActivator(PushNotificationRegistrationToken token, PushNotificationRegistrationOptions options);
```
danielayala94, line 596: If we don't need a token for unpackaged, probably we need an overload of this function.

Just my opinion, not sure what the opinion of Jon or Howard would be on this case.

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -459,12 +592,21 @@ namespace Microsoft.Windows.PushNotifications
...
 
         // Request a Push Channel with an encoded RemoteId from WNS. RemoteId is an AAD identifier GUID
         static Windows.Foundation.IAsyncOperationWithProgress<PushNotificationCreateChannelResult, PushNotificationCreateChannelStatus> CreateChannelAsync(Guid remoteId);
+
+        // Developers should call this to know with which PushNotificationRegistrationKind to call RegisterActivator. 
+        // Packaged applications should set the isInProcActivation to true if they want the current process to handle background activation. 
+        // Unpackaged applications should always set the isInProcActivation to false.
```
danielayala94, line 603: Why? Out-of-proc is unsupported for unpackaged apps.

danielayala94 (changes requested):

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -258,7 +261,135 @@ sealed partial class App : Application
...
+receiving raw notifications.
+
+**NOTE:** Packaged applications running on Windows builds earlier than 19041 will need to use ProtocolActivator</br>
+due to SetTaskEntryPointClsid() being unavailable on those builds.
+
+```cpp
+int main()
+{
+    // Take a dependency on Windows App SDK preview 
+    const UINT32 majorMinorVersion{ 0x00000008 };
```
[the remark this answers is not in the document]
pmpurifoy: Yes

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -258,7 +261,135 @@ sealed partial class App : Application
...
+
+```cpp
+int main()
+{
+    // Take a dependency on Windows App SDK preview 
+    const UINT32 majorMinorVersion{ 0x00000008 }; 
+    PCWSTR versionTag{ L"preview" }; 
+    const PACKAGE_VERSION minVersion{};
+
+    const HRESULT hr{ MddBootstrapInitialize(majorMinorVersion, versionTag, minVersion) };
```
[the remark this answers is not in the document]
pmpurifoy: If it cannot find a DDLM to connect to it will fail.

specs/PushNotifications/PushNotifications-spec.md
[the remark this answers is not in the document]
danielayala94: Where the developer should look at to correctly set the majorMinorVersion?

specs/PushNotifications/PushNotifications-spec.md
hulumane, line 91: IDL changes will be needed in the doc. Also please make sure the idl compiles when you change it in the doc.

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -258,7 +261,135 @@ sealed partial class App : Application
...
+    const HRESULT hr{ MddBootstrapInitialize(majorMinorVersion, versionTag, minVersion) }; 
+    // Check the return code. If there is a failure, display the result.
+    if (FAILED(hr)) 
+    { 
+        wprintf(L"Error 0x%X in MddBootstrapInitialize(0x%08X, %s, %hu.%hu.%hu.%hu)\n", 
+            hr, majorMinorVersion, versionTag, minVersion.Major, minVersion.Minor, minVersion.Build, minVersion.Revision); 
+        return hr; 
+    }
+
+    PushNotificationRegistrationOptions pushOptions = PushNotificationManager::GetPushActivatorOptions(true);
```
hulumane, line 289: Might want to remove that version and bootstrap stuff in the top for this sample.

specs/PushNotifications/PushNotifications-spec.md
hulumane: Correct. UnregisterActivator simple stops PushNotifications. Probably a bad idea to do this in reality.

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -258,7 +261,135 @@ sealed partial class App : Application
...
+
+    // Unpackaged applications do not need guid to register activator
+    PushNotificationActivationInfo info(
```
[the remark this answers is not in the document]
hulumane: Agreed

specs/PushNotifications/PushNotifications-spec.md
hulumane, line 289: For Unpackaged apps, what does it mean to set inproc to false? Does that mean that we launch a new process when the app is running?

specs/PushNotifications/PushNotifications-spec.md
hulumane: All this bootstrap stuff should be removed from the sample code right?

specs/PushNotifications/PushNotifications-spec.md
hulumane: Unless this is critical for the workings of our API in some way?

hulumane: It would be nice to have that support. @danielayala94 How could we support out-of-proc activation in your COM design?

specs/PushNotifications/PushNotifications-spec.md
hulumane: Unless it is critical to the workings of our API in some indirect way?

specs/PushNotifications/PushNotifications-spec.md
hulumane: Hmm can we support it for unpackaged?

specs/PushNotifications/PushNotifications-spec.md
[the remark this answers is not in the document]
pmpurifoy: We should also add overload functions for Register/UnregisterActivator. Good idea!

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -258,7 +261,135 @@ sealed partial class App : Application
...
+        pushOptions,
+        nullptr);
+
+    // Registers the unpackaged applicaiton with the PushNotificationsLongRunningTask
```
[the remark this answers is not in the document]
pmpurifoy: Probably don't have to have to mention the full impl of unpkg registration. In the appendix it mentions this component and I thought it would be helpful to have a brief comment mentioning where the LRP is being used.

specs/PushNotifications/PushNotifications-spec.md
pmpurifoy: There needs to be a check if we returned Protocol/ComActivator from GetPushActivatorFlags. And unregister accordingly.

specs/PushNotifications/PushNotifications-spec.md
pmpurifoy: When an unpkg application exits, it removes its reference from DDLM. So that PRfwk can be serviced when all unpkg references are released.

specs/PushNotifications/PushNotifications-spec.md
pmpurifoy: Right. The comment should probably be, "To completely remove notifications from the application" call this.

specs/PushNotifications/PushNotifications-spec.md
pmpurifoy: All unpkg apps must call MddBootstrapInitialize/MddBootstrapShutdown so I figured our sample code should follow the WindowsAppSDK guidelines.

When an unpkg application exits, it removes its reference from DDLM. So that PRfwk can be serviced when all unpkg references are released.

pmpurifoy: Also it adds the PRfwk to the packaged graph of the unpkg app which gives them access to our apis.

hulumane: Gotcha...might want to add a comment next to that to say this is critical code for packaged apps to function with the framework

specs/PushNotifications/PushNotifications-spec.md
hulumane: The problem here is that the Unregister pattern is actually needed to remove the COM registeration. It doesn't mean that Push will stop working in any way. It's kinda like an anti-pattern. Do you have any ideas as to how we can mitigate this?

specs/PushNotifications/PushNotifications-spec.md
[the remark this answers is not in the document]
pmpurifoy: I believe they have to get the version themselves. They will run ProjectReunionInstaller and use that installed version

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -84,38 +84,47 @@ Power Management compliance.
...
+call into the bootstrapper API to have access to the Windows App SDK. Click [here](https://docs.microsoft.com/en-us/windows/apps/windows-app-sdk/tutorial-unpackaged-deployment?tabs=cpp) for information on how to build and 
+deploy unpackaged applications with Windows App SDK.
 ```cpp
 int main()
 {
-    // Register the COM Activator GUID
-    PushNotificationActivationInfo info(
-        PushNotificationRegistrationKind::PushTrigger | PushNotificationRegistrationKind::ComActivator,
+    PushNotificationActivationInfo info = nullptr;
+    if(PushNotificationManager::IsActivatorSupported(PushNotificationRegistrationOptions::PushTrigger | PushNotificationRegistrationOptions::ComActivator))
```
hulumane, line 94: Set this to a local bool and use it everywhere below.
For example,
bool isComActivationSupported = PushNotificationManager::IsActivatorSupported(PushNotificationRegistrationOptions::PushTrigger | PushNotificationRegistrationOptions::ComActivator);

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -359,24 +387,28 @@ namespace Microsoft.Windows.PushNotifications
     };
 
     [flags]
-    enum PushNotificationRegistrationKind
+    enum PushNotificationRegistrationOptions
```
hulumane, line 390: No need to rename this. Lets keep as is

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -456,15 +488,24 @@ namespace Microsoft.Windows.PushNotifications
...
-        // Unregister any activator if present using a token and registrationKind
+        // Unregister any activator if present using a token and PushNotificationRegistrationOptions
         // 1) If kind = PushTrigger is specified, the trigger itself will be removed
-        // 2) If kind = ComActivator is specified, the Windows App SDK Background Task component will no longer act as an InProc COM Server
-        static void UnregisterActivator(PushNotificationRegistrationToken token, PushNotificationRegistrationKind kind);
+        // 2) If kind = ComActivator is specified, the Project Reunion Background Task component will no longer act as an InProc COM Server
+        static void UnregisterActivator(PushNotificationRegistrationToken token, PushNotificationRegistrationOptions options);
+
+        // Unregister the protocol activator registered for the calling application
+        static void UnregisterProtocolActivator();
```
hulumane, line 497: Please remove. Just use the UnregisterActivator function above

specs/PushNotifications/PushNotifications-spec.md
pmpurifoy: In the PushNotifications PR, Raymond asked us to change this to *Options so I was just updated the spec with what is in main.

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -456,15 +489,22 @@ namespace Microsoft.Windows.PushNotifications
         // Register an activator using an ActivationInfo context and return a RegistrationToken
         static PushNotificationRegistrationToken RegisterActivator(PushNotificationActivationInfo details);
 
-        // Unregister any activator if present using a token and registrationKind
+        // Unregister any activator if present using a token and PushNotificationRegistrationOptions
         // 1) If kind = PushTrigger is specified, the trigger itself will be removed
-        // 2) If kind = ComActivator is specified, the Windows App SDK Background Task component will no longer act as an InProc COM Server
-        static void UnregisterActivator(PushNotificationRegistrationToken token, PushNotificationRegistrationKind kind);
+        // 2) If kind = ComActivator is specified, the Project Reunion Background Task component will no longer act as an InProc COM Server
+        // 3) If kind = ProtocolActivator is specified, the application will be unregistered from the long running process that handles activation
```
sharath2727, line 495: @pmpurifoy should we also explicitly mention token to be nullptr for ProtocolActivator?

pmpurifoy:
Need to change ProtocolActivator value from 0x3 -> 0x4.

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -84,38 +84,48 @@ Power Management compliance.
...
 
-    // Registers a Push Trigger and sets up an inproc COM Server for Activations
+    // Registers the application to receive push notifications
     auto token = PushNotificationManager::RegisterActivator(info);
 
     // Check to see if the WinMain activation is due to a Push Activator
-    // Note: This API is currently not present in Windows App SDK but will be included as part of the AppLifecycle Public API spec.
-    auto activation = AppLifecycle::Activation();
-    auto kind = activation.Kind();
+    auto args = AppInstance::GetCurrent().GetActivatedEventArgs();
```
hulumane, line 110: Thanks for updating this!

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -184,21 +194,34 @@ int main()
...
+    PushNotificationActivationInfo info = nullptr;
+    if(PushNotificationManager::IsActivatorSupported(PushNotificationRegistrationOptions::PushTrigger))
+    {
+        // Register the ComActivator guid
+        info = PushNotificationActivationInfo(PushNotificationRegistrationOptions::PushTrigger,
         winrt::guid("BACCFA91-F1DE-4CA2-B80E-90BE66934EC6"));
+    }
+    else
+    {
+        winrt::throw_hresult(HRESULT_FROM_WIN32(ERROR_NOT_SUPPORTED));
```
hulumane, line 223: Should we just return an error code here rather than throw? int == error code

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -212,27 +235,34 @@ Process B (Register the inproc COM server and handle the background activation):
...
+    PushNotificationActivationInfo info = nullptr;
+    if(PushNotificationManager::IsActivatorSupported(PushNotificationRegistrationOptions::ComActivator))
+    {
+        // Register the ComActivator guid
+        info = PushNotificationActivationInfo(PushNotificationRegistrationOptions::ComActivator,
         winrt::guid("BACCFA91-F1DE-4CA2-B80E-90BE66934EC6"));
+    }
+    else
+    {
+        winrt::throw_hresult(HRESULT_FROM_WIN32(ERROR_NOT_SUPPORTED));
```
hulumane, line 247: Same here.  Just return error code?

pmpurifoy: Yup. We should return error code.

specs/PushNotifications/PushNotifications-spec.md
hulumane: Ah got it. Thanks for the context

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -84,38 +84,48 @@ Power Management compliance.
   * It will subscribe to a In-memory Push event handler hanging off the retrieved Push Channel component.
     * The app should ideally handle the event by setting the Handled property to true to prevent Background Activation from <br/>
     launching a new process.
-
+* If the application is unpackaged, it will use the ProtocolActivator to register the activator instead. The app will also need to 
+call into the bootstrapper API to have access to the Windows App SDK. Click [here](https://docs.microsoft.com/en-us/windows/apps/windows-app-sdk/tutorial-unpackaged-deployment?tabs=cpp) for information on how to build and
```
jonwis, line 88: Please word-wrap. Use VSCode's prettier extension which will pull in the .editorconfig for consistency.

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -84,38 +84,48 @@ Power Management compliance.
   * It will subscribe to a In-memory Push event handler hanging off the retrieved Push Channel component.
     * The app should ideally handle the event by setting the Handled property to true to prevent Background Activation from <br/>
     launching a new process.
-
+* If the application is unpackaged, it will use the ProtocolActivator to register the activator instead. The app will also need to 
+call into the bootstrapper API to have access to the Windows App SDK. Click [here](https://docs.microsoft.com/en-us/windows/apps/windows-app-sdk/tutorial-unpackaged-deployment?tabs=cpp) for information on how to build and 
+deploy unpackaged applications with Windows App SDK.
```
jonwis, line 89: Needs a newline after "." and before "```"

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -84,38 +84,48 @@ Power Management compliance.
...
 {
-    // Register the COM Activator GUID
-    PushNotificationActivationInfo info(
-        PushNotificationRegistrationKind::PushTrigger | PushNotificationRegistrationKind::ComActivator,
+    PushNotificationActivationInfo info = nullptr;
+    bool isComActivationSupported = PushNotificationManager::IsActivatorSupported(PushNotificationRegistrationOptions::PushTrigger | PushNotificationRegistrationOptions::ComActivator);
+    if(isComActivationSupported)
+    {
+        // Register the ComActivator guid
+        info = PushNotificationActivationInfo(PushNotificationRegistrationOptions::PushTrigger | PushNotificationRegistrationOptions::ComActivator,
```
jonwis, line 98: Is there another model instead here, like:

```
auto info = PushNotificationActivationInfo(PushNotificationRegistrationOptions::Default);
```

... where "default" triggers this "is supported" thing, but _inside_ the activation constructor? Or is there a clear reason to make the app developer have this in their application?

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -84,38 +84,48 @@ Power Management compliance.
...
-    auto activation = AppLifecycle::Activation();
-    auto kind = activation.Kind();
+    auto args = AppInstance::GetCurrent().GetActivatedEventArgs();
+    auto kind = args.Kind();
 
-    if (kind == WindowsAppSDKActivationKind::Push)
+    if (kind == ExtendedActivationKind::Push)
     {
-        auto args = activation.Arguments().as<PushNotificationReceivedEventArgs>();
+        PushNotificationReceivedEventArgs pushArgs = args.Data().as<PushNotificationReceivedEventArgs>();
```
jonwis, line 115: Not necessary; the type is already stated in the trailing `.as<T>()`

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -184,21 +194,34 @@ int main()
         }
     }
 
-    // Unregisters the inproc COM Activator before exiting
-    PushNotificationManager::UnregisterActivator(token, PushNotificationRegistrationKind::ComActivator);
+    if(isComActivationSupported)
+    {
+        // Unregister the ComActivator from inproc activation
+        PushNotificationManager::UnregisterActivator(token, PushNotificationRegistrationOptions::ComActivator);
```
jonwis, line 200: Maybe this could be `PushNotificationManager::UnregisterActivator(token)` because the registration step remembers whether it was registered for Com or launch or...

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -184,21 +194,34 @@ int main()
...
 int main()
 {
-    PushNotificationActivationInfo info(
-        PushNotificationRegistrationKind::PushTrigger,
+    PushNotificationActivationInfo info = nullptr;
+    if(PushNotificationManager::IsActivatorSupported(PushNotificationRegistrationOptions::PushTrigger))
+    {
+        // Register the ComActivator guid
+        info = PushNotificationActivationInfo(PushNotificationRegistrationOptions::PushTrigger,
         winrt::guid("BACCFA91-F1DE-4CA2-B80E-90BE66934EC6"));
```
jonwis, line 219: Add indent

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -184,21 +194,34 @@ int main()
...
+    }
 
     return 0;
 }
 ```
 
 ## In this scenario, the process that Registers the Push Trigger and the process specified as the COM server are different.
+This only works for packaged applications that support PushTrigger and ComActivator flags in RegisterActivator. Out-of-proc activation is
+not supported for unpackaged apps.
 Process A (Registration of the Push Trigger only):
```
jonwis, line 210: Space needed between paragraphs, so this should probably be:

```
This not only works ... for unpackaged apps.

Process A ... :

``cpp
```

jonwis (changes requested):

specs/PushNotifications/PushNotifications-spec.md
pmpurifoy: We have to let developer decide which flag they want to unregister. They should call UnregisterActivator with PushTrigger if they no longer want to receive PushNotifications. They should call UnregisterActivator with ComActivator to unregister the calling app from inproc activation.

A common pattern will be RegisterActivator(ComActivator | PushTrigger) -> UnregisterActivator(ComActivator).

specs/PushNotifications/PushNotifications-spec.md
pmpurifoy: Because we support out-of-proc activation for packaged apps, they are allowed to register only a PushTrigger for process A and ComActivator for process B. I think default would be confusing for this case. Also, making the developer know exactly which flag is being used for RegisterActivator will allow them to know which flag to Unregister with.

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -84,38 +84,49 @@ Power Management compliance.
...
+deploy unpackaged applications with Windows App SDK.
 
 ```cpp
 int main()
 {
-    // Register the COM Activator GUID
-    PushNotificationActivationInfo info(
-        PushNotificationRegistrationKind::PushTrigger | PushNotificationRegistrationKind::ComActivator,
+    PushNotificationActivationInfo info = nullptr;
+    bool isComActivationSupported = PushNotificationManager::IsActivatorSupported(PushNotificationRegistrationOptions::PushTrigger | PushNotificationRegistrationOptions::ComActivator);
```
danielayala94, line 95: nit: maybe some extra lines for clarity?

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -84,38 +84,49 @@ Power Management compliance.
...
+    if(isComActivationSupported)
```
danielayala94, line 96: nit: `if (isComActivationSupported)`

specs/PushNotifications/PushNotifications-spec.md
danielayala94: Same for the other code snippets.

specs/PushNotifications/PushNotifications-spec.md
danielayala94: Same for the other code snippets.

danielayala94 (approved):

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -2,120 +2,150 @@
...
 
 ## The problems today
 
-**Portal Registration**: The entire Registration flow for Push is dependent on registering an app in the <br/>
-partner center portal by [associating an app with the Windows Store](https://docs.microsoft.com/azure/notification-hubs/notification-hubs-windows-store-dotnet-get-started-wns-push-notification). As a result, non-store provisioned <br/>
-apps cannot use this feature.
+**Portal Registration**: The entire Registration flow for Push is dependent on registering an app in
+the <br/> partner center portal by
+[associating an app with the Windows Store](https://docs.microsoft.com/azure/notification-hubs/notification-hubs-windows-store-dotnet-get-started-wns-push-notification).
+As a result, non-store provisioned <br/> apps cannot use this feature.
```
jonwis, line 25: Can you make a pass and see what's going on with all these `<br/>` in the document? Markdown automatically generates appropriate linebreaks. If you want a new paragraph, just use a blank like:

```
instead of <br/> this style
```

```
use this

style.
```

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -2,120 +2,150 @@
...
-Power Management compliance.
-* If the Activation is Foreground, the app will do 2 things:
-  * It will request a Push Channel Asynchronously with an implementation of Progress and Completed event handler.
-    * Expect Channel operations to take around 2-16 minutes in some rare cases (retryable errors). In the 90th percentile case, <br/>
-    it should be fairly quick operation (in a few seconds).
-  * It will subscribe to a In-memory Push event handler hanging off the retrieved Push Channel component.
-    * The app should ideally handle the event by setting the Handled property to true to prevent Background Activation from <br/>
-    launching a new process.
+
+-   The Registration will take in a CLSID of the COM server as part of the Activator setup. The
```
jonwis, line 83: (No change required, note on writing style)

Instead of "the bar will foo" consider using "the bar foos".  "Will" suggests indeterminate future, while "bar foos" is a stronger statement that it must be so.  This document, once merged, should be read as "now" rather than "some future point." So someone who comes across this in the future doesn't need to wonder "have they already made this happen, can I use it now?" and instead says "ok, I need my bar to foo in order to accomplish this."

specs/PushNotifications/PushNotifications-spec.md
jonwis: I see. I'd assumed that a developer would register twice, once per activation kind, rather than or-ing together flags, which meant that they got unique tokens back from registration.

Is there an "unregister all my activators" in case the app can't remember the tokens it got?

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -359,24 +435,28 @@ namespace Microsoft.Windows.PushNotifications
     };
 
     [flags]
-    enum PushNotificationRegistrationKind
+    enum PushNotificationRegistrationOptions
```
jonwis, line 438: Consider a name like `PushNotificationActivators` or similar. It's no longer general options about registering activation, it's "which kind of activator does a push notification for this registration use."

jonwis (changes requested):

specs/PushNotifications/PushNotifications-spec.md
pmpurifoy: We do not have an "unregister all my activators" function. Should we consider adding an API like this?

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -2,120 +2,148 @@
...
 
-At a high level, we need to provide a way for all Win32 applications to subscribe to and receive Push notifications <br/>
-irrespective of their app type. This includes unpackaged apps and packaged win32 (MSIX Desktop Bridge, Sparse Signed Packages). <br/>
-Moreover, all Push scenarios should adhere to OS resource management policies like Power Saver, Network Attribution, <br/>
-Enterprise group policies etc. The Windows App SDK Push component will abstract away the complexities of dealing with WNS <br/>
-channel registrations and Push related activations as much as possible freeing the developer to focus on other app <br/>
-related challenges.
+At a high level, we need to provide a way for all Win32 applications to subscribe to and receive
+Push notifications irrespective of their app type. This includes unpackaged apps and packaged win32
+(MSIX Desktop Bridge, Sparse Signed Packages). Moreover, all Push scenarios should adhere to OS
```
DrusTheAxe, line 48: You let out Win32App

...(MSIX Desktop Bridge, MSIX Win32App, Sparse Signed Packages)...

Win32App aka RuntimeBehavior=Win32App is like Desktop Bridge except instead of virtualized I/O and the like goodness, Win32App processes strongly favor (classic) Win32 compability. No Helium container, Loader has the classic 18+ mechanisms in the DLL Search Order vs 8, among other things.

P.S. Sparse Signed Packages isn't really another 'app type', more like a way of unpackaged apps opting in to MSIX goodness for runtime without replacing their installer with MSIX's. But sure, can't hurt to mention it here. All good.

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -2,120 +2,148 @@
...
-* Supporting raw push for Packaged Win32(MSIX Desktop Bridge)
-* Supporting raw push for Unpackaged Win32.
-* Supporting Visual Cloud Toasts for packaged Win32 apps.
-* Supporting Visual Cloud Toasts for unpackaged Win32 apps.
-
-A Portal Registration flow through [AAD (Azure Active Directory)](https://docs.microsoft.com/azure/active-directory/develop/quickstart-register-app) will also be defined that removes the dependency of the Push Flow with <br/>
-the Partner Center Portal. The RemoteIdentifier GUID in this spec maps to the ApplicationId(ClientId) in the AAD App <br/>
-Registration process. Below  is an illustration of the proposed flow through AAD:<br/>
+
+-   Supporting raw push for Packaged Win32(MSIX Desktop Bridge)
```
DrusTheAxe, line 56: NIT: space

...for Packaged Win32 (MSIX Desktop Bridge)...

DrusTheAxe, line 56: What about MSIX RuntimeBehavior=Win32app?

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -2,120 +2,148 @@
...
+    [here](https://docs.microsoft.com/en-us/windows/apps/windows-app-sdk/tutorial-unpackaged-deployment?tabs=cpp)
+    for information on how to build and deploy unpackaged applications with Windows App SDK.
 
 ```cpp
 int main()
 {
-    // Register the COM Activator GUID
-    PushNotificationActivationInfo info(
-        PushNotificationRegistrationKind::PushTrigger | PushNotificationRegistrationKind::ComActivator,
+    PushNotificationActivationInfo info = nullptr;
```
DrusTheAxe, line 110: NIT: `PushNotificationActivationInfo info{};`

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -2,120 +2,148 @@
...
+    bool isComActivationSupported = PushNotificationManager::IsActivatorSupported(PushNotificationRegistrationOptions::PushTrigger |
```
DrusTheAxe, line 111: NIT: `bool isComActivationSupported{ Push... };`

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -2,120 +2,148 @@
...
-    // Registers a Push Trigger and sets up an inproc COM Server for Activations
+    // Registers the application to receive push notifications
     auto token = PushNotificationManager::RegisterActivator(info);
 
     // Check to see if the WinMain activation is due to a Push Activator
-    // Note: This API is currently not present in Windows App SDK but will be included as part of the AppLifecycle Public API spec.
-    auto activation = AppLifecycle::Activation();
-    auto kind = activation.Kind();
+    auto args = AppInstance::GetCurrent().GetActivatedEventArgs();
+    auto kind = args.Kind();
```
DrusTheAxe, line 129: NIT: `auto kind{ args.Kind() };`

specs/PushNotifications/PushNotifications-spec.md
DrusTheAxe: I'd think so. Handy for apps that do conditional activations (varying options) can have 1 cleanup call `UnregisterActivatorAllKinds()` and not have to track which options they original passed in

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -208,43 +252,57 @@ int main()
...
+    PushNotificationActivationInfo info = nullptr;
+    if (PushNotificationManager::IsActivatorSupported(PushNotificationRegistrationOptions::ComActivator))
+    {
+        // Register the ComActivator guid
+        info = PushNotificationActivationInfo(PushNotificationRegistrationOptions::ComActivator,
+            winrt::guid("BACCFA91-F1DE-4CA2-B80E-90BE66934EC6"));
+    }
+    else
+    {
+        return ERROR_NOT_SUPPORTED;
```
DrusTheAxe, line 270: Why not E_NOTIMPL? (the typical HRESULT used instead of Win32's ERROR_NOT_SUPPORTED)

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -318,7 +388,10 @@ else if (result.Status() == ChannelStatus::CompletedFailure)
 ```
 
 ## Manifest Registration
-For MSIX, the COM activator GUID and the exe path need to be registered in the manifest. The launch args would need to be pre set to a well-known string that defines Push Triggers.
+
+For MSIX, the COM activator GUID and the exe path need to be registered in the manifest. The launch
+args would need to be pre set to a well-known string that defines Push Triggers.
```
DrusTheAxe, line 393: NIT: pre-set ?

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -318,7 +388,10 @@ else if (result.Status() == ChannelStatus::CompletedFailure)
 ```
 
 ## Manifest Registration
-For MSIX, the COM activator GUID and the exe path need to be registered in the manifest. The launch args would need to be pre set to a well-known string that defines Push Triggers.
+
+For MSIX, the COM activator GUID and the exe path need to be registered in the manifest. The launch
```
DrusTheAxe, line 392: NIT: ...and the exe need to be...

One specifies an exe in the manifest (as a relative path/filename) so you don't really specify the path in the manifest (some manifest features do let you specify an actual/full path, but I don't think this is one of them)

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -359,24 +433,28 @@ namespace Microsoft.Windows.PushNotifications
...
 
     // An abstraction over the activation Registration flow
     runtimeclass PushNotificationActivationInfo
     {
-        // Initialize using a RegistrationKind and optionally defined parameters like manifest defined activatorId
+        // Initialize using a Registration option and optionally defined parameters like manifest defined activatorId
         // 1) If kind = PushTrigger is specified, only the Push Trigger will be Registered with Background Infra
-        // 2) If kind = ComActivator is specified, the Windows App SDK Background Task component will be Registered as an InProc COM server
-        PushNotificationActivationInfo(PushNotificationRegistrationKind kind, Guid taskClsid);
+        // 2) If kind = ComActivator is specified, the Project Reunion Background Task component will be Registered as an InProc COM server
```
DrusTheAxe, line 448: the what Background Task component? ;-)

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -456,32 +534,44 @@ namespace Microsoft.Windows.PushNotifications
...
-unpackaged Win32 process from the long running service. The requirement is mainly because the native client </br>
-today does not have support to launch Win32 apps directly in response to a Push payload unless they are </br>
-already running in which case we simply marshall the payload back to the process via a registered callback sink.
-* For packaged applications like MSIX Desktop Bridge, we buid a thin wrapper around existing Background Manager </br>
-APIs rather than re-inventing new Background Triggers. This is mainly because Background Infrastructure native </br>
-client stack is built with Power Management and resourcing concerns already addressed. Building a new stack that </br>
-addresses similar concerns is likely not the best use of our time or resources. Instead we should invest in </br>
-Windows App SDK-izing the Background Infra stack itself.
+
+-   For all OS images before 21H1 (19043), the RemoteId will be a noop and the native platform will
```
DrusTheAxe, line 558: "For all Windows releases before..."

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -456,32 +534,44 @@ namespace Microsoft.Windows.PushNotifications
...
+    send the PFN (PackageFamilyName) in the channel request protocol. The WNS server will maintain
```
DrusTheAxe, line 559: "...send the PackageFamilyName in the channel..."

Remove PFN. Please don't use that acronym as its ambiguous (PackageFamilyName and PackageFullName have the same abbreviation, so we strongly discourage using 'PFN').

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -456,32 +534,44 @@ namespace Microsoft.Windows.PushNotifications
...
+    an internal mapping of PFN -> RemoteId and will return a channelUri with the encoded RemoteId
+    after querying the map.
+-   For OS images 21H1(19043) and beyond, the Windows App SDK component will call into newly added
```
DrusTheAxe, line 562: NIT: OS images -> Windows releases

DrusTheAxe (approved):
LGTM modulo comments left

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -2,120 +2,148 @@
...
-  * It will subscribe to a In-memory Push event handler hanging off the retrieved Push Channel component.
-    * The app should ideally handle the event by setting the Handled property to true to prevent Background Activation from <br/>
-    launching a new process.
+
+-   The Registration will take in a CLSID of the COM server as part of the Activator setup. The
+    registration API will simply be a thin wrapper around the BackgroundInfra WinRT APIs and
+    abstract away the COM details from the developer including registration of the inproc Windows
+    App SDK component.
+-   The app will query ApplifeCycle APIs to retrieve an ActivationKind. (Note: This is covered in a
+    seperate AppLifeCycle API spec.)
```
riverar, line 87:
```suggestion
-   The app will query AppLifecycle APIs to retrieve an ActivationKind. (Note: This is covered in a
    seperate AppLifecycle API spec.)
```

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -184,21 +212,37 @@ int main()
...
+    }
 
     return 0;
 }
 ```
 
 ## In this scenario, the process that Registers the Push Trigger and the process specified as the COM server are different.
+
+This only works for packaged applications that support PushTrigger and ComActivator flags in
+RegisterActivator. Out-of-proc activation is not supported for unpackaged apps.
```
riverar, line 228:
```suggestion
RegisterActivator. out-of-process activation is not supported for unpackaged apps.
```

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -332,6 +405,7 @@ For MSIX, the COM activator GUID and the exe path need to be registered in the m
 ```
 
 # API Details
+
 ```c# (but really MIDL3)
```
riverar, line 409: This is not valid syntax, let's not jam up the machines parsing this file.

```suggestion
```c#
```

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -359,24 +433,28 @@ namespace Microsoft.Windows.PushNotifications
     };
 
     [flags]
-    enum PushNotificationRegistrationKind
+    enum PushNotificationRegistrationOptions
     {
         PushTrigger = 0x1, // Registers a Push Trigger with Background Infrastructure
-        ComActivator = 0x2, // Registers the Windows App SDK Background Task component as an InProc COM server
+        ComActivator = 0x2, // Registers the Project Reunion Background Task component as an InProc COM server
```
riverar, line 439:
```suggestion
        ComActivator = 0x2, // Registers the Background Task component as an in-process COM server
```

specs/PushNotifications/PushNotifications-spec.md
riverar:
```suggestion
        // 2) If kind = ComActivator is specified, the Background Task component will be Registered as an in-process COM server
```

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -456,32 +534,44 @@ namespace Microsoft.Windows.PushNotifications
         // Register an activator using an ActivationInfo context and return a RegistrationToken
         static PushNotificationRegistrationToken RegisterActivator(PushNotificationActivationInfo details);
 
-        // Unregister any activator if present using a token and registrationKind
+        // Unregister any activator if present using a token and PushNotificationRegistrationOptions
         // 1) If kind = PushTrigger is specified, the trigger itself will be removed
-        // 2) If kind = ComActivator is specified, the Windows App SDK Background Task component will no longer act as an InProc COM Server
-        static void UnregisterActivator(PushNotificationRegistrationToken token, PushNotificationRegistrationKind kind);
+        // 2) If kind = ComActivator is specified, the Project Reunion Background Task component will no longer act as an InProc COM Server
```
riverar, line 539:
```suggestion
        // 2) If kind = ComActivator is specified, the Background Task component will no longer act as an InProc COM Server
```

specs/PushNotifications/PushNotifications-spec.md
riverar: These are officially documented as _versions_, not _releases_. But otherwise agree with the suggestion.

```suggestion
-   For all versions of Windows 10 prior to Version 21H1 (build 19043), the RemoteId is unused and the native platform will
```

specs/PushNotifications/PushNotifications-spec.md
riverar:
```suggestion
    send the Package Family Name in the channel request protocol. The WNS server will maintain
```

specs/PushNotifications/PushNotifications-spec.md
riverar: Would use _versions_, see earlier comments.

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -456,32 +534,44 @@ namespace Microsoft.Windows.PushNotifications
...
-client stack is built with Power Management and resourcing concerns already addressed. Building a new stack that </br>
-addresses similar concerns is likely not the best use of our time or resources. Instead we should invest in </br>
-Windows App SDK-izing the Background Infra stack itself.
+
+-   For all OS images before 21H1 (19043), the RemoteId will be a noop and the native platform will
+    send the PFN (PackageFamilyName) in the channel request protocol. The WNS server will maintain
+    an internal mapping of PFN -> RemoteId and will return a channelUri with the encoded RemoteId
+    after querying the map.
+-   For OS images 21H1(19043) and beyond, the Windows App SDK component will call into newly added
+    Closed source APIs in the OS to actually pass a RemoteId in the Channel Request operation.
```
riverar, line 563: Don't need to call out its open/closed state. For the purposes of this spec, it's an OS provided API. Anything beyond that is out of scope.

Same suggestion applies to other instances below.

```suggestion
    APIs in the OS to pass a RemoteId in the Channel Request operation.
```

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -456,32 +534,44 @@ namespace Microsoft.Windows.PushNotifications
...
+-   For unpackaged Win32 apps, the Windows App SDK component will call into Closed source APIs to
+    register the unpackaged Win32 process with the RemoteId.
+-   A long running Windows App SDK component will be needed to intercept Push payloads from the
+    native platform and Launch the corresponding unpackaged Win32 app. We will be using Protocol
+    Activation via WinMain to launch the unpackaged Win32 process from the long running service. The
+    requirement is mainly because the native client today does not have support to launch Win32 apps
+    directly in response to a Push payload unless they are already running in which case we simply
+    marshall the payload back to the process via a registered callback sink.
```
riverar, line 571:
```suggestion
    marshal the payload back to the process via a registered callback sink.
```

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -456,32 +534,44 @@ namespace Microsoft.Windows.PushNotifications
...
+-   For packaged applications like MSIX Desktop Bridge, we buid a thin wrapper around existing
```
riverar, line 572: MSIX Desktop Bridge isn't a packaged application. And we've already established the difference between unpackaged/packaged earlier, so it's just repetitive at this point.

```suggestion
-   For packaged applications, we build a thin wrapper around existing
```

riverar (commented):
Just some cleanup nits.

hulumane (approved):

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -2,120 +2,148 @@
...
-* If the Activation is Foreground, the app will do 2 things:
-  * It will request a Push Channel Asynchronously with an implementation of Progress and Completed event handler.
-    * Expect Channel operations to take around 2-16 minutes in some rare cases (retryable errors). In the 90th percentile case, <br/>
-    it should be fairly quick operation (in a few seconds).
-  * It will subscribe to a In-memory Push event handler hanging off the retrieved Push Channel component.
-    * The app should ideally handle the event by setting the Handled property to true to prevent Background Activation from <br/>
-    launching a new process.
+
+-   The Registration will take in a CLSID of the COM server as part of the Activator setup. The
+    registration API will simply be a thin wrapper around the BackgroundInfra WinRT APIs and
```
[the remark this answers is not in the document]
pmpurifoy: Removing.

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -359,24 +435,28 @@ namespace Microsoft.Windows.PushNotifications
     };
 
     [flags]
-    enum PushNotificationRegistrationKind
+    enum PushNotificationRegistrationOptions
     {
         PushTrigger = 0x1, // Registers a Push Trigger with Background Infrastructure
```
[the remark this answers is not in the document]
pmpurifoy: Yes.

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -2,120 +2,148 @@
...
+
+-   Supporting raw push for Packaged Win32(MSIX Desktop Bridge)
+-   Supporting raw push for Unpackaged Win32.
+-   Supporting Visual Cloud Toasts for packaged Win32 apps.
+-   Supporting Visual Cloud Toasts for unpackaged Win32 apps.
+
+A Portal Registration flow through
+[AAD (Azure Active Directory)](https://docs.microsoft.com/azure/active-directory/develop/quickstart-register-app)
+will also be defined that removes the dependency of the Push Flow with the Partner Center Portal.
+The RemoteIdentifier GUID in this spec maps to the ApplicationId(ClientId) in the AAD App
```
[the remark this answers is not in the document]
pmpurifoy: It shows as "Application (client) ID" in the portal but I will just use ApplicationID.

specs/PushNotifications/PushNotifications-spec.md
pmpurifoy: We can store the RegistrationTokens for the user and revoke all tokens when they call this function. Should RegisterActivator then not return a token and we just handle it for the user? And UnregisterActivator just accept flags? (What @jonwis said above)
`void RegisterActivator(PushNotificationActivationInfo details);`

`void UnregisterActivator(PushNotificationRegistrationActivators options);

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -2,120 +2,148 @@
...
+the partner center portal by
+[associating an app with the Windows Store](https://docs.microsoft.com/azure/notification-hubs/notification-hubs-windows-store-dotnet-get-started-wns-push-notification).
+As a result, non-store provisioned apps cannot use this feature.
 
-**Client Registration**: Unpackaged 3rd party applications cannot register themselves as a Push target.<br/>
-The only way to get an implicit client side registration is to package the app as an MSIX Desktop Bridge<br/>
-or SSP which may not satisfy 3rd party requirements.
+**Client Registration**: Unpackaged 3rd party applications cannot register themselves as a Push
+target. The only way to get an implicit client side registration is to package the app as an MSIX
+Desktop Bridge or SSP which may not satisfy 3rd party requirements.
```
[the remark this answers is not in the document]
pmpurifoy: SSP = Sparse signed packages

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -2,120 +2,148 @@
...
-At a high level, we need to provide a way for all Win32 applications to subscribe to and receive Push notifications <br/>
-irrespective of their app type. This includes unpackaged apps and packaged win32 (MSIX Desktop Bridge, Sparse Signed Packages). <br/>
-Moreover, all Push scenarios should adhere to OS resource management policies like Power Saver, Network Attribution, <br/>
-Enterprise group policies etc. The Windows App SDK Push component will abstract away the complexities of dealing with WNS <br/>
-channel registrations and Push related activations as much as possible freeing the developer to focus on other app <br/>
-related challenges.
+At a high level, we need to provide a way for all Win32 applications to subscribe to and receive
+Push notifications irrespective of their app type. This includes unpackaged apps and packaged win32
+(MSIX Desktop Bridge, Sparse Signed Packages). Moreover, all Push scenarios should adhere to OS
+resource management policies like Power Saver, Network Attribution, Enterprise group policies etc.
```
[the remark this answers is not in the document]
pmpurifoy: Network Attribution is how much data an app uses. I can add this as a simple definition in parenthesis.

To see this value: goto Settings->Apps->Apps&Features->Your App.

DrusTheAxe (approved):

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -454,34 +534,49 @@ namespace Microsoft.Windows.PushNotifications
     static runtimeclass PushNotificationManager
     {
         // Register an activator using an ActivationInfo context and return a RegistrationToken
-        static PushNotificationRegistrationToken RegisterActivator(PushNotificationActivationInfo details);
+        static void RegisterActivator(PushNotificationActivationInfo details);
```
hulumane, line 537: A little concerned that we removed the token here. The token is being used to infer state like the BackgroundTaskName in the implementation today. How would that flow work without a token?

pmpurifoy: PushNotificationRegistrationToken moving to private struct and we cache generated tokens by the user to revoke when given unregister flag.

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -184,67 +213,97 @@ int main()
...
 {
-    PushNotificationActivationInfo info(
-        PushNotificationRegistrationKind::PushTrigger,
-        winrt::guid("BACCFA91-F1DE-4CA2-B80E-90BE66934EC6"));
+    PushNotificationActivationInfo info = nullptr;
+    if (PushNotificationManager::IsActivatorSupported(PushNotificationRegistrationActivators::PushTrigger))
+    {
+        // Register the ComActivator guid
+        info = PushNotificationActivationInfo(PushNotificationRegistrationActivators::PushTrigger,
+            winrt::guid("BACCFA91-F1DE-4CA2-B80E-90BE66934EC6"));
```
jonwis, line 241: If this is a sample GUID created by the application, consider hiding it here or making it invalid so people don't cut-and-paste this into their code, like:

```c++
// Be sure to create your own activator ID for each app and registration
static const winrt::guid g_myActivatorId("BACCFA91-F1DE-4CA2-xxxx-90BE66934EC6");

...

info = PushNotificationActivationInfo(PushNotificationRegistrationActivators::PushTrigger, g_myActivatorId);
```

jonwis (approved):

PR #1066, opened 2021-07-15
Add source link to MRTCore

huichen123:


rohanp-msft:
Do we need to update any of the tasks in the YML file?

huichen123:
> Do we need to update any of the tasks in the YML file?

don't think so. the source link is embedded in the pdb files.

rohanp-msft (approved):

jefgen:
Just curious, this also works for native code? The [docs](https://docs.microsoft.com/en-us/azure/devops/artifacts/symbols/setting-up-github-sourcelinking?view=azure-devops#:~:text=Source%20link%20is%20a%20set%20of%20tools%20that,the%20Microsoft.SourceLink.GitHub%20package%20reference%20to%20your%20project%20file.) seem to only mention .NET code

huichen123:
> Just curious, this also works for native code? The [docs](https://docs.microsoft.com/en-us/azure/devops/artifacts/symbols/setting-up-github-sourcelinking?view=azure-devops#:~:text=Source%20link%20is%20a%20set%20of%20tools%20that,the%20Microsoft.SourceLink.GitHub%20package%20reference%20to%20your%20project%20file.) seem to only mention .NET code.
> 
> EDIT: Also, do we need to add `/p:SourceLinkCreate=true` ?

doc has a section for "Using Source Link in C++ projects". Not sure about the command line parameter. I see the source link in the pdb files. Unfortunately current binary is not compatible due to namespace change, I cannot try by replacing binaries and debugging.

jefgen:
Thanks Hui.

FWIW, the docs [here](https://github.com/dotnet/sourcelink#using-source-link-in-c-projects) seem to imply that all you need to do is add the packages (which is what this PR does).

Looking at the build [log](https://dev.azure.com/ms/ceef01e9-e5e2-46d8-bbe1-c83450ca8182/_apis/build/builds/200019/logs/136), I do see various `/sourcelink` being added. For example:
`/sourcelink:"x64\Release\Microsoft.Windows.ApplicationModel.Resources.sourcelink.json"`

So maybe this is all that is needed. :)

PR #1101, opened 2021-07-21
Add WindowsAppSDKTelemetryResource project

ZenBird:
* Adding a script to auto generate the rc file for the resource dll that will contain the values needed to log PartB telemetry. The PartB struct contains the SDK version, experimentation flag and other data
* The .rc file for the resource project is generated from the resource.templ file checked in with the project.
* The script must be executed before running the build step on this solution.
* Adding WindowsAppSDKTelemetry.h that will reference this resource dll to construct the PartB structure.


dev/WindowsAppSDKTelemetryResource/WindowsAppSDKTelemetryResourceLib/WindowsAppSDKTelemetry.cpp
```diff
@@ -0,0 +1,47 @@
...
+    if (module)
+    {
+        char buffer[256];
+        if (0 < LoadStringA(module.get(), id, buffer, ARRAYSIZE(buffer)))
+        {
+            return buffer;
+        }
+        else
+        {
+            LOG_LAST_ERROR_MSG("Unable to load resource string, id: %d, error: %d", id, GetLastError());
```
MythiliM, line 25: Should we log telemetry using non-reunion telemetry so that we can track # of load DLL failures? (or) are there other ways to know that telemetry log failed? Maybe check for null package version from all events with COMMON_WINDOWSAPPSDK_PARAMS?

ZenBird: Is there a generic WindowsAppSDK provider that we can log against?

MythiliM: Please create one.

dev/Common/WindowsAppSDKTelemetry.h
```diff
@@ -0,0 +1,24 @@
...
+
+#ifdef __WIL_TRACELOGGING_H_INCLUDED
+#error "WIL Tracelogging.h must not be explicitly included when including this file"
+#endif
+
+char* WINDOWSAPPSDK_PACKAGE_VER();
+char* WINDOWSAPPSDK_EXPERIMENTATION_LEVEL();
+
+#define _GENERIC_PARTB_FIELDS_ENABLED \
+        TraceLoggingStruct(4, "PartB_WINDOWSAPPSDK_API"), \
```
MythiliM, line 17: Can you please change to COMMON_WINDOWSAPPSDK_PARAMS

MythiliM (approved):

ZenBird (commented):
/azp run

PR #1107, opened 2021-07-23
Shrink DDLM windows.appExtension Name to fit in the old <=39 maxlen

DrusTheAxe:
DDLM's ShortVersionTag was missing the suffix digit, e.g. "preview1" as a short tag was "p" but so was "preview2". Needs to be "p1" / "p2" ... Changed the ShortVersionTag from VersionTag[0] to VersionTag[0] + VersionTag[-1] if character is an ASCII digit (0-9).

Thus a ShortVersionTag is "" if VersionTag is "", else ShortVersionTag is an ASCII letter (A-Z) optionally followed an ASCII digit (0-9), as per the first and last characters in VersionTag.

Changed the DDLM windows.appExtension's Name to fit in the old maxlen<=39 constraint (we bulged due to ProjectReunion->WindowsAppSDK rename and the ShortVersionTag didn't support a suffix digit)

Tests pass as expected.

This requires a matching change in the aggregator build when creating the AppxManifest.xml for the DDLM package.


riverar:
@DrusTheAxe Will this impact MddBootstrap? It currently accepts a full unnumbered "preview" tag.

DrusTheAxe:
> @DrusTheAxe Will this impact MddBootstrap? It currently accepts a full unnumbered "preview" tag.

Not negatively. The digit suffix is optional. You'll need the newer bootstrap dll to handle it but the older bootstrap works for 0.8-preview and it was experimental / not-stable in 0.8 anyway so no one should be negatively affected

kythant (approved):

PR #1112, opened 2021-07-23
Roadmap updates

andrewleader:
Lots of small updates...

* Update the "Use it now" to mention 0.8 Experimental for unpackaged apps
* Rename "Localization" to "Resource management" as that's the name we're using in docs/tags/etc
* Clarify that 0.8/1.0 doesn't support incrementally adopting WinUI 3 (will be supported with XAML Islands in future).
* Remove 0.5 from roadmaps and add "Future" column for post-1.0
* Add Windowing to the front-page roadmap snippet
* Add WinUI 3 to the list of app types
* Clarify that WinUI 3 in the tables only works for WinUI 3 apps until the Future release


dayman40 (approved):

riverar:
The added qualifier (WinUI 3 apps) is confusing to me. What does this mean exactly / what are you trying to say here? C++, WPF, etc. can bring in WinUI and WAS components a la carte.

andrewleader:
@riverar using WinUI 3 from a WPF or WinForms apps isn't supported yet. You can use other WinAppSDK APIs from WPF/WinForms apps, but not WinUI 3 yet (that requires XAML Islands). Hence why that qualifier was added only to the WinUI 3 row.

riverar:
I'm struggling with the qualifier because I don't think there's anything inherently special about WPF, WinForms, etc. From a XAML compiler standpoint, completely agree -- and maybe that's what you're trying to communicate. But my understanding is that folks can still bootstrap (w/ dynamic dependencies) and create an instance of `Window` and build up their UI manually.

Happy to chat outside GitHub if you'd like too. Rest of the changes look good.

andrewleader:
I'm not sure many people know how to bootstrap and manually mix UI's like that? From a general developer's standpoint, a WPF app (an app that uses WPF for its UI) cannot easily (and in a supported fashion) use WinUI 3 controls in it. The WPF app would have to completely switch all of its UI to WinUI 3, or do hackery :)

XAML Islands would enable WPF apps to use WinUI 3 in a supported fashion. Sure, you possibly could hack things together, maybe display separate UI stacks on separate windows, but that's not an officially supported scenario and what our customers have been asking us for is to be able to use WinUI 3 controls in their existing WPF app.

riverar:
That's fair.

Can we maybe clarify via exclusion instead? E.g. `(excl. WPF/WinForms)`

I think devs will be confused with `WinUI 3 app`. They'll add WinUI to their [any] project and assume it's now "a WinUI 3 app".

Of course we can try it and revise as needed too.

andrewleader:
@riverar how about this way of conveying it? Trying something slightly different

image

riverar:
Much better, thanks for the consideration!

PR #1114, opened 2021-07-23
Add isActivatorSupported to PushNotifications (Draft)

pmpurifoy:


dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -244,4 +251,90 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
             LOG_IF_FAILED(::CoRevokeClassObject(static_cast<DWORD>(token.Cookie())));
         }
     }
+
+    bool PushNotificationManager::IsBIAvailable()
+    {
+        BackgroundTaskBuilder builder = BackgroundTaskBuilder();
```
danielayala94, line 257: Nit: This call can be inside the try block.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -244,4 +251,90 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        catch (...)
+        {
+            return false;
+        }
+        return true;
+    }
+
+    bool PushNotificationManager::IsActivatorSupported(PushNotificationRegistrationOptions const& options)
+    {
+        if (!WI_AreAllFlagsClear(options, PushNotificationRegistrationOptions::PushTrigger | PushNotificationRegistrationOptions::ComActivator |
```
danielayala94, line 271: Do we have a WI_* function that can tell you if the variable is zero essentially without passing all flags? Can you double check?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -244,4 +251,90 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        bool isBIFlagSet = WI_IsFlagSet(options, PushNotificationRegistrationOptions::PushTrigger) ||
+                                    WI_IsFlagSet(options, PushNotificationRegistrationOptions::ComActivator);
+        bool isProtocolActivatorSet = WI_IsFlagSet(options, PushNotificationRegistrationOptions::ProtocolActivator);
+
+        if (isBIFlagSet && isProtocolActivatorSet)
+        {
+            return false;
+        }
+
+        if (!AppModel::Identity::IsPackagedProcess())
```
danielayala94, line 286: You can do this instead:

```
if (AppModel::Identity::IsPackagedProcess() && isBIAvailable())
    return isBIFlagSet;
else
    return isProtocolActivatorSet;
```

pmpurifoy: Yup. This is much cleaner.

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: WI_IsSingleFlagSet checks if a single bit is set.

PR #1125, opened 2021-07-26
MRTCore use CSWinRT 1.3.3

huichen123:


kythant (approved):

manodasanW (approved):

axelandrejs (approved):

dev/MRTCore/build/versions.props
```diff
@@ -20,8 +20,8 @@
     <CurrentLiftedIXPPackageVersion>10.0.20170.1001-200713-1000.rs-onecore-dep-ixp1</CurrentLiftedIXPPackageVersion>
     <CurrentWilPackageVersion>1.0.190716.2</CurrentWilPackageVersion>
     <MuxCustomBuildTasksPackageVersion>1.0.83-winui3</MuxCustomBuildTasksPackageVersion>
-    <DotNetCoreSdkVersion>5.0.202</DotNetCoreSdkVersion>
-    <DotNetCoreRuntimeVersion>5.0.5</DotNetCoreRuntimeVersion>
```
jefgen, line 24: Does the Dot Net Runtime installer also need to be updated too?

https://github.com/microsoft/WindowsAppSDK/blob/main/build/DownloadDotNetRuntimeInstaller.ps1#L18

huichen123: runtime should be part of SDK, and in fact the SDK with correct version is already installed. I don't want to maintain that script and plan to remove it.

jefgen: Hmm, the script is also used as part of the Helix tests as well though.
See the pipeline step here: https://github.com/microsoft/WindowsAppSDK/blob/7c39e1bcb3e48ac2e07fd5707f7ab9c64326200c/build/AzurePipelinesTemplates/WindowsAppSDK-RunHelixTests-Job.yml#L62-L69

However, I'm not sure if it is *actually necessary* for the Helix tests -- as the dotnet runtime binaries are also brought along by the `WindowsAppSDK.Helix.Test.NetCore` project in the test payload.

I've created a [branch](https://github.com/microsoft/WindowsAppSDK/tree/user/jefgen/test-removing-DotNet-installer-Helix) to test removing it from the Helix test payload. The build results are [here](https://dev.azure.com/ms/ProjectReunion/_build/results?buildId=204269&view=results) (still running). If it passes then that will confirm if it's actually needed for the Helix tests or not.

jefgen: Update: The build [passed](https://dev.azure.com/ms/ProjectReunion/_build/results?buildId=204269&view=results) (though there are some unrelated issues with CodeSign validation it seems).

So I think the script could be removed.

jefgen: Created a PR here to remove the dotnet installer from the Helix pipeline, so there is one less dependency https://github.com/microsoft/WindowsAppSDK/pull/1126

PR #1126, opened 2021-07-27
Remove .NET installer script from repo (remove from Helix pipeline and MRT core pipeline)

jefgen:
The dotnet installer exe isn't necessary, as the .NET runtime bits are brought along with the `WindowsAppSDK.Helix.Test.NetCore` project in the test payload for Helix.

For MRT Core (per Hui): The runtime should be part of SDK, and in fact the SDK with correct version is already installed.


huichen123 (commented):
since you are here, can you also remove one in build-mrt.yml as well?

jefgen:
> since you are here, can you also remove one in build-mrt.yml as well?

Sure, I'll also remove the file from the repo too.

huichen123 (approved):

axelandrejs (approved):

Issue #111, opened 2020-07-14
App Lifecycle


andreww-msft:
<table>
<tr>
<th>Summary</th>
<th>Status</th>
</tr>
<tr>
<td><p><strong>Customer / job-to-be-done</strong>: App developers using WPF, WinForms, or Win32 C++ need to register for certain activation types like file associations and startup activation, need to selectively single-instance their apps, and need to use power states to selectively suspend their apps.</p><p><strong>Problem</strong>: Non-packaged apps use a different set of APIs to achieve all of the above, and they also don't have access to power state notifications</p>

<p><strong>How does our solution make things easier?</strong> All types of apps will benefit from the same single API regardless of app type, apps can now listen to power states, and WIn32 apps have a built-in easy way to single instance.</p>

</td>

<td>✅&nbsp;Problem&nbsp;validation</br>
❌ Docs</br>
🔄&nbsp;<a href="https://github.com/microsoft/ProjectReunion/pull/434">Dev&nbsp;spec</a></br>
❌&nbsp;Solution&nbsp;validation</br>
❌&nbsp;Implementation</br>
❌&nbsp;Samples
</tr>
</table>

## Summary
Provide a core set of functionality related to app activation and lifecycle. The initial release focuses on the following 4 main feature areas:
1.	Activation.
2.	Selective Multi-instancing.
3.	Restart and Recovery.
4.	Power/State Notifications.

## Rationale
* Enable all apps to register for the core activation contracts, consistently in the same way.
* Enable all apps to use platform support for single-instancing, or selective multi-instancing.
* Enable all apps to request immediate restart, or to opt in to recovery/restart in the event of app failure or system reboot.
* Enable all apps to get power/battery/user/system-state change notifications so that they can be better power/battery citizens.

This aligns with [Project Reunion's roadmap](https://github.com/Microsoft/ProjectReunion/blob/master/docs/roadmap.md):

* Enable all applications (packaged and not-packaged) to use the same mechanisms for managing app activation and lifecycle.

## Scope

| Capability | Priority |
| :---------- | :------- |
| This proposal will allow developers to use the same mechanism to register for different activation kinds regardless of app technology (classic Win32, Windows Forms, WPF, UWP or WinUI). | Must |
| This proposal will allow developers to use the same mechanism to get rich activation event arguments regardless of app technology. | Must |
| This proposal will allow users to activate any app by double-clicking it in file explorer, or typing in a path in a command window. | Must |
[...]

PR #1134, opened 2021-07-28
AppLifecycle Restart API

hamza-usmani:
Bringing in Restart API spec into main. This spec addresses the gaps in restart today and how we can expose new functions to manage restart across all app types.

### DIFF
--- a/specs/AppLifecycle/Restart/restartApi.markdown
+++ b/specs/AppLifecycle/Restart/restartApi.markdown
@@ -32,8 +32,8 @@ cannot restart immediately on-command with specific arguments/state. This applie
 
 ### Reunion Restart API components 
 
-1.  Expose a new function: **RequestRestartNow** function to enable any
-    app (win32/packaged win32, UWP, WinUI apps) to terminate and restart itself on command, and to provide an
+1.  Expose a new function: **RequestRestartNow** function to enable any packaged or unpackaged Win32
+    app to terminate and restart itself on command, and to provide an
     arbitrary command-line string for the restarted instance.
 
 
--- a/specs/AppLifecycle/Restart/restartApi.markdown
+++ b/specs/AppLifecycle/Restart/restartApi.markdown
@@ -28,18 +28,11 @@ goals of this specific API are the following:
 
 Addressing the gaps that exist today: at a high-level, Win32 apps can
 register with the OS to restart in update/hang/reboot scenarios but
-cannot restart immediately on-command with specific arguments/state. UWP
-apps can restart on-command with command-line restart arguments but
-cannot register with the OS Restart Manager for update/hang/reboot
-scenarios. These are the gaps we need to address.
+cannot restart immediately on-command with specific arguments/state. This applies to packaged (Centennial) and unpackaged apps. These are the gaps we need to address.
 
 ### Reunion Restart API components 
 
-1.  Expose a new function: **RegisterForRestart**, which allows any app (win32/packaged win32, UWP, WinUI apps)
-    to register itself for restart if it was running when a system
-    update occurs or if an app hangs/crashes unexpectedly.
-
-2.  Expose a new function: **RequestRestartNow** function to enable any
+1.  Expose a new function: **RequestRestartNow** function to enable any
     app (win32/packaged win32, UWP, WinUI apps) to terminate and restart itself on command, and to provide an
     arbitrary command-line string for the restarted instance.
 
@@ -280,53 +273,6 @@ string launchArguments);</td>
 
 ### New API Details
 
-#### RegisterForRestart
-
-<table>
-<colgroup>
-<col style="width: 40%" />
-<col style="width: 51%" />
-<col style="width: 8%" />
-</colgroup>
-<thead>
-<tr class="header">
-<th>New API </th>
-<th>Description</th>
-<th>Priority</th>
-</tr>
-</thead>
-<tbody>
-<tr class="odd">
-<td>static bool RegisterForRestart(string restartArguments, RestartContext context)</td>
-<td><p>Equivalent to the Win32 RegisterApplicationRestart function.</p>
-<p>restartArguments is the same as the pwzCommandline parameter in the Win32 RegisterApplicationRestart. It is a string of command-line arguments which will be passed as restart arguments.</p>
-<p>restartArguments become the restarted/new instance’s LaunchActivatedEventArgs.</p></td>
-<td>P3</td>
-</tr>
-<tr class="even">
-<td>[Flags] enum RestartContext</td>
-<td><p>The values passed for RestartContext indicate which cases</p>
-<p>the app <em>should be restarted.<br />
-<br />
-</em>The possible cases are the same as dwFlag values listed above in Win32’s RegisterApplicationRestart<strong>:</strong></p>
-<ul>
-<li><p>-1: ALL</p></li>
-<li><p>1: CRASH</p></li>
-<li><p>2: HANG</p></li>
-<li><p>4: PATCH</p></li>
-<li><p>8: REBOOT</p></li>
-</ul>
-<p>[Flags] enum RestartContext { All = -1, Crash = 1, Hang = 2, Patch = 4, Reboot = 8 }</p></td>
-<td>P3</td>
-</tr>
-<tr class="odd">
-<td>static bool UnregisterForRestart()</td>
-<td>Equivalent to the Win32 UnregisterApplicationRestart function.</td>
-<td>P3</td>
-</tr>
-</tbody>
-</table>
-
 #### RequestRestartNow
 
 <table>
@@ -362,40 +308,6 @@ string launchArguments);</td>
 </table>
 
 ### API Sample Code
-
-#### RegisterForRestart Sample
-```c#
-public class App {
-
-        // App was just launched
-        protected override void OnLaunched(LaunchActivatedEventArgs args)
-        {
-            // If the application was terminated for example and has now been restarted
-            if (args != NULL)
-            {
-                if (args.PreviousExecutionState == ApplicationExecutionState.Terminated)
-                {
-                    Telemetry::WriteLine("Restart activated: " + args.Arguments);
-                }
-            }
-        }
-
-        // We want to register the app for restart if termination happens only after crash
-        protected bool restartOnUpdate(){
-            bool result = AppLifecycle::RegisterForRestart("args for new instance", RestartContext::Crash);
-            Telemetry::WriteLine("RegisterForRestart: Restart Context: " + RestartContext::Crash.ToString() 
-            + ". Result: " + result ? "success" : "fail");
-
-            // Do normal app behavior
-        }
-
-        // We want to unregister for post-crash restarts
-        protected bool removeRestartOnUpdate(){
-            bool result = AppLifecycle::UnregisterForRestart();
-            Telemetry::WriteLine("UnregisterForRestart: Result: " + result ? "success" : "fail");
-        }
-}
-```
 
 ### RequestRestartNow Sample
 
@@ -440,7 +352,6 @@ Event Name: **AppRestart**
 
 | Metadata                | Description                                                                                       | Data type |
 |-------------------------|---------------------------------------------------------------------------------------------------|-----------|
-| RestartContext          | Reason for an App to register for restart and then attempt to restart                             | Enum      |
 | AppRestartFailureReason | Reason an app restart failed. This will be collected when an app requests a restart now but fails | Enum      |
 | AppName                 | Name of the app which restarted                                                                   | String    |
 | AppType                 | UWP App or Centennial App or Win32 app                                                            | String    |
--- a/specs/AppLifecycle/Restart/restartApi.markdown
+++ b/specs/AppLifecycle/Restart/restartApi.markdown
@@ -174,10 +174,13 @@ string launchArguments);</td>
  <ul>
 <li>When the app is restarted, LaunchActivatedEventArgs.PreviousExecutionState will have the value Terminated so that the app can distinguish between a resume and a restart. </li>
 </ul>
-<p>Returns a failure reason for why the restart didn’t successfully complete (AppRestartFailureReason).
+<p>If the request to restart is successful, the application will be restarted immediately. The user does not have to do anything else/handle a return value in any way.
+
+If the request fails due to a nonfatal reason, a failure reason is returned for why the restart didn’t successfully complete. This method leverages an existing enum (AppRestartFailureReason) which has the following relevant return fields.
 <ul>
-<li>This API leverages an existing Enum (AppRestartFailureReason), but not all return values are relevant.</li>
-<li>This API does not require for the app to be in foreground, and so AppRestartFailureReason::NotInForeground will never be returned</li>
+<li> InvalidUser: Could not restart for the specified user.</li>
+<li> RestartPending: A restart is already in progress.</li>
+<li> Other: Unspecified failure.</li>
 </ul>
 </p></td>
 </tr>
@@ -205,9 +208,8 @@ public class App
             switch (result)
             {
                 case AppRestartFailureReason::RestartPending:
-                    // An example of how to handle this case is updating a progress bar to the user
-                    // indicating immediate restart
-                    UpdateProgressBar();
+                    // In this case the restart could not be completed because another restart is pending and we can try to restart again after a period of time
+                    RetryRestartAfterXSeconds();
                     break;
                 case AppRestartFailureReason::Other:
                     // In this case, a dialog box will appear informing of an error requesting restart
@@ -226,7 +228,10 @@ public class App
         switch (result)
         {
             case AppRestartFailureReason::RestartPending:
-                Telemetry::WriteLine("Restart request successful.");
+                Telemetry::WriteLine("Another restart is currently pending.");
+                break;
+            case AppRestartFailureReason::InvalidUser:
+                Telemetry::WriteLine("Current user is not signed in or not a valid user.");
                 break;
             case AppRestartFailureReason::Other:
                 Telemetry::WriteLine("Failure restarting.");
--- a/specs/AppLifecycle/Restart/restartApi.markdown
+++ b/specs/AppLifecycle/Restart/restartApi.markdown
@@ -5,30 +5,35 @@
 - [Restart API](#restart-api)
 - [Contents](#contents)
 - [Background](#background)
-  - [Description](#description)
+  - [Purpose](#purpose)
 - [API](#api)
-  - [Existing API Summary](#existing-api-summary)
+  - [Existing API](#existing-api)
+    - [Summary](#summary)
   - [Existing API Details](#existing-api-details)
+    - [Existing API](#existing-api-1)
   - [New API Details](#new-api-details)
-    - [RequestRestartNow](#requestrestartnow)
-  - [Examples and Scenarios](#examples-and-scenarios)
-- [Telemetry](#telemetry)
+    - [New API](#new-api)
+    - [Description](#description)
+    - [Mechanism](#mechanism)
+    - [Examples](#examples)
 
 # Background
 
-This spec addresses the **Restart APIs** in the AppLifecycle component. At a high-level, Win32 apps can
-register with the OS to restart in update/hang/reboot scenarios but cannot restart immediately on-command 
-with specific arguments/state. This applies to packaged (Desktop Bridge/Centennial) and unpackaged apps. This API will address
+This spec addresses the **Restart APIs** in the AppInstance component. At a high-level, Win32 apps can
+register with the OS to restart in update/hang/reboot scenarios but cannot initiate an explicit restart
+with specific arguments/state. This applies to packaged (Desktop Bridge) and unpackaged apps. This API will address
 these gaps and provide the ability for any Win32 app to restart immediately.
 
-## Description
+## Purpose
 
-The goal of this feature is the new function: **RequestRestartNow**, which will enable any packaged or unpackaged Win32
-app (including Console, WinMain, Windows Forms, WPF and WinUI) to terminate and restart itself on command, and to provide an arbitrary command-line string for the restarted instance.
+The goal of this feature is the new function: **Restart**, which will enable any packaged or unpackaged Win32
+app (including Console, WinMain, Windows Forms, WPFI) to terminate and restart itself on command, and to provide an arbitrary command-line string for the restarted instance. 
+
+This behavior (restarting an app immediately) is not currently available to Win32 applications. This new method will fill the gap for Win32 applications and align with CoreApplication's existing ``CoreApplication.RequestRestartAsync``.
 
 # API
 
-## Existing API Summary
+## Existing API
 
 1.  **Restart-me-now**. CoreApplication exposes the
     [RequestRestartAsync](https://docs.microsoft.com/en-us/uwp/api/windows.applicationmodel.core.coreapplication.requestrestartasync?view=winrt-18362)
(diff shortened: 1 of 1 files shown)


riverar:
Please review spec authoring/merge guidance at https://github.com/microsoft/WindowsAppSDK/tree/main/specs.

Also I don't recall this ever going through review?

Samuel12321:
Issue #111 

Does this support restarting with identity

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -0,0 +1,357 @@
...
+    command-line.
+
+2.  **Recovery-in-progress**. While the app is doing work in its
+    recovery callback, the app must periodically call
+    [ApplicationRecoveryInProgress](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-applicationrecoveryinprogress)
+    -- if it doesn't call within the registered ping interval, WER will
+    terminate the process.
+
+
+<table>
```
DrusTheAxe, line 79: Why all the HTML? Markdown supports tables (and with much simpler syntax)

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -0,0 +1,357 @@
...
+
+- [Existing API Details](#existing-api-details)
+
+- [New API Details](#new-api-details)
+
+- [API Sample Code](#api-sample-code)
+
+[3. Telemetry](##restartApi\telemetry)
+
+[4. Resources](#resources)
```
DrusTheAxe, line 22: I think your TOC is out of sync with your actual content

SUGGESTION: Use VSCode with the [Marodnw All in One](https://marketplace.visualstudio.com/items?itemName=yzhang.markdown-all-in-one) extension, then issue commands via Ctrl-P

1. Markdown All in One: Add/Update section numbers
2. Put cursor at top of file
3. Markdown All in One: Create Table of Contents

and if/when you make edits the pair of ...Add/Update section numbers + ...Update Table of Contents keeps everything correct and current

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -0,0 +1,357 @@
+# Reunion Restart API Spec
```
DrusTheAxe, line 1: Please use [specs/spec_template.md](https://github.com/microsoft/WindowsAppSDK/blob/main/specs/spec_template.md)

You can find existing  specs @ [specs](https://github.com/microsoft/WindowsAppSDK/blob/main/specs) for reference examples. For example a really good simple example see [DeploymentAPI](https://github.com/microsoft/WindowsAppSDK/blob/main/specs/Deployment/DeploymentAPI.md). For more extensive specs [AppLifecycle Activation](https://github.com/microsoft/WindowsAppSDK/blob/main/specs/AppLifecycle/Activation/AppLifecycle%20Activation.md) and [Dynamic Dependencies]https://github.com/microsoft/WindowsAppSDK/blob/main/specs/dynamicdependencies/DynamicDependencies.md) are good examples.

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -0,0 +1,357 @@
...
+<thead>
+<tr class="header">
+<th>New API </th>
+<th>Description</th>
+<th>Priority</th>
+</tr>
+</thead>
+<tbody>
+<tr class="odd">
+<td>static AppRestartFailureReason RequestRestartNow(string restartArguments)</td>
```
DrusTheAxe, line 293: I can't tell what kind of API this is. WinRT? But it's just a function w/o mention where it lives (no runtimeclass or interface mentioned). Flat-C? But then why 'static', and the data types imply otherwise.

If you're intending WinRT I'd expect to see a MIDL3 .idl (or at least the relevant subset of one)

Please use the standard template and see other existing specs for representative examples how to express your API.

DrusTheAxe (changes requested):

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -0,0 +1,254 @@
...
+    // The new assets have downloaded and installed, and now the app has to request a restart immediately
+    protected void UpdateInstallComplete()
+    {
+        // Checking if the assets have updated successfully and there are no pending updates
+        if (!CheckForUpdate())
+        {
+            AppRestartFailureReason result = AppLifecycle::RequestRestartNow("args for new instance");
+            switch (result)
+            {
+                case AppRestartFailureReason::RestartPending:
```
aeloros, line 207: RestartPending is only returned if another restart is in progress.  Since that is more likely for the async variation in the platform, this should only ever happen if the API is called twice by the same process.

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -401,65 +396,68 @@ string launchArguments);</td>
...
+                    Telemetry::WriteLine("Restart activated: " + args.Arguments);
+                }
+            }
+        }
+
+        // We want to register the app for restart if termination happens only after crash
+        protected bool restartOnUpdate(){
+            bool result = AppLifecycle::RegisterForRestart("args for new instance", RestartContext::Crash);
+            Telemetry::WriteLine("RegisterForRestart: Restart Context: " + RestartContext::Crash.ToString() 
+            + ". Result: " + result ? "success" : "fail");
```
DrusTheAxe: NIT: Indent

```
            Telemetry::WriteLine("RegisterForRestart: Restart Context: " + RestartContext::Crash.ToString() 
                + ". Result: " + result ? "success" : "fail");
```

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -401,65 +396,68 @@ string launchArguments);</td>
...
-           }               
-    }  
+```c#
+public class App {
+       // Let's assume this method is run when an app updates assets.
+       // The new assets have downloaded and installed, and now the app has to restart immediately
+       protected void updateInstallComplete(){
+           // code here that checks the app update has installed
+
+           WCHAR buffer\[1024\];
```
DrusTheAxe: Why the backslashes? Are they needed in a code block (```blah```)?

DrusTheAxe: NIT: `WCHAR buffer[1024]{};`

DrusTheAxe: Magic constant. And why 1024? Why not 256 or 2K or...?

Also, buffer is assigned in some cases but not used?
If it's needed, why not std::wstring?
But it's unclear if this is needed at all

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -401,65 +396,68 @@ string launchArguments);</td>
...
-    }  
+```c#
+public class App {
+       // Let's assume this method is run when an app updates assets.
+       // The new assets have downloaded and installed, and now the app has to restart immediately
+       protected void updateInstallComplete(){
+           // code here that checks the app update has installed
+
+           WCHAR buffer\[1024\]; 
+           AppRestartFailureReason result = AppLifecycle::RequestRestartNow("args for new instance");
```
DrusTheAxe: NIT: {} initialization

`AppRestartFailureReason result{ AppLifecycle::RequestRestartNow("args for new instance") };`

specs/AppLifecycle/Restart/restartApi.markdown
hamza-usmani: Updated with this information.

jonwis:
Apologies for the delay here. API reviewers, please leave your comments on the document. We'll go through them at 1pm PST on January 11th, 2022.  We'll specifically be reviewing https://github.com/microsoft/WindowsAppSDK/pull/1134/files/c7c31991d80601c06ee52e8ec7c5e3929cfe7d28 in case additional items get pushed.

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -0,0 +1,259 @@
...
+not only for measuring quality but also for helping with investigations
+of failures. Following is the event which can be used to capture
+telemetry information:
+
+Event Name: **AppRestart**
+
+| Metadata                | Description                                                                                       | Data type |
+| ----------------------- | ------------------------------------------------------------------------------------------------- | --------- |
+| AppRestartFailureReason | Reason an app restart failed. This will be collected when an app requests a restart now but fails | Enum      |
+| AppName                 | Name of the app which restarted                                                                   | String    |
```
riverar, line 258: ⚠️ AppName could contain PII. Do not recommend using as-is.

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -0,0 +1,259 @@
...
+</ul>
+</p></td>
+</tr>
+</tbody>
+</table>
+
+When an application calls RequestRestartNow, the mechanism for restart will be as follows:
+1. Application calls RequestRestartNow API.
+2. The API will CreateProcess with the Agent (a helper EXE in the framework package containing the API) and will then wait for an an acknowledgment (ACK) from the Agent.
+3. The Agent will get the executable path of the application and signal an ACK to the API. The Agent will then then terminate the application and then CreateProcess with the application (restarted).
```
riverar, line 193: > The Agent will then terminate the application [...]

Am concerned about cleanup here, if the Agent handles app termination.

An app may only cleanup unmanaged resources (e.g. shared GDI handles) when terminating cleanly. Is the thinking here that the developer needs to redesign their application to have a cleanup-first-then-call-RequestRestartNow code path? Perhaps this can be redesigned so that the Agent sends back an ACK to the caller, whom can then attempt to terminate gracefully?

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -0,0 +1,259 @@
...
+ <ul>
+<li>When the app is restarted, LaunchActivatedEventArgs.PreviousExecutionState will have the value Terminated so that the app can distinguish between a resume and a restart. </li>
+</ul>
+<p>If the request to restart is successful, the application will be restarted immediately. The user does not have to do anything else/handle a return value in any way.
+
+If the request fails due to a nonfatal reason, a failure reason is returned for why the restart didn’t successfully complete. This method leverages an existing enum (AppRestartFailureReason) which has the following relevant return fields.
+<ul>
+<li> InvalidUser: Could not restart for the specified user.</li>
+<li> RestartPending: A restart is already in progress.</li>
+<li> Other: Unspecified failure.</li>
```
riverar, line 183: It's not clear what the app developer will do these reasons. `InvalidUser` leaks out some obscure internal details, `RestartPending` means I'm about to die and any sort of handling here is not guaranteed to run, and `Other` is useless.

Perhaps this can be replaced with a Boolean indicating if the request was accepted or not.

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -0,0 +1,259 @@
...
+<tr class="odd">
+<td>static AppRestartFailureReason <strong>RequestRestartNow</strong>(string restartArguments)</td>
+<td><p>Equivalent to CoreApplication.RequestRestartAsync, except that it is <em>not</em> async.
+<ul>
+<li>If the restart fails, but the user subsequently launches the app manually, the app will launch normally and no restart arguments will be passed. </li>
+<li>If the app has any in-process background tasks running when it calls this API, those tasks will be cancelled in the normal way. Out-of-process background tasks will not be affected. </li>
+</ul>
+ The restarted instance will be activated with Windows.ApplicationModel.Activation.LaunchActivatedEventArgs.</p>
+ <ul>
+<li>When the app is restarted, LaunchActivatedEventArgs.PreviousExecutionState will have the value Terminated so that the app can distinguish between a resume and a restart. </li>
```
riverar, line 175: What happens if the restart takes too long and the user starts the app manually? What is the value for `PreviousExecutionState`? Or will multiple instances start if supported, with varying states?

Perhaps an opportunity for the Agent to provide progress UI?

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -0,0 +1,259 @@
...
+of failures. Following is the event which can be used to capture
+telemetry information:
+
+Event Name: **AppRestart**
+
+| Metadata                | Description                                                                                       | Data type |
+| ----------------------- | ------------------------------------------------------------------------------------------------- | --------- |
+| AppRestartFailureReason | Reason an app restart failed. This will be collected when an app requests a restart now but fails | Enum      |
+| AppName                 | Name of the app which restarted                                                                   | String    |
+| AppType                 | UWP App or Centennial App or Win32 app                                                            | String    |
```
riverar, line 259: Nits:

```suggestion
| AppType                 | UWP app or Desktop Bridge app or Win32 app                                                            | String    |
```

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -0,0 +1,259 @@
...
+public class App 
+{
+    // Let's assume this method is run when an app updates assets.
+    // The new assets have downloaded and installed, and now the app has to request a restart immediately
+    protected void UpdateInstallComplete()
+    {
+        // Checking if the assets have updated successfully and there are no pending updates
+        if (!CheckForUpdate())
+        {
+            AppRestartFailureReason result = AppLifecycle::RequestRestartNow("args for new instance");
```
andrewleader, line 207: The class that the method is in looks like it should be **AppInstance** based on Scott's code pull request (AppLifecycle is the namespace), so should be `AppInstance.RequestRestartNow("args")` (and prolly needs updating in multiple places.

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -0,0 +1,259 @@
...
+</table>
+
+When an application calls RequestRestartNow, the mechanism for restart will be as follows:
+1. Application calls RequestRestartNow API.
+2. The API will CreateProcess with the Agent (a helper EXE in the framework package containing the API) and will then wait for an an acknowledgment (ACK) from the Agent.
+3. The Agent will get the executable path of the application and signal an ACK to the API. The Agent will then then terminate the application and then CreateProcess with the application (restarted).
+
+## Examples and Scenarios
+
+```c#
```
jeffstall, line 197: If this is real C#, then please use "." instead of "::".

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -0,0 +1,259 @@
...
+
+    // In this scenario, assume the app encounters an error during initialization.
+    // The app displays an error dialog, and after the user clicks OK on the dialog the app must restart immediately.
+    // This scenario simply displays logging for these use cases
+    protected void HandleInitializationError()
+    {
+        AppRestartFailureReason result = AppLifecycle::RequestRestartNow("args for new instance");
+        switch (result)
+        {
+            case AppRestartFailureReason::RestartPending:
```
jeffstall, line 230: How would I cancel the previous restart and "upgrade" it with new information? It would be helpful if the sample showed this.

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -0,0 +1,259 @@
...
+        {
+            AppRestartFailureReason result = AppLifecycle::RequestRestartNow("args for new instance");
+            switch (result)
+            {
+                case AppRestartFailureReason::RestartPending:
+                    // In this case the restart could not be completed because another restart is pending and we can try to restart again after a period of time
+                    RetryRestartAfterXSeconds();
+                    break;
+                case AppRestartFailureReason::Other:
+                    // In this case, a dialog box will appear informing of an error requesting restart
```
jeffstall, line 215: Can the sample show how the dialog box would get the correct message text to show the end user?

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -0,0 +1,259 @@
...
+# Contents
+
+- [Restart API](#restart-api)
+- [Contents](#contents)
+- [Background](#background)
+  - [Description](#description)
+- [API](#api)
+  - [Existing API Summary](#existing-api-summary)
+  - [Existing API Details](#existing-api-details)
+  - [New API Details](#new-api-details)
```
jeffstall, line 12: Is there supposed to be MIDL 3 or something in this file to review? There's details about enum values and stuff, but no clear MIDL section with RtCop output, etc.

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -0,0 +1,259 @@
...
+these gaps and provide the ability for any Win32 app to restart immediately.
+
+## Description
+
+The goal of this feature is the new function: **RequestRestartNow**, which will enable any packaged or unpackaged Win32
+app (including Console, WinMain, Windows Forms, WPF and WinUI) to terminate and restart itself on command, and to provide an arbitrary command-line string for the restarted instance.
+
+# API
+
+## Existing API Summary
```
jeffstall, line 31: Can this also clearly identify why an app developer would want to call the new API instead of the existing APIs? It's telling me what the old APIs do, but not being an expert in this area, I'm not clear why I would want to stop calling the old ones and start calling the new ones.

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -0,0 +1,259 @@
...
+<li>If the restart fails, but the user subsequently launches the app manually, the app will launch normally and no restart arguments will be passed. </li>
+<li>If the app has any in-process background tasks running when it calls this API, those tasks will be cancelled in the normal way. Out-of-process background tasks will not be affected. </li>
+</ul>
+ The restarted instance will be activated with Windows.ApplicationModel.Activation.LaunchActivatedEventArgs.</p>
+ <ul>
+<li>When the app is restarted, LaunchActivatedEventArgs.PreviousExecutionState will have the value Terminated so that the app can distinguish between a resume and a restart. </li>
+</ul>
+<p>If the request to restart is successful, the application will be restarted immediately. The user does not have to do anything else/handle a return value in any way.
+
+If the request fails due to a nonfatal reason, a failure reason is returned for why the restart didn’t successfully complete. This method leverages an existing enum (AppRestartFailureReason) which has the following relevant return fields.
```
jeffstall, line 179: I think this is what I'm asking the samples to show how to use properly. How does a developer correctly use this information to help the app or the end user when something goes wrong?

specs/AppLifecycle/Restart/restartApi.markdown
jeffstall, line 207: Please show real args, especially if you need to have things quoted.

riverar:
@jeffstall In the future, can you please click Start Review instead of adding separate comments? Makes reviewing much easier, thanks!

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -0,0 +1,259 @@
...
+<ul>
+<li> InvalidUser: Could not restart for the specified user.</li>
+<li> RestartPending: A restart is already in progress.</li>
```
oldnewthing, line 182: Trying to understand how RestartPending could happen. I think it means "Two threads called RequestRestartNow simultaneously. One of them was accepted, and the restart will occur with that command line. The other one was rejected, so the second request's command line won't actually run. But it's not like you can recover from this in any meaningful way because you're going to die soon. You may even die before you get a chance to act on the return value. All you can do is `Sleep(INFINITE)` and come to terms with the bad choices you've made in your life.

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -0,0 +1,259 @@
...
+</ul>
+ The restarted instance will be activated with Windows.ApplicationModel.Activation.LaunchActivatedEventArgs.</p>
+ <ul>
+<li>When the app is restarted, LaunchActivatedEventArgs.PreviousExecutionState will have the value Terminated so that the app can distinguish between a resume and a restart. </li>
+</ul>
+<p>If the request to restart is successful, the application will be restarted immediately. The user does not have to do anything else/handle a return value in any way.
+
+If the request fails due to a nonfatal reason, a failure reason is returned for why the restart didn’t successfully complete. This method leverages an existing enum (AppRestartFailureReason) which has the following relevant return fields.
+<ul>
+<li> InvalidUser: Could not restart for the specified user.</li>
```
oldnewthing, line 181: I don't see how this value could come out, since there is no `User` parameter. The `InvalidUser` reason is from `RequestRestartForUser(User user)`. So the documentation should just say "Not used."

But maybe the correct thing is to invent a new enum instead of trying to shoehorn our results into an existing enum. What happens if we want to report some new failure mode that `RequestRestartAsync` doesn't have? We have no way to add an enum value to somebody else's enum.

specs/AppLifecycle/Restart/restartApi.markdown
oldnewthing, line 183: Having an `Other` prevents us from adding failure reasons in the future in a non-breaking way. Suppose we want to report an error like `BadAroma`. But old apps expect to receive `Other`, so how do we choose whether to return `Other` or `BadAroma`?

Are we confident that we'll never need to add a new failure reason in the future? Should we document that apps should treat any unrecognized result the same as `Other`?

specs/AppLifecycle/Restart/restartApi.markdown
oldnewthing: Apps already have to deal with abnormal termination in the case of termination-on-suspension, app crash, or the user terminating from Task Manager. Resources like kernel handles and GDI resources are automatically cleaned up at process termination. Other resources like temporary files will have to be cleaned up by the restarted application, just like any other launch after termination. So no new recovery scenarios are introduced here.

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -0,0 +1,259 @@
...
+    {
+        // Checking if the assets have updated successfully and there are no pending updates
+        if (!CheckForUpdate())
+        {
+            AppRestartFailureReason result = AppLifecycle::RequestRestartNow("args for new instance");
+            switch (result)
+            {
+                case AppRestartFailureReason::RestartPending:
+                    // In this case the restart could not be completed because another restart is pending and we can try to restart again after a period of time
+                    RetryRestartAfterXSeconds();
```
oldnewthing, line 212: Why retry? It'll just say "Yup, still pending."

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -0,0 +1,259 @@
...
+            }
+        }
+    } 
+
+    // In this scenario, assume the app encounters an error during initialization.
+    // The app displays an error dialog, and after the user clicks OK on the dialog the app must restart immediately.
+    // This scenario simply displays logging for these use cases
+    protected void HandleInitializationError()
+    {
+        AppRestartFailureReason result = AppLifecycle::RequestRestartNow("args for new instance");
```
oldnewthing, line 227: Better example would be
```csharp
// Restart in safe mode. Maybe that'll avoid whatever it is that prevented us from initializing.
AppRestartFailureReason result = AppLifecycle.RequestRestartNow("/safemode");
```

oldnewthing, line 227: Variable name doesn't match type. Should be

```csharp
AppRestartFailureReason reason = ...
```
or
```csharp
AppRestartResult result = ...
```

specs/AppLifecycle/Restart/restartApi.markdown
oldnewthing: App names have generally been considered non-PII. App paths contain PII, but not the executable name itself.

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -0,0 +1,259 @@
...
+}
+```
+
+# Telemetry
+
+This section explores the telemetry needs for the Restart feature in
+four dimensions – Usage, Reliability, Performance and Failure
+Investigation. The telemetry will include the metadata that are needed
+not only for measuring quality but also for helping with investigations
+of failures. Following is the event which can be used to capture
```
oldnewthing, line 250: Who has access to this telemetry? If the app developer cannot access it, then there's no point listing it in the API documentation. (Don't tell people about things they can't access.)

specs/AppLifecycle/Restart/restartApi.markdown
riverar: I was thinking more along the lines of shared GDI resources like icons and possible impact to Desktop Bridge middleware, but I concede this all falls under abnormal termination handling.

specs/AppLifecycle/Restart/restartApi.markdown
riverar: Darn, I better rename `RafsIceCreamConeMaker.exe`.

(I see there's established [precedent for this](https://docs.microsoft.com/en-us/windows/privacy/required-windows-diagnostic-data-events-and-fields-2004#:~:text=appname%20the%20name%20of%20the%20app%20that%20has%20crashed.) so OK.)

oldnewthing: In practice, software developers do not ship a custom executable name for each customer. (`ContosoDeluxe-BobSmithFrom Denver.exe`, `ContosoDeluxe-JaneJohnsonFromChicago.exe`, `ContosoDeluxe-PatWilsonFromLosAngeles.exe`.) The executable name is the same for all customers: `ContosoDeluxe.exe`.

`RafsIceCreamConeMaker.exe` is presumably a piece of software that is marketed as "Check out this cool program  'Raf's Ice Cream Cone Maker', in which a cartoon character name Raf helps you design the perfect ice cream cone." It doesn't tell you who bought the program. Raf is the program name, not the customer name. (That's the theory at least.)

riverar: One of my favorites, PowerGrep, ships `SetupPowerGREP522RafaelRivera.exe`. (The executable is generated server-side for licensing purposes.) Hex Rays (maker of IDA) also ships custom executables with uniquely identifying details (e.g. order number, user id). But I was also thinking of scenarios where I fire up Visual Studio and just hack on a project and don't necessarily want information about it going back to the mothership. All outliers in the grand scheme, perhaps.

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -0,0 +1,259 @@
...
+
+2.  **Recovery-in-progress**. While the app is doing work in its
+    recovery callback, the app must periodically call
+    [ApplicationRecoveryInProgress](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-applicationrecoveryinprogress)
+    \-- if it doesn't call within the registered ping interval, WER will
+    terminate the process.
+
+| App Type                                                            | Applicable Restart Mechanism                                           | Gap                                                                                                                  |
+| ------------------------------------------------------------------- | ---------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------- |
+| Win32 (unpackaged), Win32/Centennial (packaged), WinUI (Centennial) | RegisterApplicationRestart for OS + app update/restart/crash scenarios | Apps can register with the OS to restart in specific app/OS states, but cannot restart on-command in a healthy state |
```
MikeHillberg, line 69: WinUI shouldn't be separate from Win32; it's just the UI stack in a Win32 app and can be packaged or unpackaged.

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -0,0 +1,259 @@
...
+</thead>
+<tbody>
+<tr class="odd">
+<td>static AppRestartFailureReason <strong>RequestRestartNow</strong>(string restartArguments)</td>
+<td><p>Equivalent to CoreApplication.RequestRestartAsync, except that it is <em>not</em> async.
+<ul>
+<li>If the restart fails, but the user subsequently launches the app manually, the app will launch normally and no restart arguments will be passed. </li>
+<li>If the app has any in-process background tasks running when it calls this API, those tasks will be cancelled in the normal way. Out-of-process background tasks will not be affected. </li>
+</ul>
+ The restarted instance will be activated with Windows.ApplicationModel.Activation.LaunchActivatedEventArgs.</p>
```
MikeHillberg, line 173: That's a CoreApplication API, what happens in a Win32 app?

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -0,0 +1,259 @@
...
+ <ul>
+<li>When the app is restarted, LaunchActivatedEventArgs.PreviousExecutionState will have the value Terminated so that the app can distinguish between a resume and a restart. </li>
+</ul>
+<p>If the request to restart is successful, the application will be restarted immediately. The user does not have to do anything else/handle a return value in any way.
```
MikeHillberg, line 177: I assume "immediately" means that the call doesn't return? (Would help understanding to call this out explicitly)

specs/AppLifecycle/Restart/restartApi.markdown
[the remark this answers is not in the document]
MikeHillberg: Agreed, WinUI shouldn't be separate from Win32; it's just the UI stack in a Win32 app and can be packaged or unpackaged.

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -0,0 +1,259 @@
...
+    \-- if it doesn't call within the registered ping interval, WER will
+    terminate the process.
+
+| App Type                                                            | Applicable Restart Mechanism                                           | Gap                                                                                                                  |
+| ------------------------------------------------------------------- | ---------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------- |
+| Win32 (unpackaged), Win32/Centennial (packaged), WinUI (Centennial) | RegisterApplicationRestart for OS + app update/restart/crash scenarios | Apps can register with the OS to restart in specific app/OS states, but cannot restart on-command in a healthy state |
+
+## Existing API Details
+
+<table>
```
MikeHillberg, line 73: Document style comment: API specs shouldn't use HTML. We want the API spec to roll into the docs, and the docs are Markdown-only. Markdown isn't set up well for tall table rows like this, but that's actually a ploy to discourage tables with tall rows. This API in the docs will become a page, so better to just make it a section.

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -0,0 +1,259 @@
...
+</colgroup>
+<thead>
+<tr class="header">
+<th>New API </th>
+<th>Description</th>
+</tr>
+</thead>
+<tbody>
+<tr class="odd">
+<td>static AppRestartFailureReason <strong>RequestRestartNow</strong>(string restartArguments)</td>
```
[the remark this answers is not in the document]
MikeHillberg: +1 to removing "request". Actually I think `Restart()` might be enough.

specs/AppLifecycle/Restart/restartApi.markdown
MikeHillberg: And what does the rest of the `AppLifecycle` API look like?

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -0,0 +1,259 @@
...
+</colgroup>
+<thead>
+<tr class="header">
+<th>Existing API</th>
+<th>Details</th>
+</tr>
+</thead>
+<tbody>
+<tr class="odd">
+<td><p>HRESULT <strong>RegisterApplicationRestart</strong>(PCWSTR pwzCommandline, DWORD dwFlags);</p></td>
```
MikeHillberg, line 86: I'm confused, this looks like a Win32 API, but the sample makes it appear to be a WinRT API?

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -0,0 +1,259 @@
...
+<li> Other: Unspecified failure.</li>
+</ul>
+</p></td>
+</tr>
+</tbody>
+</table>
+
+When an application calls RequestRestartNow, the mechanism for restart will be as follows:
+1. Application calls RequestRestartNow API.
+2. The API will CreateProcess with the Agent (a helper EXE in the framework package containing the API) and will then wait for an an acknowledgment (ACK) from the Agent.
```
johnstep, line 192: How is the helper process lifetime managed? What happens if the app is running elevated or as a different user? If the helper is not started by the app, it can handle ERROR_ELEVATION_REQUIRED from CreateProcess by trying again with ShellExecute, but that only works if the executable requires admin, and does not guarantee the same user.

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -0,0 +1,259 @@
...
+  - [New API Details](#new-api-details)
+    - [RequestRestartNow](#requestrestartnow)
+  - [Examples and Scenarios](#examples-and-scenarios)
+- [Telemetry](#telemetry)
+
+# Background
+
+This spec addresses the **Restart APIs** in the AppLifecycle component. At a high-level, Win32 apps can
+register with the OS to restart in update/hang/reboot scenarios but cannot restart immediately on-command 
+with specific arguments/state. This applies to packaged (Desktop Bridge/Centennial) and unpackaged apps. This API will address
```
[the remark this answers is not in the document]
hamza-usmani: Good callout, I'll make this clearer.

specs/AppLifecycle/Restart/restartApi.markdown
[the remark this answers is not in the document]
hamza-usmani: Yup, I agree with this. Will update accordingly.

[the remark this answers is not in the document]
hamza-usmani: There previously was a row for UWP but we were given the guidance to remove most/all UWP specific methods/entries from the spec since UWP isn't something we want to continue with in this context.

specs/AppLifecycle/Restart/restartApi.markdown
hamza-usmani: Yes, the reason for the HTML was that Markdown doesn't support tables with bullets like this. However, given this suggestion I can refactor it into its own section.

specs/AppLifecycle/Restart/restartApi.markdown
hamza-usmani: This is a Win32 API (this section outlines the existing APIs): https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-registerapplicationrestart

specs/AppLifecycle/Restart/restartApi.markdown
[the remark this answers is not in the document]
hamza-usmani: This is something we've debated in the past. We chose to go this path for mainly 2 reasons:
1. Technically a restart is being "requested", it could fail. It may not actually complete successfully and there may be another restart already in progress/pending.
2. For parity/consistency with the UWP RequestRestartAsync

specs/AppLifecycle/Restart/restartApi.markdown
hamza-usmani: We'll get the RtCop output in a MIDL section in this.

specs/AppLifecycle/Restart/restartApi.markdown
hamza-usmani: Sure, good point.

specs/AppLifecycle/Restart/restartApi.markdown
jonwis: RECOMMEND: Jon to more carefully review before sending review.  Hamza to add MIDL3 for the feature.

specs/AppLifecycle/Restart/restartApi.markdown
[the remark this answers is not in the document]
jonwis: RECOMMEND: Update flavor text

specs/AppLifecycle/Restart/restartApi.markdown
jonwis: DISCUSSION: There's a servicing gap possible, but there's a race between any restart/servicing operation; there's already an API that does this.  Mostly it's to align WinAppSDK's CoreApp with UWP's features.  Adobe has specifically asked for this win32 app-restart feature, and it's not available (yet) on UWP's CoreApplication.

specs/AppLifecycle/Restart/restartApi.markdown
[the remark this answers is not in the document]
jonwis: DISCUSSION: WinAppSDK is explicitly not focusing on UWPs.

RECOMMEND: Transmute the table to just a paragraph.

DISCUSSION: If XAML were to call this API, and XAML was being used in its experimental UWP mode, what would happen? It would just work - the API fills the Win32 gap but calls into the platform CoreApplication as necessary.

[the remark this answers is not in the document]
jonwis: RECOMMEND: Update text - it's the state of a restarted app.

specs/AppLifecycle/Restart/restartApi.markdown
jonwis: RECOMMEND: We'll move it to a section, but generally we need to revamp the template to AVOID tables.

specs/AppLifecycle/Restart/restartApi.markdown
jonwis: DISCUSSION: It's OK, showing the MSDN content. Just name the API and link to MSDN with it.

specs/AppLifecycle/Restart/restartApi.markdown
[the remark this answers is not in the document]
jonwis: DISCUSSION: This is on the AppInstance itself, so it's to restart the instance with new args. The Request- prefix is to align better with the existing Windows.ApplicationModel.CoreApplication method, which needs permission from the user.  We'll do some investigation to figure out what the Request- operation actually did in the OS.  If this API is used in a UWP it will need to be on a UI thread.

RECOMMEND: Try calling it Restart, unless we find that the Request was integral to the original design.

specs/AppLifecycle/Restart/restartApi.markdown
jonwis: DISCUSSION: The app is just CreateProcess'd with the string, and your WinMain/wmain get the string to process. So you'd say "/restart" and the API will invoke "contoso.exe /restart"

specs/AppLifecycle/Restart/restartApi.markdown
jonwis: DISCUSSION: Similar to clicking the tile twice, the API-launched and user-launched instances will come up.  Both get the "previous state = terminated" value.  The user-launched one gets whatever the tile would have passed to launch the app, the API-launched one gets the command line.  The single-instancing APIs - also on this type, AppInstance - can help developers resolve this into a single target instance.

specs/AppLifecycle/Restart/restartApi.markdown
jonwis: DISCUSSION: There's a nonzero chance that the system couldn't process launch of the agent, or couldn't open the source process to terminate. As long as the API returns one of these codes, doom is imminent.

specs/AppLifecycle/Restart/restartApi.markdown
jonwis: DISCUSSION: yes, on success, the call does not return.

RECOMMEND: Clarify a sequence diagram

specs/AppLifecycle/Restart/restartApi.markdown
jonwis: DISCUSSION: The elevation state of the app is maintained in the restarted process.

RECOMMEND: Ensure the behaviors of elevation levels is documented. Also document how much of the source environment is flowed to the target.

DISCUSSION: If the source exe is marked delete-on-close, the agent could terminate that instance then call CreateProcess and get back a "file not found", etc. The caller gets back an "other" error. Ultimately the worst possible case is the app terminates and cannot be restarted, and the user goes "huh" then clicks the tile again.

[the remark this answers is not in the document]
jonwis: DISCUSSION: The API calls CreateProcess to launch the agent.  Suggest updating text to explain better.

specs/AppLifecycle/Restart/restartApi.markdown
jonwis: RECOMMEND: Fix

specs/AppLifecycle/Restart/restartApi.markdown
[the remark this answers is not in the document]
jonwis: RECOMMEND: Update text to reflect what's happening

specs/AppLifecycle/Restart/restartApi.markdown
jonwis: SAMPLE: Improve, like "/foo" or something.

jonwis: RECOMMEND: Yes, AppInstance (when you update the IDL it'll be clear.)

specs/AppLifecycle/Restart/restartApi.markdown
jonwis: SAMPLE: It will likely say "something went wrong".  There's not much to say other than "you will need to manually fix it, user."

specs/AppLifecycle/Restart/restartApi.markdown
jonwis: SAMPLE: Fix

jonwis: SAMPLE: Yes, like above. More realism.

specs/AppLifecycle/Restart/restartApi.markdown
jonwis: DISCUSSION: Once a request has been started there's no cancelling it.

specs/AppLifecycle/Restart/restartApi.markdown
jonwis: RECOMMEND: Remove, it's inside baseball here.

specs/AppLifecycle/Restart/restartApi.markdown
jonwis: RECOMMEND: Fix. s/Centennial/Desktop Bridge/gi

specs/AppLifecycle/Restart/restartApi.markdown
MikeHillberg: So update to explain that rather than referring to LaunchActivatedEventArgs?

aeloros: The command-line is exposed via LaunchActivatedEventArgs, the object is returned via the GetActivatedEventArgs() method on the same object this API is exposed from.

specs/AppLifecycle/Restart/restartApi.markdown
hamza-usmani: Done. Only thing not yet in is the RtCop output.

specs/AppLifecycle/Restart/restartApi.markdown
hamza-usmani: Fixed

specs/AppLifecycle/Restart/restartApi.markdown
hamza-usmani: Updated

specs/AppLifecycle/Restart/restartApi.markdown
hamza-usmani: Added

specs/AppLifecycle/Restart/restartApi.markdown
hamza-usmani: Added links

specs/AppLifecycle/Restart/restartApi.markdown
hamza-usmani: Moved to sections, removed tables

specs/AppLifecycle/Restart/restartApi.markdown
hamza-usmani: Removed

specs/AppLifecycle/Restart/restartApi.markdown
hamza-usmani: Fixed

hamza-usmani: Fixed

specs/AppLifecycle/Restart/restartApi.markdown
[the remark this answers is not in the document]
hamza-usmani: Updated

specs/AppLifecycle/Restart/restartApi.markdown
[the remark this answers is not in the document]
hamza-usmani: Updated

specs/AppLifecycle/Restart/restartApi.markdown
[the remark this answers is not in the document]
hamza-usmani: Updated

specs/AppLifecycle/Restart/restartApi.markdown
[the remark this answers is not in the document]
hamza-usmani: Updated

specs/AppLifecycle/Restart/restartApi.markdown
hamza-usmani: Updated

specs/AppLifecycle/Restart/restartApi.markdown
hamza-usmani: Added a few lines on this.

specs/AppLifecycle/Restart/restartApi.markdown
hamza-usmani: Fixed

specs/AppLifecycle/Restart/restartApi.markdown
hamza-usmani: Fixed

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -64,135 +69,177 @@ There are 2 related Win32 APIs that focus on recovery prior to restart:
...
+        ExtendedActivationKind Kind { get; };
+        IInspectable Data{ get; };
+    };
+
+    runtimeclass AppInstance
+    {
+        static AppInstance GetCurrent();
+        static Windows.Foundation.Collections.IVector<AppInstance> GetInstances();
+        static AppInstance FindOrRegisterForKey(String key);
+        static Windows.ApplicationModel.Core.AppRestartFailureReason Restart(String arguments);
```
jonwis, line 191: Consider putting a "this is the new one!" marker here. Also since this is a new API in 1.1, you'll need to add the proper contract information here, like:

```idl
[contractversion(1.1)] apicontract AppLifecycleContract{}

[contract(AppLifecycleContract, 1)]
runtimeclass AppInstance
{

    // On the new thing
    [contract(AppLifecycleContract, 1.1)]
    {
        static Windows.ApplicationModel.Core.AppRestartFailureReason Restart(String arguments);
    }
}
```

specs/AppLifecycle/Restart/restartApi.markdown
```diff
@@ -64,135 +69,177 @@ There are 2 related Win32 APIs that focus on recovery prior to restart:
...
-3. The Agent will get the executable path of the application and signal an ACK to the API. The Agent will then then terminate the application and then CreateProcess with the application (restarted).
-
-## Examples and Scenarios
+- InvalidUser: Could not restart for the specified user.
+- RestartPending: A restart is already in progress.
+- Other: Unspecified failure.
+
+### Mechanism
+
+When an application calls Restart, the mechanism for restart will be as follows:
```
jonwis, line 237: Markdown nit: Put a blank line before & after number/bullet lists.

jonwis (approved):
Please fix the two things, and run a word-wrapper (check out the docs/coding-guidelines.md for suggested tooling.)  Looks fine otherwise.

specs/AppLifecycle/Restart/restartApi.markdown
hamza-usmani: Done for all lists!

specs/AppLifecycle/Restart/restartApi.markdown
hamza-usmani: Added a marker for this.

On the note on contract: Is there a template for how/where the contract information belongs? Do I just drop it into the IDL as you've commented above?

andreww-msft (approved):

PR #1135, opened 2021-07-28
Add spec for MSIX Packages

DrusTheAxe:
Define the various MSIX packages composing Windows App SDK, including the new Singleton package


specs/Deployment/MSIXPackages.md
```diff
@@ -0,0 +1,189 @@
...
+                     a 'helper process' running at MediumIL and doing work on the API's behalf.
+                     cannot do for itself (i.e. brokered access to a resource).
+* Long-Running Process (aka user service) -- functionality requiring a long-running process
+                                             (especially if its lifetime isn't directly coupled to
+                                             an app's lifetime).
+
+Main packages don't support side-by-side behavior for a user. A user can only have 1 (or 0) Main
+packages in a package family registered at a time. Thus if a user has v1.0.0.0 registered when
+1.0.0.1 is registered it's an *update*, ending with the user having v1.0.0.1 registered instead of
+v1.0.0.0. This poses servicing challenges, as updating the Main package requires any running
```
andrewleader, line 57: What happens when an app built with 1..0.0.1 is trying to be installed, and the current Main package is 1.0.0.0, and there's currently an app open that uses Main package? What kind of error will the user get when they try to install? This seems slightly worrisome as we get more and more apps? How can developers "only rely on Main package when necessary", will it somehow be detected that an app isn't actively calling APIs that use Main even though the app itself requires Main?

I'd love to understand the user impact a bit more :) Great details already though!

specs/Deployment/MSIXPackages.md
```diff
@@ -0,0 +1,189 @@
...
+int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, PWSTR pCmdLine, int nCmdShow)
+{
+    return MyFeatureWinMain(hInstance, hPrevInstance, pCmdLine, nCmdShow);
+}
+```
+
+and export `MyFeatureWinMain` from a DLL in the Framework package to get the implementation's
+other 99%.
+
+### Role - Singleton
```
huichen123, line 85: is this per user?

specs/Deployment/MSIXPackages.md
```diff
@@ -0,0 +1,189 @@
...
+
+The Singleton package supplements the Main package, for when functionality is needed by a single
```
andrewleader, line 87: Any more details on Singleton? Maybe list the push notifications scenario.

specs/Deployment/MSIXPackages.md
```diff
@@ -0,0 +1,189 @@
...
+* Minor -- minor version of the release, e.g. "0" for Windows App SDK 1.0
+* VersionTag -- optional version tag to distinguish amongst channels and revs of a channel, e.g. "-preview2" for Windows App SDK 1.0 Preview 2
+
+### Naming - SubName
+
+The following SubName values are used:
+
+| SubName | Package | Example |
+|-|-|-|
+| | Framework | Microsoft.WindowsAppRuntime.1.0-experimental1 |
```
huichen123, line 127: not exp1? char count keeps increasing

specs/Deployment/MSIXPackages.md
```diff
@@ -0,0 +1,189 @@
...
+have registered...
+
+* Microsoft.WindowsAppRuntime.1.0
+* Microsoft.WindowsAppRuntime.1.1
+* Microsoft.WindowsAppRuntime.2.0
+* Microsoft.WindowsAppRuntime.Main.1.0
+* Microsoft.WindowsAppRuntime.Main.1.1
+* Microsoft.WindowsAppRuntime.Main.2.0
+* Microsoft.WindowsAppRuntime.Singleton (version 2.0)
+* Microsoft.WindowsAppRuntime.DDLM.1.0-x8
```
huichen123, line 150: what's the reason for "x8" instead of "x86"? not intuitive with one char saving.

specs/Deployment/MSIXPackages.md
DrusTheAxe: > What happens when an app built with 1..0.0.1 is trying to be installed, and the current Main package is 1.0.0.0, and there's currently an app open that uses Main package?

Very much depends on how features are designed and implemented, how they interact with Main package content, how termination and (re)start of those processes is handled and how all that interacts with applications (if at all).

The installer (currently) calls `PackageManager.AddPackageAsync(...,DeploymentOptions::None)`. Thus running Main processes won't be interrupted and the Main package won't be updated. For those who skip the installer and directly install the MSIX packages via `PackageManager` it depends if they do likewise or specify a 'force' (e.g. `DeploymentOptions::ForceTargetApplicationShutdown`). In the latter case it's no different than if you update Calculator or other packages while their app(s) are running - if they're implemented to provide Suspend/Resume semantics it looks to a user like the app disappears then reappears. YMMV of course.

>This seems slightly worrisome as we get more and more apps?

It's a pickle no matter what we do. If no Force option is specified then running apps are uninterrupted but a main package won't be updated. OTOH if a Force option's specified it may affect running applications. In the end this isn't all that different from installing various applications, some will interrupt functionality, require me to quit apps to unblock an uninstall, require me to reboot, etc. Servicing is a hard problem space to begin with, doing so generically for all possible software even harder, and doing it seamlessly with no impact is the holy grail of servicing. Balancing those factors along (without servicing Windows downlevel) makes for an imperfect equation.

This is one of the reasons we strongly encourage Windows App SDK developers to put content into the Framework package as it has no such servicing headaches.

>How can developers "only rely on Main package when necessary"

That's guidance to Windows App SDK developers (users of Windows App SDK won't care, to them it's just how the sausage is works).

Some features can accomplish their ends with different designs; even if harder / higher cost to implement a feature it's often the better solution in the end. We've seen some cases it's simply not practical (or even possible) to avoid; in such necessary cases, it's available.

specs/Deployment/MSIXPackages.md
DrusTheAxe: Yes. All MSIX packages are registered per-user.

The installer will also provision all the packages if it can (i.e. if it's running with admin privilege i.e. elevated), but provisioning just makes packages available to all users. That doesn't change the core principle of MSIX as a per-user app model.

We have feature today needing machine-wide singleton semantics. We may in the future. We also have more tools at our disposal we haven't needed to leverage (so far) e.g. Packaged Services. We've considered the issue but lacking a driving need we haven't seen the need to drive any sort of stake in the ground how we'll implement said unknown theoretical future features :-)

specs/Deployment/MSIXPackages.md
riverar: AppExtension name is limited to <= 39 characters on builds less than 18307 and the WAS min target is RS5 17763. When the min moves forward, presumably all this can be refactored away.

specs/Deployment/MSIXPackages.md
```diff
@@ -0,0 +1,189 @@
...
+This is accomplished for the Microsoft.WindowsAppRuntime framework package via a Packaged COM
+out-of-process (OOP) server in a DDLM package, selected and managed via the Bootstrap API. See the
+[Dynamic Dependency spec](https://github.com/microsoft/WindowsAppSDK/blob/main/specs/dynamicdependencies/DynamicDependencies.md)
+for more details.
+
+## Naming
+
+Windows App SDK's MSIX packages use the following naming rules for package identity:
+
+* Name = Microsoft.WindowsAppRuntime[.SubName].<Major>.<Minor>[-VersionTag]
```
riverar, line 111: Version tags are missing when the Markdown is rendered.

```suggestion
* Name = Microsoft.WindowsAppRuntime[.SubName].\<Major>.\<Minor>[-VersionTag]
```

specs/Deployment/MSIXPackages.md
DrusTheAxe: Length. A package's Name is limited to <= 50 characters, so DDLM is cramped for space. I see the detailed naming rules aren't posted; let me dig those up.

And actually the DDLM here is wrong. DDLM names are actually `Microsoft.ProjectReunion.DDLM-<version>-<shortarchitecture>[-shortversiontag]` where `<version>` is the full MSIX version (not the project's major.minor release numbering). For instance, even though we released the  `Microsoft.ProjectReunion.0.8-preview1` framework package the associated x64 DDLM package had a Name of `Microsoft.ProjectReunion.DDLM.8000.146.628.0-x6-p`. 

(Soon to be renamed for WindowsAppRuntime, though DDLM will need be Microsoft.WinAppRuntime.DDLM... to fit within name limits)

specs/Deployment/MSIXPackages.md
DrusTheAxe: Gah! Thanks!

specs/Deployment/MSIXPackages.md
DrusTheAxe: Channel names are Stable, Preview and Experimental.
Their version tags are "" (none), "-preview" and "-experimental"
They also can have a numeric suffix e.g. "-experimental1".

That's also why we need the 'shortversiontag' e.g. -e1 == -experimental1.

And yes, names keep getting longer since 0.1 days. But so too do our requirements so it shouldn't be too surprising ;-)

I went through the exhaustive exercise to design a naming scheme that lets MSIX packages meet our versioning principles and policies while also working within the bounds of the available technology. That's also why DDLM packages have unusual names (but they also pose unusual requirements). Too many details not mentioned here but already well defined (been following those rules since 0.8-preview). Let me add that information.

specs/Deployment/MSIXPackageVersioning.md
```diff
@@ -0,0 +1,511 @@
...
+WinUI 2.x encodes MSIX versions as `M.NYYMM.DDPPP.0` (same as Windows App SDK 0.5). This runs into issues starting with WinUI 2.7, as identified by [WinUI version schema sometime can't applied to MinVersion #4008](https://github.com/microsoft/microsoft-ui-xaml/issues/4008#issuecomment-765633788).
+
+## 2.5. Decision 5: Package Names
+
+Windows App SDK 0.8+ has 3 MSIX packages. We have choices for package Name patterns for 0.8 and 1.x.
+
+Assuming Breaking Change Boundary for v0.x and v1+'s decision are `Major.Minor` version we have the
+following naming patterns...
+
+* WASfwk: `Microsoft.WindowsAppRuntime.<rmajor>.<rminor>[-tag]`
```
niklasb-ms, line 400: How do <rmajor> and <rminor> differ from <major> and <minor>?

andrewleader:
Ok so I need your help understanding the servicing/updating aspects of the Main and Singleton packages some more 😊

1) If a packaged app is using WinAppSDK, how will that impact updateability of the Main/Singleton packages? Is the Main/Singleton locked the entire time the app is open? Or only when the app is calling a specific API that calls into Main, it then temporarily locks Main till that API returns?

2) If an unpackaged app uses WinAppSDK and uses the default way (the installer), today it sounds like the Main package will be locked the entire time the app is open?

specs/Deployment/MSIXPackageVersioning.md
DrusTheAxe: rmajor+rminor are the Release version, e.g. 0 nd 8 for what was ProjectReunion 0.8
major+minor are the 1st 2 parts of the MSIX package version

I see rmajor+rminor are missing from the bulleted list @ line 406. Thought I'd added that. Fixing

DrusTheAxe:
>Is the Main/Singleton locked the entire time the app is open?

No. Only if there's a process running from the Main package. Those are expected to be short-lived and/or support suspend/resume semantics. For instance, DynDep has a short-lived helper process from the Main packge that runs and quits during one of its method calls.

Singleton's more likely to have long-running processes. This only came up the last few days so we haven't worked through the fine print yet how we're going to handle all the servicing scenarios. My hunch, the installer will add --force options so the caller can plow the earth if anything's in its way but is potentially impactful, or be like today's non-disruptive no-force behavior.

We also need to write up the guidance for folks directly installing the packages (not using the installer).


> 2. If an unpackaged app uses WinAppSDK and uses the default way (the installer), today it sounds like the Main package will be locked the entire time the app is open?

No, as we have no long-lived processes in the Main package today.
The Singleton package is a different matter (if we do nothing beyond today. But we will :-)

It's all do-able, but there's a few decisions we have to make in the installer's flow and developer guidance. In progress.

dhoehna (approved):

andrewleader (approved):

specs/Deployment/MSIXPackageVersioning.md
```diff
@@ -0,0 +1,517 @@
...
+Option C is too limiting given it can't support Patch versions beyond 6.
+
+Option D relies on the package Name including Major version. This leaves sufficient space in the
+version to encode up to 99 Minor and 99 Patch versions. Date encoding as an Epoch reduces the size
+while supporting 20+ years with only 4 digits.
+
+Option E is a variant of Option D except Patch version is restricted to 2 digits, but it's an
+unnecessary simplification.
+
+### 2.4.1. Windows App SDK 0.5 Notation
```
jonwis, line 367: Is this here for historical reasons? We're past 0.8 at this point.

specs/Deployment/MSIXPackageVersioning.md
```diff
@@ -0,0 +1,517 @@
...
+All fields are fixed-size with leading zeroes as needed. For example, Windows App SDK 0.5.0's MSIX package has a version of `0.52103.25000.0`
+
+* Major = 0
+* Minor = 5
+* Year = 21 (aka 2021)
+* Month = 03
+* Day = 25
+* Patch = 000 (if we have a 0.5.1 release this would be `001`)
+
+This doesn't scale going forward e.g. MSIX version is a DotQuadNumber aka `uint16[4]`. If 0.8 was released on May'31, 2021 MSIX version would be `0.82105.31000.0`
```
jonwis, line 387: Consider using the Prettier VSCode extension to word-wrap.

specs/Deployment/MSIXPackageVersioning.md
```diff
@@ -0,0 +1,517 @@
...
+
+### 2.4.2. WinUI 2.x Notation
+
+WinUI 2.x encodes MSIX versions as `M.NYYMM.DDPPP.0` (same as Windows App SDK 0.5). This runs into issues starting with WinUI 2.7, as identified by [WinUI version schema sometime can't applied to MinVersion #4008](https://github.com/microsoft/microsoft-ui-xaml/issues/4008#issuecomment-765633788).
+
+## 2.5. Decision 5: Package Names
+
+Windows App SDK 0.8+ has 3 MSIX packages. We have choices for package Name patterns for 0.8 and 1.x.
```
jonwis, line 395: We're done with 0.8, yes?

specs/Deployment/MSIXPackageVersioning.md
```diff
@@ -0,0 +1,517 @@
...
+|Package|Average|AverageLength|
+| --- | :--- | :---: |
+|WARfwk |Microsoft.WindowsAppRuntime.1.15-preview1|41|
+|WARmain|Microsoft.WindowsAppRuntime.Main.1.15-preview1|46|
+|WARmain|Microsoft.WindowsAppRuntime.Singleton-preview1|46|
+|WARddlm|Microsoft.WindowsAppRuntime.DDLM.1.15.12345.24680-arm64-preview1|**<span style="color:red">64</span>**|
+
+|Package|Min|MinLength|
+| --- | :--- | :---: |
+|WARfwk |Microsoft.WindowsAppRuntime.1.0-preview1|40|
```
jonwis, line 438: Should we use `WindowsAppRT` (even though it has an unfortunate acronym and association with previous OS efforts) here?

specs/Deployment/MSIXPackageVersioning.md
```diff
@@ -0,0 +1,517 @@
...
+for more details.
+
+This leads to package Name length issues even for common cases:
+
+|Package|Average|AverageLength|
+| --- | :--- | :---: |
+|WARfwk |Microsoft.WindowsAppRuntime.1.15-preview1|41|
+|WARmain|Microsoft.WindowsAppRuntime.Main.1.15-preview1|46|
+|WARmain|Microsoft.WindowsAppRuntime.Singleton-preview1|46|
+|WARddlm|Microsoft.WindowsAppRuntime.DDLM.1.15.12345.24680-arm64-preview1|**<span style="color:red">64</span>**|
```
jonwis, line 434: Could all these bits of data (versions, architecture, preview-ness) be combined into one large 128-bit value and then converted back via base-32 (a-z0-9 minus a few suspect characters like 'u' and 'c' and 's') into a single very short token? It impacts readability but moots the size concerns. This isn't a string that developers should have to manage anyhow other then copypasting it around.

Can "DDLM" shorten to "DD"?

specs/Deployment/MSIXPackageVersioning.md
```diff
@@ -0,0 +1,517 @@
...
+* Encode tag in the delimiter between version+architecture e.g. Microsoft.WindowsAppRuntime.DDLM.1.0.0.0<span style="color:red"><b>p1</b></span>arm64 using P1 for Preview1, E1=Experimental1, ...
+* Name WARddlm differently from WARfwk and WARmain e.g. use "-p1" for WARddlm regardless if WARfwk and WARmain use "-preview1"
+* ???
+
+WARddlm is needed until Dynamic Dependencies is 100% based on OS DynDep.
+
+Best case (!) WARddlm exists until the minimum supported Windows release is Cobalt.
+
+***Recommendation:*** The general package naming syntax is
+`Microsoft.ProjectReunion[.SubName]-<rmajor>.<rminor>[-tag]`. WARddlm and WARsingleton are
```
jonwis, line 466: Old Reunion name here

specs/Deployment/MSIXPackages.md
```diff
@@ -0,0 +1,217 @@
...
+[background tasks](https://docs.microsoft.com/windows/uwp/launch-resume/create-and-register-a-background-task),
+[startup tasks](https://docs.microsoft.com/uwp/api/Windows.ApplicationModel.StartupTask) executing
+on user login, [ApplicationData](https://docs.microsoft.com/uwp/api/Windows.Storage.ApplicationData)
+for data storage scoped to Windows App SDK (not any particular app using Windows App SDK), and
+more.
+
+The primary scenarios include:
+
+* Brokered Access -- processes running in an [AppContainer](https://docs.microsoft.com/windows/win32/secauthz/appcontainer-isolation) requiring
+                     access to resources that cannot be access from within the AppContainer can use
```
jonwis, line 57: Reformat doc using prettier, please

specs/Deployment/MSIXPackages.md
```diff
@@ -0,0 +1,217 @@
...
+                                             (especially if its lifetime isn't directly coupled to
+                                             an app's lifetime).
+
+Main packages don't support side-by-side behavior for a user. A user can only have 1 (or 0) Main
+packages in a package family registered at a time. Thus if a user has v1.0.0.0 registered when
+1.0.0.1 is registered it's an *update*, ending with the user having v1.0.0.1 registered instead of
+v1.0.0.0. This poses servicing challenges, as updating the Main package requires any running
+processes from the existing Main package to be terminated.
+
+For these reasons developers are strongly encouraged to only rely on the Main package when
```
jonwis, line 70: Seems important to note who "developers" means here. I wouldn't expect Contoso to know whether the WinAppSDK features are implemented in the FWK or the Main packages, nor should it impact them other than how they choose to deliver an unpackaged app.

specs/Deployment/MSIXPackages.md
```diff
@@ -0,0 +1,217 @@
...
+necessary.
+
+If a process from the Main package is necessary, it should be short-lived or support a save/restore
+execution model. The former shrinks the possibility a running process when a servicing event
+occurs. The latter ensures if Main package processes are running when a servicing event occurs, the
+running processes can be terminated (e.g. via [DeploymentOptions.ForceTargetApplicationShutdown](https://docs.microsoft.com/uwp/api/windows.management.deployment.deploymentoptions))
+to allow the servicing event to occur and restarted upon its completion, with only a transient (and
+possibly unnoticed) interruption of service.
+
+Further, if/when the Main package is necessary, developers are encouraged to put as
```
jonwis, line 80: Again, note who "developers" is - it's people writing WinAppSDK features? In which case the guidance is right, but maybe provide a little more guidance on how to do this.   Like, "put your resource access requests in the main package, but put the remainder of the implementation in the FWK" for instance.

jonwis (changes requested):

BenJKuhn (approved):

specs/Deployment/MSIXPackageVersioning.md
```diff
@@ -0,0 +1,511 @@
...
+    - [2.4.2. WinUI 2.x Notation](#242-winui-2x-notation)
+  - [2.5. Decision 5: Package Names](#25-decision-5-package-names)
+- [3. Decision Conclusion](#3-decision-conclusion)
+
+# 1. Background
+
+This document answers the question: "How are MSIX and Dynamic Dependencies defined and implemented
+to meet versioning requirements for Windows App SDK 0.8 through 1.x?"
+
+Windows App SDK's versioning requirements beyond 0.5 are defined in (insert link here).
```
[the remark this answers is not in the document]
DrusTheAxe: @andrewleader @EHO-makai got link?

specs/Deployment/MSIXPackageVersioning.md
```diff
@@ -0,0 +1,517 @@
...
+**Option C: `-p`**
+
+These all may optionally include a trailing digit indicating the Nth release of a channel for a
+version of the project.
+
+***Recommendation:*** Option A and C (when necessary).
+
+All tags have a long name (Option A) and a short (1-2 characters) name (Option C).
+
+Option C is used when Option A doesn't work due to package Name length restrictions.
```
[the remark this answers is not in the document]
DrusTheAxe: The short -p is only needed for the DDLM; the longer form works for everyone else (Fwk, Main and Singleton).

The DDLM naming is a special snowflake in other ways too (full MSIX version a.b.c.d not product release major.minor, includes architecture in its name). So low value in using the short name for everyone else too.

specs/Deployment/MSIXPackageVersioning.md
DrusTheAxe: Yes, historical. That's why it's in a sub-section and not part of any option for consideration. We started with that, quickly saw the limitations and gamed out alternatives to land on our current solution. But knowing this history helps many to understand "Why is it so complicated? Why can't it be simple?".

specs/Deployment/MSIXPackageVersioning.md
DrusTheAxe: Name += version.ToBase32() is possible but we still need to fill in the MSIX package's version as a DotQuadNumber in `appxmanifest.xml`. Base32 doesn't help us there.

Hashing was briefly considered then quickly discarded, as Deployment considers version relatively significant and won't install a lower version if a higher version's already installed.

Treating version as a uint128 was considered with data spanning byte boundaries but it buys us little and degrades usability (developers routinely look at version numbers during support and troubleshooting. A sea of numbers in logs are helpful if not totally gibberish).

The current scheme works reasonably well, balancing the information and identity needs with usability and not-hostile to devs. And this isn't a 'forever' issue. Win11 has the Dynamic Dependencies API (w/o need a the DDLM helper) so someday when that's the oldest supported version the DDLM goes away. Granted, that's years out at least, but someday.

specs/Deployment/MSIXPackageVersioning.md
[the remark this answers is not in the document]
DrusTheAxe: I'll follow up with @andrewleader and @EHO-makai to get this updated in a future PR

specs/Deployment/MSIXPackageVersioning.md
DrusTheAxe: Yes

specs/Deployment/MSIXPackageVersioning.md
DrusTheAxe: No. We can make the naming work, no need to raise shades of times best forgot 😋

PR #1146, opened 2021-07-29
Disable MRTCore managed test

huichen123:
After the build pool switched to the private pool, MRTCore managed test started to fail due to some configuration issue. However, it didn't cause build break. Now it breaks the build in release branch. Disable it in release branch for now.

### DIFF
--- a/build/build-mrt.yml
+++ b/build/build-mrt.yml
@@ -123,17 +123,17 @@ steps:
     PathtoPublish: $(Build.SourcesDirectory)/mrtcore.$(buildPlatform).$(buildConfiguration).binlog
     artifactName: binlogs
 
-- task: VSTest@2
-  displayName: 'test MRT (Managed)'
-  condition: and(succeeded(), or(eq(variables['buildPlatform'], 'x86'), eq(variables['buildPlatform'], 'x64')))
-  inputs:
-    testSelector: 'testAssemblies'
-    testAssemblyVer2: |
-      **\ManagedTest.build.appxrecipe
-    searchFolder: '${{ parameters.MRTBinariesDirectory }}\Release\$(buildPlatform)'
-    testRunTitle: 'test MRT $(buildPlatform)'
-    platform: '$(buildPlatform)'
-    configuration: '$(buildConfiguration)'
+#- task: VSTest@2
+#  displayName: 'test MRT (Managed)'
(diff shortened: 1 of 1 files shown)


axelandrejs (approved):

axelandrejs:
Can you file a ticket to reenable this?

jefgen:
Just curious, is there a reason why the `main` branch doesn't use the same pool?
It looks like `main` uses the regular Azure pipelines and Release is using `ProjectReunionESPool`.
Using different pools seems like it might lead to things getting out of sync.

huichen123:
> Just curious, is there a reason why the `main` branch doesn't use the same pool?
> It looks like `main` uses the regular Azure pipelines and Release is using `ProjectReunionESPool`.
> Using different pools seems like it might lead to things getting out of sync.

i think main is also using private pool.

jefgen:
> i think main is also using private pool.

The most [recent build](https://dev.azure.com/ms/ProjectReunion/_build/results?buildId=204789&view=logs&j=0f8a1c78-354a-516e-289c-92301b7b43b4&t=7604646d-2f1a-47a9-b735-ec19335aedd9) for the recent commit on `main` on GitHub looks like it is running on `Agent name: 'Azure Pipelines 3'`

Whereas the `release/*` branch [build](https://microsoft.visualstudio.com/ProjectReunion/_build/results?buildId=37313406&view=logs&j=0f8a1c78-354a-516e-289c-92301b7b43b4) looks to be using `ProjectReunionESPool`.

Perhaps maybe because it runs as a different pipeline? (`ProjectReunion-Foundation.Transport` instead of `microsoft.ProjectReunion`).

jefgen:
Looks like only the "`ProjectReunion-Foundation.Transport`" pipeline was updated to use the private pool (with https://github.com/microsoft/WindowsAppSDK/pull/1031).

Which means that the CI builds on `main` are building using a different pool...
(I'm not sure if this is expected or maybe `main` just hasn't moved over yet?)

jefgen:
FYI: issue filed here: #1148 (thanks Hui!)

huichen123:
> Looks like only the "`ProjectReunion-Foundation.Transport`" pipeline was updated to use the private pool (with #1031).
> 
> Which means that the CI builds on `main` are building using a different pool...
> (I'm not sure if this is expected or maybe `main` just hasn't moved over yet?)

the private pool only applies to the internal repo. Branch doesn't matter. The PR/CI build will still use shared pool.

PR #1149, opened 2021-07-30
Adding BuildAndPublishInsights.yml to CI.yml so that the headers and libs are included in the CI builds.

ZenBird:


riverar:
Will this build for folks external like me? That nuget package source does not work for us.

```xml
<add key="WindowsTelemetry-Internal" value="https://microsoft.pkgs.visualstudio.com/_packaging/0/nuget/v3/index.json" />
```

ZenBird:
Are you not able to build in Visual studio? If that is the case we may have to look for a workaround.

---
In reply to: [893986122](https://github.com/microsoft/WindowsAppSDK/pull/1149#issuecomment-893986122) [](http://example.com/codeflow?ancestors=893986122,893986122)

kythant (approved):

PR #1152, opened 2021-07-30
Move PushNotifications overrides out of GenerateDynamicDependenciesOverrides.ps1

DrusTheAxe:
Don't put non-DynamicDependencies concerns into a DynamicDependencies file

Move PushNotifications' logic in Generate**DynamicDependencies**Overrides.ps1 to its own file+task


pmpurifoy:
Thanks for separating the files into their own separate scripts!

pmpurifoy (approved):

PR #1159, opened 2021-08-02
Adding an explicit reference to resources winmd causes invalid appxmanifest.xml build error

Scottj1s:
This logic appears to be an early experiment in getting activation to work.  It now causes an invalid appxmanifest.xml build error, with the Implementation element containing a noncanonical path.  That could be fixed, but there doesn't appear to be a need for this anyway.  C# apps don't even use this logic, and C++ works without it.

Error:
...\AppxManifest.xml : error APPX0501: Validation error. error C00CE169: App manifest validation error: The app manifest must be valid as per schema: Line 40, Column 10, Reason: 'd:\samples\WindowsAppSDKInAppCpp\packages\Microsoft.WindowsAppSDK.Foundation.0.9.0-20210722.17.CI\build\native\..\..\runtimes\win10-x64\native\Microsoft.WindowsApp.Resources.dll' violates pattern constraint of '([^/\\]*[^./\\]+)(\\[^/\\]*[^./\\]+)*|([^/\\]*[^./\\]+)(/[^/\\]*[^./\\]+)*'. The element '{http://schemas.microsoft.com/appx/manifest/foundation/windows10}Path' with value 'd:\samples\Window
sAppSDKInAppCpp\packages\Microsoft.WindowsAppSDK.Foundation.0.9.0-20210722.17.CI\build\native\..\..\runtimes\win10-x64\native\Microsoft.WindowsApp.Resources.dll' failed to parse.


huichen123 (approved):

rohanp-msft:
Is the WinMD just automatically referenced without an explicit reference addition in the props file?

Scottj1s:
> Is the WinMD just automatically referenced without an explicit reference addition in the props file?

The winmd is not automatically referenced, and that doesn't appear necessary.  I'm not sure why this winmd in particular was given special treatment.  But it's definitely interfering with PR builds.

rohanp-msft:
> > Is the WinMD just automatically referenced without an explicit reference addition in the props file?
> 
> The winmd is not automatically referenced, and that doesn't appear necessary. I'm not sure why this winmd in particular was given special treatment. But it's definitely interfering with PR builds.

I can test to see if MRT Core works without the WinRT props file. Let me do that now (unless you've already done that).

rohanp-msft (approved):
:shipit:

build/NuSpecs/WindowsAppSDK-Nuget-Native.targets
```diff
@@ -30,4 +30,11 @@
     </Reference>
   </ItemGroup>
 
+  <ItemGroup>
+    <Reference Include="$(MSBuildThisFileDirectory)..\..\lib\native\Microsoft.ApplicationModel.Resources.winmd">
+      <Private>false</Private>
+      <Implementation>Microsoft.ApplicationModel.Resources.dll</Implementation>
+    </Reference>
+  </ItemGroup>
```
huichen123, line 38: if nuget doesn't handle this automatically, i am afraid WinUI may need some change to ingest.

Scottj1s: good point - will check that

PR #1162, opened 2021-08-02
Add IsActivatorSupported to PushNotifications

pmpurifoy:
An API for both packaged/unpackaged apps to call before registering an activator type for their app to decide which activator type to use to gain access to push notifications.


dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -245,4 +245,36 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        return true;
+    }
+
+    bool PushNotificationManager::IsActivatorSupported(PushNotificationRegistrationOptions const& options)
+    {
+        bool isBIFlagSet = WI_IsAnyFlagSet(options, PushNotificationRegistrationOptions::PushTrigger | PushNotificationRegistrationOptions::ComActivator);
+        bool isProtocolActivatorSet = WI_IsFlagSet(options, PushNotificationRegistrationOptions::ProtocolActivator);
+
+        THROW_HR_IF(E_INVALIDARG, isBIFlagSet && isProtocolActivatorSet); // Invalid flag combination
+        THROW_HR_IF(E_INVALIDARG, !(isBIFlagSet || isProtocolActivatorSet)); // null flags
```
hulumane, line 269: Hmm is this second state even possible? I thought the enum forced you to select atleast one activator type. If so, can we remove this second check?

pmpurifoy: Someone could do
`PushNotificationRegistrationOptions option;
 IsActivatorSupported(option)`
And option would init as 0

hulumane: Then why not check for 0 in the beginning of the function?

hulumane: Even better, we could set Undefined to 0 and check for that in the beginning of the function.

test/PushNotificationTests/APITests.cpp
```diff
@@ -203,15 +203,15 @@ namespace Test::PushNotifications
             RunTest(L"MultipleChannelClose", channelTestWaitTime());
         }
 
-        TEST_METHOD(MultipleChannelRequestUsingSameRemoteId)
-        {
-            RunTest(L"MultipleChannelRequestUsingSameRemoteId", channelTestWaitTime());
-        }
+        //TEST_METHOD(MultipleChannelRequestUsingSameRemoteId)
```
danielayala94, line 206: Why are we commenting out these tests?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -245,4 +245,36 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
             LOG_IF_FAILED(::CoRevokeClassObject(static_cast<DWORD>(token.Cookie())));
         }
     }
+
+    bool PushNotificationManager::IsBIAvailable()
```
DrusTheAxe, line 249: BI?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -245,4 +245,36 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+    {
+        try
+        {
+            BackgroundTaskBuilder builder = BackgroundTaskBuilder();
+            auto builder5 = builder.as<winrt::IBackgroundTaskBuilder5>();
```
DrusTheAxe, line 254: NIT: [ES.23: Prefer the {}-initializer syntax](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#es23-prefer-the--initializer-syntax)

```
auto builder{ BackgroundTaskBuilder() };
auto builder5{ builder.as<...>(); };
```

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -245,4 +245,36 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        }
+        catch (...)
+        {
+            return false;
```
DrusTheAxe, line 258: Silently swallowing errors is rarely wise. At least a LOG_ERROR_MSG() here so there's a trail if something goes wrong.

Imagine you have to support this code, a year from now. In the wild.
Imagine a production system is affected. There is no debugger or or invasive process.
Imagine this fails. Users don't know, other than the end behavior just doesn't happen.
What do you do?

Logs are your best friend. Learn to love them :-)

test/PushNotificationTests/APITests.cpp
DrusTheAxe, line 206: Please don't check in dead code

pmpurifoy: These tests take a little longer to run so I commented them out to quickly test IsActivatorSupported.

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: This is a scenario that we support where packaged apps don't have access to IBackgroundTaskBuilder5 for builds RS5 or below. Instead, they will call RegisterActivator with the ProtocolFlag to be launched by our LRP.

Because this is a scenario we support, should we LOG_ERROR_MSG() still?

test/PushNotificationTests/APITests.cpp
danielayala94: Please follow up with @sharath2727 about these tests.

dev/PushNotifications/PushNotificationManager.cpp
hulumane: Is there a way to perform a QI that doesn't throw exceptions? I think that would be ideal here. 

I do agree with the notion that it is strange pattern to rely on exceptions for cases like this.

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: BI = Broker Infrastructure, we check if BI supports IBackgroundTaskBuilder5 but I can rename this to IsBackgroundTaskBuilder

dev/PushNotifications/PushNotificationManager.cpp
hulumane: Ok did some digging. Instead of .as, you can use try_as. It will return nullptr if the QI fails and you can check against that instead of relying on an exception. Cleaner code.

test/TestApps/PushNotificationsTestApp/main.cpp
```diff
@@ -312,14 +365,16 @@ int main() try
...
 
     PushNotificationActivationInfo info(
         PushNotificationRegistrationOptions::PushTrigger | PushNotificationRegistrationOptions::ComActivator,
         winrt::guid(c_comServerId)); // same clsid as app manifest
 
-    g_appToken = PushNotificationManager::RegisterActivator(info);
+    if (Test::AppModel::IsPackagedProcess())
+    {
+        g_appToken = PushNotificationManager::RegisterActivator(info);
+    }
```
hulumane, line 376: Why didn't you use the IsSupported method here instead and branch the code out based on that?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -245,4 +245,37 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        THROW_HR_IF(E_INVALIDARG, options == PushNotificationRegistrationOptions::Undefined); // WI_IsFlagSet does not allow 0 as flag.
+
+        bool isBIFlagSet{ WI_IsAnyFlagSet(options, PushNotificationRegistrationOptions::PushTrigger | PushNotificationRegistrationOptions::ComActivator) };
+        bool isProtocolActivatorSet{ WI_IsFlagSet(options, PushNotificationRegistrationOptions::ProtocolActivator) };
+
+        THROW_HR_IF(E_INVALIDARG, isBIFlagSet && isProtocolActivatorSet); // Invalid flag combination
+
+        if (AppModel::Identity::IsPackagedProcess() && IsBackgroundTaskBuilderAvailable())
+        {
+            return isBIFlagSet;
```
hulumane, line 274: Probably shouldn't call this BI. Just call it isBackgroundTaskFlagSet

test/TestApps/PushNotificationsTestApp/main.cpp
pmpurifoy: I wanted to test the IsSupported method just by itself.

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: We are checking if we have support for IBackgroundTaskBuilder5 in the current build. Renamed to IsBackgroundTaskAvailable.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -245,4 +245,31 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        THROW_HR_IF(E_INVALIDARG, options == PushNotificationRegistrationOptions::Undefined); // WI_IsFlagSet does not allow 0 as flag.
+
+        bool isBackgroundTaskFlagSet{ WI_IsAnyFlagSet(options, PushNotificationRegistrationOptions::PushTrigger | PushNotificationRegistrationOptions::ComActivator) };
+        bool isProtocolActivatorSet{ WI_IsFlagSet(options, PushNotificationRegistrationOptions::ProtocolActivator) };
+
+        THROW_HR_IF(E_INVALIDARG, IsBackgroundTaskBuilderAvailable && isProtocolActivatorSet); // Invalid flag combination
+
+        if (AppModel::Identity::IsPackagedProcess() && IsBackgroundTaskBuilderAvailable())
+        {
+            THROW_HR_IF(E_INVALIDARG, isProtocolActivatorSet); // Don't allow packaged apps to call with protocol if background task available
```
hulumane, line 267: Why throw? Instead simply return false because protocol activation is not supported if Background task is available.

test/TestApps/PushNotificationsTestApp/main.cpp
pmpurifoy: This is done via scope_exit we define at the beginning of main. This block also handles cleaning up bootstrap.

hulumane: Then in that case, the scope exit should be right below the RegisterActivator.

hulumane: Also why isn't there a Protocol activator registration here. Is Main running only for packaged apps or will this implementation be shared by both packaged and unpackaged?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -245,4 +245,31 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        auto builder5{ builder.try_as<winrt::IBackgroundTaskBuilder5>() };
+        return builder5 != nullptr;
+    }
+
+    bool PushNotificationManager::IsActivatorSupported(PushNotificationRegistrationOptions const& options)
+    {
+        THROW_HR_IF(E_INVALIDARG, options == PushNotificationRegistrationOptions::Undefined); // WI_IsFlagSet does not allow 0 as flag.
+
+        bool isBackgroundTaskFlagSet{ WI_IsAnyFlagSet(options, PushNotificationRegistrationOptions::PushTrigger | PushNotificationRegistrationOptions::ComActivator) };
+        bool isProtocolActivatorSet{ WI_IsFlagSet(options, PushNotificationRegistrationOptions::ProtocolActivator) };
```
sharath2727, line 261: Optional: auto?

test/TestApps/PushNotificationsTestApp/main.cpp
pmpurifoy: It's going to be shared by packaged/unpackaged. But we can consider splitting the test app into two separate apps.

hulumane: A lot of odd things going on here in main IMHO:
1) I feel that Registration should be defined by the test as part of LaunchArgs. Automatically doing this for every test seems a bit of an anti-pattern for testing.

2) If we are running code for both packaged and unpackaged, it stands to reason that we cover both registration flows as part of this. Same for Unregister.

pmpurifoy: We can fix the unit tests to handle the concerns in (1). And I agree that we should cover both registration flows for pkg/unpkg, if we try to register with ProtocolActivator right now the function will just throw. This PR only contains the IsActivatorSupported function but we can definitely implement (2) when we start implementing unpkg.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -245,4 +245,33 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+    {
+        THROW_HR_IF(E_INVALIDARG, options == PushNotificationRegistrationOptions::Undefined); // WI_IsFlagSet does not allow 0 as flag.
+
+        auto isBackgroundTaskFlagSet{ WI_IsAnyFlagSet(options, PushNotificationRegistrationOptions::PushTrigger | PushNotificationRegistrationOptions::ComActivator) };
+        auto isProtocolActivatorSet{ WI_IsFlagSet(options, PushNotificationRegistrationOptions::ProtocolActivator) };
+
+        THROW_HR_IF(E_INVALIDARG, isBackgroundTaskFlagSet && isProtocolActivatorSet); // Invalid flag combination
+        if (AppModel::Identity::IsPackagedProcess() && IsBackgroundTaskBuilderAvailable())
+        {
+            if (isProtocolActivatorSet)
```
hulumane, line 266: Please add a comment here that we won't support protocol activation if COM activation is supported in the flow.

That mean's @sharath2727 should be calling this API in his channel registration flow as well rather than adding his own custom logic.

hulumane (approved):

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -245,4 +245,33 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
             LOG_IF_FAILED(::CoRevokeClassObject(static_cast<DWORD>(token.Cookie())));
         }
     }
+
+    bool PushNotificationManager::IsBackgroundTaskBuilderAvailable()
+    {
+        auto builder{ BackgroundTaskBuilder() };
+        auto builder5{ builder.try_as<winrt::IBackgroundTaskBuilder5>() };
+        return builder5 != nullptr;
```
DrusTheAxe, line 253: NIT: `return !builder;` is the more ModernC++ pattern

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -245,4 +245,33 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+    }
+
+    bool PushNotificationManager::IsActivatorSupported(PushNotificationRegistrationOptions const& options)
+    {
+        THROW_HR_IF(E_INVALIDARG, options == PushNotificationRegistrationOptions::Undefined); // WI_IsFlagSet does not allow 0 as flag.
```
DrusTheAxe, line 258: Remove comment (it's redundant - can't test bits when you provide no bits to test...)

In theory you could `WI_AreAllFlagsClear(options)`? Or `WI_AreAllFlagsClear(options, static_cast<int>(~0))`?
But the `options==Undefined` is clearer

dev/PushNotifications/PushNotifications.idl
```diff
@@ -24,8 +24,10 @@ namespace Microsoft.Windows.PushNotifications
     [flags]
     enum PushNotificationRegistrationOptions
     {
+        Undefined = 0x0, // Used to check if 0 was passed as a flag
```
DrusTheAxe, line 27: NIT: `...= 0` ?

DrusTheAxe (approved):

test/TestApps/PushNotificationsTestApp/main.cpp
```diff
@@ -312,15 +364,24 @@ int main() try
...
+    {
+        PushNotificationActivationInfo info(
+            PushNotificationRegistrationOptions::PushTrigger | PushNotificationRegistrationOptions::ComActivator,
+            winrt::guid(c_comServerId)); // same clsid as app manifest
 
-    g_appToken = PushNotificationManager::RegisterActivator(info);
-    
+        g_appToken = PushNotificationManager::RegisterActivator(info);
+    }
+    /* TODO: Register ProtocolActivator for unpackaged applications
```
hulumane, line 378: Nit: TODO: Register ProtocolActivator for unpackaged applications or Packaged Applications when COM activation is unsupported

hulumane (approved):

dev/PushNotifications/PushNotificationManager.cpp
jonwis, line 258: Comment is misleading as "zero" is not a flag. It's not WI_IsFlagSet's fault.

jonwis (approved):

PR #1164, opened 2021-08-02
Put objs under common obj folder

huichen123:
so that they don't pollute the source tree, and the file length is more predictable.


axelandrejs (approved):

build/AzurePipelinesTemplates/WindowsAppSDK-BuildProject-Steps.yml
```diff
@@ -66,7 +66,7 @@ steps:
       vsVersion: 16.0
       platform: '$(buildPlatform)'
       configuration: '$(buildConfiguration)'
-      msbuildArgs: '/p:AppxSymbolPackageEnabled=false /binaryLogger:$(Build.SourcesDirectory)/${{ parameters.solutionPath }}.$(buildPlatform).$(buildConfiguration).binlog /p:WindowsAppSDKVersionBuild=$(builddate_yymm) /p:WindowsAppSDKVersionRevision=$(builddate_dd)$(buildrevision) /p:VCToolsInstallDir="$(VCToolsInstallDir)\" /p:PGOBuildMode=$(PGOBuildMode) /p:WindowsAppSDKBuildPipeline=1'
+      msbuildArgs: '/restore /p:AppxSymbolPackageEnabled=false /binaryLogger:$(Build.SourcesDirectory)/${{ parameters.solutionPath }}.$(buildPlatform).$(buildConfiguration).binlog /p:WindowsAppSDKVersionBuild=$(builddate_yymm) /p:WindowsAppSDKVersionRevision=$(builddate_dd)$(buildrevision) /p:VCToolsInstallDir="$(VCToolsInstallDir)\" /p:PGOBuildMode=$(PGOBuildMode) /p:WindowsAppSDKBuildPipeline=1'
```
DrusTheAxe, line 69: Please remove

`/restore` does what? Do a `NuGet restore`?

That's unnecessary. The build pipeline explicitly does that in build\AzurePipelinesTemplates\WindowsAppSDK-BuildProject-Steps.yml line 40+ (`...displayName: 'NuGet restore ${{ parameters.solutionPath }}'...`) plus some other prep work before this task

And split across 2 (or more - see build-mrt.yml) is a good thing. If something goes wrong with the NuGet operation it's easier to identify (vs here mixed in with the build logic)

build/EnsureOutputLayout.props
```diff
@@ -19,6 +19,10 @@
     <!-- Use PlatformTarget instead of Platform to keep things more consistent (avoid 'Win32' as a build output) -->
     <BaseOutputPath>$(RepoRoot)\BuildOutput\$(Configuration)\$(PlatformTarget)\</BaseOutputPath>
 
+    <!-- Intermediate files under obj folder. Some intermediate file paths are really long, thus no unnecessary folder added -->
+    <BaseIntermediateOutputPath>$(RepoRoot)\obj\$(Configuration)\$(PlatformTarget)\$(MSBuildProjectName)\</BaseIntermediateOutputPath>
+    <IntermediateOutputPath>$(BaseIntermediateOutputPath)</IntermediateOutputPath>
```
DrusTheAxe, line 24: Nice

DrusTheAxe, line 24: What about C++/WinRT's "Generated Files" path? That's spewed directly under each project's directory (and can get disturbingly long). What controls its root, and how do we get it to use a similar path (outside the source trees)?

As a separate PR. But that's the only other artifact I'm aware of generated under source directories (/dev, /test, /installer/dev, /installer/test). It should be possible to put those on read-only media (and build successfully).

DrusTheAxe (changes requested):
Please remove the /restore and it's good to checkin

Doing likewise for C++/WinRT "Generated files" would be a good _separate_ PR

build/AzurePipelinesTemplates/WindowsAppSDK-BuildProject-Steps.yml
huichen123: yes. /restore does a restore for PackageReference. "nuget restore" doesn't fully respect directory.build.props, and puts project.assets.json in wrong place. Also it's recommended to restore using msbuild.

build/EnsureOutputLayout.props
huichen123: they are under obj too.

build/AzurePipelinesTemplates/WindowsAppSDK-BuildProject-Steps.yml
DrusTheAxe: Please follow up with @BenJKuhn and @kythant as what you describe implies the current pipeline definition is defective and needs broader changes

As /restore is unrelated to moving to /objs I suggest removing the /restore change and addressing it in a separate PR

huichen123: not defective, but cannot handle the custom path. /restore is a related change. otherwise those projects importing nuget packages with PackageReference will fail to build.

huichen123:
> Please remove the /restore and it's good to checkin
> 
> Doing likewise for C++/WinRT "Generated files" would be a good _separate_ PR

/restore is needed for this change, or msbuild will fail to find project.assets.json files. 
C++/WinRT "Generated files" are already handled.

build/AzurePipelinesTemplates/WindowsAppSDK-BuildProject-Steps.yml
DrusTheAxe: Then we should remove the explicit Restore tasks before this?

Why are we doing *both*?

huichen123: that's still needed for packages.config. Figuring out who they are and only restore those before build (that's actually what MRTCore does) is out of scope of this change.

PR #1169, opened 2021-08-03
User/ericjohnson327/transform fusion manifest

EricJohnson327:
Use xslt to create a fusion manifest from the application manifest.  This manifest will be put into a common 'manifests' folder to support unpackaged in-app use.

### DIFF
--- a/build/AzurePipelinesTemplates/WindowsAppSDK-CreateNugetPackage-Job.yml
+++ b/build/AzurePipelinesTemplates/WindowsAppSDK-CreateNugetPackage-Job.yml
@@ -47,7 +47,7 @@ jobs:
         LICENSE
       TargetFolder: '${{ parameters.fullnupkgdir }}'
 
-  # Copy the Windows App SDK full package specific target files and AppxManifest into the full package folder
+  # Copy the Windows App SDK full package specific target files, AppxManifest, and fusion manifests into the full package folder
   - task: PowerShell@2
     displayName: 'Copy Windows App SDK full package assets'
     inputs:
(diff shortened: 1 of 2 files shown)


build/AzurePipelinesTemplates/WindowsAppSDK-CreateNugetPackage-Job.yml
```diff
@@ -64,6 +64,11 @@ jobs:
         Copy-Item -Path "$targetsFilePath\Microsoft.WindowsAppSDK.Bootstrap.targets" -Destination "$fullpackagePath\build\Microsoft.WindowsAppSDK.Bootstrap.targets"
 
         Copy-Item -Path "$targetsFilePath\AppxManifest.xml" -Destination "$fullpackagePath\AppxManifest.xml"
+        $manifestPath = $fullpackagePath+'\manifests';
+        New-Item -ItemType Directory -Force -Path $manifestPath;
+        $xslt = New-Object System.Xml.Xsl.XslCompiledTransform;
+        $xslt.Load('build\TransformAppxManifest.xslt');
+        $xslt.Transform($fullpackagePath+'\AppxManifest.xml', $manifestPath+'\WindowsAppSDK.manifest');
```
Scottj1s, line 71: I would name the output something like Microsoft.WindowsAppSdk.Foundation.manifest, just to ensure it doesn't collide with other transport packages, and to provide a pattern to follow

Scottj1s (approved):

build/AzurePipelinesTemplates/WindowsAppSDK-CreateNugetPackage-Job.yml
EricJohnson327: Good thinking.  Updated

DrusTheAxe: @Scottj1s where should that be noted in the Transport Package Spec? ;-)

EricJohnson327: @DrusTheAxe I'm planning to make spec updates once the end-to-end is working

PR #1189, opened 2021-08-06
Git checkin-to-main policy

DrusTheAxe:


docs/Coding-Guidelines.md
```diff
@@ -57,3 +57,66 @@ ABI.
...
+code change is Experimental vs Preview or Stable.
+
+See [Final Decision: Release channels and cadences #637](https://github.com/microsoft/WindowsAppSDK/discussions/637)
+for the expectations of the various release channels.
+
+### Policy: Preview and Stable (aka non-Experimental)
+
+The following requirements must be met to checkin non-Experimental changes to main:
+
+1. API Review approval
```
jeffstall, line 75: Traditionally, we don't require API review to be completed before something RIs to rsmain. We just require API review before Windows ships (i.e., the GA build of Reunion ships). Are we changing policy here?

docs/Coding-Guidelines.md
```diff
@@ -57,3 +57,66 @@ ABI.
...
+2. Tests
```
jeffstall, line 76: What kind of tests? For example, automated vs. manual? Do they need to be part of the build? Do they need to run?

docs/Coding-Guidelines.md
```diff
@@ -57,3 +57,66 @@ ABI.
...
+1. **Selfhost** – a change is good but you're interested in community usage and feedback to verify
+   it's ready for Preview/Stable.
+2. **Skeleton** – API Review has approved and you've made an initial 'skeletal' implementation to
+   get the interface and infrastructure building but isn't functionally complete e.g. skeletal code
+   compiles but everything returns E_NOTIMPL/false/null/etc). Having the API surface building in
+   main can unblock documentation, test and sample writers.
+3. **Feedback** – your API shape isn't locked (or even clear) but you want and/or need feedback from
+   the community, partners and/or others before deciding the API is ready to consider for
+   non-Experimental releases. The API is not yet ready to go through API Review.
+
```
jeffstall, line 94: 4. **Under development** - the API is still under development and is used within WinAppSDK by other components. It will eventually be public to get feedback and be self-hosted.

docs/Coding-Guidelines.md
```diff
@@ -57,3 +57,66 @@ ABI.
...
+Reasons NOT to make Experimental content:
+
+1. **Prototype** – prototypes, ideas and wild experimentation should be done on a 'feature branch'
```
jeffstall, line 97: We need to determine how to have feature branches propagate throughout WinAppSDK, including Razzle. If co-engineering feature involves modifying multiple components, the feature team needs a way to get an end-to-end build.

docs/Coding-Guidelines.md
MikeHillberg: That's fine as long as it's [experimental]. We shouldn't commit an API (remove the tag) until we've reviewed it.

docs/Coding-Guidelines.md
MikeHillberg: The name "under development" is confusing, as this whole category is for under development APIs. All of the code in this category will eventually be public (or is expected to), and will get feedback and be self-hosted. So I don't think we need a 4th reason?

docs/Coding-Guidelines.md
jonwis: For Windows in-box, our guidance for multiple releases has been that APIs should be under experimental/feature configuration until they have been fully reviewed, so they show up in the prerelease SDK as "experimental" indicating that they might change. This is the same guidance - "mark your unfinished unreviewed APIs as experimental" ... the implication of this section is that if you merge to main you EITHER:

* Have finished your API review and it's in the final state
* Have not finished your API review and your API is marked experimental

(Note that some teams in-box forget to use the tagging, which occasionally leads to sadness.)

jonwis (approved):

docs/Coding-Guidelines.md
DrusTheAxe: Whatever level of tests be appropriate.

We have unit+functional + integration tests, and automated + manual. Specific mix is up to each feature team's judgement.

Issue #660, opened 2021-03-30
Accelerate Xcopy deployment availability


michaelongithub:
# Proposal: Urgent for starting in house apps. Killer condition

## Summary
Is still not available in Reunion 0.5. MSIX and the necessary cert handling is a no go for our need for apps in company desktop apps on PC's with monitor.

## Rationale
We dont want our SW engineers or server / pc staff to burdend with unwanted and unnecessary cert installs on target machines.
We will only start WINUI development if Xcopy deployment is availabel

-->
| Capability | Priority |
| :---------- | :------- |
| This proposal will allow developers to accomplish staritng much needed WINUI development unburdend on MSIX / cert knowhow and management. Not ready to train our staff (costs, availability) with this for us not needed knowledge | Must |

Thanks for considering and changing Your priorities

andrewleader:
Hey @DrusTheAxe, we think dynamic dependencies will resolve this concern - can you follow up with Michael on this thread to verify if that'll work?

driver1998:
Technically you should be able to do that now, Windows Terminal can be run if you extract the msix package and double click WindowsTerminal.exe.

DrusTheAxe:
>MSIX and the necessary cert handling is a no go...
>Rationale: We dont want our SW engineers or server / pc staff to burdened
>with unwanted and unnecessary cert installs on target machines.

@michaelongithub Not a problem. Project Reunion's MSIX packages are Windows Store signed. They install without requiring any cert management on your part.

Windows knows about the Windows Store certificates. MSIX packages signed with the Store cert (or a certificate chained to it) installs just fine. It's only if your MSIX is signed with another cert that you'd need to add it to Windows.

You can verify this for yourself by snooping at the package...

```
[D:]powershell -c get-appxpackage *reunion*


Name              : Microsoft.ProjectReunion.0.5
Publisher         : CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US
...
PackageFullName   : Microsoft.ProjectReunion.0.5_0.52103.25000.0_x64__8wekyb3d8bbwe
...
SignatureKind     : Store
...
```

[SignatureKind](https://docs.microsoft.com/uwp/api/windows.applicationmodel.package.signaturekind?view=winrt-19041)=[Store](https://docs.microsoft.com/uwp/api/windows.applicationmodel.packagesignaturekind?view=winrt-19041) means the package is Store signed.

Does this address your concern?

[...]

PR #1202, opened 2021-08-09
Initial draft of AppLocal Deployment spec.

mikebattista:
A few topics in need of feedback:

* What is the preferred terminology: AppLocal, Self-Contained, In-App?
* How does Windows App SDK AppLocal relate to .NET Self-Contained deployments?
  * Should we align on terminology?
  * Should we keep the msbuild property separate or piggyback on the existing .NET Self-Contained property? If developers want to take a framework package dependency on .NET but an AppLocal dependency on the Windows App SDK or vice versa, we should keep these separate.
* Are the requirements to redist and install the Windows App SDK Main and Singleton MSIX packages a problem for AppLocal scenarios?

We plan to deliver a preview of this in an upcoming 1.0 Preview release.

### DIFF
--- a/specs/Deployment/AppLocalDeployment.md
+++ b/specs/Deployment/AppLocalDeployment.md
@@ -1,0 +1,74 @@
+# AppLocal Deployment
+
+## Overview
+
+The default and preferred deployment model for the Windows App SDK is the Windows App SDK Framework Package which
+provides a single shared dependency across apps that can be automatically kept up to date and secure.
+
+However, developers sometimes prefer AppLocal deployment of their dependencies to minimize risk of
+automatic updates breaking compatibility and to ensure their dependencies aren't pulled out from
+under them unexpectedly. In some scenarios like xcopy deployment of apps, AppLocal deployment would
+be required.
+
+This spec captures the goals and requirements for supporting AppLocal deployment in the Windows App SDK
+as well as the developer experience for choosing this model.
+
+## Goals
+
+* Support AppLocal deployment for both packaged and unpackaged scenarios
+* Minimize effort required by Windows App SDK component owners to support all deployment and packaging models
+
+## Functional Requirements
+
+|ID|Requirement|Priority|
+|-|-|-|
+| 1 | Support packaged apps at runtime | P0 |
+| 2 | Support unpackaged apps at runtime | P0 |
+| 3 | Support [sparse package apps](https://docs.microsoft.com/windows/apps/desktop/modernize/grant-identity-to-nonpackaged-apps) at runtime | P0 |
+| 4 | Support Windows App SDK deployed via MSIX | P0 |
+| 5 | Support Windows App SDK deployed via AppLocal | P0 |
+| 6 | Support all runtime and deployment permutations downlevel to RS5 | P0 |
+| 7 | Windows App SDK components targeting stable releases MUST support all runtime and deployment permutations otherwise they are excluded from the release | P0 |
+| 8 | Windows App SDK component owners author their component registrations for packaged runtime and MSIX deployment which is then translated automatically to support other runtime and deployment permutations as necessary (e.g. RegFreeWinRT) | P0 |
+| 9 | Windows App SDK builds MUST produce artifacts required for packaged and unpackaged AppLocal deployment | P0 |
+| 10 | Build documentation for all supported runtime and deployment permutations (see [Developer Experience](#Developer-Experience)) | P0 |
+| 11 | Build samples for all supported runtime and deployment permutations (see [Developer Experience](#Developer-Experience)) | P0 |
+| 12 | Samples MUST be provided for all supported languages (see [Developer Experience](#Developer-Experience)) | P0 |
+| 13 | Build a test for each runtime and deployment permutation | P0 |
+
+## Developer Experience
+
+Project templates will only offer framework package deployment, as that is the recommended path. Developers must opt out to use AppLocal deployment.
+
+The opt out mechanism(s) must be simple and documented and must support all WinUI 3 Desktop project templates in both packaged and unpackaged scenarios.
+
+For v1.0, developers must specify the msbuild property `<WindowsAppSDKAppLocal>true</WindowsAppSDKAppLocal>` manually in their project file to switch to AppLocal deployment. Integration of this property into the Visual Studio Project Properties UI will be investigated in future releases.
+
+The [Windows App SDK Main and Singleton Packages](MSIXPackages.md) are required for unpackaged scenarios as well as some features like Notifications whose implementation relies on that package. These MSIX packages will need to be installed as a prerequisite even for AppLocal scenarios. For AppLocal deployments, the expectation is apps can redistribute these MSIX packages within their app and register them on startup using `PackageManager.AddPackageAsync(...)`. These packages are considered an undocked part of the OS so are treated more like an OS dependency rather than an AppLocal dependency, with the same compatibility promises of OS updates.
+
+## Design Guidelines for Windows App SDK Components
+
+AppLocal deployment implies two things:
+
+* Component binaries will be deployed locally within the app rather than within the shared MSIX framework package
+* Multiple instances of those binaries can all be running at the same time with multiple apps using AppLocal deployment
+
+Windows App SDK components need to support all runtime and deployment permutations so should not make assumptions that preclude any of these scenarios:
+
+* DON'T assume an app has package identity
+* DON'T assume an app has a package graph
+* DON'T depend solely on MSIX-specific features like StartupTasks or DEHs
+  * Features can depend on WARMain and still support all runtime and deployment permutations but must fail gracefully if WARMain is not installed
+* DON'T depend on a single instance of your component running in the shared MSIX framework package
+* DON'T depend on cross-process singleton resources
+  * Constant named kernel objects (e.g. mutex, event, etc.)
+  * Constant NT service name
+  * Constant absolute filename
+* DON'T depend on classic COM registration in HKCR
+* DON'T depend on installation in a common location shared across apps
+
+In addition to validating all runtime and deployment permutations within a single application, it is important to validate the following multiple application scenarios as well:
+
+* DO validate that multiple applications using a Windows App SDK component via the shared MSIX framework package install and run concurrently without affecting each other
+* DO validate that multiple applications using a Windows App SDK component via AppLocal deployment install and run concurrently without affecting each other
+* DO validate that one application using a Windows App SDK component via the shared MSIX framework package and another application using that component via AppLocal deployment both install and run concurrently without affecting each other


riverar:
Hey @mikebattista is there an existing issue thread where we can provide feedback?
https://github.com/microsoft/WindowsAppSDK/tree/main/specs#authoring-an-api-spec

mikebattista:
> Hey @mikebattista is there an existing issue thread where we can provide feedback?
> https://github.com/microsoft/WindowsAppSDK/tree/main/specs#authoring-an-api-spec

Please provide comments in the spec in the PR or here in the PR discussion. I'd prefer comments in the spec.

riverar:
Will do. @jonwis I'm concerned the community is expecting the documented process to be followed and will miss review of this.

mikebattista:
The linked process is for API design reviews which this is not. Though if filing an issue to "Support AppLocal Deployment of the Windows App SDK" will make it easier to get feedback, then I'm happy to do that.

nickrandolph:
Some initial feedback:

**Q: What is the preferred terminology: AppLocal, Self-Contained, In-App?**
A: I prefer Self-Contained however this should **only** be used if it is 100% consistent with the .NET publish --self-contained flag (see https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-publish) both in usage (ie using the --self-contained flag in the publish command) and outcome (ie packages the .net runtime). On the assumption that this is not the case I think In-App makes more sense. AppLocal is meaningless, please don't use this.

How does Windows App SDK AppLocal relate to .NET Self-Contained deployments?
**Q: Should we align on terminology?**
A: See previous answer - only if 100% consistent

**Q: Should we keep the msbuild property separate or piggyback on the existing .NET Self-Contained property? If developers want to take a framework package dependency on .NET but an AppLocal dependency on the Windows App SDK or vice versa, we should keep these separate.**
A: The two should probably be kept separate - I can't think of why a developer would want to take a dependency on one and not the other but why not leave this option open, since there are countless different deployment requirements for different developers. 

**Q: Are the requirements to redist and install the Windows App SDK Main and Singleton MSIX packages a problem for AppLocal scenarios?**
A: Yes, (unless I'm missing something) this completely negates the whole point of In-App deployment of the SDK. The goal should be xcopy deployment of apps - no installation of any component. The exception to this is where there needs to be some registration for say Identity, which should be done when app is executed (and needs corresponding Windows clean up task, for example if the app is deleted, moved etc).

DrusTheAxe:
> * Are the requirements to redist and install the Windows App SDK Main and Singleton MSIX packages

The Main+Singleton packages depend on the Framework package. For Main+Singleton to work Framework must also be installed i.e. you need it all.

**Applications using WinAppSDK** via AppLocal don't (directly) need the MSIX packages
**Main+Singleton MSIX packages** always need the Framework package.

If AppLocal requires redist+install the Main+Singleton MSIX packages then ditto the Framework package.

Does that change the AppLocal equation?

mikebattista:
> Some initial feedback:
> 
> **Q: What is the preferred terminology: AppLocal, Self-Contained, In-App?**
> A: I prefer Self-Contained however this should **only** be used if it is 100% consistent with the .NET publish --self-contained flag (see https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-publish) both in usage (ie using the --self-contained flag in the publish command) and outcome (ie packages the .net runtime). On the assumption that this is not the case I think In-App makes more sense. AppLocal is meaningless, please don't use this.

Agreed on only using Self-Contained if it meets the same definition as .NET. I don't think it does currently given the state of the MSIX packages that need to be redist. AppLocal has precedent, too, I know from CRT at least, but I'm not attached to any option.

> **Q: Should we keep the msbuild property separate or piggyback on the existing .NET Self-Contained property? If developers want to take a framework package dependency on .NET but an AppLocal dependency on the Windows App SDK or vice versa, we should keep these separate.**
> A: The two should probably be kept separate - I can't think of why a developer would want to take a dependency on one and not the other but why not leave this option open, since there are countless different deployment requirements for different developers.

My feeling too is that these should be kept separate.

> **Q: Are the requirements to redist and install the Windows App SDK Main and Singleton MSIX packages a problem for AppLocal scenarios?**
> A: Yes, (unless I'm missing something) this completely negates the whole point of In-App deployment of the SDK. The goal should be xcopy deployment of apps - no installation of any component. The exception to this is where there needs to be some registration for say Identity, which should be done when app is executed (and needs corresponding Windows clean up task, for example if the app is deleted, moved etc).

> If AppLocal requires redist+install the Main+Singleton MSIX packages then ditto the Framework package.
>Does that change the AppLocal equation?

I share the concerns here. One of the fundamental questions of this review is if the current design of those packages and the dependencies that Windows App SDK components are taking on them means that we can't truly support AppLocal/Self-Contained/In-App deployment unless we revisit those designs.

riverar:
> What is the preferred terminology: AppLocal, Self-Contained, In-App?
In no order, I'd go with perhaps self-contained, xcopy, or portable.

> How does Windows App SDK AppLocal relate to .NET Self-Contained deployments?
> .. Should we align on terminology?
It shouldn't, in any way. Not all consumers of the WAS are participants of the .NET ecosystem (e.g. C++, Rust). And I don't see any connections.

> .. Should we keep the msbuild property separate or piggyback on the existing .NET Self-Contained property? If developers want to take a framework package dependency on .NET but an AppLocal dependency on the Windows App SDK or vice versa, we should keep these separate.
Separate, again for folks not using .NET. And all of this should work without msbuild (e.g. Rust).

> .. Are the requirements to redist and install the Windows App SDK Main and Singleton MSIX packages a problem for AppLocal scenarios?
Problem? No. Silly and kind of undermines the entire point of an xcopy deployment? Yes. 😆 But hey it's better than nothing so please don't scrap anything.

gcaughey:
Q: What is the preferred terminology: AppLocal, Self-Contained, In-App?
A: I prefer self-contained if it works like .NET self-contained

Q: Should we keep the msbuild property separate or piggyback on the existing .NET Self-Contained property?
A: I'd always use them together but can see good reasons for making them distinct options.

Q: Are the requirements to redist and install the Windows App SDK Main and Singleton MSIX packages a problem for AppLocal scenarios?
A: My enterprise customers generally dislike MSIX and like the convenience of xcopy deployment. Anything requiring a separate installation is challenging for them for various reasons.

Scottj1s:
I would avoid SelfContained, precisely because it also applies to the [.NET deployment option](https://docs.microsoft.com/en-us/dotnet/core/deploying/).  What happens if/when we introduce a .NET runtime framework package, and .NET SelfContained becomes an option (vs framework-dependent)?  How will we disambiguate SelfContained-WindowsAppSDK from SelfContained-DotNet apps?  I would assume devs would want to mix and match these options.  So this seems like a recipe for confusion - I'd stick with AppLocal.

I assume "multiple instances" means "multiple versions".  Doesn't FrameworkPackage deployment also imply that multiple versions of binaries can be running simultaneously? 

There are parallel discussions on how to address other shared dependencies, like the .NET 5 runtime, the C Runtime, and C#/WinRT projection assemblies.  As with "self-contained", these may overlap with the term "component binaries", so I'd just call these out as Windows App SDK binaries.

Can we define some of the new terminology, like WARMain?  We should also not use the abbreviation WAR, as we no longer use the term WindowsAppRuntime/WinAppRuntime.  The [MSIXPackages spec needs to be updated](https://github.com/microsoft/WindowsAppSDK/issues/1215).

DrusTheAxe:
> We should also not use the abbreviation WAR, as we no longer use the term WindowsAppRuntime/WinAppRuntime

That is the correct term. The _project_ is WindowsAppSDK but the runtime _artifacts_ are WindowsAppRuntime. That includes the MSIX package identities and the DLLs (e.g. Microsoft.WindowsApp**Runtime**.dll). We haven't renamed these yet (I'm working on it).

> Can we define some of the new terminology, like WARMain? 

Like [this](https://github.com/microsoft/WindowsAppSDK/blob/main/specs/Deployment/MSIXPackageVersioning.md#123-windows-app-sdks-msix-packages)? ;-)

We generally don't use the WAR... terms, except when we're repeatedly referencing multiple MSIX packages (where using the full names would degrade comprehension). [specs/Deployment/MSIXPackageVersioning.md](https://github.com/microsoft/WindowsAppSDK/blob/main/specs/Deployment/MSIXPackageVersioning.md) is the only place I'm aware of that we do.

And occasionally in informal conversation, for same reason. WinAppSDK Framework is a mouthful, plus "WinAppSDK Main" can get confusing as WinAppSDK Main, WinAppSDK Singleton and WinAppSDK DDLM are all MSIX main packages (as in, PackageType=Main).

WARmain is just the modern lingo for what used to be informally referred to as PRmain (ditto PRfwk->WARfwk and PRddlm->WARddlm).

P.S. I did find a few references to the PR... terms across the codebase when I did the mass ProjectReunion->WindowsAppSDK rename. Tried to not just rename but also use the formal terms e.g. PRfwk -> Windows App SDK Framework, or WinAppSDK Framework.

DrusTheAxe:
> Doesn't FrameworkPackage deployment also imply that multiple versions of binaries can be running simultaneously?

Yes. Let's say (hypothetical example) Outlook uses the Foo framework package. I have Outlook running 24x7x365 on my machine, so it's running with foo v1.0.0.0. While it's running I install Foo framework v1.0.0.1. We install, say, Calculator and Foo v1.0.0.1. We _add_ v1.0.0.1 to the machine but don't disturb Outlook, who continues running using v1.0.0.0. I now launch Calculator and it's merrily running using Foo v1.0.0.1...while the Outlook process is running using v1.0.0.1. Someday when Outlook quits we'll update the resolved dependency to the newer Foo framework so next time Outlook runs it'll use v1.0.0.1.

So yes, framework packages support concurrent usage. Newer versions aren't immediately applied to everything, just everything non-disruptive and we coalesce to the newer version over time. Framework packages are very handy, hence the reason we aspire to do as much as possible via framework packages.

Alas, Framework packages can't do everything. Framework package's `appxmanifest.xml` can't declare Applications, PackagedCOM, BackgroundTasks, AppExtensions, StartupTasks and more. If you need these you need you need a Main package. The WinAppSDK Main+Singleton+DDLM packages exist for this reason - to help us do things we can't do via a framework package.

Inproc code / DLLs, data files and more work quite well via framework packages. It's usually the exe / process / out-of-process stuff that you start bumping into framework package restrictions. That's why [specs/Deployment/MSIXPackages.md](https://github.com/microsoft/WindowsAppSDK/blob/main/specs/Deployment/MSIXPackages.md) has [comments like](https://github.com/microsoft/WindowsAppSDK/blob/main/specs/Deployment/MSIXPackages.md#212-role---main)

> For these reasons Windows App SDK developers are strongly encouraged to only rely on the Main package when necessary.

[specs/Deployment/MSIXPackages.md](https://github.com/microsoft/WindowsAppSDK/blob/main/specs/Deployment/MSIXPackages.md) goes into this subject in more detail.

AdamBraden:
To complement @DrusTheAxe explanation, a main package is required to lift system services out of the OS and into store updatable packages so the WinAppSDK can rev more quickly than the OS. The spec mentions [PushNotifications](https://github.com/microsoft/WindowsAppSDK/blob/main/specs/PushNotifications/PushNotifications-spec.md) as an example of such a singleton. 

In v1.0 we have a chicken and egg challenge of getting these new main packages onto devices, and are working to solve that problem - easier done in newer versions of Windows, more challenging to get it downlevel. Until then, that means the only way to guarantee an app has all its dependencies, is for the app to register the main packages. In this scenario the application is running against an applocal copy of the framework, and thus the framework version it uses is not impacted by store updates to the shared framework. The main packages themselves will require the shared framework to be installed.

Do you feel this puts too much of a burden on xcopy/applocal deployments? Or do you think apps will avoid using such features so they can avoid registering the main packages (which would be an unsupported scenario)?

riverar:
> Until then, that means the only way to guarantee an app has all its dependencies, is for the app to register the main packages. In this scenario the application is running against an applocal copy of the framework, and thus the framework version it uses is not impacted by store updates to the shared framework.

Hm? I'm not aware of a difference between installs originating from the app vs. the user executing the runtimes installer. Why would these packages, once installed, be unservicable? Or are you referring to some sort of unsupported xcopy-everything deployment?

gcaughey:
> > Until then, that means the only way to guarantee an app has all its dependencies, is for the app to register the main packages. In this scenario the application is running against an applocal copy of the framework, and thus the framework version it uses is not impacted by store updates to the shared framework.
> 
> Hm? I'm not aware of a difference between installs originating from the app vs. the user executing the runtimes installer. Why would these packages, once installed, be unservicable? Or are you referring to some sort of unsupported xcopy-everything deployment?

In addition to this, would there be a difference between per-user vs. per-machine installations as far as package registration is concerned? Multiple users per machine is not uncommon for my customers, but they only want to install once.

AdamBraden:
> > Until then, that means the only way to guarantee an app has all its dependencies, is for the app to register the main packages. In this scenario the application is running against an applocal copy of the framework, and thus the framework version it uses is not impacted by store updates to the shared framework.
> 
> Hm? I'm not aware of a difference between installs originating from the app vs. the user executing the runtimes installer. Why would these packages, once installed, be unservicable? Or are you referring to some sort of unsupported xcopy-everything deployment?

In context of an app choosing to use applocal - the main and the shared framework packages would be serviceable by the store. regardless of installation mechanism (app registers the msix packages directly or the app/user runs the runtimes installer). The applocal copy of the framework would not be serviced by the store, that would require the app to ship an update.

riverar:
> The applocal copy of the framework would not be serviced by the store, that would require the app to ship an update.

I wasn't aware this even worked. You're saying I can stuff all the files from those WAS packages into my app and xcopy deploy the resultant mess to another machine? And the app will actually launch?

AdamBraden:
> > The applocal copy of the framework would not be serviced by the store, that would require the app to ship an update.
> 
> I wasn't aware this even worked. You're saying I can stuff all the files from those WAS packages into my app and xcopy deploy the resultant folder to another machine? And the app will actually launch?

That's what this spec is proposing. I may be getting into implementation details here, but imagine:
- In a packaged app with applocal - the framework package binaries would be copied similar to .net publish 'self contained' and tooling would write the winrt ACID registrations to the app's package manifest. The app would need to ensure the main+shared framework packages are installed for all apis to work correctly. The main package will still depend on the shared framework package. 
- In an unpackaged app with applocal - the framework package binaries would be copied similar to .net publish 'self contained' and tooling would write the winrt ACID registrations to the app's sxs/fusion manifest (app.exe.manifest). The app would need to ensure the main+shared framework packages are installed for all apis to work correctly. The main package will still depend on the shared framework package. This type of app would not use dynamic dependencies.

gcaughey:
> > > The applocal copy of the framework would not be serviced by the store, that would require the app to ship an update.
> > 
> > 
> > I wasn't aware this even worked. You're saying I can stuff all the files from those WAS packages into my app and xcopy deploy the resultant folder to another machine? And the app will actually launch?
> 
> That's what this spec is proposing. I may be getting into implementation details here, but imagine:
> 
> * In a packaged app with applocal - the framework package binaries would be copied similar to .net publish 'self contained' and tooling would write the winrt ACID registrations from the shared framework package to the app's package manifest. The app would need to ensure the main+shared framework packages are installed for all apis to work correctly. The main package will still depend on the shared framework package.
> * In an unpackaged app with applocal - the framework package binaries would be copied similar to .net publish 'self contained' and tooling would write the winrt ACID registrations from the shared framework package to the app's sxs/fusion manifest (<app>.exe.manifest). The app would need to ensure the main+shared framework packages are installed for all apis to work correctly. The main package will still depend on the shared framework package. This type of app would not use dynamic dependencies.

This unpackaged applocal scenario would be what my larger enterprise users would want. They are already familiar with installing our apps published as .NET self-contained and really do xcopy them to multiple PCs after testing them instead of using our installer as I understand it.

nickrandolph:
@AdamBraden what's the latest on this? 
I'm stunned that the team is considering an unpackaged applocal model that still requires main/shared framework package to be installed. Surely there must be a way that we can just bundle all the dependencies for the app, in the app itself and have it just work. It seem crazy that we've got this engineering masterpiece, that simply fails the simplest deployment strategy of all - simple xcopy deployment of the app.

AdamBraden:
> @AdamBraden what's the latest on this?
> I'm stunned that the team is considering an unpackaged applocal model that still requires main/shared framework package to be installed. Surely there must be a way that we can just bundle all the dependencies for the app, in the app itself and have it just work. It seem crazy that we've got this engineering masterpiece, that simply fails the simplest deployment strategy of all - simple xcopy deployment of the app.

@nickrandolph - we are aligning with the .NET term 'self-contained' , and we’ll behave the same way with all dependencies carried with the app to support xcopy deployments.

However there will be a small set of features that depend on a singleton. If managing these additional dependencies will compromise your xcopy deployments, then we recommend leveraging inbox features or lighting up the functionality at runtime when the dependencies are already available. 

For example, push notifications will be one of the features that requires a singleton, since push notifications depends on having a single socket connection shared across all apps in order to improve battery and perf… otherwise, there would be no point in us shipping push notifications if each app carried WNS themselves and had their own socket connection always running to WNS server. They might as well have their own socket connection to their own server at that point (what apps do today).

This is a point in time as we are on the journey for the Windows App SDK, and we are working to minimize scenarios when these dependencies are not available, but it may take a little while for us to get there especially on downlevel machines.

riverar:
> However there will be a small set of features that depend on a singleton. If managing these additional dependencies will compromise your xcopy deployments [...]

Isn't it more accurate to say that there is no xcopy deployment? That both packaged and unpackaged apps require the runtimes installed? You can't pick and choose with WAS at a granularity that suddenly makes xcopy work, so it doesn't really make sense to communicate that way.

mshaikhcool:
WAS is just a decoupled component framework from windows which needs to be installed separately (because it's decoupled.)  the point of WAS is that (i)it can be developed independently without changing windows and (ii)support for some downlevel versions going 1809. that's all I understand.

IIUC, there were no such promise of XCOPY deployment to begin with. https://github.com/microsoft/WindowsAppSDK/issues/660#issuecomment-819260272. Unpackaged means that one uses anything other than MSIX

DrusTheAxe:
> Unpackaged means that one uses anything other than MSIX

Correct - "unpackaged app" isn't synonymous with XCOPY-install, it just means "app not installed via MSIX". XCOPY-install, 'self-contained', etc are a subset of 'unpackaged app' solutions.

This spec outlined thinking regarding 'app local' and the discussion is helping vet and refine that thinking. As 'oft said, community feedback is valued to help us better understand needs and desires, how we are/not best meeting those and helps us chart a path forward that balances the various requirements and constraints.

Where the conversation started is less important than where it goes, so thank you (and everyone else, on this thread and elsewhere) for participating in the discussion. It's a really huge and varied ecosystem so the more voices and viewpoints shared the stronger our solutions can be :-)

mikebattista:
Supporting xcopy deployment is a primary goal here. Self-contained and xcopy deployment go hand in hand.

Notifications is one feature that requires some additional dependencies, and we'd recommend either avoiding features like that if they would compromise your self-contained/xcopy deployments, or light up the functionality only if the runtime is already available. We don't expect features requiring these additional dependencies to be the common case, and we actively discourage the approach unless the features wouldn't be meaningful without it.

> > However there will be a small set of features that depend on a singleton. If managing these additional dependencies will compromise your xcopy deployments [...]
> 
> Isn't it more accurate to say that there is no xcopy deployment? That both packaged and unpackaged apps require the runtimes installed? You can't pick and choose with WAS at a granularity that suddenly makes xcopy work, so it doesn't really make sense to communicate that way.

Are you referring to the fact that the App SDK is basically one umbrella reference vs. a la carte references to specific features? Our goal is that if you don't use a feature that requires the additional runtime dependencies, then you don't need to worry about those dependencies, regardless of your references, and we would document which features require those additional dependencies. That list should be small.

mshaikhcool:
the roadmap lists LITERALLY every single WAS features to be made available for non-MSIX apps ! 
why would anyone want to go the trouble of MSIX at this point when every single features are being made available for non MSIX ?

make the only `Clean Install-Uninstall` advantage of MSIX part of OS and maybe we can forget about Packaged | Unpackaged thing ?!

mikebattista:
Closing out as this feature has shipped and is publicly documented at https://learn.microsoft.com/en-us/windows/apps/package-and-deploy/deploy-overview.

PR #1219, opened 2021-08-11
Remove WinMDs from lib/native directory and update VSIX for merging WAS packages

kythant:
Fix MDMerge Issue with combining all the Nuget Packages by removing WinMDs files from native

Foundation puts it's WinMDs in both native and Uap10.0 whereas the other transport packages only puts it's WinMDs in Uap10.0 directory. When we combine all of the WAS Nuget Packages and put all the WinMDs from all the transport packages together into the lib directory, the result is that we hit an MDMerge issue when building a project. The MDMerge error occurs since the targets from WinUI and IXP is referecing all the winmds in the Uap10.0 and the target from Foundation is targeting the duplicate winmd in native as well.

Snippit from IXP's targets
```

  <!-- CsWinRT transitive build -->
  <ItemGroup>
     <CsWinRTInputs Include="$(MSBuildThisFileDirectory)..\lib\uap10.0\*.winmd" />
  </ItemGroup>
  
```


huichen123 (approved):
:shipit:

BenJKuhn (approved):

DrusTheAxe (approved):

Issue #1240, opened 2021-08-13
Windows App Runtime Deployment API


dkbennett:
# Proposal: Windows App Runtime Deployment API Design

## Summary
Provide API to query deployment status of the Windows App Runtime and provide means for developers to resolve or  deploy any missing packages.

The Windows App SDK consists of several framework packages and one or more main packages. The Windows Application Model supports package dependencies on frameworks, but does not support package dependencies on main packages. 

For unpackaged apps, the Windows App SDK installer can deploy all packages needed, and those apps can incorporate the installer or the packages into their own setups.  For packaged apps that is not possible, as they can only declare dependencies on the framework packages, not the main packages. This creates two problems for packaged apps using the Windows App SDK:

1. Identifying whether the status of the Windows App SDK packages are present and ready for use or if one or more packages are missing.

2). Resolving missing packages. There are two possible resolutions:
- Acquire and deploy the missing packages, if the app has the capabilities to do so.
- Direct the user to a location where they can resolve the missing packages, if it does not have the capability to acquire and/or deploy them.

This API is to provide a means of resolving both of these problems. For full-trust or packageManagement-capable packages, it can immediately resolve the missing dependency without further friction for the end user.

Proposed initial API spec is here:
https://github.com/microsoft/WindowsAppSDK/tree/user/dkbennett%2Fwasdeploymentapi

## Rationale
* Platform support for Main Package dependencies does not exist, and cannot exist on older supported versions. This fills the gap via runtime declaration and information on how to resolve the dependency.
* Some MSIX packages may have permission to resolve dependencies on their own and minimize user disruption and developer need to resolve dependencies themselves.
* Systems may be disconnected, operate in an enterprise, or otherwise not be on the latest version that software which the Windows App SDK may require. This provides a means to detect that situation and help resolve it gracefully.

This aligns with the Windows App SDK roadmap's Platform Unification and Deployment focus area. 

## Scope
| Capability | Priority |
| :---------- | :------- |
| This proposal will allow developers to check the Deployment Status of the Windows App SDK meets a minimum expected version | Must |
| This proposal will allow developers with a Full Trust or Package Management capable MSIX to deploy a missing Main Package for the Windows App Runtime. | Must |
| This proposal will provide developers with a URI for resolution of missing packages if it cannot be resolved automatically. | Could |
| This proposal will provide developers with streamlined install of missing packages for MSIX and Win32 packages that have appropriate permissions to do them. | Could |

[...]

PR #1241, opened 2021-08-13
WindowsAppRuntime Deployment API Spec

dkbennett:
Initial spec proposal for #1240 

Feedback appreciated, thank you!

### DIFF
--- a/specs/Deployment/DeploymentAPI.md
+++ b/specs/Deployment/DeploymentAPI.md
@@ -1,0 +1,136 @@
+
+# Background
+
+This is the spec for proposal [Windows App Runtime Deployment API #1240](https://github.com/microsoft/WindowsAppSDK/issues/1240).
+
+Windows App SDK (WAS) deployment consists of a Main package, multiple Framework packages, and for
+unpackaged apps a set of Dynamic Dependency packages, all of which are published by Microsoft 
+and are Store-Signed. Consumers of WAS will be both Microsoft and non-Microsoft publishers.
+In order to have WAS fully present and functional, the WAS Framework and Main packages must be 
+deployed to the system. While the Windows Application Model naturally supports framework
+dependencies it does not have a means of declaring a dependency on Main packages. This API spec
+and feature  proposal is to provide a means of identifying when the WAS dependencies are not
+satisfied, and if that is the case, provide a means to resolve the missing dependency. 
+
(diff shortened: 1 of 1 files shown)


riverar:
[just my usual complaint about [not following procedure](https://github.com/microsoft/WindowsAppSDK/tree/main/specs#authoring-an-api-spec) here but feel free to ignore]

The spec opens with "Windows App SDK (WAS) deployment" so it's not clear if you're trying to solve some sort of developer-only SDK scenario or if this is _really_ about the Windows App Runtime (WAR) and how apps built with the WAS will perform runtime detection.

Assuming this is about runtime detection... I don't really understand how to use this API. The spec, for example, states (emphasis mine):
> This is the means by which an app or program can call to verify the runtime packages is present to a minimum version needed by the loaded framework. **It should be called once after activation before trying to use the Windows App Runtime.**

What does activation here mean? I think you're trying to say: _when your app is launched, before you bootstrap, call this API_. But it's not clear what value this has over just, well, trying to bootstrap and handling the HRESULT.

Continuing my assumptions above, how does this API know what my unpackaged app needs prior to bootstrap?

dkbennett:
> [just my usual complaint about [not following procedure](https://github.com/microsoft/WindowsAppSDK/tree/main/specs#authoring-an-api-spec) here but feel free to ignore]
> 
> The spec opens with "Windows App SDK (WAS) deployment" so it's not clear if you're trying to solve some sort of developer-only SDK scenario or if this is _really_ about the Windows App SDK Runtime (WAR) and how apps built with the WAS will perform runtime detection.
> 
> Assuming this is about runtime detection... I don't really understand how to use this API. The spec, for example, states (emphasis mine):
> 
> > This is the means by which an app or program can call to verify the runtime packages is present to a minimum version needed by the loaded framework. **It should be called once after activation before trying to use the Windows App Runtime.**
> 
> What does activation here mean? I think you're trying to say: _when your app is launched, before you bootstrap, call this API_. But it's not clear what value this has over just, well, trying to bootstrap and handling the HRESULT.
> 
> Continuing my assumptions above, how does this API know what my unpackaged app needs prior to bootstrap?

My bad for not following the correct procedure, I'll remedy that. As to the other feedback, I'll try to make it more clear about what the purpose is here, which is to detect and resolve missing runtime dependency.

PR #1242, opened 2021-08-13
Start building Debug builds on CI pipeline

kythant:
Start building Debug builds on CI pipeline
Per Howard's suggestion

### DIFF
--- a/build/build-mrt.yml
+++ b/build/build-mrt.yml
@@ -118,14 +118,14 @@ steps:
     configuration: '$(buildConfiguration)'
     msbuildArguments: '/restore /binaryLogger:$(Build.SourcesDirectory)/mrtcore.$(buildPlatform).$(buildConfiguration).binlog'
 
-- ${{ if eq(variables.buildOnly, 'false') }}:
-  - task: PublishBuildArtifacts@1
-    displayName: 'Publish mrtcore binlog'
-    condition: succeededOrFailed()
-    inputs:
-      PathtoPublish: $(Build.SourcesDirectory)/mrtcore.$(buildPlatform).$(buildConfiguration).binlog
-      artifactName: binlogs
+- task: PublishBuildArtifacts@1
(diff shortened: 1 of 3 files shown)


build/build-mrt.yml
```diff
@@ -118,121 +118,122 @@ steps:
...
-      Microsoft.ApplicationModel.Resources.winmd
-    TargetFolder: '$(Build.ArtifactStagingDirectory)\mrt_raw\lib\anycpu'
-    flattenFolders: true
-
-- task: PublishBuildArtifacts@1
-  displayName: 'Publish Artifact: Binaries'
-  inputs:
-    PathtoPublish: '$(Build.ArtifactStagingDirectory)\mrt_raw'
-    ArtifactName: 'mrtcore_binaries'
+- ${{ if eq(variables.buildOnly, 'false') }}:
```
huichen123, line 121: why this check?

huichen123:
>   - template: publish-mrt.yml

this also needs update

---
Refers to: build/ProjectReunion-CI.yml:169 in 34c9923. [](commit_id = 34c9923d4ed6eec62d7a8c3aa52726c68d9b11fb, deletion_comment = False)

build/build-mrt.yml
kythant: because the copyto steps afterwards fails for various reasons

huichen123: is that because we hardcoded "Release" instead of using $(buildConfiguration)? we can skip copy, but we want to have test run. At the very minimum, we want to have the binlog published so that we have more information why build failed.

kythant: I think the test runs relies on the copto tasks

kythant:
YML is just too stupid for this to work

riverar:
Lets keep the unprofessional comments in your head. Lots of eyes on this project and all your messages go out as emails/notifications.

PR #1249, opened 2021-08-16
(Partial Fix) Dynamic Dependencies doesn't support Elevation #567

DrusTheAxe:
Dynamic Dependencies rquires PackagedCOM/WinRT activation which wasn't supported but now is as of Windows 10.0.22000.0.
[Dynamic Dependencies doesn't support Elevation #567](https://github.com/microsoft/WindowsAppSDK/issues/567) can be alleviated (I think) by updating the IsElevated check to not block if running on Windows >= 10.0.22000.0 (Win11)

This will (hopefully) unblock recent Windows releases while more comprehensive solutions are under investigation.

### DIFF
--- a/dev/DynamicDependency/MsixDynamicDependency.cpp
+++ b/dev/DynamicDependency/MsixDynamicDependency.cpp
@@ -3,6 +3,8 @@
 
 #include "pch.h"
 
+#include <VersionHelpers.h>
+
 #include "msixdynamicdependency.h"
 
 #include "MddDetourPackageGraph.h"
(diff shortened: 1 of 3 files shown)


riverar:
This is great as it also unblocks time travel debugging scenarios. Thanks!

DrusTheAxe:
Testing is showing failure to CoCreateInstance(packagedcom oop server) from an elevated process. Investigating

DrusTheAxe:
> Testing is showing failure to CoCreateInstance(packagedcom oop server) from an elevated process :-( Investigating

Turns out Windows' elevation restriction was lifted for WinRT but not PackagedCOM. That explains why DynDep's first call to CoCreateInstance(packagedcom) fails with REGDB_E_CLASSNOTREG even on the latest Win11 builds.

Verified with the COM team. Future improvements are under consideration but that doesn't change the current landscape  :-(

Ah well. It was worth a short. Can't roll 7s every time.

PR #1253, opened 2021-08-16
Windows App SDK Deployment API Design

dkbennett:
Initial spec proposal for #1240

### DIFF
--- a/specs/Deployment/DeploymentAPI.md
+++ b/specs/Deployment/DeploymentAPI.md
@@ -1,0 +1,136 @@
+
+# Background
+
+This is the spec for proposal [Windows App Runtime Deployment API #1240](https://github.com/microsoft/WindowsAppSDK/issues/1240).
+
+Windows App SDK (WAS) deployment consists of a Main package, multiple Framework packages, and for
+unpackaged apps a set of Dynamic Dependency packages, all of which are published by Microsoft 
+and are Store-Signed. Consumers of WAS will be both Microsoft and non-Microsoft publishers.
+In order to have WAS fully present and functional, the WAS Framework and Main packages must be 
+deployed to the system. While the Windows Application Model naturally supports framework
+dependencies it does not have a means of declaring a dependency on Main packages. This API spec
+and feature  proposal is to provide a means of identifying when the WAS dependencies are not
+satisfied, and if that is the case, provide a means to resolve the missing dependency. 
+
+The key scenario that is expected to have a missing dependency is for packaged MSIX which have
+declared the WAS frameworks as a dependency. In this scenario it is likely the main package would
+be missing because they cannot declare a dependency on the main package and it will not be
+deployed by the Store. For MSIX packages with permissions, this dependency could be immediately
+resolved online or offline if the package size is small and the framework carries the dependency.
+
+A class of MSIX packages that are Full Trust or have package management capabilities, such as
+Centennial MSIX packages will have the permissions to immediately resolve the dependency. These
+packages are the target for this version of the API. These packages will have the capability
+to deploy packages and satisfy the dependency. Since the Main Package for WAS is expected to be
+very small, this API has a streamlined solution for this scenario, which is to carry the WAS
+main packages inside the WAS framework, and leverage the full trust capability of the package
+to deploy the main package. A later release will address how to resolve the missing packages
+for UWPs when they become supported.
+
+
+# Description
+
+When a Windows App SDK (WAS) consuming Centennial package is deployed from the Store, it is
+expected to only have the App package and frameworks, and the WAS main package may not be present,
+or may not be at the desired version. This API enables a developer to detect when the required WAS
+main and/or framework packages are not present and take corrective action.
+
+Since this API is part of the WAS framework, the framework itself must be present.
+For MSIX packages, the framework must be listed as a dependency in the application manifest and
+will not be deployable without a min version being present. 
+
+Developers must not assume that the WAS main package is present, or that the version
+present is the minimum version required to support the feature set they expect. Instead, this API
+should be used to validate the WAS is the correct version and all required packages are present.
+
+The developer scenarios for this API is to answer the questions:
+1. Does this system have the min version of the Windows App SDK I require?
+2. What do I do to resolve this if it doesn't?
+
+The API addresses #1 by a simple query of the version they seek and a possible check for OS
+updates. #2 has two possible answers. First is direct resolution by the developer by acquiring and
+installing the packages if possible. Since the main packages are small and the target developer
+scenario will have permissions to deploy them, we will minimize developer and user pain by
+including the main package MSIX packages within the framework, and this API will allow for the
+deployment of that package if needed.
+
+For this release, packages without permission to deploy are not supported, and future versions
+will have solutions for such packages.
+
+# Examples
+
+## GetStatus
+
+This is the means by which an app or program can call to verify the runtime packages is present to
+a minimum version needed by the loaded framework. It should be called once after activation before
+trying to use the Windows App Runtime. This method returns a WindowsAppRuntime object
+that describes the current deployment state of the Windows App Runtime.
+
+Developers call this method without parameters, as all information about the version, channel, and
+architecture are derived from the current framework loaded and we assume any main packages needed
+by those frameworks are included within those framework packages.
+
+This method is good to identify if there is possible work required in the form of package install
+which may take several seconds to complete.
+
+```C++
+auto status =
+Windows::ApplicationModel::WindowsAppRuntime::Deployment::GetStatus();
+if (!status.IsOK())
+{
+    ...
+}
+```
+
+## Initialize
+
+Does a status check and if any packages are required for install, attempts to register those packages
+and get the WindowsAppRuntime into an OK state. This will perform package install if possible. The
+returned WindowsAppRuntimeStatus will contain the first error if one occurs during package install.
+
+
+```C++
+auto status =
+Windows::ApplicationModel::WindowsAppRuntime::Deployment::Initialize();
+if (!status.IsOK())
+{
+    return status.ExtendedError();
+}
+```
+
+
+# API Details
+
+```c#
+namespace Windows.ApplicationModel.WindowsAppRuntime
+{
+    /// Represents the current Deployment status of the WindowsAppRuntime
+    runtimeclass DeploymentStatus
+    {
+        /// Returns True if WindowsAppRuntime has all required packages at equal or greater version
+	    /// than the current framework requires.
+        Boolean IsOK{ get; };
+
+        /// Returns True if a Package install is required the WindowsAppRuntime to be in a good
+	    /// state.
+        Boolean PackageInstallRequired{ get; };
+
+        /// Returns the first encountered error if there was an error initializing or getting the status,
+        /// or S_OK if no error.
+        HRESULT ExtendedError{ get; };
+    }
+
+    /// Used to query deployment information for WindowsAppRuntime
+    static runtimeclass DeploymentManager
+    {
+        /// Returns the current deployment status of Windows App Runtime.
+        static DeploymentStatus GetStatus();
+
+        /// Checks the status of the WindowsAppRuntime and attempts to register any packages
+	    /// that can be registered. 
+        static DeploymentStatus Initialize();
+    }
+}
+```
+
+# Appendix


specs/Deployment/DeploymentAPI.md
```diff
@@ -0,0 +1,136 @@
...
+
+# Description
+
+When a Windows App SDK (WAS) consuming Centennial package is deployed from the Store, it is
+expected to only have the App package and frameworks, and the WAS main package may not be present,
+or may not be at the desired version. This API enables a developer to detect when the required WAS
+main and/or framework packages are not present and take corrective action.
+
+Since this API is part of the WAS framework, the framework itself must be present.
+For MSIX packages, the framework must be listed as a dependency in the application manifest and
```
DrusTheAxe, line 39: You usually say Main as a proper noun meaning the WAS Main package, and main as in a package where PackageType=Main (not any particular package). You should do likewise for Framework/framework.

You'll probably find it helpful to use WARmain, WARsingleton, WARfwk and WARsingleton like https://github.com/microsoft/WindowsAppSDK/blob/main/specs/Deployment/MSIXPackageVersioning.md#123-windows-app-sdks-msix-packages. I found a similar pain writing that spec -- writing out the full formal WinAppSDK package names _degraded_ clarity. You could up top cite the MSIXPackageVersioning.md spec for the list of WinAppSDK's MSIX packages and how you're using that terminology through the rest of this spec. Then you could say `WARfwk` here instead of `the WAS Framework` and likewise throughout the spec.

As a perk you reduce the need to update this spec in the future if/when we add (or remove) a package, since you just refer to that spec for the exact package list except where noteworthy to cite a specific package or subset.

specs/Deployment/DeploymentAPI.md
```diff
@@ -0,0 +1,136 @@
...
+will not be deployable without a min version being present. 
+
+Developers must not assume that the WAS main package is present, or that the version
+present is the minimum version required to support the feature set they expect. Instead, this API
+should be used to validate the WAS is the correct version and all required packages are present.
```
DrusTheAxe, line 44: What does "validate" mean? I'm jumping a little head here but specific in regards to 'validate'. Feel free to gloss over it here and delve it in in the API details below.

The scenario that came to mind is if WARmain is present and meets the needed minversion but package status != OK, is it 'valid'? Does it matter if the status is repairable (Tampered etc) vs not (Offline etc)?

specs/Deployment/DeploymentAPI.md
```diff
@@ -0,0 +1,136 @@
...
+
+The developer scenarios for this API is to answer the questions:
+1. Does this system have the min version of the Windows App SDK I require?
+2. What do I do to resolve this if it doesn't?
+
```
DrusTheAxe, line 49: 3. Are the packages usable?
?

Not sure if usable is quite the right term. Healthy? Ready to be used? &lt;Insert thesaurus here&gt;

specs/Deployment/DeploymentAPI.md
```diff
@@ -0,0 +1,136 @@
...
+The API addresses #1 by a simple query of the version they seek and a possible check for OS
+updates. #2 has two possible answers. First is direct resolution by the developer by acquiring and
```
DrusTheAxe, line 51: Reads a little run-on.

SUGGESTION: Split `#2 has two possible...` to a separate paragraph to aid clarity

specs/Deployment/DeploymentAPI.md
```diff
@@ -0,0 +1,136 @@
...
+installing the packages if possible. Since the main packages are small and the target developer
+scenario will have permissions to deploy them, we will minimize developer and user pain by
```
DrusTheAxe, line 53: ...we <del>will</del> minimize developer...

Present tense and active voice are generally preferred

specs/Deployment/DeploymentAPI.md
```diff
@@ -0,0 +1,136 @@
...
+including the main package MSIX packages within the framework, and this API will allow for the
```
DrusTheAxe, line 54: SUGGESTION: ...within the framework packages. This API supports deploying of the packages in the framework packages if needed.

specs/Deployment/DeploymentAPI.md
```diff
@@ -0,0 +1,136 @@
...
+deployment of that package if needed.
+
+For this release, packages without permission to deploy are not supported, and future versions
```
DrusTheAxe, line 57: SUGGESTION: Processes lacking permission to deploy MSIX packages are out-of-scope at this time.

specs/Deployment/DeploymentAPI.md
```diff
@@ -0,0 +1,136 @@
...
+will have solutions for such packages.
+
+# Examples
+
+## GetStatus
+
+This is the means by which an app or program can call to verify the runtime packages is present to
```
DrusTheAxe, line 64: packages <del>is</del> are present

specs/Deployment/DeploymentAPI.md
```diff
@@ -0,0 +1,136 @@
...
+a minimum version needed by the loaded framework. It should be called once after activation before
```
DrusTheAxe, line 65: SUGGESTION: A process should call `GetStatus()` after process initialization and before using Windows App Runtime content (e.g. in `main()` or `WinMain()`)

Using the API in `DllMain` or a global constructor would be unwise (not work at best, cause weird behavior to hangs or crashes at worst). Saying 'process initialization' or something like it's nice and generically vague but commonly understood

specs/Deployment/DeploymentAPI.md
```diff
@@ -0,0 +1,136 @@
...
+trying to use the Windows App Runtime. This method returns a WindowsAppRuntime object
+that describes the current deployment state of the Windows App Runtime.
+
+Developers call this method without parameters, as all information about the version, channel, and
+architecture are derived from the current framework loaded and we assume any main packages needed
```
DrusTheAxe, line 70: SUGGESTION: ...derived from the current WARfwk in use. The API assumes any other WinAppSDK packages needed by WARfwk  are embedded within WARfwk and available to be deployed.

specs/Deployment/DeploymentAPI.md
```diff
@@ -0,0 +1,136 @@
...
+by those frameworks are included within those framework packages.
+
+This method is good to identify if there is possible work required in the form of package install
```
DrusTheAxe, line 73: You make 2 good points here in 1 sentence

SUGGESTION: Split into 2 sentences e.g. "This method identifies if work is required for WinAppSDK to be ready for use." and defer the time-to-complete/not-instantaneous to the Initialize API's prose (GetStatus just reports the status, it doesn't change machine state, right?)

specs/Deployment/DeploymentAPI.md
```diff
@@ -0,0 +1,136 @@
...
+which may take several seconds to complete.
+
+```C++
+auto status =
+Windows::ApplicationModel::WindowsAppRuntime::Deployment::GetStatus();
```
DrusTheAxe, line 78: `auto status{ Microsoft::Windows::ApplicationModel... };`

1. Missing the Microsoft:: root namespace
2. [ES.23: Prefer the {}-initializer syntax](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#es23-prefer-the--initializer-syntax)

specs/Deployment/DeploymentAPI.md
```diff
@@ -0,0 +1,136 @@
...
+if (!status.IsOK())
+{
+    ...
+}
+```
+
+## Initialize
+
+Does a status check and if any packages are required for install, attempts to register those packages
```
DrusTheAxe, line 87: Phrasing

SUGGESTION: `This API performs a status check and...`

specs/Deployment/DeploymentAPI.md
```diff
@@ -0,0 +1,136 @@
...
+and get the WindowsAppRuntime into an OK state. This will perform package install if possible. The
+returned WindowsAppRuntimeStatus will contain the first error if one occurs during package install.
```
DrusTheAxe, line 89: This implies Initialize may do multiple things in a best-effort fashion and only return at the end returning 1st error. True?

If Initialize could, say, install 2 packages (e.g. WARmain + WARsingleton) and fails on the 1st, will it also try to install the 2nd and in the end return the 1st error encountered? This seems implied but not clearly stated.

specs/Deployment/DeploymentAPI.md
```diff
@@ -0,0 +1,136 @@
...
+
+
+```C++
+auto status =
```
DrusTheAxe, line 93: NIT: `auto status{ ... };

specs/Deployment/DeploymentAPI.md
```diff
@@ -0,0 +1,136 @@
...
+# API Details
+
+```c#
+namespace Windows.ApplicationModel.WindowsAppRuntime
+{
+    /// Represents the current Deployment status of the WindowsAppRuntime
+    runtimeclass DeploymentStatus
+    {
+        /// Returns True if WindowsAppRuntime has all required packages at equal or greater version
+	    /// than the current framework requires.
```
DrusTheAxe, line 111: NIT: indentation

riverar:
My earlier comments stand, doesn't seem like anything changed in the spec yet to address them. https://github.com/microsoft/WindowsAppSDK/pull/1241#issuecomment-899271950

Why would I use any of this? Unpackaged apps have to bootstrap right (which already provides an HRESULT)? And packaged apps will ship with deps inbox or pull them from the Store?

specs/Deployment/DeploymentAPI.md
```diff
@@ -0,0 +1,136 @@
+
+# Background
+
+This is the spec for proposal [Windows App Runtime Deployment API #1240](https://github.com/microsoft/WindowsAppSDK/issues/1240).
+
+Windows App SDK (WAS) deployment consists of a Main package, multiple Framework packages, and for
```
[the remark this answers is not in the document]
zaryaf: WAS should be changed to WinAppSDK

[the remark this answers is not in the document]
riverar: Are you suggesting this read as: Windows App SDK (WinAppSDK)? If so, I disagree. That doesn't make a lot of sense.

specs/Deployment/DeploymentAPI.md
```diff
@@ -0,0 +1,136 @@
...
+to deploy packages and satisfy the dependency. Since the Main Package for WAS is expected to be
+very small, this API has a streamlined solution for this scenario, which is to carry the WAS
+main packages inside the WAS framework, and leverage the full trust capability of the package
+to deploy the main package. A later release will address how to resolve the missing packages
+for UWPs when they become supported.
+
+
+# Description
+
+When a Windows App SDK (WAS) consuming Centennial package is deployed from the Store, it is
```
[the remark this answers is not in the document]
zaryaf: What about full trust packaged apps that are not deployed through the Store?

specs/Deployment/DeploymentAPI.md
zaryaf: "First is direct resolution by the developer by acquiring and installing the packages if possible" --> we should specify that this is the scenario wherein the app already is carrying the installer or MSIX packages (i.e., the unpackaged scenario)

specs/Deployment/DeploymentAPI.md
zaryaf: " Since the main packages are small and the target developer..." let's make it clear that this is the second option.

zaryaf: "target developer scenario will have permissions to deploy them..." might be useful to clarity again that this is the full trust packaged apps or app containers with packageManagement capability

specs/Deployment/DeploymentAPI.md
zaryaf: we can remove  "if needed" since you won't get into this scenario (resolving the Main gap) unless needed.

specs/Deployment/DeploymentAPI.md
zaryaf: "It should be called once after activation..." - could this API also not be used by apps to check the status of the packages at a later time and identify if there are missing dependencies?

specs/Deployment/DeploymentAPI.md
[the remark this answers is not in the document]
zaryaf: That's what we discussed internally with Andrew and Thomas. Has WAS actually been approved for use?

specs/Deployment/DeploymentAPI.md
zaryaf: why have GetStatus at all if Initialize can check status _and_ then resolve dependencies if any are missing?

dkbennett: GetStatus is essentially a exists check that will tell you if there is any work to do. Since the Initialize() method may do package installs, and package installs can take significant amounts of time, the GetStatus method is useful because a developer may wish to defer that potential performance-impacting work to a time that is not performance sensitive. Additionally, the developer may wish to gate functionality on availability of the WinAppSDK if they are using it for a limited set. This allows a low-impact way to test for that without attempting to do package deployment.

specs/Deployment/DeploymentAPI.md
```diff
@@ -0,0 +1,133 @@
...
+    ...
+}
+```
+
+## Initialize
+
+This API performs a status check and if any packages are required for install, attempts to register
+those packages and get the WindowsAppRuntime into an OK state. This will perform package install if
+possible. It is possible multiple packages may be required for install, and this will attempt
+to deploy all required packages. The returned WindowsAppRuntimeStatus will contain the first error
```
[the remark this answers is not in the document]
jonwis: You can listen to this on the existing `PackageCatalog` type, it probably doesn't need to be here. Although we should write some sample code showing how to mix progress from that into your application's experience.

specs/Deployment/DeploymentAPI.md
```diff
@@ -0,0 +1,141 @@
...
+        Boolean PackageInstallRequired{ get; };
+ 
+        /// Returns the first encountered error if there was an error initializing or getting the status,
+        /// or S_OK if no error.
+        HRESULT ExtendedError{ get; };
+    }
+
+    /// Used to query deployment information for WindowsAppRuntime
+    [experimental]
+    static runtimeclass DeploymentManager
```
jonwis, line 128: This steals a "good name" (`DeploymentManager`) ... should it be `WinAppSDKManager` in the `M.W.A` namespace instead?

specs/Deployment/DeploymentAPI.md
```diff
@@ -0,0 +1,141 @@
...
+    /// Represents the current Deployment status of the WindowsAppRuntime
+    [experimental]
+    runtimeclass DeploymentStatus
+    {
+        DeploymentStatus(
+            Boolean isOK,
+            Boolean packageInstallRequired,
+            HRESULT extendedError);
+
+        /// Returns True if WindowsAppRuntime has all required packages at equal or greater version
```
jonwis, line 113: When can both booleans be `true`? Or `false`?  In most WinRT "alternate success" models there's a `Result` type that encodes an enum plus an extended error code.  Like:

```c#
enum WinAppSDKDeploymentStatus {
    Ok = 0,
    PackageInstallRequired,
    PackageInstallFailed,
    Other,
}
runtimeclass WinAppSDKDeploymentResult
{
    WinAppSDKDeploymentResult(WinAppSDKDeploymentStatus status, HRESULT extendedError);
    WinAppSDKDeploymentStatus Status { get; };
    HRESULT ExtendedError { get; }
}
```

specs/Deployment/DeploymentAPI.md
```diff
@@ -0,0 +1,141 @@
...
+        /// Returns the first encountered error if there was an error initializing or getting the status,
+        /// or S_OK if no error.
+        HRESULT ExtendedError{ get; };
+    }
+
+    /// Used to query deployment information for WindowsAppRuntime
+    [experimental]
+    static runtimeclass DeploymentManager
+    {
+        /// Returns the current deployment status of Windows App Runtime package full name provided.
```
jonwis, line 130: Does this API work for any package full name? Or just the ones that are part of WinAppSDK?

What if the package referenced is superseded by another package that was deployed by servicing?

specs/Deployment/DeploymentAPI.md
```diff
@@ -0,0 +1,141 @@
...
+            Boolean packageInstallRequired,
+            HRESULT extendedError);
+
+        /// Returns True if WindowsAppRuntime has all required packages at equal or greater version
+        /// than the current framework requires.
+        Boolean IsOK{ get; };
+
+        /// Returns True if a Package install is required the WindowsAppRuntime to be in a good
+        /// state.
+        Boolean PackageInstallRequired{ get; };
```
jonwis, line 119: So the intent is the developer can know if the main (singleton, things like WNS) are present and accounted for, and they would do this by calling the API maybe twice, once with the main and once with the FXP?

specs/Deployment/DeploymentAPI.md
```diff
@@ -0,0 +1,133 @@
...
+
+## GetStatus
+
+This is the means by which an app or program can call to verify all required packages are present
+to a minimum version needed by the loaded framework.  A process should call GetStatus() after
+process initialization and before using Windows App Runtime content (e.g. in main() or WinMain()).
+
+Developers call this method without parameters, as all information about the version, channel, and
+architecture are derived from the current WARfwk loaded. This method is good to identify if there
+is possible work required in the form of package install which may take several seconds to
```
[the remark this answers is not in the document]
jonwis: As a follow-on, what threads are these APIs expected to be called on?  The samples should probably include a little more context about whether these are called in WinMain or in a message loop or on a background thread.  Maybe show a more complete example using "await/co_await" to specifically show it?

specs/Deployment/DeploymentAPI.md
```diff
@@ -0,0 +1,133 @@
...
+Since this API is part of the WinAppSDK framework, the framework itself must be present.
+For MSIX packages, the framework must be listed as a dependency in the application manifest and
+will not be deployable without a min version being present. 
+
+Developers must not assume that the WinAppSDK main package is present, or that the version
+present is the minimum version required to support the feature set they expect. Instead, this API
+should be used to validate the WinAppSDK is the correct version and all required packages are present.
+
+The developer scenarios for this API is to answer the questions:
+1. Does this system have the min version of the Windows App SDK I require and are they usable?
```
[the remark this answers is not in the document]
dkbennett: > Presumably "the main Windows App SDK package".
> 
> Can you give example circumstances of the system having the package and it not being usable? If it's not usable, how can the system say it has it?

An example of having the package and not being usable would be incorrect version, or having one package (the WindowsAppRuntime Main) but not all (WindowsAppRuntime Singleton). The WindowsAppRuntime has features which use two separate main packages, and this API simplifies the question of whether things are in a correct and expected state.

specs/Deployment/DeploymentAPI.md
```diff
@@ -0,0 +1,133 @@
...
+deployment of that package if needed.
+
+Processes lacking permission to deploy MSIX are out-of-scope at this time.
+
+# Examples
+
+## GetStatus
+
+This is the means by which an app or program can call to verify all required packages are present
+to a minimum version needed by the loaded framework.  A process should call GetStatus() after
```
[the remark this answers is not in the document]
dkbennett: > If the framework needs the packages, why is the app forced to check for the dependencies? Why doesn't the framework handle its own dependencies?

This is a limitation of the application model - frameworks cannot have dependencies. While this would be something we can address in a future Windows release, we need the WindowsAppRuntime to be backwards compatable to already-shipped versions of Windows, therefore we must work within these constraints as best we can.

specs/Deployment/DeploymentAPI.md
```diff
@@ -0,0 +1,133 @@
...
+process initialization and before using Windows App Runtime content (e.g. in main() or WinMain()).
+
+Developers call this method without parameters, as all information about the version, channel, and
+architecture are derived from the current WARfwk loaded. This method is good to identify if there
+is possible work required in the form of package install which may take several seconds to
+complete.
```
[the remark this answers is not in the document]
dkbennett: > If this call is to find out if all the required packages are installed, usable, and of an appropriate version, where are details of the packages and versions specified? I don't see how this call can know if everything needed is there if it hasn't been told what's needed.

We derive what is needed based on the version of the framework that the package is using. Each framework is built with a version and is compatible with a set of other packages - the Singleton, which works without version, and the Main, which is versioned. We currently also have a package full name optional parameter to specify the exact version, but if that isn't the framework that's being used it's going to be a failure. So this simplifies down to a single call where we already know the requested version information.

specs/Deployment/DeploymentAPI.md
```diff
@@ -0,0 +1,133 @@
...
+```
+
+## Initialize
+
+This API performs a status check and if any packages are required for install, attempts to register
+those packages and get the WindowsAppRuntime into an OK state. This will perform package install if
+possible. It is possible multiple packages may be required for install, and this will attempt
+to deploy all required packages. The returned WindowsAppRuntimeStatus will contain the first error
+if one occurs during package install. There will be no rollback of successfully installed packages
+if an error occurs on a subsequent package install.
```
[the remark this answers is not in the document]
dkbennett: Actionable failure scenarios are the usual out-of-disk-space/memory which could be checked. Access denied is a failure option from Initialize() if the package does not have permissions to deploy packages. 

Other types of deployment errors are possible but not really actionable by a developer.

specs/Deployment/DeploymentAPI.md
```diff
@@ -0,0 +1,133 @@
...
+The API addresses #1 by a simple query of the version they seek and a possible check for OS
+updates. #2 has two possible answers. First is direct resolution by the developer by acquiring and
+installing the packages if possible. Since the main packages are small and the target developer
+scenario will have permissions to deploy them, we will minimize developer and user pain by
+including the main package MSIX packages within the framework, and this API will allow for the
+deployment of that package if needed.
+
+Processes lacking permission to deploy MSIX are out-of-scope at this time.
+
+# Examples
```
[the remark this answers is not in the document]
dkbennett: Yes, that is the intent. Conceptually it's the same as initializing COM. The reason it is not built into the APIs themselves is because it may affect your usage, and deployment of packages is not instantaneous. It may take a few seconds to install the Main/Singleton the first time, and because of that we wanted to give you control over when that could happen.

The APIs that require that should also return a meaningful result if the expected runtime packages are not present and they fail for that reason.

specs/Deployment/DeploymentAPI.md
dkbennett: The intent is so the developer can know whether the reason the runtime is in a not-OK state is due to missing packages, which calling Initialize() will attempt to resolve. So the idea is that GetStatus() is a performant check, while Initialize() can also do package installs if the status is not OK and may not be instantaneous. I think the PackageInstallRequired boolean is not particularly useful since the action a developer would take (if GetStatus() is not ok, call Initialize()) is essentially the same, since Initialize() will take whatever actions are possible to resolve the not-OK state and turn it into an OK state.

The API could return a list of packages that are missing, but I don't think this is particularly actionable, since the framework contains the missing packages and calling Initialize() will attempt to deploy them if any are missing. However Initialize() could fail and for diagnostic purposes the list of missing packages may be useful, but that can also be done with telemetry.

specs/Deployment/DeploymentAPI.md
dkbennett: Agree, I will change to use something similar to this.

specs/Deployment/DeploymentAPI.md
DrusTheAxe: `WindowsAppRuntimeManager` ?

dkbennett: Are these preferable?
Namespace: Microsoft.Windows.ApplicationModel
Class: WindowsAppSDKManager

Not sure if WinAppSDKManager would be preferential to WindowsAppSDKManager, since it's sort of redundant in that we already are in the Microsoft.Windows.ApplicationModel namespace. We abbreviate App so is abbreviating Windows OK or should it stick with the full name as is in this repo project (WindowsAppSDK)?

DrusTheAxe: Windows preferable over Win. We abbreviate in rare cases we need to but full form is generally preferred and recommended

But...Runtime is preferable over SDK. As this API is to manage the Runtime components, not the SDK (same rationale as why we renamed packages from SDK to Runtime)

dkbennett:
> My earlier comments stand, doesn't seem like anything changed in the spec yet to address them. [#1241 (comment)](https://github.com/microsoft/WindowsAppSDK/pull/1241#issuecomment-899271950)
> 
> Why would I use any of this? Unpackaged apps have to bootstrap right (which already provides an HRESULT)? And packaged apps will ship with deps inbox or pull them from the Store?

Sorry I'm not intending to ignore your feedback, but rather address these questions in the spec. I think the reason for confusion is missing context.  The WindowsAppSDK is intended to function on Windows release RS5 and up, which means we have to work with all the limitations of the Application Platform as it exists on Windows RS5, otherwise this is not particularly useful since the customer base where your software will work is extremely limited to the latest Windows release.

A significant constraint with the platform is that Frameworks and Main Packages cannot depend on other main packages, or in the case of Frameworks, anything at all. In order to utilize some of the features of the platform, like Push Notifications and some other UX functionality, a Main package is needed. Currently we have two - a Singleton and Main package. Not all features require them, but some do, and this API exists to be able to answer and resolve the question of whether the runtime is in a good state as far as package deployment is concerned. And in the case of that runtime not being good, it needs to be able to try to resolve the problem.

The solution presented in this API is the ability to query and attempt to resolve any package deployment issues. It works by checking the framework which your package does depend on (because you can have dependencies on frameworks) and seeing if the associated main packages needed by that framework are present and at a supported version. If yes, then the result is an OK package status - you're good to go!  If the result is no, then we need to try to deploy those packages. Since they are small and lightweight at this time, a simple and elegant solution is to carry those packages within the framework itself. So the Initialize() call is effectively a call to GetStatus, and if the status is not OK, it attempts to deploy the packages within the framework which your package already depends. As long as your package has package deployment capability (i.e. it is full trust or has the associated capabilities), then this will work reliably to resolve the dependency.

Once the Main and Singleton packages have been deployed once - by any package that requires that release of the WindowsAppSDK - it generally doesn't need to ever be deployed again. The Store will auto-update those packages if updates do come along, or another app may carry with it an update. So we expect that the common case is that Initialize() will only ever deploy packages the first time it is called for that release of the SDK for each user. After that, status should be OK and so no work is needed. It's sort of a one-time setup call. Additionally, if Initialize() is called regularly if a user were to remove one of hte packages, such as the WindowsAppSDK Main or Singleton, then Initialize() will put them back and fix the issue. So this does give you some resiliency as long as multiple packages are required.

Does this answer your questions about why this API exists and why you would use it?

specs/Deployment/DeploymentAPI.md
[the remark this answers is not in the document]
dkbennett: It sort of depends on the app. At the simplest implementation, a single Initialize() call could be on WinMain, as in the sample. The first time the app launches, it _may_ need to deploy packages, in which case it may take a few seconds. Subsequent launches should not have that issue. 

The times where Initialize() actually deploys packages should be extremely limited to the first time the app is deployed to the system. Even other users should have a faster registration time than the initial first user deployment. This performance issue may not be a significant concern to some developers, in which case this solution is probably sufficient.

If you want a more robust handling of the potential performance issue, instead call GetStatus(), check the result, and if the result is not good, call Initialize() on another thread and you can handle the time delay more elegantly.

PR #1264, opened 2021-08-18
 Adding WIL dependency in WindowsAppSDK transport package.

ZenBird:
The insights library depends on WIL package. This change introduces the dependency on the appropriate version of WIL.

### DIFF
--- a/build/NuSpecs/Microsoft.WindowsAppSDK.Foundation.nuspec
+++ b/build/NuSpecs/Microsoft.WindowsAppSDK.Foundation.nuspec
@@ -11,5 +11,9 @@
          <projectUrl>https://aka.ms/windowsappsdk</projectUrl>
          <copyright>© Microsoft Corporation. All rights reserved.</copyright>
          <tags>Microsoft Resources Native Windows WindowsAppSDK</tags>
+
+         <dependencies>
+             <dependency id="Microsoft.Windows.ImplementationLibrary" version="[1.0.210803,)" />
+         </dependencies>
      </metadata>
 </package>


BenJKuhn:
Note that the transport packages dependencies don't map directly through to the published packages. @DefaultRyan is doing some work to sort out language-specific dependencies in the published App SDK nuget packages. If this dependency bleeds through to the public surface of the App SDK, you'll want to discuss with him how to update the App SDK nuget(s) to express this.

build/NuSpecs/Microsoft.WindowsAppSDK.Foundation.nuspec
```diff
@@ -11,5 +11,9 @@
          <projectUrl>https://aka.ms/windowsappsdk</projectUrl>
          <copyright>© Microsoft Corporation. All rights reserved.</copyright>
          <tags>Microsoft Resources Native Windows WindowsAppSDK</tags>
+
+         <dependencies>
+             <dependency id="Microsoft.Windows.ImplementationLibrary" version="[1.0.210803,)" />
```
DrusTheAxe, line 16: Why is this needed?

Sure, we use WIL in a lot of our C++ implementation but I'm not aware of a reason it's required in our API, and being 100% inline code it doesn't have any runtime component. Did someone bleed implementation through to their interface? If so, we should fix that. WIL is great for C++ development, but it shouldn't be _required_ to use Windows App SDK

ZenBird: This is specifically for WindowsAppRuntimeInsights.h. This file depends on WIL. If developers include the header, they would have to add WIL dependency in the project or this would not build.

DrusTheAxe: You should add checks in your header for WIL e.g. `#if defined(__WIL_SOMETHING)...rest of your header...#endif` so anyone #include'ing your header w/o WIL support first compiles, but gets no Insight functionality (WIL's heavily implemented that way. That's why you can include headers with and without include'ing WIL headers first and you get access to what you can and not things that don't work. That's also why you can include WIL headers multiple times and potentially get access to ever more symbols. Part of the trick is to not `#pragma once` and not wrap your whole header in `#if !defined(__FOO_H) #define __FOO_H ... #endif`. Instead you have smaller blocks e.g. in wil/foo.h something needs bar.h included first to compile WIL does `#ifdef __FOO_H #define __WIL_FOO ...declare stuff using foo.h's content... #endif`.

That makes your header adaptive, and potentially include'able multiple times (and work) e.g.

```
...
#include <insights.h>
// insight things using WIL aren't defined
#include <wil/foo.h>
#include <insights.h>
// insight things using WIL are now defined
...
```

Low friction way for folks who want to use you without negatively impacting those who don't need your content.

You could update this PR _or_ if you have folks chomping at the bit you could complete this PR and update your content via another PR.


Alternatively you could tell folks "don't include this if you don't have WIL". Leads to weird build errors if someone doesn't realize (or includes this before the right WIL header(s)) but it's not necessarily the worst answer.

riverar: Does this impact every user consuming the WAS? If so, that's bad. We don't want this dependency.

ZenBird: MRTCore includes WIL, so this dependency will come in via WindowsAppSDK.

BenJKuhn: @riverar, WIL doesn't have a DLL or other binaries or dependencies, so this shouldn't be required to consume App SDK. Also, this nuget definition is for the transport package. It's how one component publishes itself to other components in the build process. It's not a product nuget package. Those are defined elsewhere and composed later in the build process. So this is only impacting internal build stages within the process of creating App SDK.

@ZenBird and I should discuss whether this will flow through to the product Nuget package as well, but even there, keep in mind that since WIL is header-only, if we do include the dependency it will have no impact on products that don't use that functionality. I suspect there are some other things we can do to make even the dependency "pay for play". 

As an example the __has_include macro could be used here instead to provide a message when the dependency is required but not available. see: https://en.cppreference.com/w/cpp/preprocessor/include. If not present, an error message could point the user to the correct package.

ZenBird: Would this be an acceptable pattern 

#if __has_include (<wil/tracelogging.h>)
// Include WindowsAppRuntimeInsights.h
#else
#error "WIL package is required to use this header"
#endif

DrusTheAxe: has_include just checks if wil/tracelogging.h can be resolved. It doesn't mean anyone's #include'd it yet

To check if wil/tracelogging.h has been #include'd you need to test for something defined in wil/tracelogging.h in your header before referencing any wil/telemtry symbols e.g.

```
#if !defined(__WIL_TRACELOGGING_H)
#error "wil/tracelogging.h is required"
#endif
...
```

BUT...WIL doesn't use the classic macro guard game

```
// foo.h
#if !defined(__FOO_H)
#define __FOO_H
...
#endif
```

WIL wraps chunks of logic with that sort of macro guard, not entire files (that's one of the things WIL does so you can multiply include WIL headers and incrementally see more of WIL)

For what you describe **callers** would need to

```
#if __has_include (<wil/tracelogging.h>)
#include <wil/tracelogging.h>
#include <WindowsAppRuntimeInsights.h>
#else
#error "WIL package is required to use this header"
#endif
```

and the `#else` block is entirely optional on their part. And I'd expect callers would just include wil/tracelogging.h right before windowsappruntimeinsights.h and not bother with any conditional logic

The only alternative is _inside WindowAppRuntimeInsights.h_ you could not just detect but actually `#include <wil/tracelogging.h>` under the assumption if someone includes insights.h and didn't include wil tracelogging you could make it right. That doesn't work if wil/tracelogging.h has prerequisites or ordering dependencies you can't guarantee doing it for callers always works correctly.

That's why I recommend the WIL style macro guards. 

#if defined(__SOMETHING_DEFINED_IN_WIL_TRACELOGGING_H_INDICATING_ITS_BEEN_INCLUDED)
#define __WINDOWSAPPRUNTIME_INSIGHTS_H
...all the insights symbols in the current header...
#endif // __WINDOWSAPPRUNTIME_INSIGHTS_H
```
That way if someone includes insights without (or before) tracelogging the header is fine but anyone referencing Insights symbols will get compile errors. No different than if you stick `HowardIsAwesome();` in a source file without defining that function.

ZenBird: The __has_include() I alluded to earlier is to determine if the caller has included the WIL package in their project. WindowsAppRuntimeInsights.h cannot function without the package. As part of the checks before defining any of the macros/classes in WindowsAppRuntimeInsights.h, I would wrap the contents in __has_include () or report an error if the header cannot find WIL package.

DrusTheAxe: @BenJKuhn I see no need for the Transport to dictate to higher level repos what version of WIL must be used.

Yes, we want to inventory tool dependencies and get inconsistencies detected and potentially flagged in the future, but not every dependency requires strict alignment. WIL is a great example that only affects implementation so we have no technical requirement for e.g. DWriteCore to use WIL version == Foundation (or even &gt;=).

Funny story. Guess what I did yesterday?... I noticed Foundation projects were collectively referencing 4(!) different versions of WIL. Including ones from 2020 and 2019.<sup>1</sup> C++/WinRT was 'good' by comparison, just 3 different versions. I updated these to be consistent (and latest) and while we want better tooling visibility and problem detection, manually cribbing a dependency in a Transport package is both over- and under-kill.

One day I look forward to seeing buildinfo.json in every Transport package enumerating all tooling dependencies (nugets, VS, other) including versions, and a report showing inconsistencies in a Transport as well as the aggregator's reporting across all Transports it consumed. We're not there yet. But that's where we should be focusing.


<sup>1</sup> WindowsAppSDK_DLL was using the latest but lots of variability across the rest.

DrusTheAxe: > The __has_include() I alluded to earlier is to determine if the caller has included the WIL package in their project

That's not what __has_include does.

It tells you if the compiler configuration can find it, not if it's already been included.

Think of __has_include + include as a #include_if_found directive

If you want to detect if a header's already been included you can do so with #ifdef on some symbol in the header. That's been available for decades; no need to invent a new C++ keyword. Unlike __has_include, which C++ has never been able to do before

ZenBird: I am thinking __has_include() as an IfExists. I am using wil/tracelogging.h as the condition of whether WIL package has been included in the project. It could be any file under wil. WindowsAppRuntimeInsights.h requires that wil/tracelogging.h NOT be included before including itself. WindowsAppRuntimeInsights.h defines the macros that change the definition within wil/tracelogging.h

PR #1268, opened 2021-08-18
Moving from the static library based solution for telemetry to a head…

ZenBird:
…er based approach to reduce complexity.

The difference (on disk) size between the lib based approach and header based approach is 1 KB (header > lib). In favor of reducing the complexity, the decision has been made to go with a header based solution


build/AzurePipelinesTemplates/WindowsAppSDK-PublishInsightsHeader-Steps.yml
```diff
@@ -0,0 +1,26 @@
+parameters:
+  WindowsAppSDKInsightsSourceDirectory: $(Build.SourcesDirectory)\dev\WindowsAppSDK_Insights
+  WindowsAppSDKInsightsBinariesDirectory: $(Build.SourcesDirectory)\BuildOutput
```
ZenBird, line 3: >WindowsAppSDKInsightsBinariesDirectory: $(Build.SourcesDirectory)\BuildOutput [](http://example.com/codeflow?start=2&length=77)

nit: not needed.

build/AzurePipelinesTemplates/WindowsAppSDK-PublishInsightsHeader-Steps.yml
```diff
@@ -0,0 +1,26 @@
...
+
+steps:
+- task: NuGetToolInstaller@1
+  displayName: 'Use NuGet 5.6'
+  inputs:
+    versionSpec: 5.6
+  continueOnError: true
```
huichen123, line 10: is this needed?

ZenBird: Not needed. Will remove.

build/AzurePipelinesTemplates/WindowsAppSDK-PublishInsightsHeader-Steps.yml
```diff
@@ -0,0 +1,26 @@
...
+    SourceFolder: '${{ parameters.WindowsAppSDKInsightsSourceDirectory }}'
+    Contents: |
+      WindowsAppSDKInsights.h
+    TargetFolder: '$(Build.ArtifactStagingDirectory)\fullnuget\include'
+    flattenFolders: true
+
+- task: PublishBuildArtifacts@1
+  inputs:
+    PathtoPublish: '$(Build.ArtifactStagingDirectory)\fullnuget'
+    ArtifactName: 'FullNuget'
```
huichen123, line 26: this is probably handled by others. i tend to think we don't need a separate yml file.

dev/WindowsAppSDK_Insights/WindowsAppSDKInsightsLib/packages.config
```diff
@@ -1,4 +0,0 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<packages>
-  <package id="Microsoft.Windows.ImplementationLibrary" version="1.0.210803.1" targetFramework="native" />
-</packages>
```
huichen123, line 4: do we still need nuget.config, packages.config?

build/AzurePipelinesTemplates/WindowsAppSDK-PublishInsightsHeader-Steps.yml
ZenBird: Is the comment to merge this into another yml? If so, I want to keep this separate so that we are able to find the steps when we need to.

dev/WindowsAppSDK_Insights/WindowsAppSDKInsights.h
```diff
@@ -10,8 +10,50 @@
...
+    if (module)
+    {
+        char buffer[256];
+        if (0 < LoadStringA(module.get(), id, buffer, ARRAYSIZE(buffer)))
+        {
+            return buffer;
+        }
+        else
+        {
+            LOG_LAST_ERROR_MSG("Unable to load resource string, id: %d, error: %d", id, GetLastError());
```
huichen123, line 34: >error: %d [](http://example.com/codeflow?start=72&length=9)

why log again?

dev/WindowsAppSDK_Insights/WindowsAppSDKInsights.h
```diff
@@ -10,8 +10,50 @@
...
+        }
+        
+    }
+    else
+    {
+        LOG_LAST_ERROR_IF_MSG(module == nullptr, "Unable to load dll: %s", dllName);
```
huichen123, line 40: >module == nullptr [](http://example.com/codeflow?start=30&length=17)

no need to check

dev/WindowsAppSDK_Insights/WindowsAppSDKInsightsLib/packages.config
ZenBird: This file is deleted. We don't need this packages.config and nuget.config. The one in the root folder is still needed.

dev/WindowsAppSDK_Insights/WindowsAppSDKInsights.h
```diff
@@ -10,8 +10,50 @@
...
+        char buffer[256];
+        if (0 < LoadStringA(module.get(), id, buffer, ARRAYSIZE(buffer)))
+        {
+            return buffer;
+        }
+        else
+        {
+            LOG_LAST_ERROR_MSG("Unable to load resource string, id: %d, error: %d", id, GetLastError());
+        }
+
```
huichen123, line 36: remove

dev/WindowsAppSDK_Insights/WindowsAppSDKInsights.h
ZenBird: We need to know if any issues loading resources.

dev/WindowsAppSDK_Insights/WindowsAppSDKInsights.h
```diff
@@ -10,8 +10,50 @@
...
+    }
+    else
+    {
+        LOG_LAST_ERROR_IF_MSG(module == nullptr, "Unable to load dll: %s", dllName);
+    }
+
+    return {};
+}
+
+__declspec(dllexport) char* WINDOWSAPPSDK_PACKAGE_VER()
```
rohanp-msft, line 46: >__declspec(dllexport) [](http://example.com/codeflow?start=0&length=21)

needs to be inline or static to avoid LNK2005

e.g.:
`Error LNK2005 "char * __cdecl WINDOWSAPPSDK_PACKAGE_VER(void)" (?WINDOWSAPPSDK_PACKAGE_VER@@YAPEADXZ) already defined in LoggingTests.obj MrmBaseUnitTests S:\ProjectReunion\dev\MRTCore\mrt\mrm\UnitTests\mrmmin.lib(MrmTraceLogging.obj) 1 `

dev/WindowsAppSDK_Insights/WindowsAppSDKInsights.h
huichen123: LOG_LAST_ERROR_MSG is already logging last error

dev/WindowsAppSDK_Insights/WindowsAppSDKInsights.h
ZenBird: LNK 2005 means that the function is already defined. That can happen if the file and the lib are included. Are you sure that is not the case. The file itself is wrapped in #ifndef 

The __declspec(dllexport) must go.

rohanp-msft: #ifndef doesn't prevent duplication across different compilation units though, just within one.

build/ProjectReunion-BuildFoundation.yml
```diff
@@ -145,9 +145,9 @@ jobs:
         buildPlatform: 'ARM64'
         buildConfiguration: 'Release'
   steps:
-  - template: AzurePipelinesTemplates/WindowsAppSDK-BuildAndPublishInsights-Steps.yml
+  - template: AzurePipelinesTemplates/WindowsAppSDK-PublishInsightsHeader-Steps.yml
     parameters:
-      buildJobName: 'BuildAndPublishWindowsAppSDKInsights'
+      buildJobName: 'PublishWindowsAppSDKInsights'
```
huichen123, line 150: >buildJobName: 'PublishWindowsAppSDKInsights' [](http://example.com/codeflow?start=6&length=44)

not used.

build/AzurePipelinesTemplates/WindowsAppSDK-PublishInsightsHeader-Steps.yml
```diff
@@ -0,0 +1,13 @@
+parameters:
+  WindowsAppSDKInsightsSourceDirectory: $(Build.SourcesDirectory)\dev\WindowsAppSDK_Insights
+
```
huichen123, line 3: not really a parameter. may just remove the parameter and use it directly.

huichen123:
>         TraceLoggingBool(true, "UTCReplace_AppSessionGuid")

do we want this can be opt out?

---
Refers to: dev/WindowsAppSDK_Insights/WindowsAppSDKInsights.h:62 in 4e7bd28. [](commit_id = 4e7bd288fb81e981a5e8b8afee21b8ecd3d27194, deletion_comment = False)

ZenBird:
>         TraceLoggingBool(true, "UTCReplace_AppSessionGuid")

This is needed. There were WDX wide discussions to make it part of common PartB.

---
In reply to: [901503109](https://github.com/microsoft/WindowsAppSDK/pull/1268#issuecomment-901503109) [](http://example.com/codeflow?ancestors=901503109,901503109)

---
Refers to: dev/WindowsAppSDK_Insights/WindowsAppSDKInsights.h:62 in 4e7bd28. [](commit_id = 4e7bd288fb81e981a5e8b8afee21b8ecd3d27194, deletion_comment = False)

axelandrejs (approved):

build/AzurePipelinesTemplates/WindowsAppSDK-PublishInsightsHeader-Steps.yml
```diff
@@ -0,0 +1,10 @@
+steps:
+# START of copying files into their final package location.
+- task: CopyFiles@2
+  displayName: 'copy WindowsAppSDKInsights.h into output folder'
+  inputs:
+    SourceFolder: '$(Build.SourcesDirectory)\dev\WindowsAppSDK_Insights'
+    Contents: |
+      WindowsAppSDKInsights.h
+    TargetFolder: '$(Build.ArtifactStagingDirectory)\fullnuget\include'
+    flattenFolders: true
```
huichen123, line 10: I don't think it's justified to have a separate job for copy one file. This needs to be part of previous "build" job.

ZenBird: I disagree. Each module should have its own build/publish yml. The uber yml should only invoke and aggregate the results of child yml.

huichen123: restore of telemetry nuget package is part of reunion build job, but prep the telemetry header file for publish is a separate job. Also it's very inefficient to do the copy in its own job. @kythant

dev/WindowsAppSDK_Insights/WindowsAppSDKInsights.h
```diff
@@ -10,8 +10,49 @@
 #error "WIL Tracelogging.h must not be explicitly included when including this file"
 #endif
 
-char* WINDOWSAPPSDK_PACKAGE_VER();
-char* WINDOWSAPPSDK_EXPERIMENTATION_LEVEL();
+#include <wil/resource.h>
+#include <string>
+
+#define _WINDOWSAPPSDK_PACKAGE_VER_RESOURCEID             10000
+#define _WINDOWSAPPSDK_EXPERIMENTATION_FLAG_RESOURCEID    10001
```
DrusTheAxe, line 17: These are part of your API? I see nothing about INSIGHTS in their name., despite them being very much part of the Insights DLL and resources. Further, these are global scope symbols. At a minimum they need to be suitably scoped.

```
#define _WINDOWSAPPSDK_INSIGHTS_RESOURCEID_PACKAGE_VERSION   10000
#define _WINDOWSAPPSDK_INSIGHTS_RESOURCEID_CHANNEL           10001
```

But I see a broader range of issues, especially as this is an API and not just implementation detail.
1. Very C - global symbols, #defines, more
2. Unscoped names - almost no symbol indicates it's part of Insights
3. load_resource_string swallows errors...but (sometimes) throws exceptions - I don't see any HRESULT or throw, and yet std::string is used so it could throw exceptions. Is your intent to never throw and never return an error code, just return values or 'dead-values' (null, 0, etc) if something doesn't go perfectly right?
4. Assumption max-path-length-for-resource-dll is <256 chars - that's rather dated by today's standards. We need to be prepared to handle longer paths

Even if this is internal (not a public API) we can (and should) write Modern C++. No need to be restricted to Win32 C export rules in a header file meant for (very) C++ callers :-)

Here's one potential idea:

```
namespace Microsoft.Windows.Something.Insights
{
class RuntimeInformation
{
public:
    static std::string GetVersion()
    {
        const uint32_t c_versionId{ 10000 };
        static std::string value{ LoadStringFromResource(c_versionId) };
        return value;
    }

    static std::string GetChannel()
    {
        const uint32_t c_channelId{ 10001 };
        static std::string value{ LoadStringFromResource(c_channelId) };
        return value;
    }

private:
    static std::string LoadStringFromResource(uint32_t id)
    {
        static wil::unique_hmodule module = std::move(LoadResourceModule());

        // We retrieve known controlled data so we know it's max length and expect it to exist
        const int c_valueMaxLength{ 100 };
        char value[c_valueMaxLength]{};
        THROW_IF_WIN32_BOOL_FALSE_MSG(::LoadStringA(module.get(), id, value, ARRAYSIZE(value)), "Failed to load resource string id:%d", id);
        return std::string(value);
    }

    static std::string LoadResourceModule()
    {
        PCWSTR filename{ "Microsoft.WindowsAppSDK.Insights.Resource.dll" };
        wil::unique_hmodule module(::LoadLibraryA(filename));
        THROW_HR_IF_NULL_MSG(HRESULT_FROM_WIN32(GetLastError()), module, "Failed to load %hs", filename);
        return module;
    }
};
}
```

where usage is `auto version{ Microsoft::Windows::Something::Insights::RuntimeInformation::GetVersion() };` and `GetChannel`.

Symbols are scoped. Everything's in an `Insights` namespace. The class trick with static methods gives scoping so only `GetVersion` and `GetChannel` are exposed, all the implementation is private. Implementation details like what ids are used for what resources (or even that resources are used at all) is hidden inside the implementation, preventing leakage into every callsite.

This assumes an exception-based error handling strategy. If you want an error code base strategy you'd need functions returning HRESULT and moving retvals to out parameters. If you want a no-fail API then you need to replace std::string, e.g. return PCSTR to a static const value (instead of std::string), use fixed-length static buffers or static pointer to dynamically allocated memory (e.g. `std::unique_ptr<WCHAR[]>`), and so forth.

This is one of multiple possible solutions. The important point is to avoid unscoped global symbols, and to use a consistent error strategy.

DrusTheAxe (changes requested):

build/AzurePipelinesTemplates/WindowsAppSDK-PublishInsightsHeader-Steps.yml
DrusTheAxe: @ZenBird Insights isn't a module, it's an API in WinAppSDK. If it's becoming 1 .h file then...
1. it should live at `/dev/Insights` (just contains the 1 header file)
2. Update `dev/WindowsAppSDK_DLL/WindowsAppSDK_DLL.vcxproj` to copy ('publish') the header like others
3. CopyFilesToStagingDir.ps1 adds a line to copy the header to the Transport package (line it does MsixDynamicDependency.h and other headers)

WindowsAppSDK_DLL.vcxproj update (end of the file)

```
...
    <Copy SourceFiles="@(PublicHeaders)" DestinationFolder="$(OutDir)" />
    <Copy SourceFiles="$(MSBuildThisFileDirectory)..\Insights\WindowsAppSDKInsights.h" DestinationFolder="$(OutDir)" />
  </Target>
</Project>
```

This is consistent with how other WinAppSDK APIs' headers are handled.

It also makes Insight's header/API available to others in the project e.g. WindowsAppSDK_BootstrapDLL project will want to reference this soon. Cross-project should reference 'published' artifacts and not reach into source dirs (take a look at dev/DynamicDependency - there's 3 headers meant for use by others, everything else is implementation detail. But if folks used them from here they're point to the directory and have no clear distinction which headers are implementation vs interface).

DrusTheAxe: @huichen123 you do see the irony, given MRT is its own separate beast bolted onto Foundation (`/dev/mrt`, `build-mrt.yml`, `publish-mrt.yml`)? ;-)

Is that a temporary matter you expect to be (eventually) merge into WindowsAppSDK.sln, CopyFilesToStagingDir.ps1, etc?

huichen123: merging into WindowsAppSDK.sln is low priority. I'd only do that when WindowsAppSDK.sln has no build error/warning.

dev/WindowsAppSDK_Insights/WindowsAppSDKInsights.h
```diff
@@ -10,15 +10,64 @@
...
+    {
+        static wil::unique_hmodule module = std::move(LoadResourceModule());
+        const uint32_t c_ResourceMaxLength{ 100 };
+        char resourceValue[c_ResourceMaxLength]{};
+
+        THROW_IF_WIN32_BOOL_FALSE_MSG(::LoadStringA(module.get(), id, resourceValue, ARRAYSIZE(resourceValue)), "Failed to load resource string. id: %d", id);
+        return resourceValue;
+    }
+
+    static wil::unique_hmodule LoadResourceModule()
```
huichen123, line 50: >static wil::unique_hmodule LoadResourceModule() [](http://example.com/codeflow?start=4&length=47)

wondering why separate function for this.

huichen123 (approved):
:shipit:

dev/WindowsAppSDK_Insights/WindowsAppSDKInsights.h
```diff
@@ -10,15 +10,64 @@
 #error "WIL Tracelogging.h must not be explicitly included when including this file"
 #endif
 
-char* WINDOWSAPPSDK_PACKAGE_VER();
-char* WINDOWSAPPSDK_EXPERIMENTATION_LEVEL();
+#include <wil/resource.h>
+#include <string>
+
+namespace Microsoft
```
DrusTheAxe, line 16: NIT: C++ allows you to specify multiple namespaces in 1 statement e.g. `namespace A::B::C`

As of C++17 I think

ZenBird: Do we recommend C++ 17 for our consumers?

build/CopyFilesToStagingDir.ps1
```diff
@@ -97,6 +98,7 @@ PublishFile $FullBuildOutput\WindowsAppSDK_BootstrapDLL\Microsoft.WindowsAppSDK.
 PublishFile $FullBuildOutput\WindowsAppSDK_BootstrapDLL\MddBootstrap.h $NugetDir\include
 PublishFile $FullBuildOutput\WindowsAppSDK_DLL\MsixDynamicDependency.h $NugetDir\include
 PublishFile $FullBuildOutput\WindowsAppSDK_DLL\wil_msixdynamicdependency.h $NugetDir\include
+PublishFile $FullBuildOutput\WindowsAppSDK_DLL\WindowsAppSDKInsights.h $NugetDir\include\
```
DrusTheAxe, line 101: CONSIDER: Rename to WindowsAppRuntimeInsights.h

Better fit given the naming guidance we've received (when to use SDK vs Runtime)

Our DLLs and MSIX packages are being renamed from Windows App SDK to Windows App Runtime, per guidance. Only reason they haven't been renamed yet is I held off while Will worked on some restructuring. Will be changing soon

DrusTheAxe (approved):
Some minor feedback but otherwise looks ready

dev/WindowsAppSDK_Insights/WindowsAppRuntimeInsights.h
```diff
@@ -0,0 +1,67 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#pragma once
+
+#ifndef __WINDOWSAPPSDKINSIGHTS_INCLUDED
```
ZenBird, line 6: >__WINDOWSAPPSDKINSIGHTS_INCLUDED [](http://example.com/codeflow?start=8&length=32)

Rename

dev/WindowsAppSDK_Insights/WindowsAppSDKInsights.h
DrusTheAxe: Is Insights a public API or just for consumption within WinAppSDK?

I think we (everyone in WinAppSDK) are (or can) use C++17. Worst case we can tweak Insights' .h if someone finds a real blocker.

Publicly I'm not sure what level of C++ we're requiring, but definitely something. More than C is required e.g. MsixDynamicDependency.h declares `enum class foo : uint32_t...` - that `class` keyword requires C++, and that's intentionally used per API Review feedback/requirements.

If WindowsAppSDKInsights.h is a public API there should be a spec under `specs\insights` that went through API Review. Not such beast so I assumed it's an internal API, and in the Transport package so DWriteCore etc up the stack w/in WinAppSDK can call it (not because it's meant for public consumption). But perhaps I'm mistaken

ZenBird: It is technically public since we are publishing it with WindowsAppSDK. However, it is most useful for WindowsAppSDK code since it adds most value here.

PR #1274, opened 2021-08-20
Add Bootstrap C# API. Spec Bootstrap C# and get GenerationId (Flat-C/WinRT) APIs

DrusTheAxe:
Update spec for the Bootstrap C# API and Dynamic Dependency GenerationId property. 

Spec @ [Spec updates for GenerationId and Bootstrap C# API #1284](https://github.com/microsoft/WindowsAppSDK/pull/1284)

This PR also includes some internal maintenance (reorganizing some project layout).


specs/dynamicdependencies/sample-b.1.md
```diff
@@ -0,0 +1,107 @@
...
+int main()
+{
+    // Initialize access to Windows App SDK
+    const UINT32 c_majorMinorVersion{ 0x00010000 };
+    PCWSTR c_versionTag{ L"preview1" };
+    const PACKAGE_VERSION c_minVersion{};
+    wprintf(L"MddBootstrapInitialize(0x%08X, \"%s\", %hu.%hu.%hu.%hu)...\n",
+            c_majorMinorVersion, c_versionTag, c_minVersion.Major, c_minVersion.Minor, c_minVersion.Build, c_minVersion.Revision);
+    HRESULT hr{ MddBootstrapInitialize(c_majorMinorVersion, c_versionTag, c_minVersion) };
+     if (FAILED(hr))
```
riverar, line 21: Nit: extra space

riverar:
Just some minor feedback. 👍

dev/Bootstrap/CS/Microsoft.WindowsAppSDK.Bootstrap.Net/Bootstrap.cs
```diff
@@ -0,0 +1,218 @@
...
+        /// @see Initialize(uint)
+        /// @see Initialize(uint, string)
+        /// @see Shutdown()
+        public static void Initialize(uint majorMinorVersion, string versionTag, PackageVersion minVersion)
+        {
+            MddBootstrapInitialize(majorMinorVersion, versionTag, minVersion);
+        }
+
+        [DllImport("Microsoft.WindowsAppSDK.Bootstrap.dll", CharSet = CharSet.Unicode, ExactSpelling = true, PreserveSig = false)]
+        private static extern void MddBootstrapInitialize(uint majorMinorVersion, string versionTag, PackageVersion packageVersion);
```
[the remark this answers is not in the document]
DrusTheAxe: CharSet=Unicode covers it (see below). What I can't figure out why both CharSet and MarshalAs both exist. My best guess from what hints I've seen is CharSet applies to P/Invoke or a structure as a whole, but MarshalAs can be applied to an individual structure field or function parameter. If you know some definitive docs I've love to see a link :-)



https://docs.microsoft.com/en-us/dotnet/standard/native-interop/type-marshaling#default-rules-for-marshaling-common-types

.NET Type | Native Type
-- | --
string | Either char* or char16_t* depending on the CharSet of the P/Invoke or structure. See the [charset documentation.](https://docs.microsoft.com/en-us/dotnet/standard/native-interop/charset)

[Charsets and marshaling](https://docs.microsoft.com/en-us/dotnet/standard/native-interop/charset)

CharSet value | Windows | .NET Core 2.2 and earlier on Unix | .NET Core 3.0 and later and Mono on Unix
-- | -- | -- | --
Ansi | char (the system default Windows (ANSI) code page) | char (UTF-8) | char (UTF-8)
Unicode | wchar_t (UTF-16) | char16_t (UTF-16) | char16_t (UTF-16)
Auto | wchar_t (UTF-16) | char16_t (UTF-16) | char (UTF-8)

`CharSet = CharSet.Unicode` is specified so the string parameter is marshalled as expected.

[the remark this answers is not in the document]
riverar: Probably okay but beware those docs refer to .NET Core, which is newer than inbox .NET Framework. Not trying to win a nit battle here, but just covering your butt 😆

dev/Bootstrap/CS/Microsoft.WindowsAppSDK.Bootstrap.Net/Microsoft.WindowsAppSDK.Bootstrap.Net.csproj
```diff
@@ -0,0 +1,30 @@
...
+
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
+    <TargetFramework>net5.0-windows10.0.19041.0</TargetFramework>
+    <TargetPlatformMinVersion>10.0.17763.0</TargetPlatformMinVersion>
+    <Platforms>x86;x64;arm64</Platforms>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
+    <PlatformTarget>AnyCPU</PlatformTarget>
```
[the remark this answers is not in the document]
DrusTheAxe: VS must have inserted then when I was fiddling with getting Configuration Manager right

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/output#platformtarget

anycpu (default)...

Making these `<PlatformTarget>` redundant

Thanks for the note

dev/Bootstrap/CS/Microsoft.WindowsAppSDK.Bootstrap.Net/Bootstrap.cs
```diff
@@ -0,0 +1,218 @@
...
+        // Return the string as a formatted value "major.minor.build.revision".
+        public override string ToString()
+        {
+            return $"{Major}.{Minor}.{Build}.{Revision}";
+        }
+    };
+
+    internal static class NativeMethods
+    {
+        [DllImport("Microsoft.WindowsAppSDK.Bootstrap.dll", EntryPoint = "MddBootstralInitialize", CharSet = CharSet.Unicode, ExactSpelling = true, PreserveSig = false)]
```
jonwis, line 67: Spelling error on EntryPoint. Is there a test for this?

dev/Bootstrap/CS/Microsoft.WindowsAppSDK.Bootstrap.Net/Bootstrap.cs
```diff
@@ -0,0 +1,218 @@
...
+        {
+            NativeMethods.MddBootstrapShutdown();
+        }
+    }
+
+    // The Windows App SDK bootstrap initialization API. This class returns error as an HRESULT
+    // (>=0 on success, <0 on error).
+    //
+    // @see Bootstrap
+    public class Bootstrap_NoThrow
```
jonwis, line 151: Basically everything in C#-land should use exceptions, not hresults. What's the motivation for having this class? I'd just delete it. If you wanted to have a version that returned "nuh-uh, can't init" then have a method like TryInitialize instead, and it returns a `bool` instead of an `int`.

dev/DynamicDependency/API/MsixDynamicDependency.h
```diff
@@ -191,7 +191,7 @@ STDAPI MddGetIdForPackageDependencyContext(
 
 #if defined(WINDOWSAPPSDK_MICROSOFT_WINDOWS_APPLICATIONMODEL_DYNAMICDEPENDENCY_FEATURE_GENERATIONID_ENABLED) && (WINDOWSAPPSDK_MICROSOFT_WINDOWS_APPLICATIONMODEL_DYNAMICDEPENDENCY_FEATURE_GENERATIONID_ENABLED == 1)
 /// Return the package graph's current generation id.
-UINT32 MddGetGenerationId() noexcept;
+STDAPI_(UINT32) MddGetGenerationId() noexcept;
```
jonwis, line 194: @MikeHillberg was tracking important changes in 1.0, this is one of them since it meaningfully changes the signature.

jonwis (changes requested):
Looking pretty good. Would be useful to have a test, and see the comment about not needing the `_Nothrow` variant.

dev/Bootstrap/CS/Microsoft.WindowsAppSDK.Bootstrap.Net/Bootstrap.cs
riverar: `PreserveSig` is `true` by default. Folks calling into an wrapper of unmanaged Win32 `noexcept` APIs may be surprised by a sudden exception created out of thin air.

I'm okay with the deletion if you provide the `TryInitialize` variant.

dev/Bootstrap/CS/Microsoft.WindowsAppSDK.Bootstrap.Net/Bootstrap.cs
DrusTheAxe: I thought so too but "Entrypoint" doesn't compile. "EntryPoint" (cap P) is right - per VS compiling the project and [docs](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.dllimportattribute.entrypoint?view=net-5.0)

dev/Bootstrap/CS/Microsoft.WindowsAppSDK.Bootstrap.Net/Bootstrap.cs
DrusTheAxe: Idiomatic = +1

TryInitialize = Good idea, but return boolean means you lose the HRESULT. Which would you suggest:

#1

```c#
public static bool TryInitialize(..., out int hresult)
```

#2

```c#
public static int TryInitialize(...)
```

#3

```c#
public static int Initialize_NoThrow(...)
```

or ?

dev/Bootstrap/CS/Microsoft.WindowsAppSDK.Bootstrap.Net/Bootstrap.cs
riverar: I think he was referring to `MddBootstralInitialize`, note the missing `p` in bootstrap.

dev/Bootstrap/CS/Microsoft.WindowsAppSDK.Bootstrap.Net/Bootstrap.cs
riverar: I like `#1`!

DrusTheAxe: Nope, int is correct. HRESULT is a typedef for LONG i.e. 32-bit signed integer. Hence the reason you often seen values as ginormous negative integers

Hex at least is less cryptic. `0x80070005` == `-2147024891` == `E_ACCESS_DENIED`. Or maybe I've been doing this too long I don't see integers I just see severity, facility, code ;-)

DrusTheAxe: I like #1 too. Very C#esque e.g. Int.TryParse.

Thanks @jonwis and @riverar for the feedback!

See latest commit for revised syntax

DrusTheAxe:
> Looking pretty good. Would be useful to have a test, and see the comment about not needing the `_Nothrow` variant.

We have IntegrationTests/gates in the internal repo. They currently use the 0.8-preview Samples' `MddBootstrap.cs` as a placeholder, to be updated once this is available. That will serve multiple ends:

* tests exercising this API
* IntegrationTests exercising E2E flow
* gates to prevent regressions
* simple starter projects for other devs to use as a starter on their unpackaged tests (@MikeHillberg was chomping at the bit ;-) 
* working reference for Sample updates

dev/Bootstrap/CS/Microsoft.WindowsAppSDK.Bootstrap.Net/Bootstrap.cs
DrusTheAxe: Doh!

dhoehna (approved):

PR #1279, opened 2021-08-20
Include WIL before referencing WindowsAppRuntimeInsights.h

ZenBird:
This change will inform the caller that one must include the WIL package before including the header in their project.

### DIFF
--- a/dev/WindowsAppSDK_Insights/WindowsAppRuntimeInsights.h
+++ b/dev/WindowsAppSDK_Insights/WindowsAppRuntimeInsights.h
@@ -4,16 +4,15 @@
 #pragma once
 
 #if __has_include(<wil/tracelogging.h>)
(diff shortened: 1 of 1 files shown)


DrusTheAxe (approved):
Cool!

And of course a PR wouldn't be complete without a NIT 😉
No need to indent lines 7-15.
Notice how line 11 isn't indented from line 10... That's not only fine, that's my preference

I've seen 3 styles for preprocessor blocks

```
#ifndef FOO
#define FOO
#endif
```

```
#ifndef FOO
    #define FOO
#endif
```

```
#ifndef FOO
#    define FOO
#endif
```

Personally I prefer the first unless the logic's exceeding complex and in one block, then I prefer 3. But in the end it's style so...eh. Readable. Maintainble. Consistent. After that it's personal fashion preferences 😋

PR #1281, opened 2021-08-20
User/default/vsix split

DefaultRyan:
This splits the VSIX extension into C++ and C# components.

Doing this improves the developer experience, because it relaxes the requirement to have both the C# and C++ toolsets installed. This is also a requirement for us to get included as a Visual Studio component. (Hence, the Component subfolders).

At the moment, I'm leaving the original VSIX intact, because I want to have a discussion about what to do with it. I favor deprecating it for 1.0 and directing devs towards the new language-specific extensions. This also allows us to shed the original extension's baggage where its Id and Url still had "ProjectReunion" and couldn't be changed.

One other thing, with this extension, I've been battling a weird MSBuild issue that I can't quite root-cause. Using MSBuild's [binary logging](https://github.com/dotnet/msbuild/blob/main/documentation/wiki/Binary-Log.md), and determined that all the *.vsixmanifest files in subfolders of the project file are being indiscriminately added as items to the project. Or to put another way, given this structure:
```
+- extension.csproj
   +- Component
      |- source.extension.vsixmanifest
   +- Standalone
      |- source.extension.vsixmanifest
```

Both the Component and Standalone files are being added as Project Items (verified in the binlog). This can cause another part of the build process to simply pick one of the files as "the" manifest for some subset of the build process. I've combed through our project/props/targets files to find and explanation and I've come up empty. It's worth noting that we're using this same pattern in both [cppwinrt](https://github.com/microsoft/cppwinrt/tree/master/vsix) and (Single-Project MSIX)[https://docs.microsoft.com/en-us/windows/apps/windows-app-sdk/single-project-msix?tabs=csharp-vs2019], where it apparently works.

I've gotten past it for the moment in the C++/C# split by putting each proj into its own subfolder (before that, all the projs were in the Extensions folder, and they were all grabbing each others vsixmanifest, which would break the build).
I'll continue to debug that issue as I work on adjusting the Component versions, but I'm sending up the Bat Signal in case somebody has seen a similar issue and can help me figure it out.


alwu-msft:
> One other thing, with this extension, I've been battling a weird MSBuild issue that I can't quite root-cause. Using MSBuild's binary logging, and determined that all the *.vsixmanifest files in subfolders of the project file are being indiscriminately added as items to the project.

My bet is that this is an intersection of:
1. The VSIX project is a .NET SDK-style project 
2. The [glob pattern](https://docs.microsoft.com/en-us/dotnet/core/project-sdk/overview#default-includes-and-excludes) for the `None` default include used by .NET SDK-style projects
3. The fact that the invocation of the `FindSourceVsixManifest` task passed in the `None` itemgroup as the set of items to consider, and the task itself simply returns the first item named `source.extension.vsixmanifest`.
```xml
  <Target Name="FindSourceVsixManifest"
      Condition="'$(CreateVsixContainer)' == 'true' or '$(CopyVsixManifestToOutput)' == 'true'">
    <FindVsixManifest ItemsToConsider="@(None)"
              ProjectName="$(MSBuildProjectName)">
      <Output TaskParameter="VsixManifest" ItemName="SourceVsixManifest"/>
    </FindVsixManifest>
  </Target>
```

alwu-msft:
> <Project ToolsVersion="Current">

Given your (necessary) changes to the repository ID used by the C++ templates, there's not much point in keeping this project... any new VSIX you create with it is going to be unable to create C++ projects.

---
Refers to: dev/VSIX/Extension/Original/WindowsAppSDK.Extension.csproj:1 in 3af171d. [](commit_id = 3af171d23f9fe1c4929823008a43d7fa385aa2e1, deletion_comment = False)

dev/VSIX/Extension/Cs/WindowsAppSDK.Cs.Extension.csproj
```diff
@@ -0,0 +1,274 @@
...
+      XmlInputPath="$(VsixManifestSource)"
+      Query="/ns:PackageManifest/ns:Metadata/ns:Tags"
+      Value="$(_OriginalVsixTags)"
+      Namespaces="&lt;Namespace Prefix='ns' Uri='http://schemas.microsoft.com/developer/vsx-schema/2011' /&gt;" />
+    <XmlPoke
+      XmlInputPath="$(VsixManifestSource)"
+      Query="/ns:PackageManifest/ns:Metadata/ns:Preview"
+      Value="$(_OriginalVsixPreview)"
+      Namespaces="&lt;Namespace Prefix='ns' Uri='http://schemas.microsoft.com/developer/vsx-schema/2011' /&gt;" />
+  </Target>
```
alwu-msft, line 273: This should be factored out into `/dev/VSIX/Extension/Directory.Build.Targets` to avoid duplication (it's significant enough logic that we don't want to maintain multiple copies).

dev/VSIX/Extension/Cpp/Component/source.extension.vsixmanifest
```diff
@@ -0,0 +1,55 @@
...
+    </InstallationTarget>
+    <InstallationTarget Id="Microsoft.VisualStudio.Enterprise" Version="[16.9, 17.0)">
+      <ProductArchitecture>x86</ProductArchitecture>
+    </InstallationTarget>
+    <InstallationTarget Id="Microsoft.VisualStudio.Enterprise" Version="[17.0, 18.0)">
+      <ProductArchitecture>amd64</ProductArchitecture>
+    </InstallationTarget>
+  </Installation>
+  <Prerequisites>
+    <Prerequisite Id="Microsoft.VisualStudio.Component.CoreEditor" Version="[16.0,18.0)" DisplayName="Visual Studio core editor" />
```
alwu-msft, line 37: >Microsoft.VisualStudio.Component.CoreEditor [](http://example.com/codeflow?start=22&length=43)

I thought component VSIXs aren't allowed to take a dependency on CoreEditor unless they're part of a core component?

DefaultRyan: Correct, and good catch.

I was attempting to do this in two separate commits, but in all my attempts to try and fix the MSBuild weirdness, I started getting a little scope creep. I'll just finish this off and complete both the language split and component splits right now.

dev/VSIX/Extension/Cs/WindowsAppSDK.Cs.Extension.csproj
DefaultRyan: Thanks. I was starting to think the same thing.

DefaultRyan:
> > One other thing, with this extension, I've been battling a weird MSBuild issue that I can't quite root-cause. Using MSBuild's binary logging, and determined that all the *.vsixmanifest files in subfolders of the project file are being indiscriminately added as items to the project.
> 
> My bet is that this is an intersection of:
> 
> 1. The VSIX project is a .NET SDK-style project
> 2. The [glob pattern](https://docs.microsoft.com/en-us/dotnet/core/project-sdk/overview#default-includes-and-excludes) for the `None` default include used by .NET SDK-style projects
> 3. The fact that the invocation of the `FindSourceVsixManifest` task passes in the `None` itemgroup as the set of items to consider, and the task itself simply returns the first item named `source.extension.vsixmanifest`.
> 
> ```
>   <Target Name="FindSourceVsixManifest"
>       Condition="'$(CreateVsixContainer)' == 'true' or '$(CopyVsixManifestToOutput)' == 'true'">
>     <FindVsixManifest ItemsToConsider="@(None)"
>               ProjectName="$(MSBuildProjectName)">
>       <Output TaskParameter="VsixManifest" ItemName="SourceVsixManifest"/>
>     </FindVsixManifest>
>   </Target>
> ```

I wasn't aware of the glob pattern for .NET projects. That's another nice bit of evidence to build up a better picture, but it doesn't help me understand why this issue isn't afflicting the cppwinrt or single-project MSIX VSIXs.

I don't watn to waste much more time digging into this, so unless somebody has some inspiration, I'll probably just continue separating the Component/Standalone projects into their own subfolders with

DefaultRyan:
> > 
> 
> Given your (necessary) changes to the repository ID used by the C++ templates, there's not much point in keeping this project... any new VSIX you create with it is going to be unable to create C++ projects.
> 
> Refers to: dev/VSIX/Extension/Original/WindowsAppSDK.Extension.csproj:1 in [3af171d](https://github.com/microsoft/WindowsAppSDK/commit/3af171d23f9fe1c4929823008a43d7fa385aa2e1). [](commit_id = [3af171d](https://github.com/microsoft/WindowsAppSDK/commit/3af171d23f9fe1c4929823008a43d7fa385aa2e1), deletion_comment = False)

I'm going to remove all the templates and content from that project, but keep it around a little longer. Just long enough to decide if we're deleting, deprecating, or abandoning the existing extension (and finalizing the deprecation text).

alwu-msft:
Neither of those projects is an SDK-style project, so they don't have the globbing pattern. Your two options are:
1) Place the individual .csprojs low enough in the directory hierarchy that there is only single sibling/descendant source.extension.vsixmanifest
2) Add the following property to disable the default `None` globbing (note that you'll have to manually specify all `None` items yourself): `<EnableDefaultNoneItems>false</EnableDefaultNoneItems>`

---
In reply to: [904136891](https://github.com/microsoft/WindowsAppSDK/pull/1281#issuecomment-904136891) [](http://example.com/codeflow?ancestors=904136891,904136891)

DefaultRyan:
> Neither of those projects is an SDK-style project, so they don't have the globbing pattern. Your two options are:
> 
> 1. Place the individual .csprojs low enough in the directory hierarchy that there is only single sibling/descendant source.extension.vsixmanifest
> 2. Add the following property to disable the default `None` globbing (note that you'll have to manually specify all `None` items yourself): `<EnableDefaultNoneItems>false</EnableDefaultNoneItems>`
> 
> In reply to: [904136891](https://github.com/microsoft/WindowsAppSDK/pull/1281#issuecomment-904136891)

Thanks, that helped me get it working. Got the new updates pushed.

alwu-msft (approved):
:shipit:

PR #1284, opened 2021-08-20
Spec updates for GenerationId and Bootstrap C# API

DrusTheAxe:
Update spec for the Bootstrap C# API and Dynamic Dependency GenerationId property. Spec update (changes in specs/dynamicdependencies)

* specs/dynamicdependencies/DynamicDependencies.md
* specs/dynamicdependencies/sample-b.1.md

**Why Bootstrap C# API?** The Bootstrapper provides a Flat-C API taking multiple optional parameters. The Bootstrap C# API is provided as a .NET assembly with the following features:

Function overloads to make optional parameters optional
Bootstrap APIs scoped to a standard namespace and class
Choice of error handling via throw exceptions or returned HRESULT
C# friendly syntax implementing the necessary P/Invoke definition to adds the Bootstrap Flat-C API..
NOTE: This is a revised form of MddBootstrap.cs in 0.8-preview1's samples incorporating improvements based on community feedback.

NOTE: This provides a natural C# form of the Bootstrap API, which has only been available as a Flat-C API. The Bootstrap API is not provided as a WinRT API due to the nature of the Bootstrap API's required use before accessing the Windows App SDK. Although possible for C# to use a WinRT API via RegFreeWinRT it requires additional developer effort (Fusion manifest isn't seamlessly integrated into the C# developer experience). This C# API complements the Flat-C format to provide a natural and frictionless C# developer experience, to a degree not as easily accomplished with a WinRT API at this time.

**Why Generation Id?** Generation Id is a unique value changed whenever a process' package graph is altered. This is used to quickly and efficiently detect if the package graph has changed since the last time one examined it. This is of import if package graph sensitive data is computed and cached e.g. when the package graph is used as a search order for resources that are cached or otherwise expensive to recompute. One can remember the generation id when retrieving the package graph to compute some data, then later, before using that data, get the current Generation Id to check if the package graph is the same and the data's safe to use, or the package graph has changed requiring query'ing the package graph, scrapping the previously computed (now stale) data and recomputing it.


specs/dynamicdependencies/DynamicDependencies.md
```diff
@@ -1250,14 +1252,17 @@ STDAPI MddGetResolvedPackageFullNameForPackageDependency(
 STDAPI MddGetIdForPackageDependencyContext(
     _In_ MDD_PACKAGEDEPENDENCY_CONTEXT packageDependencyContext,
     _Outptr_result_maybenull_ PWSTR* packageDependencyId);
+
+/// Return the package graph's current generation id.
+STDAPI_(UINT32) MddGetGenerationId() noexcept;
```
jonwis, line 1257: Is the `noexcept` necessary here? Leaking exceptions across a DLL boundary is utterly verboten, so if this is an export I'm not sure it's required.  (Some compilers, like cl, have an "assume extern-c methods, like this one, don't throw.")

specs/dynamicdependencies/DynamicDependencies.md
```diff
@@ -1289,6 +1294,231 @@ This header contains the Lifetime Management API
...
+        /// @param minVersion the minimum version to use
+        /// @see Initialize(uint)
+        /// @see Initialize(uint, string)
+        /// @see Shutdown()
+        public static int Initialize(uint majorMinorVersion, string versionTag, PackageVersion minVersion)
+        {
+            return NativeMethods.MddBootstrapInitialize(majorMinorVersion, versionTag, minVersion);
+        }
+
+        /// Undo the changes made by Initialize().
```
jonwis, line 1507: I'd remove all this and link to your existing PR. Better to not ossify potentially-living code in documentation, it's code.

specs/dynamicdependencies/sample-b.1.md
```diff
@@ -0,0 +1,107 @@
...
+
+```c++
+#include <iostream>
+
+#include <MddBootstrap.h>
+
+int main()
+{
+    // Initialize access to Windows App SDK
+    const UINT32 c_majorMinorVersion{ 0x00010000 };
```
jonwis, line 15: `uint32_t` - prefer native types when possible.

specs/dynamicdependencies/sample-b.1.md
```diff
@@ -0,0 +1,107 @@
...
+    PCWSTR c_versionTag{ L"preview1" };
```
jonwis, line 16: Just use `auto` here.

specs/dynamicdependencies/sample-b.1.md
```diff
@@ -0,0 +1,107 @@
...
+                hr, c_majorMinorVersion, c_versionTag,
+                c_minVersion.Major, c_minVersion.Minor, c_minVersion.Build, c_minVersion.Revision);
+        return hr;
+    }
+
+    // Do interesting stuff...
+    std::cout << "Hello World!\n";
+
+    // Cleanup
+    MddBootstrapShutdown();
```
jonwis, line 33: Consider showing adding a `wil::unique_call` here that handles calling this shutdown method, even on exceptions.

specs/dynamicdependencies/sample-b.1.md
```diff
@@ -0,0 +1,107 @@
...
+            Console.WriteLine("Hello World!");
+
+            // Cleanup
+            Bootstrap.Shutdown();
+        }
+    }
+}
+```
+
+## C# (Return HRESULT on Error)
```
jonwis, line 75: I don't think this sample is necessary.

jonwis (changes requested):

specs/dynamicdependencies/DynamicDependencies.md
DrusTheAxe: `noexcept` is actually on all the existing exports, added during a code review due to feedback last year by you, Jevan or EHO (I forgot who). It's here for consistency and is harmless (at least, if specified). I haven't heard if VS assumes extern-C implies `noexcept`. I'll look into it for a future PR, easy enough to remove if redundant.

specs/dynamicdependencies/sample-b.1.md
DrusTheAxe: auto == `const wchar_t *`
PCWSTR == `typedef _Nullterminated_ const wchar_t*`

Is SAL dead to us?

specs/dynamicdependencies/DynamicDependencies.md
DrusTheAxe: Not following what you mean by "all this"

API Reviews are usually IDL and Flat-C headers, i.e. all interface. C# has doesn't do that last I looked. That's fine, the spec should be C#-like, just omit all bodies as N/A for specs+review? For instance this

```
        public static void Shutdown()
        {
            NativeMethods.MddBootstrapShutdown();
        }
```

should be spec'd as

```
        public static void Shutdown();
```

?

specs/dynamicdependencies/sample-b.1.md
DrusTheAxe: As discussed on the other PR, failure to call shutdown is a minor nit. Is automatic cleanup important? And if so, is that just a C++ thing or should we also provide it for C#?

specs/dynamicdependencies/DynamicDependencies.md
jonwis: Ah, yes.  Put the _interface_ here and leave the implementation in the .cs file, thanks.

Issue #927, opened 2021-06-11
[PushNotifications] Take a strong reference for PushNotificationManager object in CreateChannelAsync API 


sharath2727:
Description:
	Do we need an RAII type to hold the module lock while we are running in the background, to prevent us from being unloaded prematurely?

The problem is that since you do a resume_background in CreateChannelAsync API, the caller thread is going to resume execution and might even decide to release the PushNotificationManager (since it no longer needs it; if it needs it again, it’ll go get another one).

This means that the background thread is running with no active PushNotifciationManager object, so the DLlCanUnloadNow will return “Yes, you can unload me because I have no outstanding objects.”

The DLL then gets unloaded while your background thread is running.

Resolution: Avoid this by taking a strong reference to the PushNotificationManager object.

PR #1290, opened 2021-08-21
Channel Request for unpackaged applications

sharath2727:
Enable CreateChannelAsync API to support channel requests for both packaged and unpackaged applications. The API is modified to do two tasks:
1. If PushTrigger is supported then we call into Notification platform through frameworkudk to request channel using remoteId (GUID obtained from Azure Portal). As a fallback mechanism if the frameworkudk returns E_NOTIMPL then we go by the legacy way of requesting channels.
2. If PushTrigger isn't supported we consider the application as unpackaged and we go ahead register the application and request a channel for a given appusermodelId.


dev/PushNotifications/PushNotificationChannel.h
```diff
@@ -18,6 +19,8 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
 
     private:
         const winrt::Windows::Networking::PushNotifications::PushNotificationChannel m_channel{ nullptr };
+        winrt::Windows::Foundation::DateTime m_Extime{};
```
danielayala94, line 22: nit:
```suggestion
        winrt::Windows::Foundation::DateTime m_extime{};
```

dev/PushNotifications/PushNotificationChannel.h
```diff
@@ -18,6 +19,8 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        winrt::Windows::Foundation::Uri m_Uri{ nullptr };
```
danielayala94, line 23: nit:
```suggestion
        winrt::Windows::Foundation::Uri m_uri{ nullptr };
```

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -17,6 +17,9 @@
 #include "PushNotificationChannel.h"
 #include "externs.h"
 #include <string_view>
+#include <frameworkudk/pushnotifications.h>
+#include "..\..\BuildOutput\Debug\x64\PushNotificationsLongRunningTask.ProxyStub\NotificationsLongRunningProcess_h.h"
```
danielayala94, line 21: This should be `#include <NotificationsLongRunningProcess_h.h>`.

If there are compile issues, we need to talk offline. But I think I already did everything needed to avoid referencing this long path.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -62,13 +65,68 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
         }
     }
 
+    // RegisterUnpackagedapplication
+    void PushNotificationManager::RegisterApplicationHelper(wil::unique_cotaskmem_string &unpackagedAppUserModelId, winrt::guid remoteId)
```
danielayala94, line 69: Don't forget to change the name of the function! `RegisterUnpackagedApplicationHelper`, or something similar.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -62,13 +65,68 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                wil::com_ptr<INotificationsLongRunningPlatform> reunionEndpoint{
+                    wil::CoCreateInstance<NotificationsLongRunningPlatform, INotificationsLongRunningPlatform>(CLSCTX_LOCAL_SERVER) };
+
+                char processName[1024];
+                GetModuleFileNameExA(GetCurrentProcess(), NULL, processName, sizeof(processName) / sizeof(processName[0]));
+
+                winrt::throw_hresult(reunionEndpoint->RegisterFullTrustApplication(processName, remoteId, &unpackagedAppUserModelId));
+            }
+
+        return;
```
danielayala94, line 86: Not needed. Please remove.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -62,13 +65,68 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+            THROW_IF_FAILED(::CoInitializeEx(nullptr, COINITBASE_MULTITHREADED));
+
+            auto scopeExit = wil::scope_exit(
+                [&]() { CoUninitialize(); });
+
+            {
+                wil::com_ptr<INotificationsLongRunningPlatform> reunionEndpoint{
+                    wil::CoCreateInstance<NotificationsLongRunningPlatform, INotificationsLongRunningPlatform>(CLSCTX_LOCAL_SERVER) };
+
+                char processName[1024];
```
danielayala94, line 80: Are we sure 1024 is a reasonable size or is it too much?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -62,13 +65,68 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                GetModuleFileNameExA(GetCurrentProcess(), NULL, processName, sizeof(processName) / sizeof(processName[0]));
```
danielayala94, line 81: Does `GetCurrentProcess()` give you the full path to the exe?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -62,13 +65,68 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        wil::unique_cotaskmem_string& channelUri,
+        winrt::Windows::Foundation::DateTime& channelExpiryTime,
+        winrt::guid remoteId)
+    {
+        wchar_t appUserModelId[APPLICATION_USER_MODEL_ID_MAX_LENGTH + 1] = {};
+        UINT32 appUserModelIdSize = ARRAYSIZE(appUserModelId);
+
+        THROW_IF_FAILED(GetCurrentApplicationUserModelId(&appUserModelIdSize, appUserModelId));
+
+        /* channel request result */
```
danielayala94, line 99: I'm not sure of the value of this comment. Remove?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -62,13 +65,68 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        HRESULT operationalCode;
+        wil::unique_cotaskmem_string channelId;
+        ABI::Windows::Foundation::DateTime channelexpirytime{};
```
danielayala94, line 102: nit:
```suggestion
        ABI::Windows::Foundation::DateTime channelExpiryTime{};
```

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -62,13 +65,68 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+
+        HRESULT hr = PushNotifications_CreateChannelWithRemoteIdentifier(
+            appUserModelId,
+            remoteId,
+            &operationalCode,
+            &channelId,
+            &channelUri,
+            &channelexpirytime);
+
+        wprintf(L"appIdentifier: %s\n", appUserModelId);
```
danielayala94, line 112: Please remove.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -91,16 +149,48 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
             try
             {
-                PushNotificationChannelManager channelManager{};
-                winrt::PushNotificationChannel pushChannelReceived{ nullptr };
+                if (IsActivatorSupported(PushNotificationRegistrationOptions::PushTrigger))
+                {
+                    wil::unique_cotaskmem_string channelUri;
+                    winrt::Windows::Foundation::DateTime channelExpiryTime;
+
+                    hresult hr = CreateChannelWithRemoteIdHelper(channelUri, channelExpiryTime, remoteId);
```
danielayala94, line 157: Is this type `winrt::hresult` ? Maybe let's make explicit it's winrt?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -62,13 +65,68 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+            appUserModelId,
+            remoteId,
+            &operationalCode,
+            &channelId,
+            &channelUri,
+            &channelexpirytime);
+
+        wprintf(L"appIdentifier: %s\n", appUserModelId);
+
+        THROW_HR_IF(hr, (hr != E_NOTIMPL) && (hr != S_OK));
```
danielayala94, line 114: Do we need to explain in a comment why we are not throwing in the `E_NOTIMPL` case?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -62,13 +65,68 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                winrt::throw_hresult(reunionEndpoint->RegisterFullTrustApplication(processName, remoteId, &unpackagedAppUserModelId));
+            }
+
+        return;
+    }
+
+    winrt::hresult PushNotificationManager::CreateChannelWithRemoteIdHelper(
+        wil::unique_cotaskmem_string& channelUri,
+        winrt::Windows::Foundation::DateTime& channelExpiryTime,
+        winrt::guid remoteId)
```
danielayala94, line 92: The standard order of parameters are: input params go first, then specify the output params.

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -46,12 +47,20 @@ void NotificationsLongRunningPlatformImpl::WaitForWinMainEvent()
...
+STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::RegisterFullTrustApplication(
+    _In_ PCSTR processName, GUID remoteId, _Out_ LPWSTR* appId) noexcept
 {
     auto lock = m_lock.lock_shared();
-    RETURN_HR_IF(WPN_E_PLATFORM_UNAVAILABLE, m_shutdown);
-    
-    return E_NOTIMPL;
-}
 
+    UNREFERENCED_PARAMETER(processName);
```
danielayala94, line 56: Seems like you are planning to address this as part of your Registry work?

In the PR description, be very clear about the scope of this PR and what will be addressed later.

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -46,12 +47,20 @@ void NotificationsLongRunningPlatformImpl::WaitForWinMainEvent()
...
+
+    GUID guidReference;
+    HRESULT hr = CoCreateGuid(&guidReference);
```
danielayala94, line 59: Are you going to do anything with the HRESULT from CoCreateGuid?

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -46,12 +47,20 @@ void NotificationsLongRunningPlatformImpl::WaitForWinMainEvent()
...
+
+    StringFromCLSID(guidReference, appId);
```
danielayala94, line 61: Do we want to return in case this fails?

danielayala94 (changes requested):

dev/PushNotifications/PushNotificationManager.cpp
riverar: Recommend you use `wil::GetModuleFileName*` instead. It will handle growing the buffer size as needed to accommodate the output correctly. Then you can also delete the `1024` above.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -62,13 +65,68 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+    {
+        wchar_t appUserModelId[APPLICATION_USER_MODEL_ID_MAX_LENGTH + 1] = {};
```
riverar, line 94: This doesn't look right. `APPLICATION_USER_MODEL_ID_MAX_LENGTH` already includes space for the null terminator.

dev/PushNotifications/PushNotificationChannel.h
hulumane, line 22: Avoid abbreviations like ext in your code. This is m_expirationTime at the end of the day.

dev/PushNotifications/PushNotificationManager.cpp
hulumane: Yeah this long path that inferences debug and x64 specifically cannot be right. Please fix.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -62,13 +65,68 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+    // RegisterUnpackagedapplication
+    void PushNotificationManager::RegisterApplicationHelper(wil::unique_cotaskmem_string &unpackagedAppUserModelId, winrt::guid remoteId)
+    {
+            THROW_IF_FAILED(::CoInitializeEx(nullptr, COINITBASE_MULTITHREADED));
+
+            auto scopeExit = wil::scope_exit(
+                [&]() { CoUninitialize(); });
+
+            {
+                wil::com_ptr<INotificationsLongRunningPlatform> reunionEndpoint{
```
hulumane, line 77: This is not reunionEndpoint anymore. It is notificationPlatform

dev/PushNotifications/PushNotificationManager.cpp
hulumane, line 69: Is remoteId the PFN for the unpackaged app? Just clarrifying

dev/PushNotifications/PushNotificationManager.cpp
hulumane: Agreed with @riverar . In general, avoid hardcoded constants in your local variables.

dev/PushNotifications/PushNotificationManager.cpp
hulumane: Also needs SAL right?

dev/PushNotifications/PushNotificationManager.cpp
hulumane: Agreed. We shouldn't be adding comments that are self descriptive in code unless there are gotchas or the logic is complex in some form and needs explanation.

dev/PushNotifications/PushNotificationManager.cpp
hulumane: This definitely needs a comment unlike the prior case where it wasn't necessary : )

dev/PushNotifications/PushNotificationManager.cpp
danielayala94: @hulumane Yeah I think it's fair enough.

dev/PushNotifications/PushNotificationManager.cpp
danielayala94: I noticed that I didn't add the references to the Reunion DLL vcxproj to look at the Output directory of the LRP proxy stub. I just gave guidance to @sharath2727 to address this, it should be straightforward.

dev/PushNotifications/PushNotificationManager.cpp
sharath2727: yes pavan that is what it is.

dev/PushNotifications/PushNotificationManager.cpp
sharath2727: done

dev/PushNotifications/PushNotificationManager.cpp
sharath2727: Yeah I missed it. Thanks.

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -22,13 +22,33 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        {
+            return m_channel.ExpirationTime();
+        }
+        else
+        {
+            return m_channelExpirationTime;
+        }
     }
     void PushNotificationChannel::Close()
     {
```
loneursid, line 54: Calling this function will end up in a crash, when the object has been constructed using the new constructor. In which case m_channel is initialized to nullptr.

loneursid (changes requested):

loneursid:
I'm not a fan of the implementation for PushNotificationChannel. We have two independent sets of member variables and we use one or the other depending on the constructor called to initialize the object. Each function ends up using only one set of variables (depending which group was initialized at construction). We really have two independent implementations that share the same class, when they should really each be implemented in their own class (and have a common interface).

dev/PushNotifications/PushNotificationManager.h
```diff
@@ -21,11 +21,13 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        static void RegisterUnpackagedApplicationHelper(winrt::guid remoteId, _Out_ wil::unique_cotaskmem_string& unpackagedAppUserModelId);
         static winrt::hresult CreateChannelWithRemoteIdHelper(
-            wil::unique_cotaskmem_string& channelUri,
-            winrt::Windows::Foundation::DateTime& channelexpirytime,
-            winrt::guid remoteId);
+            winrt::guid remoteId,
+            _Out_ wil::unique_cotaskmem_string& channelUri,
+            _Out_ wil::unique_cotaskmem_string& channelId,
+            _Out_ wil::unique_cotaskmem_string& appUserModelId,
+            winrt::Windows::Foundation::DateTime& channelExpiryTime);
```
danielayala94, line 30: Nit: `_Out_`

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -62,13 +64,75 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+    void PushNotificationManager::RegisterUnpackagedApplicationHelper(
+        winrt::guid remoteId,
+        _Out_ wil::unique_cotaskmem_string &unpackagedAppUserModelId)
+    {
+            THROW_IF_FAILED(::CoInitializeEx(nullptr, COINITBASE_MULTITHREADED));
+
+            auto scopeExit = wil::scope_exit(
+                [&]() { CoUninitialize(); });
+
+            {
```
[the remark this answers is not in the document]
sharath2727: yeah we don't need it.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -91,16 +155,50 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-                winrt::PushNotificationChannel pushChannelReceived{ nullptr };
+                if (IsActivatorSupported(PushNotificationRegistrationOptions::PushTrigger))
+                {
+                    wil::unique_cotaskmem_string channelUri;
+                    winrt::Windows::Foundation::DateTime channelExpiryTime;
+                    wil::unique_cotaskmem_string channelId;
+                    wil::unique_cotaskmem_string appUserModelId;
+
+                    hresult hr = CreateChannelWithRemoteIdHelper(remoteId, channelUri, channelId, appUserModelId, channelExpiryTime);
+
```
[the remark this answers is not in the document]
sharath2727: CreateChannelWithRemoteIdHelper is taking care of the all the errors.

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -46,12 +47,21 @@ void NotificationsLongRunningPlatformImpl::WaitForWinMainEvent()
...
     auto lock = m_lock.lock_shared();
-    RETURN_HR_IF(WPN_E_PLATFORM_UNAVAILABLE, m_shutdown);
-    
-    return E_NOTIMPL;
-}
 
+    UNREFERENCED_PARAMETER(processName);
+
+    GUID guidReference;
+    THROW_IF_FAILED(CoCreateGuid(&guidReference));
```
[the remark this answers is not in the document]
sharath2727: Yes pavan. That is why if you see UNREFERENCED_PARAMETER(processName); this was temporary.This will be fixed in the upcoming commit. Thanks.

dev/PushNotifications/PushNotificationManager.cpp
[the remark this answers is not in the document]
hulumane: But thats not what this code is doing. It looks for hr = S_OK, hr = E_NOT_IMPL and it bails the if check.
What should really happen is that we should throw for errors not equal to E_NOT_IMPL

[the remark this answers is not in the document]
hulumane: The intention should be for back-off retry logic to kick in for failures that are not E_NOT_IMPL.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -62,13 +64,75 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        /*
+           As the channel request frameworkUDK API is introduced recently, it is
+           not applicable for older versions. So for error E_NOTIMPL we fallback
+           to calling into
+           CreatePushNotificationChannelForApplicationAsync
+           to request for channels
+        */
+        THROW_HR_IF(hr, (hr != E_NOTIMPL) && (hr != S_OK));
+        THROW_HR_IF(operationalCode, (hr == S_OK) && (operationalCode != S_OK));
```
hulumane, line 124: I would remove these 2 line checks and do it like this:
THROW_HR_IF(hr, FAILED(hr) && hr != E_NOT_IMPL);
THROW_IF_FAILED(operationalCode); // when we reach this line, it is assumed that hr == S_OK

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -62,13 +64,75 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+
+        return;
+    }
+
+    winrt::hresult PushNotificationManager::CreateChannelWithRemoteIdHelper(
+        winrt::guid remoteId,
+        _Out_ wil::unique_cotaskmem_string& channelUri,
+        _Out_ wil::unique_cotaskmem_string& channelId,
+        _Out_ wil::unique_cotaskmem_string& appUserModelIdString,
+        winrt::Windows::Foundation::DateTime& channelExpiryTime
```
hulumane: last one is missing SAL: channelExpiryTime

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -62,13 +64,75 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+            &channelexpirytime);
+
+        /*
+           As the channel request frameworkUDK API is introduced recently, it is
+           not applicable for older versions. So for error E_NOTIMPL we fallback
+           to calling into
+           CreatePushNotificationChannelForApplicationAsync
+           to request for channels
+        */
+        THROW_HR_IF(hr, (hr != E_NOTIMPL) && (hr != S_OK));
```
hulumane, line 111: I think these THROW_HR_IF conditions aren't well derfined. We should do something like this instead in the following order:

THROW_IF_FAILED(hr);

Let the upper layer caller decide how to act on E_NOT_IMPL or any other failed HRESULT's.

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -46,12 +47,58 @@ void NotificationsLongRunningPlatformImpl::WaitForWinMainEvent()
     m_shutdownTimerManager->Wait();
 }
 
-// Example of one function. We will add more as we need them.
-STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::RegisterFullTrustApplication(_In_ PCWSTR /*processName*/, _In_ GUID /*remoteId*/, _Out_ GUID* /*appId*/) noexcept
+void NotificationsLongRunningPlatformImpl::AddToRegistry(const std::wstring& processName, std::wstring appId)
```
hulumane, line 50: appId should also be const&?

dev/PushNotifications/PushNotificationChannel.h
```diff
@@ -18,6 +19,10 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
 
     private:
         const winrt::Windows::Networking::PushNotifications::PushNotificationChannel m_channel{ nullptr };
+        winrt::Windows::Foundation::DateTime m_channelExpirationTime{};
```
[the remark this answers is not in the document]
hulumane: This wasn't condensed. Is that coming in another iteration?

dev/PushNotifications/PushNotificationsLongRunningTask.ProxyStub/NotificationsLongRunningProcess.idl
```diff
@@ -11,7 +11,7 @@ import "ocidl.idl";
 [pointer_default(unique)]
 interface INotificationsLongRunningPlatform : IUnknown
 {
-    HRESULT RegisterFullTrustApplication([in] LPCWSTR processName, [in] GUID remoteId, [out] GUID* appId);
+    HRESULT RegisterFullTrustApplication([in] LPCWSTR processName, [in] GUID remoteId, [out] LPWSTR* appId);
```
hulumane, line 14: PWSTR. Not LPWSTR

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -46,12 +47,58 @@ void NotificationsLongRunningPlatformImpl::WaitForWinMainEvent()
...
+        nullptr,
+        appId.c_str(),
+        REG_SZ,
+        processName.c_str(),
+        static_cast<DWORD>(sizeof(WCHAR) * processName.size())));
+}
+
+void NotificationsLongRunningPlatformImpl::GetAppIdentifier(const std::wstring processName)
+{
+    if (m_appIdMap.find(processName) == m_appIdMap.end())
```
hulumane, line 78: Why are you storing this in a map? You should directly be querying the registry to see if the mapping can be found. If not, Create a new GUID and save it in the registry.

dev/PushNotifications/PushNotificationsLongRunningTask/platform.h
```diff
@@ -28,8 +28,11 @@ Microsoft::WRL::RuntimeClass<
 
     bool m_initialized = false;
     bool m_shutdown = false;
+    std::map<std::wstring, std::wstring> m_appIdMap;
```
hulumane, line 31: Please get rid of this. Its not needed. You are querying registry.

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -46,12 +47,58 @@ void NotificationsLongRunningPlatformImpl::WaitForWinMainEvent()
...
+
+        AddToRegistry(processName, guidStr.get());
+        m_appIdMap[processName] = guidStr.get();
+    }
 }
 
+STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::RegisterFullTrustApplication(
+    _In_ PCWSTR processName, GUID remoteId, _Out_ PWSTR* appId) noexcept try
+{
+    auto lock = m_lock.lock_shared();
```
hulumane, line 94: @danielayala94 What is the guidance for making this threadsafe per your design? I think this should check for the state of m_shutdown right? Please take a look and advise.

hulumane: Also I think this should be a write lock and not a read lock since it is modifying a registry based mapping.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -62,13 +64,63 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
         }
     }
 
+    void PushNotificationManager::RegisterUnpackagedApplicationHelper(
+        const winrt::guid& remoteId,
+        _Out_ wil::unique_cotaskmem_string &unpackagedAppUserModelId)
+    {
+        wchar_t processName[1024];
```
jonwis, line 71: This pattern showed up in your other PR.. Use `wil::GetModuleFileNameExW` instead of a giant stack buffer.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -62,13 +64,63 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        THROW_HR_IF(ERROR_FILE_NOT_FOUND, GetModuleFileNameExW(GetCurrentProcess(), NULL, processName, sizeof(processName) / sizeof(processName[0])) == 0);
+
+        THROW_IF_FAILED(::CoInitializeEx(nullptr, COINITBASE_MULTITHREADED));
```
jonwis, line 74: See `wil::CoInitializeEx` - https://github.com/microsoft/wil/blob/492c01bb535daadf719d4445d6107aadf1e60812/include/wil/resource.h#L4347

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -62,13 +64,63 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+
+        auto scopeExit = wil::scope_exit(
+            [&]() { CoUninitialize(); });
+
+        wil::com_ptr<INotificationsLongRunningPlatform> notificationPlatform{
```
jonwis, line 79: Since you named the interface in the template parameter, you can just say `auto notificationPlatform { ... }` here

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -62,13 +64,63 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+            wil::CoCreateInstance<NotificationsLongRunningPlatform, INotificationsLongRunningPlatform>(CLSCTX_LOCAL_SERVER) };
+
+        winrt::check_hresult(notificationPlatform->RegisterFullTrustApplication(processName, remoteId, &unpackagedAppUserModelId));
```
jonwis, line 82: Don't mix hresult helpers; the rest of this file uses WIL, so `THROW_IF_FAILED`

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -62,13 +64,63 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+    }
+
+    winrt::hresult PushNotificationManager::CreateChannelWithRemoteIdHelper(const winrt::guid& remoteId, ChannelDetails& channelInfo)
```
jonwis, line 85: Should this method be returning a `std::optional<ChannelDetails>` instead of using an out-reference-param and returning an hresult?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -62,13 +64,63 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        THROW_HR_IF(operationalCode, (hr == S_OK) && (operationalCode != S_OK));
+
+        winrt::copy_from_abi(channelInfo.channelExpiryTime, &channelexpirytime);
+        channelInfo.appUserModelId = wil::make_cotaskmem_string(appUserModelId);
+
+        return hr;
+    }
+
     winrt::IAsyncOperationWithProgress<winrt::Microsoft::Windows::PushNotifications::PushNotificationCreateChannelResult, winrt::Microsoft::Windows::PushNotifications::PushNotificationCreateChannelStatus> PushNotificationManager::CreateChannelAsync(const winrt::guid &remoteId)
     {
```
jonwis, line 121: In nearly all `...Async` methods, you need to "capture a strong this" with this magic dance at the start of the method:

```c++
auto strong = get_strong();
```

... that ensures `this` has a reference after the `co_await` points below.

dev/PushNotifications/PushNotificationManager.h
```diff
@@ -3,9 +3,17 @@
 
 #pragma once
 #include "Microsoft.Windows.PushNotifications.PushNotificationManager.g.h"
+#include <windows.foundation.h>
 
 namespace winrt::Microsoft::Windows::PushNotifications::implementation
 {
+    struct ChannelDetails {
```
jonwis, line 10: `{` goes on a new line

dev/PushNotifications/PushNotificationManager.h
```diff
@@ -20,6 +28,8 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
     private:
         static bool IsChannelRequestRetryable(const winrt::hresult& hrException);
         static bool IsBackgroundTaskBuilderAvailable();
+        static void RegisterUnpackagedApplicationHelper(const winrt::guid& remoteId, _Out_ wil::unique_cotaskmem_string& unpackagedAppUserModelId);
+        static winrt::hresult CreateChannelWithRemoteIdHelper(const winrt::guid& remoteId, ChannelDetails& channelInfo);
```
jonwis, line 32: Don't put `private static` methods in the header. Instead, just declare them in the .cpp file as loose functions.

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -46,12 +47,58 @@ void NotificationsLongRunningPlatformImpl::WaitForWinMainEvent()
...
+    THROW_IF_WIN32_ERROR(RegSetKeyValue(
+        hKeyResult.get(),
+        nullptr,
+        appId.c_str(),
+        REG_SZ,
+        processName.c_str(),
+        static_cast<DWORD>(sizeof(WCHAR) * processName.size())));
+}
+
+void NotificationsLongRunningPlatformImpl::GetAppIdentifier(const std::wstring processName)
```
jonwis, line 76: "Get-" methods usually return something. Maybe this is "Load-" instead? Or maybe this method returns the GUID string it constructed?

jonwis (changes requested):

dev/PushNotifications/PushNotificationsLongRunningTask.ProxyStub/NotificationsLongRunningProcess.idl
```diff
@@ -19,14 +19,16 @@ interface IWpnForegroundSink : IUnknown
 [pointer_default(unique)]
 interface INotificationsLongRunningPlatform : IUnknown
 {
-    HRESULT RegisterFullTrustApplication([in] LPCWSTR processName, [in] GUID remoteId, [out] GUID* appId);
+
+    HRESULT RegisterFullTrustApplication([in] LPCWSTR processName, [in] GUID remoteId, [out] LPWSTR* appId);
+
+    //HRESULT RegisterFullTrustApplication([in] LPCWSTR processName, [in] GUID remoteId, [out] GUID* appId);
```
danielayala94, line 25: Why is there a commented signature for RegisterFullTrustApplication?

loneursid: That's my fault. A bad operation when merging WNP_LRP into this branch. It should have been deleted.

dev/PushNotifications/PushNotificationManager.h
pmpurifoy: Moved to externs.h

dev/PushNotifications/PushNotificationsLongRunningTask.ProxyStub/NotificationsLongRunningProcess.idl
pmpurifoy: PWSTR can't used as a parameter. You would have to use [out, string] PWSTR.

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: PushNotificationManager is a static runtime class. Do we have to get a strong reference to PushNotificationManager if it is static?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -62,13 +64,63 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        wil::com_ptr<INotificationsLongRunningPlatform> notificationPlatform{
+            wil::CoCreateInstance<NotificationsLongRunningPlatform, INotificationsLongRunningPlatform>(CLSCTX_LOCAL_SERVER) };
+
+        winrt::check_hresult(notificationPlatform->RegisterFullTrustApplication(processName, remoteId, &unpackagedAppUserModelId));
+    }
+
+    winrt::hresult PushNotificationManager::CreateChannelWithRemoteIdHelper(const winrt::guid& remoteId, ChannelDetails& channelInfo)
+    {
+        wchar_t appUserModelId[APPLICATION_USER_MODEL_ID_MAX_LENGTH] = {};
+        UINT32 appUserModelIdSize = ARRAYSIZE(appUserModelId);
```
DrusTheAxe, line 88: NIT: `UINT32 appUserModelIdSize{ ARRAYSIZE(appUserModelId) };`

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -62,13 +64,63 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+
+        THROW_IF_FAILED(GetCurrentApplicationUserModelId(&appUserModelIdSize, appUserModelId));
+
+        THROW_HR_IF(E_INVALIDARG, (appUserModelIdSize > APPLICATION_USER_MODEL_ID_MAX_LENGTH) || (appUserModelIdSize == 0));
+
+        HRESULT operationalCode;
```
DrusTheAxe, line 94: NIT: `HRESULT operationalCode{};`

Always initialize variables

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -62,13 +64,63 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+           Public WinRT API CreatePushNotificationChannelForApplicationAsync
+           to request for channels
+        */
+        THROW_HR_IF(hr, (hr != E_NOTIMPL) && (hr != S_OK));
+        THROW_HR_IF(operationalCode, (hr == S_OK) && (operationalCode != S_OK));
+
+        winrt::copy_from_abi(channelInfo.channelExpiryTime, &channelexpirytime);
+        channelInfo.appUserModelId = wil::make_cotaskmem_string(appUserModelId);
+
+        return hr;
```
DrusTheAxe, line 117: It's weird to see a function both THROW_HR and return hr

Best to pick one error strategy for a function and stick with it

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -91,16 +143,46 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                    {
+                        co_return winrt::make<PushNotificationCreateChannelResult>(
+                            winrt::make<PushNotificationChannel>(channelInfo.channelUri.get(), channelInfo.channelId.get(),  channelInfo.appUserModelId.get(), channelInfo.channelExpiryTime),
+                            hr,
+                            PushNotificationChannelStatus::CompletedSuccess);
+                    }
+                    else if (hr == E_NOTIMPL)
+                    {
+                        PushNotificationChannelManager channelManager{};
+                        winrt::PushNotificationChannel pushChannelReceived{ nullptr };
```
DrusTheAxe, line 161: NIT: do you need `nullptr` or will void-constructor do the equivalent?

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -46,14 +47,61 @@ void NotificationsLongRunningPlatformImpl::WaitForWinMainEvent()
     m_shutdownTimerManager->Wait();
 }
 
-// Example of one function. We will add more as we need them.
-STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::RegisterFullTrustApplication(_In_ PCWSTR /*processName*/, _In_ GUID /*remoteId*/, _Out_ GUID* /*appId*/) noexcept
+void NotificationsLongRunningPlatformImpl::AddToRegistry(const std::wstring& processName, std::wstring appId)
+{
+    std::wstring subkey = L"Software\\Microsoft\\Windows\\CurrentVersion\\PushNotifications\\LongRunningProcess";
```
DrusTheAxe, line 52: `PCWSTR`

You make a string here (ctor, malloc, memcpy) so you can just .c_str() and then destroy it (dtor, free). Expensive way to have a string literal.

C++ is cool, but C++ isn't always the best answer ;-)

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: This variable should just be set. It immediately gets assigned in the next line.

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
pmpurifoy: Using ApplicationData instead of Registry. Removed.

dev/PushNotifications/PushNotificationsLongRunningTask.ProxyStub/NotificationsLongRunningProcess.idl
danielayala94: What's wrong with adding [string]?

dev/PushNotifications/PushNotificationChannel.h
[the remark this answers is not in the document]
sharath2727: done

dev/PushNotifications/PushNotificationManager.cpp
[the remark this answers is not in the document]
sharath2727: I have addressed this now in the caller. Please re-review pavan. Thanks.

dev/PushNotifications/PushNotificationManager.cpp
sharath2727: https://github.com/microsoft/WindowsAppSDK/issues/927 - I have an issue tracking this already. I will address this a little later as it needs further discussion. Thanks.

dev/PushNotifications/PushNotificationManager.cpp
sharath2727: For Now I would like to go by the returning hresult here as based on that - whether it is S_OK, E_NOTIMPL or other errors I prefer the caller to make decisions.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -94,20 +143,60 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                                remoteId);
+
+                            co_return winrt::make<PushNotificationCreateChannelResult>(
+                                winrt::make<PushNotificationChannel>(pushChannelReceived),
+                                S_OK,
+                                PushNotificationChannelStatus::CompletedSuccess);
+                        }
+                        else
+                        {
+                            winrt::check_hresult(hr);
```
hulumane, line 184: Don't we want to throw here? So that the back-off retry logic fires?

hulumane: Ah never mind this macro throws. We should be good then.

hulumane (approved):

dev/PushNotifications/PushNotificationManager.cpp
danielayala94: I have opened a bug for this -> #1365

dev/PushNotifications/PushNotificationManager.cpp
jonwis: This was resolved, but without a change to `wil::CoInitializeEx(...)` - intentional?

jonwis (approved):

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -45,58 +46,60 @@ void NotificationsLongRunningPlatformImpl::WaitForLifetimeEvent()
...
+        }
     }
+
+    GUID guidReference;
+    THROW_IF_FAILED(CoCreateGuid(&guidReference));
+
+    wil::unique_cotaskmem_string guidStr;
+    THROW_IF_FAILED(StringFromCLSID(guidReference, &guidStr));
+
+    m_storage.Values().Insert(guidStr.get(), winrt::box_value(processName.c_str()));
```
danielayala94, line 69: nit: can this just be `values.Insert(guidStr.get(), winrt::box_value(processName.c_str()));`

No need to address though.

dev/PushNotifications/PushNotificationManager.cpp
danielayala94: I'm taking this. I'm gonna send a new iteration.

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
danielayala94: Might take this in the big PR.

danielayala94 (approved):

PR #1298, opened 2021-08-24
Add HybridCRT docs

DrusTheAxe:
Docs on why and how we use the hybrid CRT


docs/Coding-Guidelines.md
```diff
@@ -20,7 +36,7 @@ Windows-specific helpers rather than creating your own.
 
 **DO** use 4-space indentation instead of tab characters.
 
-##### Catching C++/WinRT Exceptions and HRESULT
+##### 1.1.1.0.1. Catching C++/WinRT Exceptions and HRESULT
```
jeffstall, line 39: This is worse than an IP address. Perhaps have separate docs for how to write C++ code vs. C# code vs. formatting vs. how to properly configure the CRT vs. all of Reunion. I don't think merging everything into one doc is helping readability.

jeffstall (approved):
:shipit:

docs/Coding-Guidelines.md
DrusTheAxe: Snicker. It did point out how we borked the header styles. I fixed that.

As for one doc...'twas only a matter of time. We talked about splitting out content to separate docs and this becomes more of an index/links-to-everything-else. Looks like the day has come. Splitting up...

DrusTheAxe: Ahhhh. Much better, and room to grow. Thanks Jeff!

docs/Coding-Guidelines/HybridCRT.md
```diff
@@ -0,0 +1,133 @@
...
+containing C/C++ code are built using the 'hybrid CRT' technique.
+
+## What is the 'Hybrid CRT'?
+
+The 'Hybrid CRT' is a technique using the
+[Universal CRT](https://docs.microsoft.com/en-us/cpp/windows/universal-crt-deployment?view=msvc-160)
+AND the static CRT to get functional coverage without the overhead of the static CRT or the external
+dependency of the dynamic CRT.
+
+Visual Studio offers the CRT in 2: DLLs needed at runtime (dynamic) and static
```
riverar, line 15: Nit: in 2 what?

DrusTheAxe: Doh! Thanks

BenJKuhn: This probably warrants a bit of explanation. @DrusTheAxe, this probably should be a footnote. 

The CRT, depending on who you ask, includes up to about a dozen DLLs. The most commonly used are ucrtbase.dll and vcruntime.dll. ucrtbase.dll contains most of the common "C" functions implemented in the CRT, has a binary-stable interface, and is provided in the OS. Binary-stable in this case means that code linked to run against earlier versions of the DLL will still run against newer versions. vcruntime###.dll is coupled to the compiler version. It contains compiler-support functions that are implicitly added to the generated code. For instance, stack unwind is implemented in vcruntime in __CxxFrameHandler4. Calls from a client binary to the frame handler are generated by the compiler itself behind the scenes. You would never write code to call this directly.

There are also a number of C++ support DLL's for STL, C++ CX, PPL, MFC, and other features. Those modules all have non-binary stable interfaces. So compiled code needs to use the DLL version that matches the compiler.

In static linking, a binary with application code also contains the code for the C runtime functions, vcruntime code, and other CRT related support libraries. This leads to a larger binary with a larger working set, but it's fully self-contained. No need to redist the CRT. This is commonly used for binaries that can be copied around with no installation. Since the linker can discard unused code, it's somewhat pay-for-play. You don't get everything, though sometimes it may include more code than is actually reachable because of the limits of dependency analysis.

In dynamic linking, all of these are referenced in external DLLs, with UCRTBase coming from the OS, and vcruntime and others are provided via either a framework package or the VC redist. The result is that many DLL's can share the same instance of the code, and get size and performance advantages.

Hybrid CRT is a compromise. vcruntime, STL support, etc. are linked into the application binary, but ucrtbase.dll is referenced dynamically and supplied by the OS. Hybrid linking is a good bit smaller than fully static linkage (typically around 1/3 the overhead, IIRC) and gets the benefit of using the shared copy of ucrtbase.dll. It also avoids the need to use the vcredist or framework packages for the CRT.

For App SDK, this approach avoids potential headaches with CRT references coming from both the App SDK and the application using it while minimizing the performance and binary size overhead of doing so.

Ben

BenJKuhn: And I just read past the comment and see much of this is captured already in some detail. Oops. I see this is really just asking for the wording to be better (hence the "nit"). The CRT is many DLLs, but for this discussion we can primarily focus on vcruntime___.dll (C/C++ runtime support functions), ucrtbase.dll (ANSI C functions, mostly), and maybe msvcp140.dll (STL support). The others are more obscure and don't impact the substance of the discussion.

BenJKuhn (approved):

PR #1321, opened 2021-08-26
Prototype implementation of Push Notification Telemetry

loneursid:
This is an early attemp at implement telemetry for PushNotifications.

I've added 4 telemetry events (2 making use of a correlation vector) and verified that they where getting logged correctly (including the cv and the PartB data specific to WindowsAppSDK).

There are still a number of issues to be resolved:
1. How to correctly link with the WindowsAppSDKInsight project? Where does that project reside?
   1.1. For now, I've added the WindowsAppSDKInsight header to the comon folder but this is probably not the correct approach as I had to define the two functions that were declared in that header, because the linker couldn't fint their implementations (I thought Telemetry was all implemented in header files!)
2. How does one obtain a correletion vector from the WindowsAppSDK and how will this correlation vector be kept in sink across the WindowsAppSDK and the CoreOS?


loneursid:
Thanks everyone for the feedback. There's a new PR now open for the actual feature implementation.

PR #1331, opened 2021-08-27
Adding Telemetry to PowerNotifications

huzaifa-d:


huzaifa-d:
\azp run

dev/PowerNotifications/PowerNotifications.h
```diff
@@ -7,8 +7,71 @@
...
+    IMPLEMENT_TRACELOGGING_CLASS(PowerNotifications, "Microsoft.WindowsAppSDK.System.PowerNotifications",
+        // {a1b12e2c-12d9-564e-2ea1-2894ffcc7cc5}
+        (0xa1b12e2c, 0x12d9, 0x564e, 0x2e, 0xa1, 0x28, 0x94, 0xff, 0xcc, 0x7c, 0xc5));
+
+    public:
+
+        DEFINE_COMPLIANT_TELEMETRY_EVENT_STRING(RegisterEvent, PDT_ProductAndServiceUsage, value);
+        DEFINE_COMPLIANT_TELEMETRY_EVENT_STRING(UnregisterEvent, PDT_ProductAndServiceUsage, value);
+
+        DEFINE_COMPLIANT_MEASURES_ACTIVITY(EnergySaverStatusCallbackActivity, PDT_ProductAndServicePerformance);
```
ZenBird, line 23: > DEFINE_COMPLIANT_MEASURES_ACTIVITY [](http://example.com/codeflow?start=7&length=35)

Why are these activities and not just events?

huzaifa-d: Because they are at the boundary of external/OS API calls and we wish to log it if it fails

dev/PowerNotifications/PowerNotifications.h
```diff
@@ -7,8 +7,71 @@
...
+
+class PowerNotifications : public wil::TraceLoggingProvider
+{
+    IMPLEMENT_TRACELOGGING_CLASS(PowerNotifications, "Microsoft.WindowsAppSDK.System.PowerNotifications",
+        // {a1b12e2c-12d9-564e-2ea1-2894ffcc7cc5}
+        (0xa1b12e2c, 0x12d9, 0x564e, 0x2e, 0xa1, 0x28, 0x94, 0xff, 0xcc, 0x7c, 0xc5));
+
+    public:
+
+        DEFINE_COMPLIANT_TELEMETRY_EVENT_STRING(RegisterEvent, PDT_ProductAndServiceUsage, value);
```
brialmsft, line 20: Minor: consider changing from "value" to something more descriptive (e.g. functionName, dataName, etc.)

brialmsft: Also, isn't capturing the name explicitly redundant if the event is happening in the context of an activity that's going to have the same information? If RegisterEvent fires with "EnergySaverStatus" as its parameter, and that happens under an activity named "EnergySaverStatusRegisterActivity", the parameter may not be doing much. Unless there's some reason that the activity name is harder to use when processing the telemetry for some reason (most of my experience is with deciphering raw events in ETL files, rather than using the full telemetry pipeline.

brialmsft (approved):

dev/PowerNotifications/PowerNotifications.h
huzaifa-d: Changed. 
For the other comment, I am using Activities to log API calls and not all of the Register Activities will result in a registration on the FrameworkUDK side. So I added the events to explicitly log it and differentiate. Hope this makes sense.

ZenBird (approved):
:shipit:

dev/PowerNotifications/PowerNotifications.h
brialmsft: > not all of the Register Activities will result in a registration on the FrameworkUDK side
I'm not questioning whether the event is necessary, I'm asking whether the *parameter* of the event adds any information.
The key question for whether the parameter is redundant is "is it possible for AddCallback/RemoveCallback to be called outside the context of one of these entrypoint activities?". If not, then the event should always have an activity tag that embeds the same information that's in the parameter.

dev/PowerNotifications/PowerNotifications.h
```diff
@@ -467,22 +566,26 @@ namespace winrt::Microsoft::Windows::System::Power
             // DisplayStatus Functions
             Power::DisplayStatus DisplayStatus()
             {
+                auto activity = PowerNotifications::DisplayStatusUpdateActivity::Start();
```
DrusTheAxe, line 569: Are there perf implications to all these WIL activities? How often do these fire, and how much work happens? "Lots" and "little" would make the activity overhead significant.

huzaifa-d: I have added an Activity for every API call. ie for every Power API set like PowerStatus an activity for register, unregister, update value and callback. I haven't done telemetry before so don't know if this is enough/more than needed. @andreww-msft thoughts?
@ZenBird could you answer the perf query?

ZenBird: activities are indeed expensive. I would rather log events with HR and wrap the functions in a try-catch block and log the failing hr from there.

dev/PowerNotifications/PowerNotifications.h
```diff
@@ -281,71 +235,103 @@ namespace winrt::Microsoft::Windows::System::Power
...
-                std::scoped_lock<std::mutex> lock(m_mutex);
-                if (!RegisteredForEvents(eventObj))
+                try
+                {
+                    PowerNotifications::AddCallbackEvent(fn.name.c_str());
+                    auto& eventObj{ fn.event() };
+                    std::scoped_lock<std::mutex> lock(m_mutex);
+                    if (!RegisteredForEvents(eventObj))
+                    {
+                        PowerNotifications::RegisterEvent(fn.name.c_str());
```
brialmsft, line 245: Readability feedback: because of the way things are named, it really sounds like RegisteredForEvents and RegisterEvent are related to each other; it's not clear on reading this that RegisteredForEvents is functional, but RegisterEvent is telemetry. Might consider massaging the naming.

brialmsft: Maybe TraceRegister instead of RegisterEvent? I think verb naming fits better than noun naming here anyway.

huzaifa-d: Great point. Changed.

brialmsft: Oh yeah, changing the class name was also a good call.

brialmsft (approved):

dev/PowerNotifications/PowerNotifications.h
```diff
@@ -7,11 +7,30 @@
...
+    DEFINE_COMPLIANT_TELEMETRY_EVENT_STRING(RegisterTrace, PDT_ProductAndServiceUsage, eventName);
+    DEFINE_COMPLIANT_TELEMETRY_EVENT_STRING(UnregisterTrace, PDT_ProductAndServiceUsage, eventName);
+    DEFINE_COMPLIANT_TELEMETRY_EVENT_PARAM3(FailureTrace, PDT_ProductAndServiceUsage,
+        PCWSTR, eventName, PCWSTR, APIName, PCSTR, exceptionString);
+};
 
 namespace winrt::Microsoft::Windows::System::Power
 {
+
+
```
ZenBird, line 33: nit: Remove spaces

dev/PowerNotifications/PowerNotifications.h
andreww-msft: Re activities: yes that seems like overkill. ZenBird's suggestion makes sense to reduce the perf hit.

dev/PowerNotifications/PowerNotifications.h
```diff
@@ -7,11 +7,30 @@
...
+class PowerNotificationsTelemetry : public wil::TraceLoggingProvider
+{
+    IMPLEMENT_TRACELOGGING_CLASS(PowerNotificationsTelemetry, "Microsoft.WindowsAppSDK.System.PowerNotifications",
+        // {a1b12e2c-12d9-564e-2ea1-2894ffcc7cc5}
+        (0xa1b12e2c, 0x12d9, 0x564e, 0x2e, 0xa1, 0x28, 0x94, 0xff, 0xcc, 0x7c, 0xc5));
+
+public:
+
+    DEFINE_COMPLIANT_TELEMETRY_EVENT_STRING(AddCallbackTrace, PDT_ProductAndServiceUsage, eventName);
```
jvarela64, line 20: All of these TELEMETRY macros define a Telemetry-sampling level event. This needs to be at MEASURES sampling level so it fires. This would be the macros: DEFINE_COMPLIANT_MEASURES_EVENT_STRING (or similar).
@huzaifa-d we can work together to place the correct event sampling level for each event.

jvarela64: Although, this set of events seem to complement information together, so they should all be MEASURES.

jvarela64 (changes requested):
My comments are only on the telemetry level when defining events. For OKRs, understanding what apps are registering for notifications works well.

dev/PowerNotifications/PowerNotifications.h
jvarela64: Correct, Telemetry sampling is very seldom used. It even has to be turned on OTA, temporarily in retail, so the use case is almost extra diagnostic info after uncovering bugs. Standard use for any telemetry that we know we will consume is Measures-level. There is internal wiki information.

dev/PowerNotifications/PowerNotifications.h
```diff
@@ -206,42 +237,80 @@ namespace winrt::Microsoft::Windows::System::Power
...
-                    fn.unregisterListener();
+                    PowerNotificationsTelemetry::FailureTrace(fn.name.c_str(), L"RemoveCallback", ex.what());
+                    throw ex;
                 }
             }
 
             void RaiseEvent(PowerFunctionDetails fn)
             {
-                std::thread thread([fn]() {
-                    fn.event()(nullptr, nullptr);
```
jonwis, line 232: https://github.com/microsoft/cppwinrt/pull/858#discussion_r583120467 - there's a way to do this using C++/winrt and coroutines, so your methods become:

```c++
winrt::fire_and_forget RaiseEvent(PowerFunctionDetails fn)
{
    auto lifetime = get_strong();
    fn.event()(nullptr, nullptr);
}
```

dev/PowerNotifications/PowerNotifications.h
```diff
@@ -206,42 +237,80 @@ namespace winrt::Microsoft::Windows::System::Power
...
             event_token AddCallback(PowerFunctionDetails fn, const PowerEventHandler& handler)
             {
-                auto& eventObj{ fn.event() };
-                std::scoped_lock<std::mutex> lock(m_mutex);
-                if (!RegisteredForEvents(eventObj))
+                try
                 {
-                    fn.registerListener();
+                    PowerNotificationsTelemetry::AddCallbackTrace(fn.name.c_str());
+                    auto& eventObj{ fn.event() };
```
jonwis, line 243: What's motivating all these try/catch handlers? My estimation is that this should be moved out of here so there's a noexcept method like:

```c++
void PowerNotificationsTelemetry::AddCallbackTrace(PowerFunctionDetails const& fn) noexcept try
{
 
}
catch(...)
{
    FailureTrace(/* whatever */)
}
```

So the code here remains just `PowerNotificationsTelemetry::AddCallbackTrace(fn)` ... put the error handling where it belongs (in the place that generates errors) and not here.

jonwis (changes requested):
Generally OK, but maybe there's a way to eliminate the try/catch madness.

dev/PowerNotifications/PowerNotifications.h
huzaifa-d: Just curious: is this necessarily better?

dev/PowerNotifications/PowerNotifications.h
huzaifa-d: Kept only one instance.

dev/PowerNotifications/PowerNotifications.h
huzaifa-d: No longer applicable as we now only use two events.

huzaifa-d:
Moving over to https://github.com/microsoft/WindowsAppSDK/pull/1379 due to merge issues.

PR #1332, opened 2021-08-27
Fix MRT Core binary vulcanization for API scanning.

jefgen:
We need to set additional Linker options in order to Vulcanize to "Level 3" for the API scanning tools.

**How verified:**
Alex was able to run the API Scan tool locally with private binaries and they now work with the tool instead of generating an error.
(Thanks Alex!)

**Testing**
Used the MaxDep tool with the level 3 option and it is able to load the DLL/PDB files without error(s).


rohanp-msft (approved):
:shipit:

dev/MRTCore/mrt/Core/src/MRM.vcxproj
```diff
@@ -125,6 +125,8 @@
       <LinkIncremental>false</LinkIncremental>
       <GenerateDebugInformation>true</GenerateDebugInformation>
```
huichen123, line 126: no longer neede?

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/src/Microsoft.ApplicationModel.Resources.vcxproj
```diff
@@ -154,6 +154,8 @@
       <OptimizeReferences>true</OptimizeReferences>
       <LinkIncremental>false</LinkIncremental>
       <GenerateDebugInformation>true</GenerateDebugInformation>
```
huichen123, line 156: and this?

dev/MRTCore/mrt/Core/src/MRM.vcxproj
```diff
@@ -125,6 +125,8 @@
...
       <!-- EnableCOMDATFolding and OptimizeReferences are set to true below for all build configurations. -->
```
rohanp-msft, line 127: >below [](http://example.com/codeflow?start=70&length=5)

This is my mistake I think, but since you're here, can you change this to above?

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/src/Microsoft.ApplicationModel.Resources.vcxproj
```diff
@@ -154,6 +154,8 @@
       <OptimizeReferences>true</OptimizeReferences>
```
huichen123, line 154: ><OptimizeReferences>true</OptimizeReferences> [](http://example.com/codeflow?start=6&length=45)

also this?

huichen123:
>       

this needs to move down

---
Refers to: dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/src/Microsoft.ApplicationModel.Resources.vcxproj:152 in 61aa2d9. [](commit_id = 61aa2d9cf46ef30ada825bf867f83d102668fc3a, deletion_comment = False)

dev/MRTCore/mrt/Core/src/MRM.vcxproj
jefgen: IIUC, I think we still need this as it sets `/DEBUG`.

Oh, I think you're saying that the Release build doesn't need this since all configurations get it...?

huichen123:
>       <EnableCOMDATFolding>true</EnableCOMDATFolding>

also this

---
Refers to: dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/src/Microsoft.ApplicationModel.Resources.vcxproj:153 in 61aa2d9. [](commit_id = 61aa2d9cf46ef30ada825bf867f83d102668fc3a, deletion_comment = False)

dev/MRTCore/mrt/Core/src/MRM.vcxproj
huichen123: based on document, /Profile seems enough. not sure though

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/src/Microsoft.ApplicationModel.Resources.vcxproj
```diff
@@ -154,6 +154,8 @@
...
       <LinkIncremental>false</LinkIncremental>
       <GenerateDebugInformation>true</GenerateDebugInformation>
+      <!-- /PROFILE implies /OPT:REF /OPT:NOICF /INCREMENTAL:NO /FIXED:NO -->
+      <Profile>true</Profile>
```
rohanp-msft, line 158: ><Profile>true</Profile> [](http://example.com/codeflow?start=6&length=23)

Not relevant for API scanning, but I noticed this:
For this DLL, unlike mrm.dll, we're not setting `EnableCOMDATFolding`, `OptimizeReferences` and `GenerateDebugInformation` to `true` for Debug. Should we?

dev/MRTCore/mrt/Core/src/MRM.vcxproj
jefgen: IIUC, I think we need both /DEBUG and /PROFILE, as that causes /DEBUGTYPE:CV,FIXUP to be set.
(I suppose we could also set them manually as well if we wanted to be sure.)

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/src/Microsoft.ApplicationModel.Resources.vcxproj
rohanp-msft: This too is separate from your change. So feel free to ignore the comment.

dev/MRTCore/mrt/Core/src/MRM.vcxproj
jefgen: Sure, no problem. Thanks for noticing it!

dev/MRTCore/mrt/Core/src/MRM.vcxproj
huichen123: I really wish we have a simpler way to verify.

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/src/Microsoft.ApplicationModel.Resources.vcxproj
huichen123: i think "<EnableCOMDATFolding>true</EnableCOMDATFolding>
      <OptimizeReferences>true</OptimizeReferences>" are wrong flags.

dev/MRTCore/mrt/Core/src/MRM.vcxproj
jefgen: Hm, re-reading the document: IIUC, setting `Profile=True` implies `EnableCOMDATFolding=No` and `OptimizeReferences=Yes`. 
But they recommend enabling `EnableCOMDATFolding`. But this can only be done by overriding the Profile settings.
So we'll need to add additional linker commands.

jefgen: For testing I think we can use maxdep with -l3 and see if it reports any errors.

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/src/Microsoft.ApplicationModel.Resources.vcxproj
jefgen: I pushed additional changes with some more clean-up. Hopefully it makes things a bit clearer.

FWIW, I don't think we want to `EnableCOMDATFolding` for Debug builds, based on the comments on the page here:
https://docs.microsoft.com/en-us/cpp/build/reference/opt-optimizations?view=msvc-160

(Note: `/OPT:ICF` is `EnableCOMDATFolding=true`).

> /OPT:ICF can merge identical data or functions, it can change the function names that appear in stack traces. It can also make it impossible to set breakpoints in certain functions or to examine some data in the debugger, and can take you into unexpected functions when you single-step through your code. The behavior of the code is identical, but the debugger presentation can be very confusing. Therefore, we do not recommend that you use /OPT:ICF in debug builds unless the advantages of smaller code outweigh these disadvantages.

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/src/Microsoft.ApplicationModel.Resources.vcxproj
```diff
@@ -124,22 +104,32 @@
       <GenerateWindowsMetadata>false</GenerateWindowsMetadata>
       <ModuleDefinitionFile>Microsoft.ApplicationModel.Resources.def</ModuleDefinitionFile>
       <AdditionalDependencies>$(OutDir)..\mrm\MRM.lib;onecoreuap.lib;%(AdditionalDependencies)</AdditionalDependencies>
+      <!-- This sets /DEBUG -->
+      <GenerateDebugInformation>true</GenerateDebugInformation>
+      <!-- Note: /PROFILE implies the following 4 settings.
+           However, since /PROFILE is only supported on VS Enterprise we will also set them explicitly below as well. -->
```
huichen123, line 110: >we will also set them explicitly below as well [](http://example.com/codeflow?start=70&length=46)

why? does /Profile not work in the APIScan task in the build pipeline?

jefgen: The build pipeline uses VS Enterprise, so it's not an issue.
I wanted to set them here in case an external person is building with VS Community edition.
I think we'd still want to set `EnableCOMDATFolding=false` so they have a good debugging experience and also so that we don't end up with differences in the binary based on what edition of VS you use to build.

huichen123: don't think it's a big deal. if you want that, just don't set profile.

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/src/Microsoft.ApplicationModel.Resources.vcxproj
```diff
@@ -69,28 +69,7 @@
     <WholeProgramOptimization>true</WholeProgramOptimization>
     <LinkIncremental>false</LinkIncremental>
   </PropertyGroup>
-  <PropertyGroup Label="Configuration" Condition="'$(Configuration)|$(Platform)'=='Debug|ARM'">
-    <DesktopCompatible>true</DesktopCompatible>
-  </PropertyGroup>
-  <PropertyGroup Label="Configuration" Condition="'$(Configuration)|$(Platform)'=='Release|ARM'">
```
rohanp-msft, line 75: Appreciate the simplification here and elsewhere!

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/src/Microsoft.ApplicationModel.Resources.vcxproj
jefgen: Hmm, I'm not sure I understand... maybe I'm missing something.
The vcxproj file by default sets `/profile`, so someone would need to edit/change to remove it.
By settings these as well that way VS Community builds would be closer to the same as VS Enterprise builds, which I think would be a good thing... (?).

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/src/Microsoft.ApplicationModel.Resources.vcxproj
jefgen: Thanks! :)

dev/MRTCore/mrt/Core/src/MRM.vcxproj
```diff
@@ -96,9 +96,20 @@
       <AdditionalDependencies>$(OutDir)..\mrmmin\mrmmin.lib;rpcrt4.lib;onecoreuap.lib;%(AdditionalDependencies)</AdditionalDependencies>
       <ModuleDefinitionFile>MRM.def</ModuleDefinitionFile>
       <SubSystem>Windows</SubSystem>
+      <!-- This sets /DEBUG -->
```
DrusTheAxe, line 99: Remove

No need documenting MSBuild/VS' properties

dev/MRTCore/mrt/Core/src/MRM.vcxproj
```diff
@@ -96,9 +96,20 @@
...
       <GenerateDebugInformation>true</GenerateDebugInformation>
-      <EnableCOMDATFolding>true</EnableCOMDATFolding>
+      <!-- Note: /PROFILE implies the following 4 settings.
+           However, since /PROFILE is only supported on VS Enterprise we will also set them explicitly below as well. -->
```
DrusTheAxe, line 102: Would it be more helpful to define something in a root props and use it where needed, e.g. create `/Profile.props` that does lines 101-112 and then in this file just `<Import...Profile.props/>` ?

dev/MRTCore/mrt/Core/src/MRM.vcxproj
```diff
@@ -121,10 +132,12 @@
...
-      <!-- Configure the release build binary to be as required by internal API scanning tools. -->
-      <LinkIncremental>false</LinkIncremental>
-      <GenerateDebugInformation>true</GenerateDebugInformation>
-      <!-- EnableCOMDATFolding and OptimizeReferences are set to true below for all build configurations. -->
+      <!-- Note: /PROFILE is required for API scanning, and is set above for all build configurations.
+           However, this also sets /OPT:NOICF which disables EnableCOMDATFolding. For Release builds
+           it is suggested to re-enable this, but the only way to do so is by setting additional command
+           line options to the linker, after /profile is passed.
+       -->
+      <AdditionalOptions>/opt:icf %(AdditionalOptions)</AdditionalOptions>
```
DrusTheAxe, line 140: Isn't /OPT:ICF the same as `<EnableCOMDATFolding>true</EnableCOMDATFolding>` ?

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/src/Microsoft.ApplicationModel.Resources.vcxproj
huichen123: with the other flags set explicitly, there is not need to set <profile>true</profile>

jefgen: I thought about doing that, but `/profile` also ends up setting `/debugtype:cv,fixup` and there is no explicit option to set that.
I think we could try adding additional command line options to get the same effect.
(Though the APIScan documentation seems to imply that `/profile` is the preferred way to do it.)

dev/MRTCore/mrt/Core/src/MRM.vcxproj
jefgen: Sure, I'll remove these comments.

dev/MRTCore/mrt/Core/src/MRM.vcxproj
jefgen: Right, `/OPT:ICT` is the same as `EnableCOMDATFolding`, but the issue is that setting `Profile=true` actually **disables** `/OPT:ICF`, as it implies `/OPT:NOICF`.
The only way to get it back is by adding `AdditionalOptions` which get passed after `/profile`, and turn it back on.
(It kind of feels like a kludge, but this is actually what the APIScan docs suggest doing...)

dev/MRTCore/mrt/Core/src/MRM.vcxproj
jefgen: Good idea!
What about something like `WindowsAppSDK.Vulcanize.Cpp.props` in the top-level dev folder (since this is specific to C++) ?

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/src/Microsoft.ApplicationModel.Resources.vcxproj
jefgen: I pushed changes which now set explicit flags, so we aren't using `/profile` anymore -- so there shouldn't be anymore issues with different versions of VS.

dev/MRTCore/mrt/Core/src/MRM.vcxproj
jefgen: I pushed changes which no longer use `Profile`, so we don't need to turn things back on again after they have been turned off.

dev/MRTCore/mrt/Core/src/MRM.vcxproj
```diff
@@ -107,25 +103,15 @@
       <Optimization>Disabled</Optimization>
       <WarningVersion>18</WarningVersion>
     </ClCompile>
+    <Link>
+      <EnableCOMDATFolding>false</EnableCOMDATFolding>
```
jefgen, line 107: I turned off COMDAT folding for **Debug** builds per https://docs.microsoft.com/en-us/cpp/build/reference/opt-optimizations?view=msvc-160 as it can make debugging harder:
> ... it can change the function names that appear in stack traces. It can also make it impossible to set breakpoints in certain functions or to examine some data in the debugger, and can take you into unexpected functions when you single-step through your code.

axelandrejs (approved):

dev/MRTCore/mrt/Core/src/MRM.vcxproj
jefgen: Per offline conversation with Rohan, going to use `WindowsAppSDK.ApiScan.Cpp.props` instead.

DrusTheAxe: If Vulcanize is C/C++ specific then yes on the .Cpp part

I'd omit the WindowsAppSDK prefix. We have other *.props in the  root and non have the project name as a prefix. So first pass I suggest `\Vulcanize.Cpp.props`, and think about <del>if</del> when we refactor what should the structure look like? Should the be a subdir? Called what? `/build` is for the ADO build pipeline. Should there be a subdir under and existing directory or a new root? `/build/msbuild` or...?

So go with Vulcanize.Cpp.props for now to unblock this work, and then cleaning up file structure is a simple (later) PR.

dev/WindowsAppSDK.ApiScan.Cpp.props
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
```
huichen123, line 1: could be out of the scope of this PR. we may hook this up in Directory.build.props so that it applies to all binaries. Then of course it cannot be cpp only.

dev/MRTCore/mrt/Core/src/MRM.vcxproj
jefgen: Thanks!
Yes, the Vulcanization settings are specific to C/C++. There's different ones for C#/Managed code, so we'd likely want to have that be a separate file. Or we could probably share the same file, but since this only handles C/C++ for now, it feels like we'd want to denote that with adding `.Cpp` to the name.

I'll move the file to the root level folder and use the name `ApiScan.Cpp.props` (so no prefix of WindowsAppSDK). Also using ApiScan instead of Vulcanize, in order to avoid using internal-only names.

dev/WindowsAppSDK.ApiScan.Cpp.props
jefgen: Hm, that's a good idea. I'm not as familiar with the C#/Managed settings though, so I'd prefer to leave that to a separate PR later on.

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/src/Microsoft.ApplicationModel.Resources.vcxproj
```diff
@@ -5,6 +5,7 @@
   <Import Project="..\..\packages\Microsoft.SourceLink.Common.1.1.0-beta-21055-01\build\Microsoft.SourceLink.Common.props" Condition="Exists('..\..\packages\Microsoft.SourceLink.Common.1.1.0-beta-21055-01\build\Microsoft.SourceLink.Common.props')" />
   <Import Project="..\..\packages\Microsoft.Windows.CppWinRT.2.0.210806.1\build\native\Microsoft.Windows.CppWinRT.props" Condition="Exists('..\..\packages\Microsoft.Windows.CppWinRT.2.0.210806.1\build\native\Microsoft.Windows.CppWinRT.props')" />
   <Import Project="..\..\..\..\WindowsAppSDK.Build.Cpp.props" />
```
huichen123, line 7: >WindowsAppSDK.Build.Cpp.props [](http://example.com/codeflow?start=31&length=29)

wondering why this is under \dev.

huichen123 (approved):
:shipit:

dev/MRTCore/mrt/Microsoft.ApplicationModel.Resources/src/Microsoft.ApplicationModel.Resources.vcxproj
jefgen: It's no longer under `\dev` anymore. 🙂

huichen123: i am talking about the one above yours "WindowsAppSDK.Build.Cpp.props"" :)

jefgen: Ah, yeah good question.. I'm not sure why that one is under \dev either...

PR #1282, opened 2021-08-20
Moved Bootstrap dll to fix VS/AnyCPU projects

DrusTheAxe:
Moved the dll from runtimes\lib\native\PLATFORM to runtimes\win-PLATFORM\native

Fixes https://task.ms/33477857, #1217 and maybe #893

PR #1333, opened 2021-08-28
User/drustheaxe/undo bootstrap move

DrusTheAxe:
Something went very wrong in the Aggregator's NuGet creation. This moved the bootstrap DLL to the right location but caused other files to get dropped out of the NuGet. Something's off in the aggregator but damned if I can figure it out right now, and resident experts are OOF. Reverting this change for now. Will take another run at it at a later date, when pipeline experts are around to advise


rohanp-msft:
Does https://github.com/microsoft/WindowsAppSDK/pull/1282 need to be undone too?

rohanp-msft (approved):

PR #1336, opened 2021-08-28
Introducing unpackaged foreground activation for PushNotifications

pmpurifoy:
This PR allows unpackaged applications to declare foreground handlers for their requested PushNotificationChannel.


danielayala94 (changes requested):

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -48,18 +51,75 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
     }
 
     void PushNotificationChannel::PushReceived(winrt::event_token const& token) noexcept
     {
-        m_channel.PushNotificationReceived(token);
+        if (IsPackagedApp())
+        {
+            m_channel.PushNotificationReceived(token);
+        }
+        else
```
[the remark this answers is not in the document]
pmpurifoy: This bool is guard for registering with the LRP only once.

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -48,18 +51,75 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+    }
+
+    HRESULT __stdcall PushNotificationChannel::InvokeAll(ULONG length, byte* start) noexcept try
+    {
+        std::vector<uint8_t> vec;
+
+        for (int i = 0; (ULONG)i < length; i++)
+        {
+            vec.push_back(start[i]);
+        }
```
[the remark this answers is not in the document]
pmpurifoy: Yup. Need to remove.

dev/PushNotifications/PushNotificationsLongRunningTask/ForegroundSinkManager.h
```diff
@@ -0,0 +1,23 @@
...
+#include <vector>
+
+class ForegroundSinkManager
+{
+public:
+    ForegroundSinkManager() = default;
+
+    void AddSink(std::wstring processName, IWpnForegroundSink* sink);
+
+    void Remove(std::wstring processName);
```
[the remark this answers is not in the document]
pmpurifoy: This is the remove one sink function. One sink holds all the handlers which we can remove with just the processName.

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -55,3 +55,31 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::RegisterF
...
+    m_foregroundSinkManager.AddSink(processName, sink);
+    return S_OK;
+}
+
+STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::SendBackgroundNotification(_In_ PCWSTR processName, _In_ byte* payload, _In_ ULONG payloadSize)
+{
+    return S_OK;
+}
+
+void NotificationsLongRunningPlatformImpl::UnregisterForegroundActivator(_In_ PCWSTR processName)
```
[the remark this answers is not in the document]
pmpurifoy: This will be used by the LRP in the future to remove the foreground sink when the RefCount == 1.

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -48,18 +51,75 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
 
     winrt::event_token PushNotificationChannel::PushReceived(winrt::Windows::TypedEventHandler<winrt::Microsoft::Windows::PushNotifications::PushNotificationChannel, winrt::Microsoft::Windows::PushNotifications::PushNotificationReceivedEventArgs> handler)
     {
-        return m_channel.PushNotificationReceived([weak_self = get_weak(), handler](auto&&, auto&& args)
+        if (IsPackagedApp())
```
hulumane, line 54: This check is incorrect. Packaged apps that don't use the BI APIs (unsupported) will need to subscribe their FG sink with the LRP. This is because, we will treat these apps as unpackaged essentially. I believe you need to use IsActivatorSupported here.

dev/PushNotifications/PushNotificationsLongRunningTask/ForegroundSinkManager.cpp
```diff
@@ -0,0 +1,26 @@
...
+{
+    m_foregroundMap.erase(processName);
+}
+
+bool ForegroundSinkManager::InvokeForegroundHandlers(std::wstring processName, byte* payload, ULONG payloadSize)
+{
+    if (m_foregroundMap.find(processName) != m_foregroundMap.end())
+    {
+        m_foregroundMap[processName]->InvokeAll(payloadSize, payload);
+        return true;
```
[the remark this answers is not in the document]
pmpurifoy: I think this should be wrapped in LOG_IF_FAILED because we found the registered sink but something happened on calling InvokeAll.

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -48,18 +51,75 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-            if (strong)
+            if (!m_isRegisteredWithLRP)
             {
-                handler(*strong, winrt::make<winrt::Microsoft::Windows::PushNotifications::implementation::PushNotificationReceivedEventArgs>(args));
-            };
-        });
+                wil::com_ptr<INotificationsLongRunningPlatform> notificationsLongRunningPlatform{
+                    wil::CoCreateInstance<NotificationsLongRunningPlatform, INotificationsLongRunningPlatform>(CLSCTX_LOCAL_SERVER) };
+
+                wchar_t processName[1024];
```
hulumane, line 72: Dangerous to assume that processName which includes the file path is max of 1024 characters.

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -48,18 +51,75 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                m_isRegisteredWithLRP = true;
+            }
+            return m_foregroundHandlers.add(handler);
+        }
     }
 
     void PushNotificationChannel::PushReceived(winrt::event_token const& token) noexcept
     {
-        m_channel.PushNotificationReceived(token);
+        if (IsPackagedApp())
```
hulumane, line 85: Same issue as above here. This check is incorrect.

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -48,18 +51,75 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        {
+            m_channel.PushNotificationReceived(token);
+        }
+        else
+        {
+            m_foregroundHandlers.remove(token);
+        }
```
hulumane, line 92: This implementation of unregistering the eventhandler has just leaked the long running foreground sink in the LRP. What you essentially need to do is to remove the foreground sink in LRP if the total number of event handlers registered is 0.

dev/PushNotifications/PushNotificationChannel.cpp
pmpurifoy: This function checks if the app is Packaged/BiAvailable but the function name is not accurate. Will rename.

dev/PushNotifications/PushNotificationChannel.cpp
pmpurifoy: Function needs to be renamed.

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -48,18 +51,75 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+    }
+    CATCH_RETURN()
+
+    bool PushNotificationChannel::IsBackgroundTaskBuilderAvailable()
+    {
+        return winrt::Windows::Metadata::ApiInformation::IsMethodPresent(L"Windows.ApplicationModel.Background.BackgroundTaskBuilder", L"SetTaskEntryPointClsid");
+    }
+
+    // Determines if the caller should be treated as packaged app or not.
+    bool PushNotificationChannel::IsPackagedApp()
```
hulumane, line 116: Would rename the function to something like: bool IsPackagedAppScenario().

IsPackagedApp gives the wrong impression if the app is packaged but BI isn't supported.

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -48,18 +51,75 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+    {
+        if (AppModel::Identity::IsPackagedProcess() && IsBackgroundTaskBuilderAvailable())
```
hulumane, line 118: just return AppModel::Identity::IsPackagedProcess() && IsBackgroundTaskBuilderAvailable();

if else block is not needed.

dev/PushNotifications/PushNotificationChannel.cpp
pmpurifoy: Do we want the behavior to be unregister the LRP on 0 handlers then reregister on the next handler? 

Daniel and I were talking about the lifetime of the sink in the LRP and it is difficult to define when to remove it. One idea is to call UnregisterForegroundActivator when the PushNotificationChannel dtor is called, but the dtor will never be called because it still has a reference in the LRP. We could override AddRef/Release but I am not sure this is the way to go. 

Another idea is to have a cleanup function in the LRP that loops through its currently registered apps and checks their ref counts and removes them if their refcount == 1.

Any suggestions here?

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -48,18 +51,75 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        {
+            m_channel.PushNotificationReceived(token);
+        }
+        else
+        {
+            m_foregroundHandlers.remove(token);
+        }
+    }
+
+    HRESULT __stdcall PushNotificationChannel::InvokeAll(ULONG length, byte* start) noexcept try
```
hulumane: this is byte* payload.  Please lets not call it start.

dev/PushNotifications/PushNotificationChannel.h
```diff
@@ -3,24 +3,39 @@
...
+
         winrt::Windows::Foundation::Uri Uri();
         winrt::Windows::Foundation::DateTime ExpirationTime();
         void Close();
 
         winrt::event_token PushReceived(winrt::Windows::Foundation::TypedEventHandler<Microsoft::Windows::PushNotifications::PushNotificationChannel, Microsoft::Windows::PushNotifications::PushNotificationReceivedEventArgs> handler);
         void PushReceived(winrt::event_token const& token) noexcept;
 
+        // IWpnForegroundSink
+        HRESULT __stdcall InvokeAll(ULONG length, byte* start) noexcept;
```
hulumane, line 26: This is byte* payload. Please lets not call it start. Also please reverse the order of the arguments. payload first and then length. Also missing SAL notation.

dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
```diff
@@ -23,49 +23,108 @@ namespace winrt
...
-        return m_backgroundTaskInstance.GetDeferral();
+        if (m_isBIAvailable)
+        {
+            THROW_HR_IF_NULL_MSG(E_ILLEGAL_METHOD_CALL, m_backgroundTaskInstance, "Foreground activation cannot call this.");
+
+            return m_backgroundTaskInstance.GetDeferral();
+        }
+        else
+        {
+            return nullptr;
```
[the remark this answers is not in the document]
hulumane: Yes. I would implement a dummy Deferral here an just return that. Otherwise we crash.

dev/PushNotifications/PushNotificationChannel.cpp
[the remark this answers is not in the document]
pmpurifoy: Looks like there is MAX_PATH constant which equals 260.

dev/PushNotifications/PushNotificationReceivedEventArgs.h
```diff
@@ -17,9 +19,17 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
         bool Handled();
         void Handled(bool value);
 
+        byte* GetByteArrayFromBuffer(winrt::Windows::Storage::Streams::IBuffer buffer);
+
     private:
-        const winrt::Windows::Storage::Streams::IBuffer m_rawNotification{};
+
+        byte* m_rawNotification = nullptr;
```
hulumane: Bad idea to have raw pointers. This should be std::vector<byte> instead.

dev/PushNotifications/PushNotificationChannel.cpp
pmpurifoy: Looks like there is a MAX_PATH constant which equals 260. Will switch to this constant.

dev/PushNotifications/PushNotificationReceivedEventArgs.h
```diff
@@ -17,9 +19,17 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+
         const winrt::Windows::ApplicationModel::Background::IBackgroundTaskInstance m_backgroundTaskInstance{};
         const winrt::Windows::Networking::PushNotifications::PushNotificationReceivedEventArgs m_args = nullptr;
+
+        ULONG m_length;
```
hulumane: This should be deleted. Store the bytes in a vector please and infer length from the vector instead.

dev/PushNotifications/PushNotificationChannel.cpp
hulumane: Thats even worse. We can have file paths greater than 260 characters right? Then what?

dev/PushNotifications/PushNotificationChannel.cpp
hulumane: You remove it when the total number of foreground event handlers goes to 0. Otherwise it is ok to leak.

On the LRP side, when you try to invoke the FG Sink and you receive an RPC error, it is a good indication that the app is no longer in the FG. In that case, you will need to delete the foreground sink in the LRP and call into the background (protocol) activator. 

Makes sense? : )

hulumane: I would also remove it when the dtor of channel is called to be super safe.

dev/PushNotifications/PushNotificationChannel.h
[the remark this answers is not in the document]
jonwis: Well, `_In_reads_(length)`

dev/PushNotifications/PushNotificationChannel.cpp
pmpurifoy: Yup sounds good!

dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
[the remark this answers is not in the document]
jonwis: Yes, I'd return a dummy Deferral object.

dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
```diff
@@ -23,49 +23,108 @@ namespace winrt
...
+        {
+            THROW_HR_IF_NULL_MSG(E_ILLEGAL_METHOD_CALL, m_args, "Background activation cannot call this.");
+
+            m_args.Cancel(value);
+        }
+        else
+        {
+            if (value)
+            {
+                wchar_t processName[1024];
```
jonwis, line 117: Can this code (the get module file name) be deduplicated with the above as well?

dev/PushNotifications/PushNotificationsLongRunningTask/ForegroundSinkManager.h
```diff
@@ -0,0 +1,23 @@
...
+#include <unordered_set>
+#include <algorithm>
+#include <vector>
+
+class ForegroundSinkManager
+{
+public:
+    ForegroundSinkManager() = default;
+
+    void AddSink(std::wstring processName, IWpnForegroundSink* sink);
```
hulumane, line 14: Just Add. Not AddSink since Sink is implied in the class name

dev/PushNotifications/PushNotificationsLongRunningTask.ProxyStub/NotificationsLongRunningProcess.idl
```diff
@@ -6,12 +6,25 @@ import "ocidl.idl";
...
 [object]
 [uuid(60FC21B2-B396-4D49-94F0-7555869FB93C)]
 [pointer_default(unique)]
 interface INotificationsLongRunningPlatform : IUnknown
 {
     HRESULT RegisterFullTrustApplication([in] LPCWSTR processName, [in] GUID remoteId, [out] GUID* appId);
+
+    HRESULT RegisterForegroundActivator([in] IWpnForegroundSink* sink, [in] LPCWSTR processName);
+
+    HRESULT SendBackgroundNotification([in] LPCWSTR processName, [in, string] byte data[], [in] ULONG length);
```
hulumane, line 26: What is this for? I don't get it.

hulumane: Ah just read through your code. This is if the developer intercepts the payload in the FG but for some strange reason also chooses to trigger the background activator?

dev/PushNotifications/PushNotificationReceivedEventArgs.h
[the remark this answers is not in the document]
jonwis: Why is this a raw pointer vs. a `vector<uint8_t>` or a `std::unique_ptr<uint8_t[]>` ?

dev/PushNotifications/PushNotificationsLongRunningTask.ProxyStub/NotificationsLongRunningProcess.idl
pmpurifoy: Yes. This is to simulate .Handled().

hulumane: Hmm why don't you have IWpnSink::InvokeAll return a specific error code if Handled == false. Then you could eliminate this API and handle everything from LRP side.

hulumane: Or InvokeAll could have an additional parameter out bool handled?

dev/PushNotifications/PushNotificationReceivedEventArgs.h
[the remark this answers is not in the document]
jonwis: (Applies everywhere - please no raw pointers except at ABI boundaries.)

dev/PushNotifications/PushNotificationsLongRunningTask.ProxyStub/NotificationsLongRunningProcess.idl
```diff
@@ -6,12 +6,25 @@ import "ocidl.idl";
 
 #include "../PushNotifications-Constants.h"
 
+[object]
+[uuid(25604D55-9B17-426F-9D67-2B11B3A65598)]
+[pointer_default(unique)]
+interface IWpnForegroundSink : IUnknown
+{
+    HRESULT InvokeAll([in] ULONG length, [in, string] byte data[]);
```
jonwis, line 14: 'string' means zero-terminated for length. I think you meant `[in, size_is(length)] byte*` ?

dev/PushNotifications/PushNotificationsLongRunningTask/ForegroundSinkManager.cpp
```diff
@@ -0,0 +1,26 @@
...
+void ForegroundSinkManager::Remove(std::wstring processName)
+{
+    m_foregroundMap.erase(processName);
+}
+
+bool ForegroundSinkManager::InvokeForegroundHandlers(std::wstring processName, byte* payload, ULONG payloadSize)
+{
+    if (m_foregroundMap.find(processName) != m_foregroundMap.end())
+    {
+        m_foregroundMap[processName]->InvokeAll(payloadSize, payload);
```
jonwis: This forces a dual-lookup, once to find another to index in.  The usual pattern is:

```c++
auto it = map.find(p);
if (it != map.end())
{
    it->second->InvokeAll(...);
}
```

dev/PushNotifications/PushNotificationsLongRunningTask/ForegroundSinkManager.cpp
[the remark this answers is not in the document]
jonwis: See `SUCCEEDED_LOG(exp)`

dev/PushNotifications/PushNotificationsLongRunningTask/ForegroundSinkManager.h
```diff
@@ -0,0 +1,23 @@
...
+
+    void Remove(std::wstring processName);
+
+    bool InvokeForegroundHandlers(std::wstring processName, byte* payload, ULONG payloadSize);
```
jonwis, line 18: Should these be `const&` instead of value type?

jonwis, line 18: Should this be a `winrt::com_array<>` or a `gsl::span` ? Don't use raw pointers unless you have to.

jonwis (approved):
Few changes, nothing major.

dev/PushNotifications/PushNotificationsLongRunningTask.ProxyStub/NotificationsLongRunningProcess.idl
pmpurifoy: Does the background handler get triggered once the .Handled(bool) is called?

dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
```diff
@@ -90,7 +87,7 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
 
     bool PushNotificationReceivedEventArgs::Handled()
     {
-        if (m_isBIAvailable)
+        if (m_isUnpackagedApp)
```
danielayala94, line 90: Wait, are the code blocks in their opposite side? I would expect to return `m_handledUnpackaged` if this is an unpkg app.

dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
```diff
@@ -66,7 +63,7 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
 
     winrt::event_token PushNotificationReceivedEventArgs::Canceled(winrt::BackgroundTaskCanceledEventHandler const& handler)
     {
-        if (m_isBIAvailable)
+        if (m_isUnpackagedApp)
```
danielayala94, line 66: Code blocks are in the wrong side.

dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
```diff
@@ -25,34 +25,31 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
     winrt::com_array<uint8_t> PushNotificationReceivedEventArgs::Payload()
     {
-        return { m_rawNotification, m_rawNotification + (m_length * sizeof(uint8_t)) };
+        return { m_rawNotificationPayload.data(), m_rawNotificationPayload.data() + (m_rawNotificationPayload.size() * sizeof(uint8_t)) };
     }
 
     winrt::BackgroundTaskDeferral PushNotificationReceivedEventArgs::GetDeferral()
     {
-        if (m_isBIAvailable)
+        if (m_isUnpackagedApp)
```
danielayala94, line 52: Code blocks are in the wrong side.

dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
```diff
@@ -80,7 +77,7 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
 
     void PushNotificationReceivedEventArgs::Canceled(winrt::event_token const& token) noexcept
     {
-        if (m_isBIAvailable)
+        if (m_isUnpackagedApp)
```
danielayala94, line 80: Wrong condition. You want:
```suggestion
        if (!m_isUnpackagedApp)
```

dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
```diff
@@ -104,23 +101,23 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
 
     void PushNotificationReceivedEventArgs::Handled(bool value)
     {
-        if (m_isBIAvailable)
+        if (m_isUnpackagedApp)
```
danielayala94, line 104: It seems like we changed the var name but now `true` has a different meaning.

dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
```diff
@@ -25,34 +25,31 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
 {
     PushNotificationReceivedEventArgs::PushNotificationReceivedEventArgs(winrt::IBackgroundTaskInstance const& backgroundTask):
         m_backgroundTaskInstance(backgroundTask),
-        m_rawNotification(GetByteArrayFromBuffer(backgroundTask.TriggerDetails().as<RawNotification>().ContentBytes())),
-        m_isBIAvailable(true) {}
+        m_rawNotificationPayload(BuildPayload(backgroundTask.TriggerDetails().as<RawNotification>().ContentBytes().data(), backgroundTask.TriggerDetails().as<RawNotification>().ContentBytes().Length())),
+        m_isUnpackagedApp(true) {}
```
danielayala94, line 29: Probably at this point, the best to do is to rename this var. All checks throughout the code makes me think if this is in fact `m_isPackagedApp`.

danielayala94 (changes requested):

dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
danielayala94: I would just rename this variable to `m_isPackagedApp`. The code works as if that were it's real name.

dev/PushNotifications/PushNotificationsLongRunningTask.ProxyStub/NotificationsLongRunningProcess.idl
hulumane: If the developer intercepts the payload in the foreground and sets args.Handled = true, then the background handler will not get invoked. Otherwise the background handler would also get invoked with the same payload.

dev/PushNotifications/PushNotificationsLongRunningTask/ForegroundSinkManager.cpp
```diff
@@ -0,0 +1,26 @@
+﻿#pragma once
+
+#include "pch.h"
+
+#include <NotificationsLongRunningProcess_h.h>
+#include <ForegroundSinkManager.h>
+
+void ForegroundSinkManager::Add(std::wstring processName, IWpnForegroundSink* sink)
+{
+    m_foregroundMap[processName] = sink;
```
hulumane, line 10: None of these operations have locks or are threadsafe.

dev/PushNotifications/PushNotificationsLongRunningTask/ForegroundSinkManager.cpp
```diff
@@ -0,0 +1,26 @@
...
+void ForegroundSinkManager::Remove(std::wstring processName)
+{
+    m_foregroundMap.erase(processName);
+}
+
+bool ForegroundSinkManager::InvokeForegroundHandlers(std::wstring processName, byte* payload, ULONG payloadSize)
+{
+    if (m_foregroundMap.find(processName) != m_foregroundMap.end())
+    {
+        LOG_IF_FAILED(m_foregroundMap[processName]->InvokeAll(payloadSize, payload));
```
hulumane, line 22: You should first check if you get an RPC error. If you do, delete the foreground sink and invoke the background handler.

hulumane: I guess you would be returning false for the RPC error case.

dev/PushNotifications/PushNotificationsLongRunningTask/ForegroundSinkManager.h
```diff
@@ -0,0 +1,23 @@
...
+    ForegroundSinkManager() = default;
+
+    void Add(std::wstring processName, IWpnForegroundSink* sink);
+
+    void Remove(std::wstring processName);
+
+    bool InvokeForegroundHandlers(std::wstring processName, byte* payload, ULONG payloadSize);
+
+private:
+    std::unordered_map<std::wstring, Microsoft::WRL::ComPtr<IWpnForegroundSink>> m_foregroundMap = {};
```
hulumane, line 21: Please add a comment somewhere to say that "An app can only have one active foreground sink with Long Running Process" and that event handler accounting is done in WinSDK."

dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
```diff
@@ -23,49 +23,105 @@ namespace winrt
...
+        {
+            if (!value)
+            {
+                wchar_t processName[MAX_PATH];
+                THROW_HR_IF(ERROR_FILE_NOT_FOUND, GetModuleFileNameExW(GetCurrentProcess(), NULL, processName, sizeof(processName) / sizeof(processName[0])) == 0);
+
+                wil::com_ptr<INotificationsLongRunningPlatform> notificationsLongRunningPlatform{
+                    wil::CoCreateInstance<NotificationsLongRunningPlatform, INotificationsLongRunningPlatform>(CLSCTX_LOCAL_SERVER) };
+
+                THROW_IF_FAILED(notificationsLongRunningPlatform->SendBackgroundNotification(processName, m_rawNotificationPayload.data(), m_rawNotificationPayload.size()));
```
hulumane, line 120: This pattern is incorrect. All this method should be doing is to set m_handledUnpackaged. 

The actual Background operation is done in the foreground activation thread.

dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
```diff
@@ -23,49 +23,105 @@ namespace winrt
...
+            return m_handledUnpackaged;
+        }
     }
 
     void PushNotificationReceivedEventArgs::Handled(bool value)
     {
-        THROW_HR_IF_NULL_MSG(E_ILLEGAL_METHOD_CALL, m_args, "Background activation cannot call this.");
-
-        m_args.Cancel(value);
+        if (m_isUnpackagedApp)
```
hulumane, line 104: I'm really confused with m_unpackagedApp Vs m_unpackagedAppScenario. I think you mean the latter since packaged apps can also go through the unpackaged scenario. I would reword all the variables to clearly indicate the context.

dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
pmpurifoy: Added GetCurrentProcessPath in externs.h

dev/PushNotifications/PushNotificationsLongRunningTask/ForegroundSinkManager.cpp
pmpurifoy: Added lock to ForegroundSinkManager.

dev/PushNotifications/PushNotificationDummyDeferral.h
```diff
@@ -0,0 +1,31 @@
...
+#include <winrt/Windows.ApplicationModel.background.h>
+
+struct PushNotificationDummyDeferral : winrt::implements <PushNotificationDummyDeferral, winrt::Windows::ApplicationModel::Background::IBackgroundTaskDeferral>
+{
+    PushNotificationDummyDeferral() {}
+
+    void Complete()
+    {
+        return;
+    }
```
danielayala94, line 16: No need of fix. It can just be:

```suggestion
    void Complete() { };
```

danielayala94 (approved):

dev/PushNotifications/PushNotificationsLongRunningTask/ForegroundSinkManager.cpp
```diff
@@ -0,0 +1,36 @@
...
+{
+    auto lock = m_lock.lock_exclusive();
+
+    auto it = m_foregroundMap.find(processName);
+    if (it != m_foregroundMap.end())
+    {
+        BOOL foregroundHandled = true;
+        if (FAILED(it->second->InvokeAll(payloadSize, payload.data(), &foregroundHandled)))
+        {
+            Remove(processName);
```
hulumane, line 30: Deadlock here. You are already in an exclusivelock and the same thread tries to acquire another exclusive lock

dev/PushNotifications/PushNotificationsLongRunningTask.ProxyStub/NotificationsLongRunningProcess.idl
```diff
@@ -6,12 +6,27 @@ import "ocidl.idl";
...
 [pointer_default(unique)]
 interface INotificationsLongRunningPlatform : IUnknown
 {
     HRESULT RegisterFullTrustApplication([in] LPCWSTR processName, [in] GUID remoteId, [out] GUID* appId);
+
+    HRESULT RegisterForegroundActivator([in] IWpnForegroundSink* sink, [in] LPCWSTR processName);
+
+    HRESULT UnregisterForegroundActivator([in] LPCWSTR processName);
+
+    HRESULT SendBackgroundNotification([in] LPCWSTR processName, [in] ULONG length, [in, size_is(length)] byte* data);
```
hulumane, line 28: Please remove from idl. No longer being used anywhere.

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -48,18 +53,73 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-        m_channel.PushNotificationReceived(token);
+        if (IsPackagedAppScenario())
+        {
+            m_channel.PushNotificationReceived(token);
+        }
+        else
+        {
+            auto lock = m_lock.lock_exclusive();
+            m_foregroundHandlers.remove(token);
+            if (!--m_foregroundHandlerCount)
```
hulumane, line 93: Nit: Don't rely on int to bool conversion as a practise. Always better to say != 0 to be precise.

dev/PushNotifications/PushNotificationsLongRunningTask/ForegroundSinkManager.cpp
```diff
@@ -0,0 +1,36 @@
...
+    m_foregroundMap[processName] = sink;
+}
+
+void ForegroundSinkManager::Remove(std::wstring const& processName)
+{
+    auto lock = m_lock.lock_exclusive();
+    m_foregroundMap.erase(processName);
+}
+
+bool ForegroundSinkManager::InvokeForegroundHandlers(std::wstring const& processName, winrt::com_array<uint8_t> const& payload, ULONG const& payloadSize)
```
hulumane, line 20: Who calls this method? Is there a pipe from the LongRunningSink that will be calling into this as a seperate PR?

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -55,3 +55,25 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::RegisterF
...
+STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::UnregisterForegroundActivator(_In_ PCWSTR processName)
+{
+    RETURN_HR_IF(WPN_E_PLATFORM_UNAVAILABLE, m_shutdown);
+    auto lock = m_lock.lock_exclusive();
+
+    m_foregroundSinkManager.Remove(processName);
+    return S_OK;
+}
+
+STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::SendBackgroundNotification(_In_ PCWSTR processName, _In_ ULONG payloadSize, _In_ byte* payload)
```
hulumane, line 76: Please Remove unused APIs

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -48,18 +53,73 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                if (auto strong = weak_self.get())
+                {
+                    handler(*strong, winrt::make<winrt::Microsoft::Windows::PushNotifications::implementation::PushNotificationReceivedEventArgs>(args));
+                };
+            });
+        }
+        else
+        {
+            auto lock = m_lock.lock_exclusive();
+            if (!m_foregroundHandlerCount++)
```
hulumane, line 69: Nit: From a clarity perspective, prefer that we don't rely on int to bool conversion. Better to say != 0 or > 0 and convert it into the bool equivalent.

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -55,3 +55,25 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::RegisterF
     return E_NOTIMPL;
 }
 
+STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::RegisterForegroundActivator(_In_ IWpnForegroundSink* sink, _In_ PCWSTR processName)
+{
+    RETURN_HR_IF(WPN_E_PLATFORM_UNAVAILABLE, m_shutdown);
```
hulumane, line 60: Unfortunately m_shutdown is not being protected by the lock. So you can run into race conditions. @danielayala94 What is the recommended pattern here?

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -55,3 +55,25 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::RegisterF
...
+    auto lock = m_lock.lock_exclusive();
+
+    m_foregroundSinkManager.Add(processName, sink);
+    return S_OK;
+}
+
+STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::UnregisterForegroundActivator(_In_ PCWSTR processName)
+{
+    RETURN_HR_IF(WPN_E_PLATFORM_UNAVAILABLE, m_shutdown);
```
hulumane, line 69: Same race issue with shutdown. Not being protected by the lock. @danielayala94

PR #1337, opened 2021-08-28
Add basic telemetry to Push Notifications

loneursid:
Add basic telemetry support for the WindowsAppSDK (WAS) Push Notification feature so we can:
1. measure adoption
2. identify and investigate issues in the WAS layer

New telemetry events:
* PushNotificationTelemetry::ChannelRequestedByApi
* PushNotificationTelemetry::ChannelClosedByApi
* PushNotificationTelemetry::ActivatorRegisteredByApi
* PushNotificationTelemetry::ActivatorUnegisteredByApi

Open Questions
* Is there value in logging registration / unregistration of activators?
* Is there calue in logging registration to the PushReceived event?

How tested?
Ran all our unit tests and verified the events were correctly logged (including PartB data) using WPA


dev/PushNotifications/PushNotificationTelemetry.h
```diff
@@ -0,0 +1,74 @@
...
+        {
+            if (c_maxEventLimit >= UpdateLogEventCount())
+            {
+                TraceLoggingClassWriteMeasure(
+                    "ChannelRequestedByApi",
+                    TelemetryPrivacyDataTag(PDT_ProductAndServicePerformance),
+                    _GENERIC_PARTB_FIELDS_ENABLED,
+                    TraceLoggingHexUInt32(hr, "OperationResult"),
+                    TraceLoggingBool(isAppPackaged, "IsAppPackaged"),
+                    TraceLoggingWideString(to_hstring(remoteId).data(), "RemoteId"));
```
ZenBird, line 30: Can this be replaced by DEFINE_COMPLIANT_TRACELOGGING_EVENT_PARAM3(ChannelRequestedByApi, int, operationResult, bool, isAppPackaged, PCWSTR, remoteId); ?

and 

PushNotificationTelemetry::ChannelRequestedByApi(hr, isAppPackaged, to_hstring(remoteId).data());


This is so that you don't have to reference _GENERIC_PARTB_FIELDS_ENABLED in your code.

loneursid: It almost can but the solution you're proposing will fail to compile because it re-use the same eventId for DEFINE_COMPLIANT_TRACELOGGING_EVENT_PARAM3 as was used for DEFINE_EVENT_METHOD. The function signatures will be too similar and the compiler will get all confused.

We can go around this by introducing a new name for the eventId but this somewath obfuscates the code, i.e.: the name you'll see in the logs, won't be the same that the provider exposes in its API.

loneursid: What is wrong with refering to _GENERIC_PARTB_FIELDS_ENABLED in the code?

ZenBird: _GENERIC_PARTB_FIELDS_ENABLED is an internal implementation detail. It should not bleed into any other components. Ideally, it should be hidden similar to PartA. We are not able to hide it because the telemetry is MACRO based.

loneursid: I don't think it's a reasonable expectation. _GENERIC_PARTB_FIELDS_ENABLED is very public and you won;t be able to police its usage for ever. People will see it and use it when they need it.

Have you thought about introducing a new TraceLoggingClassWriteMeasure or whatever TraceLoggingWrite function it wraps that would have applied _GENERIC_PARTB_FIELDS_ENABLED. If people have access to the correct construct they need in their code, that would certainly help deter people from _GENERIC_PARTB_FIELDS_ENABLED.

loneursid: For example, if there was a version of TraceLoggingClassWriteMeasure that applied _GENERIC_PARTB_FIELDS_ENABLED, I'd be very happy to use it.

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -35,12 +36,16 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
         try
         {
             m_channel.Close();
+
+            PushNotificationTelemetry::ChannelClosedByApi(S_OK);
         }
         catch (...)
         {
             auto channelCloseException = hresult_error(to_hresult());
             if (channelCloseException.code() != HRESULT_FROM_WIN32(ERROR_NOT_FOUND))
```
danielayala94, line 45: What if the error is ERROR_NOT_FOUND? Don we want to log this too?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -97,6 +98,11 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
 
                 pushChannelReceived = co_await channelManager.CreatePushNotificationChannelForApplicationAsync();
 
+                PushNotificationTelemetry::ChannelRequestedByApi(
+                    S_OK,
+                    AppModel::Identity::IsPackagedProcess(),
```
danielayala94, line 103: We can cache this value at the beginning of the function instead of doing two calls to this API.

dev/PushNotifications/PushNotificationTelemetry.h
```diff
@@ -0,0 +1,105 @@
...
+                    "ChannelClosedByApi",
+                    TelemetryPrivacyDataTag(PDT_ProductAndServicePerformance),
+                    _GENERIC_PARTB_FIELDS_ENABLED,
+                    TraceLoggingHexUInt32(hr, "OperationResult"));
+            }
+        }
+        CATCH_LOG()
+
+    DEFINE_EVENT_METHOD(ActivatorRegisteredByApi)(
+        winrt::hresult hr) noexcept try
```
danielayala94, line 53: We are missing information, like the flags and appId (either AppUserModelId or processName, depending on the case).

dev/PushNotifications/PushNotificationTelemetry.h
```diff
@@ -0,0 +1,105 @@
...
+                    TelemetryPrivacyDataTag(PDT_ProductAndServicePerformance),
+                    _GENERIC_PARTB_FIELDS_ENABLED,
+                    TraceLoggingHexUInt32(hr, "OperationResult"));
+            }
+        }
+        CATCH_LOG()
+
+
+    DEFINE_EVENT_METHOD(ActivatorUnregisteredByApi)(
+        winrt::hresult hr) noexcept try
```
danielayala94, line 68: Same as the event above: at least we should know the flags and appId.

dev/PushNotifications/PushNotificationChannel.cpp
loneursid: Yes we do. Thanks for cathing this.

dev/PushNotifications/PushNotificationManager.cpp
loneursid: But won't the first use of the API disapear once unpackaged has been implemented?

dev/PushNotifications/PushNotificationTelemetry.h
loneursid: Done. Options have been added to RegisterActivatorByAPi and UnregisterActivaterByAPi events. AppUserModelId has been added to all reunion PushNotification telemetry events.

dev/PushNotifications/PushNotificationTelemetry.h
loneursid: Done

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -269,7 +269,7 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
             LOG_IF_FAILED(::CoRevokeClassObject(static_cast<DWORD>(token.Cookie())));
         }
 
-        PushNotificationTelemetry::ActivatorUnregisteredByApi(S_OK);
+        PushNotificationTelemetry::ActivatorUnregisteredByApi(S_OK, options);
```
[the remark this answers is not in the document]
danielayala94: In this case, we may want to do:

```
void PushNotificationManager::UnregisterActivator(PushNotificationRegistrationToken const& token, PushNotificationRegistrationOptions const& options)
{
    try
    {
        // ...
    }
    catch(...)
    {
        HRESULT hrError = wil::ResultFromCaughtException();
        PushNotificationTelemetry::ActivatorUnregisteredByApi(hrError, options);
        THROW_HR(hrError)
    }
    PushNotificationTelemetry::ActivatorUnregisteredByApi(S_OK, options);
}
```

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -248,7 +248,7 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
         PushNotificationRegistrationToken token = { cookie, registeredTaskFromBuilder };
         scopeExitToCleanRegistrations.release();
 
-        PushNotificationTelemetry::ActivatorRegisteredByApi(S_OK);
+        PushNotificationTelemetry::ActivatorRegisteredByApi(S_OK, details.Options());
```
danielayala94, line 251: Same issue as in UnregisterActivator: we can throw but we don't log the error code.

danielayala94 (changes requested):

dev/PushNotifications/PushNotificationTelemetry.h
```diff
@@ -16,67 +16,80 @@ class PushNotificationTelemetry : public wil::TraceLoggingProvider
...
-                    TelemetryPrivacyDataTag(PDT_ProductAndServicePerformance),
-                    _GENERIC_PARTB_FIELDS_ENABLED,
-                    TraceLoggingHexUInt32(hr, "OperationResult"));
-            }
+            TraceLoggingClassWriteMeasure(
+                "ActivatorUnregisteredByApi",
+                TelemetryPrivacyDataTag(PDT_ProductAndServicePerformance),
+                _GENERIC_PARTB_FIELDS_ENABLED,
+                TraceLoggingHexUInt32(hr, "OperationResult"),
+                TraceLoggingHexUInt32(static_cast<std::underlying_type_t<RegistrationOptions>>(options),
```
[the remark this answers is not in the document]
loneursid: Options is of type RegistrationOptions, wich is an enum. I'm just casting it to its underlying type, which is uint32. I could cast directly to uint32 (static_cast<uint32> but then it's not as foul proof. It's unlikely, but someone could change the underlying type. The cast I'm doing would catch that change (if the new type can't be safely cast to a uint32), a bare static_cast to uint32 would not.

There will be a std::to_underlying() function to cast enums in C++ 23.

dev/PushNotifications/PushNotificationManager.cpp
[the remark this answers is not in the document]
loneursid: Do we usually log everything in telemetry? I had tried to log actual successes and failures but not thinks like a developer passing invalid args. hence why I hadn't wrapped the entire function inside a try / catch construct. I did that same for CreateChannelAsync, where I did not wrap the function in a try / catch construct and whilingly let exceptions like invalid args go by without logging them. 

Is it our practice to systematically wrap a function in a try / catch and log everything, even dev not respecting the calling contract?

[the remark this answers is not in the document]
loneursid: Done. I've wrapped CreateChannelAsync, RegisterActivator and UnregisterActivator in a try / catch contruct

dev/PushNotifications/PushNotificationManager.cpp
loneursid: Also addressed

dev/PushNotifications/PushNotificationManager.cpp
danielayala94: Thanks for pointing it out. Then the approach as is it's fine.

danielayala94 (approved):

PR #1339, opened 2021-08-30
Removed PushNotificationToken

pmpurifoy:
This PR modifies all registration APIs for PushNotifications and renames PushNotificationRegistrationOptions->PushNotificationRegistrationActivators.

static void RegisterActivator(Microsoft::Windows::PushNotifications::PushNotificationActivationInfo const& details);
static void UnregisterActivator(Microsoft::Windows::PushNotifications::PushNotificationRegistrationActivators const& activators);

Developers will no longer have to keep track of the registration tokens.


dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -63,7 +63,7 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
         }
     }
 
-    winrt::IAsyncOperationWithProgress<winrt::Microsoft::Windows::PushNotifications::PushNotificationCreateChannelResult, winrt::Microsoft::Windows::PushNotifications::PushNotificationCreateChannelStatus> PushNotificationManager::CreateChannelAsync(const winrt::guid &remoteId)
+    winrt::IAsyncOperationWithProgress<winrt::Microsoft::Windows::PushNotifications::PushNotificationCreateChannelResult, winrt::Microsoft::Windows::PushNotifications::PushNotificationCreateChannelStatus> PushNotificationManager::CreateChannelAsync(const winrt::guid& remoteId)
     {
```
jonwis, line 67: I saw another PR like this - any member method that uses co_await needs to capture a strong reference to "this" like:

```c++
IAsyncOperation<Q> Foo::ThingAsync()
{
    auto lifetime = get_strong();
    // the rest of the code
}
```

... without this you can accidentally final-deref the type while an async operation is running.

jonwis (changes requested):

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -223,26 +233,45 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-            auto taskRegistration = token.TaskRegistration();
-            THROW_HR_IF_NULL(HRESULT_FROM_WIN32(ERROR_NOT_FOUND), taskRegistration);
-            taskRegistration.Unregister(true);
+            s_taskRegistration.Unregister(true);
+            s_taskRegistration = nullptr;
         }
 
-        // Check for COM flag, a valid cookie and if there are outstanding locks on the PushNotificationBackgroundTask class factory
-        if (WI_IsFlagSet(options, PushNotificationRegistrationOptions::ComActivator) && token.Cookie() && winrt::get_module_lock() == 0)
+        if (s_cookie)
```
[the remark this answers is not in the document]
pmpurifoy: s_cookie is currently just static uint32_t s_cookie = 0;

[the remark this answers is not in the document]
pmpurifoy: Should change wrap this in a wil object?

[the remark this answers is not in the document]
danielayala94: Check this out: `unique_com_class_object_cookie`

https://github.com/Microsoft/wil/wiki/RAII-resource-wrappers

dev/PushNotifications/PushNotificationManager.h
```diff
@@ -6,16 +6,21 @@
 
 namespace winrt::Microsoft::Windows::PushNotifications::implementation
 {
+    static winrt::Windows::ApplicationModel::Background::IBackgroundTaskRegistration s_taskRegistration = nullptr;
+    static uint32_t s_cookie = 0;
```
danielayala94, line 10: Seems like you can use `wil::unique_com_class_object_cookie here`.

dev/PushNotifications/PushNotificationManager.h
```diff
@@ -6,16 +6,21 @@
...
-        static void UnregisterActivator(Microsoft::Windows::PushNotifications::PushNotificationRegistrationToken const& token, Microsoft::Windows::PushNotifications::PushNotificationRegistrationOptions const& options);
+        static void RegisterActivator(Microsoft::Windows::PushNotifications::PushNotificationActivationInfo const& details);
+        static void UnregisterActivator(Microsoft::Windows::PushNotifications::PushNotificationRegistrationActivators const& activators);
+        static void UnregisterAllActivators();
 
-        static winrt::Windows::Foundation::IAsyncOperationWithProgress<winrt::Microsoft::Windows::PushNotifications::PushNotificationCreateChannelResult, winrt::Microsoft::Windows::PushNotifications::PushNotificationCreateChannelStatus> CreateChannelAsync(const winrt::guid &remoteId);
+        static winrt::Windows::Foundation::IAsyncOperationWithProgress<winrt::Microsoft::Windows::PushNotifications::PushNotificationCreateChannelResult, winrt::Microsoft::Windows::PushNotifications::PushNotificationCreateChannelStatus> CreateChannelAsync(const winrt::guid& remoteId);
```
danielayala94, line 21: nit: I think we've been sticking to the `winrt::guid const& remoteId` syntax

dev/PushNotifications/PushNotifications.idl
```diff
@@ -22,26 +22,29 @@ namespace Microsoft.Windows.PushNotifications
     };
 
     [flags]
-    enum PushNotificationRegistrationOptions
+    enum PushNotificationRegistrationActivators
     {
-        Undefined = 0x0, // Used to check if 0 was passed as a flag
+        Undefined = 0, // Default enum to catch nullptr case when no flag is defined
```
danielayala94, line 27: Technically there isn't a nullptr concept in enums, we simply don't have flags.

dev/PushNotifications/PushNotifications.idl
```diff
@@ -22,26 +22,29 @@ namespace Microsoft.Windows.PushNotifications
...
         PushTrigger = 0x1, // Registers a Push Trigger with Background Infrastructure
-        ComActivator = 0x2, // Registers the Windows App SDK Background Task component as an InProc COM server
-        ProtocolActivator = 0x4 // Registers an application with the PushNotificationsLongRunningTask to be activated via protocol
+        ComActivator = 0x2, // Registers the Background Task component as an in-process COM server
+        ProtocolActivator = 0x4, // Registers an application with the PushNotificationsLongRunningTask to be activated via protocol
     };
 
-    [experimental]
+    // An abstraction over the activation Registration flow
```
danielayala94, line 33: What does this mean?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -17,11 +17,11 @@
 #include "PushNotificationChannel.h"
 #include "externs.h"
 #include <string_view>
+#include <iostream>
```
danielayala94, line 20: What API is using this library?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -128,22 +128,27 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
         DWORD cookie = 0;
         IBackgroundTaskRegistration registeredTask = nullptr;
         BackgroundTaskBuilder builder = nullptr;
 
-        if (WI_IsFlagSet(registrationOptions, PushNotificationRegistrationOptions::PushTrigger))
+        if (WI_IsFlagSet(registrationOptions, PushNotificationRegistrationActivators::PushTrigger))
         {
+            {
+                auto lock = s_lock.lock_exclusive();
+                THROW_HR_IF(E_INVALIDARG, s_taskRegistration);
```
danielayala94, line 149: Is it intentional to throw if the developer calls RegisterActivator twice for the same registrationOption?

dev/PushNotifications/PushNotificationManager.h
```diff
@@ -6,16 +6,21 @@
 
 namespace winrt::Microsoft::Windows::PushNotifications::implementation
 {
+    static winrt::Windows::ApplicationModel::Background::IBackgroundTaskRegistration s_taskRegistration = nullptr;
+    static uint32_t s_cookie = 0;
+    static wil::srwlock s_lock;
```
danielayala94, line 11: nit: `s_activatorInfoLock`, so we put it clear why we need the lock.

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: Must be leftover from testing.

dev/PushNotifications/PushNotifications.idl
pmpurifoy: Not sure why this is there but it should just be experimental.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -161,15 +160,20 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
             GUID taskClsid = details.TaskClsid();
             THROW_HR_IF(E_INVALIDARG, taskClsid == GUID_NULL);
 
-            auto registrationOptions = details.Options();
-            THROW_HR_IF(E_INVALIDARG, WI_AreAllFlagsClear(registrationOptions, PushNotificationRegistrationOptions::PushTrigger | PushNotificationRegistrationOptions::ComActivator));
+            auto registrationActivators = details.Activators();
```
DrusTheAxe, line 163: NIT: `auto registrationActivators{ details.Activators() };`

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -161,15 +160,20 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+            THROW_HR_IF(E_INVALIDARG, WI_AreAllFlagsClear(registrationActivators, PushNotificationRegistrationActivators::PushTrigger | PushNotificationRegistrationActivators::ComActivator));
 
             DWORD cookie = 0;
             IBackgroundTaskRegistration registeredTask = nullptr;
```
DrusTheAxe, line 167: NIT: `IBackgroundTaskRegistration registeredTask{};`

see previous coment

dev/PushNotifications/PushNotificationManager.h
```diff
@@ -6,16 +6,21 @@
 
 namespace winrt::Microsoft::Windows::PushNotifications::implementation
 {
+    static winrt::Windows::ApplicationModel::Background::IBackgroundTaskRegistration s_taskRegistration = nullptr;
+    static wil::unique_com_class_object_cookie s_cookie;
+    static wil::srwlock s_activatorInfoLock;
```
DrusTheAxe, line 11: These are static global variables. Doesn't this mean if 2 .cpp #include this header they each get their own copy of the variables?

Are you intending this to be unique to a single global variable across the feature?

DrusTheAxe (commented):
LGTM (some minor comments) except for the question about the static  global variables

dev/PushNotifications/PushNotificationManager.h
pmpurifoy: It should be unique to just PushNotificationManager.cpp

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -239,8 +236,13 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+            if (WI_IsFlagSet(registrationActivators, PushNotificationRegistrationActivators::ComActivator))
             {
+                {
+                    auto lock = s_activatorInfoLock.lock_exclusive();
+                    THROW_HR_IF_MSG(E_INVALIDARG, s_cookie, "ComActivator already registered.");
+                }
+
                 GetWaitHandleForArgs().create();
 
                 THROW_IF_FAILED(::CoRegisterClassObject(
```
danielayala94, line 248: You should just call `s_cookie.reset()`. Please look in ReSearch how this wil wrapper is used.

dev/PushNotifications/PushNotificationManager.h
danielayala94: I don't see a reason for the statics to be on the header. They can be placed in the cpp, so no other headers can reference them.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -251,73 +253,82 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-        }
+            auto lock = s_activatorInfoLock.lock_exclusive();
+            if (s_taskRegistration)
+            {
+                s_taskRegistration.Unregister(true);
+                s_taskRegistration = nullptr;
+            }
 
-        if (s_cookie)
+            if (s_cookie)
```
danielayala94, line 320: No need of an if check. Just calling reset is fine.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -251,73 +253,82 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-                REGCLS_MULTIPLEUSE,
-                &cookie));
-        }
-
-            PushNotificationRegistrationToken token = { cookie, registeredTaskFromBuilder };
             scopeExitToCleanRegistrations.release();
 
-            PushNotificationTelemetry::ActivatorRegisteredByApi(S_OK, details.Options());
+            auto lock = s_activatorInfoLock.lock_exclusive();
+            s_cookie.reset(cookie);
```
danielayala94, line 264: This is wrong. The usage of this wil wrapper should be something like:

`CoRegisterClassObject(..., &s_cookie);`

Then you just call `reset()` whenever you need to revoke the class object.

Having said that, probably it's not `s_cookie` but `s_comActivatorRegistration` or similar;

danielayala94 (changes requested):

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -251,51 +258,77 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
                 registeredTaskFromBuilder = builder.Register();
             }
 
-            PushNotificationRegistrationToken token = { cookie, registeredTaskFromBuilder };
             scopeExitToCleanRegistrations.release();
 
-            PushNotificationTelemetry::ActivatorRegisteredByApi(S_OK, details.Options());
+            auto lock = s_activatorInfoLock.lock_exclusive();
+            s_cookie.reset(cookie);
```
danielayala94, line 264: What's the reason of doing this in two steps instead of this?

`CoRegisterClassObject(..., &s_cookie);`

Also, probably it's not `s_cookie` but `s_comActivatorRegistration` or similar;

pmpurifoy: In the previous implementation, we didn't want to return the s_cookie unless we made it to the end of the function. We can just call s_cookie.reset() in the scope_exit.

danielayala94 (approved):

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -216,86 +220,110 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-        catch (...)
+        catch(...)
         {
             HRESULT hrError = wil::ResultFromCaughtException();
-            PushNotificationTelemetry::ActivatorRegisteredByApi(hrError,
-                details == nullptr ? PushNotificationRegistrationOptions::Undefined : details.Options());
+            PushNotificationTelemetry::ActivatorRegisteredByApi(hrError, 
+                details == nullptr ? PushNotificationRegistrationActivators::Undefined : details.Activators());
 
             THROW_HR(hrError);
```
jonwis, line 274: Consider removing this and just saying `throw;` here which propagates the current exception instead of effectively creating a new one.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -216,86 +220,110 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-            if (WI_IsFlagSet(options, PushNotificationRegistrationOptions::PushTrigger))
+            auto lock = s_activatorInfoLock.lock_exclusive();
+            if (WI_IsFlagSet(activators, PushNotificationRegistrationActivators::PushTrigger))
             {
-                auto taskRegistration = token.TaskRegistration();
-                THROW_HR_IF_NULL(HRESULT_FROM_WIN32(ERROR_NOT_FOUND), taskRegistration);
-                taskRegistration.Unregister(true);
+                THROW_HR_IF_NULL_MSG(HRESULT_FROM_WIN32(ERROR_NOT_FOUND), s_pushTriggerRegistration, "PushTrigger not registered.");
+                s_pushTriggerRegistration.Unregister(true);
+                s_pushTriggerRegistration = nullptr;
```
jonwis, line 287: So push triggers are one-registration-per-process? Should this be refcounted somehow?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -216,86 +220,110 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+            }
+
+            s_comActivatorRegistration.reset();
+        }
+        catch(...)
+        {
+            HRESULT hrError = wil::ResultFromCaughtException();
+            PushNotificationTelemetry::ActivatorUnregisteredByApi(hrError, PushNotificationRegistrationActivators::PushTrigger | PushNotificationRegistrationActivators::ComActivator);
```
jonwis, line 323: Does the OSS version of WIL's telemetry helpers support creating Activities?  Those can watch the thread for thrown exceptions and log on them, rather than you having to wrap everything in try/catch(e){log(e);throw;} sequences.

jonwis (approved):

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: PushTriggers are one-registration-per-process. Also, this is a static winrt object, does it need ref counting?

dev/PushNotifications/PushNotificationManager.cpp
loneursid: Captured here: https://github.com/microsoft/WindowsAppSDK/issues/1391

PR #1340, opened 2021-08-30
Unpackaged notification unit tests

loneursid:
Adding unit tests for unpackaged notifications


test/PushNotificationTests/APITests.cpp
```diff
@@ -188,72 +197,142 @@ namespace Test::PushNotifications
...
+            RunTestUnpackaged(L"BackgroundActivationTest", testWaitTime()); // Need to launch one time to enable background activation.
+
+            auto LocalBackgroundTask1 = winrt::create_instance<winrt::Windows::ApplicationModel::Background::IBackgroundTask>(c_comServerId, CLSCTX_ALL);
+            auto mockBackgroundTaskInstance1 = winrt::make<MockBackgroundTaskInstance>();
+
+            auto LocalBackgroundTask2 = winrt::create_instance<winrt::Windows::ApplicationModel::Background::IBackgroundTask>(c_comServerId, CLSCTX_ALL);
+            auto mockBackgroundTaskInstance2 = winrt::make<MockBackgroundTaskInstance>();
+
+            VERIFY_NO_THROW(LocalBackgroundTask1.Run(mockBackgroundTaskInstance1));
+            VERIFY_NO_THROW(LocalBackgroundTask2.Run(mockBackgroundTaskInstance2));
```
danielayala94, line 211: Nit: This is the same code as in the equivalent packaged-app test. We probably want to have one common function for this and call it in both tests. It's fine, I won't block the PR because of this.

test/PushNotificationTests/APITests.cpp
```diff
@@ -173,6 +173,15 @@ namespace Test::PushNotifications
             VERIFY_NO_THROW(LocalBackgroundTask.Run(mockBackgroundTaskInstance));
         }
 
+        TEST_METHOD(BackgroundActivation_Unpackaged)
+        {
+            RunTestUnpackaged(L"BackgroundActivationTest", testWaitTime()); // Need to launch one time to enable background activation.
+
+            auto LocalBackgroundTask = winrt::create_instance<winrt::Windows::ApplicationModel::Background::IBackgroundTask>(c_comServerId, CLSCTX_ALL);
+            auto mockBackgroundTaskInstance = winrt::make<MockBackgroundTaskInstance>();
+            VERIFY_NO_THROW(LocalBackgroundTask.Run(mockBackgroundTaskInstance));
```
danielayala94, line 182: Similar comment as in `MultipleBackgroundActivation_Unpackaged` test.

danielayala94 (commented):
Just a couple of nits. I'm good with this approach 😊.

test/PushNotificationTests/APITests.cpp
loneursid: These are unit tests, having two with some duplicated code itsn't a big deal at this time. Plus, we have a deliverable to move as many tests as possible from the test app, into this file. When this happen, there will be significant code churn and refactoring opportunities

test/PushNotificationTests/APITests.cpp
loneursid: Similar answer :)

danielayala94 (approved):

PR #1342, opened 2021-08-31
Fix ".Experimental" VSIX renaming

DefaultRyan:
The newly-split Cpp and Cs VSIX projects now use `.Experimental` in the Id for experimental builds, but there was logic leftover from the old legacy VSIX that said `.Preview`.


rohanp-msft (approved):
:shipit:

alwu-msft:
This is superseded by https://github.com/microsoft/WindowsAppSDK/pull/1341.

PR #1346, opened 2021-08-31
Build Information spec

DrusTheAxe:
Spec for the build information 'API' (really just a bunch of constants), for use by code and toolchains to facilitate automating apps automagically referencing the WinAppSDK version they're build against.

We use a form of this today in the build pipeline and parts of the build. This specs a public form of the information, derived from current practice but refined (inconsistencies smoothed out etc)


specs/Deployment/BuildInformation.md
```diff
@@ -0,0 +1,226 @@
...
+  - [3.4. WindowsAppSDK-VersionInfo.xml](#34-windowsappsdk-versioninfoxml)
+
+# 1. Background
+
+Windows App SDK supports deployment as a set of MSIX packages for use at runtime, but some
+operations and APIs require build information about the release. This spec defines public symbols
+provided so developers can easily use Windows App SDK APIs at runtime referencing the version of
+Windows App SDK they use at build-time.
+
+For example,  `MddBootstrapInitialize(majorMinorRelease, versionTag, minVersion)` requires...
```
BenJKuhn, line 18: Assuming this is code compiled into the client, can we provide default values for these arguments so that when most developers want to just "use the version I'm building with", they can provide no explicit arguments?

specs/Deployment/BuildInformation.md
```diff
@@ -0,0 +1,226 @@
...
+## 2.1. Examples - C++
+
+```c++
+#include <WindowsAppSDK-VersionInfo.h>
+
+int main()
+{
+    const uint32_t c_majorMinorRelease{ WINDOWSAPPSDK_RELEASE_MAJOR_MINOR };
+    auto c_versionTag{ WINDOWSAPPSDK_RELEASE_VERSION_TAG_W };
+    const PACKAGE_VERSION c_minVersion{ WINDOWSAPPSDK_MSIX_VERSION_UINT64 };
```
riverar, line 65: This example shows `minVersion` getting initialized with what appears to be the _current_ framework package version delivered via the WAS release. This means apps compiled over several releases of the WAS may create linkage to different versions of WAS framework packages and introduce unintended behavior.

I understand that sounds a bit contrived as a developer would have to update the associated WAS package version for this to happen but consider that WAS has a bunch of components/tooling within and developers may want to get fixes to those without changing app targets.

To do so, however, would require a few more stable constants, e.g. WINDOWSAPPSDK_MSIX_**MIN**VERSION_UINT64 , xxx_VERSION_1_0, etc. (See also: WINVER.)

specs/Deployment/BuildInformation.md
DrusTheAxe: Not today

Baking those constants into APIs is problematic due to build-chain chicken/egg ordering issues). I have some ideas how we might be able to tackle that in the future (or at least get closer) but it's not a simple or singular answer so not in this pass.

specs/Deployment/BuildInformation.md
DrusTheAxe: > This example shows `minVersion` getting initialized with what appears to be the _current_ framework package version delivered via the WAS release.

Yes. This means "I expect to run with WinAppSDK.version >= what I build against"

> This means apps compiled over several releases of the WAS may create linkage to different versions of WAS framework packages

Yes. ByDesign. See [2.2. Decision 2: Breaking Change Boundary for Version 1.x](https://github.com/microsoft/WindowsAppSDK/blob/main/specs/Deployment/MSIXPackageVersioning.md#22-decision-2-breaking-change-boundary-for-version-1x) for more details.

> I understand that sounds a bit contrived as a developer would have to update the associated WAS package version (e.g. PackageReference) for this to happen but consider that WAS has a bunch of components/tooling within and developers may want to get fixes to those without changing app targets. (The WAS does proport to follow semver.)

Sure. Trivial example: I build with 1.2.0 (stable) which has Microsoft.WindowsAppRuntime.1.2 version 2.3.4.5 (yeah, exact values are more complicated but not relevant for the point. Roll with it). That means you need to run with Microsoft.WindowsAppRuntime.1.2 version >= 2.3.4.5. And lets say you find a DivideByZero bug which is fixed in 1.2.1 with MSIX package version 2.3.4.6, and the fix requires you do nothing more than run with the updated MSIX packages - no need to recompile all your code with the newer NuGet. This is trivially accomplished with a 1 line change from

```
const PACKAGE_VERSION c_minVersion{ WINDOWSAPPSDK_MSIX_VERSION_UINT64 };
```

to

```
const PACKAGE_VERSION c_minVersion{ 0x0002000300040006 };
```

> To do so, however, would require a few more stable constants, e.g. WINDOWSAPPSDK_MSIX_**MIN**VERSION_UINT64 , xxx_VERSION_1_0, etc. (See also: WINVER.)

That assumes if you build with WinAppSDK 1.2.3 that you could run with the runtime bits for WinAppSDK 1.2.2.

That's a dangerous assumption. Yes, it _may_ be possible, in some cases. But as a general rule it's not. In the general  case, you can't/won't run with older bits than you're built against.

In the advanced scenario where you need this more nuanced behavior, you need more nuanced code (like above, or equivalent - you decide the minversion, not simply whatever-i-compiled-against).

riverar: Ah, I was operating with the assumption that WAS was principally semver compliant ([per the docs](https://github.com/microsoft/WindowsAppSDK/blob/f05b0d0d7a7fefbfdad590c4e72fe463e052d426/docs/README.md#versioning)). Given that's not the case, we should scrub the docs.

DrusTheAxe: > Ah, I was operating with the assumption that WAS was principally semver compliant ([per the docs](https://github.com/microsoft/WindowsAppSDK/blob/main/docs/README.md#Versioning)). Given that's not the case, we should scrub the docs.

In what way is WAS not principally server compliant? Are you specifically referring to this sentence

> ...We'll use SemVer 2 to clearly identify breaking API shape levels that let those APIs evolve forward towards new functionality while not compromising existing apps.

?

DrusTheAxe: +@andrewleader @EHO-makai  for visibility

riverar: > In what way is WAS not principally server [sic] compliant? Are you specifically referring to this sentence [...]

You stated earlier that WAS 1.2.3 is not compatible with 1.2.2 runtimes. That is not semver compliant.

The WAS repo documentation states the project will [use SemVer 2 to clearly identify breaking API shape levels that let those APIs evolve forward towards new functionality while not compromising existing apps.](https://github.com/microsoft/WindowsAppSDK/blob/main/docs/README.md#:~:text=we'll%20use%20semver%202%20to%20clearly%20identify%20breaking%20api%20shape%20levels%20that%20let%20those%20apis%20evolve%20forward%20towards%20new%20functionality%20while%20not%20compromising%20existing%20apps.)

We could have fun and pick that sentence apart, but I believe WAS [release channel / servicing](https://github.com/microsoft/WindowsAppSDK/discussions/637) was also designed around NuGet semver.

I'm okay with the PR, so happy to move the discussion, but I think we really need to tread carefully and communicate clearly that apps compiled against WAS 1.0.3 may blow up on WAR 1.0.2. It's not supported. Semver is out the window for all of WAS as a result.

DrusTheAxe: > You stated earlier that WAS 1.2.3 is not compatible with 1.2.2 runtimes. That is not semver compliant.

Let me clarify, to make sure we're on the same page

* If you build with 1.2.3 you're expecting to run with >=1.2.3
* If you build with 1.2.3 you're not expecting to run with <1.2.3

Aren't those valid statements?

What about SemVer says you can build against 1.2.3 and expect to run on 1.2.2?

riverar: I did expect to build with WAS 1.2.3 and successfully target WAR 1.2.2. I had semver on the mind, and the overall SDK only incremented its patch level. Why would I expect otherwise?

This is similar to how the .NET SDK operates today. .NET SDK 5.0.1234, for example, can target .NET Runtime 5.0 (and below, to some point). And similar to how the Windows 11 SDK can build apps that run on Windows 10.

riverar: > What about SemVer says you can build against 1.2.3 and expect to run on 1.2.2?

Missed your question -- The PATCH component of the version [is used to signal](https://semver.org/#:~:text=patch%20version%20when%20you%20make%20backwards%20compatible%20bug%20fixes.) backwards compatible bug fixes. WAS 1.2.2 to WAS 1.2.3 signals that Microsoft made a small bug fix and maintains complete compatibility with WAS 1.2.2 and all its sub-components.

DrusTheAxe: > I did expect to build with WAS 1.2.3 and successfully target WAR 1.2.2

Thank you for the input. That's interesting, as it hadn't been considered a priority. Yes, it's possible, but in the ItJustWorks-simplest answer, one is expected to require &gt;= what you build with. I'll bring this back to folks to chew on.

The build info here is a bill-of-goods of what's the build-time artifacts about/good-for. Certainly if we change a bit in code you don't use no one says you must run with the newer/matching version. But it's hard to understand that in the abstract. Still quite possible, e.g. if you want to build with 1.0.9876 &lt;g&gt; but run with >= 1.0.1 you can do that via `MddBootstrapInitalize(...,minversion=0x0001000000010000)`. You just don't get a pretty magic constant for that.

You _do_ get a set of constants so you could write

```
PACKAGE_VERSION minVersion{};
minVersion.Major = WINDOWSAPPSDK_MSIX_VERSION_MAJOR;
minVersion.Minor = 123;
minVersion.Build = 456;
minVersion.Revision = 789;
MddBootstrapInitialize(0x00010000, nullptr, minVersion);
```

and you're get at runtime Microsoft.WindowsAppRuntime.1.0 version >= 0.123.456.789, regardless what you build against.

P.S. I now understand the nuance of what you meant when you mentioned `WINVER`. The comment about " .NET SDK 5.0.1234, for example, can target .NET Runtime 5.0" and Win11SDK/10runtime made it clear. Thanks!

riverar: Sounds good, thanks! In the meantime, I'll make sure we (and others) don't hardcode those constants and then get out of sync. Looking very much forward to this PR getting in!

jonwis (approved):

specs/Deployment/BuildInformation.md
```diff
@@ -0,0 +1,226 @@
...
+using Microsoft.Windows.ApplicationModel.DynamicDependency;
+
+namespace MyNamespace
+{
+    class MyClass
+    {
+        static void Main()
+        {
+            // Initialize access to Windows App SDK
+            var majorMinorRelease = Microsoft.WindowsApp.Release.MajorMinor;
```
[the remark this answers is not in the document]
DrusTheAxe: WindowsAppRuntime seems the better fit

specs/Deployment/BuildInformation.md
```diff
@@ -0,0 +1,226 @@
...
+            var versionTag = Microsoft.WindowsApp.Release.VersionTag;
+            Microsoft.WindowsApp.PackageVersion minVersion{ Microsoft.WindowsApp.MSIX.Version.UInt64 };
+            Bootstrap.Initialize(majorMinorRelease, versionTag, minVersion));
```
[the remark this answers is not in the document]
DrusTheAxe: I'd like that too but we have build pipeline work to sort out to enable that.

This addresses the urgent need - hand-editing these values, some of which can be pretty cryptic (MSIX version info's only really knowable if you crack open a package and check its manifest, or install it and Get-AppxPackage). This is also a bit verbose for clarity's sake. You could just as easily replace

```
var majorMinorRelease = Microsoft.WindowsApp.Release.MajorMinor;
var versionTag = Microsoft.WindowsApp.Release.VersionTag;
Microsoft.WindowsApp.PackageVersion minVersion{ Microsoft.WindowsApp.MSIX.Version.UInt64 };
Bootstrap.Initialize(majorMinorRelease, versionTag, minVersion));
```

with

```
Bootstrap.Initialize(Microsoft.WindowsApp.Release.MajorMinor, Microsoft.WindowsApp.Release.VersionTag, new Microsoft.WindowsApp.PackageVersion(Microsoft.WindowsApp.MSIX.Version.UInt64)));
```

or

```
using Microsoft.WindowsApp;
...
Bootstrap.Initialize(Release.MajorMinor, Release.VersionTag, new PackageVersion(MSIX.Version.UInt64)));
```

etc. But now we're getting into C#isms (would an Alias be preferable to using? etc) with no obvious best answer. We can iron out preferred C# syntax via the samples and docs process (and if it differs we can always update this doc later with that hashed out result).

specs/Deployment/BuildInformation.md
```diff
@@ -0,0 +1,226 @@
...
+* Minimum required version of MSIX packages
+
+If a developer builds an unpackaged app with Windows App SDK 1.0 Experimental1 with MSIX package
+version of 0.218.840.0, they first must initialize Bootstrap API with this information. This is
+accomplished with the call
+
+```c++
+const uint32_t c_majorMinorRelease{ (1u << 16) | 0 };
+auto c_versionTag{ "experimental1" };
+const PACKAGE_VERSION c_minVersion{ (0LLU << 48) | (218LLU << 32) | (840LLU << 16) | 0 };
```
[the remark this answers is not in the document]
DrusTheAxe: See line 40+ :-)

This is the raw "here's what you have to do without this spec"

Line 40+ is the "here's how much nicer it is with this spec"

specs/Deployment/BuildInformation.md
```diff
@@ -0,0 +1,226 @@
...
+
+And each time the WinAppSDK used by the build is updated this must be changed to match. This spec
+allows developers to use constants provided by WinAppSDK ensuring the application's runtime use of
+WinAppSDK's build information always matches the WinAppSDK used at build-time (without manual
+developer intervention):
+
+```c++
+const uint32_t c_majorMinorRelease{ WINDOWSAPPSDK_RELEASE_MAJOR_MINOR };
+auto c_versionTag{ WINDOWSAPPSDK_RELEASE_VERSION_TAG_W };
+const PACKAGE_VERSION c_minVersion{ WINDOWSAPPSDK_MSIX_VERSION_UINT64 };
```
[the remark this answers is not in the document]
DrusTheAxe: The version _is_ MSIX specific, or at least, how we express WinAppRuntime's version in MSIX (we use different values in other tech, e.g. NuGet). If WinAppSDK was used in an AppLocal way you'd have the WinAppSDK runtime (bunch of files with your app) but there'd be no MSIX package and nowhere would this UInt64 value exist.

If you're still comfortable with Runtime instead of MSIX then sure, we can roll with that

specs/Deployment/BuildInformation.md
```diff
@@ -0,0 +1,226 @@
...
+
+If a developer builds an unpackaged app with Windows App SDK 1.0 Experimental1 with MSIX package
+version of 0.218.840.0, they first must initialize Bootstrap API with this information. This is
+accomplished with the call
+
+```c++
+const uint32_t c_majorMinorRelease{ (1u << 16) | 0 };
+auto c_versionTag{ "experimental1" };
+const PACKAGE_VERSION c_minVersion{ (0LLU << 48) | (218LLU << 32) | (840LLU << 16) | 0 };
+THROW_IF_FAILED(MddBootstrapInitialize(c_majorMinorRelease, c_versionTag, minVersion));
```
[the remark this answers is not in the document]
DrusTheAxe: > This `MddBootstrap` is still WinAppSDK-specific, right?

Yes

> Maybe that's a v2, where we have WinAppSDKInitialize()

Indeed.

We've discussed it (multiple times) but held off so far. We're actively avoiding reinventing the `CoreApplication` requried-monolithic problem. Features that need initialization work are providing their own init function. We can create a WinAppSDKInitialize() 'super init' that calls all those, as a _convenience_, so long as we don't _require_ it.

But we have few init functions so far so ROI's low to date. But yes, a future consideration.

DrusTheAxe:
@jonwis Any preference on filenames?

* WindowsAppSDK-VersionInfo.cs
* WindowsAppSDK-VersionInfo.h
* WindowsAppSDK-VersionInfo.json
* WindowsAppSDK-VersionInfo.xml

Or should this be...
* Microsoft.WindowsAppSDK.VersionInfo.*
* Microsoft.WindowsAppRuntime.VersionInfo.*
* ?

BenJKuhn (approved):

specs/Deployment/BuildInformation.md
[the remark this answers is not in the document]
jonwis: In my mind, this is the "version of the Windows App Runtime your app wants." MSIX is an implementation detail.  I'd go with "Runtime" as the magic token here, rather than MSIX. I wouldn't block on it, though, because in 1.1 you're going to add that single-like WinAppSDKInitialize() which hides all these constants behind a generated method. :)

specs/Deployment/BuildInformation.md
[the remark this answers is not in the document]
DrusTheAxe: Actually...WindowsAppSDK is a better fit given the MSIX->Runtime change

PR #1354, opened 2021-09-01
Revert "MRTCore namespace from Microsoft.ApplicationModel.Resources to Microsoft.Windows.ApplicationModel.Resources"

kythant:
Reverts microsoft/WindowsAppSDK#1328

Temporarily revert this change to unblock the checks in the nightly build.


huichen123:
@MikeHillberg  when do you expect we will have a new WinUI transport package?

PR #1358, opened 2021-09-01
Notification delivery + Protocol Activation

danielayala94:
This PR implements `INotificationListener`, which will handle incoming notifications from wpncore and route them to unpackaged apps via foreground or through protocol activation (a.k.a. background notifications).

Also, I have introduced `NotificationListenerManager`, which will cache the long running sinks for every app tracked on the LRP.

Finally, in this PR I intend to fill the remaining implementation gaps in the LRP platform (not 100% yet because we're waiting for this PR to be merged into WNP_LRP branch, specially because it has the logic to store appId/processName in disk -> #1290) and adding Protocol activation functionality in the SDK.


dev/AppLifecycle/AppInstance.cpp
```diff
@@ -337,14 +337,20 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
             // protocol, except the catch-all LaunchActivatedEventArgs case.
             if (!contractArgument.empty())
             {
-                if (contractData.empty())
+                if (CompareStringOrdinal(contractArgument.data(), static_cast<int>(contractArgument.size()), L"WindowsAppSdkPushServer", -1, TRUE) == CSTR_EQUAL)
```
danielayala94, line 340: This is not what I intended to do, something happened during merging. Fixing.

dev/PushNotifications/PushNotificationReceivedEventArgs.h
```diff
@@ -10,8 +10,12 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
 
         PushNotificationReceivedEventArgs(winrt::Windows::ApplicationModel::Background::IBackgroundTaskInstance const& backgroundTask);
         PushNotificationReceivedEventArgs(winrt::Windows::Networking::PushNotifications::PushNotificationReceivedEventArgs const& args);
+
+        PushNotificationReceivedEventArgs(std::wstring& payload);
+
         PushNotificationReceivedEventArgs(byte* const& payload, ULONG const& length);
 
+
```
danielayala94, line 18: No need of extra line.

dev/AppLifecycle/AppInstance.cpp
```diff
@@ -337,14 +337,20 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
             // protocol, except the catch-all LaunchActivatedEventArgs case.
             if (!contractArgument.empty())
             {
-                if (contractData.empty())
+                // If the contractData is empty, handle any aliased encoded launches.
```
aeloros, line 340: This comment is no longer correct given your change.

Edit: I think if you just reword it to not be about contractData being empty it will make sense.  Maybe just calling out that we're looking for an aliased launches.

dev/AppLifecycle/AppInstance.cpp
```diff
@@ -337,14 +337,20 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
...
+                    if (contractData.empty())
                     {
                         contractData = GenerateEncodedLaunchUri(L"App", c_pushContractId);
                         contractArgument = c_protocolArgumentString;
                     }
+                    else // Command line contained a background notification payload
+                    {
+                        kind = ExtendedActivationKind::Push;
+                        Uri pushNotificationUri{ contractData };
+                        data = winrt::Microsoft::Windows::PushNotifications::Deserialize(pushNotificationUri);
```
aeloros, line 352: What is stopping you from just using the table lookup here instead of calling Deserialize directly?

dev/PushNotifications/GetRawNotificationEventArgs.h
```diff
@@ -10,8 +10,19 @@ constexpr PCWSTR c_pushContractId = L"Windows.Push";
 
 namespace winrt::Microsoft::Windows::PushNotifications
 {
-    static winrt::Windows::Foundation::IInspectable Deserialize(winrt::Windows::Foundation::Uri const&)
+    static winrt::Windows::Foundation::IInspectable Deserialize(winrt::Windows::Foundation::Uri const& uri)
     {
+        // Verify if the uri contains a background notification payload.
+        // Otherwise, we expect to process the notification in a background task.
+        std::wstring uriAsString{ uri.AbsoluteUri() };
+        if (uriAsString._Starts_with(L"-payload:"))
```
danielayala94, line 18: I think this can be simplified if the command line payload were URI compliant from the beginning. I'll try my best to get a better solution.

dev/AppLifecycle/AppInstance.cpp
danielayala94: It is only for unpackaged apps. For packaged apps, it should be empty.

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -51,16 +57,42 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::RegisterF
...
     return E_NOTIMPL;
 }
 
+STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::RegisterActivator(_In_ PCWSTR processName) noexcept
+{
+    auto lock = m_lock.lock_shared();
+    RETURN_HR_IF(WPN_E_PLATFORM_UNAVAILABLE, m_shutdown);
+
+    m_notificationListenerManager.AddListener(processName);
+
```
danielayala94, line 72: We should cancel the timer if we go from 0 to 1 listeners in the manager.

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -51,16 +57,42 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::RegisterF
...
+STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::UnregisterActivator(_In_ PCWSTR processName) noexcept
+{
+    auto lock = m_lock.lock_shared();
+    RETURN_HR_IF(WPN_E_PLATFORM_UNAVAILABLE, m_shutdown);
+
+    m_notificationListenerManager.RemoveListener(processName);
+    m_foregroundSinkManager->Remove(processName);
+
+    // TODO: Cleanup storage app entry after Sharath's PR
+
```
danielayala94, line 85: Also, if the removed listener happens to be the last one, we should signal the event.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -161,7 +161,27 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
             auto registrationOptions = details.Options();
-            THROW_HR_IF(E_INVALIDARG, WI_AreAllFlagsClear(registrationOptions, PushNotificationRegistrationOptions::PushTrigger | PushNotificationRegistrationOptions::ComActivator));
+
+            auto isBackgroundTaskFlagSet{ WI_IsAnyFlagSet(registrationOptions, PushNotificationRegistrationOptions::PushTrigger | PushNotificationRegistrationOptions::ComActivator) };
+            THROW_HR_IF(E_INVALIDARG, isBackgroundTaskFlagSet && !IsActivatorSupported(registrationOptions));
+
+            auto isProtocolActivatorSet{ WI_IsFlagSet(registrationOptions, PushNotificationRegistrationOptions::ProtocolActivator) };
+            THROW_HR_IF(E_INVALIDARG, isProtocolActivatorSet && !IsActivatorSupported(registrationOptions));
+
+            if (isProtocolActivatorSet)
```
hulumane, line 171: nit: Instead of returning in the if ( .. ) block, I prefer an if ( ) else { } pattern so that we return only at the end of function call.

dev/PushNotifications/PushNotificationReceivedEventArgs.h
```diff
@@ -10,6 +10,9 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
 
         PushNotificationReceivedEventArgs(winrt::Windows::ApplicationModel::Background::IBackgroundTaskInstance const& backgroundTask);
         PushNotificationReceivedEventArgs(winrt::Windows::Networking::PushNotifications::PushNotificationReceivedEventArgs const& args);
+
+        PushNotificationReceivedEventArgs(std::wstring& payload);
```
hulumane, line 14: const ?

dev/PushNotifications/PushNotificationsLongRunningTask/ForegroundSinkManager.cpp
```diff
@@ -27,7 +24,7 @@ bool ForegroundSinkManager::InvokeForegroundHandlers(std::wstring const& process
         BOOL foregroundHandled = true;
         if (FAILED(it->second->InvokeAll(payloadSize, payload.data(), &foregroundHandled)))
```
hulumane, line 25: There is a bug in the foregroundsink map. All the foregroundsinks need to be stored as AgileRef's in the map so that we can invoke the main process from a different apartment context. Otherwise if the ForegroundSink was intialized in a different apartment from the callback invocation thread, all calls will end up failing.

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.h
```diff
@@ -0,0 +1,17 @@
+﻿#pragma once
+
+#include <FrameworkUdk/PushNotificationsRT.h>
+
+class NotificationListener : public Microsoft::WRL::RuntimeClass<::ABI::Microsoft::Internal::PushNotifications::INotificationListener>
```
hulumane, line 5: A NotificationListener is not a per app component. It is supposed to cover all apps. So this tells me that we have to rename these 2 components:
NotificationListener -> AppListener?
NotificationListenerManager -> NotificationListener (This is the real listener IMHO)

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListenerManager.cpp
```diff
@@ -0,0 +1,44 @@
...
+{
+    m_foregroundSinkManager = foregroundSinkManager;
+
+    for (std::wstring& processName : processNameList)
+    {
+        AddListener(processName);
+    }
+}
+
+void NotificationListenerManager::AddListener(std::wstring processName)
```
hulumane, line 15: AddAppActivator: If we decide to go with AppActivator instead of NotificationListener

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListenerManager.cpp
```diff
@@ -0,0 +1,44 @@
...
+    THROW_IF_FAILED(MakeAndInitialize<NotificationListener>(&listener, m_foregroundSinkManager, appIdFromProcessName));
+    THROW_IF_FAILED(PushNotifications_RegisterNotificationSinkForFullTrustApplication(appIdFromProcessName.c_str(), listener.Get()));
+
+    if (m_notificationListeners.find(processName) == std::end(m_notificationListeners))
+    {
+        m_notificationListeners.insert({ processName, listener });
+    }
+}
+
+void NotificationListenerManager::RemoveListener(std::wstring processName)
```
hulumane, line 37: RemoveAppActivator

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListenerManager.cpp
```diff
@@ -0,0 +1,44 @@
...
+        THROW_IF_FAILED(PushNotifications_RegisterFullTrustApplication(L"SomeAppIdentifier", remoteId));
+    }
+
+    ComPtr<NotificationListener> listener;
+    THROW_IF_FAILED(MakeAndInitialize<NotificationListener>(&listener, m_foregroundSinkManager, appIdFromProcessName));
+    THROW_IF_FAILED(PushNotifications_RegisterNotificationSinkForFullTrustApplication(appIdFromProcessName.c_str(), listener.Get()));
+
+    if (m_notificationListeners.find(processName) == std::end(m_notificationListeners))
+    {
+        m_notificationListeners.insert({ processName, listener });
```
hulumane, line 33: No locks??

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListenerManager.cpp
```diff
@@ -0,0 +1,44 @@
...
+    }
+}
+
+void NotificationListenerManager::RemoveListener(std::wstring processName)
+{
+    // To be completed after Sharath's PR
+    std::wstring appIdFromProcessName = L"appId";
+    LOG_IF_FAILED(PushNotifications_UnregisterNotificationSinkForFullTrustApplication(processName.c_str()));
+
+    m_notificationListeners.erase(processName);
```
hulumane, line 43: No locks?

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListenerManager.h
```diff
@@ -0,0 +1,20 @@
...
+public:
+    NotificationListenerManager() {};
+
+    void Initialize(std::shared_ptr<ForegroundSinkManager> foregroundSinkManager, std::vector<std::wstring>& processNameList);
+
+    void AddListener(std::wstring processName);
+    void RemoveListener(std::wstring processName);
+
+private:
+    std::map<std::wstring, Microsoft::WRL::ComPtr<NotificationListener>> m_notificationListeners;
```
hulumane, line 18: The listener parameter should be an agileref.

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListenerManager.cpp
```diff
@@ -0,0 +1,44 @@
...
+    {
+        m_notificationListeners.insert({ processName, listener });
+    }
+}
+
+void NotificationListenerManager::RemoveListener(std::wstring processName)
+{
+    // To be completed after Sharath's PR
+    std::wstring appIdFromProcessName = L"appId";
+    LOG_IF_FAILED(PushNotifications_UnregisterNotificationSinkForFullTrustApplication(processName.c_str()));
```
hulumane, line 41: I'm assuming that we mean to pass a GUID here?

hulumane: Then we should pass appIdFromProcessName instead right?

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -49,16 +55,42 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::RegisterF
...
     return E_NOTIMPL;
 }
 
+STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::RegisterActivator(_In_ PCWSTR processName) noexcept
+{
+    auto lock = m_lock.lock_shared();
+    RETURN_HR_IF(WPN_E_PLATFORM_UNAVAILABLE, m_shutdown);
+
+    m_notificationListenerManager.AddListener(processName);
```
hulumane, line 69: AddListener throws. This method is noexcept

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -49,16 +55,42 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::RegisterF
...
+
+    return S_OK;
+}
+
+STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::UnregisterActivator(_In_ PCWSTR processName) noexcept
+{
+    auto lock = m_lock.lock_shared();
+    RETURN_HR_IF(WPN_E_PLATFORM_UNAVAILABLE, m_shutdown);
+
+    m_notificationListenerManager.RemoveListener(processName);
```
hulumane, line 79: These functions throw

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListenerManager.cpp
danielayala94: Yes

danielayala94: Actually it's a bit working too hard passing processName. We can directly pass appId everywhere.

dev/PushNotifications/PushNotificationsLongRunningTask/ForegroundSinkManager.cpp
danielayala94: Yeah, I noticed that comment in the WNP_LRP PR. @pmpurifoy will address it.

aeloros (approved):
I only reviewed stuff from my area, which look fine.

Thanks!

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -161,7 +161,24 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+
+                wil::com_ptr<INotificationsLongRunningPlatform> notificationPlatform{
+                    wil::CoCreateInstance<NotificationsLongRunningPlatform, INotificationsLongRunningPlatform>(CLSCTX_LOCAL_SERVER) };
+
+                wil::unique_cotaskmem_string processName;
+                THROW_IF_FAILED(GetCurrentProcessPath(processName));
+                THROW_IF_FAILED(notificationPlatform->RegisterActivator(processName.get()));
+            }
 
             DWORD cookie = 0;
```
hulumane, line 183: Why not put the whole block below in the else clause? It's not relevant to protocol right?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -161,7 +161,24 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+
+            auto isBackgroundTaskFlagSet{ WI_IsAnyFlagSet(registrationOptions, PushNotificationRegistrationOptions::PushTrigger | PushNotificationRegistrationOptions::ComActivator) };
+            THROW_HR_IF(E_INVALIDARG, isBackgroundTaskFlagSet && !IsActivatorSupported(registrationOptions));
+
+            auto isProtocolActivatorSet{ WI_IsFlagSet(registrationOptions, PushNotificationRegistrationOptions::ProtocolActivator) };
+            THROW_HR_IF(E_INVALIDARG, isProtocolActivatorSet && !IsActivatorSupported(registrationOptions));
+
+            if (isProtocolActivatorSet)
+            {
+                auto coInitialize = wil::CoInitializeEx();
```
hulumane, line 173: Why is this block still here? It can be moved to the block below right? 
if (WI_IsFlagSet(options, PushNotificationRegistrationOptions::ProtocolActivator))

hulumane: Basically the logic in this API should be simple if (protocol) { ... } else if (backgroundtrigger) ... else {}.

Right now the ordering and structure could use some work

dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
```diff
@@ -119,5 +131,17 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
             m_handledUnpackaged = value;
         }
     }
+
+    std::string PushNotificationReceivedEventArgs::Utf16ToUtf8(_In_z_ const PCWSTR utf16)
```
hulumane, line 135: isn't PCWSTR const by default? PWSTR is non-const

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListenerManager.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+void NotificationListenerManager::RemoveListener(std::wstring appId)
+{
+    auto lock = m_lock.lock_exclusive();
+
+    THROW_HR_IF(E_INVALIDARG, appId.empty());
+
+    LOG_IF_FAILED(PushNotifications_UnregisterNotificationSinkForFullTrustApplication(appId.c_str()));
+
+    m_notificationListeners.erase(appId);
```
hulumane, line 43: Should we be doing this here?

if (m_notificationListeners.size() == 0)
{
   SignalEvent();
}

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -49,24 +60,115 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::RegisterF
...
-STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::RegisterForegroundActivator(_In_ IWpnForegroundSink* sink, _In_ PCWSTR processName)
+STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::RegisterActivator(_In_ PCWSTR processName) noexcept try
+{
+    auto lock = m_lock.lock_shared();
+    RETURN_HR_IF(WPN_E_PLATFORM_UNAVAILABLE, m_shutdown);
+
+    std::wstring appId = GetAppIdentifier(processName);
+    m_notificationListenerManager.AddListener(appId, processName);
+
+    if (!m_notificationListenerManager.IsEmpty())
```
hulumane, line 77: This if check is not necessary. If the previous statement succeeded, then it can never be empty right?

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -49,24 +60,115 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::RegisterF
...
 }
 
-STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::RegisterForegroundActivator(_In_ IWpnForegroundSink* sink, _In_ PCWSTR processName)
+STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::RegisterActivator(_In_ PCWSTR processName) noexcept try
+{
+    auto lock = m_lock.lock_shared();
+    RETURN_HR_IF(WPN_E_PLATFORM_UNAVAILABLE, m_shutdown);
+
+    std::wstring appId = GetAppIdentifier(processName);
+    m_notificationListenerManager.AddListener(appId, processName);
```
hulumane, line 75: Should the app be added to storage here?

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -49,24 +60,115 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::RegisterF
...
+    }
+
+    return S_OK;
+}
+CATCH_RETURN()
+
+STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::RegisterForegroundActivator(_In_ IWpnForegroundSink* sink, _In_ PCWSTR processName) noexcept try
 {
     auto lock = m_lock.lock_exclusive();
     RETURN_HR_IF(WPN_E_PLATFORM_UNAVAILABLE, m_shutdown);
```
hulumane, line 108: Should all this be THROW_HR_IF for consistency?

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -49,24 +60,115 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::RegisterF
...
+std::map<std::wstring, std::wstring> NotificationsLongRunningPlatformImpl::GetFullTrustApps()
+{
+    std::map<std::wstring, std::wstring> mapOfFullTrustApps;
+
+    // Get list of full trust apps with valid channels from wpncore
+    wil::unique_cotaskmem_array_ptr<PWSTR> appIds;
+    PushNotifications_GetFullTrustApplicationsWithChannels(appIds.addressof(), appIds.size_address<ULONG>());
+
+    // Load list of apps from Storage
+    auto values{ m_storage.Values() };
```
hulumane, line 138: No locks? If you put a lock here, please make sure that the caller does not share the same lock(deadlock case)

dev/PushNotifications/PushNotificationsLongRunningTask/platform.h
```diff
@@ -18,18 +15,28 @@ struct __declspec(uuid(PUSHNOTIFICATIONS_IMPL_CLSID_STRING)) NotificationsLongRu
 
     STDMETHOD(RegisterFullTrustApplication)(_In_ PCWSTR processName, _In_ GUID remoteId, _Out_ GUID* appId) noexcept;
 
+    STDMETHOD(RegisterActivator)(_In_ PCWSTR processName) noexcept;
+
+    STDMETHOD(UnregisterActivator)(_In_ PCWSTR processName) noexcept;
+
     STDMETHOD(RegisterForegroundActivator)(_In_ IWpnForegroundSink* sink, _In_ PCWSTR processName);
```
hulumane, line 22: noexcept for these 2 as well

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
hulumane: All the other checks as well

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -49,24 +60,115 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::RegisterF
...
+STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::RegisterForegroundActivator(_In_ IWpnForegroundSink* sink, _In_ PCWSTR processName) noexcept try
 {
     auto lock = m_lock.lock_exclusive();
     RETURN_HR_IF(WPN_E_PLATFORM_UNAVAILABLE, m_shutdown);
 
-    m_foregroundSinkManager.Add(processName, sink);
+    std::wstring appId = GetAppIdentifier(processName);
+
+    m_notificationListenerManager.AddListener(appId, processName);
+    m_foregroundSinkManager->Add(processName, sink);
```
hulumane, line 113: @pmpurifoy  Why do we use processname as the currency for ForegroundSinkManager instead of appId? Is there a reason for that? @danielayala94 Why have 2 different currencies?

danielayala94: At the time the foreground sink was introduced, there wasn't any need of using AppId as the sink ID. Also, if from the SDK we get processName, we can directly use it to do any operation in the foreground sink manager.

But, in the big picture there is no consistency: the appId is the key (therefore, ID) in the storage and the listener ID too. I'm going to change the foreground sink manager ID to be appId.

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
danielayala94: I looked at the documentation and there is no exception safety anywhere, so I assume if AddListener doesn't throw then we have at least one element in the listener manager.

dev/PushNotifications/PushNotificationManager.cpp
danielayala94: There is no such if in this function. Also, if we return at a single place then we have to return a null token (i.e. cookie is 0 and there is no registeredTask). It can improved after the whole feature is checked in and as part of token removal PR -> #1339.

danielayala94: It should be: `if (Protocol) else { if(BI){} if(COM){} }`. We can't do an if-else on BI and then COM or vice versa.

Right now, returning a token is a problem for simplifying blocks if we are supposed to return in just one place. It can be addressed as part of the token removal PR -> #1339

dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
danielayala94: `true`

dev/PushNotifications/PushNotificationManager.cpp
hulumane: This actually sounds reasonable: if(Protocol){} else { if(BI){} if(COM){} }

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
danielayala94: Good point

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
hulumane: Sounds like a plan. We shouldn't use processName anywhere but at the entry points of LRP Platform. Always convert to appId and use that as the currency in internal components.

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
danielayala94: It should only called during Initialize(), which already acquired a lock. I added a comment explaining this.

dev/PushNotifications/PushNotificationsLongRunningTask/platform.h
danielayala94: I should be more attentive to such details.

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -46,60 +54,133 @@ void NotificationsLongRunningPlatformImpl::WaitForLifetimeEvent()
...
+        {
+            winrt::hstring processName{ winrt::unbox_value<winrt::hstring>(values.Lookup(appIds[i])) };
+            mapOfFullTrustApps.emplace(reinterpret_cast<PWSTR>(appIds[i]), processName.c_str());
+        }
+    }
+
+    return mapOfFullTrustApps;
+}
+
+// Code directly pasted from Sharath's PR, adjusted to return std::wstring
```
danielayala94, line 158: Okay Sharath's work is merged. We can remove this.

loneursid (approved):

PR #1364, opened 2021-09-02
Workaround UDK include issue

loneursid:
There is a bug in the FrameworkUDK resulting in a file not found during the build. The file that can't be found is PushNotificationsRT.h. The fix is simple: replace angle brackets with double quotes and it has been posted to the correct rs_xxx branches. Still, the fix hasn't been ingested into the WindowAppSDK.

This work around addresses the build failure and allows the Unpackaged Push Notifications feature (which depends on the FrameworkUDK) to compile and work properly. It does that by placing one copy of the file inside the project folder where it can be easily located by the  build tools. 

This is only a temporary solution to allow the feature to ship with WindowsAppSDK v1.0 and the added files will have to be removed once the FrameworkUDK has been updated.

The file contents are identical across all three platforms: amd64, arm64 and x86, so keeping a single copy of it works accross al platforms.

Other options considered but rejected:
1) Adding the path to PushNotificationsRT.h inside FrameworkUDK package to the project settings. This works, but it means version number of the package get hardcoded in the project file and the build will break as soon as the FrameworkUDK gets updated.
2) Using a script to copy the file from the package to a location where the build can easily access it. It has the same issue has the option above, though the script could be made to fail silently if the file can't found. The build itself wouldn't break then, because the next version of the FrameworkUDK will include a fix that will ensure PushNotificationsRT.h is found inside the package.


danielayala94 (approved):

dev/PushNotifications/PushNotificationsRT.h
```diff
@@ -0,0 +1,299 @@
+/* Header file automatically generated from pushnotificationsrt.idl */
+/*
+ * File built with Microsoft(R) MIDLRT Compiler Engine Version 10.00.0231 
+ */
+
+#pragma warning( disable: 4049 )  /* more than 64k source lines */
```
hulumane, line 6: I'm assuming that the header isn't getting generated for some reason from the udk? Do we know when the udk fix is coming?

hulumane (approved):

dev/PushNotifications/PushNotificationsRT.h
loneursid: No. It's completely out of our hands at this point. It's guarantee to hapen before Reunion v1.0 launches though. 

The fix has been checked-in into the rs_wdx_dxp_ixp1 branch and cherry-picked into ni_release_svc_reunion21110. From there the UDK must be taken and ingested into Project reunion.

PR #1366, opened 2021-09-03
Ensure order of WARfwk install to avoid a Deployment issue

DrusTheAxe:
Ensure order of WARfwk is arm64 > x64 > x86 to avoid a bug in appxsvc erroneously using e.g. x86 on an x64 machine if there's only WARfwk x86 and not x64

The installer has an ordered list to process packages - [fwk, main, singleton, ddlm] - and for each ordered as [x86, x64, arm64]. If you installed x86 first the appxsvc would then try to load the Environment dll and fail on x64(+arm64) systems due to a bug ("I see no x64 packages in this family but I do see an x86, I'll load that..."). This was purely if there's no native-arch-matching package in the family; long as WARfwk-x64 is on the machine the Environment dll would be loaded by the x64 appxsvc process, even if there's a newer version of other architectures (like I said, a bug).

But to workaround that in the short-term we can reorder the installer's list of packages so 'the most applicable supported architecture' for a system is always processed before any other package in the family is considered, i.e. order the packages [arm64, x64, x86] and e.g. Deployment on an arm64 system will find the arm64 package before x64|x86, and likewise on an x64 system will find the x64 package before x86.

This bug is unique to Win11/Cobalt builds with new plumbing used by the Environment dll. It's not an issue on older systems, and soon to be a non-issue even on Win11/Cobalt once that's fixed.

### DIFF
--- a/installer/dev/packages.h
+++ b/installer/dev/packages.h
@@ -36,34 +36,34 @@ namespace WindowsAppRuntimeInstaller {
         WAR_FRAMEWORK_X86_LISTENTRY
     #endif
 
-    #if defined(WAR_MAIN_X86_LISTENTRY)
-        WAR_MAIN_X86_LISTENTRY
+    #if defined(WAR_MAIN_ARM64_LISTENTRY)
+        WAR_MAIN_ARM64_LISTENTRY
     #endif
     #if defined(WAR_MAIN_X64_LISTENTRY)
         WAR_MAIN_X64_LISTENTRY
     #endif
-    #if defined(WAR_MAIN_ARM64_LISTENTRY)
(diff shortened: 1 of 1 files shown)


jefgen (approved):
Thanks @DrusTheAxe !

riverar:
Do you have a rough affected build range?

DrusTheAxe:
> Do you have a rough affected build range?

Of Windows or WinAppSDK ?

riverar:
@DrusTheAxe Was just looking for which cobalt builds you were referring to re:

> This bug is unique to Win11/Cobalt builds with new plumbing used by the Environment dll. It's not an issue on older systems, and soon to be a non-issue even on Win11/Cobalt once that's fixed.

PR #1377, opened 2021-09-07
Remove bootstrap.targets

huichen123:
bootstrap.targets is no longer needed. The copy is handled by nuget. Its existence only causes AnyCPU build to fail.


DrusTheAxe:
> bootstrap.targets is no longer needed. The copy is handled by nuget. Its existence only causes AnyCPU build to fail.

I don't think we're there yet. I'd taken a pass at it but the 'fix' caused more harm than help and reverted. I've got a plan to make another run at the problem, but it's a timesuck testing the various permutations so I haven't gotten through it yet; burning through higher pri fires first.

DrusTheAxe (changes requested):
This presumes the change I previously made wasn't reverted?

DrusTheAxe:
FYI the fix was add the dll to the net5... directory, remove the copy as VS handles it AND remove the dll from the runtime dirs. Having it in both lib+runtime confused VS to no end. But finally sorted out.

This PR's now obsolete and can be closed w/o checkin

PR #1385, opened 2021-09-08
Telemetry Update for Unpackaged Apps

loneursid:
Log the processName as an alternative to the package name as the appId for unpackaged apps.


dev/PushNotifications/PushNotificationTelemetry.h
```diff
@@ -116,21 +118,61 @@ class PushNotificationTelemetry : public wil::TraceLoggingProvider
         return m_eventCount;
     }
 
-    wchar_t m_appUserModelId[APPLICATION_USER_MODEL_ID_MAX_LENGTH] = {};
+    inline bool IsPackagedApp()
+    {
+        static bool isPackagedApp = AppModel::Identity::IsPackagedProcess();
```
loneursid, line 123: Is this a valid approach in COM /Winrt? 

It's valid and thread safe C++ code but I'm concerned that it may not play well with Com/WinRT.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -167,10 +169,7 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
 
                             pushChannelReceived = co_await channelManager.CreatePushNotificationChannelForApplicationAsync();
 
-                            PushNotificationTelemetry::ChannelRequestedByApi(
-                                S_OK,
-                                AppModel::Identity::IsPackagedProcess(),
-                                remoteId);
+                            PushNotificationTelemetry::ChannelRequestedByApi(S_OK, remoteId);
```
sharath2727, line 172: when going by the legacy approach - remoteId doesn't make sense. Can you distinguish between legacy and newer approach?

dev/PushNotifications/PushNotificationTelemetry.h
```diff
@@ -116,21 +118,61 @@ class PushNotificationTelemetry : public wil::TraceLoggingProvider
...
+
+    std::wstring InitAppIdUnpackaged() noexcept
+    {
+        std::wstring appId;
+
+        wil::unique_cotaskmem_string processName;
+        auto result = wil::GetModuleFileNameExW(GetCurrentProcess(), nullptr, processName);
+        if (result == ERROR_SUCCESS)
+        {
+            appId = CensorFilePath(processName.get());
```
sharath2727, line 168: why are we storing the processname as appId? I think we need to call into the LRP and access the storage and get the appId right?

danielayala94: It is just a field name. The GUID doesn't mean anything if not tied to a process name anyway.

dev/PushNotifications/PushNotificationTelemetry.h
```diff
@@ -116,21 +118,61 @@ class PushNotificationTelemetry : public wil::TraceLoggingProvider
...
+        static bool isPackagedApp = AppModel::Identity::IsPackagedProcess();
+
+        return isPackagedApp;
+    }
+
+    inline const wchar_t* GetAppId()
+    {
+        static std::wstring appId = IsPackagedApp() ? InitAppIdPackaged() : InitAppIdUnpackaged();
+
+        return appId.c_str();
```
danielayala94, line 132: Isn't this a bug? c_str() is scoped to appId lifetime.

dev/PushNotifications/PushNotificationTelemetry.h
```diff
@@ -116,21 +118,61 @@ class PushNotificationTelemetry : public wil::TraceLoggingProvider
...
+    {
+        if (path)
+        {
+            path = !PathIsFileSpecW(path) ? PathFindFileNameW(path) : path;
+        }
+
+        return path;
+    }
+
+    std::wstring InitAppIdUnpackaged() noexcept
```
danielayala94, line 160: This function can throw while it's marked as noexcept.

danielayala94: You need a try-CATCH_LOG pattern here.

dev/PushNotifications/PushNotificationTelemetry.h
loneursid: appId is a local static

dev/PushNotifications/PushNotificationTelemetry.h
loneursid: appId is not the best name choice; I'll replace it with appName. I was looking for an abstraction that would be the PFN for packaged app and the process name for unpackaged. AppId seemed like a good idea at the time but it's already way to overloaded.

loneursid: Changed to appName

dev/PushNotifications/PushNotificationTelemetry.h
loneursid: good catch. Remove the noexcept as higher level functions in the class are already wrapped-up in try / catch

dev/PushNotifications/PushNotificationManager.cpp
loneursid: yet, CreateChannelAsync was called with a remoteId (even if it'a a null one) and this is really all this is logging.

loneursid: What are the options we want to log beyond "legacy"?

dev/PushNotifications/PushNotificationTelemetry.h
```diff
@@ -31,8 +32,9 @@ class PushNotificationTelemetry : public wil::TraceLoggingProvider
                 _GENERIC_PARTB_FIELDS_ENABLED,
                 TraceLoggingHexUInt32(hr, "OperationResult"),
                 TraceLoggingWideString(to_hstring(remoteId).data(), "RemoteId"),
```
danielayala94, line 34: Nit: Is `TraceLoggingGuid` supported in WAS?

danielayala94, line 34: Nit: Is TraceLoggingGuid supported in WAS?

danielayala94 (approved):

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -118,6 +118,8 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
 
     winrt::IAsyncOperationWithProgress<winrt::Microsoft::Windows::PushNotifications::PushNotificationCreateChannelResult, winrt::Microsoft::Windows::PushNotifications::PushNotificationCreateChannelStatus> PushNotificationManager::CreateChannelAsync(const winrt::guid &remoteId)
     {
+        bool usingLegacyImpl = false;
```
DrusTheAxe, line 121: NIT: `bool usingLegacyImpl{};`

DrusTheAxe, line 121: NIT: `usingLegacyImplementation`

Avoid abbreviations

dev/PushNotifications/PushNotificationTelemetry.h
```diff
@@ -31,9 +31,10 @@ class PushNotificationTelemetry : public wil::TraceLoggingProvider
                 TelemetryPrivacyDataTag(PDT_ProductAndServicePerformance),
                 _GENERIC_PARTB_FIELDS_ENABLED,
                 TraceLoggingHexUInt32(hr, "OperationResult"),
-                TraceLoggingBool(isAppPackaged, "IsAppPackaged"),
                 TraceLoggingWideString(to_hstring(remoteId).data(), "RemoteId"),
-                TraceLoggingWideString(GetAppUserModelId(), "AppUserModelId"));
+                TraceLoggingBool(usingLegacyImpl, "UsingLegacyImpl"),
+                TraceLoggingBool(IsPackagedApp(), "IsAppPackaged"),
```
DrusTheAxe, line 36: Isn't 'App' redundant here?

SUGGESTION: IsPackaged

dev/PushNotifications/PushNotificationTelemetry.h
```diff
@@ -116,21 +120,61 @@ class PushNotificationTelemetry : public wil::TraceLoggingProvider
...
+        static const std::wstring appName = IsPackagedApp() ? GetAppNamePackaged() : GetAppNameUnpackaged();
+
+        return appName.c_str();
+    }
 
-    wchar_t* GetAppUserModelId()
+    std::wstring GetAppNamePackaged() noexcept
     {
-        if (m_appUserModelId[0] == '\0')
+        wchar_t appUserModelId[APPLICATION_USER_MODEL_ID_MAX_LENGTH] = {};
```
DrusTheAxe, line 139: NIT: `wchar_t app[BLAH]{};`

dev/PushNotifications/PushNotificationTelemetry.h
```diff
@@ -116,21 +120,61 @@ class PushNotificationTelemetry : public wil::TraceLoggingProvider
...
+        if (result != ERROR_SUCCESS)
+        {
+            wcscpy_s(appUserModelId, L"AppUserModelId not found");
+            LOG_WIN32(result);
+        }
+
+        return appUserModelId;
+    }
+
+    PCWSTR CensorFilePath(_In_opt_ PCWSTR path) noexcept
```
DrusTheAxe, line 152: NIT: Remove `_In_opt_`

InOpt is redundant with PCWSTR

DrusTheAxe (approved):
LGTM modulo minor comments

dev/PushNotifications/PushNotificationTelemetry.h
loneursid: done

dev/PushNotifications/PushNotificationManager.cpp
loneursid: done

loneursid: done

dev/PushNotifications/PushNotificationTelemetry.h
loneursid: I kept the IsAppPackaged, to match the corresponding GetAppName.

dev/PushNotifications/PushNotificationTelemetry.h
loneursid: done

dev/PushNotifications/PushNotificationTelemetry.h
loneursid: done

PR #1386, opened 2021-09-08
Update FrameworkUDK

kythant:
Update FrameworkUDK


dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/src/Microsoft.Windows.ApplicationModel.Resources.vcxproj
```diff
@@ -174,7 +174,7 @@
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets">
-    <Import Project="..\..\packages\Microsoft.FrameworkUdk.$(Udk-Arch)fre.10.0.22431.1002-210816-1731.ni-release-svc-reunion2110\build\native\Microsoft.FrameworkUdk.$(Udk-Arch)fre.targets" Condition="Exists('..\..\packages\Microsoft.FrameworkUdk.$(Udk-Arch)fre.10.0.22431.1002-210816-1731.ni-release-svc-reunion2110\build\native\Microsoft.FrameworkUdk.$(Udk-Arch)fre.targets')" />
```
BenJKuhn, line 177: Please consolidate the dependencies into a common props file, since this is starting to be a widely duplicated bit of information.

DrusTheAxe:
You missed one. When changing dependencies (FrameworkUdk or anything else) please update the entire Foundation repo (this PR only hits 80%)

```
[C:]findstr /m /s /i frameworkudk packages.config *.vcxproj
dev\MRTCore\mrt\Microsoft.ApplicationModel.Resources\src\Microsoft.ApplicationModel.Resources.vcxproj
dev\MRTCore\mrt\Microsoft.ApplicationModel.Resources\src\packages.config
dev\WindowsAppRuntime_DLL\packages.config
dev\WindowsAppRuntime_DLL\WindowsAppRuntime_DLL.vcxproj
test\PowerNotifications\PowerNotifications.vcxproj
```

DrusTheAxe:
Gah! WindowsAppRuntime_DLL.vcxproj doesn't specify a version (just DelayLoadDLL, which is needed specifically in this one spot not a general thing).

But general sentiment stands - dependency changes need to be repo wide. Consider this further ammunition for @BenJKuhn's 'make a common props file' request :-)

kythant:
Yea I am thinking about reusing the version.props that we have. Unfortunately package.config won't be able to read into those.

DrusTheAxe:
> Yea I am thinking about reusing the version.props that we have. Unfortunately package.config won't be able to read into those.

version.props isn't the right answer? We shouldn't force FrameworkUdk (+WIL, +...) upon every project in the repo. Better to be opt-in so projects that don't fit the norm can deviate e.g. I very much do NOT want CppWinRT included in certain projects as it's not a seamlessly positive-only experience.

package.config is a problem. Perhaps a `tools\CheckDependencies.cmd`/ps1 that scans everything in the repo and reports which uses what. For instance, by default `-Level Warning` would flag anything inconsistent or problematic, but `-Level All` would enumerate everything for a singular master list. Make it easy to see which projects use what, as well as flag potential issues (I assume inconsistencies are bad as a general rule, though we'll undoubtedly have exceptions at some point for good reasons).

dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/src/Microsoft.Windows.ApplicationModel.Resources.vcxproj
BenJKuhn: per discussion, this is more complex than it sounds, and may not be exactly what we want. So we can defer this improvement to a later PR / change.

BenJKuhn (approved):

huichen123 (approved):
:shipit:

PR #1388, opened 2021-09-08
Microsoft.WindowsAppRuntime.Bootstrap.Net.dll not in Transport package

DrusTheAxe:
Microsoft.WindowsAppRuntime.Bootstrap.Net.dll was copied to the Transport's staging area not then not actually copied into the Transport package.

This may not be the right end directory in the NuGet but getting it into the NuGet unblocks other folks (docs, samples, etc). Doing testing but won't be finished as quickly as this will be available so unblock ASAP, and if it needs to move to other location(s) I'll relocate it via a follow up PR

[Bug# 35891850](https://task.ms/35891850)

### DIFF
--- a/build/CopyFilesToStagingDir.ps1
+++ b/build/CopyFilesToStagingDir.ps1
@@ -110,7 +110,7 @@ PublishFile $FullBuildOutput\WindowsAppRuntime_DLL\Microsoft.Windows.System.Powe
 # Native (not managed, AppLocal / no MSIX)
 PublishFile $FullBuildOutput\WindowsAppRuntime_BootstrapDLL\Microsoft.WindowsAppRuntime.Bootstrap.dll $NugetDir\runtimes\lib\native\$Platform
 PublishFile $FullBuildOutput\WindowsAppRuntime_BootstrapDLL\Microsoft.WindowsAppRuntime.Bootstrap.pdb $NugetDir\runtimes\lib\native\$Platform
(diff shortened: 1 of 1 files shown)


build/CopyFilesToStagingDir.ps1
```diff
@@ -110,6 +110,7 @@ PublishFile $FullBuildOutput\WindowsAppRuntime_DLL\Microsoft.Windows.System.Powe
 # Native (not managed, AppLocal / no MSIX)
 PublishFile $FullBuildOutput\WindowsAppRuntime_BootstrapDLL\Microsoft.WindowsAppRuntime.Bootstrap.dll $NugetDir\runtimes\lib\native\$Platform
 PublishFile $FullBuildOutput\WindowsAppRuntime_BootstrapDLL\Microsoft.WindowsAppRuntime.Bootstrap.pdb $NugetDir\runtimes\lib\native\$Platform
+PublishFile $FullBuildOutput\Microsoft.WindowsAppRuntime.Bootstrap.Net\Microsoft.WindowsAppRuntime.Bootstrap.Net.dll $NugetDir\runtimes\lib\native\$Platform
```
huichen123, line 113: >native [](http://example.com/codeflow?start=140&length=6)

why .net dll in native folder?

DrusTheAxe: It put it next to the current Bootstrap. dll for now. I suspect that's the wrong place, but not sure of the right one - yet. Easy enough to move once I do some digging, but I didn't want to block others (docs etc) while I do

kythant (approved):

PR #771, opened 2021-04-27
In C#/.NET projects, pick up image and resw files for PRI generation.

rohanp-msft:
This wasn't done in all cases. See the note in this page https://docs.microsoft.com/en-us/windows/apps/project-reunion/mrtcore/mrtcore-overview:
"
Before you can use MRT Core to retrieve strings and images in a WinUI 3 project that uses C#/.NET 5, you must ensure that these resources are configured so that they can be indexed in the resources.pri file. Otherwise, these resources cannot be retrieved by MRT Core.

Strings: Make sure the Build Action property for resources files (.resw) is set to PRIResource. This property is set automatically if you add a new Resources File (.resw) item to your project.
Images: Make sure the Build Action property for image files are set to Content. This property is set automatically if you add an existing image to a folder named Assets in your project.
"

Addresses: #613 

**How verified**
With the .NET/C# MRT Core sample app, added a PNG, a BMP, a JPG and a RESW file. After building the app, the resources do show up in the PRI as expected.

PR #1389, opened 2021-09-09
Fix default file inclusion of .RESW and image files in .NET projects

### LINKED ISSUE #980: Adding an icon to a package with build action None still adds it to the .pri resource
**Describe the bug**
When building a class library with the ability to pack into a nuget package, an included icon for the nuget package will be referenced by the .pri file.
This results is a nuget package that fails to deploy, due to the lack of this file in the lib folder, but it could also mean that resources are getting deployed that the developer never explicitly told it to include.

**Steps to reproduce the bug**

1. Create a new WinUI class library (C#):
2. Add an icon in the folder above, and include it in the project for nuget icons. Project should look like this:
```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFramework>net5.0-windows10.0.19041.0</TargetFramework>
    <TargetPlatformMinVersion>10.0.17763.0</TargetPlatformMinVersion>
    <RootNamespace>WinUIClassLib</RootNamespace>
    <RuntimeIdentifiers>win10-x86;win10-x64;win10-arm64</RuntimeIdentifiers>
    <UseWinUI>true</UseWinUI>
    <PackageIcon>icon.png</PackageIcon> <!-- add this -->
  </PropertyGroup>

  <ItemGroup>
    <None Include="../icon.png" Link="icon.png" Pack="true" PackagePath="\icon.png" /> <!-- add this -->
    <PackageReference Include="Microsoft.ProjectReunion" Version="0.8.0" />
    <PackageReference Include="Microsoft.ProjectReunion.Foundation" Version="0.8.0" />
    <PackageReference Include="Microsoft.ProjectReunion.WinUI" Version="0.8.0" />
  </ItemGroup>
</Project>
```
3. Right-click the project and select "Pack"
4. Inspect the .pri file that is generated and placed in the nuget package. It'll contain a reference to this icon file. But since it isn't packed with the package below the lib folder, it'll fail to deploy this package.

Full repro:
[WinUIClassLib.zip](https://github.com/microsoft/WindowsAppSDK/files/6712327/WinUIClassLib.zip)

I did find this in `Microsoft.ApplicationModel.Resources.PriGen.targets` but to me it seems incorrect to assume `None` files should be included:
```xml
  <!-- The .NET/C# project system does not add the .resw files added to the project to PRIResource, unlike the UWP
       project system (this is expected since .resw files were historically UWP specific). That is needed for the
       .resw files to be indexed during PRI generation. Likewise, image files need to be added to Content for them
       to be indexed. These files exist in the None list by default (meaning without app developer intervention).
       We address this requirement here so that the app developer doesn't need to manually intervene. -->
  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp'">
    <!-- Add .resw files to PRIResource so they are indexed. -->
    <PRIResource Include="@(None->WithMetadataValue('Extension', '.resw'))" />
    <!-- Add image files to PRIResource so they are indexed. -->
    <Content Include="@(None->WithMetadataValue('Extension', '.png'))" />
    <Content Include="@(None->WithMetadataValue('Extension', '.bmp'))" />
    <Content Include="@(None->WithMetadataValue('Extension', '.jpg'))" />
    <Content Include="@(None->WithMetadataValue('Extension', '.dds'))" />
    <Content Include="@(None->WithMetadataValue('Extension', '.tif'))" />
    <Content Include="@(None->WithMetadataValue('Extension', '.tga'))" />
    <Content Include="@(None->WithMetadataValue('Extension', '.gif'))" />
  </ItemGroup>
```
Especially this comment:
> Likewise, image files need to be added to Content for them to be indexed. These files exist in the None list by default (meaning without app developer intervention). We address this requirement here so that the app developer doesn't need to manually intervene. 

IMHO if I chose not to include files (implicitly or explicitly), it shouldn't just be chosen for me. The NuGet packaging scenario is one obvious scenario, but there could just as well be resources here that I never intended to actually ship but just keep for other reasons. And even if I _explicitly_ set a build action to `None` (as above), it gets forced in anyway. This doesn't seem like the correct way to configure default items.
I believe the way to do default globbing/includes should be done in the SDK instead. See here for references: https://docs.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props-desktop

I did confirm by removing this entry from the Foundation nuget package, the problem goes away.

This seems to have been introduced by #771

There is also no simple workaround that I can see. Even doing this will cause it to be added regardless:
```xml
<ItemGroup>
    <Content Remove="icon.png" />
</ItemGroup>
```
You can however do this:
```
  <ItemGroup>
    <Content Remove="icon.png" />
    <None Remove="icon.png" />
  </ItemGroup>
```
But that is very counter-intuitive, and now I can't pack the icon with my nuget package.

**Expected behavior**
Don't include files with build action `None` into the resource file.

**Version Info**
VS16.10

NuGet package version: 
0.8.0

| Windows 10 version                  | Saw the problem? |
| :--------------------------------- | :-------------------- |
| Insider Build (xxxxx)              |    |
| May 2020 Update (19041)            |    |
| November 2019 Update (18363)       |    |
| May 2019 Update (18362)            |    |
| October 2018 Update (17763)        |    |

**Additional context**

rohanp-msft:
The existing solution (added in #771) has an issue: app devs cannot opt out an image or a RESW file from PRI inclusion.

Closes #980 

**How verified (commit 11)**

Setup:
1. Uninstall the existing VSIX.
2. Install WindowsAppSDK.Extension.Experimental.vsix from the output of the build pipeline run.
3. Create a new app using the "WinUI.Desktop.Cs.PackagedApp" template.
4. Install Microsoft.WindowsAppRuntime.1.0.msix manually.

Verification:
(BA below means Build Action)
1. Copy paste a RESW file, a .PNG file, and a .JPG file into the folder containing the .CSPROJ file. Build and run. Examine PRI. Examine bin placing.
2. Opt-out the RESW file and the PNG file. Build and run. Examine PRI. Examine bin placing.
3. Opt-in the files that were opted out. Build and run. Examine PRI. Examine bin placing.
4. Restart VS. Rebuild and run. Examine PRI.
5. Opt-out files. Restart VS. Build. Examine PRI.
6. Set `EnableDefaultPRIResourceItems` to false. BA of RESW file changed to `None`. Remove set. BA changes back to `PRIResource`.
7. Set `EnableDefaultContentItems` to false. BA of image file changed to `None`. Remove set. BA changes back to `Content`. (BA of image file that was `None` already remains None and is `None` after set is removed too.)
8. Set `EnableDefaultItems` to false. All image and RESW files disappeared in Solution Explorer (regardless of their BA). Remove set. All files come back.

**Note**:
- *Issue 1*: There is an issue I found for which I'll create a tracking item and link to this PR:
  - Let's say you create a new app. You build and run it. That generates a resources.pri.
  - You then copy-paste some resource files into the project folder. The Build Action of these files will be set correctly by default (which is the point of the change).
  - You then hit F5.
  - The resources.pri in the build output after the F5 Run doesn't have the new resources. I suspect the PRI is not regenerated.
  - If the app is rebuilt, then the PRI's contents are correct.
- *Issue 2*: This solution does not work for resource files that are added from outside the project folder. For them, manual action is needed to set the Build Action to `PRIResource` or `Content`. I'll create a tracking item for this too and link to this PR.


build/NuSpecs/Microsoft.WindowsAppSDK.Foundation.props
```diff
@@ -8,4 +8,14 @@
...
 
+  <!--
+      At least for now, we want MRT Core to not be used in WAPPROJ projects. There, we want the in-box PRI generation tooling to be used.
+      Likewise, we don't want this out-of-box PRI generation tooling to be used with UWPs.
+  -->
+  <PropertyGroup>
+    <EnableCoreMrtTooling Condition="'$(EnableCoreMrtTooling)'=='' and '$(WindowsAppContainer)'!='true' and '$(MSBuildProjectExtension)' != '.wapproj'">true</EnableCoreMrtTooling>
+  </PropertyGroup>
+
+  <Import Project="$(MSBuildThisFileDirectory)Microsoft.Windows.ApplicationModel.Resources.props" Condition="Exists('$(MSBuildThisFileDirectory)Microsoft.Windows.ApplicationModel.Resources.props') and '$(EnableCoreMrtTooling)'=='true'" />
```
rainersigwald, line 19: Nit: in the case of missing files that should always be in your NuGet package/installed/present, it's usually preferable to avoid the `Exists` condition and just let MSBuild error on the import. Otherwise you have to spend time debugging "the project loaded fine but didn't behave right" and eventually figure out that it was because a file was missing.

```suggestion
  <Import Project="$(MSBuildThisFileDirectory)Microsoft.Windows.ApplicationModel.Resources.props" Condition="'$(EnableCoreMrtTooling)'=='true'" />
```

rainersigwald (approved):

huichen123 (approved):
:shipit:

build/NuSpecs/Microsoft.WindowsAppSDK.Foundation.props
rohanp-msft: Done!

rohanp-msft:
@rainersigwald and @huichen123 
When you get some time, would you revisit the change please? I fixed a bug: `EnableDefaultContentItems` should only affect Content. I also added `EnableDefaultPRIResourceItems`, something that doesn't exist currently, so that app devs can switch off default inclusion of PRIResource items, just like they can for Content, Compile, etc.

Disclaimer: Verification of commit 6 is pending. There are no new changes that I think will cause the fix to stop working. I will still do the verification, of course, but I want to be sure that things like the introduction of `EnableDefaultPRIResourceItems` is not unacceptable (if it was, that'll make my time-expensive verification a waste).

dev/MRTCore/packaging/MrtCore.props
```diff
@@ -0,0 +1,32 @@
...
+    defined in a .NET SDK file that's imported after this one, but ItemGroup definitions are evaluated in a later pass.
+    See https://docs.microsoft.com/en-us/visualstudio/msbuild/build-process-overview?view=vs-2019#evaluation-phase  
+  -->
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)'=='.NETCoreApp' and '$(EnableDefaultItems)'=='true' and '$(EnableDefaultContentItems)'!='false'">
+    <None Remove="$(_ReunionImageGlobs)" />
+    <Content Include="$(_ReunionImageGlobs)" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
+  </ItemGroup>
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)'=='.NETCoreApp' and '$(EnableDefaultItems)'=='true' and '$(EnableDefaultPRIResourceItems)'!='false'">
+    <None Remove="**/*.resw" />
+    <PRIResource Include="**/*.resw" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
```
rainersigwald, line 29: I think this makes sense but it did give me pause for a second. Specifically, the idea that `$(EnableDefaultPRIResourceItems)` doesn't control whether the files get globbed, just what item they get globbed into, is a bit surprising. Is that preferable to always removing resw from None (when using your package) and only conditionally including it?

I am fine with either way but want to make sure you think about it explicitly.

rainersigwald (approved):
> I want to be sure that things like the introduction of `EnableDefaultPRIResourceItems` is not unacceptable

I think this is a very reasonable extension of the current pattern. With MSBuild there's always a risk of name collision but this is very specific and doesn't appear to be currently used.

@dsplaisted do you have any concerns from the SDK perspective?

build/NuSpecs/Microsoft.WindowsAppSDK.Foundation.props
```diff
@@ -8,4 +8,11 @@
     <WindowsAppSdkFoundation>true</WindowsAppSdkFoundation>
   </PropertyGroup>
 
+  <!-- At least for now, we want MRT Core to not be used in WAP or UWP projects. There, we want the in-box PRI generation tooling to be used. -->
+  <PropertyGroup>
+    <EnableCoreMrtTooling Condition="'$(EnableCoreMrtTooling)'=='' and '$(WindowsAppContainer)'!='true' and '$(MSBuildProjectExtension)'!='.wapproj'">true</EnableCoreMrtTooling>
```
dsplaisted, line 13: It seems like this condition won't entirely work now that it's moved to a .props file.  The `EnableCoreMrtTooling` and `WindowsAppContainer` properties wouldn't be set yet.

dsplaisted (approved):
The globbing looks good, but it seems like there may be issues with moving the `EnableCoreMrtTooling` logic to .props files.

A possible way to work around this would be to move the logic that sets `EnableCoreMrtTooling` back to the .targets, import `MrtCore.props` unconditionally, and then add a check for `EnableCoreMrtTooling` to all the ItemGroups in `MrtCore.props`.

dev/MRTCore/packaging/MrtCore.props
dsplaisted: I think leaving resw files out of `None` would more closely match what we now do in the .NET SDK.  If you turn `EnableDefaultCompileItems` off, the source files will not be part of `None`.  Originally the behavior was different, but we changed it to address [some bugs](https://github.com/dotnet/sdk/issues/1157) in Visual Studio behavior.

So I would lean towards removing `*.resw` from `None` regardless of whether `EnableDefaultPRIResourceItems` is set.  It's up to you, but if you want to go the other way I'd suggest testing what happens if you set the property to false and then add a resw file to the project in Visual Studio, both via the new item template as well as add / existing item.

build/NuSpecs/Microsoft.WindowsAppSDK.Foundation.props
rohanp-msft: Excellent find...

When I moved the property definition, that didn't occur to me. The import then worked in my verification, so I didn't question it. I see now that the reason it worked is that my conditions check if the properties are not equal to fixed strings. Empty strings will of course satisfy that.

Will fix!

I wish MSBuild logs captured where a property was assigned its value, not just what its value is. That would have helped in figuring out why this import was working.

dsplaisted: [MSBuild binary logs](https://msbuildlog.com) actually do capture the locations where properties are reassigned.  I'm not sure if that tells you where it was first set, but if it is changed then it will show you where.

image

rohanp-msft: That's good to know! I think your suspicion about the first assignment is right. In my case, the property was not reassigned a value and I didn't see the location of the assignment in the logs, just its value.

rainersigwald: > I wish MSBuild logs captured where a property was assigned its value, not just what its value is.

https://github.com/dotnet/msbuild/issues/2711. We need to figure out an efficient way to do it.

dev/MRTCore/packaging/MrtCore.props
rohanp-msft: Went with Daniel's suggestion to remove `*.resw` from `None` regardless of whether `EnableDefaultPRIResourceItems` is set.

dev/MRTCore/packaging/MrtCore.props
```diff
@@ -0,0 +1,37 @@
...
+    See https://docs.microsoft.com/en-us/dotnet/core/project-sdk/overview#default-includes-and-excludes.
+
+    We match the semantics of EnableDefaultCompileItems here: if EnableDefaultCompileItems is false, .CS files,
+    for example, are not added to the project, as opposed to added as None.
+
+    Yes, the properties used here, including EnableCoreMrtTooling, are defined in MSBuild files imported after
+    this file. This is correct though because ItemGroup definitions are evaluated in a later pass. See
+    https://docs.microsoft.com/en-us/visualstudio/msbuild/build-process-overview?view=vs-2019#evaluation-phase.
+  -->
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)'=='.NETCoreApp' and '$(EnableCoreMrtTooling)'=='true'">
```
huichen123, line 26: >'$(EnableCoreMrtTooling)'=='true'"> [](http://example.com/codeflow?start=74&length=35)

this is not true

rohanp-msft: Item groups are evaluated in a later pass than properties. Please see https://docs.microsoft.com/en-us/visualstudio/msbuild/build-process-overview?view=vs-2019#evaluation-phase

PR #1394, opened 2021-09-09
Rename MRTCore build filename

huichen123:
Microsoft.Windows.ApplicationModel.Resources.References.targets is the longest filename in the WindowsAppSDK package. When app path is a little deep, restoring WindowsAppSDK package may fail if any file path is longer than MAX_PATH. To not contributing to this, now shrink MRTCore targets/props filename (replace Microsoft.Windows.ApplicationModel.Resources with MrtCore).


rohanp-msft (approved):
:shipit:

rohanp-msft:
dev/MRTCore/packaging/README.md needs to be updated too

PR #1396, opened 2021-09-09
Shorten VSIX DisplayName

DefaultRyan:
The DisplayName part of a VSIX has a maximum length of 50 characters. An example of a current name is:
`Windows App SDK C++ VS 2022 Templates` (37 characters).

In experimental builds, we append ` (Experimental)` to get
`Windows App SDK C++ VS 2022 Templates (Experimental)` (52 characters).

This takes over the limit. In this change, I shorten the name by removing a few spaces, resulting in
`WindowsAppSDK C++ VS2022 Templates (Experimental)` (49 characters).


andrewleader:
How about `Windows App SDK C++ VS 2022 Experimental Templates`, which is exactly 50? 🙂

riverar:
Or alternatively, shorten the suffix to `Windows App SDK C++ VS 2022 Templates (Exp.)` [44].

DefaultRyan:
> Or alternatively, shorten the suffix to `Windows App SDK C++ VS 2022 Templates (Exp.)` [44]. Stepping back, how is any of this really going to work with loc? 🤔

That's a good question. The current MSBuild approach of labelling the experimental VSIX manifest is not going to scale for localization. I've filed a task to do some refactoring to enable this next month.

DefaultRyan:
Force committing to get some pipelines unblocked.

Issue #1287, opened 2021-08-20
RS5 Compatibility: GetCurrentPackageInfo2 not available through DynDep


huzaifa-d:
**Describe the bug**

APIs that use FrameworkUDK fail to be called on RS5 systems.

**Steps to reproduce the bug**

The CPPTestApp from ReunionInternal can be used to repro.
Changes in user/modanish/PowerNotif-IntegTest branch.
1. Build the solution and deploy the test on a RS5 VM.
2. Run the test using te.exe
3. Optionally, run the test app and press the "Power Supply Status" button.

**Expected behavior**

API call expected to succeed.

**Screenshots**

![image](https://user-images.githubusercontent.com/16077119/130300766-1b90d819-fdec-4de1-852a-011954d89da6.png)

**Version Info**

NuGet package version: 

1.0.0-20210819.1-CI

RS5_Release_SVC_2110c

**Additional context**

asklar:
@JesseCol

JesseCol:
I don't think this is a FrameworkUdk issue -- as far as I can tell, the GetCurrentPackageInfo2 simply doesn't exist in RS5.

huzaifa-d:
I agree. The only thing is that my APIs have a reliance on FUDK so I didn't know how to title the bug better. Other AppSDK APIs seem to be running fine.

asklar:
thanks - so this isn't a fudk issue but either the app/test needs to be updated, or the API needs to be polyfilled on rs5

huzaifa-d:
Agreed.

DrusTheAxe:
The problem is `GetCurrentPackageInfo2` doesn't exist in RS5. Dynamic Dependencies needs to use it when available and thus play nice on all releases down to RS5.

Working on it...

DrusTheAxe:
> The problem is `GetCurrentPackageInfo2` doesn't exist in RS5

There's a 2nd problem: `GetPackageInfo2` doesn't exist in RS5 either (shocking, I know :P).

I've got a branch with the former addressed. The latter is a little more involved due to internal implementation reasons, threading the code through several layers to address the problem is a little complex (in the time-consuming sense, not rocket-science). Workingh on it.

PR #1397, opened 2021-09-10
Remove import references to Get[Current]PackageInfo2 causing load failures on RS5

DrusTheAxe:
This fixes [RS5 Compatibility: GetCurrentPackageInfo2 not available through DynDep #1287](https://github.com/microsoft/WindowsAppSDK/issues/1287)

Get[Current]PackageInfo2 was added in 19H1 so import references cause load failures on RS5.

Reshuffled code to only call Get[Current]PackageInfo2 if available (19H1+) and fallback to Get[Current]PackageInfo (v1) on earlier systems. Get[Current]PackageInfo2 access is now done dynamically (LoadLibrary/GetrProcAddress) when necessary to avoid load-time resolution failures.

Verified import references were here yesterday and gone today by comparing `LINK /DUMP /IMPORTS Microsoft.WindowsAppRuntime.dll`

Also fixes most warnings in WindowsAppSDK.sln (too much noise, masking real potential issues).


dev/DynamicDependency/API/MddDetourPackageGraph.cpp
```diff
@@ -120,18 +142,35 @@ HRESULT WINAPI MddDetourPackageGraphInitialize() noexcept
     // Detour package graph APIs to our implementation
     FAIL_FAST_IF_WIN32_ERROR(DetourUpdateThread(GetCurrentThread()));
     FAIL_FAST_IF_WIN32_ERROR(DetourAttach(&(PVOID&)TrueGetCurrentPackageInfo, DynamicGetCurrentPackageInfo));
-    FAIL_FAST_IF_WIN32_ERROR(DetourAttach(&(PVOID&)TrueGetCurrentPackageInfo2, DynamicGetCurrentPackageInfo2));
-    if (IsWindows10_20H1OrGreater())
+    //
+    // NOTE: GetCurrentPackageInfo2 requires >=19H1
+    // NOTE: GetCurrentPackageInfo3 requires >=20H1
```
riverar, line 147: Curious: Why the sliding API window here versus just using lowest common denominator (GetCurrentPackageInfo)? Seems like it would simplify the code below.

test/EnvironmentManagerTests/EnvironmentManagerCentennialTests.cpp
```diff
@@ -104,7 +104,7 @@ namespace WindowsAppSDKEnvironmentManagerTests
         // Remove the value
         // setting the value to empty is the same as deleting the variable
         VERIFY_NO_THROW(environmentManager.SetEnvironmentVariable(c_EvKeyName, L""));
-        VERIFY_ARE_EQUAL(0, ::GetEnvironmentVariable(c_EvKeyName, nullptr, 0));
+        VERIFY_ARE_EQUAL(0u, ::GetEnvironmentVariable(c_EvKeyName, nullptr, 0));
```
riverar, line 107: Nit: Changes don't seem relevant to the GetPackageInfo fix here. Ditto for similar below.

riverar (approved):

dev/DynamicDependency/API/MddDetourPackageGraph.cpp
```diff
@@ -120,18 +142,35 @@ HRESULT WINAPI MddDetourPackageGraphInitialize() noexcept
...
-        TrueGetCurrentPackageInfo3 = dllGetCurrentPackageInfo3;
+        // Also grab GetPackageInfo2 if available (see MddGetPackageInfo1Or2() for details)
+        auto dllGetPackageInfo2{ reinterpret_cast<GetPackageInfo2Function>(GetProcAddress(dllApisetAppmodelRuntime_1_3, "GetPackageInfo2")) };
+        FAIL_FAST_HR_IF_NULL(HRESULT_FROM_WIN32(GetLastError()), dllGetPackageInfo2);
+        TrueGetPackageInfo2 = dllGetPackageInfo2;
 
-        FAIL_FAST_IF_WIN32_ERROR(DetourAttach(&(PVOID&)TrueGetCurrentPackageInfo3, DynamicGetCurrentPackageInfo3));
+        if (IsWindows10_20H1OrGreater())
+        {
+            HMODULE dllApisetAppmodelRuntimeInternal_1_6{ LoadLibraryExW(L"api-ms-win-appmodel-runtime-internal-l1-1-6.dll", nullptr, 0) };
```
riverar, line 166: Note: Does this work on Windows 11 22000+ where apiset is `api-ms-win-appmodel-runtime-internal-l1-1-9.dll`?

test/EnvironmentManagerTests/EnvironmentManagerCentennialTests.cpp
DrusTheAxe: The build was getting messy producing lots of warnings causing me some angst to overlook a problem at first. So I took a pass cleaning most of the warnings (simple things that cut a lot of noise), so I could clearly verify my fix.

dev/DynamicDependency/API/MddDetourPackageGraph.cpp
DrusTheAxe: Yes

Naming scheme is blahblah-l1-major-minor, so that's version 1.9 of that APIset. APIsets follow a strong breaking-change-requires-major-change policy so if 1-9 is present so too is 1-0 thru 1-8. 1-9 _adds_ symbols over 1-8; it's generally problematic to delete or change existing symbols (not impossible but not usually done, for obvious functional reasons as well as technical). When that needs to happen it's time to make 2-0.

That doesn't mean everything in 1-9 is functional (if it's internal). If, say, 1-5 added DoFoo(), DoFoo wasn't public and it's not obsolete (e.g. everything that used it doesn't anymore let alone if it's problematic now) then we can't remove DoFoo but we can gut the implementation e.g. `{ return E_NOTIMPL; }`.

riverar: Cool, just wanted to double check apisets we're versioned as such. Wasn't sure. Lgtm.

dev/DynamicDependency/API/MddDetourPackageGraph.cpp
```diff
@@ -109,6 +125,12 @@ VERSIONHELPERAPI IsWindowsVersionOrGreaterEx(
     return VerifyVersionInfoW(&osvi, VER_MAJORVERSION | VER_MINORVERSION | VER_SERVICEPACKMAJOR | VER_BUILDNUMBER, c_conditionMask) != FALSE;
 }
 
+VERSIONHELPERAPI IsWindows10_19H1OrGreater()
```
dhoehna, line 128: Shouldn't the code look for a feature and not a version?

dhoehna:
For the Environment Manager tests, all look good.

dev/DynamicDependency/API/MddDetourPackageGraph.cpp
sylveon: There isn't a feature check corresponding to this API. It was introduced in 19H1, and the docs say it as such.

riverar: I think the suggestion is to GetProcAddress poke around and delete all the version checks.

DrusTheAxe: Yes, but the lineage of some of this isn't as clearcut as it would appear. I'll make a future PR replacing the version checks if/where possible.

dhoehna (approved):

DrusTheAxe:
Weird build failure. Something wrong with the build infrastructure? @BenJKuhn @kythant @alexlamtest @kmahone

PR #1399, opened 2021-09-10
Implement GetRuntimeClassName correctly to support C#/WinRT downcasting of RCW

Scottj1s:
Without this change, C#/WinRT cannot project the interface correctly.  
See also:
https://github.com/microsoft/CsWinRT/issues/966

### DIFF
--- a/dev/AppLifecycle/LaunchActivatedEventArgs.h
+++ b/dev/AppLifecycle/LaunchActivatedEventArgs.h
@@ -19,6 +19,11 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
         {
             m_kind = ActivationKind::Launch;
         }
+        
+        winrt::hstring GetRuntimeClassName() const noexcept override
+        {
+            return winrt::hstring(L"Windows::ApplicationModel::Activation.LaunchActivatedEventArgs"sv);
+        }
 
         static winrt::Windows::Foundation::IInspectable Deserialize(winrt::Windows::Foundation::Uri const& uri)
         {


Scottj1s:
@aeloros can you drive this PR?  i.e., build and test to confirm?

dev/AppLifecycle/LaunchActivatedEventArgs.h
```diff
@@ -19,6 +19,11 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
         {
             m_kind = ActivationKind::Launch;
         }
+        
+        winrt::hstring GetRuntimeClassName() const noexcept override
+        {
+            return winrt::hstring(L"Windows::ApplicationModel::Activation.LaunchActivatedEventArgs"sv);
```
kennykerr, line 25: These should be dots.
```suggestion
            return winrt::hstring(L"Windows.ApplicationModel.Activation.LaunchActivatedEventArgs"sv);
```

DrusTheAxe: Aren't there constants generated by MIDL for ACIDs?

DrusTheAxe (approved):

jonwis:
Seems OK, but why does the type need to do this at all? Should the type be defined as:

```c++
    struct LaunchActivatedEventArgs : public LaunchActivatedEventArgsT<LaunchActivatedEventArgs, IInternalValueMarshalable>, 
        ActivatedEventArgsBase
    {
```

... which would use the C++/WinRT RTC machinery?

kennykerr:
> ... which would use the C++/WinRT RTC machinery?

It's pretending to be something it's not. If it were really the runtime class in question, then yes C++/WinRT would wire this up automatically.

Scottj1s:
Abandoning PR, per @aeloros

PR #1403, opened 2021-09-11
TerminalVelocity docs + infrastructure support for PushNotifications

DrusTheAxe:
Documented how to use TerminalVelocity

This includes basic infrastructure for PushNotifications velocity support. Checks are defined but not implemented.

@pmpurifoy you'll want to grab this and add checks in the appropriate places in PushNotifications' implementation to NOOP the behavior. See the docs for more details. We can check this in and you can make a PR with PushNotifications' use to NOOP implemnetation, or you can grab the branch and party on it so we ultimately check this in as a joint effort, or you can grab the bits and make your own branch and PR. Your call. How would you like to proceed?


dev/Common/TerminalVelocityFeatures-PushNotifications.h
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+// THIS FILE IS AUTOMATICALLY GENERATED; DO NOT EDIT IT
+
+// INPUT FILE: dev\common\TerminalVelocityFeatures-DeploymentAPI.xml
+// OPTIONS: -Channel Experimental -Language C++ -Namespace Microsoft.Windows.PushNotifications
```
DHowett, line 7: FWIW: Terminal does not check these files in because they're generated by an always-run build step, even in local VS builds. Is it expected that this file will stay in sync when the person holding the repo changes which channel they're building for?

DHowett (dismissed):
Open question about the checked-in file 😁

pmpurifoy:
> Documented how to use TerminalVelocity
> 
> This includes basic infrastructure for PushNotifications velocity support. Checks are defined but not implemented.
> 
> @pmpurifoy you'll want to grab this and add checks in the appropriate places in PushNotifications' implementation to NOOP the behavior. See the docs for more details. We can check this in and you can make a PR with PushNotifications' use to NOOP implemnetation, or you can grab the branch and party on it so we ultimately check this in as a joint effort, or you can grab the bits and make your own branch and PR. Your call. How would you like to proceed?

Let's check this in an I will make a separate PR to NOOP PushNotifications implementation.

dev/Common/TerminalVelocityFeatures-PushNotifications.h
DrusTheAxe: >Is it expected

No. But folks generally don't do inner-loop work for specific channels; it's heading for main, then cherrypick if need be. We have other files we checkin a baseline for inner-loop but the build generates updates/overrides; those inner-loop files aren't channel-savvy either. Something we'll need to address longer-term but not a problem today.

DrusTheAxe:
> Let's check this in an I will make a separate PR to NOOP PushNotifications implementation.

Sounds good to me.

So you Approve the PR?...

pmpurifoy (approved):
I will in a PR after this add experimental checks to PushNotifications APIs

PR #1406, opened 2021-09-13
Protocol Activation bug fixes

danielayala94:
This PR fixes a couple of issues:

1. **Truncated payloads delivered to the app**: the Push payload is now processed correctly to comply with AppLifecycle command line expectations (i.e. truncation happened at the first space character - other activation types are already escaping such characters, now we do it too. Then, the payload is unescaped at GetRawNotificationEventArgs.h).
2. **Payloads with special characters crashed the app**: Special characters weren't playing okay with wstring->string conversion at PushNotificationReceivedEventArgs, because WideCharToMultiByte was erroring out with ERROR_INSUFFICIENT_BUFFER.

### DIFF
--- a/dev/PushNotifications/GetRawNotificationEventArgs.h
+++ b/dev/PushNotifications/GetRawNotificationEventArgs.h
@@ -18,7 +18,10 @@ namespace winrt::Microsoft::Windows::PushNotifications
         {
             if (pair.Name() == L"payload")
             {
-                std::wstring payloadAsWstring{pair.Value()};
+                // Convert escaped components to its normal content
+                // from the conversion done in the LRP (see NotificationListener.cpp)
+                std::wstring payloadAsEscapedWstring{ pair.Value() };
+                std::wstring payloadAsWstring{ winrt::Windows::Foundation::Uri::UnescapeComponent(payloadAsEscapedWstring) };
                 return winrt::make<winrt::Microsoft::Windows::PushNotifications::implementation::PushNotificationReceivedEventArgs>(payloadAsWstring);
             }
         }
--- a/dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
+++ b/dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
@@ -132,16 +132,36 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
         }
     }
 
-    std::string PushNotificationReceivedEventArgs::Utf16ToUtf8(_In_z_ PCWSTR utf16)
+    std::string PushNotificationReceivedEventArgs::Utf16ToUtf8(_In_ std::wstring const& utf16string)
     {
-        int size_needed = WideCharToMultiByte(CP_UTF8, 0, utf16, -1, NULL, 0, nullptr, nullptr);
+        int size_needed = WideCharToMultiByte(
+            CP_UTF8,
+            0,
+            utf16string.data(),
+            static_cast<int>(utf16string.length()),
+            nullptr,
+            0,
+            nullptr,
+            nullptr);
+
         THROW_LAST_ERROR_IF(size_needed == 0);
 
-        // size_needed minus the null character
-        std::string utf8(size_needed - 1, 0);
-        int size = WideCharToMultiByte(CP_UTF8, 0, utf16, size_needed - 1, &utf8[0], size_needed - 1, nullptr, nullptr);
+        std::string utf8string;
+        utf8string.resize(size_needed);
+
+        int size = WideCharToMultiByte(
+            CP_UTF8,
+            0,
+            utf16string.data(),
+            static_cast<int>(utf16string.length()),
+            &utf8string[0],
+            size_needed,
+            nullptr,
+            nullptr);
+
         THROW_LAST_ERROR_IF(size == 0);
-        return utf8;
+
+        return utf8string;
     }
 }
 
--- a/dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
+++ b/dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
@@ -15,6 +15,8 @@
 #include <PushNotificationDummyDeferral.h>
 #include "ValueMarshaling.h"
 
+#include "utils.h"
+
 namespace winrt
 {
     using namespace Windows::ApplicationModel::Background;
@@ -55,7 +57,7 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
 
     std::vector<uint8_t> PushNotificationReceivedEventArgs::BuildPayload(std::wstring const& payload)
     {
-        std::string payloadToSimpleString = Utf16ToUtf8(payload.c_str());
+        std::string payloadToSimpleString = ConvertWideStringToUtf8String(payload);
 
         return { payloadToSimpleString.c_str(), payloadToSimpleString.c_str() + (payloadToSimpleString.length() * sizeof(uint8_t)) };
     }
@@ -130,38 +132,6 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
         {
             m_handledUnpackaged = value;
         }
-    }
-
-    std::string PushNotificationReceivedEventArgs::Utf16ToUtf8(_In_ std::wstring const& utf16string)
-    {
-        int size_needed = WideCharToMultiByte(
-            CP_UTF8,
-            0,
-            utf16string.data(),
-            static_cast<int>(utf16string.length()),
-            nullptr,
-            0,
-            nullptr,
-            nullptr);
-
-        THROW_LAST_ERROR_IF(size_needed == 0);
-
-        std::string utf8string;
-        utf8string.resize(size_needed);
-
-        int size = WideCharToMultiByte(
-            CP_UTF8,
-            0,
-            utf16string.data(),
-            static_cast<int>(utf16string.length()),
-            &utf8string[0],
-            size_needed,
-            nullptr,
-            nullptr);
-
-        THROW_LAST_ERROR_IF(size == 0);
-
-        return utf8string;
     }
 }
 
--- a/dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
+++ b/dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
@@ -1,4 +1,5 @@
 ﻿#include "pch.h"
+#include "utils.h"
 
 HRESULT NotificationListener::RuntimeClassInitialize(
     std::shared_ptr<ForegroundSinkManager> foregroundSinkManager,
@@ -21,12 +22,12 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationListener::OnRawNotificationReceived
 
     if (!m_foregroundSinkManager->InvokeForegroundHandlers(m_appId, payloadArray, payloadLength))
     {
-        // Command line format: ----WindowsAppSDKPushServer:-Payload:"<payloadAsEscapedUriFormat>"
-        std::wstring commandLine = L"----WindowsAppSDKPushServer:-Payload:\"";
+        // Command line format: ----WindowsAppRuntimePushServer:-Payload:"<payloadAsEscapedUriFormat>"
+        std::wstring commandLine = L"----WindowsAppRuntimePushServer:-Payload:\"";
 
         // Escape special characters to follow command line standards for any app activation type in AppLifecycle
         // (See AppInstance.cpp and Serialize() from other activation types)
-        std::wstring payloadAsWideString = GetPayloadAsWideString(payloadLength, payload);
+        std::wstring payloadAsWideString = ConvertByteArrayToWideString(payloadLength, payload);
         auto payloadAsEscapedUriFormat = winrt::Windows::Foundation::Uri::EscapeComponent(payloadAsWideString.c_str());
 
         commandLine.append(payloadAsEscapedUriFormat);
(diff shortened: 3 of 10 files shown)


dev/PushNotifications/GetRawNotificationEventArgs.h
```diff
@@ -18,7 +18,10 @@ namespace winrt::Microsoft::Windows::PushNotifications
         {
             if (pair.Name() == L"payload")
             {
-                std::wstring payloadAsWstring{pair.Value()};
+                // Convert escaped components to its normal content
+                // from the conversion done in the LRP (see NotificationListener.cpp)
+                std::wstring payloadAsEscapedWstring{ pair.Value() };
+                std::wstring payloadAsWstring{ winrt::Windows::Foundation::Uri::UnescapeComponent(payloadAsEscapedWstring) };
```
hulumane, line 24: Where exactly was Uri::EscapeComponent called? Just trying to reference the encoding issue : )

hulumane: Ok nmind. I see the code changes below.

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
```diff
@@ -45,14 +52,25 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationListener::OnRawNotificationReceived
...
     std::string utf8(size_needed - 1, 0);
-    int size = WideCharToMultiByte(CP_UTF8, 0, m_processName.c_str(), size_needed - 1, &utf8[0], size_needed - 1, nullptr, nullptr);
+    int size = WideCharToMultiByte(CP_UTF8, 0, wideString.c_str(), size_needed - 1, &utf8[0], size_needed - 1, nullptr, nullptr);
     THROW_LAST_ERROR_IF(size == 0);
     return utf8;
 }
+
+const std::wstring NotificationListener::GetPayloadAsWideString(unsigned int payloadLength, byte* payload)
+{
+    int size_needed = MultiByteToWideChar(CP_UTF8, 0, reinterpret_cast<LPCSTR>(payload), payloadLength, nullptr, 0);
```
hulumane, line 69: PCSTR.

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
```diff
@@ -45,14 +52,25 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationListener::OnRawNotificationReceived
...
+const std::string NotificationListener::ConvertWideStringToUtf8String(std::wstring const& wideString)
 {
-    int size_needed = WideCharToMultiByte(CP_UTF8, 0, m_processName.c_str(), -1, NULL, 0, nullptr, nullptr);
+    int size_needed = WideCharToMultiByte(CP_UTF8, 0, wideString.c_str(), -1, NULL, 0, nullptr, nullptr);
     THROW_LAST_ERROR_IF(size_needed == 0);
 
     // size_needed minus the null character
     std::string utf8(size_needed - 1, 0);
-    int size = WideCharToMultiByte(CP_UTF8, 0, m_processName.c_str(), size_needed - 1, &utf8[0], size_needed - 1, nullptr, nullptr);
+    int size = WideCharToMultiByte(CP_UTF8, 0, wideString.c_str(), size_needed - 1, &utf8[0], size_needed - 1, nullptr, nullptr);
```
hulumane, line 62: Nit: indent to multiline

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
```diff
@@ -45,14 +52,25 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationListener::OnRawNotificationReceived
...
     return utf8;
 }
+
+const std::wstring NotificationListener::GetPayloadAsWideString(unsigned int payloadLength, byte* payload)
+{
+    int size_needed = MultiByteToWideChar(CP_UTF8, 0, reinterpret_cast<LPCSTR>(payload), payloadLength, nullptr, 0);
+    THROW_LAST_ERROR_IF(size_needed == 0);
+
+    std::wstring payloadAsWideString(size_needed, 0);
+    size_needed = MultiByteToWideChar(CP_UTF8, 0, reinterpret_cast<LPCSTR>(payload), payloadLength, &payloadAsWideString[0], size_needed);
```
hulumane, line 73: nit: indent to multi line

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
hulumane, line 69: Nit: indent to multiline

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
hulumane, line 73: PCSTR

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
```diff
@@ -22,10 +22,17 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationListener::OnRawNotificationReceived
     if (!m_foregroundSinkManager->InvokeForegroundHandlers(m_appId, payloadArray, payloadLength))
     {
         std::string commandLine = "----WindowsAppSDKPushServer:-Payload:\"";
-        commandLine.append(reinterpret_cast<char*>(payload), payloadLength);
+
+        // Escape special characters to follow command line standards for any app activation type in AppLifecycle
+        // (See AppInstance.cpp and Serialize() from other activation types)
+        std::wstring payloadAsWideString = GetPayloadAsWideString(payloadLength, payload);
```
hulumane, line 28: Is there an API to directly add escape components to UTF-8 string? Right now we have UTF8->UTF16->UTF8 which seems to be highly redundant

hulumane: I'm thinking that the utf16 conversion is simply unecessary overhead here.

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
```diff
@@ -45,14 +52,25 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationListener::OnRawNotificationReceived
 }
 CATCH_RETURN()
 
-const std::string NotificationListener::ConvertProcessNameToUtf8String()
+const std::string NotificationListener::ConvertWideStringToUtf8String(std::wstring const& wideString)
 {
-    int size_needed = WideCharToMultiByte(CP_UTF8, 0, m_processName.c_str(), -1, NULL, 0, nullptr, nullptr);
+    int size_needed = WideCharToMultiByte(CP_UTF8, 0, wideString.c_str(), -1, NULL, 0, nullptr, nullptr);
```
hulumane, line 57: Nit: indent to multiline

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
danielayala94: There's no such API ☹️. The closest I found is [UrlEscapeA](https://docs.microsoft.com/en-us/windows/win32/api/shlwapi/nf-shlwapi-urlescapea) which seem to be able to do the same encoding as the winrt Uri does. The caveats are:

1. The input is not a URL.
2. It does the conversion up to a maximum of INTERNET_MAX_URL_LENGTH (2k characters, evidently insufficient).

What I can do is to use the ShellExecuteEx version for wide strings. This way I avoid the payload wstring->string conversion, as well as the process name wstring->string conversion too.

hulumane: Can you quickly elaborate on the total number of string conversions that we do for a Push scenario in Reunion today? 

At the end of the day, we give the developer a stream of bytes. So we want to avoid any unecessary wide character utf-16 conversions unless absolutely necessary.

danielayala94: We have another constraint at AppLifecycle: the command line is a wide string. This is why we end up doing another conversion at PushNotificationReceivedEventArgs to convert it to a byte array. So, the big picture looks like this:

WNS -> wpncore -> NotificationListener : byte array
NotificationListener -> ShellExecuteEx : byte array -> wide string (+1 conversion)
**AppLifecycle -> GetRawNotificationEventArgs -> PushNotificationReceivedEventArgs : wide string**
GetRawNotificationEventArgs (BuildPayload) : wide string -> byte array (+1 conversion)

dev/PushNotifications/GetRawNotificationEventArgs.h
hulumane, line 24: I would test the encoding decoding scenarios with payload limits < 8k. We should make sure that
a) Command Line is able to accept such large strings.
b) Any additional buffer from encoding Escape characters is accounted for in the command line
c) The Byte stream from the service is unaltered.

dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
```diff
@@ -132,16 +132,36 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
         }
     }
 
-    std::string PushNotificationReceivedEventArgs::Utf16ToUtf8(_In_z_ PCWSTR utf16)
+    std::string PushNotificationReceivedEventArgs::Utf16ToUtf8(_In_ std::wstring const& utf16string)
```
hulumane, line 135: Nit: Should be a global function IMHO and not part of PushNotificationREceivedEventArgs

dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
```diff
@@ -132,16 +132,36 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+
         THROW_LAST_ERROR_IF(size_needed == 0);
 
-        // size_needed minus the null character
-        std::string utf8(size_needed - 1, 0);
-        int size = WideCharToMultiByte(CP_UTF8, 0, utf16, size_needed - 1, &utf8[0], size_needed - 1, nullptr, nullptr);
+        std::string utf8string;
+        utf8string.resize(size_needed);
+
+        int size = WideCharToMultiByte(
```
hulumane, line 152: Nit: No need for defining two variables size and size_needed. I would just define size once and reuse it below

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
```diff
@@ -45,14 +50,26 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationListener::OnRawNotificationReceived
 }
 CATCH_RETURN()
 
-const std::string NotificationListener::ConvertProcessNameToUtf8String()
+const std::wstring NotificationListener::GetPayloadAsWideString(unsigned int payloadLength, byte* payload)
```
hulumane, line 53: To be consistent with the naming schema, this should be named as Utf8ToUtf16. The API should be global

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
```diff
@@ -45,14 +50,26 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationListener::OnRawNotificationReceived
...
 {
-    int size_needed = WideCharToMultiByte(CP_UTF8, 0, m_processName.c_str(), -1, NULL, 0, nullptr, nullptr);
+    int size_needed = MultiByteToWideChar(
```
hulumane, line 55: Just call it size since you are reusing this variable twice in the API calls below.

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
hulumane: Also it might make sense to have all string conversion utility helpers in a seperate header definition file so that they can be resued everywhere in our code base.

dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
danielayala94: Agree. Let's have a wpnlib for WAS 😊.

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
```diff
@@ -21,21 +21,26 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationListener::OnRawNotificationReceived
 
     if (!m_foregroundSinkManager->InvokeForegroundHandlers(m_appId, payloadArray, payloadLength))
     {
-        std::string commandLine = "----WindowsAppSDKPushServer:-Payload:\"";
-        commandLine.append(reinterpret_cast<char*>(payload), payloadLength);
-        commandLine.append("\"");
+        // Command line format: ----WindowsAppSDKPushServer:-Payload:"<payloadAsEscapedUriFormat>"
+        std::wstring commandLine = L"----WindowsAppSDKPushServer:-Payload:\"";
```
danielayala94, line 25: I don't know how this hasn't been changed yet.

hulumane (approved):

dev/PushNotifications/GetRawNotificationEventArgs.h
danielayala94: Results:

**a)** WNS still rejects 8K payloads ☹️. 5K works okay.
**b) and c)** 5K of chinese characters were correctly handled, delivering the payload to the unpackaged app.

Issue #1217, opened 2021-08-10
AnyCPU Class libraries are broken again


dotMorten:
**Describe the bug**

Internal bug https://microsoft.visualstudio.com/OS/_workitems/edit/36498869

When building an AnyCPU class library, the build is broken [again](https://github.com/microsoft/WindowsAppSDK/issues/939). Since it's not an application, the architecture runtime limitation shouldn't be in effect.
The issue appears to be due to an incorrect condition in `microsoft.windowsappsdk.foundation\1.0.0-experimental1\build\Microsoft.WindowsAppSDK.Bootstrap.targets`:

```xml
  <Target Name="BinPlaceBootstrapDll" Condition="'$(AppxPackage)' != 'true'" AfterTargets="Build">
    <Copy SourceFiles="$(MSBuildThisFileDirectory)..\runtimes\lib\native\$(_WindowsAppSDKFoundationPlatform)\Microsoft.WindowsAppSDK.Bootstrap.dll" DestinationFolder="$(OutDir)" OverwriteReadOnlyFiles="true" ContinueOnError="false"/>
  </Target>
```

A class library isn't an AppxPackage, but it isn't a running application either.

**Steps to reproduce the bug**

Compile a class lib using reunion:
```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFramework>net5.0-windows10.0.18362.0</TargetFramework>
    <TargetPlatformMinVersion>10.0.17763.0</TargetPlatformMinVersion>
    <Configurations>Debug;Release</Configurations>
    <Platforms>AnyCPU</Platforms>
  </PropertyGroup>

  <ItemGroup>
     <PackageReference Include="Microsoft.WindowsAppSDK" Version="1.0.0-experimental1" />
  </ItemGroup>
  
</Project>
```

Observe the error:

>1>C:\Users\mn\.nuget\packages\microsoft.windowsappsdk.foundation\1.0.0-experimental1\build\Microsoft.WindowsAppSDK.Bootstrap.targets(11,5): error MSB3030: Could not copy the file "C:\Users\mn\.nuget\packages\microsoft.windowsappsdk.foundation\1.0.0-experimental1\build\..\runtimes\lib\native\AnyCPU\Microsoft.WindowsAppSDK.Bootstrap.dll" because it was not found.

I managed to fix it locally by editing the above condition in the nuget package to this:
[...]

PR #1413, opened 2021-09-14
Copy Bootstrap .NET dll. Don't copy Bootstrap DLL if OutputType=Library

DrusTheAxe:
This corrects 2 problems:

1. Microsoft.WindowsAppSDK.Bootstrap.targets copied the Bootstrap dll from `...\runtimes\lib\**cpu**\...` and handled it when `**cpu**` = `Win32` but not when it's `AnyCPU`. Either should grab the `x86` dll
2. Microsoft.WindowsAppRuntime.Bootstrap.Net.dll wasn't getting copied for .NET projects

I tested with various permutations of C++ and C# projects.  This doesn't affect C++ projects (were doing the right thing before, this doesn't _negatively_ impact them). OTOH C# projects now build successfully with content as expected.

This addresses [AnyCPU Class libraries are broken again #1217](https://github.com/microsoft/WindowsAppSDK/issues/1217#)

[Bug 33477857](https://task.ms/33477857)

### DIFF
--- a/build/NuSpecs/Microsoft.WindowsAppSDK.Bootstrap.targets
+++ b/build/NuSpecs/Microsoft.WindowsAppSDK.Bootstrap.targets
@@ -5,10 +5,13 @@
   <PropertyGroup>
     <_WindowsAppSDKFoundationPlatform Condition="'$(Platform)' == 'Win32'">x86</_WindowsAppSDKFoundationPlatform>
     <_WindowsAppSDKFoundationPlatform Condition="'$(Platform)' != 'Win32'">$(Platform)</_WindowsAppSDKFoundationPlatform>
+    <_WindowsAppSDKBootstrapPlatform Condition="'$(_WindowsAppSDKFoundationPlatform)' == 'AnyCPU'">x86</_WindowsAppSDKBootstrapPlatform>
+    <_WindowsAppSDKBootstrapPlatform Condition="'$(_WindowsAppSDKFoundationPlatform)' != 'AnyCPU'">$(_WindowsAppSDKFoundationPlatform)</_WindowsAppSDKBootstrapPlatform>
   </PropertyGroup>
 
-  <Target Name="BinPlaceBootstrapDll" Condition="'$(AppxPackage)' != 'true'" AfterTargets="Build">
-    <Copy SourceFiles="$(MSBuildThisFileDirectory)..\runtimes\lib\native\$(_WindowsAppSDKFoundationPlatform)\Microsoft.WindowsAppRuntime.Bootstrap.dll" DestinationFolder="$(OutDir)" OverwriteReadOnlyFiles="true" ContinueOnError="false"/>
+  <Target Name="BinPlaceBootstrapDll" Condition="'$(AppxPackage)' != 'true' and '$(OutputType)' != 'Library'" AfterTargets="Build">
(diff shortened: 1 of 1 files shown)


DrusTheAxe:
I'm considering inverting the test, changing

```
<Target Name="BinPlaceBootstrapDll" Condition="'$(AppxPackage)' != 'true' and '$(OutputType)' != 'Library'" AfterTargets="Build">
```

to

```
<Target Name="BinPlaceBootstrapDll" Condition="'$(AppxPackage)' != 'true' and ('$(OutputType)' == 'Exe' or '$(OutputType)' == 'WinExe')" AfterTargets="Build">
```

i.e. replace "if !=Library" with "if ==Exe or WinExe"

According to [MSBuild's docs](https://docs.microsoft.com/visualstudio/msbuild/common-msbuild-project-properties?view=vs-2019) `OutputType` has 4 possible values: Library (default), Exe, Module, WinExe, but I've also noticed other values e.g. Microsoft.UI.Xaml.Markup.Compiler.Interop.targets supports `AppContainerExe`. Plus 'module' seems exceedingly rare at best. Overall I find precedent for all possible answers (==X and !=Y) so I'm open to suggestions if there's "common practice" or other compelling suggestions.

build/NuSpecs/Microsoft.WindowsAppSDK.Bootstrap.targets
```diff
@@ -5,10 +5,13 @@
   <PropertyGroup>
     <_WindowsAppSDKFoundationPlatform Condition="'$(Platform)' == 'Win32'">x86</_WindowsAppSDKFoundationPlatform>
     <_WindowsAppSDKFoundationPlatform Condition="'$(Platform)' != 'Win32'">$(Platform)</_WindowsAppSDKFoundationPlatform>
+    <_WindowsAppSDKBootstrapPlatform Condition="'$(_WindowsAppSDKFoundationPlatform)' == 'AnyCPU'">x86</_WindowsAppSDKBootstrapPlatform>
```
huichen123, line 8: in WindowsAppSDK-Nuget-Native.C.props, we are still trying to copy from path based on _WindowsAppSDKFoundationPlatform.

dotMorten: Wouldn't it make more sense to make it empty (or add !AnyCPU in line 7) so it isn't set at all for anycpu, and then deal with it not set?

DrusTheAxe: Native.C is only used by C/C++ projects, where AnyCPU doesn't exist?

DrusTheAxe: > make it empty (or add !AnyCPU in line 7) so it isn't set at all for anycpu, and then deal with it not set?

It could also be handled that way, but the conditional logic gets a bit messier (thinking long-term as well  as today).

This defines _WindowsAppSDKBootstrapPlatform as the key subdir to grab from, which 1-1 matches _WindowsappSDKFoundationPlatform except when building AnyCPU. So we precompute the answer here and then just use it below. That's a tad cleaner than using undefined/"" for AnyCPU and then computing it below.

I was careful to define a Bootstrap-specific property instead of trying to rely on / change the Foundation one, to avoid unwelcome entanglements down the road.

I tried to make the copy entirely overrideable e.g. msbuild /p:WindowsAppSDKBootstrapCopy=false to disable the copy, =true to do it and "" to "do the default", but it got a bit twisty and busted. I probably botched the logic but wasn't clearing up quickly so not worth the time sink. Not required today so I deferred that degree of control to a future date if/when desired. But this precompute-then-just-use leaves open the door to (more easily) go there, if/when desired

huichen123 (approved):

PR #1415, opened 2021-09-14
Add get_strong to PushNotificationManager::CreateChannelAsync 

pmpurifoy:
Need to get a strong reference to PushNotificationManager to keep it alive during the call to CreateChannelAsync.

### DIFF
--- a/dev/PushNotifications/PushNotificationManager.cpp
+++ b/dev/PushNotifications/PushNotificationManager.cpp
@@ -42,117 +42,10 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
     static wil::unique_com_class_object_cookie s_comActivatorRegistration;
     static wil::srwlock s_activatorInfoLock;
 
-    inline constexpr auto c_maxBackoff{ 5min };
-    inline constexpr auto c_initialBackoff{ 60s };
-    inline constexpr auto c_backoffIncrement{ 60s };
-
-    const HRESULT WNP_E_NOT_CONNECTED = static_cast<HRESULT>(0x880403E8L);
-    const HRESULT WNP_E_RECONNECTING = static_cast<HRESULT>(0x880403E9L);
-    const HRESULT WNP_E_BIND_USER_BUSY = static_cast<HRESULT>(0x880403FEL);
-
-    bool PushNotificationManager::IsChannelRequestRetryable(const hresult& hr)
-    {
-        switch (hr)
-        {
-        case HRESULT_FROM_WIN32(ERROR_TIMEOUT):
-        case WNP_E_NOT_CONNECTED:
-        case WPN_E_OUTSTANDING_CHANNEL_REQUEST:
-        case WNP_E_RECONNECTING:
-        case WNP_E_BIND_USER_BUSY:
-        case HRESULT_FROM_WIN32(RPC_S_SERVER_UNAVAILABLE):
-            return true;
-        default:
-            return false;
-        }
-    }
 
     winrt::IAsyncOperationWithProgress<winrt::Microsoft::Windows::PushNotifications::PushNotificationCreateChannelResult, winrt::Microsoft::Windows::PushNotifications::PushNotificationCreateChannelStatus> PushNotificationManager::CreateChannelAsync(winrt::guid const& remoteId)
     {
-        try
-        {
-            THROW_HR_IF(E_INVALIDARG, (remoteId == winrt::guid()));
-
-            // API supports channel requests only for packaged applications for v0.8 version
-            THROW_HR_IF(E_NOTIMPL, !AppModel::Identity::IsPackagedProcess());
-
-            auto cancellation{ co_await winrt::get_cancellation_token() };
-
-            cancellation.enable_propagation(true);
-
-            // Allow to register the progress and complete handler
-            co_await resume_background();
-
-            auto progress{ co_await winrt::get_progress_token() };
-
-            uint8_t retryCount = 0;
-            winrt::hresult channelRequestResult = E_PENDING;
-            PushNotificationChannelStatus status = PushNotificationChannelStatus::InProgress;
-
-            PushNotificationCreateChannelStatus
-                channelStatus = { status, channelRequestResult, retryCount };
-
-            progress(channelStatus);
-
-            for (auto backOffTime = c_initialBackoff; ; backOffTime += c_backoffIncrement)
-            {
-                try
-                {
-                    PushNotificationChannelManager channelManager{};
-                    winrt::PushNotificationChannel pushChannelReceived{ nullptr };
-
-                    pushChannelReceived = co_await channelManager.CreatePushNotificationChannelForApplicationAsync();
-
-                    PushNotificationTelemetry::ChannelRequestedByApi(
-                        S_OK,
-                        AppModel::Identity::IsPackagedProcess(),
-                        remoteId);
-
-                    co_return winrt::make<PushNotificationCreateChannelResult>(
-                        winrt::make<PushNotificationChannel>(pushChannelReceived),
-                        S_OK,
-                        PushNotificationChannelStatus::CompletedSuccess);
-
-                }
-                catch (...)
-                {
-                    auto channelRequestException = hresult_error(to_hresult(), take_ownership_from_abi);
-
-                    if ((backOffTime <= c_maxBackoff) && IsChannelRequestRetryable(channelRequestException.code()))
-                    {
-                        channelStatus.extendedError = channelRequestException.code();
-                        channelStatus.status = PushNotificationChannelStatus::InProgressRetry;
-                        channelStatus.retryCount = ++retryCount;
-
-                        progress(channelStatus);
-                    }
-                    else
-                    {
-
-                        PushNotificationTelemetry::ChannelRequestedByApi(
-                            channelRequestException.code(),
-                            AppModel::Identity::IsPackagedProcess(),
-                            remoteId);
-
-                        co_return winrt::make<PushNotificationCreateChannelResult>(
-                            nullptr,
-                            channelRequestException.code(),
-                            PushNotificationChannelStatus::CompletedFailure);
-                    }
-                }
-
-                co_await winrt::resume_after(backOffTime);
-            }
-        }
-        catch (...)
-        {
-            HRESULT hrError = wil::ResultFromCaughtException();
-            PushNotificationTelemetry::ChannelRequestedByApi(
-                hrError,
-                AppModel::Identity::IsPackagedProcess(),
-                remoteId);
-
-            THROW_HR(hrError);
-        }
+        return make_self<factory_implementation::PushNotificationManager>()->CreateChannelAsync(remoteId);
     }
 
     void PushNotificationManager::RegisterActivator(PushNotificationActivationInfo const& details)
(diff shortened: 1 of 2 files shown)


dev/PushNotifications/PushNotificationManager.h
```diff
@@ -4,6 +4,23 @@
 #pragma once
 #include "Microsoft.Windows.PushNotifications.PushNotificationManager.g.h"
 
+namespace winrt::Microsoft::Windows::PushNotifications::implementation
```
hulumane, line 7: Can you please move this back down to where the definition was earlier? It makes it easier to diff and compare the changes.

dev/PushNotifications/PushNotificationManager.h
```diff
@@ -4,6 +4,23 @@
...
+{
+    struct PushNotificationManager;
```
hulumane, line 9: Please remove this. Redfinition is not needed if you move the factory implementation back to its old location at the bottom of the file.

dev/PushNotifications/PushNotificationManager.h
```diff
@@ -4,6 +4,23 @@
...
+}
+
+namespace winrt::Microsoft::Windows::PushNotifications::factory_implementation
+{
+    struct PushNotificationManager : PushNotificationManagerT<PushNotificationManager, implementation::PushNotificationManager, static_lifetime>
+    {
+        winrt::Windows::Foundation::IAsyncOperationWithProgress<winrt::Microsoft::Windows::PushNotifications::PushNotificationCreateChannelResult, winrt::Microsoft::Windows::PushNotifications::PushNotificationCreateChannelStatus> CreateChannelAsync(winrt::guid const& remoteId);
+
+        private:
+            static bool IsChannelRequestRetryable(const winrt::hresult& hrException);
```
hulumane, line 19: Nit: indents for these APIs don't line up.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -358,3 +251,119 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        default:
+            return false;
+        }
+    }
+
+    winrt::Windows::Foundation::IAsyncOperationWithProgress<winrt::Microsoft::Windows::PushNotifications::PushNotificationCreateChannelResult, winrt::Microsoft::Windows::PushNotifications::PushNotificationCreateChannelStatus> PushNotificationManager::CreateChannelAsync(winrt::guid const& remoteId)
+    {
+        try
+        {
+            auto strong_this{ get_strong() };
```
hulumane, line 286: Is this line the only change from the original implementation?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -358,3 +251,119 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+
+    inline constexpr auto c_maxBackoff{ 5min };
+    inline constexpr auto c_initialBackoff{ 60s };
+    inline constexpr auto c_backoffIncrement{ 60s };
+
+    const HRESULT WNP_E_NOT_CONNECTED = static_cast<HRESULT>(0x880403E8L);
+    const HRESULT WNP_E_RECONNECTING = static_cast<HRESULT>(0x880403E9L);
+    const HRESULT WNP_E_BIND_USER_BUSY = static_cast<HRESULT>(0x880403FEL);
+
+    bool PushNotificationManager::IsChannelRequestRetryable(const hresult& hr)
```
hulumane, line 266: Thinking about this a bit more, there is no reason that IsChannelRequestRetryable should be a member of PushNotificationManager. 
Recommend: Remove this API from the header file and just call it IsChannelRequestRetryable here so that it is a global function for this namespace

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: Yup. The only change to the implementation that was needed was to get a strong reference.

hulumane: Gotcha. Makes sense.

hulumane: Alternatively, you can also use the get_weak pattern in the caller API

hulumane (approved):

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -358,3 +251,119 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+
+    inline constexpr auto c_maxBackoff{ 5min };
+    inline constexpr auto c_initialBackoff{ 60s };
+    inline constexpr auto c_backoffIncrement{ 60s };
+
+    const HRESULT WNP_E_NOT_CONNECTED = static_cast<HRESULT>(0x880403E8L);
+    const HRESULT WNP_E_RECONNECTING = static_cast<HRESULT>(0x880403E9L);
+    const HRESULT WNP_E_BIND_USER_BUSY = static_cast<HRESULT>(0x880403FEL);
+
+    bool IsChannelRequestRetryable(const hresult& hr)
```
DrusTheAxe, line 266: Function's not declared inline

`inline bool IsChannelRequestRetryable...`

or

`constexpr bool IsChannelRequestRetryable...`

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -358,3 +251,119 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+    }
+
+    winrt::Windows::Foundation::IAsyncOperationWithProgress<winrt::Microsoft::Windows::PushNotifications::PushNotificationCreateChannelResult, winrt::Microsoft::Windows::PushNotifications::PushNotificationCreateChannelStatus> PushNotificationManager::CreateChannelAsync(winrt::guid const& remoteId)
+    {
+        try
+        {
+            auto strong_this{ get_strong() };
+            THROW_HR_IF(E_INVALIDARG, (remoteId == winrt::guid()));
+
+            // API supports channel requests only for packaged applications for v0.8 version
```
DrusTheAxe, line 289: for version what? :P

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -358,3 +251,119 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+            auto cancellation{ co_await winrt::get_cancellation_token() };
+
+            cancellation.enable_propagation(true);
+
+            // Allow to register the progress and complete handler
+            co_await resume_background();
+
+            auto progress{ co_await winrt::get_progress_token() };
+
+            uint8_t retryCount = 0;
```
DrusTheAxe, line 301: NIT: `uint8_t retryCount{};`

Obligatory favorite link ;-) [ES.23: Prefer the {}-initializer syntax](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#es23-prefer-the--initializer-syntax)

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -358,3 +251,119 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+            winrt::hresult channelRequestResult = E_PENDING;
+            PushNotificationChannelStatus status = PushNotificationChannelStatus::InProgress;
+
+            PushNotificationCreateChannelStatus
+                channelStatus = { status, channelRequestResult, retryCount };
+
+            progress(channelStatus);
+
+            for (auto backOffTime = c_initialBackoff; ; backOffTime += c_backoffIncrement)
```
DrusTheAxe, line 310: So the logic is?

```
timeout=1min
DO
{
    JustDoIt()
    if not-done
        timeout += 1min
        co_await winrt::resume_after(timeout)
} UNTIL timeout>5min
```

So it tries for 1m, then 2m, 3m, 4m, 5m then gives up after 15minutes?

Is that intentional?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -358,3 +251,119 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+            {
+                try
+                {
+                    PushNotificationChannelManager channelManager{};
+                    winrt::PushNotificationChannel pushChannelReceived{ co_await channelManager.CreatePushNotificationChannelForApplicationAsync() };
+
+                    PushNotificationTelemetry::ChannelRequestedByApi(
+                        S_OK,
+                        AppModel::Identity::IsPackagedProcess(),
+                        remoteId);
```
DrusTheAxe, line 320: Seems odd how line 315 is 1line 149 chars long

but this is 4 lines that could be expressed as 1line 126 chars long

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: Yes, this is intentional. @sharath2727 can explain more here.

oldnewthing:
A much less intrusive version would be to put a `wil::winrt_module_reference` on the coroutine stack. Because really the only thing the strong reference is used for is the DLL reference.

DrusTheAxe (approved):

dev/PushNotifications/PushNotificationManager.cpp
sharath2727: Yeah that was intentional @DrusTheAxe . Based on some initial analysis we have done, we realized Most of the time the errors are benign so we want to improve the reliability of fetching the channel. Basically we are trying after 1m, 2m, 3m, 4m, 5m and then give up. If we add up in the worst case scenario we are spending 15mins to complete this process.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -264,15 +265,15 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
 
             scopeExitToCleanRegistrations.release();
 
-            auto lock = s_activatorInfoLock.lock_exclusive();
+            auto lock{ s_activatorInfoLock.lock_exclusive() };
             s_pushTriggerRegistration = registeredTaskFromBuilder;
 
             PushNotificationTelemetry::ActivatorRegisteredByApi(S_OK, details.Activators());
         }
```
jonwis, line 273: Remove blank line here

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -264,15 +265,15 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
 
         catch(...)
         {
-            HRESULT hrError = wil::ResultFromCaughtException();
```
jonwis, line 275: This block (and others like it) should not throw a _new_ exception. Instead, use "re-throw" (just the `throw` keyword.) Like so:

```c++
catch(...)
{
    PushNotificationTelemetry::ActivatorRegisteredByApi(
        wil::ResultFromCaughtException(),
        details ? details.Activators() : PushNotificationRegistrationActivators::Undefined);
    throw;
}
```

That preserves the original exception state while allowing you a chance to clean up.

dev/WindowsAppRuntime_DLL/WindowsAppRuntime_DLL.vcxproj
```diff
@@ -383,7 +383,7 @@
...
     <Import Project="..\..\packages\Microsoft.FrameworkUdk.x86fre.10.0.22453.1006-210916-1712.ni-release-svc-reunion2110\build\native\Microsoft.FrameworkUdk.x86fre.targets" Condition="Exists('..\..\packages\Microsoft.FrameworkUdk.x86fre.10.0.22453.1006-210916-1712.ni-release-svc-reunion2110\build\native\Microsoft.FrameworkUdk.x86fre.targets')" />
     <Import Project="..\..\packages\Microsoft.Windows.CppWinRT.2.0.210806.1\build\native\Microsoft.Windows.CppWinRT.targets" Condition="Exists('..\..\packages\Microsoft.Windows.CppWinRT.2.0.210806.1\build\native\Microsoft.Windows.CppWinRT.targets')" />
-    <Import Project="..\..\packages\Microsoft.Windows.ImplementationLibrary.1.0.210803.1\build\native\Microsoft.Windows.ImplementationLibrary.targets" Condition="Exists('..\..\packages\Microsoft.Windows.ImplementationLibrary.1.0.210803.1\build\native\Microsoft.Windows.ImplementationLibrary.targets')" />
```
jonwis, line 386: Looking forward to when these are centralized for the project...

jonwis (changes requested):

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -73,6 +73,7 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
 
         try
         {
+            wil::winrt_module_reference moduleRef{};
```
hulumane, line 76: Might want to move it one level above the try catch. If developers add something outside the try catch for example, this would no longer apply.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -149,13 +149,8 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
         {
-            HRESULT hrError = wil::ResultFromCaughtException();
-            PushNotificationTelemetry::ChannelRequestedByApi(
-                hrError,
-                AppModel::Identity::IsPackagedProcess(),
-                remoteId);
-
-            THROW_HR(hrError);
+            PushNotificationTelemetry::ChannelRequestedByApi(wil::ResultFromCaughtException(), AppModel::Identity::IsPackagedProcess(), remoteId);
+            throw;
```
hulumane, line 153: Why throw a new exception. Can we rethrow instead?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -264,20 +259,16 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
         catch(...)
         {
-            HRESULT hrError{ wil::ResultFromCaughtException() };
-            PushNotificationTelemetry::ActivatorRegisteredByApi(hrError, 
-                details == nullptr ? PushNotificationRegistrationActivators::Undefined : details.Activators());
-
-            THROW_HR(hrError);
+            PushNotificationTelemetry::ActivatorRegisteredByApi(wil::ResultFromCaughtException(),
+                details ? details.Activators() : PushNotificationRegistrationActivators::Undefined);
+            throw;
```
hulumane, line 271: same

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -304,9 +295,8 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
         }
         catch (...)
         {
-            HRESULT hrError{ wil::ResultFromCaughtException() };
-            PushNotificationTelemetry::ActivatorUnregisteredByApi(hrError, activators);
-            THROW_HR(hrError);
+            PushNotificationTelemetry::ActivatorUnregisteredByApi(wil::ResultFromCaughtException(), activators);
+            throw;
```
hulumane, line 299: here too

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -329,9 +319,8 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
         }
         catch(...)
         {
-            HRESULT hrError{ wil::ResultFromCaughtException() };
-            PushNotificationTelemetry::ActivatorUnregisteredByApi(hrError, PushNotificationRegistrationActivators::PushTrigger | PushNotificationRegistrationActivators::ComActivator);
-            THROW_HR(hrError);
+            PushNotificationTelemetry::ActivatorUnregisteredByApi(wil::ResultFromCaughtException(), PushNotificationRegistrationActivators::PushTrigger | PushNotificationRegistrationActivators::ComActivator);
+            throw;
```
hulumane, line 323: here too

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: This was from @jonwis guidance here:

https://github.com/microsoft/WindowsAppSDK/pull/1415#discussion_r719935295

hulumane: Ah ok. Gotcha

hulumane (approved):

PR #1424, opened 2021-09-15
Adding trace logging to Environment Manager

dhoehna:
This PR adds simple trace logging to Environment Manager.  Every API surface call has a simple trace logging method.

This PR also has some warning fixes.


dev/EnvironmentManager/API/Microsoft.Windows.System.EnvironmentManager.Insights.h
```diff
@@ -0,0 +1,45 @@
...
+            scopeAsString = L"Process";
+        }
+        else if (scope == winrt::Microsoft::Windows::System::implementation::EnvironmentManager::Scope::User)
+        {
+            scopeAsString = L"User";
+        }
+        else
+        {
+            scopeAsString = L"Machine";
+        }
```
DrusTheAxe, line 39: No need for scope to be std::wstring as all you're doing is malloc+memcpy(+free) for a string constant

SUGGESTION: Refactor as a conversation function

```
constexpr PCWSTR ScopeToString(winrt::...::Scope scope)
{
    if (scope == ...::Process)
    {
        return L"Process";
    }
    ...
}
```

and change this function to

```
static void LogWithScopeAndMessage(...)
{
    auto scopeAsString{ ScopeToString(scope) };
    TraceLoggingClassWrite("CustomEvent", TraceLoggingValue(scopeAsString, "scope") ...);
}
```

dev/EnvironmentManager/API/Microsoft.Windows.System.EnvironmentManager.cpp
```diff
@@ -16,20 +17,23 @@ namespace winrt::Microsoft::Windows::System::implementation
...
+        EnvironmentManagerInsights::StringTelemetryEvent(L"Making EM for process");
         Microsoft::Windows::System::EnvironmentManager environmentManager{ nullptr };
         environmentManager = winrt::make<implementation::EnvironmentManager>(Scope::Process);
         return environmentManager;
     }
 
     Microsoft::Windows::System::EnvironmentManager EnvironmentManager::GetForUser()
     {
+        EnvironmentManagerInsights::StringTelemetryEvent(L"Making EM for User");
         Microsoft::Windows::System::EnvironmentManager environmentManager{ nullptr };
```
DrusTheAxe, line 29: NIT: remove nullptr?

Does EnvironmentManager have a void ctor? If so then nullptr can be removed.
(if it doesn't but ctor(ptr=null) is legal then consider adding a void ctor :-)

test/EnvironmentManagerTests/EnvironmentVariableHelper.h
```diff
@@ -165,7 +165,7 @@ inline std::wstring GetEnvironmentVariableForProcess(const std::wstring variable
     return environmentVariableValue;
 }
 
-inline void InjectIntoPath(bool isProcess, bool isUser, const std::wstring& pathPart, int index)
+inline void InjectIntoPath(bool isProcess, bool isUser, const std::wstring& pathPart, UINT index)
```
DrusTheAxe, line 168: `UINT`? How very retro :P

`UINT32` or `uint32_t` is more hip these days

test/EnvironmentManagerTests/EnvironmentVariableHelper.h
```diff
@@ -213,7 +213,7 @@ inline void InjectIntoPath(bool isProcess, bool isUser, const std::wstring& path
         }
     }
 
-    int currentIndex{ 0 };
+    UINT currentIndex{ 0 };
```
DrusTheAxe, line 216: NIT: `...currentIndex{};`

No need for the explicit 0. The default initialization sets that

DrusTheAxe (approved):
LGTM modulo comments noted

DrusTheAxe (approved):
LGTM modulo feedback

dev/EnvironmentManager/API/Microsoft.Windows.System.EnvironmentManager.Insights.h
```diff
@@ -0,0 +1,45 @@
+﻿#pragma once
+#include "Microsoft.Windows.System.EnvironmentManager.h"
+
+class EnvironmentManagerInsights : public wil::TraceLoggingProvider
+
+{
+    IMPLEMENT_TRACELOGGING_CLASS(
+        EnvironmentManagerInsights,
+        "Microsoft.Windows.System.EnvironmentManager.Insights",
```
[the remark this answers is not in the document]
dhoehna: Now that I think about it, insights is redundant.  Should this go into Microsoft.WIndows.System.EnvironmentManager.h instead?

dev/EnvironmentManager/API/Microsoft.Windows.System.EnvironmentManager.Insights.h
```diff
@@ -0,0 +1,45 @@
...
+        else if (scope == winrt::Microsoft::Windows::System::implementation::EnvironmentManager::Scope::User)
+        {
+            scopeAsString = L"User";
+        }
+        else
+        {
+            scopeAsString = L"Machine";
+        }
+
+        TraceLoggingClassWrite("CustomEvent", TraceLoggingValue(scopeAsString.c_str(), "scope"),
```
ZenBird, line 41: >TraceLoggingClassWrite [](http://example.com/codeflow?start=8&length=22)

This will not log the PartB information. Any reason why you are calling TraceLoggingClassWrite explicitly instead of defining an event (CustomEvent) and logging the data as payload?

dev/EnvironmentManager/API/Microsoft.Windows.System.EnvironmentManager.cpp
dhoehna: Yep.  nullptr constructor is made by default.  In fact, this has to be done in this case.

https://docs.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/author-apis#if-youre-authoring-a-runtime-class-in-a-windows-runtime-component

Whether you have IDL-declared constructor(s) or not, a constructor overload that takes std::nullptr_t is generated on your projected type. **Calling the std::nullptr_t constructor is the first of two steps in consuming the runtime class from the same compilation unit.** For more details, and a code example, see Consume APIs with C++/WinRT.

test/EnvironmentManagerTests/EnvironmentVariableHelper.h
dhoehna: I don't want to be retro.

dev/EnvironmentManager/API/Microsoft.Windows.System.EnvironmentManager.cpp
dhoehna: In fact.  If you try anything else the compiler will get stuck in a loop.

dev/EnvironmentManager/API/Microsoft.Windows.System.EnvironmentManager.Insights.h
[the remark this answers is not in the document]
DrusTheAxe: Your call. Depends if you want your insights support in its own file or in the same as other stuff.

dev/EnvironmentManager/API/Microsoft.Windows.System.EnvironmentManager.Insights.h
dhoehna: image

By partB i'm guessing you mean the message?  Both are logged.

I'm calling TraceLoggingClassWrite because that is how I saw how to call a custom event.  Is there a better way to call a custom event?

I could convert the scope to a string and add it to a std::wstring and call StringTelemetryEvent and that way there isn't a need for a customer event.

Is there a different way to call a custom event?

ZenBird: Nope. PartB contains the information about the SDK and other fields that must be logged with all data coming from WindowsAppSDK. This will help us produce metrics on the usage and stability of code in different versions of WindowsAppSDK.

ZenBird: You could have a member similar to m_scope like m_scopeString and log that as part of the payload in the event. Alternately, you could log the enum (as int) as part of the payload.

dev/EnvironmentManager/API/Microsoft.Windows.System.EnvironmentManager.Insights.h
```diff
@@ -0,0 +1,45 @@
...
+        EnvironmentManagerInsights,
+        "Microsoft.Windows.System.EnvironmentManager.Insights",
+        // {fbce41fa-f7bc-538e-11f2-abe1b9d2ab23}
+        (0xfbce41fa, 0xf7bc, 0x538e, 0x11, 0xf2, 0xab, 0xe1, 0xb9, 0xd2, 0xab, 0x23));
+
+    DEFINE_COMPLIANT_MEASURES_EVENT_STRING(StringTelemetryEvent, // Event name 
+        MICROSOFT_KEYWORD_MEASURES, // Check wil/traceloggingconfig.h for a full list 
+        stringvalue/* value to be logged*/);
+
+    DEFINE_COMPLIANT_TELEMETRY_EVENT_PARAM2(ScopeTelemetryEvent, //Event name
```
ZenBird, line 17: >ScopeTelemetryEvent [](http://example.com/codeflow?start=44&length=19)

Where is this event being logged? Isn't this event identical to the LogWithScopeAndMessage function with the difference that LogWithScopeAndMessage converts the scope to string while this logs the scope as an enum.

dhoehna: You commented before I pushed. 

This was an attempt to log the scope and string before I moved onto the custom event.  It is gone now and replaced with a customer even "LogMessageAndScope"

ZenBird (approved):
:shipit:

PR #1428, opened 2021-09-16
Fix signing pattern to capture all winmd and dll in the transport package

kythant:
Fix signing pattern to capture all winmd and dll in the transport package

### DIFF
--- a/build/ProjectReunion-BuildFoundation.yml
+++ b/build/ProjectReunion-BuildFoundation.yml
@@ -196,6 +196,7 @@ jobs:
     signingPattern: |
         **/*.winmd
         **/*.dll
+        **/*.exe
   dependsOn:
     - Build
     - PublishMRT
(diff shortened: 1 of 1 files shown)


build/ProjectReunion-BuildFoundation.yml
```diff
@@ -194,9 +194,8 @@ jobs:
 - job: SignBinariesAndPublishSymbols
   variables:
     signingPattern: |
-        lib\uap10.0\*.winmd
-        lib\net5.0-windows\*.dll
-        runtimes\*\*\*.dll
+        **/*.winmd
+        **/*.dll
```
DrusTheAxe, line 198: Also needs `**/*.exe*` ?

DrusTheAxe (approved):

build/ProjectReunion-BuildFoundation.yml
kythant: Yea

PR #1429, opened 2021-09-16
Condense ProxyStubs to WindowsAppRuntime.ProxyStub.dll

pmpurifoy:
Condensed DynamicDependencyDataStore/DynamicDependencyLifetimeManager.ProxyStub.dll into WindowsAppRuntime.ProxyStub.dll. Other features that need to define their com interfaces for their ProxyStubs will define them here.

TODO: Change the dll paths in the appxmanifest of DataStore/DDLM packages.


dev/WindowsAppRuntime.ProxyStub/WindowsAppRuntime.idl
```diff
@@ -16,13 +17,32 @@ interface IDynamicDependencyDataStore : IUnknown
     HRESULT GetPackageFullName([out, retval] LPWSTR* packageFullName);
 };
 
+[object]
```
jonwis, line 20: Feels a little strange to _move_ this file here, rather than `#include`-ing it? Or having the common proxystub implementation reference the `_p.c` files and compile them?

dev/WindowsAppRuntime.ProxyStub/WindowsAppRuntime.ProxyStub.vcxproj
```diff
@@ -161,6 +159,34 @@
     <LinkIncremental>false</LinkIncremental>
     <CustomBuildAfterTargets>PostBuildEvent</CustomBuildAfterTargets>
   </PropertyGroup>
+  <ItemDefinitionGroup Condition="'$(Configuration)'=='Debug'">
```
DrusTheAxe, line 162: Where'd this come from?

This is doing the hybrid CRT game, but we shouldn't be mucking with that in individual projects anymore. \Directory.build.props imports \HybridCRT.props with all the right settings so everything in the repository gets hybrid CRT support.

Remove this block and restore line 45 ?

dev/WindowsAppRuntime.ProxyStub/WindowsAppRuntime.idl
DrusTheAxe: Yes. I don't understand why this was _moved_ (ditto the coclass below). You're implying we should only have 1 .idl for the whole project. Most unwise.

DrusTheAxe (changes requested):

DrusTheAxe:
I don't understand how this is intended to work. You're coming "all COM proxy/stub" into 1 dll, but we have packaged COM in Main + Singleton + DDLM packages. That's different versioning models, as well as different processes. For instance, the DDLM processes are tiny - they shouldn't be loading Push or other support.

PackagedCOM requires `<Application>` scope `<Extension>`s today so that means main packages. Where does 'the' single WAR.ps.dll come into play?

BadMojo(TM).

If you want to advocate for 1 proxy/stub dll _per MSIX package_ we could discuss further. But that's not what's going on here.

PR #1431, opened 2021-09-16
Fixes for the E2E Unpackaged scenario

danielayala94:
While doing extensive testing for the unpackaged apps E2E scenario, I caught a couple of bugs:

**1. Bug on foreground notifications**: On platform initialization, if the app list size is 0, the NotificationListenerManager was never initialized with a ForegroundSinkManager. Even after adding a listener, the ForegroundSinkManager reference on NotificationListenerManager remained empty. Now we initialize the NotificationListenerManager appropriately regardless of the app list size.
**2. NotificationSink update bug:** Whenever a channel request arrives, wpncore cleans the corresponding NotificationSink of the app. In such case, the LRP doesn't know if the NotificationSink (a NotificationListener in LRP jargon) is still valid. This caused new push notifications to be dropped and block the channel. With the fix, whenever RegisterForegroundActivator or RegisterLongRunningActivator are called, we remove the existing NotificationListener and replace it with a new one, doing a best effort to have an up-to-date listener. **This may still require a RegisterLongRunningActivator call from WAS channel request, if applicable.**


dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListenerManager.cpp
```diff
@@ -15,30 +18,31 @@ void NotificationListenerManager::Initialize(std::shared_ptr<ForegroundSinkManag
...
-        AgileRef agileListener;
-        THROW_IF_FAILED(AsAgile(listener.Get(), &agileListener));
-        m_notificationListeners.insert({ appId, agileListener });
-    }
+    auto lock = m_lock.lock_exclusive();
+
+    // Make sure we keep the long running sink up-to-date with wpncore.
+    ComPtr<INotificationListener> newListener;
+    THROW_IF_FAILED(MakeAndInitialize<NotificationListener>(&newListener, m_foregroundSinkManager, appId, processName));
+    THROW_IF_FAILED(PushNotifications_RegisterNotificationSinkForFullTrustApplication(appId.c_str(), newListener.Get()));
```
robertzhou-wpn, line 29: Not sure, Just wondering why we don't need to  call "THROW_IF_FAILED(PushNotifications_RegisterFullTrustApplication(appId.c_str(), GUID_NULL))" as the original code does?

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListenerManager.cpp
```diff
@@ -15,30 +18,31 @@ void NotificationListenerManager::Initialize(std::shared_ptr<ForegroundSinkManag
...
+
+    // Before updating the listener in the map,
+    // remove any previous listener that belongs to the app.
+    m_notificationListeners.erase(appId);
```
hulumane, line 33: I'm not sure I understand this. Why update the map with a new Listener if one already exists? We only need one listner per app right?

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListenerManager.cpp
hulumane: Right. If the registration does not exist, this call to register a sink will likely fail. What was the reasoning for deleting that here?

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListenerManager.cpp
```diff
@@ -15,30 +18,31 @@ void NotificationListenerManager::Initialize(std::shared_ptr<ForegroundSinkManag
 
 void NotificationListenerManager::AddListener(std::wstring appId, std::wstring processName)
 {
-    auto lock = m_lock.lock_exclusive();
-
     THROW_HR_IF(E_INVALIDARG, appId.empty());
     THROW_HR_IF(E_INVALIDARG, processName.empty());
 
-    if (m_notificationListeners.find(appId) == std::end(m_notificationListeners))
```
hulumane, line 23: Personally, I would just revert the old code and move the lock down. I don't see why we need any modifications to this method.

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListenerManager.h
```diff
@@ -9,7 +9,10 @@ class NotificationListenerManager
 public:
     NotificationListenerManager() {};
 
-    void Initialize(std::shared_ptr<ForegroundSinkManager> foregroundSinkManager, std::map<std::wstring, std::wstring>& appIdList);
+    // This function has to be called after initializing the ForegroundSinkManager during Platform initialization
+    void Initialize(std::shared_ptr<ForegroundSinkManager> foregroundSinkManager);
+
+    void SetAppIdMapping(std::map<std::wstring, std::wstring>& appIdList);
```
hulumane, line 15: The map should be a const reference?

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListenerManager.cpp
danielayala94: It's a mistake. Thanks for catching this!

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListenerManager.cpp
danielayala94: Please read the description. The faulty scenario is:

- App launches the first time
- App requests a channel
- App calls PushReceived -> The SDK calls the LRP to register a foreground sink (and a NotificationListener). We set a NotificationSink in wpncore.
- App closes
- App launches for the second time
- App requests a channel -> **Wpncore cleans the NotificationSink. Push notifications won't arrive to the LRP, and we have a stale Listener on the LRP.**

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListenerManager.cpp
danielayala94: Please looked at the scenario described in my other comment.

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListenerManager.cpp
danielayala94: Actually I think I can directly insert the new value without erasing first (a replacement). Let me look into this.

hulumane: I have some concerns with this fix. Before I voice them out though, can you please point me to where we clean the NotificationSink in wpncore as part of the last step?

hulumane: Note: We will have to slightly tweak our flow for LRP Sink registration if channel requests are delting the sink in wpncore.

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListenerManager.cpp
```diff
@@ -15,30 +18,31 @@ void NotificationListenerManager::Initialize(std::shared_ptr<ForegroundSinkManag
 
 void NotificationListenerManager::AddListener(std::wstring appId, std::wstring processName)
```
DrusTheAxe, line 19: const & ?

DrusTheAxe, line 19: is appId case-sensitive?

loneursid (changes requested):
All unpackaged unit tests are failing on this PR and needs to be addressed.

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListenerManager.cpp
pmpurifoy: AppId here is just guid as a string so it shouldn't have to worry about being case-sensitive.

pmpurifoy:
All unpkg tests are working now.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -196,6 +196,12 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
                         PushNotificationChannelManager channelManager{};
                         winrt::PushNotificationChannel pushChannelReceived{ co_await channelManager.CreatePushNotificationChannelForApplicationAsync(unpackagedAppUserModelId.get()) };
 
+                        auto notificationPlatform{ wil::CoCreateInstance<NotificationsLongRunningPlatform, INotificationsLongRunningPlatform>(CLSCTX_LOCAL_SERVER) };
```
hulumane, line 199: Please move this block to a helper function.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -261,17 +267,11 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-                wil::com_ptr<INotificationsLongRunningPlatform> notificationPlatform{
-                    wil::CoCreateInstance<NotificationsLongRunningPlatform, INotificationsLongRunningPlatform>(CLSCTX_LOCAL_SERVER) };
-
-                wil::unique_cotaskmem_string processName;
-                THROW_IF_FAILED(GetCurrentProcessPath(processName));
-                THROW_IF_FAILED(notificationPlatform->RegisterLongRunningActivator(processName.get()));
-
                 {
                     auto lock = s_activatorInfoLock.lock_exclusive();
+                    wil::unique_cotaskmem_string unpackagedAppUserModelId;
```
hulumane, line 272: Why is this block in a lock? There is no variable that needs a lock here.

dev/PushNotifications/PushNotificationManager.h
```diff
@@ -17,7 +17,7 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
         static void UnregisterAllActivators();
 
         static winrt::Windows::Foundation::IAsyncOperationWithProgress<winrt::Microsoft::Windows::PushNotifications::PushNotificationCreateChannelResult, winrt::Microsoft::Windows::PushNotifications::PushNotificationCreateChannelStatus> CreateChannelAsync(winrt::guid const remoteId);
-
+
```
hulumane, line 20: Blank line added? Why?

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
```diff
@@ -22,26 +21,21 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationListener::OnRawNotificationReceived
 
     if (!m_foregroundSinkManager->InvokeForegroundHandlers(m_appId, payloadArray, payloadLength))
     {
-        // Command line format: ----WindowsAppRuntimePushServer:-Payload:"<payloadAsEscapedUriFormat>"
-        std::wstring commandLine = L"----WindowsAppRuntimePushServer:-Payload:\"";
+        std::string commandLine = "----WindowsAppRuntimePushServer:-Payload:\"";
+        commandLine.append(reinterpret_cast<char*>(payload), payloadLength);
```
hulumane, line 25: Why are we re-introducing this escape character bug? @danielayala94 ?

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
```diff
@@ -50,3 +44,15 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationListener::OnRawNotificationReceived
     return S_OK;
 }
 CATCH_RETURN()
+
+const std::string NotificationListener::ConvertProcessNameToUtf8String()
```
hulumane, line 48: Not needed. Just use the helper method provided by @sharath2727 in his PR.

hulumane: I would remove this method and just call his Helper with m_processName wherever needed

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.h
```diff
@@ -10,6 +10,8 @@ class NotificationListener : public Microsoft::WRL::RuntimeClass<::ABI::Microsof
     STDMETHOD(OnRawNotificationReceived)(unsigned int payloadLength, _In_ byte* payload, _In_ HSTRING correlationVector) noexcept;
 
 private:
+    const std::string ConvertProcessNameToUtf8String();
```
hulumane, line 13: Remove. Not needed. Use @sharath2727 's helper method

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListenerManager.cpp
```diff
@@ -3,42 +3,46 @@
...
+    auto lock = m_lock.lock_exclusive();
+
+    // Make sure we keep the long running sink up-to-date with wpncore.
+    ComPtr<INotificationListener> newListener;
+    THROW_IF_FAILED(MakeAndInitialize<NotificationListener>(&newListener, m_foregroundSinkManager, appId, processName));
+    THROW_IF_FAILED(PushNotifications_RegisterNotificationSinkForFullTrustApplication(appId.c_str(), newListener.Get()));
+
+    // Before updating the listener in the map,
+    // remove any previous listener that belongs to the app.
+    m_notificationListeners.erase(appId);
```
hulumane, line 33: Why are we erasing stuff? It isn't necessary. Just replace whatever exists in the map

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
jonwis: Saw something like this in the other PR as well.

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
```diff
@@ -22,26 +21,21 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationListener::OnRawNotificationReceived
...
+        shellExecuteInfo.cbSize = sizeof(SHELLEXECUTEINFOA);
         shellExecuteInfo.fMask = SEE_MASK_NOCLOSEPROCESS | SEE_MASK_DOENVSUBST;
-        shellExecuteInfo.lpFile = m_processName.c_str();
+        shellExecuteInfo.lpFile = processNameAsUtf8String.c_str();
         shellExecuteInfo.lpParameters = commandLine.c_str();
 
         shellExecuteInfo.nShow = SW_NORMAL;
 
-        if (!ShellExecuteEx(&shellExecuteInfo))
+        if (!ShellExecuteExA(&shellExecuteInfo))
```
jonwis, line 38: New code should use the -W variants. Why the shift to -A?

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
DrusTheAxe: > use the helper method provided by @sharath2727 Sharath Manchala FTE in his PR.

link?

jonwis (approved):
Ok, but please apply @hulumane 's feedback.

loneursid:
> All unpkg tests are working now.

Verified, they do indeed all pass now as long as you have this FrameworkUDK fix installed on your devide: https://microsoft.visualstudio.com/OS/_git/os.2020/pullrequest/6502196

thanks.

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
danielayala94: @pmpurifoy Please look at the existing code in WNP_LRP. This seems to be an incorrect merge.

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
danielayala94: This function was used before the Protocol activation fixes. @pmpurifoy needs to fix the merge with WNP_LRP (i.e. remove this function).

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListenerManager.cpp
danielayala94: Yes. `insert_or_assign` should be used instead.

danielayala94 (commented):
@pmpurifoy I cannot block the PR because I opened it. But please do a thorough review of all files edited on this PR against the current WNP_LRP code. I noticed that NotificationListener is missing part of the fixes for Protocol.

loneursid (changes requested):
This cannot be merged into WNP_LRP as it would regress other changes. It seems that WNP_LRP was merged, then the merge was reverted. This means that if this branch is merged into WNP_LRP as is, we'll regress previous changes. Furthermore, if WNP_LRP was then merged into main, the regression would propagate to main. (the branch is currently bringing changes to 182 files, most of them outside of the Notifications firectories...)

@pmpurifoy, please work with @jonwis to see what is the correct somution here. I've the feeling that this branch will have to be reset to the initial merge of WNP_LRP or the the commit prior to that. That would mean reaplying the few "nits" PR that have been merged since then (as they would be lost after the reset operation.)

Also not that whatever issues you faced the first time you merged WNP_LRP and prompted the revert, will have to be adressed sooner or later. This branch will have to be merged into WNP_LRP at some point and that merge will need to be clean and not introduce any unwanted issues.

PR #1432, opened 2021-09-17
Add arm64 support for Installer

### LINKED ISSUE #479: Installer build and runtime detection of ARM64 is incorrect
_M_ARM64 does not detect properly when doing build configuration, and runtime detection of Arm64 also appears to lie and say it is x86. This causes problems in both the build of the installer to include the arm64 packages in the exe as well as runtime detection of when to install those packages. This issue tracks resolution of both problems with building an installer that carries arm64 and deploying it correctly to arm64 devices.

_Originally posted by @riverar in https://github.com/microsoft/ProjectReunion/pull/444#discussion_r577202068_

### LINKED ISSUE #714: Rewrite Installer Tests to use TAEF
Per feedback from #656 , change tests to use TAEF

dkbennett:
Fix #479
Fix #714

* Changes system architecture detection to use IsWow64Process2 since GetSystemInfo is not truthful about ARM64.
* Updates the test architecture detection to also use IsWow64Process2.
* Enables ARM64 packages to be present in all installer builds so any architecture version of the installer will do the right thing on any system in which it runs.
* Adds missing tests for Singleton.msix.
* Fixes project to build ARM64.
* Fixes tests to use TAEF #714 - because it was required in order for arm64 to build, as MSTest did not support ARM64.


installer/dev/install.cpp
```diff
@@ -66,9 +66,24 @@ namespace WindowsAppRuntimeInstaller {
...
+            systemArchitecture = ProcessorArchitecture::X86;
+            break;
+        case IMAGE_FILE_MACHINE_AMD64:
+            systemArchitecture = ProcessorArchitecture::X64;
+            break;
+        case IMAGE_FILE_MACHINE_ARM64:
+            systemArchitecture = ProcessorArchitecture::Arm64;
+            break;
+        default:
+            THROW_HR(HRESULT_FROM_WIN32(ERROR_NOT_SUPPORTED));
```
DrusTheAxe, line 85: SUGGESTION: `THROW_HR_MSG(..., "nativeMachine=%hu", nativeMachine);`

In case we get an unrecognized value, log the value as just knowing the callstack that this line failed won't be as helpful

installer/test/InstallerFunctionalTests/helpers.cpp
```diff
@@ -110,8 +110,19 @@ namespace WindowsAppRuntimeInstallerTests
 
     ProcessorArchitecture GetSystemArchitecture()
     {
```
DrusTheAxe, line 112: Isn't this the same as the code above? Why the duplication?

DrusTheAxe (approved):
LGTM modulo comments

riverar (approved):

installer/test/InstallerFunctionalTests/helpers.cpp
dkbennett: One is in the test project, the other is part of the installer.

PR #1435, opened 2021-09-17
Use remoteId as const value

sharath2727:
Pass remoteId as const value to CreateChannelAsync. As the value is const we don't need to pass by reference.

### DIFF
--- a/dev/PushNotifications/PushNotificationManager.h
+++ b/dev/PushNotifications/PushNotificationManager.h
@@ -16,7 +16,7 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
         static void UnregisterActivator(Microsoft::Windows::PushNotifications::PushNotificationRegistrationActivators const& activators);
         static void UnregisterAllActivators();
 
-        static winrt::Windows::Foundation::IAsyncOperationWithProgress<winrt::Microsoft::Windows::PushNotifications::PushNotificationCreateChannelResult, winrt::Microsoft::Windows::PushNotifications::PushNotificationCreateChannelStatus> CreateChannelAsync(winrt::guid const& remoteId);
+        static winrt::Windows::Foundation::IAsyncOperationWithProgress<winrt::Microsoft::Windows::PushNotifications::PushNotificationCreateChannelResult, winrt::Microsoft::Windows::PushNotifications::PushNotificationCreateChannelStatus> CreateChannelAsync(winrt::guid const remoteId);
 
         static bool IsActivatorSupported(Microsoft::Windows::PushNotifications::PushNotificationRegistrationActivators const& activators);
     };
--- a/dev/PushNotifications/PushNotificationManager.cpp
+++ b/dev/PushNotifications/PushNotificationManager.cpp
@@ -118,7 +118,7 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
     }
     CATCH_RETURN()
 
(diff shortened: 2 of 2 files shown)


DrusTheAxe:
Works either way. Efficiency is quibbling -- GUID=16 bytes so pass-by-reference is a pointer plus dereference, pass-by-value may be the same or more efficient with registers or whatnot (depends how the compiler handles larger structures, calling convention, etc)

@jonwis is there general guidance these days on parameters as `const GUID` vs `const GUID&` ?

loneursid:
> Works either way. Efficiency is quibbling -- GUID=16 bytes so pass-by-reference is a pointer plus dereference, pass-by-value may be the same or more efficient with registers or whatnot (depends how the compiler handles larger structures, calling convention, etc)
> 
> @jonwis is there general guidance these days on parameters as `const GUID` vs `const GUID&` ?

The reason for this change isn't about performance though, it's about fixing an async big. The function is async and makes use of the GUID parameter. If the GUID is passed by reference, then we need to take a copy inside of the function to ensure the object lifetime matches the function's. Here letting the compiler make copy the GUID seems the right thing to do (less code).

loneursid (approved):

hulumane (approved):

dev/PushNotifications/PushNotificationManager.h
```diff
@@ -16,7 +16,7 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
         static void UnregisterActivator(Microsoft::Windows::PushNotifications::PushNotificationRegistrationActivators const& activators);
         static void UnregisterAllActivators();
 
-        static winrt::Windows::Foundation::IAsyncOperationWithProgress<winrt::Microsoft::Windows::PushNotifications::PushNotificationCreateChannelResult, winrt::Microsoft::Windows::PushNotifications::PushNotificationCreateChannelStatus> CreateChannelAsync(winrt::guid const& remoteId);
+        static winrt::Windows::Foundation::IAsyncOperationWithProgress<winrt::Microsoft::Windows::PushNotifications::PushNotificationCreateChannelResult, winrt::Microsoft::Windows::PushNotifications::PushNotificationCreateChannelStatus> CreateChannelAsync(winrt::guid const remoteId);
```
hulumane, line 19: I'm curious that the idl auto generated guid& instead of guid as the header. Is that true and does this impact the projection in any negative way?

DrusTheAxe: @kennykerr ?

sharath2727: Yeah kenny. Thanks for your response. The above API is a coroutine and the value of remoteId is not held on the stack if the API is suspended. So that is the reason I had to convert this to const value.

hulumane: @kennykerr Should we have c++/winrt generate the pass by value argument automatically for co-routines rather than pass by ref knowing some of the limitations here with argument lifetime?

kennykerr: We could, although its probably more reliable to build with code analysis enabled and then you end up with a compiler warning/error:

```C++
winrt::Windows::Foundation::IAsyncOperation<winrt::guid> Async(winrt::guid const& value)
{
    co_return value;
}
```

```
main.cpp(5): warning C26811: Lifetime of the memory referenced by parameter ''value'' might end by the time the coroutine is resumed (lifetime.1).
```

image

sharath2727: This is good to learn. However, I am wondering which is the right approach here? Enabling code analysis or make the C++/WinRT generate differently for coroutines?

kennykerr: C++/WinRT has no idea how you are going to implement the method - you may not use coroutines at all - and once the code is generated all control is lost. However, the C++ compiler can *always* validate that the code is safe regardless of how the code evolves over time. Therefore, having the compiler check the code's correctness seems far more useful.

PR #1442, opened 2021-09-18
Add PRI generation support for C++ libraries

rohanp-msft:
Turns out that it's not true that `TargetPlatformIdentifier` is UAP for .NET version less than 5.0. I tested with a Windows Forms app with .NET v4.8. Yes, we require .NET 5 or greater for WAS, but there is no need to have a check here.

We now just set `TargetPlatformIdentifierAdjusted` to UAP for all cases. MrtCore.PriGen.targets will not be imported for projects where we don't want to use it. That's currently already the case for WAPPROJ only.

**How verified**
Commit 4 verification:
- Create a new WinUI C# app using the template.
- Go into the NuGet package. Copy paste file contents.
- Add a RESW file. Rebuild and run.
- Confirm PRI generated. Confirm correct.


dev/MRTCore/packaging/MrtCore.PriGen.targets
```diff
@@ -11,8 +11,8 @@
     <!-- Any .NET 5 or greater version is valid here! -->
     <TargetPlatformIdentifierAdjusted Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp' and $([MSBuild]::VersionGreaterThanOrEquals($(TargetFrameworkVersion), '5.0'))">UAP</TargetPlatformIdentifierAdjusted>
 
-    <!-- Add support for console apps. -->
-    <TargetPlatformIdentifierAdjusted Condition="'$(OutputType)' == 'Exe'">UAP</TargetPlatformIdentifierAdjusted>
+    <!-- Add support for console apps and libraries. -->
+    <TargetPlatformIdentifierAdjusted Condition="'$(OutputType)' == 'Exe' or '$(OutputType)' == 'Library'">UAP</TargetPlatformIdentifierAdjusted>
```
DrusTheAxe, line 15: What about OutputType=WinExe ?

DrusTheAxe:
Is PRI generation required or optional? If I don't want to use MRT can disable PRI generation? (or is it opt-in, I need to do something to generate PRI files, not merely use WinAppSDK)?

rohanp-msft:
> Is PRI generation required or optional? If I don't want to use MRT can disable PRI generation? (or is it opt-in, I need to do something to generate PRI files, not merely use WinAppSDK)?

It's on by default, but it can be turned off by setting `EnableCoreMrtTooling` to false.

dev/MRTCore/packaging/MrtCore.PriGen.targets
rohanp-msft: Will do!

huichen123, line 15: >TargetPlatformIdentifierAdjusted [](http://example.com/codeflow?start=5&length=32)

if this is not UAP, what's the result of this file?

rohanp-msft: Build error: error APPX3217: SDK folder containing 'Windows.props' for 'Windows 10.0.19041.0' cannot be located.

huichen123: then why do we need to the condition?

dev/MRTCore/packaging/MrtCore.PriGen.targets
```diff
@@ -1,18 +1,13 @@
...
-  <!-- With .NET 5 and above, the propery TargetPlatformIdentifier is not UAP but rather Windows.
-       Likewise, for other console apps, TargetPlatformIdentifier is Windows.
-       PRI generation in VS apps was developed for UWPs originally. Some of the binaries involved assume
+  <!-- PRI generation in VS apps was developed for UWPs originally. Some of the binaries involved assume
        that the app is a UWP. Change the value provided to said binaries to UAP. -->
   <PropertyGroup>
     <TargetPlatformIdentifierAdjusted>$(TargetPlatformIdentifier)</TargetPlatformIdentifierAdjusted>
 
-    <!-- Any .NET 5 or greater version is valid here! -->
-    <TargetPlatformIdentifierAdjusted Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp' and $([MSBuild]::VersionGreaterThanOrEquals($(TargetFrameworkVersion), '5.0'))">UAP</TargetPlatformIdentifierAdjusted>
```
rohanp-msft, line 12: @huichen123 please see the updated PR description for the reason for this removal. Please let me know if you're OK with it.
Also @axelandrejs if you have time

dev/MRTCore/packaging/MrtCore.PriGen.targets
rohanp-msft: Done

rohanp-msft: I'm sorry, I don't understand. We need to change the TPI to UAP to avoid this error. There may be other places in the managed helper code too where there is a dependency on the TPI being UAP.

huichen123: if this code only works when TargetPlatformIdentifierAdjusted is UAP, we need to set TargetPlatformIdentifierAdjusted to UAP unconditionally to make this file at least self consistent. If for any TPI this is undesirable, it needs to be filtered to not import this targets file.

dev/MRTCore/packaging/MrtCore.PriGen.targets
rohanp-msft: Been discussed in another comment thread.

dev/MRTCore/packaging/MrtCore.PriGen.targets
rohanp-msft: OK, I unconditionally set it in the file now. The import is already done conditionally. Right now, we don't do it only for WAPPROJ.

huichen123 (approved):
:shipit:

PR #1451, opened 2021-09-20
Support foregroundtask registration for packaged applications

sharath2727:


test/TestApps/PushNotificationsDemoPackage/Package.appxmanifest
```diff
@@ -11,7 +11,7 @@
 	<Identity
 	  Name="PushNotificationsWin32App"
 	  Publisher="CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US"
-	  Version="1.0.0.0" />
+	  Version="1.0.2.0" />
```
loneursid, line 15: Why the bump in version number? Is this intentional?

test/TestApps/PushNotificationsDemoApp/main.cpp
```diff
@@ -18,7 +18,7 @@ winrt::Windows::Foundation::IAsyncOperation<PushNotificationChannel> RequestChan
     // To obtain an AAD RemoteIdentifier for your app,
     // follow the instructions on https://docs.microsoft.com/azure/active-directory/develop/quickstart-register-app
     auto channelOperation = PushNotificationManager::CreateChannelAsync(
-        winrt::guid("ccd2ae3f-764f-4ae3-be45-9804761b28b2"));
+        winrt::guid("0160ee84-0c53-4851-9ff2-d7f5a87ed914"));
```
loneursid, line 21: Why the change in the guid? is this intentional or required?

test/TestApps/PushNotificationsDemoPackage/PushNotificationsDemoPackage.wapproj
```diff
@@ -48,6 +48,7 @@
     <GenerateTestArtifacts>True</GenerateTestArtifacts>
     <AppxBundlePlatforms>$(Platform)</AppxBundlePlatforms>
     <HoursBetweenUpdateChecks>0</HoursBetweenUpdateChecks>
+    <AppxPackageDir>C:\Users\vemancha\Desktop\</AppxPackageDir>
```
pmpurifoy, line 51: Make sure you cleanup local changes 😄

dev/PushNotifications/PushNotificationChannel.h
```diff
@@ -29,6 +30,17 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
         // IWpnForegroundSink
         HRESULT __stdcall InvokeAll(_In_ ULONG length, _In_ byte* payload, _Out_ BOOL* foregroundHandled) noexcept;
 
+        //INotificationHandler
+        HRESULT __stdcall OnRawNotificationReceived(unsigned int payloadLength, _In_ byte* payload, _In_ HSTRING /*correlationVector */) noexcept
+        {
+            BOOL foregroundHandled = false;
+            THROW_IF_FAILED(InvokeAll(payloadLength, payload, &foregroundHandled));
+            THROW_HR_IF(E_NOT_VALID_STATE, !foregroundHandled);
```
pmpurifoy, line 38: In this scenario, when the foreground isn't handled, we need to find a way to route the args to background activation. So that means protocol activation will have to happen here 😄 . Check the LRP to see how protocol activation is currently being done. Maybe start a commons file shared between LRP/WAS.

test/TestApps/PushNotificationsDemoPackage/PushNotificationsDemoPackage.wapproj
hulumane: This should be a relative path I imagine.

dev/PushNotifications/PushNotificationChannel.h
```diff
@@ -29,6 +30,17 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
         // IWpnForegroundSink
         HRESULT __stdcall InvokeAll(_In_ ULONG length, _In_ byte* payload, _Out_ BOOL* foregroundHandled) noexcept;
 
+        //INotificationHandler
+        HRESULT __stdcall OnRawNotificationReceived(unsigned int payloadLength, _In_ byte* payload, _In_ HSTRING /*correlationVector */) noexcept
```
hulumane, line 34: We probably should file a bug here to use the correlation vector in our telemetry?

dev/PushNotifications/PushNotificationChannel.h
hulumane: Correct. If the foreground event is not handled, we should re-launch the app with the payload using Protocol. That would mean that this line of code needs to go away:
 THROW_HR_IF(E_NOT_VALID_STATE, !foregroundHandled);

hulumane: Also please test this corner case scenario : )

test/TestApps/PushNotificationsDemoPackage/PushNotificationsDemoPackage.wapproj
sharath2727: @pmpurifoy that is why still in draft mode :)

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -89,15 +89,36 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                        pushArgs.Handled(true);
+                        handler(*strong, pushArgs);
+                    };
+                });
+            }
+            else
+            {
+                auto lock = m_lock.lock_exclusive();
+
+                THROW_HR_IF_NULL(E_UNEXPECTED, m_channelInfo.channelUri.c_str());
```
hulumane, line 108: This is incorrect. It should be:
THROW_HR_IF_NULL(E_UNEXPECTED, m_channelInfo);

You essentially want to return an error if channelInfo is not set, right? So use the winrt reference, not the fields in the reference.

dev/PushNotifications/PushNotificationChannel.h
sharath2727: I agree to the fact that we should not have the line set.  Firslty, let me reiterate we are dealing with packaged applications. Let me put the scenarios here:

1. Foreground handled - args.Cancel is set to true - this will avoid background task to be triggered
2. Foreground not handled - args.Cancel is set to false - then if the background task is also registered by the user then backgroundtask is triggered.

The above cases are handled if we remove the foregroundhandled check. Do we really need the protocol activation here.?

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -89,15 +89,36 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+
+                wchar_t appUserModelId[APPLICATION_USER_MODEL_ID_MAX_LENGTH] = {};
+                UINT32 appUserModelIdSize{ ARRAYSIZE(appUserModelId) };
+
+                THROW_IF_FAILED(GetCurrentApplicationUserModelId(&appUserModelIdSize, appUserModelId));
+
+                THROW_IF_FAILED(PushNotifications_RegisterNotificationSinkForFullTrustApplication(appUserModelId, this));
+
+                return m_foregroundHandlers.add(handler);
+
```
hulumane, line 118: Nit: Too many blank lines. Please remove them all before the brace.

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -89,15 +89,36 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                THROW_IF_FAILED(GetCurrentApplicationUserModelId(&appUserModelIdSize, appUserModelId));
+
+                THROW_IF_FAILED(PushNotifications_RegisterNotificationSinkForFullTrustApplication(appUserModelId, this));
+
+                return m_foregroundHandlers.add(handler);
+               
+
+            }
+            // One more check if channelInfo struct is null - we should throw E_UNEXPECTED
```
hulumane, line 121: Remove comment.

dev/PushNotifications/PushNotificationChannel.h
hulumane: I'm not sure I understand the question. The contract is straightforward here: If the developer chooses not to intercept the foreground event, then the expectation is that the background trigger is executed. In this case, there is a noop. So it is a bug : )

hulumane: I also think that packaged apps using BI suffer the same limitation here. We should file a seperate bug for that and fix it as well.

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -120,7 +141,23 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                m_channel.PushNotificationReceived(token);
+            }
+            else
+            {
+                auto lock = m_lock.lock_exclusive();
+
+                wchar_t appUserModelId[APPLICATION_USER_MODEL_ID_MAX_LENGTH] = {};
+                UINT32 appUserModelIdSize{ ARRAYSIZE(appUserModelId) };
+
+                THROW_IF_FAILED(GetCurrentApplicationUserModelId(&appUserModelIdSize, appUserModelId));
```
robertzhou-wpn, line 155: Nit: looks to me to register and unregister a sink, we will first get the appUserModelId. Can we put the logic of getting the appUserModelId into a function, which can then be called in both places?

dev/PushNotifications/PushNotificationChannel.cpp
sharath2727: Yeah I am aware of this. I have taken a note. Will update the PR :)

dev/PushNotifications/PushNotificationChannel.h
sharath2727: Yeah I created one: https://github.com/microsoft/WindowsAppSDK/issues/1464

dev/PushNotifications/PushNotificationChannel.cpp
sharath2727: Actually m_channelInfo is a struct static allocation object and not a pointer object. So checking on the channelUri string should be the right thing to do?

test/TestApps/PushNotificationsDemoApp/main.cpp
sharath2727: Planning to change the GUID as the newer guid works for both scenarios - packaged and unpackaged.

dev/PushNotifications/PushNotificationChannel.cpp
hulumane: Check out std::variant here. https://en.cppreference.com/w/cpp/utility/variant. You could technically use this and check if it is invalid or not?

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -89,15 +88,35 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                auto lock = m_lock.lock_exclusive();
+
+                THROW_HR_IF_NULL(E_UNEXPECTED, m_channelInfo.channelUri.c_str());
+
+                wil::unique_cotaskmem_string appUserModelId;
+                THROW_IF_FAILED(GetAppUserModelId(appUserModelId));
+
+                THROW_IF_FAILED(PushNotifications_RegisterNotificationSinkForFullTrustApplication(appUserModelId.get(), this));
+
+                winrt::event_token token = m_foregroundHandlers.add(handler);
```
hulumane, line 114: No need of having a token local variable here. Just return this at the end.

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -89,15 +88,35 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+            else
+            {
+                auto lock = m_lock.lock_exclusive();
+
+                THROW_HR_IF_NULL(E_UNEXPECTED, m_channelInfo.channelUri.c_str());
+
+                wil::unique_cotaskmem_string appUserModelId;
+                THROW_IF_FAILED(GetAppUserModelId(appUserModelId));
+
+                THROW_IF_FAILED(PushNotifications_RegisterNotificationSinkForFullTrustApplication(appUserModelId.get(), this));
```
hulumane, line 112: You only want to register the foreground sink if the foregroundHandlerCount is 0. Otherwise, we keep the existing sink.

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -120,7 +139,23 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                m_channel.PushNotificationReceived(token);
+            }
+            else
+            {
+                auto lock = m_lock.lock_exclusive();
+
+                wil::unique_cotaskmem_string appUserModelId;
+                THROW_IF_FAILED(GetAppUserModelId(appUserModelId));
+
+                THROW_IF_FAILED(PushNotifications_UnregisterNotificationSinkForFullTrustApplication(appUserModelId.get()));
```
hulumane, line 153: We only unregister the sink if the foregroundHandlerCount == 0

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -148,6 +183,44 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
     }
     CATCH_RETURN()
 
+    HRESULT __stdcall PushNotificationChannel::OnRawNotificationReceived(unsigned int payloadLength, _In_ byte* payload, _In_ HSTRING /*correlationVector */) noexcept
+    {
+        try
```
hulumane, line 188: try catch. Don't really understand this flow at all.

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -148,6 +183,44 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        {
+            BOOL foregroundHandled = false;
+            THROW_IF_FAILED(InvokeAll(payloadLength, payload, &foregroundHandled));
+            THROW_HR(E_INVALIDARG);
+        }
+        catch(...)
+        {
+            //TODO: Capture the HRESULT in telemetry
+            winrt::com_array<uint8_t> payloadArray{ payload, payload + (payloadLength * sizeof(uint8_t)) };
```
hulumane, line 197: This is all shared code between the LRP and WinSDK. So it should be defined in a common header in a utility function.

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -148,6 +183,44 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+            std::string commandLine = "----WindowsAppRuntimePushServer:-Payload:\"";
+            commandLine.append(reinterpret_cast<char*>(payload), payloadLength);
+            commandLine.append("\"");
+
+            wil::unique_cotaskmem_string processName;
+            THROW_IF_FAILED(GetCurrentProcessPath(processName));
+
+            const std::string processNameAsUtf8String = ConvertProcessNameToUtf8String(processName.get());
```
hulumane, line 205: Did you look into @danielayala94 's code change for special characters and protocol launch? I think we are introducing the same bug here which he fixed. Hence its important to share code through utility helpers so that we don't redo the same mistakes.

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -148,6 +183,44 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
     }
     CATCH_RETURN()
 
+    HRESULT __stdcall PushNotificationChannel::OnRawNotificationReceived(unsigned int payloadLength, _In_ byte* payload, _In_ HSTRING /*correlationVector */) noexcept
+    {
+        try
+        {
+            BOOL foregroundHandled = false;
+            THROW_IF_FAILED(InvokeAll(payloadLength, payload, &foregroundHandled));
```
hulumane, line 191: foregroundHandled should be used along with channelInfo != nullptr to determine protocol launch

dev/PushNotifications/externs.h
```diff
@@ -21,3 +21,26 @@ inline HRESULT GetCurrentProcessPath(wil::unique_cotaskmem_string& processName)
...
+    UINT32 appIdSize{ ARRAYSIZE(appId) };
+
+    GetCurrentApplicationUserModelId(&appIdSize, appId);
+
+    appUserModelId = wil::make_unique_string<wil::unique_cotaskmem_string>(appId);
+
+    return S_OK;
+}
+
+inline const std::string ConvertProcessNameToUtf8String(std::wstring processName)
```
hulumane, line 36: @danielayala94 Already introduced a function for this. We should not copy paste but introduce Utility Helpers with shared common code.

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -148,6 +183,44 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+
+            const std::string processNameAsUtf8String = ConvertProcessNameToUtf8String(processName.get());
+
+            SHELLEXECUTEINFOA shellExecuteInfo{};
+            shellExecuteInfo.cbSize = sizeof(SHELLEXECUTEINFOA);
+            shellExecuteInfo.fMask = SEE_MASK_NOCLOSEPROCESS | SEE_MASK_DOENVSUBST;
+            shellExecuteInfo.lpFile = processNameAsUtf8String.c_str();
+            shellExecuteInfo.lpParameters = commandLine.c_str();
+
+            shellExecuteInfo.nShow = SW_NORMAL;
```
hulumane, line 213: Did you test the case where the foregroundHandled = false? We should ensure that a new background process is launched with the payload.

dev/PushNotifications/PushNotificationChannel.cpp
hulumane: BTW c_str() can never be null for a string. It can only be empty.

dev/PushNotifications/PushNotificationChannel.cpp
sharath2727: Agreed. I totally missed that. I have integrated that change into my code and validated it. I have created a helper for the same. Thanks.

dev/PushNotifications/PushNotificationChannel.cpp
sharath2727: question here: should we not always register the latest sink? @hulumane.

dev/PushNotifications/PushNotificationChannel.cpp
sharath2727: For now I have moved this code into externs.h which the files in PushNotifications/ are using. I need to see how I can make both LRP and WinSDK use the same code when daniel's changes and this PR merges into WNP. It makes my life easier then. I will take this as a bug: https://github.com/microsoft/WindowsAppSDK/issues/1481

Thanks.

dev/PushNotifications/externs.h
sharath2727: Currently moved all this into externs.h. Will create a utility to be used between LRP and PushNotifications. bug: https://github.com/microsoft/WindowsAppSDK/issues/1481

dev/PushNotifications/PushNotificationChannel.cpp
hulumane: You misunderstand the design for foreground sink. The way @pmpurifoy  implemented this is using a single sink registration which is refcounted for every eventHandler that the developer subscribes too. 

Another way to achieve the same result is to AddRef the COM ref instead of having a foregroundHandlerCount (we can delete this). That might actually be more optimal.

sharath2727: https://github.com/microsoft/WindowsAppSDK/issues/1489 - created a bug for this as this is an improvement and will not impact the current design. Thanks.

dev/PushNotifications/PushNotificationChannel.h
sharath2727: https://github.com/microsoft/WindowsAppSDK/issues/1490

dev/PushNotifications/PushNotificationUtility.cpp
```diff
@@ -0,0 +1,59 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
danielayala94, line 1: Should the utility implementation be in the header file? I would imagine that this file has to be published to be consumed by the LRP. But also, if the LRP compiles before the SDK, wouldn't a compile error happen since we were looking for the utility file not having published yet?

Would dev/Common be the right place for this code?

dev/PushNotifications/PushNotificationUtility.h
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#pragma once
+#include "pch.h"
+
+const std::wstring ConvertByteArrayToWideString(unsigned int payloadLength, byte* payload) noexcept;
```
danielayala94, line 7: Nit:

```suggestion
const std::wstring ConvertByteArrayToWideString(unsigned int payloadLength, _In_reads_(payloadLength) byte* payload) noexcept;
```

dev/PushNotifications/PushNotificationUtility.cpp
```diff
@@ -0,0 +1,59 @@
...
+        reinterpret_cast<PCSTR>(payload),
+        payloadLength,
+        &payloadAsWideString[0],
+        size);
+    THROW_LAST_ERROR_IF(size == 0);
+
+    return payloadAsWideString;
+}
+
+void ProtocolLaunchHelper(unsigned int payloadLength, _In_ byte* payload) noexcept
```
danielayala94, line 31: This can't be noexcept. For instance, you are dealing with STL strings.

dev/PushNotifications/PushNotificationUtility.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+    std::wstring payloadAsWideString(size, 0);
+    size = MultiByteToWideChar(
+        CP_UTF8,
+        0,
+        reinterpret_cast<PCSTR>(payload),
+        payloadLength,
+        &payloadAsWideString[0],
+        size);
+    THROW_LAST_ERROR_IF(size == 0);
```
danielayala94, line 26: THROW but noexcept?

danielayala94 (changes requested):

dev/PushNotifications/PushNotificationUtility.cpp
sharath2727: Create a common utility file between LRP and PushNotifications #1481 - That is the reason I have created this bug. Temporarily I have created the files PushNotificationUtility.h/cpp in dev/PushNotifications. As responses might be little longer it should not block the PR from merging atleast in WNP_LRP. By then once we know where these files should move to we can address it as part of a bug. Let me know if you think otherwise. Thanks.

sharath2727: @danielayala94 - realized that PushNotifications.vcxitems is a shared folder. So we don't need to create a new common folder and with the suggestions from Howard, we should be able to call into PushNotificationUtility.h/cpp from anywhere within LRP and beyond. Thanks.

dev/PushNotifications/PushNotificationUtility.h
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#pragma once
+#include "pch.h"
+
+const std::wstring ConvertByteArrayToWideString(unsigned int payloadLength, _In_reads_(payloadLength) byte* payload);
+void ProtocolLaunchHelper(unsigned int payloadLength, _In_reads_(payloadLength) byte* payload);
+
+inline HRESULT GetAppUserModelId(wil::unique_cotaskmem_string& appUserModelId)
```
jonwis, line 10: Consider having this method return the `wil::unique_cotaskmem_string` instead.

dev/PushNotifications/PushNotificationUtility.h
```diff
@@ -0,0 +1,20 @@
...
+{
+    wchar_t appId[APPLICATION_USER_MODEL_ID_MAX_LENGTH] = {};
+    UINT32 appIdSize{ ARRAYSIZE(appId) };
+
+    GetCurrentApplicationUserModelId(&appIdSize, appId);
```
jonwis, line 15: Presumably this should check for the one "no package identity" error and return null, then throw otherwise?

jonwis (changes requested):
Two things to change, the rest are optional - use of that "count" thing to track number of registered handlers, and the missing function-level try/catch on the noexcept method. Otherwise looks pretty good, glad to see this coming together.

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -89,15 +89,32 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                        pushArgs.Handled(true);
+                        handler(*strong, pushArgs);
+                    };
+                });
+            }
+            else
+            {
+                auto lock = m_lock.lock_exclusive();
+
+                if (!m_foregroundHandlerCount++)
```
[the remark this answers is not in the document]
jonwis: ... or if that's too hard, put the `Count++` _after_ you've registered and added the handler.

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -120,7 +137,24 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-            m_channel.PushNotificationReceived(token);
+            if (m_channel)
+            {
+                m_channel.PushNotificationReceived(token);
+            }
+            else
+            {
+                auto lock = m_lock.lock_exclusive();
+
+                if (!--m_foregroundHandlerCount)
```
robertzhou-wpn, line 148: Looks like the counter here will have the similar problem as Jon mentioned in the counter adding case. can we move the count operation after the operations?

dev/PushNotifications/PushNotificationUtility.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+    return payloadAsWideString;
+}
+
+void ProtocolLaunchHelper(unsigned int payloadLength, _In_reads_(payloadLength) byte* payload)
+{
+    // Command line format: ----WindowsAppRuntimePushServer:-Payload:"<payloadAsEscapedUriFormat>"
+    std::wstring commandLine = L"----WindowsAppRuntimePushServer:-Payload:\"";
```
hulumane, line 34: This is common code pasted from the LRP. Can we use a utility helper method here?

dev/PushNotifications/PushNotificationUtility.h
```diff
@@ -0,0 +1,20 @@
...
+
+    appUserModelId = wil::make_unique_string<wil::unique_cotaskmem_string>(appId);
```
hulumane, line 17: I believe this throws for the no memory case. But your helper method doesn't handle that.

dev/PushNotifications/PushNotificationUtility.h
hulumane, line 15: If this API call throws, then the caller method should wrap a try catch_return. If the API returns HRESULT, then we need a RETURN_IF_FAILED here or a THROW_IF_FAILED with a method level try-catch_return()

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -89,15 +89,32 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                {
+                    if (auto strong = weak_self.get())
+                    {
+                        auto pushArgs = winrt::make<winrt::Microsoft::Windows::PushNotifications::implementation::PushNotificationReceivedEventArgs>(args);
+                        pushArgs.Handled(true);
+                        handler(*strong, pushArgs);
+                    };
+                });
+            }
+            else
```
hulumane, line 104: Comment needed here that this else is specifically for the RemoteId case when we directly call into the UDK from the SDK to requestChannel from RemoteId

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -148,6 +182,19 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
     }
     CATCH_RETURN()
 
+    HRESULT __stdcall PushNotificationChannel::OnRawNotificationReceived(unsigned int payloadLength, _In_ byte* payload, _In_ HSTRING /*correlationVector */) noexcept
+    {
+        BOOL foregroundHandled = false;
```
hulumane, line 187: Please set this to true instead. Default behavior should be that if the dev doesn't set this to true, then we don't do protocol launch.

dev/PushNotifications/PushNotificationChannel.h
```diff
@@ -29,6 +30,9 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
         // IWpnForegroundSink
         HRESULT __stdcall InvokeAll(_In_ ULONG length, _In_ byte* payload, _Out_ BOOL* foregroundHandled) noexcept;
 
+        //INotificationHandler
```
hulumane, line 33: Nit: Missing space
// INotificationHandler

dev/PushNotifications/PushNotificationUtility.h
hulumane: I thibk this is missing a THROW_IF_FAILED @sharath2727 ?

jonwis: It's OK to have it return null, but the caller has to know that through a comment in the header - "method returns nullptr when the package does not have identity."

sharath2727: Sure. I think throwing on the error codes: https://docs.microsoft.com/en-us/windows/win32/api/appmodel/nf-appmodel-getcurrentapplicationusermodelid should take care of this.

dev/PushNotifications/PushNotificationUtility.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+    // Escape special characters to follow command line standards for any app activation type in AppLifecycle
+    // (See AppInstance.cpp and Serialize() from other activation types)
+    std::wstring payloadAsWideString = ConvertByteArrayToWideString(payloadLength, payload);
```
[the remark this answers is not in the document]
aeloros: I can't remember if we talked about this before or not,  but I believe we would be better served not creating some new payload model.  Instead we should create a contractId for the push server, and use the already existing infrastructure to pass properties through the query string.  This might be an example of a fragile design since it requires continued upkeep to make it special.

loneursid:
So, I've messed up. I've acidentaly merged this branch into WNP_LRP. As a result GIT closed this PR and the branch was deleted. I've reverted the merge from WNP_LRP and restored the branch. Still trying to figure out how to re-open this PR.

jonwis (commented):
Hmm

dev/PushNotifications/PushNotificationUtility.cpp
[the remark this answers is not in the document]
sharath2727: @jonwis /@aeloros - I see howard introducing a lot of string conversion APIs in dev/common: https://github.com/microsoft/WindowsAppSDK/pull/1513. Once the PR merges some of these conversion APIs in my .h/.cpp will be removed. 

@aeloros - "Instead we should create a contractId for the push server, and use the already existing infrastructure to pass properties through the query string. This might be an example of a fragile design since it requires continued upkeep to make it special" -  can you please help me understand this further? Maybe I could work on this going ahead to simplify things here.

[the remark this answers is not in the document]
aeloros: Sure, the command-line on the receiving end (when GetActivatedEventArgs() parses it) already has a format that supports passing payload information.  Specifically I'm talking about the ability to pass a Uri that contains a key/value pair (the query string).

There are problems with the current implementation of GetActivatedEventArgs() in that it doesn't tolerate malformed command-lines well.  I have a change out to fix that, but it changes the expectations of the format as well.  

I'm mostly suggesting here that we should try to find a better way to utilize the command-line.  One that doesn't have the calling developer designing custom formats to support payloads.  We should instead find a way to utilize the Uri format and serialize + encode data into that format.  

This also lends itself to creating shared functions in the code base that generate that command-line.  I've started to do that in a super simple way with GenerateCommandLine() inside of ActivationRegistrationManager.cpp.  It could be moved and enhanced or accompanied with other helper functions to make a useful reusable toolset for callers to create command-lines in a consistent way.

I don't think any of this is the end of the world if not done.  At this point in the conversation though, it probably would be intentional technical debt.  Obviously we should avoid that when possible, but dealing with technical debt isn't the only factor in our lives. :)

PR #1455, opened 2021-09-20
Fix publishing of Deployment API winmd and C# projections

dkbennett:
Adds publishing of the Deployment API winmd into the nuget, along with adding and publishing the C# projections.


DrusTheAxe (approved):
:-)

PR #1462, opened 2021-09-21
For MRT Core, exclude file names and file paths from logging as they contain PII.

rohanp-msft:
I verified that the change builds fine.
There is an issue where none of the log data (HR values, etc.) are showing up in the ETL logs... Outside of that, I confirmed the events are logged. Proceeding with check-in. Will follow up on the issue separately.

### DIFF
--- a/dev/MRTCore/mrt/mrm/include/mrm/common/MrmTraceLogging.h
+++ b/dev/MRTCore/mrt/mrm/include/mrm/common/MrmTraceLogging.h
@@ -32,7 +32,7 @@ class MrtRuntimeTraceLoggingProvider : public wil::TraceLoggingProvider
 };
 
 // Remove PII from string
-__inline PCWSTR RemovePiiFromString(PCWSTR)
+inline PCWSTR RemovePiiUserProfileFilename(PCWSTR)
 {
     // TODO: Once the feature below is available, this function can be removed.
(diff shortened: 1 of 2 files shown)


dev/MRTCore/mrt/mrm/include/mrm/common/MrmTraceLogging.h
```diff
@@ -31,22 +31,10 @@ class MrtRuntimeTraceLoggingProvider : public wil::TraceLoggingProvider
     DEFINE_COMPLIANT_MEASURES_EVENT_PARAM4(MeasureGenericEvent, PDT_ProductAndServicePerformance, PCWSTR, functionName, PCWSTR, message1, PCWSTR, message2, int, hresult);
 };
 
-// case insensitive prefix match of "*:\users"
-__inline PCWSTR RemoveUsernameFromPath(_In_ PCWSTR filePath)
+// Remove PII from string
+__inline PCWSTR RemovePiiFromString(PCWSTR)
```
DrusTheAxe, line 35: `inline` ?

DrusTheAxe, line 35: Rather generic name. `RemovePiiUserProfileFilename` would be more accurate

DrusTheAxe (approved):
Minor comments

jefgen (approved):
:shipit:

PR #1468, opened 2021-09-22
Reduce Microsoft.Windows.ApplicationModel.Resources.dll binary size by 70KB by disabling RTTI.

jefgen:
MRT Core doesn't use [RTTI](https://en.wikipedia.org/wiki/Run-time_type_information), but the `*.vcxproj` file that is used for building `Microsoft.Windows.ApplicationModel.Resources.dll` doesn't explicitly disable it -- and it is enabled by default. We can turn it off, which saves ~70 KB.

| File    | Size Before     | Size After     |
|---------|----------------:|---------------:|
| Microsoft.Windows.ApplicationModel.Resources.dll | 178,688 | 107,520 |

**Delta** = **-71,168 bytes (69.5 KB)**

Note: I already disabled it for `MRM.dll` in the PR here: https://github.com/microsoft/WindowsAppSDK/pull/904, which saved 37 KB.


huichen123 (approved):

rohanp-msft (approved):
Big improvement! Thanks!

PR #1471, opened 2021-09-22
Various fixes for the project templates

alwu-msft:
A number of improvements to our project templates are contained herein:
* Remove `#include <winrt/microsoft.ui.dispatching.co_await.h>` from all C++ templates' `pch.h` files since that header file no longer exists
* Rename `Images/` directory to `Assets/` in the project templates that use Single-project MSIX Packaging in order to match the convention employed by the UWP XAML project templates
* Update `launchSettings.json` in the C# Single-project MSIX Packaging project template so that the launch profile uses the correct command name
* Actually promote the C++ Single-project MSIX Packaging project template from Experimental
* The C++ Single-project MSIX Packaging project template needs to explicitly set the `AppContainerApplication` property as it's needed by the C++ project system's `.props` files but the `WindowsPackageType` property its default value is derived from  doesn't have its default value calculated until our own `.targets`.


dev/VSIX/ProjectTemplates/Desktop/CppWinRT/PackagedApp/BlankApp/pch.h
```diff
@@ -21,5 +21,4 @@
 #include <winrt/Microsoft.UI.Xaml.Media.h>
 #include <winrt/Microsoft.UI.Xaml.Navigation.h>
 #include <winrt/Microsoft.UI.Xaml.Shapes.h>
-#include <winrt/Microsoft.UI.Dispatching.h>
-#include <winrt/microsoft.ui.dispatching.co_await.h>
```
BenJKuhn, line 25: Consider commenting this out and leaving in with a comment as to where to get it, so there's a breadcrumb for folks that want to add this support.

BenJKuhn (approved):

DefaultRyan:
> * Remove `#include <winrt/Microsoft.UI.Dispatching.h>` from all C++ templates' `pch.h` files since that header file no longer exists

It appears as though this file still exists, and you removed `#include <winrt/microsoft.ui.dispatching.co_await.h>` instead. That's correct, right?

alwu-msft:
> > * Remove `#include <winrt/Microsoft.UI.Dispatching.h>` from all C++ templates' `pch.h` files since that header file no longer exists
> 
> It appears as though this file still exists, and you removed `#include <winrt/microsoft.ui.dispatching.co_await.h>` instead. That's correct, right?

Oh, good catch! I copy/pasted the wrong line into the description.

dev/VSIX/ProjectTemplates/Desktop/CppWinRT/PackagedApp/BlankApp/pch.h
alwu-msft: I dropped the relnote on async support into the `readme.txt` that is included with newly created C++/WinRT projects.

dotMorten:
@evelynwu-msft This PR says to use WIL and its resume_foreground method instead for co_await support. Not that I'm looking forward to needed even more dependencies just to switch to a UI thread, but WIL has [removed those methods for DispatcherQueue](https://github.com/microsoft/wil/pull/199) so there doesn't seem to be any alternative right now?

evelynwu-msft:
@dotMorten I'm not familiar with what's going on with WIL (my primary goal was to remove the build break from our templates), but @patrickeganMSFT is working on #1501 to add WIL as a Nuget package reference to our templates and so may be able to address your concerns.

patrickeganMSFT:
@dotMorten The `winrt::resume_foreground` coroutine was replaced with `wil::resume_foreground` defined in [wil/cppwinrt.h](https://github.com/oldnewthing/wil/blob/01ae067f7d5a5ae05134f1a433c206e4aa381acb/include/wil/cppwinrt.h#L476). #1501 adds a reference to the WIL nuget to the C++ project templates so users can use the `resume_foreground` without having to go through the trouble of manually referencing the WIL nuget and its headers themselves.

dotMorten:
@patrickeganMSFT Yes I'm aware, but the overload that takes a `winrt::Microsoft::UI::Dispatching::DispatcherQueue` isn't in the current master branch (it was just removed), or in the latest nuget package.
I had to add this to my own project, to get it to compile:
```c++
template<>
struct wil::details::dispatcher_traits<winrt::Microsoft::UI::Dispatching::DispatcherQueue>
{
    using Priority = winrt::Microsoft::UI::Dispatching::DispatcherQueuePriority;
    using Handler = winrt::Microsoft::UI::Dispatching::DispatcherQueueHandler;
    using Scheduler = dispatcher_TryEnqueue;
};
```

dotMorten:
> #1501 adds a reference to the WIL nuget to the C++ project templates so users can use the resume_foreground without having to go through the trouble of manually referencing the WIL nuget

It is still yet another reference to deal with and mantain, and nuget in C++ is already a bad experience as it is. It feels like a step back to remove out of the box functionality like this (functionality that I would consider basic in a WinUI app).

patrickeganMSFT:
`winrt::Windows::System::DispatcherQueue` is declared in the master and the most recent nuget. Other than namespace, what is the difference between that and `winrt::Microsoft::UI::Dispatching::DispatcherQueue`?

Excluding resume_foreground is a hole so it should be added to the templates. Are there changes that need to be made to get resume_foreground into the C++ project templates, other than including the WIL nuget and #include?

dotMorten:
>  Other than namespace, what is the difference between that and winrt::Microsoft::UI::Dispatching::DispatcherQueue?

The main difference is that is the DispatcherQueue type you get back from the UIElement's DispatcherQueue property, so that's what we need :-D

patrickeganMSFT:
I see, is it possible that this was a mistake in WIL? It seems odd that we'd be removing `winrt::Microsoft::UI::Dispatching::DispatcherQueue` from WIL when XAML is using it.

dotMorten:
No idea. Perhaps @dunhor can shed some light on why this was removed. Having said that, I still don't get why this was removed from the out-of-the-box libraries. Shouldn't the included WinUI headers be adding this instead?

dunhor:
Please see https://github.com/microsoft/wil/pull/199#issuecomment-928655918

PR #1473, opened 2021-09-22
Fix Bootstrap.net assembly support

DrusTheAxe:
Fixed the Bootstrap.Net dll location so VS automagically picks it up in C# projects so references work as expected.

Remove the copy from the .props since VS will do it correctly for us.

Tested with VS projects to nail the working solution.


AdamBraden:
lgtm

AdamBraden (approved):

PR #1482, opened 2021-09-23
Remove VSEXT from Component VSIX

DefaultRyan:
This change removes the single-project MSIX extension pack from the Component flavor of the VSIX. The two Windows App SDK VSIXs and the single-project VSIX are going to be handled as individual components in that scenario.

Also updated the VS SDK references to the latest prerelease.

### DIFF
--- a/dev/VSIX/Extension/Cpp/Dev17/WindowsAppSDK.Cpp.Extension.Dev17.csproj
+++ b/dev/VSIX/Extension/Cpp/Dev17/WindowsAppSDK.Cpp.Extension.Dev17.csproj
@@ -49,7 +49,7 @@
     <Content Include="$(RepoRoot)LICENSE">
       <IncludeInVSIX>true</IncludeInVSIX>
     </Content>
-    <Content Include="ExtensionPack.vsext">
+    <Content Condition="'$(Deployment)'=='Standalone'" Include="ExtensionPack.vsext">
       <IncludeInVSIX>true</IncludeInVSIX>
     </Content>
     </ItemGroup>
@@ -57,11 +57,11 @@
     <Reference Include="System" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.VisualStudio.SDK" Version="17.0.0-previews-1-31325-273" ExcludeAssets="runtime">
+    <PackageReference Include="Microsoft.VisualStudio.SDK" Version="17.0.0-previews-4-31709-430" ExcludeAssets="runtime">
       <IncludeAssets>compile; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
     <PackageReference Include="System.Net.Http" Version="4.3.4" />
-    <PackageReference Include="Microsoft.VSSDK.BuildTools" Version="17.0.3177-preview3">
+    <PackageReference Include="Microsoft.VSSDK.BuildTools" Version="17.0.4207-preview4">
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
--- a/dev/VSIX/Extension/Cpp/Dev17/WindowsAppSDK.Cpp.Extension.Dev17.csproj
+++ b/dev/VSIX/Extension/Cpp/Dev17/WindowsAppSDK.Cpp.Extension.Dev17.csproj
@@ -37,7 +37,6 @@
     <ContentNugetPackages Include="$(PkgMicrosoft_Windows_SDK_BuildTools)\*.nupkg" />
     <ContentNugetPackages Include="$(PkgMicrosoft_Windows_CppWinRT)\*.nupkg" />
     <ContentNugetPackages Include="$(PkgMicrosoft_WindowsAppSDK)\*.nupkg" />
-    <ContentNugetPackages Include="$(PkgMicrosoft_Windows_SDK_BuildTools)\*.nupkg" />
     <Content Include="@(ContentNugetPackages)">
       <IncludeInVSIX>true</IncludeInVSIX>
       <VSIXSubPath>Packages</VSIXSubPath>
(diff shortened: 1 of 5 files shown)


dev/VSIX/Extension/Cpp/Dev17/WindowsAppSDK.Cpp.Extension.Dev17.csproj
```diff
@@ -49,19 +48,19 @@
...
+    <Content Condition="'$(Deployment)'=='Standalone'" Include="ExtensionPack.vsext">
       <IncludeInVSIX>true</IncludeInVSIX>
     </Content>
     </ItemGroup>
   <ItemGroup>
     <Reference Include="System" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.VisualStudio.SDK" Version="17.0.0-previews-1-31325-273" ExcludeAssets="runtime">
+    <PackageReference Include="Microsoft.VisualStudio.SDK" Version="17.0.0-previews-4-31709-430" ExcludeAssets="runtime">
```
rohanp-msft, line 59: >Version="17.0.0-previews-4-31709-430" [](http://example.com/codeflow?start=59&length=37)

The two version updates in this change, we don't want them made in WindowsAppSDK.Cpp.Extension.Dev17.csproj and WindowsAppSDK.Cs.Extension.Dev17.csproj?

DefaultRyan: Did you mean the Dev16 projects? We're not updating them because they are targeting Dev16, and so they're using the Dev16 VS SDK.

rohanp-msft: Shoot. Yeah, I meant Dev16.
OK, got it.

dev/VSIX/Extension/Original/WindowsAppSDK.Extension.csproj
```diff
@@ -54,11 +54,11 @@
     <Reference Include="System" />
   </ItemGroup>
   <ItemGroup>
```
rohanp-msft, line 56: Just going off of names...:

We create these VSIXs:
WindowsAppSDK.Cpp.Extension.Dev16.Experimental.vsix
WindowsAppSDK.Cpp.Extension.Dev17.Experimental.Component.vsix
WindowsAppSDK.Cpp.Extension.Dev17.Experimental.Standalone.vsix
WindowsAppSDK.Cs.Extension.Dev16.Experimental.vsix
WindowsAppSDK.Cs.Extension.Dev17.Experimental.Component.vsix
WindowsAppSDK.Cs.Extension.Dev17.Experimental.Standalone.vsix
WindowsAppSDK.Extension.Experimental.vsix

Based on those names and the names of the 5 CSPROJ files  under dev\vsix\Extension, I suspect this file corresponds to WindowsAppSDK.Extension.Experimental.vsix. Assuming I'm right, given that WindowsAppSDK.Extension.Experimental.vsix targets both Dev16 and 17, should the version updates in this file be done? The version updates are only wanted for Dev16, right?

DefaultRyan: Technically, the WindowsAppSDK.Extension.Experimental is about to be removed. It's only there as a placeholder until we release 1.0 to the marketplace and replace it with C++ and C# specific flavors.

dev/VSIX/Extension/Cs/Dev17/WindowsAppSDK.Cs.Extension.Dev17.csproj
```diff
@@ -48,19 +47,19 @@
     <Content Include="$(RepoRoot)LICENSE">
       <IncludeInVSIX>true</IncludeInVSIX>
     </Content>
-    <Content Include="ExtensionPack.vsext">
+    <Content Condition="'$(Deployment)'=='Standalone'" Include="ExtensionPack.vsext">
       <IncludeInVSIX>true</IncludeInVSIX>
     </Content>
```
rohanp-msft, line 52: Are we OK with this extension being included in both the Component and Standalone cases in the Dev16 cases?

DefaultRyan: We don't build a Component extension for Dev16, and it's very unlikely we ever will. But it's simple enough to add for future-proofing.

dev/VSIX/Extension/Original/WindowsAppSDK.Extension.csproj
rohanp-msft: Got it. Still, better to leave it as-is so we don't break the VS 2022 scenario, no? I say that because devs like me may use the WindowsAppSDK.Extension.Experimental.vsix on VS2022 without knowing.

DefaultRyan: Also, the version-less extension targets both Dev16 and Dev17. And in order to target Dev17 at all, it needs to use the Dev17 version of the VS SDK.

rohanp-msft: I made another mistake... My comment above should have read:
"Got it. Still, better to leave it as-is so we don't break the VS 2019 scenario, no? I say that because devs like me may use the WindowsAppSDK.Extension.Experimental.vsix on VS2019 without knowing."

> in order to target Dev17 at all, it needs to use the Dev17 version of the VS SDK.

Got it. What will happen in the Dev16 case?

DefaultRyan: It should continue to work, as-is, because the versionless extension doesn't have any localization or code in it.

I'd recommend not using it though, because it's not an accurate representation of what our customers will be using. Again, the *only* reason I haven't yet removed this project, is because we'll probably want to publish one last update to "deprecate" it and point our customers towards the new replacement extensions. As soon as we do that, I'm deleting it.

rohanp-msft: OK, thanks for explaining.

rohanp-msft (approved):
:shipit:

alwu-msft (approved):
:shipit:

PR #1483, opened 2021-09-23
Update Deployment API Implementation

dkbennett:
Updates the Deployment API Implementation to the latest spec update.
#1253


dev/Deployment/Deployment.idl
```diff
@@ -1,46 +1,40 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
-namespace Microsoft.Windows.ApplicationModel.WindowsAppSDK
+namespace Microsoft.Windows.ApplicationModel.WindowsAppRuntime
 {
     /// Represents the current Deployment status of the WindowsAppRuntime
+    [experimental][flags]
```
jonwis, line 7: Why is this 'flags'? If it's flags, the values need to be single-bit values - 0x1, 0x2, 0x4, etc. I think instead this is just a linear list, and the 'flags' is not required.

dev/Deployment/DeploymentManager.cpp
```diff
@@ -59,30 +66,43 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppSDK::implementa
...
-        return status;
-    }
+        DeploymentStatus status{};
+        if (SUCCEEDED(verifyResult))
+        {
+            status = DeploymentStatus::Ok;
+        }
+        else
+        {
+            status = DeploymentStatus::PackageInstallRequired;
```
jonwis, line 76: Feels like this should be a more precise mapping? Are all "fails" implying install required? Or can a fail like "no internet" be made more precise? Probably OK for now.

dev/Deployment/DeploymentManager.cpp
```diff
@@ -59,30 +66,43 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppSDK::implementa
...
+        {
+            status = DeploymentStatus::Ok;
+        }
+        else
+        {
+            status = DeploymentStatus::PackageInstallFailed;
+        }
+
+        auto initializeResult{ winrt::make<implementation::DeploymentResult>(status, deployPackagesResult) };
+        return initializeResult;
```
jonwis, line 105: Nit: Consider just `return winrt::make<...>(...)`

dev/Deployment/DeploymentManager.h
```diff
@@ -3,18 +3,18 @@
...
     {
         DeploymentManager() = default;
-        static winrt::Microsoft::Windows::ApplicationModel::WindowsAppSDK::DeploymentStatus GetStatus();
-        static winrt::Microsoft::Windows::ApplicationModel::WindowsAppSDK::DeploymentStatus GetStatus(hstring const& packageFullName);
-        static winrt::Microsoft::Windows::ApplicationModel::WindowsAppSDK::DeploymentStatus Initialize();
-        static winrt::Microsoft::Windows::ApplicationModel::WindowsAppSDK::DeploymentStatus Initialize(hstring const& packageFullName);
+        static winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::DeploymentResult GetStatus();
+        static winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::DeploymentResult Initialize();
     private:
+        static winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::DeploymentResult GetStatus(hstring const& packageFullName);
```
jonwis, line 16: Note - to reference the metadata type here, just say the last part of the namespace:

```c++
static WindowsAppRuntime::DeploymentResult GetStatus(hstring const& packageFullName);
static WindowsAppRuntime::DeploymentResult Initialize(hstring const& packageFullName);
```

jonwis (approved):
Looks ok, nothing glaring. A few flavor/style notes.

dev/Deployment/DeploymentManager.cpp
dkbennett: Fixed

dev/Deployment/DeploymentManager.cpp
dkbennett: For checking status currently, yes a failure to find the registered package implies a package install is required. I expect we will need additional status types later on, such as PackageInstallPending or PackageInstallInProgress and similar results if the package is missing and we are in the process of installing it (such as a web request).

dev/Deployment/Deployment.idl
dkbennett: I was considering that there could be concurrent states, such as PackageInstallRequired and PackageInstallFailed or PackageInstallPending and have more specific states indicated later on similar to PackageStatus. But we can also infer that - if the status is PackageInstallFailed then it should be obvious that a package install is required, and same with Pending or other states. So I removed flags and kept as is.

PR #1485, opened 2021-09-24
Remove Experimental-ness from Deployment API

dkbennett:
Removes [experimental] attributes from the WindowsAppRuntime deployment API and sets the feature to being always available.


DrusTheAxe (approved):

jonwis (approved):
Super exciting!

PR #1500, opened 2021-09-28
Add source link to product code

DefaultRyan:
Adding source link allows us to see and download source code when debugging (by inserting some metadata into the debugger about how to retrieve the file from source control). This is taking the same approach as #1066 and applying it to the rest of the product code in the repo.

Verified by:
- Running a pipeline build
- Downloaded some build artifacts (e.g. DynamicDependency.DataStore.exe)
- Loaded binary into windbg
- Examine a symbol in the binary (e.g. DynamicDependency_DataStore!WinMain)
- Verify that the debugger now has source code info

With this change, I now see the following:
```
0:000> ln DynamicDependency_DataStore!WinMain
Browse module
Set bu breakpoint

 [D:\a\_work\1\s\dev\DynamicDependencyDataStore\DynamicDependency.DataStore\winmain.cpp @ 58] SrcSrv Command: https://raw.githubusercontent.com/microsoft/WindowsAppSDK/4d33f4785b0e412be9b9e86cdf24105153103613/dev/DynamicDependencyDataStore/DynamicDependency.DataStore/winmain.cpp
(00403f10)   DynamicDependency_DataStore!WinMain   |  (00404090)   DynamicDependency_DataStore!Microsoft::WRL::Module<2,Microsoft::WRL::Details::DefaultModule<2> >::DecrementObjectCount
Exact matches:
    DynamicDependency_DataStore!WinMain (struct HINSTANCE__ *, struct HINSTANCE__ *, char *, int)
```
Notice `winmain.cpp @ 58`. In my windbg, this is a hyperlink that I can click, or I can manually open:
```
windbg> .open -a 403f10
```

And I get the source code in my debugger:
```
int WINAPI WinMain(HINSTANCE /*hInstance*/, HINSTANCE /*hPrevInstance*/, PSTR /*lpCmdLine*/, int /*nCmdShow*/)
{
    RETURN_IF_FAILED(::CoInitializeEx(nullptr, COINITBASE_MULTITHREADED));

    wil::unique_event endOfTheLine(::CreateEventW(nullptr, TRUE, FALSE, nullptr));
...
```

### DIFF
--- a/dev/Detours/packages.config
+++ b/dev/Detours/packages.config
@@ -3,5 +3,4 @@
   <package id="Microsoft.Build.Tasks.Git" version="1.1.0-beta-20204-02" targetFramework="native" developmentDependency="true" />
   <package id="Microsoft.SourceLink.Common" version="1.1.0-beta-21055-01" targetFramework="native" developmentDependency="true" />
   <package id="Microsoft.SourceLink.GitHub" version="1.1.0-beta-20204-02" targetFramework="native" developmentDependency="true" />
-  <package id="Microsoft.Windows.ImplementationLibrary" version="1.0.210803.1" targetFramework="native" />
 </packages>
--- a/dev/Bootstrap/CS/Microsoft.WindowsAppRuntime.Bootstrap.Net/Microsoft.WindowsAppRuntime.Bootstrap.Net.csproj
+++ b/dev/Bootstrap/CS/Microsoft.WindowsAppRuntime.Bootstrap.Net/Microsoft.WindowsAppRuntime.Bootstrap.Net.csproj
@@ -23,4 +23,15 @@
     <AllowUnsafeBlocks>false</AllowUnsafeBlocks>
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.SourceLink.Common" Version="1.1.0-beta-21055-01">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.0-beta-20204-02">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
+  </ItemGroup>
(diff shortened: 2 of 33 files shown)


dev/Detours/packages.config
```diff
@@ -0,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="Microsoft.Build.Tasks.Git" version="1.1.0-beta-20204-02" targetFramework="native" developmentDependency="true" />
```
huichen123, line 3: >1.1.0-beta-20204-02 [](http://example.com/codeflow?start=51&length=19)

i hope one day we can have those versions defined in some common properties, therefore everyone is on the same page, and update only requires changing one place.

huichen123 (approved):
:shipit:

dev/Detours/packages.config
jonwis: Definitely something to look at in the future - also common .props file that projects reference so we don't have so many edits.

dev/WindowsAppRuntime_BootstrapDLL/packages.config
```diff
@@ -1,4 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
```
jonwis, line 1: Is there a way to at least centralize these `packages.config` files across the solution?

jonwis (approved):

dev/Detours/packages.config
DrusTheAxe: Nice idea but alas, packages.config has no #include/import/etc mechanism :-(

dev/WindowsAppRuntime_BootstrapDLL/packages.config
DrusTheAxe: Yes but not easily (VS doesn't provide a mechanism but we can, with work). It's come up in ES discussions. Ideas so far include...

* build\packages.props as a single shared master list of WIL, C++/WinRT, etc definitions/versions instead of in each *proj file
* new tooling (UpdatePackagesConfig.ps1 etc) to crawl the directory tree rewriting packages.config content per packages.props
* pipeline checks all packages.config match (and match versions in packages.props)
* Non-nuget packaging tech (eg VCPKG)

This overlaps the project-management topic we've discussed but haven't tackled yet. But that day's getting closer... :-)

DrusTheAxe (approved):

PR #1501, opened 2021-09-28
Add reference to WIL nuget to C++ project templates

patrickeganMSFT:
The C++ project templates do not have access to the resume_foreground coroutine because they do not currently reference the WIL nuget. This changes adds a reference to WIL.

Note that the version of the WIL nuget referenced, 1.0.210803.1, does not contain the new wil::resume_foreground coroutine. Rather, this PR does the preliminary work to include the WIL nuget. Once a newer version of the WIL nuget is created, we will update the version referenced here to get make wil::resume_foreground available in the C++ project templates.

### DIFF
--- a/dev/MRTCore/build/versions.props
+++ b/dev/MRTCore/build/versions.props
@@ -24,6 +24,7 @@
     <DotNetCoreRuntimeVersion>5.0.8</DotNetCoreRuntimeVersion>
     <DetoursVersion>4.0.1</DetoursVersion>
     <CppWinRTVersion>2.0.210806.1</CppWinRTVersion>
+    <WILVersion>1.0.210803.1</WILVersion>
     <MicrosoftSourceLinkAzureReposVersion>1.0.0</MicrosoftSourceLinkAzureReposVersion>
     <VCRTForwardersVersion>1.0.6</VCRTForwardersVersion>
     <WebView2Version>0.9.573-internal</WebView2Version>
--- a/dev/VSIX/ProjectTemplates/Desktop/CppWinRT/PackagedApp/BlankApp/pch.h
+++ b/dev/VSIX/ProjectTemplates/Desktop/CppWinRT/PackagedApp/BlankApp/pch.h
@@ -22,3 +22,4 @@
 #include <winrt/Microsoft.UI.Xaml.Navigation.h>
 #include <winrt/Microsoft.UI.Xaml.Shapes.h>
 #include <winrt/Microsoft.UI.Dispatching.h>
+#include <wil/cppwinrt.h>
--- a/dev/VSIX/ProjectTemplates/Desktop/CppWinRT/SingleProjectPackagedApp/pch.h
+++ b/dev/VSIX/ProjectTemplates/Desktop/CppWinRT/SingleProjectPackagedApp/pch.h
@@ -22,3 +22,4 @@
 #include <winrt/Microsoft.UI.Xaml.Navigation.h>
 #include <winrt/Microsoft.UI.Xaml.Shapes.h>
 #include <winrt/Microsoft.UI.Dispatching.h>
+#include <wil/cppwinrt.h>
--- a/dev/VSIX/ProjectTemplates/Neutral/CppWinRT/RuntimeComponent/pch.h
+++ b/dev/VSIX/ProjectTemplates/Neutral/CppWinRT/RuntimeComponent/pch.h
@@ -10,3 +10,4 @@
 #include <winrt/Microsoft.UI.Xaml.Markup.h>
 #include <winrt/Microsoft.UI.Xaml.Navigation.h>
 #include <winrt/Microsoft.UI.Dispatching.h>
+#include <wil/cppwinrt.h>
--- a/dev/VSIX/ProjectTemplates/UWP/CppWinRT/BlankApp/pch.h
+++ b/dev/VSIX/ProjectTemplates/UWP/CppWinRT/BlankApp/pch.h
@@ -17,3 +17,4 @@
 #include <winrt/Microsoft.UI.Xaml.Navigation.h>
 #include <winrt/Microsoft.UI.Xaml.Shapes.h>
 #include <winrt/Microsoft.UI.Dispatching.h>
+#include <wil/cppwinrt.h>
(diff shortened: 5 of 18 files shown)


dev/MRTCore/build/versions.props
```diff
@@ -24,6 +24,7 @@
     <DotNetCoreRuntimeVersion>5.0.8</DotNetCoreRuntimeVersion>
     <DetoursVersion>4.0.1</DetoursVersion>
     <CppWinRTVersion>2.0.210806.1</CppWinRTVersion>
+    <WILVersion>1.0.210803.1</WILVersion>
```
alwu-msft, line 27: ><WILVersion>1.0.210803.1</WILVersion> [](http://example.com/codeflow?start=4&length=37)

Is this used?

evelynwu-msft (approved):
:shipit:

dev/MRTCore/build/versions.props
DrusTheAxe, line 27: The intent is to standard versions of tooling/dependencies?

Please do this at the PROJECT level. Not just MRT. This should be versions.props in the root and all *proj files should be updated for this standardization. Not just MRT.

We've discussed this a few times but haven't gone here _yet_. Congratulations for being first to the key :-)

There's still the problem of `packages.config`. Given the lack of any #include/import/etc mechanism the best ideas to date include:

* Create a script to crawl the directory tree and update all `packges.config` based on `/versions.props`. Then version updates are (1) edit `/versions.props` and (2) run `/tools/PackagesConfig.cmd -Update` which runs `/tools/PackagesConfig.ps1 -Update` (the usual .ps1/.cmd pairing)
* Same script accepts `-Verify` to scan all `packages.config` and report any version mismatches with `/versions.props`
* Add a task in build pipelines to run `PackagesConfig.cmd -Verify` so builds always report on inconsistencies
* Add `PackagesConfig.*` to the Transport package so other repos can similarly use it. That probably means a 3rd option: `-Update`, `-Verify` and `-Version filename` that defaults to `/versions.props` but lets folks override

We _may_ eventually want to make release builds _enforce_ this consistency but there's always exceptions to rules so better to walk before we run and just _report_ inconsistencies for now.

Thoughts?

patrickeganMSFT: I'm not sure. I was looking for places where CppWinRTVersion was defined/used and saw this. I was hesitant to add this because it was outside of the dev/VSIX directory, but figured it'd be better to add then remove if unnecessary rather than omit and forget.

With that in mind, if this doesn't contribute to making the WIL nuget available to the C++ project templates, we should remove it. Do you agree?

patrickeganMSFT: I must admit that this was somewhat of a hunt-and-peck change. I was adding WILVersion anywhere I saw CppWinRTVersion used/defined. If this file does not affect the C++ project templates then adding the definition here is probably a mistake.

In regards to the issue with `packages.config`, I don't think I know enough about how this project is built to have any meaningful thoughts on it.

DrusTheAxe: > With that in mind, if this doesn't contribute to making the WIL nuget available to the C++ project templates, we should remove it. Do you agree?

This is in `dev/MRTCore/build`. What's that got to do with the templates? Aren't they under `dev/VSIX`?

patrickeganMSFT: This was a mistake and is removed in the most recent push.

dev/VSIX/ProjectTemplates/Desktop/CppWinRT/PackagedApp/BlankApp/pch.h
```diff
@@ -21,4 +21,5 @@
 #include <winrt/Microsoft.UI.Xaml.Media.h>
 #include <winrt/Microsoft.UI.Xaml.Navigation.h>
 #include <winrt/Microsoft.UI.Xaml.Shapes.h>
-#include <winrt/Microsoft.UI.Dispatching.h>
\ No newline at end of file
+#include <winrt/Microsoft.UI.Dispatching.h>
+#include <wil/cppwinrt.h>
```
jonwis, line 25: IIRC, this header needs to come _before_ any C++/WinRT headers are included.  See https://github.com/microsoft/wil/blob/a762808b1cd32b208b6f7ecd3a2f3ed32b9c3815/include/wil/cppwinrt.h#L78 - samples using this template should have failed to compile with this ordering.

dev/VSIX/ProjectTemplates/Desktop/CppWinRT/SingleProjectPackagedApp/pch.h
```diff
@@ -21,4 +21,5 @@
 #include <winrt/Microsoft.UI.Xaml.Media.h>
 #include <winrt/Microsoft.UI.Xaml.Navigation.h>
 #include <winrt/Microsoft.UI.Xaml.Shapes.h>
-#include <winrt/Microsoft.UI.Dispatching.h>
\ No newline at end of file
+#include <winrt/Microsoft.UI.Dispatching.h>
+#include <wil/cppwinrt.h>
```
jonwis, line 25: here too

dev/VSIX/ProjectTemplates/Neutral/CppWinRT/RuntimeComponent/pch.h
```diff
@@ -9,4 +9,5 @@
 #include <winrt/Microsoft.UI.Xaml.Data.h>
 #include <winrt/Microsoft.UI.Xaml.Markup.h>
 #include <winrt/Microsoft.UI.Xaml.Navigation.h>
-#include <winrt/Microsoft.UI.Dispatching.h>
\ No newline at end of file
+#include <winrt/Microsoft.UI.Dispatching.h>
+#include <wil/cppwinrt.h>
```
jonwis, line 13: Here too

dev/VSIX/ProjectTemplates/UWP/CppWinRT/BlankApp/pch.h
```diff
@@ -16,4 +16,5 @@
 #include <winrt/Microsoft.UI.Xaml.Media.h>
 #include <winrt/Microsoft.UI.Xaml.Navigation.h>
 #include <winrt/Microsoft.UI.Xaml.Shapes.h>
-#include <winrt/Microsoft.UI.Dispatching.h>
\ No newline at end of file
+#include <winrt/Microsoft.UI.Dispatching.h>
+#include <wil/cppwinrt.h>
```
jonwis, line 20: Here too

jonwis (changes requested):

PR #1502, opened 2021-09-28
Set TDBuild to commit by default and refactor parameters

kythant:
image

No idea why

### DIFF
--- a/build/WindowsAppSDK-TDBuild.yml
+++ b/build/WindowsAppSDK-TDBuild.yml
@@ -2,7 +2,7 @@ name: $(BuildDefinitionName)_$(date:yyMM).$(date:dd)$(rev:rrr)
 parameters:
 - name: GitAction
   type: string
-  default: 'None'
+  default: 'COMMIT'
   values:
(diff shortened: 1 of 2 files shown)


build/WindowsAppSDK-TDBuild.yml
```diff
@@ -21,5 +21,3 @@ jobs:
         dev\VSIX\Extension\Cpp\Common\VSPackage.resx
         dev\VSIX\Extension\Cs\Common\VSPackage.resx
       teamId: $(TdBuildTeamIDVSIX)
-      authId: $(TdBuildAuthID)
-      authKey: $(TdBuildToken)
```
alwu-msft, line 25: >authKey [](http://example.com/codeflow?start=6&length=7)

The parameter in the template was called `AuthToken`, not `authKey`. (Also, does the different casing matter?)

kythant: No it was always AuthKey?

alwu-msft (approved):
:shipit:

PR #1511, opened 2021-09-29
Deployment API Fix for Singleton not having major.minor in the package name

### LINKED ISSUE #1510: Deployment API does not properly match Singleton in production builds
**Describe the bug**

Two issues:
1) Package naming scheme has a version identifer in the package name of the form ".Major.Minor-Tag", except that Singleton packages do not have the ".Major.Minor" part and only has the "-Tag" at the end. This difference was not accounted for in the Deployment API's GetStatus() call, which causes mismatches on real builds such that GetStatus will never find the Singleton package, even when it is registered becuase it is seeking a different package name.

2) Test packages are not following the exact naming scheme, which is why the API tests pass.

More details are in the package naming scheme spec:
https://github.com/microsoft/WindowsAppSDK/blob/main/specs/Deployment/MSIXPackageVersioning.md

**Steps to reproduce the bug**
* Deploy nightly build MSIX and nuget
* Create project and use the DeploymentManager::GetStatus() method. Observe package install required.
* Run the installer from the nightly to install the entire SDK packages. 
* Run the GetStatus() in the project again. Observe package install still required.

dkbennett:
Fix  #1510

* Updates Deployment Test packages to have package names consistent with the versioning spec.
* Fixes GetStatus to now differentiate between Versioned and Unversioned main packages it checks against.
* Specifies Singleton as an unversioned package type.

Tested
* Deployment API tests pass with the new package names, including the unversioned Singleton name that is different from the framework version identifier.


dev/Deployment/DeploymentManager.cpp
```diff
@@ -42,16 +42,38 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
...
+            }
+            else if (package.versionType == PackageVersionType::Unversioned)
+            {
+                // Prefix + Subtypename + VersionTag + Suffix
+                packageFamilyName = WINDOWSAPPRUNTIME_PACKAGE_NAME_PREFIX WINDOWSAPPRUNTIME_PACKAGE_SUBTYPENAME_DELIMETER + package.identifier + packageNameVersionTag + WINDOWSAPPRUNTIME_PACKAGE_NAME_SUFFIX;
+            }
+            else
+            {
+                // Other version types not supported.
+                FAIL_FAST_HR(HRESULT_FROM_WIN32(ERROR_UNSUPPORTED_TYPE));
```
DrusTheAxe, line 75: NIT: `FAIL_FAST_HR_MSG(..., "VersionType:%d PackageName:%s", package.versionType, package.name(?))` would be convenient if something goes wrong. Data context in the log message with the code context simplifies troubleshooting :-)

DrusTheAxe (approved):
LGTM. 1 minor comment

jonwis (approved):

PR #1513, opened 2021-09-30
Refactored utf16->utf8 support from DynDep to common

DrusTheAxe:
Moved DynamicDependency's `to_utf8` functions to `Microsoft.Utf8.h` (and renamed `ToUtf8`) as the complementary inverse of the `ToUtf16` functions already there.

Deleted DynDep's now-obsolete `utf8.h` and changed to the `Microsoft.Utf8` namespace functions. All tests pass as expected


hulumane (approved):

dev/Common/Microsoft.Utf8.h
```diff
@@ -45,8 +49,52 @@ template<> inline winrt::hstring ToUtf16(PCSTR utf8)
...
+{
+    return ToUtf16(utf8.c_str());
+}
+
 inline winrt::hstring ToHString(PCSTR utf8)
 {
     return ToUtf16<winrt::hstring>(utf8);
 }
+
+// UTF16->UTF8 conversions
```
asklar, line 62: @DrusTheAxe  could you clarify why are all these needed / why don't we use `winrt::to_hstring` and `winrt::to_string` ?

PR #1516, opened 2021-09-30
Update the "Microsoft.Internal.WinUI.Helix" package version to 0.5.4

jefgen:
This updates the "`Microsoft.Internal.WinUI.Helix`" package version to `0.5.4` to pick up the fix for the re-try logic when the Helix infra fails, causing endless timing out.

From my PR in the WinUI repo:

> We recently ran into an issue in `WindowsAppSDK` where the `ProcessHelixFiles.ps1` script was taking over ~2 hours to run (due to Helix infrastructure being down) as it never timed out / hit the maximum number of retries.
> Looking at the logs, the retry logic was never hitting the max attempts of 10 times to obtain the results from Helix.
> It turns out that these variables need to be explicitly referred to in the "global" scope, otherwise the individual functions end up redeclaring the same variable. (We want there to be a maximum number of retries overall for all functions.)

### DIFF
--- a/build/AzurePipelinesTemplates/WindowsAppSDK-RunHelixTests-Job.yml
+++ b/build/AzurePipelinesTemplates/WindowsAppSDK-RunHelixTests-Job.yml
@@ -25,7 +25,7 @@ jobs:
     # This version number should be kept in sync with the value in these files:
     #   build\Helix\packages.config
     #   test\WindowsAppSDK.Helix.TestCommon\WindowsAppSDK.Helix.TestCommon.csproj
-    winUIHelixVer: 0.5.2
+    winUIHelixVer: 0.5.4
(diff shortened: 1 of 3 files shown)


kythant:
That's alot of places to update a version in just like for the FrameworkUDK. Not going to block on this but something I should look at later to improve. Just a comment.

kythant (approved):

jefgen:
> That's alot of places to update a version in just like for the FrameworkUDK. Not going to block on this but something I should look at later to improve. Just a comment.

Yeah, it's a bit better in the internal repo where there is a shared file with the version `WindowsAppSDK-Nuget-Versions.yml`.
It would be good to do the same here as well, but I wanted to quickly update the Nuget version.

kmahone (approved):

PR #1522, opened 2021-10-01
Unconditionally restore vsext to Dev16 VSIX

DefaultRyan:
#1482 removed the MSIX vsext from the Component extension, but this was mistakenly done for the Dev16 extension. The Dev16 extension doesn't have a component-specific build with a different filename, so it builds twice, with the Component clobbering the Standalone.

This reverts the part of #1482 that touched the Dev16 projects.

Long-term, we should tidy up this part of the build so we do less work (i.e. don't build Dev16 twice).


evelynwu-msft (approved):
:shipit:

PR #1535, opened 2021-10-04
Update Calls to GetMrtPackagingOutput to follow the Project Reference protocol.

jlaanstra:
This PR updates the MSBuild calls inside GetMrtPackagingOutput to closer match the MSBuild project reference protocol as documented at https://github.com/dotnet/msbuild/blob/main/documentation/ProjectReference-Protocol.md and implemented by the ResolveProjectReferences target in https://github.com/dotnet/msbuild/blob/main/src/Tasks/Microsoft.Common.CurrentVersion.targets

- Update to use _MSBuildProjectReferenceExistent, which is what ResolveProjectReferences uses.
- Add SetTargetFramework to the Properties list, so packaging output from projects that do multi-targeting is correctly resolved.
- Respects _MSBuildProjectReferenceExistent.GlobalPropertiesToRemove.

### DIFF
--- a/dev/MRTCore/packaging/MrtCore.PriGen.targets
+++ b/dev/MRTCore/packaging/MrtCore.PriGen.targets
@@ -1223,7 +1223,7 @@
       SkipNonexistentTargets="true"
       ContinueOnError="!$(BuildingProject)"
       RemoveProperties="%(_MSBuildProjectReferenceExistent.GlobalPropertiesToRemove)$(_GlobalPropertiesToRemoveFromProjectReferences)">
-      <Output TaskParameter="TargetOutputs" ItemName="_PackagingOutputsFromOtherProjects"/>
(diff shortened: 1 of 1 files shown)


jlaanstra:
> Can all instances of `AssignProjectConfiguration` (and possibly some other targets as well) be safely removed now? For example, `AssignProjectConfiguration` is in `GetPackagingOutputsDependsOn`.

Probably. Let me have a look at that.

rohanp-msft:
> @rohanp-msft Need to keep those targets in `GetPackagingOutputsDependsOn` as those targets actually ensure `_MSBuildProjectReferenceExistent` is initialized.

Thanks for looking!

rohanp-msft:
> @rohanp-msft Looks like the build passed except for some limitations of building from my fork.

All clear now @jlaanstra. Thanks, @jefgen!

Issue #1277, opened 2021-08-20
RoGetActivationFactoryDetour doesn't support reentrancy?


DrusTheAxe:
### Discussed in https://github.com/microsoft/WindowsAppSDK/discussions/1246

<div type='discussions-op-text'>

<sup>Originally posted by **riverar** August 15, 2021</sup>
My app is creating an instance of `XamlControlsResources` which involves `RoGetActivationFactory`. In at least one scenario, this involves creating an instance of `PropertyMetadata` while fulfilling this request, which results in another call to `RoGetActivationFactory`. But because of the `unique_lock` usage in the Mdd hook, this results in an exception.

I don't have a repro for you at this time (hence this is a discussion, not a proper issue), but consider the following stack:
```
00 KERNELBASE!RaiseException
01 Microsoft_WindowsAppSDK!_CxxThrowException
02 Microsoft_WindowsAppSDK!std::_Throw_Cpp_error
03 Microsoft_WindowsAppSDK!std::_Throw_C_error
04 Microsoft_WindowsAppSDK!std::_Check_C_return
05 Microsoft_WindowsAppSDK!std::_Mutex_base::lock
06 Microsoft_WindowsAppSDK!std::unique_lock<std::mutex>::{ctor}
07 Microsoft_WindowsAppSDK!MddCore::WinRTModuleManager::GetThreadingType
08 Microsoft_WindowsAppSDK!MddCore::WinRT::GetThreadingModel
09 Microsoft_WindowsAppSDK!WinRTGetThreadingModel_PackageGraph
0a Microsoft_WindowsAppSDK!WinRTGetThreadingModel
0b Microsoft_WindowsAppSDK!GetActivationLocation
0c Microsoft_WindowsAppSDK!RoGetActivationFactoryDetour
0d Microsoft_UI_Xaml_Controls!winrt::impl::get_runtime_activation_factory_impl<0>
0e Microsoft_UI_Xaml_Controls!winrt::impl::factory_cache_entry<winrt::Microsoft::UI::Xaml::PropertyMetadata,winrt::Microsoft::UI::Xaml::IPropertyMetadataFactory>::call<<lambda_073aad94565dae974f0649b9221facfc> & __ptr64>
0f Microsoft_UI_Xaml_Controls!InitializeDependencyProperty
10 Microsoft_UI_Xaml_Controls!XamlControlsResourcesProperties::EnsureProperties
[...]

PR #1538, opened 2021-10-05
UndockedRegFreeWinRT reentrancy bug

DrusTheAxe:
The UndockedRegFreeWinRT (URFW) support for the package graph deadlocks on `std::mutex` due to reentrancy.

Fixed by changing `std::mutex` to `std::recursive_mutex`.

[RoGetActivationFactoryDetour doesn't support reentrancy? #1277](https://github.com/microsoft/WindowsAppSDK/issues/1277)

NOTE: I've added tests to repro the issue but something's not not quite right about them as I can't reproduce the issue. But XAML has demonstrative proof via local testing the reentrancy issue is real _AND_ it's fixed by switching `mutex` to `recursive_mutex`. Visual inspection of the code verifies that. Rather than block critical path of folks awaiting this fix we'll go with the fix we know works and I'll follow up separate with the XAML folks to try to identify how to update the tests.


DrusTheAxe:
`Sorry! Our services aren't available right now.`

Azure DevOps (Helix?) problem on the x86 run :-(

Let's try again...

Scottj1s (approved):

kennykerr:
https://devblogs.microsoft.com/oldnewthing/20201112-00/?p=104444

PR #1547, opened 2021-10-05
Fix instancing mutex name generation to support multiple keys per app.

aeloros:
The mutex used to synchronize key names was not properly including the key name in it.  This changes it so that the provided key is part of the unique named mutex, and that it is properly escaped to remove the backslash character.

### DIFF
--- a/dev/AppLifecycle/AppInstance.cpp
+++ b/dev/AppLifecycle/AppInstance.cpp
@@ -398,7 +398,8 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
 
     bool AppInstance::TrySetKey(std::wstring const& key)
     {
-        std::wstring mutexName = m_moduleName + L"_Mutex";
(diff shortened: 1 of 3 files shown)


dev/AppLifecycle/AppInstance.cpp
```diff
@@ -398,7 +398,8 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
 
     bool AppInstance::TrySetKey(std::wstring const& key)
     {
-        std::wstring mutexName = m_moduleName + L"_Mutex";
+        auto escapedKey = std::regex_replace(key, std::wregex(L"\\\\"), L"_");
+        std::wstring mutexName = wil::str_printf<std::wstring>(L"%s_%s_Mutex", m_moduleName.c_str(), escapedKey.c_str());
```
DrusTheAxe, line 402: Is there a limit on maxlength of a named mutex?

Not worried about it short-term but if there is a limit it _may_ be something we want to consider dealing with longer term

PR #1563, opened 2021-10-08
DynamicDependencies: Change IsElevated check from FAIL_FAST to RETURN(ErrorNotSupported)

DrusTheAxe:
https://task.ms/36535227 

Change DynamicDependencies+Bootstrap is-elevated response from FailFast to ReturnError(NOT_SUPPORTED)

Added test case for Bootstrap API elevation invocation. Did NOT add a test case for DynamicDependencies elevation invocation as the tests aren't wired up to facilitate that (wrong process configurations). Doing this would require bigly-invasive restructurings so not justified; you can't get there w/o going thru the bootstrapper anyway so more theoretical than actual problem. Will consider in the future when more test work occurs.

Also:

* Tweaked DevCheck.cmd powershell invocation to work regardless of machine config (i.e. like we do elsewhere)
* Fixed file misname `winrt_msixdynamicdepednency` to the correct `...dependency` (I keep transposing those 2 letters. My nemesis :-)
* Fixed GetCurrentPackageInfoTests to run IsolationLevel=Class to avoid false errors due to process pollution if you run all DynDep tests in 1 pass. NOTE: IsolationLevel=Method would be too fine grained as the Setup fixture wouldn't happen in the right process context (plus test execution time would be greatly extended). No real need to go =Method and negative value so nah, =Class will do.

### DIFF
--- a/tools/DevCheck.cmd
+++ b/tools/DevCheck.cmd
@@ -8,7 +8,7 @@ powershell -ExecutionPolicy Unrestricted -NoLogo -NoProfile %~dpn0.ps1 %*
 GOTO TheEnd
 
 :Help
-powershell -c Get-Help %~dpn0.ps1 -full
+powershell -ExecutionPolicy Unrestricted -NoLogo -NoProfile -c Get-Help %~dpn0.ps1 -full
 
 :TheEnd
 ENDLOCAL


riverar:
Thanks, a very welcome change. I've dove into the debugger multiple times already to debug a crash on startup, only to realize I was launching elevated.

PR #1581, opened 2021-10-12
Updating WinUI C# project templates to target .NET 6 when building in Dev17

RealTommyKlein:
~~This pull request updates the WinUI C# project templates to target .NET 6 when the project is built in Dev17.  This is controlled by changing two MSBuild properties, `TargetFramework` and `AssetTargetFallback`, to reference a `net6.0` target framework moniker instead of `net5.0`.~~

~~Because the Dev16 and Dev17 VSIXs package the same project templates, we cannot change the emitted template based on whether the VSIX is for Dev16 or Dev17.  Instead, the .NET version determination is made at the created project's build-time, based on `VisualStudioVersion`.  If `VisualStudioVersion` is at least `17.0` (indicating we're building in Dev17 or later), the project uses the .NET 6 TFM - otherwise it uses the .NET 5 TFM.~~

_Note: the above description was for the initial implementation, which was replaced in favor of a [new approach](https://github.com/microsoft/WindowsAppSDK/pull/1581#issuecomment-941790045).  Leaving the original description above so the subsequent discussion makes sense._

This pull request updates new projects created via the VS2022 WinUI C# project templates to target .NET 6 instead of .NET 5, as VS2022 only ships with .NET 6 and the new project experience in VS2022 would be broken unless the developer had also installed .NET 5.  New projects created using the VS2019 WinUI C# project templates will continue to target .NET 5.

This is implemented by having the standalone C# VSIXs for VS2019 (`WindowsAppSDK.Cs.Extension.Dev16.csproj`) and VS2022 (`WindowsAppSDK.Cs.Extension.Dev17.csproj`) inject a custom parameter into the C# project templates .vsmanifests, similar to how the .vsmanifests are updated with the `Microsoft.WindowsAppSDK` NuGet package version they should reference.

### DIFF
--- a/dev/VSIX/Extension/Cs/Dev16/WindowsAppSDK.Cs.Extension.Dev16.csproj
+++ b/dev/VSIX/Extension/Cs/Dev16/WindowsAppSDK.Cs.Extension.Dev16.csproj
@@ -26,6 +26,8 @@
     <UICulture>en-US</UICulture>
     <!-- This MUST match VSPackage.PackageGuidString -->
     <PackageGuidString>{B14EB1D0-DD99-4930-A232-23275B4F8ED6}</PackageGuidString>
+    <!-- Dev16 templates should target .NET 5 -->
+    <DotNetVersion>net5.0</DotNetVersion>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' " />
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' " />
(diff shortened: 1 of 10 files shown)


riverar:
But why does the Visual Studio version influence which .NET I use? Why is this change being made?

dotMorten:
> Yes, these are vstemplates for consistency with our C++ templates

They'll still use a .vstemplate to get the generated nuget packaged integrated into the vsix. But it'll just be a single one for all 3 .NET ones.

RealTommyKlein:
I spent some time looking at the .NET CLI template approach, but would need more to ramp up on how they work and how to integrate the particulars of our templates (e.g. the wapproj, and the equivalent of our injections into the .vstemplate like `WindowsAppSDKNupkgVersion`).

Latest commit causes the `WindowsAppSDK.Cs.Extension.Dev16` VSIX to create projects which use .NET 5, while the `WindowsAppSDK.Cs.Extension.Dev17` VSIX will use .NET 6.

dotMorten:
In my mind that is a perfectly acceptable behavior for v1.0.

kinkycat91 (commented):
X

PR #1582, opened 2021-10-12
Allow Unpackaged Apps to use WindowsAppRuntime Framework Package via Targets Files

ShashankNay:
This change introduces .targets and .cs/.cpp files within the Microsoft.WindowsAppSDK NuGet that enables access to packaged content at runtime. It will allow an existing unpackaged app to use APIs from framework package binaries by simply adding the Microsoft.WindowsAppSDK NuGet package reference and some project properties. The `MddBootstrapInitializer.cpp` and `MddBootstrapModuleInitializer.cs` files will call into `MddBootstrapInitialize` when the app launches and before any APIs within the Microsoft.WindowsAppSDK are called as well as Main. This is done using ModuleInitializers  in C# and a static struct in C++.

Both native and managed projects include `Microsoft.WindowsAppSDK.MddCommon.targets` which will determine whether the `GenerateBootstrapCpp`/`GenerateBootstrapCS` targets run by examining the `WindowsAppSdkBootstrapInitialize` property. If it is not explicitly set, it will be set to true only if `WindowsPackageType` is None and `WindowsAppSDKSelfContained` is not true(the application is unpackaged and framework dependent). The `MddBootstrapInitializer.cpp` and `MddBootstrapModuleInitializer.cs` files use `WindowsAppSDK-VersionInfo.h`/`WindowsAppSDK-VersionInfo.cs` to pick up the appropriate version numbers to query framework packages installed on a machine and load binaries from there when necessary. An important distinction between the C++ and C# file is the lack of a `MddBootstrapShutdown` call in the C# file. This is done in C++ through the destructor.

### DIFF
--- a/build/ProjectReunion-BuildFoundation.yml
+++ b/build/ProjectReunion-BuildFoundation.yml
@@ -155,7 +155,7 @@ jobs:
     inputs:
       command: 'custom'
       arguments: 'restore ${{ parameters.WindowsAppRuntimeInsightsSourceDirectory }}\packages.config -ConfigFile ${{ parameters.WindowsAppRuntimeInsightsSourceDirectory }}\nuget.config -PackagesDirectory ${{ parameters.WindowsAppRuntimeInsightsSourceDirectory }}\packages'
-  
+
   - template: AzurePipelinesTemplates\WindowsAppSDK-BuildProject-Steps.yml
     parameters:
       solutionPath: 'dev\Bootstrap\CS\Microsoft.WindowsAppRuntime.Bootstrap.Net\Microsoft.WindowsAppRuntime.Bootstrap.Net.csproj'
@@ -169,7 +169,7 @@ jobs:
       SourceFolder: '$(buildOutputDir)\$(buildConfiguration)\$(buildPlatform)\Microsoft.WindowsAppRuntime.Bootstrap.Net'
       Contents: |
         Microsoft.WindowsAppRuntime.Bootstrap.Net.dll
-      TargetFolder: '$(Build.ArtifactStagingDirectory)\FullNuget\lib\net5.0-windows10.0.18362.0'
+      TargetFolder: '$(Build.ArtifactStagingDirectory)\FullNuget\lib\net5.0-windows10.0.17763.0'
       flattenFolders: false
 
   - task: PublishBuildArtifacts@1


DrusTheAxe:
> load the assembly/app for inspection/analysis. It's my understanding the module initializer will kick in and execute then, in that process.

That would be bad.

@Scottj1s ?

riverar:
Consider an app that exposes a static Version property (e.g. AssemblyFileVersion).

```csharp
public class App
{
    [ModuleInitializer]
    public static void Initializer() => Environment.Exit(-1);
    public static int Version => 42;
}
```

Here's how this example plays out:

```powershell
PS> [Reflection.Assembly]::LoadFile("C:\path\to\app.exe")
PS> "App version: {0}" -f [Namespace.App]::Version

[process exited with code 4294967295]
```

If this was a unpackaged WAS app, Bootstrap would execute for powershell.exe and what happens after that is undefined. Could break a CI pipeline, introduce weird behavior, etc.

Scottj1s:
otoh, that's just a form of opting out, after first having opted in.  this seems like an edge case that we shouldn't be concerned with.

---
In reply to: [942987325](https://github.com/microsoft/WindowsAppSDK/pull/1582#issuecomment-942987325) [](http://example.com/codeflow?ancestors=942987325,941366005)

Scottj1s:
> At risk of sounding like a broken record, I'm not sure if this PR is the right solution. Feels like a Visual Studio template tweak with the few lines of bootstrap/bootstrap.cs needed would've solved this much cleaner.

@riverar , we considered that, but functionality built into project templates does not support ad hoc availability (existing projects).  that said, we could certainly also introduce several unpackaged project templates, which do what you suggest instead of opting into the automatic background initialization here.

nickrandolph:
Hi all, I have just been playing around with preview3 and it took me a long time to realise why my app was failing to even get into my custom Main method - it would crash with a COMException in relation to the Bootstrap (I was intentionally testing without the dependencies installed). I easily found the AutoInitialise method which was causing the exception but it took a lot longer (until I found this thread) to understand how I could override the auto-generated code.
I don't like the injection of module level code - it doesn't lend itself to be easily overridden, or allow for app-level diagnostic reporting/messaging to the user.
Is there a reason this initialization isn't simply called from the auto-generated Program.Main?

DrusTheAxe:
TL;DR Your concern isn't with the bootstrapper API but rather our versioning base principle for 1.x. This is a topic in its own right. If you'd like to continue the discussion I recommend starting a Discussion thread (or Issue?) rather than comments in a closed PR.

>_(that is, ship a WAS minor release to service tooling without requiring a whole new runtime to be installed/deployed)_

That's right. This is a by-product of our decision to restrict your **runtime** to the **major.minor** release you build with.

>Developer updates WAS to 1.1--a minor bump, which should be backwards compatible with 1.0 machines

But it's not, because our versioning principle for 1.x says you only run with what you build at the major.minor boundary (not just major). If you change your app to build with 1.8 instead of 1.7 you need the 1.8 runtime. The 1.7 runtime is now irrelevant to you.

NOTE: It doesn't matter if 1.1 is ABI compatible with 1.0 because this versioning rule prevents using it. 1.1 could be called 2.0 for all it matters, your 1.0 app won't run with anything but the 1.0 runtime.

DrusTheAxe:
> Hi all, I have just been playing around with preview3 and it took me a long time to realise why my app was failing to even get into my custom Main method - it would crash with a COMException in relation to the Bootstrap (I was intentionally testing without the dependencies installed). I easily found the AutoInitialise method which was causing the exception but it took a lot longer (until I found this thread) to understand how I could override the auto-generated code. I don't like the injection of module level code - it doesn't lend itself to be easily overridden, or allow for app-level diagnostic reporting/messaging to the user. Is there a reason this initialization isn't simply called from the auto-generated Program.Main?

Hi @nickrandolph. I recommend creating a new Issue to further discuss, or perhaps a new Discussion thread. Either way a better venue than comments on a closed PR.

PR #1586, opened 2021-10-12
Add C++savvy API over Bootstrap's Flat-C API

DrusTheAxe:
Experimental C++ API over the Bootstrap Flat-C API.  This is experimental to garner feedback on the API.

Provides equivalent goodness as the C# API does for .NET developers -- overloading, automatic call to Shutdown, yadda yadda. Notably:

* All symbols in the Microsoft::Windows::ApplicationModel::DynamicDependency::Bootstrap namespace
* `TryInitialize()` returns HRESULT vs `Initialize()` throws exceptions
* `*Initialize()` overloaded for convenience
* `Shutdown()` for completeness
* `AutoShutdown` class for use by folks directly using `MddBootstrapInitialize()` e.g.

```
RETURN_IF_FAILED(MddBootstrapInitialize(...));
Microsoft::Windows::ApplicationModel::DynamicDependency::Bootstrap::AutoShutdown bootstrapShutdown;
```

* `AutoInitialize` class wrapping up all the above e.g.

```
Microsoft::Windows::ApplicationModel::DynamicDependency::Bootstrap::AutoInitialize bootstrapInitialize;
```

or

```
Microsoft::Windows::ApplicationModel::DynamicDependency::Bootstrap::AutoInitialize bootstrapInitialize(0x00040002, "experimental7");
```

As-is it assumes `#include <WindowsAppSDK-VersionInfo.h>` occurs before `#include <MsixDynamicDependency.h>`. An open question for feedback: Should it? If we #include the version info header do we also need to make it optional?

Another Q: `[Try]Initialize(majorMinor, versionTag)` takes optional parameters for both. Is that a bad thing? Remove the optional parameters? If anything's specified everything needs to be or should it assume `versionTag=null` and `minVersion=0.0.0.0`? e.g. is `TryInitialize(0x01234567)` better as

```
inline HRESULT TryInitialize(uint32_t majorMinor)
{
    return TryInitialize(majorMinor, nullptr, PackageVersion());
}
```

or the current

```
inline HRESULT TryInitialize(uint32_t majorMinor)
{
    return TryInitialize(majorMinor, WINDOWSAPPSDK_RELEASE_VERSION_TAG_W, WINDOWSAPPSDK_RELEASE_VERSION_UINT64);
}
```

?

Should `[Try]Initialize()` be overloaded to accept `PackageVersion` **and** `uint64_t version`? And `uint16_t major, minor, build, revision`? i.e. PackageVersion is accepted as a parameter but raw values are also OK?

The [spec](https://github.com/microsoft/WindowsAppSDK/blob/main/specs/dynamicdependencies/DynamicDependencies.md) will be updated later (before moving this out of experimental) after feedback and validation/refinement of the design has occurred.

NOTE: This PR's aged out so it's superseded by https://github.com/microsoft/WindowsAppSDK/pull/2191


kanismohammed:
@emrdev is added to the review. #Closed

dev/Common/TerminalVelocityFeatures-DynamicDependency.h
```diff
@@ -3,21 +3,21 @@
...
+// INPUT FILE: dev\\Common\\TerminalVelocityFeatures-DynamicDependency.xml
+// OPTIONS: -Channel Experimental -Language C++ -Namespace Microsoft.Windows.ApplicationModel.DynamicDependency
 
 // Feature constants
-#define WINDOWSAPPRUNTIME_MICROSOFT_WINDOWS_APPLICATIONMODEL_DYNAMICDEPENDENCY_FEATURE_GENERATIONID_ENABLED 1
+#define WINDOWSAPPRUNTIME_MICROSOFT_WINDOWS_APPLICATIONMODEL_DYNAMICDEPENDENCY_FEATURE_BOOTSTRAPCPPINITIALIZATION_ENABLED 1
 
 #if defined(__cplusplus)
 
 namespace Microsoft::Windows::ApplicationModel::DynamicDependency
```
Jaiganeshkumaran, line 14: Can these APIs be under the winrt namespace when using alongside C++/WinRT so that you don't have any namespace collisions? I know it's not a WinRT but still, consider keeping it under namespace winrt for consistency.

DrusTheAxe: When would you have namespace collisions?

These aren't WinRT APIs so it'd be pretty odd for them to appear in a winrt:: namespace.

Hmmm. These are primarily used in implementation of a feature, not using a feature. The header would appear to users if a Flat-C API needed the #define'd constants, but in that case the __cplusplus block isn't relevant. Only reason someone using WinAppSDK would need the that is if a feature provided a header with inline code and that needed to be compiled in/out.

It should be possible to omit the __cplusplus block in the common case e.g.

```
#if defined(__cplusplus) && defined(WINDOWSAPPRUNTIME_MICROSOFT_..._CPLUSPLUS_ENABLED)
```

Thanks for bringing this up

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.h
```diff
@@ -26,4 +26,182 @@ STDAPI MddBootstrapInitialize(
...
+
+        ~AutoShutdown()
+        {
+            Shutdown();
+        }
+    };
+
+#if defined(_CPPUNWIND) && defined(WIL_ENABLE_EXCEPTIONS)
+    // Automate Boostrap initialization and shutdown.
+    class AutoInitialize
```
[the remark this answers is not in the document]
DrusTheAxe: `auto mddBootstrapCleanup{ ::Microsoft::Windows::ApplicationModel::DynamicDependency::Bootstrap::Initialize() };`

PR #1612, opened 2021-10-18
Strip experimental bits from winmd and Cswinrt projections

DefaultRyan:
We need to ensure that experimental APIs are not shipped in the winmds and C# projections of stable/preview builds. This change has a few components working together:

### TerminalVelocity and IDL ###

The TerminalVelocity tool now puts additional code in the C++ headers intended to be consumed by MidlRT. It looks like this:
```
#ifdef __midlrt
namespace feature
{
  feature_name Feature_SomeWidgetFeature = { <state>, FALSE };
}
#endif
```
Where `<state>` depends on the TerminalVelocity state calculated from the current channel and feature's xml specification. IDL files defining this feature in WinRT will no longer use `[experimental]` directly, but will instead use `[feature(Feature_SomeWidgetFeature)]`.

There were two WinRT features that I updated to use the new syntax: PushNotifications and EnvironmentManager

### Mdmerge to generate final output winmds ###

Added a new MSBuild target in the WindowsAppRuntime_DLL project that will transform the "raw" winmds with `[feature]` into the final winmds we use to ship and generate CSWinRT projections. This uses mdmerge and its command-line option, `/transformExperimental:transform`. This option transforms APIs with the feature attribute:
- If the feature is `AlwaysDisabled`, the API is removed.
- If the feature is `DisabledByDefault`, the feature attribute is replaced with the experimental attribute.

This approach has the benefit of not adding any real inner-loop friction, and still allows local unit tests to consume the unstripped winmd with the experimental APIs present.

### Connect up the logic in the build pipelines ###

Finally, I modified our pipeline ymls to define the build channel in WindowsAppSDK-CommonVariables.yml and pass that down to the build ymls via the 'channel' parameter. WindowsAppSDK-BuildProject-Steps.yml already had this parameter and was passing it to the TerminalVelocity tool, but there was no plumbing to override the default value of 'experimental'.

Also modified CopyFilesToStagingDir.ps1 to copy the final stripped winmds instead of the internal winmds with `[feature`].

### Verification ###

For verification, I initially committed a version where the build channel was set to 'stable', and ran our nightly build pipeline, with the results [here](https://microsoft.visualstudio.com/ProjectReunion/_build/results?buildId=40393457&view=results). Both **Microsoft.Windows.PushNotifications.winmd** and **Microsoft.Windows.System** are empty, and both corresponding CSWinRT projections are absent. I will run a CI build as well and verify that things look good there, and then I'll change the build channel back to `'experimental'`.

### DIFF
--- a/docs/Coding-Guidelines/Experimental.md
+++ b/docs/Coding-Guidelines/Experimental.md
@@ -72,3 +72,22 @@ populate `-i`. Finally, it uses `-n:9` to split the output winmds by namespace, 
 
 When it is time to consume winmds for generating public projections (e.g. C#/WinRT) and to copy
 winmds for publishing to NuGet, be sure to use the stripped winmds, not the intermediate winmds.
+
+# Rationale and rejected approaches
+
+## Use [experimental] directly in IDL, with channel-based stripping
+
+This has the benefit of clarity in the IDL file, because `[experimental]` states a clearer intent than `[feature(Feature_MyFeature)]`.
+However, this would have one major drawback, and one inconvenience:
+1. The state of the WinRT API stripping would no longer be tied to the state of the implementation. This would increase the risk
+of accidentally having a disabled implementation behind a visible API, or having a live implementation hidden by a stripped API.
+2. Stripping of the WinRT API would no longer be controlled by TerminalVelocity, but would be controlled inside
+the MSBuild of the project, which would neccessitate plumbing that pipeline build parameter into the build.
+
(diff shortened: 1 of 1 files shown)


DefaultRyan:
And here is the successful CI run in 'stable' mode. https://dev.azure.com/ms/ProjectReunion/_build/results?buildId=231782&view=results

Now to change the checkin back to 'experimental' for the main branch.

PR #1630, opened 2021-10-21
Default enable `Deploy` action for Single-project MSIX Packaging projects

evelynwu-msft:
When a project is added to a solution, VS will query if the project is deployable and if so enable the `Deploy` action by default. For projects using Single-project MSIX Packaging, the question of whether or not the project is deployable is answered by the Single-project MSIX Packaging Tools extension; however, this extension is only activated if the `Msix` project capability is present, and that project capability is added by the WinAppSdk Nuget package. The issue is that projects using `PackageReference` restore their Nuget packages *after* VS has queried for deployability, so our projects were always appearing to be not deployable because without the project capability our extension was not being activated.

The solution is to explicitly include the project capability in the project template. This ensures that our extension gets activated and can properly inform VS that the project is in fact deployable, thereby allowing the `Deploy` action to be enabled by default. The project capability added by the template is guarded by a "killswitch" property that our Nuget package sets; the intent is that once our Nuget package has been restored and its `.props`/`.targets` have been imported, there's no longer a need for the project file itself to define the project capability as it is now redundant.

This PR also includes a couple other improvements to the project templates:
* `AppContainerApplication` was incorrectly set to `true` in the C++ Single-project MSIX Packaging template in an earlier PR; it needs to remain `false`. This also incidentally results in VS correctly displaying the project in Solution Explorer as "(Desktop)".
* Remove `.vcxproj.user` from the C++ project template. `.user` files are generally ignored by source control which makes it a poor choice for ensuring that an appropriate value for the `DebuggerFlavor` property has been set. Instead, logic has been added to the Nuget package to handle this behavior.
* Add an additional launch profile to the C# Single-project MSIX Packaging template in order to ease the experience of converting a packaged app into an unpackaged app.

This PR addresses internal bugs [31483194](http://task.ms/31483194) and [36705202](http://task.ms/36705202).

### DIFF
--- a/dev/VSIX/ProjectTemplates/Desktop/CSharp/SingleProjectPackagedApp/ProjectTemplate.csproj
+++ b/dev/VSIX/ProjectTemplates/Desktop/CSharp/SingleProjectPackagedApp/ProjectTemplate.csproj
@@ -13,11 +13,6 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <AppxManifest Include="Package.appxmanifest">
-      <SubType>Designer</SubType>
-    </AppxManifest>
-  </ItemGroup>
-  <ItemGroup>
     <Content Include="Assets\SplashScreen.scale-200.png" />
     <Content Include="Assets\LockScreenLogo.scale-200.png" />
     <Content Include="Assets\Square150x150Logo.scale-200.png" />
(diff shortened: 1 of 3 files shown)


evelynwu-msft:
> is there a corresponding .targets change?
@huichen123 https://microsoft.visualstudio.com/WinUI/_git/Microsoft-WinUI-Sdk/pullrequest/6587707

PR #1649, opened 2021-10-26
UndockedRegFreeWinRT doesn't check Dynamic packages for WinRT Metadata

### LINKED ISSUE #2382: ApiInformation.IsPropertyPresent throws in an unpackaged app
### Describe the bug

When I call `ApiInformation.IsPropertyPresent("Microsoft.UI.Xaml.Media.Imaging.BitmapImage", "IsAnimatedBitmap")` in an unpackaged app it is throwing an exception:

1. Create new “Blank App, Packaged (WinUI 3 in Desktop)” project.
2. Edit the .csproj to add:

```xml
  <PropertyGroup>
    <WindowsPackageType>None</WindowsPackageType>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.WindowsAppSDK" Version="1.0.2" />
  </ItemGroup>
```
3. Change the `MainWindow.xaml.cs` to have `myButton_Click` method as:

```C#
private void myButton_Click(object sender, RoutedEventArgs e)
{
    myButton.Content = ApiInformation.IsPropertyPresent("Microsoft.UI.Xaml.Media.Imaging.BitmapImage", "IsAnimatedBitmap");
}
```

4. Build the app for `Release`, `x64`.
5. run the .exe in the `bin\x64\Release\net6.0-windows10.0.19041.0\win10-x64` folder
6. Click the button

### Expected behavior

The app shouldn't crash because `ApiInformation.IsPropertyPresent` should either return `true` or `false` and not throw an exception.

### Actual behavior

```
System.InvalidOperationException
  HResult=0x80131509
  Message=Exception of type 'System.InvalidOperationException' was thrown.
  Source=WinRT.Runtime
  StackTrace:
   at WinRT.ExceptionHelpers.<ThrowExceptionForHR>g__Throw|20_0(Int32 hr) in D:\a\1\s\src\WinRT.Runtime\ExceptionHelpers.cs:line 82
```

```
 	WinRT.Runtime.dll!WinRT.ExceptionHelpers.ThrowExceptionForHR.__Throw|20_0(int hr) Line 82	C#
>	WinRT.Runtime.dll!WinRT.ExceptionHelpers.ThrowExceptionForHR(int hr) Line 85	C#
 	Microsoft.Windows.SDK.NET.dll!ABI.Windows.Foundation.Metadata.IApiInformationStaticsMethods.IsPropertyPresent(WinRT.IObjectReference _obj, string typeName, string propertyName) Line 493	C#
 	Microsoft.Windows.SDK.NET.dll!Windows.Foundation.Metadata.ApiInformation.IsPropertyPresent(string typeName, string propertyName) Line 88	C#
 	App9.dll!App9.MainWindow.myButton_Click(object sender, Microsoft.UI.Xaml.RoutedEventArgs e) Line 34	C#
 	Microsoft.WinUI.dll!WinRT._EventSource_global__Microsoft_UI_Xaml_RoutedEventHandler.EventState.GetEventInvoke.AnonymousMethod__1_0(object sender, Microsoft.UI.Xaml.RoutedEventArgs e) Line 7875	C#
 	Microsoft.WinUI.dll!ABI.Microsoft.UI.Xaml.RoutedEventHandler.Do_Abi_Invoke.AnonymousMethod__0(Microsoft.UI.Xaml.RoutedEventHandler invoke) Line 43206	C#
 	Microsoft.WinUI.dll!ABI.Microsoft.UI.Xaml.RoutedEventHandler.Do_Abi_Invoke(System.IntPtr thisPtr, System.IntPtr sender, System.IntPtr e) Line 43215	C#
```

### Screenshots

_No response_

### NuGet package version

1.0.2

### Packaging type

Unpackaged

### Windows version

Windows 10 version 21H1 (19043, May 2021 Update)

### IDE

Visual Studio 2022-preview

### Additional context

cc @mattleibow

DrusTheAxe:
Fixes https://task.ms/36717508

Add PACKAGE_FILTER_STATIC|DYNAMIC in `ResolveThirdPartyType()` so metadata resolution also doesn't just search static package graph entries (and fail to find metadata in dynamic packages)

Kudos to @huichen123 for spotting the problem

Manually tested by applying privates to the UndockedRegFreeWinRT IntegratedTest (URFW_ExeUsingWinAppSDK.exe) and tracing code flow in the debugger for the various permutations of usage.

closes #2382


EHO-makai (dismissed):
Feels like there is a missing test or tests here?

dev/UndockedRegFreeWinRT/typeresolution.cpp
```diff
@@ -428,8 +428,8 @@ namespace UndockedRegFreeWinRT
         _COM_Outptr_opt_result_maybenull_ IMetaDataImport2** ppMetaDataImport,
         _Out_opt_ mdTypeDef* pmdTypeDef)
     {
-        // Walk the dynamic package graph looking for the requested metadata
-        const uint32_t filter{ PACKAGE_FILTER_HEAD | PACKAGE_FILTER_DIRECT | PACKAGE_FILTER_IS_IN_RELATED_SET | PACKAGE_FILTER_DYNAMIC };
+        // Walk the package graph looking for the requested metadata
+        const uint32_t filter{ PACKAGE_FILTER_HEAD | PACKAGE_FILTER_DIRECT | PACKAGE_FILTER_IS_IN_RELATED_SET | PACKAGE_FILTER_DYNAMIC | PACKAGE_FILTER_STATIC };
```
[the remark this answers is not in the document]
DrusTheAxe: Doc update is coming. In the meantime...

* PACKAGE_PROPERTY_STATIC - search the package graph's static entries
* PACKAGE_PROPERTY_DYNAMIC - search the package graph's dynamic entries

Packaged apps are born with entries in their package graph, aka static package graph entries, aka the static portion of the package graph or more simply, the static package graph.

Unpackaged apps are born with an empty package graph.

The Dynamic Dependency API adds entries dynamically to the package graph, aka dynamic package graph entries, aka the dynamic portion of the package graph or more simply, the dynamic package graph

For backwards compatibility reasons `GetCurrentPackageInfo[2](...filter...)` assumes if `filter` doesn't contain `PACKAGE_FILTER_DYNAMIC` then only look at the static package graph. IOW calling `GetCurrentPackageInfo[2]()` without `PACKAGE_FILTER_DYNAMIC` is implicitly the same as calling GetCurrentPackageInfo[2]()` with `PACKAGE_FILTER_STATIC`.

This is explained in the comment block in DynamicGetCurrentPackageInfo2()` in `dev\DynamicDependency\API\MddDetourPacckageGraph.cpp`:

```c++
    // Legacy callers may not be aware of dynamic packages and a common mistake is to assume PACKAGE_FILTER_HEAD will return the
    // main package as the first entry which is not going to be true when packages have been added dynamically to the head of the graph.
    // Maintain pre-dynamic behavior by requiring the caller to opt into receiving dynamic packages.
```

Thus `GetCurrentPackageInfo[2]()` are dynamic-oblivous unless explicitly told to consider dynamic package graph entries.

OTOH `GetCurrentPackageInfo3()` assumes the caller is dynamic-savvy so it doesn't make any such backwards compatibility game. If you specify neither option to `GetCurrentPackageInfo3()` it's equivalent to specifying both (i.e. search static+dynamic entries).

dev/UndockedRegFreeWinRT/typeresolution.cpp
```diff
@@ -428,40 +428,69 @@ namespace UndockedRegFreeWinRT
...
-        // Only find the type if the it is an unpackaged app. Packaged apps can have their exe on their package graph,
-        // which will allow type resolution against adjacent WinMDs.
-        if (hr == HRESULT_FROM_WIN32(APPMODEL_ERROR_NO_PACKAGE))
+        // Walk the package graph looking for the requested metadata
+        const uint32_t filter{ PACKAGE_FILTER_HEAD | PACKAGE_FILTER_DIRECT | PACKAGE_FILTER_IS_IN_RELATED_SET | PACKAGE_FILTER_DYNAMIC | PACKAGE_FILTER_STATIC };
+        uint32_t bufferLength{};
+        uint32_t packagesCount{};
+        bool processHasStaticPackageGraph{};
+        HRESULT hr{ HRESULT_FROM_WIN32(GetCurrentPackageInfo(filter, &bufferLength, nullptr, &packagesCount)) };
+        if (hr == HRESULT_FROM_WIN32(ERROR_INSUFFICIENT_BUFFER))
```
BenJKuhn, line 437: Since this code follows an early return pattern, it might be cleaner to finish the usage of hr very locally. It gets reused sporadically, which requires a bit more careful checking across the function.
{
 HRESULT hr{ HRESULT_FROM_WIN32(GetCurrentPackageInfo(filter, &bufferLength, nullptr, &packagesCount)) };
 if (hr != HRESULT_FROM_WIN32(ERROR_INSUFFICIENT_BUFFER))
 {
  return hr; // Also I think there's a will macro to still log / track the failure in this case
 }
}
// do the thing.

dev/UndockedRegFreeWinRT/typeresolution.cpp
```diff
@@ -428,40 +428,69 @@ namespace UndockedRegFreeWinRT
...
+            RETURN_IF_FAILED(GetProcessExeDir(&exeDir));
             hr = FindTypeInDirectoryWithNormalization(
                 pMetaDataDispenser,
                 pszFullName,
                 exeDir,
                 phstrMetaDataFilePath,
                 ppMetaDataImport,
                 pmdTypeDef);
-
             if (hr == RO_E_METADATA_NAME_NOT_FOUND)
```
BenJKuhn, line 482: RETURN_HR_IF would make this more concise and add WIL logging.

dev/UndockedRegFreeWinRT/typeresolution.cpp
BenJKuhn: Also, is there any chance of this succeeding with null buffer (e.g. can package info length be zero?)

BenJKuhn (approved):
Comments are suggestions / stylistic. Non-blocking, so signing off.

dev/UndockedRegFreeWinRT/typeresolution.cpp
DrusTheAxe: > Since this code follows an early return pattern, it might be cleaner to finish the usage of hr very locally

NP. `const HRESULT hr` and `hrFindBlahBlah` multiple/specific names are common practices to avoid reuse complexities.

DrusTheAxe: > return hr; // Also I think there's a will macro to still log / track the failure in this case

That would emit WIL messages even for lookups that work and successfully does-not-find what you're looking for. So we avoid false positive messages / noise but merely `return`.

But I changed the called `FindTypeInDirectoryWithNormalization` to use WIL checks (instead of silent `if (SUCCEEDED(hr))`) so if something does go _wrong_ we'll get some information. Probably should go over the whole URFW and update it to use WIL, but that's a bit out of scope here

DrusTheAxe: > is there any chance of this succeeding with null buffer (e.g. can package info length be zero?)

If `packageCount>0` then no, bufferLength is `>0`

No change of buffer overrun

dev/UndockedRegFreeWinRT/typeresolution.cpp
DrusTheAxe: Except we don't want to log a WIL message in this case. We could invert the logic

```
// ...comment block...
RETURN_HR_IF(hr, hr != RO_E_METADATA_NAME_NOT_FOUND);
return HRESULT_FROM_WIN32(APPMODEL_ERROR_NO_PACKAGE);
```

but reading double-negatives is a bit more cryptic so not a helpful change in this case

DrusTheAxe:
> Feels like there is a missing test or tests here?

Tests exist in the aggregator pipeline

As noted in the description, I manually tested by using those tests to see the repro fail, then updated microsoft.windowsappruntime.dll and watched the test succeed. Plus a lot of interesting walk-the-debugger-through-the-code.

Functional tests will be added here (under `/test/UndockedRegFreeWinRT/`) after we address some issues in our test infrastructure.

PR #1676, opened 2021-10-28
Notify the Shell when association data changes.

aeloros:
When changes happen to associations, the shell must be notified.  If it is not, then it may take a long time to show updated icons or other resources from the registration change.  This updates the AppLifecycle APIs to automatically notify the shell every time a registration happens.

### DIFF
--- a/dev/AppLifecycle/Association.cpp
+++ b/dev/AppLifecycle/Association.cpp
@@ -461,4 +461,9 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
             }
         }
     }
+
+    void NotifyShellAssocChanged()
+    {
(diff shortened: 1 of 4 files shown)


dev/AppLifecycle/Association.cpp
```diff
@@ -461,4 +461,9 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
             }
         }
     }
+
+    void NotifyShellAssocChanged()
+    {
+        SHChangeNotify(SHCNE_ASSOCCHANGED, SHCNF_IDLIST, 0, 0);
```
ChrisGuzak, line 467: note, this is a very expensive notification, causing the taskbar to re-pain on Windows 10 and below. use it carefully.

ChrisGuzak: the last 2 params are pointers, thus `nullptr`

Issue #1121, opened 2021-07-26
WAS runtime cannot be found/loaded by apps on Windows Server 2019 LTSC


riverar:
**Describe the bug**
After installing the WAS runtime on Windows Server 2019 LTSC, apps attempting to bootstrap cannot locate the WAS runtime.

Are W.A.M.AppExtensions supported on LTSC? This doesn't work either (PowerShell):
```powershell
[Windows.ApplicationModel.AppExtensions.AppExtensionCatalog,Windows.ApplicationModel,ContentType=WindowsRuntime]::Open(
"com.microsoft.reunion.ddlm-0.8-x64-p")
```

**Steps to reproduce (C++)**
TBD

**Steps to reproduce (Rust)**
1. Enable sideloading of packages via Settings
2. Deploy runtime .\ProjectReunion-0.8Preview-Install-x64.exe
3. `git clone https://github.com/microsoft/windows-samples-rs/`
4. `cd windows-samples-rs/windows_app_sdk/skeleton`
5. `cargo run`

**Steps to reproduce (Pre-compiled, x64)**

1. Enable sideloading of packages via Settings
2. Deploy runtime .\ProjectReunion-0.8Preview-Install-x64.exe
3. Download [precompiled_skeleton.zip](https://github.com/microsoft/WindowsAppSDK/files/6876148/precompiled_skeleton.zip)
4. Run `skeleton.exe`
5. Observe (self-imposed) error due to internal DDLM 0x80070490 "Element not found."

ProjectReunion-0.8Preview-Install-x64.exe
Windows Server 2019 17763.1

Zymlex:
Same problem.

DrusTheAxe:
Manifested `windows.appExtension` is in the uap3 namespace (i.e. a long time ago) and the `AppExtension` API is likewise supported for a long time so it's probably a server-specific issue.

>After installing the WAS runtime on Windows Server 2019 LTSC

Did the installer successfully install the packages? You can verify from a command line
```
$(get-appxpackage *reunion*).packagefullname
```
and you should see the 2x Frameworks (x86+x64), 1x Main (x64) and 2x DDLM packages (x86+x64).

Do you have the Desktop Experience(?) enabled?

Zymlex:
```
Microsoft.ProjectReunion.0.8-preview_8000.146.628.0_x86__8wekyb3d8bbwe
[...]

PR #1681, opened 2021-10-29
Unpackaged Windows App SDK app doesn't run on RS5/19H1

DrusTheAxe:
https://task.ms/36777674

The bootstrapper relies on AppExtension to enumerate candidate DDLMs for the specified criteria. However AppExtension.Find*(name) only matches packages on <=19H1 if the caller is packaged and declares a matching AppExtensionHost (this must-have-matching-AppExtensionHost requirement was removed in 20H1). Since unpackaged apps have no `appxmanifest.xml` to declare an AppExtensionHost `MddBootstrapInitalize()` fails on RS5+19H1.

Bootstrap implementation is altered to workaround this AppExtension/RS5+19H1 limitation. The spec will be updated but here's the synopsis

1. Bootstrap enumerates potential DDLMs via `PackageManager.FindsPackagesByPackageType(framework)` instead of `AppExtension.OpenCatalog(ddlm)`
2. Bootstrap uses a DesktopBridge application process (`DynamicDependencyLifetimeManagerShadow.exe`) instead of a PackagedCOM OOP Server to inform Deployment the WinAppSDK Framework package-in-use
3. the package-in-use process quits when...

...it's told to quit via an named event (i.e. `MddBootstrapShutdown` signals to it)
or
...the caller process terminated (w/o calling `MddBootstrapShutdown`) instead of last COM reference to the PackagedCOM OOP Server was Release'd
DDLMshadow waits on the named event (signal from MddBootstrapInitialize) and caller process handle (`OpenProcess`) and quits when either pops.

`MddBootstrapInitalize()` activates the DDLMshadow app with arguments including the caller's process id (so DDLMshadow can get a handle to the process) and a unique id (to ensure the caller process hasn't quit and processid reused for a new process before DDLMshadow uses it).

This results in N DDLM processes (1 per Bootstrap-calling process) instead of 1 DDLM process (the PackagedCOM
 OOP Server for all Bootstrap-calling processes). IOW the DDLMshadow app process is multi-instanced vs DDLM PackagedCOM process is single-instanced.

NOTE: Parts of this will be used to address the elevation issue (#567). More is needed to solve that but this work isn't entirely just-for-downlevel :-)

This "DDLMviaEnumeration" codepath is used instead of the DDLMviaAppExtension when running on <=19H1 -or- the environment variable `MICROSOFT_WINDOWSAPPRUNTIME_DDLM_ENUMERATION` exists (e.g. `set MICROSOFT_WINDOWSAPPRUNTIME_DDLM_ENUMERATION=1`). The latter was used to verify same net behavior on >19H1 despite different implementation.

All related UnitTests pass as expected.

Additional E2E tests are being added in the aggregator as a separate PR.

NOTE: This change requires the aggregator be updated adding a file to DDLM MSIX packages named `Microsoft.WindowsAppRuntime.Release=<rmajor>.<rminor>.<rpatch>` specifying the release version (e.g. 1.0.0). This is necessary as the MSIX package versioning scheme currently does not specify the `rmajor` in the package's identity. This was handled via AppExtension but lacking that is only available if we "Find, Load and Parse appxmanifest.xml" at runtime &lt;shudder&gt;. This file let's use much more efficiently determine the release major version, necessary so we match the expected release e.g. so `MddBootstrapInitialize(0x00010002,...) doesn't match 0.2.*, 2.2.*, ...


dev/DynamicDependency/DynamicDependencyLifetimeManagerShadow/winmain.cpp
```diff
@@ -0,0 +1,71 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#include "pch.h"
+
+const HRESULT WINDOWSAPPSDK_DDLM_SHUTDOWN{ MAKE_HRESULT(SEVERITY_SUCCESS, FACILITY_ITF, 0x8000) };
+const HRESULT WINDOWSAPPSDK_DDLM_CALLER_TERMINATED{ MAKE_HRESULT(SEVERITY_SUCCESS, FACILITY_ITF, 0x8001) };
+const HRESULT WINDOWSAPPSDK_DDLM_CALLER_NOT_FOUND{ MAKE_HRESULT(SEVERITY_SUCCESS, FACILITY_ITF, 0x8002) };
+
+int WINAPI WinMain(HINSTANCE /*hInstance*/, HINSTANCE /*hPrevInstance*/, PSTR /*lpCmdLine*/, int /*nCmdShow*/)
```
huichen123, line 10: >WinMain [](http://example.com/codeflow?start=11&length=7)

wondering how this app informs deployment WAS is in use?

riverar:
Sounds like this is also related to https://github.com/microsoft/WindowsAppSDK/issues/1121

dev/DynamicDependency/DynamicDependencyLifetimeManagerShadow/winmain.cpp
DrusTheAxe: It's a DesktopBridge application so it's a packaged process, which suffices to inform Deployment a running process has a dependency on WARfwk and thus WARfwk is in use.

The PackagedCOM OOP Server serves the same purpose. A process running with package identity and a manifested dependency on WARfwk.

See the added application Id="DDLM" in `test/DynamicDependency/data/DynamicDependencyLifetimeManager.Msix/appxmanifest.xml`. That's just the test package; adding an equivalent snippet to the DDLM package's manifest in the Aggregator repo

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -251,7 +282,105 @@ void RemoveFrameworkFromPath(PCWSTR frameworkPath)
...
+    auto eventName{ wil::str_printf<wil::unique_cotaskmem_string>(L"%u;%s;%s", GetCurrentProcessId(), packageFullName.c_str(), c_uniqueIdAsString.c_str()) };
+    wil::unique_event event;
+    event.create(wil::EventOptions::ManualReset, eventName.get());
+
+    WCHAR lifetimeManagerApplicationUserModelId[APPLICATION_USER_MODEL_ID_MAX_LENGTH]{};
+    uint32_t lifetimeManagerApplicationUserModelIdLength{ static_cast<uint32_t>(ARRAYSIZE(lifetimeManagerApplicationUserModelId)) };
+    PCWSTR c_packageRelativeApplicationId{ L"DDLM" };
+    THROW_IF_WIN32_ERROR(FormatApplicationUserModelId(packageFamilyName.c_str(), c_packageRelativeApplicationId, &lifetimeManagerApplicationUserModelIdLength, lifetimeManagerApplicationUserModelId));
+
+    wil::com_ptr_nothrow<IApplicationActivationManager> aam{
```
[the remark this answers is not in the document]
DrusTheAxe: `wil::CoCreateInstance` throws on error.

As opposed to `wil::CoCreateInstanceNoThrow` and `wil::CoCreateInstanceFailFast`

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -251,7 +282,105 @@ void RemoveFrameworkFromPath(PCWSTR frameworkPath)
...
+        wil::CoCreateInstance<IApplicationActivationManager>(CLSID_ApplicationActivationManager, /*CLSCTX_LOCAL_SERVER*/CLSCTX_INPROC_SERVER)
```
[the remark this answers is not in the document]
DrusTheAxe: Leftover from earlier iteration. Removing.

Thanks!

dev/Common/IsWindowsVersion.h
```diff
@@ -0,0 +1,50 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#ifndef __ISWINDOWSVERSION_H
```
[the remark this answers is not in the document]
DrusTheAxe: VS detects the classic `#ifndef...#define...#endif` pattern and treats it equivalent to `#pragma once` (same perf benefit)

Since `#pragma` behavior isn't guaranteed to be standard and the classic pattern gives the same results it seems the better approach. Plus there are some cases you want to multiply include a header file with different/more behavior (WIL does this).

EHO-makai (approved):

dev/DynamicDependency/DynamicDependencyLifetimeManagerShadow/winmain.cpp
```diff
@@ -0,0 +1,71 @@
...
+        RETURN_LAST_ERROR_MSG("DDLM.Wait: %ls", eventName);
+    }
+    else
+    {
+        RETURN_HR_MSG(E_UNEXPECTED, "DDLM.Wait.Unexpected: %ls", eventName);
+    }
+    return 0;
+}
+
+STDAPI_(BOOL) DllMain(_In_opt_ HINSTANCE hinst, DWORD reason, _In_opt_ void*)
```
Scottj1s, line 64: why?

dev/DynamicDependency/DynamicDependencyLifetimeManagerShadow/winmain.cpp
```diff
@@ -0,0 +1,71 @@
...
+    // Open the named event created by the process calling the Bootstrap API (the process that caused us to run)
+    // If we can't it's because the process has terminated (or already called MddBootstrapShutdown())
+    // so we have nothing to do and we're out.
+    wil::unique_event_nothrow endOfTheLine;
+    RETURN_HR_IF_MSG(WINDOWSAPPSDK_DDLM_CALLER_NOT_FOUND, !endOfTheLine.try_open(eventName), "DDLM EventName=%ls", eventName);
+
+    // Open a handle to the calling process so we can watch for its termination
+    // (in case the process terminates without calling MddBootstrapShutdown())
+    wil::unique_handle callerProcessHandle{ OpenProcess(SYNCHRONIZE, FALSE, callerProcessId) };
+    RETURN_LAST_ERROR_IF_NULL(callerProcessHandle);
```
Scottj1s, line 39: consider standardized error reporting, via HRESULT_FROM_WIN32

dev/DynamicDependency/DynamicDependencyLifetimeManagerShadow/winmain.cpp
DrusTheAxe: Ummm leftover from a copy/paste?

Deleting. Thanks!

dev/Common/IsWindowsVersion.h
```diff
@@ -0,0 +1,50 @@
...
+#define __ISWINDOWSVERSION_H
+
+#include <VersionHelpers.h>
+
+namespace WindowsVersion
+{
+VERSIONHELPERAPI IsWindowsVersionOrGreaterEx(
+    const WORD majorVersion,
+    const WORD minorVersion,
+    const WORD servicePackMajor,
```
jonwis, line 14: Other readers - `VERSIONHELPERAPI` is already `static`

dev/DynamicDependency/DynamicDependencyLifetimeManagerShadow/winmain.cpp
DrusTheAxe: Error reporting _is_ standardized. OpenProcess is a BOOL/GetLastError Win32 API. `RETURN_LAST_ERROR_IF_NULL` returns `HRESULT_FROM_WIN32(GetLastError(x))` if x is null. Name might be a little misleading if you forget Win32 error land called it LastError but all good

dev/DynamicDependency/DynamicDependencyLifetimeManagerShadow/winmain.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+    // Wait for our calling process to tell us to quit OR our calling process terminates.
+    // Either way, we've served our purpose and should quit.
+    HANDLE waitForHandles[]{ endOfTheLine.get(), callerProcessHandle.get() };
+    auto rc{ WaitForMultipleObjects(static_cast<DWORD>(ARRAYSIZE(waitForHandles)), waitForHandles, FALSE, INFINITE) };
```
jonwis, line 44: Teeny nit: This `static_cast<DWORD>()` shouldn't be required.

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -251,7 +282,105 @@ void RemoveFrameworkFromPath(PCWSTR frameworkPath)
...
+    std::wstring packageFullName;
+    FindDDLMViaEnumeration(majorMinorVersion, versionTag, minVersion, packageFamilyName, packageFullName);
+
+    // Create the named event used later to signal to the lifetime manager it's time to quit
+    // The named event has the syntax: "<processid>;<packagefullname>;<uniqueid>"
+    GUID uniqueId{};
+    THROW_IF_FAILED(CoCreateGuid(&uniqueId));
+    const auto c_uniqueIdAsString{ winrt::to_hstring(uniqueId) };
+    auto eventName{ wil::str_printf<wil::unique_cotaskmem_string>(L"%u;%s;%s", GetCurrentProcessId(), packageFullName.c_str(), c_uniqueIdAsString.c_str()) };
+    wil::unique_event event;
```
jonwis, line 363: `wil::unique_event` has a ctor that accepts these options, so you can just say:

`wil::unique_event event { wil::EventOptions::ManualReset }`

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -251,7 +282,105 @@ void RemoveFrameworkFromPath(PCWSTR frameworkPath)
...
+    event.create(wil::EventOptions::ManualReset, eventName.get());
+
+    WCHAR lifetimeManagerApplicationUserModelId[APPLICATION_USER_MODEL_ID_MAX_LENGTH]{};
+    uint32_t lifetimeManagerApplicationUserModelIdLength{ static_cast<uint32_t>(ARRAYSIZE(lifetimeManagerApplicationUserModelId)) };
```
jonwis, line 367: Generally, `static_cast` of `ARRAYSIZE` shouldn't be necessary.

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -251,7 +282,105 @@ void RemoveFrameworkFromPath(PCWSTR frameworkPath)
...
+    std::wstring packageFamilyName;
+    std::wstring packageFullName;
+    FindDDLMViaEnumeration(majorMinorVersion, versionTag, minVersion, packageFamilyName, packageFullName);
+
+    // Create the named event used later to signal to the lifetime manager it's time to quit
+    // The named event has the syntax: "<processid>;<packagefullname>;<uniqueid>"
+    GUID uniqueId{};
+    THROW_IF_FAILED(CoCreateGuid(&uniqueId));
+    const auto c_uniqueIdAsString{ winrt::to_hstring(uniqueId) };
+    auto eventName{ wil::str_printf<wil::unique_cotaskmem_string>(L"%u;%s;%s", GetCurrentProcessId(), packageFullName.c_str(), c_uniqueIdAsString.c_str()) };
```
Scottj1s, line 362: does this need to have "Global\" prefix for inter-session access?

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
DrusTheAxe: Neat! For years that produced compiler warnings when I built x64 (ARRAYSIZE=size_t=uint64 but DWORD=uint32) but something must have changed as it now builds fine. Progress marches on :-)

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
DrusTheAxe: Do OOP COM servers offer inter-session access?

This is per-user and it's almost impossible to have the same user logged in twice, or at least it used to be. Has that changed lately?

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
DrusTheAxe: Cute. More useful if it wasn't a named event (`event.create()`) but good to know.

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -337,10 +466,188 @@ CLSID FindDDLM(
...
+    std::wstring& ddlmPackageFamilyName,
+    std::wstring& ddlmPackageFullName)
+{
+    // Find the best fit
+    bool foundAny{};
+    PACKAGE_VERSION bestFitVersion{};
+    winrt::hstring bestFitPackageFamilyName{};
+    winrt::hstring bestFitPackageFullName{};
+
+    // We need to look for DDLM packages in the package family for release <major>.<minor> and <versiontag>
```
Scottj1s, line 487: nice comment!

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -337,10 +466,188 @@ CLSID FindDDLM(
...
+    const UINT16 majorVersion{ HIWORD(majorMinorVersion) };
+    const UINT16 minorVersion{ LOWORD(majorMinorVersion) };
+    WCHAR packageNamePrefix[PACKAGE_NAME_MAX_LENGTH + 1]{};
+    if (!g_test_ddlmPackageNamePrefix.empty())
+    {
+        FAIL_FAST_IF_FAILED(StringCchCopyW(packageNamePrefix, ARRAYSIZE(packageNamePrefix), g_test_ddlmPackageNamePrefix.c_str()));
+    }
+    else
+    {
+        wsprintf(packageNamePrefix, L"microsoft.winappruntime.ddlm.%hu.", minorVersion);
```
Scottj1s, line 509: consider wsprintf_s

dev/Common/IsWindowsVersion.h
DrusTheAxe: Yep. Surprised me when first cooking the bits but conveniently F12 in VS quickly answered why `static VERSIONHELPERAPI` caused a compiler error

dev/DynamicDependency/DynamicDependencyLifetimeManagerShadow/winmain.cpp
DrusTheAxe: Not anymore

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
DrusTheAxe: Q: Is wsprintf_s or StringCchPrintfW preferable these days?

There's a few references. I'll see about tweaking them in a future PR

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -337,10 +466,188 @@ CLSID FindDDLM(
...
+        const auto packageName{ packageId.Name() };
+        const auto packageNameLength{ packageName.size() };
+        if (packageNameLength < packageNamePrefixLength + packageNameSuffixLength)
+        {
+            // The package's Name can't match the expected prefix and/or suffix. Skip it
+            continue;
+        }
+        if (CompareStringOrdinal(packageName.c_str(), static_cast<int>(packageNamePrefixLength), packageNamePrefix, static_cast<int>(packageNamePrefixLength), TRUE) != CSTR_EQUAL)
+        {
+                // The package's Name prefix doesn't match the expected value. Skip it
```
Scottj1s, line 545: indentation (also lines 553,4)

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -337,10 +466,188 @@ CLSID FindDDLM(
...
+        //       to work down to RS5. So instead we'll use GetPackagePathByFullName()
+        //       as that exists since Win81 (and can be significantly faster than
+        //       Package.InstalledLocation).
+        const auto packageFullName{ packageId.FullName() };
+        wil::unique_cotaskmem_string packagePathBufferDynamic;
+        uint32_t packagePathLength{};
+        const auto rc{ GetPackagePathByFullName(packageFullName.c_str(), &packagePathLength, nullptr) };
+        if (rc != ERROR_INSUFFICIENT_BUFFER)
+        {
+            THROW_HR_MSG(HRESULT_FROM_WIN32(rc), "Enumeration: %ls", packageFullName.c_str());
```
Scottj1s, line 577: there are a couple throws in the loop, which would short-circuit any later success - should these be continues?

DrusTheAxe: No. General rule of thumb, error out on errors.

Possible outcomes some line of code:

* Success
* Failure
* Error

Success + Failure (e.g. not found) are expected cases we handle
Error (out-of-memory and others) terminate processing and return to caller. Errors aren't swallowed/ignored; trying to proceed could do more harm than good.

There's (pardon the pun) exceptions to these rules but that's the general practice.

DrusTheAxe:
P.S. The PR's titled RS5/19H1 but the problem's also on 19H2 i.e. Windows &lt;20H1 is affected

PR #1685, opened 2021-10-30
feat: Added global:: to MddBootstrapAutoInitializer.

### LINKED ISSUE #1686: MddBootstrapAutoInitializer does not compile with enabled ImplicitUsings
Bug with the intersection of Microsoft.WindowsAppSDK.Runtime.Version with System.Version.

It needs to use `global::` prefixes.

Env:
`Microsoft.WindowsAppSDK`: `1.0.0-preview3`
`WindowsPackageType`: `None`

HavenDV:
to support `ImplicitUsings` and fix a bug with the intersection of `Microsoft.WindowsAppSDK.Runtime.Version` with `System.Version`.

Fixes #1686

### DIFF
--- a/dev/WindowsAppRuntime_BootstrapDLL/MddBootstrapAutoInitializer.cs
+++ b/dev/WindowsAppRuntime_BootstrapDLL/MddBootstrapAutoInitializer.cs
@@ -1,23 +1,19 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
-using System.Runtime.CompilerServices;
-using Microsoft.WindowsAppSDK;
-using Microsoft.WindowsAppSDK.Runtime;
-
 namespace Microsoft.Windows.ApplicationModel.DynamicDependency.BootstrapCS
 {
     class AutoInitialize
     {
-        [ModuleInitializer]
+        [global::System.Runtime.CompilerServices.ModuleInitializer]
         internal static void AccessWindowsAppSDK()
         {
-            uint majorMinorVersion = Release.MajorMinor;
-            string versionTag = Release.VersionTag;
-            var minVersion = new PackageVersion(Version.UInt64);
+            uint majorMinorVersion = global::Microsoft.WindowsAppSDK.Release.MajorMinor;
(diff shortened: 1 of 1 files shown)


DrusTheAxe (approved):

riverar:
@andrewleader If this gets approved, it sets a precedent that Windows App SDK supports .NET implicit usings and you may need to create some governance around type naming, avoiding clashes across namespaces, etc. if it doesn't already exist.

DrusTheAxe (changes requested):
Will discuss with the team tomorrow

DrusTheAxe:
> ⚠️ @andrewleader Andrew Leader (HE/HIM) FTE If this gets approved, it sets a precedent that Windows App SDK supports .NET _implicit usings_ and you may need to create some governance around type naming, avoiding clashes across namespaces, etc. if it doesn't already exist.

As I understand it the code here neither endorses nor rejects ImplicitUsings. The references in here are fully scoped so _regardless_ if ImplicitUsings are used or not, this code works as expected. Or am I missing something here?

This is a very WinAppSDK perspective -- how we implement WinAppSDK shouldn't require you use (or avoid) a specific technology; au contraire, we aspire to work with all tech devs may choose.

You raise an interesting issue re ImplicitUsings but it's a broader issue, for any C# code provided to an app as source e.g. wouldn't the generated WinUI main source file face the same issues?

riverar:
> The references in here are fully scoped so _regardless_ if ImplicitUsings are used or not, this code works as expected. Or am I missing something here?

Looks like it blew up hence the PR, see https://github.com/microsoft/WindowsAppSDK/issues/1686.

RealTommyKlein:
@Scottj1s - no, the Xaml compiler fully qualifies all types with the `global::` prefix, like what this PR does (FWIW, LGTM 👍)

DrusTheAxe (approved):
LGTM, and ditto @RealTommyKlein and @Scottj1s

DrusTheAxe:
@HavenDV thanks for reporting the issue and offering a fix. We discussed offline the issue and fix (goodness) and potentially related concerns (turns out not a problem). I approved the fix (thanks!) and kicked off a run of the build pipeline. Assuming that completes successfully it can be completed. I'll check back here to see how it goes

kinkycat91 (commented):
Jju

PR #1698, opened 2021-11-01
API spec for new DirectWrite font selection APIs

niklasb-ms:


specs/dwritecore/DWriteCore.md
```diff
@@ -78,33 +78,275 @@ with interoperability issues.
...
 restriction enables DWriteCore to efficiently construct the system font collection without depending on a
 cross-process or persistent cache. An example use case for the new factory type would be a web browser
 rendering process with very limited permissions.
 
-A _bitmap render target_ (IDWriteBitmapRenderTarget) is an API object that encapsulates a system memory bitmap
+# Text Rendering
+
+A _bitmap render target_ (`IDWriteBitmapRenderTarget`) is an API object that encapsulates a system memory bitmap
 and enables rendering glyphs to the bitmap. On Windows, the object encapsulates a GDI memory DC with a GDI
 device-dependent bitmap (DIB) selected into it. However, on non-Windows platforms, there is no GDI DC,
```
PeterCon, line 100: Is the DC a device-*dependent* or device-*independent* bitmap? ("DIB" suggests the latter)

specs/dwritecore/DWriteCore.md
```diff
@@ -78,33 +78,275 @@ with interoperability issues.
...
+issues. The methods exposed by the `IDWriteFontSet4` interface implement a hybrid approach that offers the
+advantages of the typographic font family model while mitigating compatibility issues.
+
+The following sections compare the older font family models with the typographic font family model,
+explain compatibility challenges posed by changing font family models, and finally explains how those
+challenges can be overcome by using the `IDWriteFontSet4` methods.
+
+## RBIZ Font Family Model
+
+The font family model used by GDI is sometimes called the "four-font model" or "RBIZ" model. Each font
```
PeterCon, line 128: To be accurate, this wasn't GDI's family model. For example, GDI can support a family with multiple weights (not just RB) provided name IDs 1 and 2 are set accordingly in the fonts. RBIZ was a model assumed early on by apps (Word) and it was compatibility with apps that led font developers to assume name IDs 1 and 2, and GDI, *require* the RBIZ model.

I'd word this, "... used by GDI apps..."

specs/dwritecore/DWriteCore.md
```diff
@@ -78,33 +78,275 @@ with interoperability issues.
...
+a time when computer font families rarely had more than two weights. However, a side effect was that
+subsequent fonts could only have two weights within an RBIZ family if they wanted to work with
+existing applications.
+
+Suppose you wanted to add a heavier "Black" weight to the Arial font family. Logically, this font is
+part of the Arial family, so you might expect to select it by setting `lfFaceName` to "Arial" and 
+`lfWeight` to `FW_BLACK`. However, there is no way for an application user to choose between three
+weights using a two-state bold button. The solution was to give the new font a different family name,
+so the user could select it by choosing "Arial Black" from the list of font families. Likewise, there
+is no way to choose from among different widths in the same font family if your only affordances are
```
PeterCon, line 159: Who's the target audience? The term "affordance" might not be familiar to all readers. (Just recently I used the term and then had to reword when someone asked for clarification.) You might say, "UI controls" instead.

specs/dwritecore/DWriteCore.md
```diff
@@ -78,33 +78,275 @@ with interoperability issues.
...
 
+# Font Selection
+
+The new `IDWriteFontSet4` interface exposes new methods for selecting fonts from a font set. The new methods
+make it possible to transition to the _typographic font family model_ while maintaining compatibility with
+existing applications, documents, and fonts.
+
+Font selection (sometimes called font matching or font mapping) is the process of selecting the available
+fonts that best match input parameters passed in by an application. The input parameters are sometimes
+referred to collectively as a _logical font_. A font selection algorithm matches the logical font (the
```
PeterCon, line 112: Anticipating discussion that follows, you might insert: "... as a _logical font_. A logical font includes a font family name plus other attributes indicating a particular typeface within the family."

specs/dwritecore/DWriteCore.md
```diff
@@ -78,33 +78,275 @@ with interoperability issues.
...
+
+## Weight-Stretch-Style Font Family Model
+
+The weight-stretch-style font family model is the original font family model used by DirectWrite before
+the typographic font family model was introduced. It is also known as weight-width-slope (WWS). In the
+WWS model, fonts within the same family can be differented by three properties: weight (`DWRITE_FONT_WEIGHT`),
+stretch (`DWRITE_FONT_STRETCH`), and style (`DWRITE_FONT_STYLE`).
+
+The WWS model is more flexible than the RBIZ model in two ways. First, fonts in the same family can
+be differentiated by stretch (or width). Second, there can be more than two weights in the same family.
```
PeterCon, line 181: Perhaps append, "..., as well as weight and italic (style)."

specs/dwritecore/DWriteCore.md
```diff
@@ -78,33 +78,275 @@ with interoperability issues.
...
+Advantages of the typographic font family model are:
+
+  - Fonts can be grouped into families as intended by the designer, rather than being split into
+    subfamilies due to limitations of the font family model.
+
+  - An application can automatically select the correct "opsz" axis value based on the font size
+    rather than exposing different optical sizes to the user as different font families.
+
+  - Arbitrary instances of variable fonts can be selected. For example, if a variable font supports
+    weights in the range 100-900, the typographic model can select _any_ weight in this range. The
```
PeterCon, line 240: In case the reader isn't familiar with variable fonts, you might say, "... in the continuous range 100–900"

specs/dwritecore/DWriteCore.md
```diff
@@ -78,33 +78,275 @@ with interoperability issues.
...
+------------------------|--------------|-----------|-----------|--------|---------------|-------|-------|-------|------
+Legacy-Regular.ttf      | Legacy       | 400       | 5         | 0      | Legacy        | 400   | 100   | 0     | 0
+Legacy-Bold.ttf         | Legacy       | 700       | 5         | 0      | Legacy        | 700   | 100   | 0     | 0
+Legacy-Black.ttf        | Legacy       | 900       | 5         | 0      | Legacy        | 900   | 100   | 0     | 0
+Legacy-Soft.ttf         | Legacy Soft  | 400       | 5         | 0      | Legacy        | 400   | 100   | 0     | 0
+Legacy-SoftBold.ttf     | Legacy Soft  | 700       | 5         | 0      | Legacy        | 700   | 100   | 0     | 0
+Legacy-SoftBlack.ttf    | Legacy Soft  | 900       | 5         | 0      | Legacy        | 900   | 100   | 0     | 0
+
+The "Legacy" typographic family has three weights and each weight has regular and "Soft" variants.
+However, these fonts predate OpenType 1.8, so there is no "SOFT" axis, or any other axis that cannot
```
PeterCon, line 282: You might insert, "If these were new fonts, they could be implemented as declaring a SOFT design axis."

specs/dwritecore/DWriteCore.md
```diff
@@ -78,33 +78,275 @@ with interoperability issues.
...
+Document and application compatibility issues are mitigated by allowing different types of font
+names to be specified and also by providing a method of deriving axis values from weight, stretch,
+and style parameters.
+
+Suppose a document specifies a family name, weight, stretch, and style, but no axis values. The
+application can first map the weight, stretch, style, and font size to axis values by calling the
+`IDWriteFontSet4::GetDerivedFontAxisValues` method. The application can then pass both the family
+name and axis values to `IDWriteFontSet4::GetMatchingFonts`. `GetMatchingFonts` returns a list of
+matching fonts in priority order, and the result is appropriate whether the specified family name
+is a typographic family name, weight-stretch-style family name, RBIZ family name, or full name.
```
PeterCon, line 309: Since you've been using WWS elsewhere already, you might say "WWS family name" here.

specs/dwritecore/DWriteCore.md
```diff
@@ -78,33 +78,275 @@ with interoperability issues.
...
+If the specified family has an "opsz" axis, the appropriate optical size is automatically selected
```
PeterCon, line 310: "If the specified family has an 'opsz' axis..." and no specific opsz value is specified?

specs/dwritecore/DWriteCore.md
```diff
@@ -78,33 +78,275 @@ with interoperability issues.
...
+
+The hybrid font selection model is implemented by the following `IDWriteFontSet4` methods:
+
+  - The `GetDerivedFontAxisValues` method converts font size, weight, stretch, and style parameters
+    to the corresponding axis values. Any explicit axis values passed in by the client are excluded
+    from the derived axis values.
+
+  - The `GetMatchingFonts` method returns a prioritized list of matching fonts given a family name
+    and array of axis values. As described above, the family name parameter can be a typographic
+    family name, weight-stretch-style family name, RBIZ family name, or full name. (Full name is
```
PeterCon, line 329: "WSS family name"?

specs/dwritecore/DWriteCore.md
```diff
@@ -78,33 +78,275 @@ with interoperability issues.
...
+ - New methods for selecting fonts from a font set
 
-The new DWRITE_FACTORY_TYPE_ISOLATED2 enum value complements the existing factory types. It is similar to
-DWRITE_FACTORY_TYPE_ISOLATED, but is more locked down in two ways. First, it only caches data in-process, and
+# Factory Creation
+
+The new `DWRITE_FACTORY_TYPE_ISOLATED2` enum value complements the existing factory types. It is similar to
+`DWRITE_FACTORY_TYPE_ISOLATED`, but is more locked down in two ways. First, it only caches data in-process, and
 will not attempt to either read from or write to a cross-process font cache or persisted cache file. Second,
 a restricted factory only enumerates well-known fonts in the system font collection. The well-known-fonts
```
PeterCon, line 91: Is this true only for Windows, or for other platforms as well?

specs/dwritecore/DWriteCore.md
niklasb-ms: Good catch. Should be device-independent.

specs/dwritecore/DWriteCore.md
niklasb-ms: Yes, I understand RBIZ is not actually GDI's font family model, but it's become the de facto model in the GD application ecosystem. I'm rewording this section to hopefully make that more clear.

specs/dwritecore/DWriteCore.md
niklasb-ms: It's easy to avoid the word affordance here.

specs/dwritecore/DWriteCore.md
niklasb-ms: Sounds good.

specs/dwritecore/DWriteCore.md
niklasb-ms: It's true on other platforms as well, but we haven't yet publicly released DWriteCore on other platforms, so details are TBD.

specs/dwritecore/DWriteCore.md
```diff
@@ -78,33 +78,272 @@ with interoperability issues.
...
+weight, stretch, style, and font size to the `IDWriteFontSet4::GetDerivedFontAxisValues` method
+to map the input parameters to _derived axis values_. If the application also has explicit axis
+values, it should pass these to `GetDerivedAxisValues` as well. The derived axis values will
+not include any axes there were passed in explicitly. This ensures that explicit axes take
+precedence over derived axis values.
+
+Once the application has axis values (derived, explicit, or both), it can pass the family name
+and axis values to the `IDWriteFontSet4::GetMatchingFonts` method. This method returns an
+`IDWriteFontSet4` representing a list of matching fonts in priority order. The specified family
+name can be a typographic family name, WWS family name, RBIZ family name, or full name.
```
jlnordin, line 314: >full name [](http://example.com/codeflow?start=77&length=9)

"full name" is later identified as a compatibility feature to GDI, but I'm not sure what a "full name" would look like. I think an example would be helpful.

specs/dwritecore/DWriteCore.md
```diff
@@ -208,10 +908,12 @@ enum DWRITE_FACTORY_TYPE
     // Warning: Make sure any additional new enum values are consistent between different versions of
     // this header in DWriteCore and the Windows SDK.
     DWRITE_FACTORY_TYPE_ISOLATED2
```
jlnordin, line 910: >DWRITE_FACTORY_TYPE_ISOLATED2 [](http://example.com/codeflow?start=4&length=29)

`ISOLATED2` doesn't tell me much about its behavior. Some options:

- `DWRITE_FACTORY_TYPE_ISOLATED_TO_PROCESS`
- `DWRITE_FACTORY_TYPE_ISOLATED_NO_CROSS_PROCESS_CACHE`
- `DWRITE_FACTORY_TYPE_RESTRICTED`

Using a number suffix for the interfaces makes sense to me, but we should be able to be a little more descriptive with the enumerant name.

For the record, this should not be considered a comprehensive list of suggestions -- I'm just hoping we can find a more descriptive name.

niklasb-ms: This enum value was added in February as part of https://github.com/microsoft/WindowsAppSDK/pull/198. I proposed the name DWRITE_FACTORY_TYPE_RESTRICTED, but Raymond Chen and other reviewers thought that was confusing and suggested ISOLATED2. You can see the comments in the original PR linked above.

specs/dwritecore/DWriteCore.md
niklasb-ms: Makes sense.

specs/dwritecore/DWriteCore.md
```diff
@@ -78,33 +78,272 @@ with interoperability issues.
 New APIs introduced by DWriteCore include:
 
  - The exported function is renamed `DWriteCoreCreateFactory` to avoid a naming collision with the
-   `DWriteCreateFactory` function exported by the Windows DLL (DWrite.dll).
- - A new DWRITE_FACTORY_TYPE enumeration for creating an isolated factory object
+   `DWriteCreateFactory` function exported by the system version of DirectWrite (DWrite.dll)
+ - A new `DWRITE_FACTORY_TYPE` enumeration value for creating an isolated factory object
```
riverar, line 82: Nit: DWRITE_FACTORY_TYPE here is an _enumeration_. It contains a number of _enumerators_.

```suggestion
 - A new `DWRITE_FACTORY_TYPE` enumeration for creating an isolated factory object
```

specs/dwritecore/DWriteCore.md
```diff
@@ -78,33 +78,272 @@ with interoperability issues.
...
  - A new method for getting pixel data from a bitmap render target
+ - New methods for selecting fonts from a font set
 
-The new DWRITE_FACTORY_TYPE_ISOLATED2 enum value complements the existing factory types. It is similar to
-DWRITE_FACTORY_TYPE_ISOLATED, but is more locked down in two ways. First, it only caches data in-process, and
+# Factory Creation
+
+The new `DWRITE_FACTORY_TYPE_ISOLATED2` enum value complements the existing factory types. It is similar to
```
riverar, line 88:
```suggestion
The new `DWRITE_FACTORY_TYPE_ISOLATED2` enum complements the existing factory types. It is similar to
```

specs/dwritecore/DWriteCore.md
```diff
@@ -78,33 +78,272 @@ with interoperability issues.
...
+    from the derived axis values.
+
+  - The `GetMatchingFonts` method returns a prioritized list of matching fonts given a family name
+    and array of axis values. As described above, the family name parameter can be a typographic
+    family name, WWS family name, RBIZ family name, or full name. (Full name is allowed because
+    GDI allows font matching by full name.)
+
+The following other DirectWrite APIs also use the hybrid font selection algorithm:
+
+  - [`IDWriteFontCollection2::GetMatchingFonts`](https://docs.microsoft.com/en-us/windows/win32/api/dwrite_3/nf-dwrite_3-idwritefontcollection2-getmatchingfonts)
```
riverar, line 331: Nit: Removed language from URL to avoid forcing non-English folks into English docs.

```suggestion
  - [`IDWriteFontCollection2::GetMatchingFonts`](https://docs.microsoft.com/windows/win32/api/dwrite_3/nf-dwrite_3-idwritefontcollection2-getmatchingfonts)
```

riverar (approved):
Fantastic docs, just a few tiny nits. Thanks!

niklasb-ms:
Thanks! I've incorporated your feedback. Thanks for taking a look!

---
In reply to: [798475190](https://github.com/microsoft/WindowsAppSDK/pull/1698#issuecomment-798475190) [](http://example.com/codeflow?ancestors=798475190,798475190)

specs/dwritecore/DWriteCore.md
```diff
@@ -78,33 +78,275 @@ with interoperability issues.
...
+ - A new `DWRITE_FACTORY_TYPE` enumerator for creating an isolated factory object
  - A new method for getting pixel data from a bitmap render target
+ - New methods for selecting fonts from a font set
+ - A new factory method to release unreferenced fonts
 
-The new DWRITE_FACTORY_TYPE_ISOLATED2 enum value complements the existing factory types. It is similar to
-DWRITE_FACTORY_TYPE_ISOLATED, but is more locked down in two ways. First, it only caches data in-process, and
+# Factory Creation
+
+The new `DWRITE_FACTORY_TYPE_ISOLATED2` enumerator complements the existing factory types. It is similar to
```
drott, line 89: Could "_ISOLATED2" become a more speaking name? Or is this not part of the current change? E.g. `_ISOLATED_IN_PROCESS` or `_ISOLATED_NO_OOP`?

specs/dwritecore/DWriteCore.md
```diff
@@ -78,33 +78,275 @@ with interoperability issues.
...
+
+  - An application can automatically select the correct "opsz" axis value based on the font size
+    rather than exposing different optical sizes to the user as different font families.
+
+  - Arbitrary instances of variable fonts can be selected. For example, if a variable font supports
+    weights in the continuous range 100-900, the typographic model can select _any_ weight in this range.
+    The older font family models can only choose the nearest weight from among the named instances defined
+    by the font.
+
+However, migrating to the typographic font family model raises potential compatibility issues. The
```
drott, line 245: From these paragraphs it's not 100% clear to me how the typographic family model captures both static instances with certain "classic" attributes and variable fonts. Perhaps it'd make sense to add some clarification on that? (I may have missed something).

specs/dwritecore/DWriteCore.md
```diff
@@ -78,33 +78,275 @@ with interoperability issues.
...
+to a "wdth" (`DWRITE_FONT_AXIS_TAG_WIDTH`) axis value. `DWRITE_FONT_STYLE` can be mapped to a combination
+of "ital" and "slnt" (`DWRITE_FONT_AXIS_TAG_ITALIC` and `DWRITE_FONT_AXIS_TAG_SLANT`) axis values. 
+
+Another registered axis is "opsz" (`DWRITE_FONT_AXIS_TAG_OPTICAL_SIZE`). An optical font family such as
+Sitka includes fonts that differ along the "opsz" axis, meaning they are designed to be used at different
+point sizes. The WWS font family model does not have an optical size axis, so the Sitka font family must
+be split into multiple WWS font families: "Sitka Small", "Sitka Text", "Sitka Subheading", and so on. Each
+WWS font family corresponds to a different optical size, and it's left to the user to specify the right
+WWS family name for a given font size. With the typographic font family model, the user can simply choose
+"Sitka", and the application can automatically set the "opsz" axis value based on the font size.
```
drott, line 230: How is "automatically" to be implemented? Would the application need to "rematch" the font for each size? Or would the application call font matching ones, then "opsz" would be implemented in rendering calls? 

It's hard for me to think of `opsz` as part of font matching, unless explicitly specified as search criterion.

specs/dwritecore/DWriteCore.md
```diff
@@ -78,33 +78,275 @@ with interoperability issues.
...
+
+Advantages of the typographic font family model are:
+
+  - Fonts can be grouped into families as intended by the designer, rather than being split into
+    subfamilies due to limitations of the font family model.
+
+  - An application can automatically select the correct "opsz" axis value based on the font size
```
drott, line 237: Similar to above: Could you elaborate on how "automatically" works?

specs/dwritecore/DWriteCore.md
```diff
@@ -78,33 +78,275 @@ with interoperability issues.
...
+
+  - Arbitrary instances of variable fonts can be selected. For example, if a variable font supports
+    weights in the continuous range 100-900, the typographic model can select _any_ weight in this range.
+    The older font family models can only choose the nearest weight from among the named instances defined
+    by the font.
+
+However, migrating to the typographic font family model raises potential compatibility issues. The
+next section describes these issues and how they are mitigated by the hybrid font selection model.
+
+## Hybrid Font Selection Model
```
drott, line 248: The CSS spec https://www.w3.org/TR/css-fonts-4/#font-matching-algorithm documents in detail the algorithm at how one arrives at a match, in a model that is similar to WWS. Perhaps it would be helpful to specify as well how this new DWrite API finds the nearest match? 

Does the output have some kind of matching score? Would it be possible to specify priorities, i.e. it's more important to find an exact match for dimension m than for dimension n, or some such?

I am thinking about how this implementation could be used from the browser so that the output matches the behavior of what the CSS spec expects - or whether the CSS spec can only be implemented by building own data structures.

specs/dwritecore/DWriteCore.md
```diff
@@ -78,33 +78,275 @@ with interoperability issues.
...
+
+  - "Legacy Soft" does not match a typographic family name, so only fonts in the WWS family
+    are considered. These can be differentiated by weight, so there is no ambiguity.
+
+  - "Legacy" matches both a typographic family and a WWS family. All fonts in the typographic
+    family are considered, but presence in the WWS family is used as a tie-breaker. For example,
+    Legacy Bold has the same weight as Legacy Soft Bold, but the former is preferred because
+    it is in the "Legacy" WWS family.
+
+Document and application compatibility issues are mitigated by allowing different types of font
```
drott, line 301: Do I understand correctly that this adaptation internally relies on string processing of the family name, i.e. detecting the "Soft" or "Black" suffix etc.?

specs/dwritecore/DWriteCore.md
```diff
@@ -78,33 +78,275 @@ with interoperability issues.
...
+example, these might come from a document's backing store. The application can first pass the
+weight, stretch, style, and font size to the `IDWriteFontSet4::GetDerivedFontAxisValues` method
+to map the input parameters to _derived axis values_. If the application also has explicit axis
+values, it should pass these to `GetDerivedAxisValues` as well. The derived axis values will
+not include any axes there were passed in explicitly. This ensures that explicit axes take
+precedence over derived axis values.
+
+Once the application has axis values (derived, explicit, or both), it can pass the family name
+and axis values to the `IDWriteFontSet4::GetMatchingFonts` method. This method returns an
+`IDWriteFontSet4` representing a list of matching fonts in priority order. The specified family
```
drott, line 315: Similar to above, it would be useful to know how the priority order is computed or whether there is control over the sorting logic (i.e. priority on a certain dimension/axis?)

specs/dwritecore/DWriteCore.md
```diff
@@ -170,8 +412,469 @@ DWRITE_BITMAP_DATA_BGRA32 TextRenderer::GetBitmapData(_In_ IDWriteBitmapRenderTa
...
+        << (allowSimulations ? L"true" : L"false")
+        << L"):\n";
+
+    // Get the matching fonts for the specified family name and axis values.
+    wil::com_ptr<IDWriteFontSet4> matchingFonts;
+    THROW_IF_FAILED(fontSet->GetMatchingFonts(
+        familyName,
+        axisValues.data(),
+        static_cast<UINT32>(axisValues.size()),
+        allowSimulations,
```
drott, line 466: As discussed in emails, it might make sense to have allowSimulation be an enum or bitmask to control what simulation type should be disallowed. (compare CSS font-synthesis property), i.e. only allowing italic synthetisation, only allowing bold synthetisation, (small-caps?).

drott (commented):
Some additional thoughts after the initial feedback in email. Thank you for sending out this proposal.

specs/dwritecore/DWriteCore.md
niklasb-ms: This is not part of the current change, but is good feedback. I was initially going to call this `DWRITE_FACTORY_TYPE_RESTRICTED` but this was thought confusing in its own way.

specs/dwritecore/DWriteCore.md
niklasb-ms: For variable fonts, an `IDWriteFontFace` object always represents a particular instance of the variable font. It may be a named instance or an arbitrary instance, but the axis values are specified at the time the font face object is created. That way, the `IDWriteFontFace` interface can work exactly the same for variable fonts as for static fonts when it comes to getting metrics, rendering glyphs, etc. Font matching returns a prioritized list of `IDWriteFontFace` objects based on the input family name and axis values. Thus, the mapping of input axis values to variable font instances happens at font matching time.

specs/dwritecore/DWriteCore.md
niklasb-ms: For matching purposes, a "distance" is computed along each axis between the requested and actual value. Well-known axes have a multiplier that is applied distance in order to give behavior as close as possible to under the weight-stretch-style model.

That said, it's possible none of the newer APIs are appropriate for HTML engines to use because the CSS spec explicitly defines a font matching algorithm based on weight, stretch, and style. How the font family name is determined is not spelled out by the CSS spec, but it's kind of implied that it should be the weight-stretch-style name.

specs/dwritecore/DWriteCore.md
niklasb-ms: Maybe I can provide more detail in the spec.

specs/dwritecore/DWriteCore.md
niklasb-ms: String processing is sometimes used to derive the weight-stretch-style family name, but it depends on the font.

For newer fonts with a STAT (style attributes) table, there are rules in OpenType 1.8 for how to construct family names and face names. A STAT table may be present in either a static or variable font, but is required only for variable fonts.

For static fonts without a STAT table, name ID 1 is the family name (what I've referred to here as the RBIZ family name), name ID 16 is the preferred/typographic family name, and name ID 22 is the WWS family name. If neither name 16 or 17 is present then all three family names are the same. If there's a name ID 16 but no 22 then a font may set a flag declaring itself to be a "WWS-only" family, in which case name ID 16 is used for both the typographic family name and the WWS family name.

Otherwise, if there's a name ID 16 but no 22 and no WWS-only flag, then DWrite uses a complicated heuristic to derived the WWS family name. This heuristic was first used by Windows Presentation Foundation (WPF) and later adopted by DWrite, and is described in a white paper called "WPF Font Selection Model" by Mikhail Leonov and David Brown. That white paper also describes the algorithm for ranking matching fonts based on weight, stretch, and style.

specs/dwritecore/DWriteCore.md
niklasb-ms: Sounds reasonable. (DWrite doesn't support small cap simulation.)

PR #1707, opened 2021-11-02
Update WIL and CppWinRT versions

DrusTheAxe:
WIL and CppWinRT were _inconsistently_ updated; bring ALL of the project forward, not just WindowsAppSDK_DLL.

Also removed some unnecessary casts


jefgen (approved):
:shipit:

PR #1737, opened 2021-11-05
Fix x86 instancing appId generation.

aeloros:
Currently appIds used for association data and instancing delineation are not consistent across instances of execution for x86.  The std::hasher returns a type of size_t which is then expanded in StringCchPrintf.  The data is 32bit on x86 platforms which leads to the printf code pulling in garbage for the first half of the number.  This causes the number being different every time.

This fix keeps the appId based off of 64bit numbers, but expands the 32bit version to be 64bit and fills in the upper half with another copy of the 32bit value.

This will likely be a temporary fix though, as during investigating the root cause it was discovered that the values are not consistent between x86 and x64.  This is much more of an edge case (x86 and x64 variants of the same app, on the same machine needing to work with the same data...)  The solution for everything here would be to switch to a hasher that provides the same results for all architectures.  The most immediate need though is to get x86 functioning, and then follow-up later with a more invasive change to the algorithm used.

### DIFF
--- a/dev/AppLifecycle/Association.cpp
+++ b/dev/AppLifecycle/Association.cpp
@@ -67,9 +67,16 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
 
         std::hash<std::wstring> hasher;
         auto hash = hasher(seed);
-;
+        uint64_t hash64 = static_cast<uint64_t>(hash);
+
+        // Simulate a larger hash on 32bit platforms to keep the id length consistent.
+        if constexpr (sizeof(size_t) < sizeof(uint64_t))
+        {
+            hash64 = (static_cast<uint64_t>(hash) << 32) | static_cast<uint64_t>(hash);
+        }
+
         wchar_t hashString[17]{}; // 16 + 1 characters for 64bit value represented as a string with a null terminator.
-        THROW_IF_FAILED(StringCchPrintf(hashString, _countof(hashString), L"%I64x", hash));
+        THROW_IF_FAILED(StringCchPrintf(hashString, _countof(hashString), L"%I64x", hash64));
 
         std::wstring result{ c_progIdPrefix };
         result += hashString;


zadesai (approved):

dev/AppLifecycle/Association.cpp
```diff
@@ -67,9 +67,16 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
 
         std::hash<std::wstring> hasher;
         auto hash = hasher(seed);
-;
+        uint64_t hash64 = static_cast<uint64_t>(hash);
```
DrusTheAxe, line 70: NIT: `uint64_t hash64{...};'  --> [Prefer {}=initializer syntax](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#es23-prefer-the--initializer-syntax)

NIT: `auto hash64{...};` --> auto instead of redundantly specifying type

NIT: `uint64_t hash64{ hash };` --> even better, drop the cast and let the compiler do it via the initializer syntax

dev/AppLifecycle/Association.cpp
```diff
@@ -67,9 +67,16 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
...
+
+        // Simulate a larger hash on 32bit platforms to keep the id length consistent.
+        if constexpr (sizeof(size_t) < sizeof(uint64_t))
```
DrusTheAxe, line 73: if constexpr is a thing? Haven't seen that before. Neat!

dev/AppLifecycle/Association.cpp
```diff
@@ -67,9 +67,16 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
...
+        {
+            hash64 = (static_cast<uint64_t>(hash) << 32) | static_cast<uint64_t>(hash);
+        }
+
         wchar_t hashString[17]{}; // 16 + 1 characters for 64bit value represented as a string with a null terminator.
-        THROW_IF_FAILED(StringCchPrintf(hashString, _countof(hashString), L"%I64x", hash));
+        THROW_IF_FAILED(StringCchPrintf(hashString, _countof(hashString), L"%I64x", hash64));
```
DrusTheAxe, line 79: NIT: `StringCchPrintfW` or `wsprintf_s`

DrusTheAxe (approved):
Some minor nits

aeloros:
I filed this to track the cross-architecture bug.

https://github.com/microsoft/WindowsAppSDK/issues/1738

kinkycat91:
#1742

kinkycat91:
> #
Duplicate of #

PR #1739, opened 2021-11-05
fix Hot Reload regression, incompatible with PublishReadyToRun

Scottj1s:


DefaultRyan (approved):

DefaultRyan:
The MSBuild code looks good, just need to find out if the fix is correct for @LyalinDotCom.

PR #1742, opened 2021-11-06
Update to use new Main package Name MicrosoftCorporationII.WinAppRuntime.Main

DrusTheAxe:


BenJKuhn (approved):

riverar:
Can we get some context here? How do I explain _this_ one to users? 🍿

BenJKuhn:
You won't have to explain anything. This is not a name that shows in UI. The package will still show as published by Microsoft in settings and install dialogs. If you poke around GitHub, you'll see there are a few open source packages produced by Microsoft that use this publisher ID.

riverar:
@BenJKuhn Fair enough, will check it out. I ask because Microsoft publishes stuff in the store with this identity and it _does_ show up as Microsoft Corporation II and is very unfortunate.

PR #1751, opened 2021-11-09
IsLifetimeManagerViaEnumeration() always returned true

DrusTheAxe:
https://task.ms/36938053

Rewrite IsWindows10_*OrGreater() to workaround GetVersionEx() lying to the caller (us) if the process' exe wasn't 'built for the OS' (PE Header indicating it's built for 20H1+ or (somehow) same via a SxS manifest). This caused IsLifetimeManagerViaEnumeration() to always evaluate to using the Enumeration algorithm even on newer platforms where we wanted the AppExtension algorithm.

Also changed the testing environment variable to MICROSOFT_WINDOWSAPPRUNTIME_DDLM_ALGORITHM=0|1 (Enumeration|AppExtension).

Changes tested via test collateral as well as walking through the relevant code in the debugger bit fiddling results to ensure ultimately 100 branch coverage.

### DIFF
--- a/dev/Common/IsWindowsVersion.h
+++ b/dev/Common/IsWindowsVersion.h
@@ -12,7 +12,7 @@ inline bool IsExportPresent(
     PCWSTR filename,
     PCSTR functionName)
 {
-    wil::unique_hmodule dll{ LoadLibraryExW(filename, nullptr, LOAD_WITH_ALTERED_SEARCH_PATH) };
+    wil::unique_hmodule dll{ LoadLibraryExW(filename, nullptr, 0) };
     if (dll)
     {
         auto function{ GetProcAddress(dll.get(), functionName) };
--- a/dev/Common/IsWindowsVersion.h
+++ b/dev/Common/IsWindowsVersion.h
@@ -8,42 +8,32 @@
 
 namespace WindowsVersion
 {
-VERSIONHELPERAPI IsWindowsVersionOrGreaterEx(
-    const WORD majorVersion,
-    const WORD minorVersion,
-    const WORD servicePackMajor,
-    const WORD buildNumber)
+inline bool IsExportPresent(
+    PCWSTR filename,
+    PCSTR functionName)
 {
-    OSVERSIONINFOEXW osvi{ sizeof(osvi) };
-    osvi.dwMajorVersion = majorVersion;
-    osvi.dwMinorVersion = minorVersion;
-    osvi.wServicePackMajor = servicePackMajor;
-    osvi.dwBuildNumber = buildNumber;
-
-    const DWORDLONG c_conditionMask{
-        VerSetConditionMask(
-            VerSetConditionMask(
-                VerSetConditionMask(
-                    VerSetConditionMask(
-                        0, VER_MAJORVERSION, VER_GREATER_EQUAL),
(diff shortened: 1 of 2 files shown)


BenJKuhn (approved):

riverar:
> or (somehow) same via a SxS manifest)

FYI: Is done via fusion manifest, `<supportedOS Id="{8e0f7a12-bfb3-4fe8-b9a5-48fd50a15a9a}"/>`. More info https://docs.microsoft.com/en-us/windows/win32/sysinfo/targeting-your-application-at-windows-8-1

Alternatives if interested that don't need GetProcAddress, manifests, etc.: RtlGetVersion, PEB members OS[Major|Minor]Version

dev/Common/IsWindowsVersion.h
```diff
@@ -8,42 +8,32 @@
...
-                VerSetConditionMask(
-                    VerSetConditionMask(
-                        0, VER_MAJORVERSION, VER_GREATER_EQUAL),
-                    VER_MINORVERSION, VER_GREATER_EQUAL),
-                VER_SERVICEPACKMAJOR, VER_GREATER_EQUAL),
-            VER_BUILDNUMBER, VER_GREATER_EQUAL)
-    };
-
-    return VerifyVersionInfoW(&osvi, VER_MAJORVERSION | VER_MINORVERSION | VER_SERVICEPACKMAJOR | VER_BUILDNUMBER, c_conditionMask) != FALSE;
+    wil::unique_hmodule dll{ LoadLibraryExW(filename, nullptr, LOAD_WITH_ALTERED_SEARCH_PATH) };
```
huichen123, line 15: >LOAD_WITH_ALTERED_SEARCH_PATH [](http://example.com/codeflow?start=63&length=29)

according to MSDN: If this value is used and lpFileName specifies a relative path, the behavior is undefined.

DrusTheAxe: Good point. And it's unnecessary; flags were used in an early iteration and I forgot to scrub it. Fixed

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -292,11 +292,24 @@ bool IsLifetimeManagerViaEnumeration()
...
-    if (GetEnvironmentVariableW(L"MICROSOFT_WINDOWSAPPRUNTIME_DDLM_ENUMERATION", nullptr, 0) > 0)
+    // Select the LifetimeManager implementation if the environment variable
+    // MICROSOFT_WINDOWSAPPRUNTIME_DDLM_ALGORITHM is defined (for testing scenarios)
+    // where:
+    //     envvar=0 => Enumeration
+    //     envvar=1 => AppExtension
     {
-        return true;
+        WCHAR value[1 + 1]{};
+        if (GetEnvironmentVariableW(L"MICROSOFT_WINDOWSAPPRUNTIME_DDLM_ALGORITHM", value, ARRAYSIZE(value)) > 0)
```
huichen123, line 302: >GetEnvironmentVariableW [](http://example.com/codeflow?start=12&length=23)

If lpBuffer is not large enough to hold the data, the return value is the buffer size, in characters, required to hold the string and its terminating null character and the contents of lpBuffer are undefined.

PR #1783, opened 2021-11-12
Added MSIX license support

DrusTheAxe:
Added support to install licenses if/when MSIX packages are Store signed and not installed by the Store*

We have a stub 'License API' checked in here that. When the build pipeline makes a Store-signed package it adds the necessary goo to get the license installed.

This PR updates the installer. A follow up PR will do likewise for other cases (e.g. Deployment API)

* Store installs licenses when installing Store-signed packages. When others install Store-signed packages (e.g. WindowsAppRuntimeInstaller.exe) we need to install the license or we get errors at runtime This prevents those errors from happening.


BenJKuhn (approved):

installer/dev/install.cpp
```diff
@@ -224,22 +238,87 @@ namespace WindowsAppRuntimeInstaller {
...
+    void ShowErrorMessage(const HRESULT hr)
+    {
+        if (SUCCEEDED(hr))
+        {
+            std::wcout << std::endl;
+            return;
+        }
+
+        PWSTR message{};
+        if (FormatMessage(FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM,
```
jonwis, line 313: Seems like a good WIL helper to add, maybe file an issue to track?

installer/dev/install.h
```diff
@@ -13,5 +26,11 @@ namespace WindowsAppRuntimeInstaller {
         bool isFramework{};
     };
 
-    HRESULT DeployPackages(const bool quiet) noexcept;
+    HRESULT Deploy(const WindowsAppRuntimeInstaller::Options options) noexcept;
+
+    HRESULT DeployPackages(const WindowsAppRuntimeInstaller::Options options);
```
jonwis, line 31: Code seems like an odd mix of exceptional and HRESULTish error patterns.

installer/dev/install.cpp
DrusTheAxe: Was thinking same as I was cribbing it (actually surprised it's not already in WIL).

Ask tracked via https://task.ms/37017993

installer/dev/install.h
DrusTheAxe: Pre-existing condition.

There's some odd mixes as you say, but we'll bake those out as we do more work in this space next year. I made a few minor improvements as needed to make (and test!) this PR but didn't want to range too far afield into other code paths as we're prepping for release.

We'll get there, just not today.

PR #1793, opened 2021-11-15
Add Licensing API

DrusTheAxe:
Add an API to install our licenses.

Those who directly install the WindowsAppSDK packages (i.e. don't run WindowsAppRuntimeInstaller.exe and don't use the Deployment API) need to install the *.msix packages AND their licenses.

Easiest way to do that is to

```
#include <WindowsAppRuntime-Licensing.h>
...
const uint32_t c_majorMinorVersion{ WINDOWSAPPSDK_RELEASE_MAJORMINOR };
PCWSTR c_versionTag{ WINDOWSAPPSDK_RELEASE_VERSION_TAG_W };
const auto hr{ WindowsAppRuntime_InstallLicenses(c_majorMinorVersion, c_versionTag) };
if (FAILED(hr))
{
    ...error...
}
```

NOTE: There's also `WindowsAppRuntime_InstallLicensesInPackage()` but that's primarily for testing.

See `WindowsAppRuntime-Licensing.h` for more details.

### DIFF
--- a/dev/Licensing/MsixLicensing.cpp
+++ b/dev/Licensing/MsixLicensing.cpp
@@ -6,11 +6,92 @@
 
 #include <WindowsAppRuntime-License.h>
 
-STDAPI MsixInstallLicenses() noexcept try
+STDAPI MsixInstallLicenses(
+    UINT32 majorMinorVersion,
+    PCWSTR versionTag) noexcept try
 {
-    //TODO
-    // FOREACH file IN pkgdir\MSIX\*_license.xml
-    //  InstallLicenseFile(file)
-    return S_OK;
+    // Licenses are only needed for Stable releases (versionTag=null or "")
+    if (versionTag && (versionTag[0] != L'\0'))
+    {
+        return S_OK;
+    }
+
+    const UINT16 majorVersion{ HIWORD(majorMinorVersion) };
+    const UINT16 minorVersion{ LOWORD(majorMinorVersion) };
+    WCHAR packageFamilyName[PACKAGE_FAMILY_NAME_MAX_LENGTH + 1]{};
+    wsprintf(packageFamilyName, L"Microsoft.WindowsAppRuntime.%hu.%hu_8wekyb3d8bbwe", majorVersion, minorVersion);
+
+    // Find a framework package in the family (any will do)
+    winrt::Windows::Management::Deployment::PackageManager packageManager;
+    winrt::hstring currentUser;
+    const auto c_packageTypes{ winrt::Windows::Management::Deployment::PackageTypes::Framework };
+    for (auto package : packageManager.FindPackagesForUserWithPackageTypes(currentUser, packageFamilyName, c_packageTypes))
+    {
+        // Install the license(s)
+        const auto packageId{ package.Id() };
+        const auto packageFullName{ packageId.FullName() };
+        RETURN_IF_FAILED(MsixInstallLicensesInPackage(packageFullName.c_str()));
+        return S_OK;
+    }
+
+    // No framework package found in the family
+    RETURN_HR_MSG(HRESULT_FROM_WIN32(ERROR_NOT_FOUND), "PackageFamilyName:%ls", packageFamilyName);
 }
 CATCH_RETURN();
+
+STDAPI MsixInstallLicensesInPackage(PCWSTR frameworkPackageFullName) noexcept try
+{
+    // Get the package's path
+    //
+    // NOTE: Use GetStagedPackagePathByFullName as don't care if the package is registered.
+    //
+    // NOTE: Don't use GetStagedPackagePathByFullName2 as we want the *install* location
+    //       (not e.g. modifiable location). Plus Get...2() is new in 20H1 (too new for us)
+    //       where Get...() exists down to Win81. In this case simplest is bestest.
+    wil::unique_cotaskmem_string packagePathBufferDynamic;
+    uint32_t packagePathLength{};
+    const auto rc{ GetStagedPackagePathByFullName(frameworkPackageFullName, &packagePathLength, nullptr) };
+    if (rc != ERROR_INSUFFICIENT_BUFFER)
+    {
+        RETURN_HR_MSG(HRESULT_FROM_WIN32(rc), "PackageFullName:%ls", frameworkPackageFullName);
+    }
+    auto packagePath{ wil::make_cotaskmem_string(nullptr, packagePathLength) };
+    RETURN_IF_WIN32_ERROR(GetStagedPackagePathByFullName(frameworkPackageFullName, &packagePathLength, packagePath.get()));
+
+    // Build path for licenses
+    std::filesystem::path licensePath{ packagePath.get() };
+    const auto c_packageRelativeLicensePath{ L"MSIX" };
+    licensePath /= c_packageRelativeLicensePath;
+    auto licenseFilespec{ licensePath };
+    licenseFilespec /= L"*_license.xml";
+
+    // Deploy the licenses (if any)
+    ::Microsoft::Windows::ApplicationModel::Licensing::Installer licenseInstaller;
+    WIN32_FIND_DATA findFileData{};
+    wil::unique_hfind hfind{ FindFirstFileW(licenseFilespec.c_str(), &findFileData) };
+    if (!hfind)
+    {
+        const auto lastError{ GetLastError() };
+        RETURN_HR_IF_MSG(HRESULT_FROM_WIN32(lastError), (lastError != ERROR_FILE_NOT_FOUND) && (lastError != ERROR_PATH_NOT_FOUND),
+                         "FindFirstFile:%ls", licenseFilespec.c_str());
+        return S_OK;
+    }
+    for (;;)
+    {
+        // Install the license file
+        auto licenseFilename{ licensePath };
+        licenseFilename /= findFileData.cFileName;
+        RETURN_IF_FAILED_MSG(licenseInstaller.InstallLicenseFile(licenseFilename.c_str()),
+                             "LicenseFile:%ls", licenseFilename.c_str());
+
+        // Next! (if any)
+        if (!FindNextFileW(hfind.get(), &findFileData))
+        {
+            const auto lastError{ GetLastError() };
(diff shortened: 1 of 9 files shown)


BenJKuhn (approved):

dev/Licensing/MsixLicensing.cpp
```diff
@@ -6,11 +6,92 @@
...
+        // Install the license file
+        auto licenseFilename{ licensePath };
+        licenseFilename /= findFileData.cFileName;
+        RETURN_IF_FAILED_MSG(licenseInstaller.InstallLicenseFile(licenseFilename.c_str()),
+                             "LicenseFile:%ls", licenseFilename.c_str());
+
+        // Next! (if any)
+        if (!FindNextFileW(hfind.get(), &findFileData))
+        {
+            const auto lastError{ GetLastError() };
```
jonwis, line 90:
```c++
RETURN_LAST_ERROR_IF(GetLastError() != ERROR_NO_MORE_FILES);
```

dev/Licensing/MsixLicensing.cpp
```diff
@@ -6,11 +6,92 @@
...
+    //       where Get...() exists down to Win81. In this case simplest is bestest.
+    wil::unique_cotaskmem_string packagePathBufferDynamic;
+    uint32_t packagePathLength{};
+    const auto rc{ GetStagedPackagePathByFullName(frameworkPackageFullName, &packagePathLength, nullptr) };
+    if (rc != ERROR_INSUFFICIENT_BUFFER)
+    {
+        RETURN_HR_MSG(HRESULT_FROM_WIN32(rc), "PackageFullName:%ls", frameworkPackageFullName);
+    }
+    auto packagePath{ wil::make_cotaskmem_string(nullptr, packagePathLength) };
+    RETURN_IF_WIN32_ERROR(GetStagedPackagePathByFullName(frameworkPackageFullName, &packagePathLength, packagePath.get()));
```
jonwis, line 59: Consider lifting this block into a helper function, then eventually into WIL, like `wil::cotaskmem_string GetStagedPackagedPath(PCWSTR packageFullName)`

jonwis: ... which you already have below, as `GetPathForPackage` ? Reuse it here, put a definition into a header used by this directory.

dev/WindowsAppRuntime_BootstrapDLL/WindowsAppRuntime-Licensing.cpp
```diff
@@ -6,9 +6,125 @@
...
+        THROW_WIN32_MSG(lastError, "Error in LoadLibrary: %d (0x%X) loading %ls", lastError, lastError, windowsAppRuntimeDllFilename.c_str());
+    }
+    return windowsAppRuntimeDll;
+}
+
+static HMODULE LoadMsixLicensingApisInPackage(
+    PCWSTR packageFullName)
+{
+    auto path{ GetPathForPackage(packageFullName) };
+    return LoadMsixLicensingApisAtPath(path);
```
jonwis, line 50: Consider merging all these little temporaries together, like:

```c++
return LoadMsixLicensingApisAtPath(GetPathForPackage(packageFullName));
```

dev/WindowsAppRuntime_BootstrapDLL/WindowsAppRuntime-Licensing.cpp
```diff
@@ -6,9 +6,125 @@
...
+    {
+        THROW_HR_MSG(HRESULT_FROM_WIN32(rc), "PackageFullName:%ls", packageFullName);
+    }
+    auto packagePath{ wil::make_cotaskmem_string(nullptr, packagePathLength) };
+    THROW_IF_WIN32_ERROR(GetStagedPackagePathByFullName(packageFullName, &packagePathLength, packagePath.get()));
+    std::filesystem::path path{ packagePath.get() };
+    return path;
+}
+
+static HMODULE LoadMsixLicensingApisAtPath(
```
jonwis, line 32: This method should return `wil::unique_hmodule` so the callers don't have to wonder about lifetime.

dev/Licensing/MsixLicensing.cpp
DrusTheAxe: There's a bunch of refactoring and optimizing that'll happen soon, just not wise to do today. We'll pull this work into main/1.1 (and work out some merge issues) and _then_ refactor a bit, when time isn't so crunched.

dev/Licensing/MsixLicensing.cpp
DrusTheAxe: Forgot about RETURN_LAST_ERROR_IF. Thanks for the reminder. Will update after we do the merge to main/1.1

DrusTheAxe: +@sachintaMSFT as reminder/visibility

dev/WindowsAppRuntime_BootstrapDLL/WindowsAppRuntime-Licensing.cpp
DrusTheAxe: Broken up this way is useful (easier) for debugging and same overhead in the end when the optimizer gets through with it (I would hope).

dev/WindowsAppRuntime_BootstrapDLL/WindowsAppRuntime-Licensing.cpp
DrusTheAxe: I thought about that but thought there was general preference to return e.g. HMODULE and let callers decide how to handle it. Or is that more relevant for other codebases and here we can/should e.g. go pervasively wil::unique_hmodule and not HMODULE?

MarkIngramUK:
I'm looking for a bit more info on this new function - we manually call `MddBootstrapInitialize`, so should that now be prefaced with `WindowsAppRuntime_InstallLicenses`? I've tested the 1.0 release, and `WindowsAppRuntime_InstallLicenses` doesn't need to be called for the app to start, so it doesn't appear to be required at this point in time? I'm not clear what this function is for, or when it's needed; can you offer any further info?

DrusTheAxe:
`WindowsAppRuntime_InstallLicenses()` exists for those who install the MSIX packages directly. If so, `WindowsAppRuntime_InstallLicenses()` should be called when you install the *.msix. Installing *.msix and associated licenses are a one-time hit. Not something you need to do every time your app runs.

There's 4 scenarios how WinAppSDK can be installed on a system:

1. Packaged app distributed via the MS Store. App calls the Deployment API.
2. Packaged app not distributed via the MS Store (i.e. sideloaded). App calls the Deployment API.
3. Run `WindowsAppRuntimeInstaller.exe`.
4. Install the individual MSIX packages. Installer needs to call WindowsAppRuntime_InstallLicenses()

1+2 amount to the same thing. Regardless how your packaged app gets on the box, call the Deployment API. The Deployment API's implementation installs the Main and Singleton packages if necessary AND installs their associated licenses.

3 is equally simple. The installer installs WinAppSDK's MSIX packages if necessary AND installs their associated licenses.

4 is where your app's unpackaged AND you don't use the installer. You can download the MSIX packages and install them (`packageManager.AddPackageAsync()` etc), but for WinAppSDK to work the associated licenses need to be installed too. Thus you'd also need to call `WindowsAppRuntime_InstallLicenses()`.


>doesn't need to be called for the app to start

Depends on the app. For example, the bootstrapper and DynamicDependencies use Windows functionality that requires a valid license (if the packages are Store-signed). Some behavior doesn't need the licenses but devil's very much in the details.

NOTE: These licenses are for WinAppSDK's packages (not your app).

It all depends how MSIX packages are signed. Store-signed packages need these licenses to work correctly. For WinAppSDK's stable channel releases we Store-sign the framework, main and singleton packages. WinAppSDK's experimental and preview channels' MSIX packages are Authenticode-signed. Not Store-signed so no (Store-)licenses needed. We didn't discover the issue until recently as 0.8 didn't happen to use Windows functionality needing licenses (framework packages don't need licenses) and we only Store-sign packages for stable releases.

P.S. If you're a packaged app when the Store installs your package it actually does _2_ things: use the [PackageManager API](https://docs.microsoft.com/uwp/api/Windows.Management.Deployment.PackageManager?view=winrt-22000) to install your package and (you guessed it) _install a license for you package_.

(OK it may do a bit more than those 2 things. But for purposes of this discussion it's those 2 that matter 😀)


>If I do call the function, it fails with error result 0x800700e8.

Can you elaborate? Did you have WinAppSDK's packages installed? How'd you install them?

DrusTheAxe:
+@zaryaf for visibility

riverar:
And it's my understanding you deliver Microsoft Store signed packages to ensure they can be serviced. Had you signed them differently--to avoid the licensing goop--they'd be unservicable (or require an alternate servicing pipeline).

Is that right-ish?

DrusTheAxe:
> And it's my understanding you deliver Microsoft Store signed packages to ensure they can be serviced. Had you signed them differently--to avoid the licensing goop--they'd be unservicable (or require an alternate servicing pipeline).
> 
> Is that right-ish?

Ish &lt;g&gt;

The licensing goop is an implementation detail necessary for Store-signed packages to work as expected in all scenarios (not just when WinAppSDK is Store-distributed).

The driving factor why we prefer to be Store-signed is Store-signed packages install in some cases where non-Store-signed packages don't e.g. there's a policy administrators can disable to prevent non-Store-signed package installation, and some admins do enable it. Being Store-signed reduces unnecessary friction (for users, developers and us - win-win-win)

We considered the `appLicensing` capability but that's an incomplete solution for our needs. It lets your Store-signed packages work as expected without a license but it's a restricted capability and (the kicker) only supported on &gt;=Win11. Windows App SDK runs down to RS5. That said, we're considering adding this in 1.1 as an optimization. That would allow us to change our license handling to skip the license work when run on &gt;=Win11. Not terribly expensive work but even if small, why pay the cost if we don't need to? 😀

MarkIngramUK:
Hi, thanks for all the info!
We only bootstrap the framework for developer builds within Visual Studio. We use CMake, so don't have any "special" Visual Studio projects which manage this stuff for us (just normal C++ projects generated via CMake), where the dev builds will bootstrap the framework, but for retail builds it won't, as the build server will create an MSIX package after compilation.

PR #1843, opened 2021-11-24
Take version from .editorconfig from os.2020 to get C++ rules

ChrisGuzak:
I did not try to merge with the existing rules, assuming they were not chosen carefully. If I have that wrong reject this PR.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,134 +1,206 @@
+# editorconfig: http://editorconfig.org/
+# Help developers standardize spaces, tabs, encoding, end-line characters across editors
+
+# top-most .editorconfig file
 root = true
 
-# Rules in this file were initially inferred by Visual Studio IntelliCode from the S:\Repos\dep.controls codebase based on best match to current usage at 11/15/2018
-# You can modify the rules from these initially generated values to suit your own policies
-# You can learn more about editorconfig here: https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference
-[*.cs]
-
-#Core editorconfig formatting - indentation
-
-#use soft tabs (spaces) for indentation
-indent_style = space
-indent_size = 4
-charset = utf-8-bom
+# defaults for all files
+[*]
+charset = utf-8
 end_of_line = crlf
+indent_size = 4
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
 
-#Formatting - indentation options
+# markdown, diff overrides
+# two trailing spaces are required for <br/> and hard line-breaks in markdown files
+# see: (https://daringfireball.net/projects/markdown/syntax#p) and (http://spec.commonmark.org/0.27/#hard-line-break)
+[*.{md,diff}]
+trim_trailing_whitespace = false
 
-#indent switch case contents.
+[*.{md,xml,gprops,man,natvis}]
+indent_size = 2
+
+# manifest validation tool requires BOM
+[*.man]
+charset = utf-8-bom
+
+# XML-based MSBuild and Visual Studio files
+[*.{props,targets,settings,*proj,vcxitems,filters}]
+indent_size = 2
+
+# Exceptions to the above *proj wildcard
+[*.vdproj]
+indent_size = 4
+
+# Visual Studio uses hard tabs for SLN files, so don't fight it
+[*.sln]
+indent_style = tab
+
+# YAML overrides
+[*.{yml,yaml}]
+indent_size = 2
+
+# package.json overrides
+# Updating package.json with NPM will revert indentation to 2 spaces so to
+# reduce churn, let's align to NPM and specify indent size 2
+[package.json]
+indent_size = 2
+
+# JSON formats
+[{*.json,*.testlist,testmd.definition}]
+indent_size = 2
+
+[*.{rc}]
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.{c,w}]
+trim_trailing_whitespace = true
+insert_final_newline = true
+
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -1,134 +1,206 @@
...
-#do not place System.* using directives before other using directives
-dotnet_sort_system_directives_first = false
+# JSON formats
+[{*.json,*.testlist,testmd.definition}]
+indent_size = 2
 
-#Formatting - spacing options
+[*.{rc}]
+trim_trailing_whitespace = true
+insert_final_newline = true
```
DrusTheAxe, line 57: Line 56+57 are dupes of line 13+14 in the `[*]` section. Does this act is an _override_ to `[*]` definitions or is `[*]` a peer only used when a matching extension's not otherwise found?

If these are peers a lot of extensions are missing values.

If these are overrides to `[*]` then there's a lot of redundancies in the file e.g. delete lines 35, 56+57, 60+61, 64+65 and probably a few others.

.editorconfig
```diff
@@ -1,134 +1,206 @@
...
 
-#require NO space between a cast and the value
+[*.{c,w}]
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.{c++,cc,cpp,cppm,cxx,h,h++,hh,hpp,hxx,inl,ipp,ixx,tlh,tli,idl}]
```
DrusTheAxe, line 63: Can *.ps1 and *.psm1 be handled here too?

.editorconfig
```diff
@@ -1,134 +1,206 @@
...
+# YAML overrides
+[*.{yml,yaml}]
+indent_size = 2
 
-#Formatting - organize using options
+# package.json overrides
+# Updating package.json with NPM will revert indentation to 2 spaces so to
+# reduce churn, let's align to NPM and specify indent size 2
+[package.json]
+indent_size = 2
```
DrusTheAxe, line 49: Should also have a packages.config for our broad NuGet usage.

SUGGESTION:
```
# packages.config override
[packages.config]
indent_size = 2
```

.editorconfig
```diff
@@ -1,134 +1,206 @@
...
+# markdown, diff overrides
+# two trailing spaces are required for <br/> and hard line-breaks in markdown files
+# see: (https://daringfireball.net/projects/markdown/syntax#p) and (http://spec.commonmark.org/0.27/#hard-line-break)
+[*.{md,diff}]
+trim_trailing_whitespace = false
 
-#use soft tabs (spaces) for indentation
-indent_style = space
-indent_size = 4
+[*.{md,xml,gprops,man,natvis}]
```
DrusTheAxe, line 22: And `xsd`

.editorconfig
```diff
@@ -1,134 +1,206 @@
...
+indent_size = 2
+
+# manifest validation tool requires BOM
```
DrusTheAxe, line 25: Are you referring to Windows manifest files? N/A for the project

.editorconfig
DrusTheAxe, line 22: .man has its own section on line 26. Is it preferable to mix or have dedication/separate definitions?

P.S. See other comment re .man

DrusTheAxe (changes requested):

.editorconfig
DrusTheAxe: Per the [docs](https://editorconfig-specification.readthedocs.io/#id6) the rules are additive/overrides:

>Files are read top to bottom and the most recent rules found take precedence. If multiple EditorConfig files have matching sections, the rules from the closer EditorConfig file are read last, so pairs in closer files take precedence.

So anything defined for [*] at the top doesn't need to be redundantly defined later

DrusTheAxe:
> I did not try to merge with the existing rules, assuming they were not chosen carefully. If I have that wrong reject this PR.

This looks like a notable improvement. Thanks for offering this Chris

.editorconfig
ChrisGuzak: I want to keep this in sync with the os.2020 version.
You are correct that this is not needed but being explicit and not depending on the other declarations argues for keeping them. Let me know if you strongly disagree.

.editorconfig
ChrisGuzak: yes, but in this change, I want to copy os.2020. if you want that please start with a PR against that.

.editorconfig
ChrisGuzak: not in this change.

.editorconfig
ChrisGuzak: yes, windows packaging system manifests.

.editorconfig
DrusTheAxe: Are you thinking we should keep this and os.2020 copies in sync? Changes here port to there and vice versa?

>being explicit and not depending on the other declarations argues for keeping them

Problem is it's _half_ explicit. If you really think explicit is good then lines 9-14 should be in every section. If the additive/override model better then we should embrace that. The half-and-half is what I question. This is particularly important if you think WinAppSDK and os.2020 should be kept in sync, as that precludes WinAppSDK from fixing the half-and-half approach.

I prefer the override approach as the more manageable and maintainable solution. Less is more.

Thoughts?

ChrisGuzak: good point. I want to get this in for now to avoid wasting time in PRs. 

I filed http://task.ms/37331332 to go all in on override.

DrusTheAxe (approved):
LGTM

raghavthind2005 (approved):

DrusTheAxe:
There's been some drift (enhancements) in the OS.2020 .editorconfig since when this PR was created, plus the PR is problematic due to drif t(it targets main not develop, I can't git checkout the branch for some weird reason...).

I made a [new PR with the latest from OS.2020](https://github.com/microsoft/WindowsAppSDK/pull/3363) to address this. Closing this PR and following up thru that PR.

PR #1875, opened 2021-12-06
Introduce ToastNotification Registration APIs

sharath2727:
Introduce the ToastNotifications.idl with stub implementations:
- Implements ToastNotificationManager functions Register/UnregisterActivator.
- Implement ToastActivationCallback to allow packaged/unpackaged apps be com activatable when receiving a ToastActivation
- Add unit test app ToastNotificationsTestApp and TAEF tests that invoke Register/Unregister functions.
- Add ToastNotificationUtility with helper function to register assets in registry and add ComServer registration for unpackaged apps


ToastNotificationTests/ToastNotificationTests.vcxproj
```diff
@@ -0,0 +1,167 @@
+<?xml version="1.0" encoding="utf-8"?>
```
danielayala94, line 1: ToastNotificationTests is in the root. Please move to the /test folder.

dev/ToastNotifications/ToastNotificationUtility.cpp
```diff
@@ -0,0 +1,192 @@
...
+        nullptr,
+        REG_OPTION_NON_VOLATILE,
+        KEY_ALL_ACCESS,
+        nullptr,
+        &hKey,
+        nullptr));
+
+    std::wstring comRegistrationExeString{ L"\"" };
+    comRegistrationExeString.append(processName.get());
+    comRegistrationExeString.append(L"\" ");
```
danielayala94, line 126: Is this backslash really needed?

dev/ToastNotifications/ToastNotificationUtility.cpp
```diff
@@ -0,0 +1,192 @@
...
+    comRegistrationExeString.append(L"----ToastActivated:");
+
+    RegisterValue(hKey, nullptr, reinterpret_cast<const BYTE*>(comRegistrationExeString.c_str()), REG_SZ, (comRegistrationExeString.size() * sizeof(wchar_t)));
```
danielayala94, line 129: Have we tested this using wide-string characters (not ASCII)?

dev/ToastNotifications/ToastNotificationUtility.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+void UnRegisterComServer(std::wstring const& clsid)
+{
+    wil::unique_hkey hKey;
+    std::wstring subKey{ clsIdPath + clsid + L"\\LocalServer32" };
+
+    THROW_IF_FAILED(RegDeleteKeyEx(
+        HKEY_CURRENT_USER,
+        subKey.c_str(),
+        KEY_WOW64_64KEY,
```
danielayala94, line 140: What does this mean?

dev/ToastNotifications/ToastNotificationUtility.h
```diff
@@ -0,0 +1,48 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#pragma once
+#include "pch.h"
+#include <winrt/Windows.Foundation.h>
+#include <ToastActivatedEventArgs.h>
+
+const std::wstring appIdentifierPath = L"Software\\Classes\\AppUserModelId\\";
```
danielayala94, line 9: Nit, var naming:

```suggestion
const std::wstring c_appIdentifierPath = L"Software\\Classes\\AppUserModelId\\";
```

Same for the var below.

dev/ToastNotifications/ToastNotificationUtility.h
```diff
@@ -0,0 +1,48 @@
...
+
+
+winrt::event<winrt::Windows::Foundation::EventHandler<winrt::Microsoft::Windows::ToastNotifications::ToastActivatedEventArgs>>& GetToastHandlers();
+int& GetToastHandleCount();
+
+inline const winrt::hstring COM_ACTIVATOR_KEY = L"ComActivatorKey";
+
+inline const int GUID_LENGTH = 36;
+
+inline std::wstring ConvertPathToKey(std::wstring path)
```
danielayala94, line 20: You can do `std::wstring&`. Make sure to avoid this unnecessary copy.

dev/ToastNotifications/ToastNotificationManager.cpp
```diff
@@ -0,0 +1,142 @@
...
+                REGCLS_MULTIPLEUSE,
+                &s_toastcomActivatorRegistration));
+        }
+
+        GetWaitHandleForArgs().create();
+    }
+
+    void ToastNotificationManager::UnregisterActivator()
+    {
+        THROW_HR_IF_MSG(HRESULT_FROM_WIN32(ERROR_NOT_FOUND), !s_toastcomActivatorRegistration, "ComActivator not registered.");
```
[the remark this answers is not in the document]
danielayala94: Also, is there a THROW_WIN32_ERROR or something like that?

dev/ToastNotifications/ToastNotificationUtility.cpp
```diff
@@ -0,0 +1,192 @@
...
+        &hKey,
+        nullptr));
+
+    RegisterValue(hKey, L"DisplayName", reinterpret_cast<const BYTE*>(displayName.c_str()), REG_EXPAND_SZ, displayName.size() * sizeof(wchar_t));
+    RegisterValue(hKey, L"IconUri", reinterpret_cast<const BYTE*>(iconUri.AbsoluteUri().c_str()), REG_EXPAND_SZ, iconUri.AbsoluteUri().size() * sizeof(wchar_t));
+
+    std::wstring wideStringClsid{ clsid.get() };
+    RegisterValue(hKey, L"CustomActivator", reinterpret_cast<const BYTE*>(wideStringClsid.c_str()), REG_SZ, wideStringClsid.size() * sizeof(wchar_t));
+
+    return;
```
danielayala94, line 105: Please remove.

dev/ToastNotifications/ToastNotificationUtility.h
```diff
@@ -0,0 +1,48 @@
...
+#include "pch.h"
+#include <winrt/Windows.Foundation.h>
+#include <ToastActivatedEventArgs.h>
+
+const std::wstring appIdentifierPath = L"Software\\Classes\\AppUserModelId\\";
+const std::wstring clsIdPath = L"Software\\Classes\\CLSID\\";
+
+
+winrt::event<winrt::Windows::Foundation::EventHandler<winrt::Microsoft::Windows::ToastNotifications::ToastActivatedEventArgs>>& GetToastHandlers();
+int& GetToastHandleCount();
```
danielayala94, line 14: Why not just return an int?

```suggestion
int GetToastHandleCount();
```

dev/ToastNotifications/ToastActivationCallback.cpp
```diff
@@ -0,0 +1,45 @@
...
+    auto appProperties = winrt::CoreApplication::Properties();
+    winrt::ToastActivatedEventArgs activatedEventArgs = winrt::make<winrt::Microsoft::Windows::ToastNotifications::implementation::ToastActivatedEventArgs>(invokedArgs, userInput);
+
+    if (GetToastHandleCount())
+    {
+        GetToastHandlers()(*this, activatedEventArgs);
+    }
+    else
+    {
+        appProperties.Insert(ACTIVATED_EVENT_ARGS_KEY, activatedEventArgs);
```
danielayala94, line 40: Wait, is this key shared with push event args?

dev/ToastNotifications/ToastNotificationUtility.h
```diff
@@ -0,0 +1,48 @@
...
+
+inline const winrt::hstring COM_ACTIVATOR_KEY = L"ComActivatorKey";
```
danielayala94, line 16: Where is this being used?

danielayala94 (changes requested):

dev/ToastNotifications/ToastNotificationUtility.cpp
```diff
@@ -0,0 +1,192 @@
...
+    using namespace Windows::ApplicationModel::Core;
+}
+
+std::wstring RetrieveUnpackagedToastGuid()
+{
+    wil::unique_cotaskmem_string processName;
+    THROW_IF_FAILED(GetCurrentProcessPath(processName));
+
+    std::wstring wideStringProcessName{ processName.get() };
+    std::wstring subKey{ L"Software\\Classes\\AppUserModelId\\" + ConvertPathToKey(wideStringProcessName) };
```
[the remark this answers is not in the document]
danielayala94: Seems like ConvertPathToKey takes care of slashes?

dev/ToastNotifications/ToastActivationCallback.cpp
```diff
@@ -0,0 +1,45 @@
+﻿#include "pch.h"
+#include <ToastActivationCallback.h>
+#include "ToastActivatedEventArgs.h"
+#include "externs.h"
```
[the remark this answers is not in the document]
pmpurifoy: We need externs here to get access to the global event GetWaitHandleForArgs().

dev/ToastNotifications/ToastNotificationUtility.cpp
```diff
@@ -0,0 +1,192 @@
...
+        subKey.c_str(),
+        0,
+        nullptr,
+        REG_OPTION_NON_VOLATILE,
+        KEY_ALL_ACCESS,
+        nullptr,
+        &hKey,
+        nullptr));
+
+    WCHAR registeredGuidBuffer[GUID_LENGTH + 4]; // GUID length + '{' + '}' + '\0'
```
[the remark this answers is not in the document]
pmpurifoy: Oops. Should be 3

dev/ToastNotifications/ToastActivationCallback.h
```diff
@@ -0,0 +1,38 @@
...
+struct ToastActivationCallback : winrt::implements<ToastActivationCallback, INotificationActivationCallback>
+{
+    HRESULT __stdcall Activate(
+        LPCWSTR appUserModelId,
+        LPCWSTR invokedArgs,
+        [[maybe_unused]] NOTIFICATION_USER_INPUT_DATA const* data,
+        [[maybe_unused]] ULONG dataCount);
+};
+
+struct ToastActivationCallback_factory : winrt::implements<ToastActivationCallback_factory, IClassFactory>
```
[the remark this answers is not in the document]
sharath2727: Not really. Will keep this comment open to see if I could find some guidelines for it.

dev/ToastNotifications/ToastNotificationManager.cpp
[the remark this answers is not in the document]
sharath2727: I checked found THROW_WIN32_MSG that doesn't take condition parameter. :(

dev/ToastNotifications/ToastActivationCallback.h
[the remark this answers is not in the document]
pmpurifoy: Since PushNotificationBackgroundTaskFactory is the previous usage, let's follow it same style.
```suggestion
struct ToastActivationCallbackFactory : winrt::implements<ToastActivationCallbackFactory, IClassFactory>
```

dev/ToastNotifications/ToastActivationCallback.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+namespace winrt
+{
+    using namespace Microsoft::Windows::ToastNotifications;
+    using namespace Windows::ApplicationModel::Core;
+    using namespace Windows::Foundation;
+    using namespace Windows::Foundation::Collections;
+}
+
+HRESULT __stdcall ToastActivationCallback::Activate(
```
hulumane, line 19: This method should be wrapped in a try catch_return

dev/ToastNotifications/ToastActivationCallback.cpp
```diff
@@ -0,0 +1,45 @@
...
+{
+    winrt::IMap<winrt::hstring, winrt::hstring> userInput{ winrt::single_threaded_map<winrt::hstring, winrt::hstring>() };
+    for (unsigned long i = 0; i < dataCount; i++)
+    {
+        userInput.Insert(data[i].Key, data[i].Value);
+    }
+
+    winrt::ToastActivatedEventArgs activatedEventArgs = winrt::make<winrt::Microsoft::Windows::ToastNotifications::implementation::ToastActivatedEventArgs>(invokedArgs, userInput);
+
+    if (GetToastHandleCount() > 0)
```
hulumane, line 33: Comment needed for the if else. It's unclear what each case is for. Per my understanding, one case is specifically for the Process Launch scenario while the other case is for the scenario where Process has already launched and you have registered event handlers.

dev/ToastNotifications/ToastActivationCallback.h
```diff
@@ -0,0 +1,38 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#pragma once
+
+#include "NotificationActivationCallback.h"
+
+struct ToastActivationCallback : winrt::implements<ToastActivationCallback, INotificationActivationCallback>
+{
+    HRESULT __stdcall Activate(
```
hulumane, line 10: Good practise to tag all methods which return HRESULT with noexcept

dev/ToastNotifications/ToastActivationInfo.cpp
```diff
@@ -6,18 +6,25 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
 {
     winrt::Microsoft::Windows::ToastNotifications::ToastActivationInfo ToastActivationInfo::CreateFromActivationGuid(winrt::guid const& taskClsid)
     {
-        throw hresult_not_implemented();
+        THROW_HR_IF_MSG(E_NOTIMPL, !AppModel::Identity::IsPackagedProcess(), "Not applicable for unpackaged applications");
+
+        return winrt::make<ToastActivationInfo>(taskClsid, nullptr);
```
hulumane, line 11: REcommend having 2 specialized constructors: One that takes in a GUID and another that takes in ToastAssets

dev/ToastNotifications/ToastActivationInfo.cpp
```diff
@@ -6,18 +6,25 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
 {
     winrt::Microsoft::Windows::ToastNotifications::ToastActivationInfo ToastActivationInfo::CreateFromActivationGuid(winrt::guid const& taskClsid)
     {
-        throw hresult_not_implemented();
+        THROW_HR_IF_MSG(E_NOTIMPL, !AppModel::Identity::IsPackagedProcess(), "Not applicable for unpackaged applications");
```
hulumane, line 9: E_NOTSUPPORTED  is more accurate

dev/ToastNotifications/ToastActivationInfo.cpp
```diff
@@ -6,18 +6,25 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
...
+
+        return winrt::make<ToastActivationInfo>(taskClsid, nullptr);
     }
+
     winrt::Microsoft::Windows::ToastNotifications::ToastActivationInfo ToastActivationInfo::CreateFromToastAssets(winrt::Microsoft::Windows::ToastNotifications::ToastAssets const& assets)
     {
-        throw hresult_not_implemented();
+        THROW_HR_IF_MSG(E_NOTIMPL, AppModel::Identity::IsPackagedProcess(), "Not applicable for packaged applications");
```
hulumane, line 16: E_NOTSUPPORTED  is more accurate

dev/ToastNotifications/ToastActivationInfo.h
```diff
@@ -7,10 +7,15 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
...
         ToastActivationInfo() = default;
 
+        ToastActivationInfo(winrt::guid const& taskClsid, winrt::Microsoft::Windows::ToastNotifications::ToastAssets const& assets) : m_taskClsid(taskClsid), m_assets(assets) {};
         static winrt::Microsoft::Windows::ToastNotifications::ToastActivationInfo CreateFromActivationGuid(winrt::guid const& taskClsid);
         static winrt::Microsoft::Windows::ToastNotifications::ToastActivationInfo CreateFromToastAssets(winrt::Microsoft::Windows::ToastNotifications::ToastAssets const& assets);
         winrt::guid TaskClsid();
         winrt::Microsoft::Windows::ToastNotifications::ToastAssets Assets();
+
+    private:
+        const winrt::guid m_taskClsid;
```
hulumane, line 17: Set this to GUID_NULL by default

dev/ToastNotifications/ToastNotificationManager.cpp
```diff
@@ -0,0 +1,135 @@
...
+    using namespace Windows::ApplicationModel::Core;
+}
+
+namespace winrt::Microsoft::Windows::ToastNotifications::implementation
+{
+    static wil::unique_com_class_object_cookie s_toastcomActivatorRegistration;
+
+    winrt::Microsoft::Windows::ToastNotifications::ToastNotificationManager ToastNotificationManager::Default()
+    {
+        throw hresult_not_implemented();
```
hulumane, line 39: Implement this please. It will return a DEfault ToastManager object

dev/ToastNotifications/ToastNotificationUtility.cpp
```diff
@@ -0,0 +1,190 @@
...
+        return guidWideStr;
+    }
+    else
+    {
+        THROW_HR_IF(status, FAILED_WIN32(status));
+    }
+    return registeredGuidBuffer;
+}
+
+std::wstring RetrieveToastGuid()
```
hulumane, line 67: The method name is incorrect. It does not necessarily return a GUID. It should be called RetrieveAppId()

dev/ToastNotifications/ToastNotificationUtility.cpp
```diff
@@ -0,0 +1,190 @@
...
+        wil::unique_cotaskmem_string newToastGuidString;
+        THROW_IF_FAILED(StringFromCLSID(newToastGuid, &newToastGuidString));
+
+        std::wstring guidWideStr{ newToastGuidString.get() };
+        RegisterValue(hKey, L"ToastGUID", reinterpret_cast<const BYTE*>(guidWideStr.c_str()), REG_SZ, guidWideStr.size() * sizeof(wchar_t));
+        return guidWideStr;
+    }
+    else
+    {
+        THROW_HR_IF(status, FAILED_WIN32(status));
```
hulumane, line 62: Why are we returning an error if the GUID is found? Am I missing something? We should return the GUID which is stored in the registry for that case.

pmpurifoy: The two cases we want to cover is ERROR_NOT_FILE_FOUND/ERROR_SUCCESS, any other error should be thrown.

hulumane: Correct. So this is a bug right? It needs another else if

pmpurifoy: Doesn't FAILED_WIN32(ERROR_SUCCESS) return true? I've seen other instances of this pattern with reading from the registry.

hulumane: Ah good point. I totally missed this. You tested that right? :)

dev/ToastNotifications/ToastNotificationUtility.cpp
```diff
@@ -0,0 +1,190 @@
...
+    HRESULT status = RegGetValueW(
+        hKey.get(),
+        nullptr,
+        L"CustomActivator",
+        RRF_RT_REG_SZ,
+        nullptr,
+        &registeredGuidBuffer,
+        &bufferLength);
+
+    if (status == ERROR_FILE_NOT_FOUND)
```
hulumane, line 181: Don't we want to THROW for File_Not_FOUND instead of returning empty string?

dev/ToastNotifications/ToastNotificationUtility.cpp
pmpurifoy: Yup! I ran the unpackaged test app twice and was able to retrieve the ToastGuid/ComActivatorGuid the second time after creating them.

dev/ToastNotifications/ToastNotificationManager.cpp
```diff
@@ -0,0 +1,135 @@
...
+        {
+            wil::unique_cotaskmem_string processName;
+            THROW_IF_FAILED(GetCurrentProcessPath(processName));
+
+            std::wstring appIdentifier{ RetrieveToastGuid() };
+            THROW_IF_FAILED(PushNotifications_RegisterFullTrustApplication(appIdentifier.c_str(), GUID_NULL));
+
+            std::wstring storedComActivatorString{ RetrieveComActivatorGuid(appIdentifier) };
+
+            if (storedComActivatorString.empty())
```
hulumane, line 57: We should move this empty string check to RetrieveComActivatorGUID. It should create a new GUID if it hasn't created one already

hulumane: ie. It should mimic RetreiveToastGuid above

pmpurifoy: I believe I used the stored com activator as the condition for whether or not we have registered assets/comserver. If we move the com activator guid creation inside RetrieveComActivatorGuid (which makes more sense), we don't know if have registered assets/comserver before. We could reregister everything but I thought that would be weird. Another option is to return a std::pair<winrt::guid, bool> for whether or not we were created yet. What do you think?

hulumane: RetrieveCOMActivatorGuid can take in the Assets along with AppIdentifier and replace the assets in registry. I think its ok to replace existing Assets if they exist already: no harm. The app just decided to change its branding(Display name and icon) which is a real scenario right?

hulumane: And you can rename the function to RegisterComAssetsAndGuid. It takes in appid and assets. Figures out  a new com activator guid if not present already and re-registers everything.

dev/ToastNotifications/ToastNotificationUtility.cpp
```diff
@@ -0,0 +1,190 @@
...
+        subKey.c_str(),
+        0,
+        nullptr,
+        REG_OPTION_NON_VOLATILE,
+        KEY_ALL_ACCESS,
+        nullptr,
+        &hKey,
+        nullptr));
+
+    std::wstring comRegistrationExeString{ L"\"" };
```
hulumane, line 122: I was taught this cool new trick to avoid \"

std::wstring comRegistrationExeString{ LR(""") };

dev/ToastNotifications/ToastNotificationUtility.cpp
```diff
@@ -0,0 +1,190 @@
...
+    RegisterValue(hKey, L"IconUri", reinterpret_cast<const BYTE*>(activationInfo.IconPath().AbsoluteUri().c_str()), REG_EXPAND_SZ, activationInfo.IconPath().AbsoluteUri().size() * sizeof(wchar_t));
+
+    std::wstring wideStringClsid{ clsid.get() };
+    RegisterValue(hKey, L"CustomActivator", reinterpret_cast<const BYTE*>(wideStringClsid.c_str()), REG_SZ, wideStringClsid.size() * sizeof(wchar_t));
+}
+
+void RegisterComServer(wil::unique_cotaskmem_string const& processName, wil::unique_cotaskmem_string const& clsid)
+{
+    wil::unique_hkey hKey;
+    std::wstring subKey{ c_clsIdPath + clsid.get() + L"\\LocalServer32" };
```
hulumane, line 109: Trick:
std::wstring subKey{ c_clsIdPath + clsid.get() + LR("\LocalServer32") };

dev/ToastNotifications/ToastNotificationUtility.cpp
hulumane: See my comment with using LR above

dev/ToastNotifications/ToastNotificationUtility.h
```diff
@@ -0,0 +1,47 @@
...
+winrt::event<winrt::Windows::Foundation::EventHandler<winrt::Microsoft::Windows::ToastNotifications::ToastActivatedEventArgs>>& GetToastHandlers();
+int& GetToastHandleCount();
+
+inline const int GUID_LENGTH = 36;
+
+inline std::wstring ConvertPathToKey(std::wstring path)
+{
+    for (int i = 0; i < path.length(); i++)
+    {
+        if (path[i] == '\\')
```
hulumane, line 22: LR

dev/ToastNotifications/ToastNotificationUtility.cpp
```diff
@@ -0,0 +1,190 @@
...
+    using namespace Windows::ApplicationModel::Core;
+}
+
+std::wstring RetrieveUnpackagedToastGuid()
+{
+    wil::unique_cotaskmem_string processName;
+    THROW_IF_FAILED(GetCurrentProcessPath(processName));
+
+    std::wstring wideStringProcessName{ processName.get() };
+    std::wstring subKey{ c_appIdentifierPath + ConvertPathToKey(wideStringProcessName) };
```
hulumane, line 23: This needs to have in comments as to how a full key looks like.

dev/ToastNotifications/ToastNotificationUtility.cpp
```diff
@@ -0,0 +1,190 @@
...
+    comRegistrationExeString.append(L"\" ");
+    comRegistrationExeString.append(L"----ToastActivated:");
+
+    RegisterValue(hKey, nullptr, reinterpret_cast<const BYTE*>(comRegistrationExeString.c_str()), REG_SZ, (comRegistrationExeString.size() * sizeof(wchar_t)));
+}
+
+void UnRegisterComServer(std::wstring const& clsid)
+{
+    wil::unique_hkey hKey;
+    std::wstring subKey{ c_clsIdPath + clsid + L"\\LocalServer32" };
```
hulumane, line 133: LR("\LocalServer32")

dev/ToastNotifications/ToastNotificationUtility.cpp
sharath2727: LR"(\LocalServer32)"

test/TestApps/ToastNotificationsTestApp/ToastNotificationsTestApp.cpp
```diff
@@ -88,8 +133,10 @@ std::map<std::string, bool(*)()> const& GetSwitchMapping()
 {
     static std::map<std::string, bool(*)()> switchMapping = {
         { "VerifyFailedRegisterActivatorUsingNullClsid", &VerifyFailedRegisterActivatorUsingNullClsid },
+        { "VerifyFailedRegisterActivatorUsingNullClsid_Unpackaged", &VerifyFailedRegisterActivatorUsingNullClsid_Unpackaged},
         { "VerifyFailedRegisterActivatorUsingNullAssets", &VerifyFailedRegisterActivatorUsingNullAssets },
         { "VerifyRegisterActivatorandUnRegisterActivatorUsingClsid", &VerifyRegisterActivatorandUnRegisterActivatorUsingClsid },
+        { "VerifydRegisterActivatoandUnRegisterActivatorUsingAssets_Unpackaged", &VerifydRegisterActivatoandUnRegisterActivatorUsingAssets_Unpackaged },
```
pmpurifoy, line 139: Verifyd->Verify

test/TestApps/ToastNotificationsTestApp/ToastNotificationsTestApp.cpp
```diff
@@ -4,52 +4,97 @@
...
+    winrt::ToastNotificationManager::Default().RegisterActivator(activationInfo);
 
-    ToastNotificationManager::Default().RegisterActivator(activationInfo);
+    winrt::ToastNotificationManager::Default().UnregisterActivator();
 
-    ToastNotificationManager::Default().UnregisterActivator();
+    return true;
+}
+
+bool VerifydRegisterActivatoandUnRegisterActivatorUsingAssets_Unpackaged()
```
pmpurifoy, line 89: VerifyRegisterActivatorAndUnregisterActivatorUsingAssets_Unpackaged()

test/ToastNotificationTests/APITests.cpp
```diff
@@ -134,14 +175,19 @@ namespace Test::ToastNotifications
...
-            // TODO
+            RunTestUnpackaged(L"VerifyFailedRegisterActivatorUsingNullAssets_Unpackaged", testWaitTime());
         }
 
         TEST_METHOD(VerifyRegisterActivatorandUnRegisterActivatorUsingClsid)
         {
             RunTest(L"VerifyRegisterActivatorandUnRegisterActivatorUsingClsid", testWaitTime());
         }
 
+        TEST_METHOD(VerifydRegisterActivatoandUnRegisterActivatorUsingAssets_Unpackaged)
```
pmpurifoy, line 186: VerifyRegisterActivatorAndUnregisterActivatorUsingAssets_Unpackaged()

dev/ToastNotifications/ToastActivationCallback.h
```diff
@@ -0,0 +1,38 @@
...
+        LPCWSTR invokedArgs,
+        [[maybe_unused]] NOTIFICATION_USER_INPUT_DATA const* data,
+        [[maybe_unused]] ULONG dataCount) noexcept;
+};
+
+struct ToastActivationCallbackFactory : winrt::implements<ToastActivationCallbackFactory, IClassFactory>
+{
+    STDMETHODIMP CreateInstance(_In_opt_ IUnknown* aggregateInterface, _In_ REFIID interfaceId, _Outptr_ VOID** object) noexcept final try
+    {
+        RETURN_HR_IF(CLASS_E_NOAGGREGATION, aggregateInterface != nullptr);
```
hulumane, line 21: RETURN_HR_IF_NULL

hulumane: Actually why do we need this check at all? This variable is a noop in any case

dev/ToastNotifications/ToastActivationCallback.cpp
```diff
@@ -0,0 +1,49 @@
...
+    {
+        userInput.Insert(data[i].Key, data[i].Value);
+    }
+
+    winrt::ToastActivatedEventArgs activatedEventArgs = winrt::make<winrt::Microsoft::Windows::ToastNotifications::implementation::ToastActivatedEventArgs>(invokedArgs, userInput);
+
+    
+    if (GetToastHandleCount() > 0)
+    {
+        /* Process is already Launched get toast event handlers for the activatedeventargs */
```
hulumane, line 36: Please change this comment a little. Something like this:
/* We get and invoke the active Foreground toast event handlers with the activatedeventargs since the process has already launched*/

dev/ToastNotifications/ToastActivationCallback.cpp
```diff
@@ -0,0 +1,49 @@
...
+        GetToastHandlers()(*this, activatedEventArgs);
+    }
+    else
+    {
+        /* Process Launch */
```
hulumane, line 41: Please change the comment here to something like this:
/* Since this activation results in a process launch, we cache the activatedEventArgs in the COM static store and fire an event to let the main thread know that it is okay to infer into AppLifeCycle::GetToastActivatedEventArgs(). */

dev/ToastNotifications/ToastActivationInfo.cpp
```diff
@@ -0,0 +1,38 @@
...
+        THROW_HR_IF(E_INVALIDARG, (taskClsid == winrt::guid(GUID_NULL)));
+
+        return winrt::make<ToastActivationInfo>(taskClsid);
+    }
+
+    winrt::Microsoft::Windows::ToastNotifications::ToastActivationInfo ToastActivationInfo::CreateFromToastAssets(winrt::Microsoft::Windows::ToastNotifications::ToastAssets const& assets)
+    {
+        THROW_HR_IF_MSG(E_ILLEGAL_METHOD_CALL, AppModel::Identity::IsPackagedProcess(), "Not applicable for packaged applications");
+
+        THROW_HR_IF(E_INVALIDARG, (assets == nullptr));
```
hulumane, line 20: THROW_HR_IF_NULL(E_INVALIDARG, assets)

hulumane: Although strictly speaking, can this actually be nullptr? Its a pass by reference right?

dev/ToastNotifications/ToastNotificationManager.cpp
```diff
@@ -0,0 +1,115 @@
...
+#include <winrt/Windows.Storage.h>
+#include <winrt/base.h>
+#include "ToastActivationCallback.h"
+#include "externs.h"
+#include "ToastNotificationUtility.h"
+#include <frameworkudk/pushnotifications.h>
+
+static winrt::event<winrt::Windows::Foundation::EventHandler<winrt::Microsoft::Windows::ToastNotifications::ToastActivatedEventArgs>> g_toastHandlers;
+
+static int g_toastHandlerCount = 0;
```
hulumane, line 14: Curious if we really need this variable? Can't we infer this from g_toastHandlers assuming it has a Size() method?

dev/ToastNotifications/ToastNotificationManager.cpp
```diff
@@ -0,0 +1,115 @@
...
+        s_toastcomActivatorRegistration.reset();
+
+        if (!AppModel::Identity::IsPackagedProcess())
+        {
+            std::wstring storedComActivatorString{ RegisterComActivatorGuidAndAssets(nullptr /* ToastActivationInfo */)};
+
+            UnRegisterComServer(storedComActivatorString);
+            UnRegisterAppIdentifierFromRegistry();
+
+            // TODO: Remove ToastGuid reference from LRP
```
hulumane, line 80: Seperate Task for this?

dev/ToastNotifications/ToastNotificationUtility.cpp
```diff
@@ -0,0 +1,213 @@
...
+        nullptr,
+        L"CustomActivator",
+        RRF_RT_REG_SZ,
+        nullptr,
+        &registeredGuidBuffer,
+        &bufferLength);
+
+    if (status == ERROR_FILE_NOT_FOUND)
+    {
+        wil::unique_cotaskmem_string processName;
```
hulumane, line 192: This if block needs a check: THROW_HR_IF_NULL(E_UNEXPECTED, details)

dev/ToastNotifications/ToastNotificationUtility.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+        std::wstring guidWideStr{ newToastGuidString.get() };
+        RegisterValue(hKey, L"ToastGUID", reinterpret_cast<const BYTE*>(guidWideStr.c_str()), REG_SZ, guidWideStr.size() * sizeof(wchar_t));
+        return guidWideStr;
+    }
+    else
+    {
+        THROW_HR_IF(status, FAILED_WIN32(status));
+    }
+    return registeredGuidBuffer;
```
hulumane, line 66: nit: move this return to the else block

dev/ToastNotifications/ToastNotificationUtility.cpp
sharath2727: THROW_HR_IF(E_INVALIDARG, (details == nullptr)); -  we are already checking the details in the RegisterActivator API i.e. the caller of this API.

dev/ToastNotifications/ToastNotificationManager.cpp
sharath2727: @pmpurifoy  can you please let me know which task of yours that you are going to handle this?

dev/ToastNotifications/ToastActivationInfo.cpp
sharath2727: @hulumane another design question here: What if the assets is not nullptr but if the asset properties i.e. the displayname or the iconuri is a nullptr should we throw pavan?

dev/ToastNotifications/ToastNotificationManager.cpp
```diff
@@ -0,0 +1,115 @@
...
+static int g_toastHandlerCount = 0;
+
+winrt::event<winrt::Windows::Foundation::EventHandler<winrt::Microsoft::Windows::ToastNotifications::ToastActivatedEventArgs>>& GetToastHandlers()
+{
+    return g_toastHandlers;
+}
+
+int& GetToastHandleCount()
+{
+    return g_toastHandlerCount;
```
robertzhou-wpn, line 23: Not sure, but I do not find logic in the PR to change this value. I am wondering when and where this value will be changed.

dev/ToastNotifications/ToastNotificationManager.cpp
pmpurifoy: https://github.com/microsoft/WindowsAppSDK/pull/1923

This PR has the LRP remove toast function.

dev/ToastNotifications/ToastNotificationManager.cpp
pmpurifoy: There is no size method but there is a bool to check registered delegates which would cleaner.

https://docs.microsoft.com/en-us/uwp/cpp-ref-for-winrt/event#eventoperator-bool

dev/ToastNotifications/ToastActivationInfo.cpp
pmpurifoy: If IconUri is blank, shell uses a default icon. I'm not sure about the DisplayName case.

dev/ToastNotifications/ToastActivationCallback.h
pmpurifoy: We have the same check in PushNotificationBackgroundTask.h. Which is also an unused variable similar to this. To be honest I'm unsure what CLASS_E_NOAGGREGATION does. We can probably remove it.

riverar: Callers that pass in a `aggregateInterface` are expecting an aggregate object (providing access to various inner and outer interfaces/methods). I think it's important you keep this in place to fulfill the contract, indicating to callers that aggregation is not supported.

See also: https://docs.microsoft.com/en-us/windows/win32/api/unknwn/nf-unknwn-iclassfactory-createinstance

dev/ToastNotifications/ToastActivationInfo.cpp
sharath2727: Discussed this offline, we are going to throw if the developer fails to provide display name or the iconuri. Thanks.

dev/ToastNotifications/ToastNotificationUtility.cpp
pmpurifoy: Yup. This works.

dev/ToastNotifications/ToastNotificationUtility.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+    wil::unique_hkey hKey;
+    THROW_IF_FAILED(RegCreateKeyEx(
+        HKEY_CURRENT_USER,
+        subKey.c_str(),
+        0,
+        nullptr,
+        REG_OPTION_NON_VOLATILE,
+        KEY_ALL_ACCESS,
+        nullptr,
```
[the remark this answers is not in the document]
pmpurifoy:
```suggestion
        nullptr /* lpSecurityAttributes */,
```

dev/ToastNotifications/ToastNotificationUtility.cpp
```diff
@@ -0,0 +1,212 @@
...
+        THROW_IF_FAILED(StringFromCLSID(comActivatorGuid, &comActivatorGuidString));
+
+        RegisterAssets(appIdentifier, details.Assets(), comActivatorGuidString);
+        RegisterComServer(processName, comActivatorGuidString);
+
+        return comActivatorGuidString.get();
+    }
+    else
+    {
+        THROW_HR_IF(status, FAILED_WIN32(status));
```
hulumane, line 209: I believe we have a wil check for win32 error codes. Trying to find the right macro to use here

hulumane: THROW_IF_WIN32_ERROR

hulumane (approved):

dev/ToastNotifications/ToastNotificationManager.cpp
```diff
@@ -0,0 +1,108 @@
...
+    void ToastNotificationManager::RegisterActivator(winrt::Microsoft::Windows::ToastNotifications::ToastActivationInfo const&  details)
+    {
+        THROW_HR_IF_MSG(E_INVALIDARG, s_toastcomActivatorRegistration, "ComActivator already registered.");
+
+        THROW_HR_IF_NULL(E_INVALIDARG, details);
+
+        std::wstring storedComActivatorString;
+        if (!AppModel::Identity::IsPackagedProcess())
+        {
+            storedComActivatorString = RegisterComActivatorGuidAndAssets(details);
```
hulumane, line 45: As a general pattern, it's weird to call Register in an Unregister function. What you really need is a helper called GetActivatorGuid that used by this method and RegisterComActivatorGuidAndAssets.

hulumane: The comment is meant for the UnregisterActivator API below

dev/ToastNotifications/ToastNotificationUtility.h
```diff
@@ -0,0 +1,49 @@
...
+
+void RegisterAssets(std::wstring const& appId, winrt::Microsoft::Windows::ToastNotifications::ToastAssets const& activationInfo, wil::unique_cotaskmem_string const& clsid);
+
+void RegisterComServer(wil::unique_cotaskmem_string const& processName, wil::unique_cotaskmem_string const& clsid);
+
+void UnRegisterComServer(std::wstring const& clsid);
+
+void UnRegisterAppIdentifierFromRegistry();
+
+HRESULT GetActivatorGuid(std::wstring& activatorGuid);
```
hulumane, line 47: mark as noexcept

dev/ToastNotifications/ToastNotificationUtility.cpp
```diff
@@ -0,0 +1,221 @@
...
+    std::wstring subKey{ c_appIdentifierPath + appIdentifier };
+
+    THROW_IF_FAILED(RegDeleteKeyEx(
+        HKEY_CURRENT_USER,
+        subKey.c_str(),
+        KEY_WOW64_64KEY,
+        0));
+}
+
+HRESULT GetActivatorGuid(std::wstring& activatorGuid)
```
hulumane, line 159: Put this in a try catch return block. Methods either return HRESULT or they throw. They don't do both.

dev/ToastNotifications/ToastNotificationUtility.cpp
```diff
@@ -0,0 +1,222 @@
...
+        THROW_IF_FAILED(CoCreateGuid(&newToastGuid));
+
+        wil::unique_cotaskmem_string newToastGuidString;
+        THROW_IF_FAILED(StringFromCLSID(newToastGuid, &newToastGuidString));
+
+        std::wstring guidWideStr{ newToastGuidString.get() };
+        RegisterValue(hKey, L"ToastGUID", reinterpret_cast<const BYTE*>(guidWideStr.c_str()), REG_SZ, guidWideStr.size() * sizeof(wchar_t));
+        return guidWideStr;
+    }
+    else
```
danielayala94, line 62: Technically you don't need the else block.

dev/ToastNotifications/ToastNotificationUtility.h
```diff
@@ -0,0 +1,49 @@
...
+#include <ToastActivatedEventArgs.h>
+
+const std::wstring c_appIdentifierPath{ LR"(Software\Classes\AppUserModelId\)" };
+const std::wstring c_clsIdPath{ LR"(Software\Classes\CLSID\)" };
+const std::wstring c_quote{ LR"(")" };
+const std::wstring c_toastActivatedArgument{ L" ----ToastActivated:" };
+
+winrt::event<winrt::Windows::Foundation::EventHandler<winrt::Microsoft::Windows::ToastNotifications::ToastActivatedEventArgs>>& GetToastHandlers();
+
+inline const int GUID_LENGTH = 39; // GUID + '{' + '}' + '/0
```
danielayala94, line 16: nit:

```suggestion
inline const int GUID_LENGTH = 39; // GUID + '{' + '}' + '/0'
```

danielayala94 (approved):
Just nits as comments. Also, it seems like this PR is a bit more than introducing the registration APIs. Please update the description accordingly.

PR #1970, opened 2022-01-12
Forward integrate main into WNP_ToastNotifications

loneursid:
FI main into pseudo feature branch - no conflicts

### DIFF
--- a/build/ProjectReunion-BuildFoundation.yml
+++ b/build/ProjectReunion-BuildFoundation.yml
@@ -436,20 +436,18 @@ jobs:
     prereleaseVersionTag: nightly
     buildPool: $(ProjectReunionBuildPool)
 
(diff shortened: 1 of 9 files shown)


danielayala94:
It seems like we don't need #1968. It is a subset of this PR.

danielayala94 (approved):

loneursid:
@danielayala94 - We do, we do also need 1968! :) This PR is targeting the old (created by Sharath) pseudo feature branch, while 1968 is targeting the new full featured feature branch. I want to keep both up to date so when merging the old one into the new one, we'll only see changes brought on for toast notifications.

danielayala94:
@loneursid Ahhh I see. Thanks for clarifying. Didn't see it was because of a different target.

PR #1976, opened 2022-01-12
Add ToastRegistration functions to PushNotificationsLongRunningTask

pmpurifoy:
- Includes AddToastRegistration([in] LPCWSTR processName, [in] LPCWSTR toastGuid) & RemoveToastRegistration([in] LPCWSTR processName) to PushNotificationsLongRunningTask so unpackaged applications can receive Toasts
- Adds ToastNotificationManager to handle all ToastRegistrations


dev/PushNotifications/PushNotificationsLongRunningTask.ProxyStub/NotificationsLongRunningProcess.idl
```diff
@@ -28,6 +28,10 @@ interface INotificationsLongRunningPlatform : IUnknown
     HRESULT RegisterForegroundActivator([in] IWpnForegroundSink* sink, [in] LPCWSTR processName);
 
     HRESULT UnregisterForegroundActivator([in] LPCWSTR processName);
+
+    HRESULT AddToastRegistration([in] LPCWSTR processName, [in] LPCWSTR toastGuid);
```
hulumane, line 32: If I understand the design correctly, this isn't always a GUID. It can also be an AUMID which is the appIdentifier for the packaged app. So this needs to be renamed appropriately. One idea is to just call it appId and leve it there.

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
```diff
@@ -28,3 +32,17 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationListener::OnRawNotificationReceived
...
 }
 CATCH_RETURN()
+
+STDMETHODIMP_(HRESULT __stdcall) NotificationListener::OnToastNotificationReceived(
+    ABI::Microsoft::Internal::ToastNotifications::INotificationProperties* notificationProperties,
+    ABI::Microsoft::Internal::ToastNotifications::INotificationTransientProperties* notificationTransientProperties) noexcept try
+{
+    auto lock = m_lock.lock_exclusive();
+
+    std::wstring toastGuid{ m_toastRegistrationManager->GetToastRegistration(m_processName) };
```
hulumane, line 42: SAme question from earlier. Is this always a GUID though?

dev/PushNotifications/PushNotificationsLongRunningTask.ProxyStub/NotificationsLongRunningProcess.idl
pmpurifoy: It can be GUID/AUMID. I will rename to appId.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -205,6 +206,8 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
 
                         wil::unique_cotaskmem_string processName;
                         THROW_IF_FAILED(GetCurrentProcessPath(processName));
+                        std::wstring appId{ RetrieveAppId() };
```
sharath2727, line 209: If you see above - there is a parameter called unpackagedAppUserModelId which you get it from the LRP (note you also register this Identifier in the wpn platform by calling registerfulltrustapplication)? Should you be using the same appId and register it in the registry?

sharath2727: If you call ToastNotificationManager::RegisterActivator for the retrieveAppId after getting the channel then it will create a new registration in the wpnplatform for the appId. Is that intended? One for toasts and one for raw for the same processname?

hulumane: Please infer to the design doc from @danielayala94 which runs through the mapping flow. In essence, we decided that the ToastREgistrationId and PushREgistrationId are different entities. The reason for this is simple, we don't want to bring up LRP(Long Running Process) for toast scenarios for perf reasons unless we also have a dependency on Push.

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.h
```diff
@@ -1,16 +1,24 @@
...
 
 class NotificationListener : public Microsoft::WRL::RuntimeClass<::ABI::Microsoft::Internal::PushNotifications::INotificationListener>
 {
 public:
-    HRESULT RuntimeClassInitialize(std::shared_ptr<ForegroundSinkManager> foregroundSinkManager, std::wstring appId, std::wstring processName);
+    HRESULT RuntimeClassInitialize(
+        std::shared_ptr<ForegroundSinkManager> foregroundSinkManager,
+        std::shared_ptr<ToastRegistrationManager> toastRegistrationManager,
+        std::wstring appId,
+        std::wstring processName);
```
hulumane, line 14: This should be marked as noexcept

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -133,6 +143,28 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::Unregiste
 }
 CATCH_RETURN()
 
+STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::AddToastRegistration(_In_ PCWSTR processName, _In_ PCWSTR toastGuid) noexcept try
```
hulumane: Please remove inferences to toastGuid everywhere.

dev/ToastNotifications/ToastNotificationManager.cpp
```diff
@@ -71,7 +77,10 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
             UnRegisterComServer(storedComActivatorString);
             UnRegisterAppIdentifierFromRegistry();
 
-            // TODO: Remove ToastGuid reference from LRP
+            wil::unique_cotaskmem_string processName;
+            THROW_IF_FAILED(GetCurrentProcessPath(processName));
+            auto notificationPlatform{ PushNotificationHelpers::GetNotificationPlatform() };
+            THROW_IF_FAILED(notificationPlatform->RemoveToastRegistration(processName.get()));
```
hulumane, line 83: We should also call this when the Push Activator is unregistered right?

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
pmpurifoy: This has been changed in the latest commit, not sure why it's not being reflected here yet.

dev/ToastNotifications/ToastNotificationManager.cpp
pmpurifoy: RemoveToastRegistration is called by UnregisterLongRunningActivator which is called when unregistering Push activator.

dev/ToastNotifications/ToastNotificationManager.cpp
```diff
@@ -71,7 +77,10 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
             UnRegisterComServer(storedComActivatorString);
             UnRegisterAppIdentifierFromRegistry();
 
-            // TODO: Remove ToastGuid reference from LRP
+            wil::unique_cotaskmem_string processName;
+            THROW_IF_FAILED(GetCurrentProcessPath(processName));
+            auto notificationPlatform{ PushNotificationHelpers::GetNotificationPlatform() };
```
hulumane, line 82: Where is the platform->AddToastREgistration() call in ToastNotificationManager? Its kinda strange to see Remove without a corresponding Add

pmpurifoy: The call happens in PushNotificationManager::CreateChannelAsync line 210 when a channel is successfully created. The mapping doesn't happen until a sink is registered.

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -133,6 +143,28 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::Unregiste
 }
 CATCH_RETURN()
 
+STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::AddToastRegistration(_In_ PCWSTR processName, _In_ PCWSTR appId) noexcept try
+{
+    auto lock{ m_lock.lock_exclusive() };
+    THROW_HR_IF(WPN_E_PLATFORM_UNAVAILABLE, m_shutdown);
+
+    m_toastRegistrationManager->Add(processName, appId);
```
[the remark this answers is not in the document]
pmpurifoy: There was an ordering issue when the sink is created. Currently the code reads AddToastRegistration, then RegisterLongRunningActivator. We can swap the order and then add the check, how does that sound?

dev/ToastNotifications/ToastNotificationManager.cpp
hulumane: Got it. I hope we are using the same appId generated in CreateChannelAsync when we call into REgisterToastActivator()

dev/PushNotifications/PushNotificationsLongRunningTask.ProxyStub/NotificationsLongRunningProcess.idl
```diff
@@ -28,6 +28,10 @@ interface INotificationsLongRunningPlatform : IUnknown
     HRESULT RegisterForegroundActivator([in] IWpnForegroundSink* sink, [in] LPCWSTR processName);
 
     HRESULT UnregisterForegroundActivator([in] LPCWSTR processName);
+
+    HRESULT AddToastRegistration([in] LPCWSTR processName, [in] LPCWSTR appId);
```
hulumane, line 32: Can we call this AddToastREgistrationMapping?

dev/PushNotifications/PushNotificationsLongRunningTask.ProxyStub/NotificationsLongRunningProcess.idl
```diff
@@ -28,6 +28,10 @@ interface INotificationsLongRunningPlatform : IUnknown
...
+
+    HRESULT RemoveToastRegistration([in] LPCWSTR processName);
```
hulumane, line 34: CAn we call this RemoveToastRegistrationMapping?

dev/ToastNotifications/ToastNotificationManager.cpp
pmpurifoy: The key for this mapping is processName->appId(generated GUID/AUMID). It is the same key in LRP.

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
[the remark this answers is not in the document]
hulumane: That sounds reasonable!

dev/ToastNotifications/ToastNotificationManager.cpp
pmpurifoy: We can call AddToastRegistration in RegisterToastActivator, but there wouldn't be a sink yet. I guess we have to call it here as well. Should we not throw if there is no sink yet as suggested in another comment?

hulumane: Do not throw there. Just return S_OK if there is no long running sink

hulumane: And yes. Please call AddToastREgistrationMapping in REgisterToastActivator. There is no other way to fix this.

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -133,6 +143,28 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::Unregiste
 }
 CATCH_RETURN()
 
+STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::AddToastRegistrationMapping(_In_ PCWSTR processName, _In_ PCWSTR appId) noexcept try
+{
+    auto lock{ m_lock.lock_exclusive() };
+    THROW_HR_IF(WPN_E_PLATFORM_UNAVAILABLE, m_shutdown);
+
+    m_toastRegistrationManager->Add(processName, appId);
```
hulumane, line 151: Please add a check to ensure that the longrunning sink has the process before calling these APIs

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -133,6 +143,28 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::Unregiste
 }
 CATCH_RETURN()
 
+STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::AddToastRegistrationMapping(_In_ PCWSTR processName, _In_ PCWSTR appId) noexcept try
```
hulumane, line 146: Why are you not calling this in CreatePushChannel anymore?

pmpurifoy: It should be in CreatePushChannel. I added it back.

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListenerManager.cpp
```diff
@@ -44,6 +45,12 @@ void NotificationListenerManager::RemoveListener(std::wstring appId)
     m_notificationListeners.erase(appId);
 }
 
+bool NotificationListenerManager::HasSinkForAppId(std::wstring const& appId)
+{
+    auto it = m_notificationListeners.find(appId);
+    return it != m_notificationListeners.end();
```
robertzhou-wpn, line 51: what about using m_notificationListeners.contains()?

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListenerManager.cpp
```diff
@@ -44,6 +45,12 @@ void NotificationListenerManager::RemoveListener(std::wstring appId)
     m_notificationListeners.erase(appId);
 }
 
+bool NotificationListenerManager::HasSinkForAppId(std::wstring const& appId)
+{
+    auto it = m_notificationListeners.find(appId);
```
robertzhou-wpn, line 50: Looks like we use auto lock = m_lock.lock_shared() when we check whether the listens map is empty or not.  should we also get a shared lock first in this method?

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListenerManager.cpp
pmpurifoy: The .contains() API is with c++20 and we are not allowed to upgrade :(

dev/PushNotifications/PushNotificationsLongRunningTask/ToastRegistrationManager.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+void ToastRegistrationManager::Remove(std::wstring const& appId)
+{
+    auto lock = m_lock.lock_exclusive();
+    m_registrationMap.erase(appId);
+}
+
+std::wstring ToastRegistrationManager::GetToastRegistration(std::wstring const& processName)
+{
+    auto lock = m_lock.lock_exclusive();
```
hulumane, line 18: this should be a shared lock.

hulumane (approved):

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -21,6 +21,7 @@
 #include "NotificationsLongRunningProcess_h.h"
 #include <TerminalVelocityFeatures-PushNotifications.h>
 #include "PushNotificationUtility.h"
+#include "ToastNotificationUtility.h"
```
hulumane, line 24: What are your plans on adding unit tests for these new APIs?

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListenerManager.cpp
```diff
@@ -37,15 +38,22 @@ void NotificationListenerManager::RemoveListener(std::wstring appId)
...
+    auto lock{ m_lock.lock_exclusive() };
 
     LOG_IF_FAILED(PushNotifications_UnregisterNotificationSinkForFullTrustApplication(appId.c_str()));
 
     m_notificationListeners.erase(appId);
 }
 
+bool NotificationListenerManager::HasSinkForAppId(std::wstring const& appId)
+{
+    auto lock{ m_lock.lock_shared() };
```
hulumane, line 50: I'm curious about this pattern. You are calling this in an exclusive lock while the method has a shared lock. You are infact deadlocking.

hulumane: Actually never mind. This is a public helper. I think you are good

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
```diff
@@ -1,12 +1,16 @@
...
+
+#include "pch.h"
 #include "../PushNotificationUtility.h"
 
 HRESULT NotificationListener::RuntimeClassInitialize(
     std::shared_ptr<ForegroundSinkManager> foregroundSinkManager,
+    std::shared_ptr<ToastRegistrationManager> toastRegistrationManager,
     std::wstring appId,
-    std::wstring processName)
+    std::wstring processName) noexcept
```
danielayala94, line 10: Is shared_ptr assignment exception safe?

danielayala94: For sure wstring is not, so this function can't be marked as noexcept.

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
```diff
@@ -28,3 +32,17 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationListener::OnRawNotificationReceived
     return S_OK;
 }
 CATCH_RETURN()
+
+STDMETHODIMP_(HRESULT __stdcall) NotificationListener::OnToastNotificationReceived(
+    ABI::Microsoft::Internal::ToastNotifications::INotificationProperties* notificationProperties,
```
danielayala94, line 37: Nit: you can do `using namespace` to shorten these lengthy ABI namespaces in the code.

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
```diff
@@ -28,3 +32,17 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationListener::OnRawNotificationReceived
...
+    ABI::Microsoft::Internal::ToastNotifications::INotificationTransientProperties* notificationTransientProperties) noexcept try
+{
+    auto lock = m_lock.lock_exclusive();
```
danielayala94, line 40: 1. The code isn't substantially modifying  the members, the map operation is just a query. This can be a `lock_shared`.
2. Please scope the lock to `appId` assignment. We don't want go all thru the UDK and hit wpncore (RPC call) while being in a lock.

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: Going to include unit tests for register/unregistering the toast from the LRP. I have plans for another PR that contains unit testing for the rest of the LRP functions :)

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
```diff
@@ -1,12 +1,16 @@
-﻿#include "pch.h"
+﻿#pragma once
+
+#include "pch.h"
 #include "../PushNotificationUtility.h"
 
 HRESULT NotificationListener::RuntimeClassInitialize(
     std::shared_ptr<ForegroundSinkManager> foregroundSinkManager,
+    std::shared_ptr<ToastRegistrationManager> toastRegistrationManager,
     std::wstring appId,
```
danielayala94, line 9: Can the wstrings be referenced?

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListenerManager.cpp
```diff
@@ -37,15 +38,22 @@ void NotificationListenerManager::RemoveListener(std::wstring appId)
...
 
-    auto lock = m_lock.lock_exclusive();
+    auto lock{ m_lock.lock_exclusive() };
 
     LOG_IF_FAILED(PushNotifications_UnregisterNotificationSinkForFullTrustApplication(appId.c_str()));
 
     m_notificationListeners.erase(appId);
 }
 
+bool NotificationListenerManager::HasSinkForAppId(std::wstring const& appId)
```
danielayala94, line 48: Please change the name to something like `HasListenerForAppId`. Let's avoid mixing terminology when possible.

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -20,11 +20,20 @@ void NotificationsLongRunningPlatformImpl::Initialize()
     // This is in case we later realize there are no apps to be tracked in the LRP.
     m_lifetimeManager.Setup();
 
-    m_storage = Storage::ApplicationData::Current().LocalSettings().CreateContainer(
-        L"LRP", Storage::ApplicationDataCreateDisposition::Always);
+    m_rawStorage = Storage::ApplicationData::Current().LocalSettings().CreateContainer(
+        L"Raw", Storage::ApplicationDataCreateDisposition::Always);
+
+    m_toastStorage = Storage::ApplicationData::Current().LocalSettings().CreateContainer(
```
danielayala94, line 26: Can the toast storage be managed by the ToastRegistrationManager? It seems odd to me that the actual registration process happens partially in the component, but this step doesn't.

dev/PushNotifications/PushNotificationsLongRunningTask/platform.h
```diff
@@ -23,13 +23,18 @@ struct __declspec(uuid(PUSHNOTIFICATIONS_IMPL_CLSID_STRING)) NotificationsLongRu
...
     STDMETHOD(UnregisterForegroundActivator)(_In_ PCWSTR processName) noexcept;
 
+    STDMETHOD(AddToastRegistrationMapping)(_In_ PCWSTR processName, _In_ PCWSTR appId) noexcept;
+
+    STDMETHOD(RemoveToastRegistrationMapping)(_In_ PCWSTR processName) noexcept;
 private:
 
     std::map<std::wstring, std::wstring> GetFullTrustApps();
     const std::wstring GetAppIdentifier(std::wstring const& processName);
+    const std::wstring BuildAppIdentifier(std::wstring const& processName);
```
danielayala94, line 33: Where is this being used?

dev/PushNotifications/externs.h
```diff
@@ -20,3 +20,4 @@ inline HRESULT GetCurrentProcessPath(wil::unique_cotaskmem_string& processName)
 {
     return wil::GetModuleFileNameExW(GetCurrentProcess(), nullptr, processName);
 };
+
```
danielayala94, line 23: Nit: please revert.

danielayala94 (changes requested):

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
hulumane: It has to be noexcept try catch return. Methods that return HRESULT should not throw exceptions.

dev/PushNotifications/PushNotificationsLongRunningTask/platform.h
pmpurifoy: This being used to replace GetAppIdentifier. GetAppIdentifier is only a getter now and BuildAppIdentifier will build an appId if it hasn't been built yet.

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListenerManager.cpp
pmpurifoy: We don't need this function anymore since we have GetAppIdentifier.

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
pmpurifoy: Can you explain a little about what this means?

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
pmpurifoy: This is a great idea. I'll move the storage to the ToastRegistrationManager class.

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
danielayala94: I mean, expect a wstring reference as a parameter instead of copying: 
```suggestion
    std::wstring& appId,
```

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
danielayala94: Not resolved yet.

dev/PushNotifications/PushNotificationsLongRunningTask/ToastRegistrationManager.h
```diff
@@ -8,15 +8,26 @@
 class ToastRegistrationManager
 {
 public:
-    ToastRegistrationManager() = default;
+    ToastRegistrationManager()
```
danielayala94, line 11: Can you please move the ctor impl to the cpp?

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -70,7 +63,7 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::RegisterF
     auto lock = m_lock.lock_exclusive();
     THROW_HR_IF(WPN_E_PLATFORM_UNAVAILABLE, m_shutdown);
 
-    const std::wstring appIdentifier = GetAppIdentifier(processName);
+    const std::wstring appIdentifier{ BuildAppIdentifier(processName) };
```
danielayala94, line 66: Can you add a comment over both GetAppIdentifier/BuildAppIdentifier saying that each function assume the caller is under a lock? Then both functions don't need locks.

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -120,9 +113,7 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::RegisterF
     auto lock = m_lock.lock_exclusive();
     THROW_HR_IF(WPN_E_PLATFORM_UNAVAILABLE, m_shutdown);
 
-    const std::wstring appId = GetAppIdentifier(processName);
-
-    m_notificationListenerManager.AddListener(appId, processName);
+    const std::wstring appId{ GetAppIdentifier(processName) };
```
danielayala94, line 116: If there is no appId for such processName, then we are going to add a sink anyway. I think we should throw with ERROR_NOT_FOUND. Thoughts?

Also, are there any SDK scenarios where we first do something else in the LRP than registering the app? i.e. register foreground activator for App1, then registering App1?

test/PushNotificationTests/LRPTests.cpp
```diff
@@ -179,5 +179,45 @@ namespace Test::PushNotifications
             VERIFY_IS_FALSE(isRunning);
         }
 
+        TEST_METHOD(AddToastRegistrationMappingNoSink)
+        {
+            VERIFY_SUCCEEDED(::CoInitializeEx(nullptr, COINITBASE_MULTITHREADED));
```
danielayala94, line 184: Do you have a task to move CoInitializeEx/CoUninitialize to the class setup/cleanup tests function?

test/PushNotificationTests/LRPTests.cpp
```diff
@@ -179,5 +179,45 @@ namespace Test::PushNotifications
...
+            auto scopeExit = wil::scope_exit(
+                [&]() { VERIFY_NO_THROW(CoUninitialize()); });
+
+            winrt::com_ptr notificationPlatform{ winrt::try_create_instance<INotificationsLongRunningPlatform>(_uuidof(NotificationsLongRunningPlatform), CLSCTX_ALL) };
+            VERIFY_IS_NOT_NULL(notificationPlatform.get());
+
+            wil::unique_cotaskmem_string unpackagedAumid;
+            VERIFY_NO_THROW(notificationPlatform->RegisterFullTrustApplication(L"TAEF.exe", c_remoteId, &unpackagedAumid));
```
danielayala94, line 216: The appId can be a class constant.

danielayala94 (changes requested):

test/PushNotificationTests/LRPTests.cpp
pmpurifoy: Moving to methodinit/unit in iteration

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
pmpurifoy: We decided as a team to return S_OK on adding foreground sinks and toast registratrions. Remove will not fail if the appId doesn't exist.

To my knowledge there is no other SDK scenarios that could break this.

test/TestApps/PushNotificationsDemoApp/main.cpp
```diff
@@ -92,16 +91,7 @@ winrt::Microsoft::Windows::PushNotifications::PushNotificationChannel RequestCha
...
-    {
-        // Major.Minor version, MinVersion=0 to find any framework package for this major.minor version
-        const UINT32 c_Version_MajorMinor{ 0x00040001 };
-        const PACKAGE_VERSION minVersion{};
-        RETURN_IF_FAILED(MddBootstrapInitialize(c_Version_MajorMinor, nullptr, minVersion));
-    }
-
-    ToastAssets assets(L"ToastNotificationApp", winrt::Windows::Foundation::Uri{ LR"(C:\Windows\System32\WindowsSecurityIcon.png)" });
-    auto activationInfo = ToastActivationInfo::CreateFromToastAssets(assets);
+    auto activationInfo = ToastActivationInfo::CreateFromActivationGuid(winrt::guid("28C29657-DB85-49D2-9974-C61094CA8280"));
```
danielayala94, line 94: Nice to see all these test improvements! 😊

danielayala94 (approved):
:shipit:

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListenerManager.cpp
```diff
@@ -21,11 +22,11 @@ void NotificationListenerManager::AddListener(std::wstring const& appId, std::ws
...
     THROW_HR_IF(E_INVALIDARG, processName.empty());
 
-    auto lock = m_lock.lock_exclusive();
+    auto lock{ m_lock.lock_exclusive() };
 
     // Make sure we keep the long running sink up-to-date with wpncore.
     ComPtr<INotificationListener> newListener;
-    THROW_IF_FAILED(MakeAndInitialize<NotificationListener>(&newListener, m_foregroundSinkManager, appId, processName));
+    THROW_IF_FAILED(MakeAndInitialize<NotificationListener>(&newListener, m_foregroundSinkManager, m_toastRegistrationManager, appId, processName));
     THROW_IF_FAILED(PushNotifications_RegisterNotificationSinkForFullTrustApplication(appId.c_str(), newListener.Get()));
```
hulumane, line 30: Out of proc API call in a lock.

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListenerManager.cpp
```diff
@@ -37,7 +38,7 @@ void NotificationListenerManager::RemoveListener(std::wstring appId)
 {
     THROW_HR_IF(E_INVALIDARG, appId.empty());
 
-    auto lock = m_lock.lock_exclusive();
+    auto lock{ m_lock.lock_exclusive() };
 
     LOG_IF_FAILED(PushNotifications_UnregisterNotificationSinkForFullTrustApplication(appId.c_str()));
```
hulumane, line 43: Out of proc API call in lock

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -76,7 +78,12 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::RegisterL
     auto lock = m_lock.lock_shared();
     THROW_HR_IF(WPN_E_PLATFORM_UNAVAILABLE, m_shutdown);
 
-    const std::wstring appId = GetAppIdentifier(processName);
+    const std::wstring appId{ GetAppIdentifier(processName) };
+    if (appId.empty())
```
hulumane, line 82: Add a comment here that NotificationsLongRunningPlatformImpl::RegisterFullTrustApplication should always be called before this or we silently ignore the API call

hulumane: Here and everywhere else where we have this check.

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -90,11 +97,12 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::Unregiste
     auto lock = m_lock.lock_shared();
     THROW_HR_IF(WPN_E_PLATFORM_UNAVAILABLE, m_shutdown);
 
-    const std::wstring appId = GetAppIdentifier(processName);
+    const std::wstring appId{ GetAppIdentifier(processName) };
```
hulumane, line 100: If this is empty, shouldn't we bail out of the function?

hulumane: I think we should and in addition probably need to call into this code before bailing out right?

if (m_notificationListenerManager.IsEmpty())
    {
        m_lifetimeManager.Setup();
    }

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -126,13 +137,38 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::Unregiste
     auto lock = m_lock.lock_exclusive();
     THROW_HR_IF(WPN_E_PLATFORM_UNAVAILABLE, m_shutdown);
 
-    const std::wstring appId = GetAppIdentifier(processName);
+    const std::wstring appId{ GetAppIdentifier(processName) };
```
hulumane, line 140: Shouldn't we bail out if this is empty?

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -126,13 +137,38 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::Unregiste
...
+{
+    auto lock{ m_lock.lock_exclusive() };
+    THROW_HR_IF(WPN_E_PLATFORM_UNAVAILABLE, m_shutdown);
+
+    // Don't register unless there is a sink available.
+    if (GetAppIdentifier(processName).empty())
+    {
+        return S_OK;
+    }
+    m_toastRegistrationManager->Add(processName, toastAppId);
```
hulumane, line 157: If this returns HRESULT, we should THROW_IF_FAILED

pmpurifoy: ToastRegistrationManager.Add is void function.

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
hulumane: This code i believe basically sets up LRP to shutdown in x seconds if there are no long running sinks registered. Please confirm though. @danielayala94 added this earlier.

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListenerManager.cpp
```diff
@@ -22,26 +22,31 @@ void NotificationListenerManager::AddListener(std::wstring const& appId, std::ws
...
     THROW_HR_IF(E_INVALIDARG, processName.empty());
 
-    auto lock{ m_lock.lock_exclusive() };
+    
 
     // Make sure we keep the long running sink up-to-date with wpncore.
     ComPtr<INotificationListener> newListener;
-    THROW_IF_FAILED(MakeAndInitialize<NotificationListener>(&newListener, m_foregroundSinkManager, m_toastRegistrationManager, appId, processName));
+    {
+        auto lock{ m_lock.lock_exclusive() };
```
hulumane, line 30: read lock right? Not exclusive.

dev/ToastNotifications/ToastNotificationUtility.cpp
```diff
@@ -188,6 +188,9 @@ CATCH_RETURN()
 
 std::wstring RegisterComActivatorGuidAndAssets(winrt::Microsoft::Windows::ToastNotifications::ToastActivationInfo const& details)
 {
+    std::wstring toastAppId{ RetrieveToastAppId() };
+    THROW_IF_FAILED(PushNotifications_RegisterFullTrustApplication(toastAppId.c_str(), GUID_NULL));
```
hulumane, line 192: Unclear why this was added here and not done seperately. Can you elaborate?

pmpurifoy: Since this is where the ComActivator/Assets are being put in the registry, I put the registration here as well because the toastAppId is being generated here. We can pull this out of this function and put it in RegisterActivator. What do you think?

hulumane: Yes. I think we should do that and break it up. This is specifically for the COM activator and Nitro scenario in my interpretation.

hulumane (approved):

PR #1979, opened 2022-01-13
Enable SBOM generation for WindowsAppSDK

kythant:
SBOM generation is enabled in each platform and configuration of the MRT and the main Foundation build. 
The json from the generations will be placed in mrtcore_binaries and windowsappsdk_binaries build artifacts for each platform and configuration.


BenJKuhn (approved):

DefaultRyan (approved):
Looking good. Could you send me a link to that new ManifestGenerator task?

PR #1986, opened 2022-01-14
Generate cert

DrusTheAxe:
Generate the test certificate instead of relying on the checked in files (MSTest.*).

TL;DR changes include:

1. DevCheck creates .user directory (think of it like temp as a local dir for per-developer artifacts that aren't temporary)
2. DevCheck creates a _password protected_ test certificate
3. MakeMSIX.targets to run MakeAppx.exe and SignTool.exe
4. Change projects creating *.MSIX to use MakeMSIX.targets (obsoleting the previous NMake- and .cmd-based logic)

MakeCert.exe is deprecated. Recommendation points to Powershell cmdlets, but those can't create a .pfx w/o a password. Doing the right thing requires a password hanging around so later we can "signtool.exe ...foo.pfx /p password blah.msix", and thus is born .user/winappsdk.certificate.test.*

* .cer = Certificate. DevCheck adds to your certificate store
* .pfx = Private key used to sign test .msix
* .pwd = Password used to create and unlock the .pfx
* .thumbprint = Certificate thumbprint

While easy enough to update existing .cmd to use an environment variable with the content of a file e.g.

```
for /F "delims=" %i in (winappsdk.certificate.test.pwd) do set content=!content!%i
```

NMake makefiles can't accomplish the same :-( So I bit the bullet and created an MSBuild replacement. MakeMSIX.targets provides the common logic, and the various test package projects were updated to use it.


MakeMsix.targets
```diff
@@ -0,0 +1,78 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <!-- Define the property MakeMsixVerbosity=high for additional logging (default=normal) -->
+  <PropertyGroup>
+    <MakeMsixVerbosity Condition="'$(MakeMsixVerbosity)' == ''">normal</MakeMsixVerbosity>
```
aeloros, line 7: Is there any sane way to determine a more dynamic default than 'normal' based on the MSBuild verbosity?

MakeMsix.targets
```diff
@@ -0,0 +1,78 @@
...
+    <MakeMsixSigned Condition="'$(MakeMsixSigned)' == ''">true</MakeMsixSigned>
```
aeloros, line 8: Is this always true, or would we ever want false?  I'm not suggesting getting rid of it.  I'm more curious if (like the above) it can be linked to some external force, but default to true here.  For instance, is there some way to say "if the property on the pipeline page is defined, then let it override this.  Otherwise default to true"?

riverar:
I don't recommend you use PFX files on disk at all. Instead, the certificate(s) would ideally be loaded into Azure HSM (either by yourself or by Azure Code Signing team) and you use an appropriate cloud-enabled code signing tool.

Today, for example, I sign our app's binaries with Azure HSM + azuresigntool.

MakeMsix.targets
```diff
@@ -0,0 +1,78 @@
...
+  </PropertyGroup>
+
+  <Target Name="MakeMSIX"
+    Inputs="@(MakeMsixInputs);@(MakeMsixInputsWithLocations);@(MakeMsixInputAssets);$(MakeMsixOutputFilename)"
```
aeloros, line 12: How are MakeMsixInputs and MakeMsixInputsWithLocations different?  Is it the same list but one does not have a path component?

aeloros: Also, why is the output an input?

MakeMsix.targets
```diff
@@ -0,0 +1,78 @@
...
+    Outputs="$(OutDir)$(TargetName)"
```
aeloros, line 13: How is $(MakeMsixOutputFilename) different than $(OutDir)$(TargetName)?

MakeMsix.targets
```diff
@@ -0,0 +1,78 @@
...
+    <!-- Ensure MakeMsixOutputFilename is defined (defining it here if necessary, as per Microsoft.Cpp.Default.props line ~189) -->
+    <PropertyGroup>
+      <MakeMsixOutputFilename Condition="'$(MakeMsixOutputFilename)' == ''">$(ProjectName).msix</MakeMsixOutputFilename>
+    </PropertyGroup>
+    <Message Text="MakeMSIX: TargetName=$(MakeMsixOutputFilename)" Importance="$(MakeMsixVerbosity)" />
+    <Message Text="MakeMSIX: OutDir=$(OutDir)" Importance="$(MakeMsixVerbosity)" />
+
+    <PropertyGroup>
+      <_MakeMsix_WorkDir>$(OutDir)msix</_MakeMsix_WorkDir>
+      <_MakeMsix_OutDir>$(OutDir)</_MakeMsix_OutDir>
```
aeloros, line 34: Please use the already defined $(AppxPackageDir) which is set in EnsureOutputLayout.props.  This creates a centralized place for MSIX output in the build that works for the differently managed pipeline and local dev builds.  The property should already be set, and if it is not, then there is an ordering issue to be diagnosed.

DrusTheAxe:
> I don't recommend you use PFX files on disk at all. Instead, the key material should ideally be loaded into Azure Key Vault (either by yourself or automatically by the Azure Code Signing team) and tapped into via an appropriate cloud-enabled code signing tool.
> 
> Today, for example, I sign our app's binaries with Azure Key Vault + azuresigntool.

This is for inner loop for test packages.

Requiring AzureKeyVault for this would negatively impact dev inner loop. Plus I'm not sure we could enable AzureKeyVault to grant "everyone" access to this; we _do_ desire enabling non-Microsofties to build all of WinAppSDK (build WindowsAppRuntime.sln and run its tests, among other things). We're working to address that (thank you for your patience :-)

Build pipelines (e.g. creating release .msix) do use AzureKeyVault (among other things).

MakeMsix.targets
```diff
@@ -0,0 +1,78 @@
...
+    <MakeMsixSigned Condition="'$(MakeMsixSigned)' == ''">true</MakeMsixSigned>
+  </PropertyGroup>
+
+  <Target Name="MakeMSIX"
+    Inputs="@(MakeMsixInputs);@(MakeMsixInputsWithLocations);@(MakeMsixInputAssets);$(MakeMsixOutputFilename)"
+    Outputs="$(OutDir)$(TargetName)"
+    AfterTargets="AfterBuild">
+
+    <!-- Initial state diagnostics -->
+    <Message Text="MakeMSIX Pre: MSBuildProjectName=$(MSBuildProjectName)" Importance="$(MakeMsixVerbosity)" />
```
aeloros, line 17: Interesting approach.  So $(MakeMasixVerbosity) is all or nothing?  There aren't varying levels of verbosity in this target?  I would have expected subsets of different verbosity, rather than everything changing based on the property.

MakeMsix.targets
```diff
@@ -0,0 +1,78 @@
...
+    <Copy SourceFiles="@(MakeMsixInputsWithLocations)"
+          DestinationFiles="@(MakeMsixInputsWithLocations->'$(_MakeMsix_WorkDir)\%(TargetFile)')"
+          SkipUnchangedFiles="true" />
+
+    <Message Text="MakeMSIX: Copy @(MakeMsixInputAssets) to $(_MakeMsix_WorkDir)\Assets" Importance="$(MakeMsixVerbosity)" Condition="'@(MakeMsixInputAssets)' != ''"/>
+    <MakeDir Directories="$(_MakeMsix_WorkDir)\Assets" Condition="'@(MakeMsixInputAssets)' != '' and !Exists('$(_MakeMsix_WorkDir)\Assets')" />
+    <Copy SourceFiles="@(MakeMsixInputAssets)" DestinationFolder="$(_MakeMsix_WorkDir)\Assets" SkipUnchangedFiles="true" Condition="'@(MakeMsixInputAssets)' != ''"/>
+
+    <Message Text="MakeMSIX: Command: $(_MakeMsixCommand)" Importance="$(MakeMsixVerbosity)" />
+    <Exec Command="$(_MakeMsixCommand)" />
```
aeloros, line 59: Consider utilizing the publicly accessible Task instead of Exec.

MakeMsix.targets
```diff
@@ -0,0 +1,78 @@
...
+      <_SignMsix_PWD>$(_SignMsix_UserDir)winappsdk.certificate.test.pwd</_SignMsix_PWD>
+      <_SignMsix_Password>$([System.IO.File]::ReadAllText('$(_SignMsix_PWD)').TrimEnd())</_SignMsix_Password>
+    </PropertyGroup>
+    <Message Text="MakeMSIX: _SignMsix_PWD: '$(_SignMsix_PWD)'" Importance="$(MakeMsixVerbosity)" />
+    <Message Text="MakeMSIX: _SignMsix_Password: ...redacted..." Importance="$(MakeMsixVerbosity)" />
+    <PropertyGroup>
+      <_SignMsixCommand>$(_SignMsix_Executable) sign $(_SignMsix_Parameters) $(_SignMsix_Options) /f "$(_SignMsix_PFX)" /p "$(_SignMsix_Password)" "$(_MakeMsix_OutFilename)"</_SignMsixCommand>
+    </PropertyGroup>
+    <Message Text="MakeMSIX: Sign: $(_SignMsixCommand)" Importance="$(MakeMsixVerbosity)" />
+    <Exec Command="$(_SignMsixCommand)" Condition="'$(MakeMsixSigned)' == 'true'"/>
```
aeloros, line 76: Consider utilizing the publicly accessible Task instead of Exec.

test/TestApps/PushNotificationsTestAppPackage/PushNotificationsTestAppPackage.wapproj
```diff
@@ -42,7 +42,13 @@
     <EntryPointProjectUniqueName>..\PushNotificationsTestApp\PushNotificationsTestApp.vcxproj</EntryPointProjectUniqueName>
     <GenerateAppxPackageOnBuild>true</GenerateAppxPackageOnBuild>
     <GenerateAppInstallerFile>False</GenerateAppInstallerFile>
-    <PackageCertificateKeyFile>$(SolutionDir)temp\MSTest.pfx</PackageCertificateKeyFile>
+    <PackageCertificateKeyFile>$(SolutionDir).user\winappsdk.certificate.test.pfx</PackageCertificateKeyFile>
```
aeloros, line 45: Are these project specific?  Couldn't this be done as a centralized target for the whole tree rather than in each project file?  We already have places for that type of thing.

test/TestApps/AppLifecycleTestPackage/AppLifecycleTestPackage.wapproj
```diff
@@ -58,7 +58,13 @@
     <EntryPointProjectUniqueName>..\AppLifecycleTestApp\AppLifecycleTestApp.vcxproj</EntryPointProjectUniqueName>
     <GenerateAppxPackageOnBuild>true</GenerateAppxPackageOnBuild>
     <GenerateAppInstallerFile>False</GenerateAppInstallerFile>
-    <PackageCertificateKeyFile>$(SolutionDir)temp\MSTest.pfx</PackageCertificateKeyFile>
+    <PackageCertificateKeyFile>$(SolutionDir).user\winappsdk.certificate.test.pfx</PackageCertificateKeyFile>
+<_SignMsix_UserDir>$(SolutionDir).user\</_SignMsix_UserDir>
```
aeloros, line 62: Are these still used here?

This project correctly uses the public Appx targets.  I don't think that utilizes _SignMsix* stuff.  The cert is the only thing that should be changed here, right?

installer/test/CreateInstallerTestPackages/Makefile.mak
```diff
@@ -21,7 +21,7 @@ SIGNTOOL_OPTS=/v
 TARGET_BASENAME=Microsoft.WindowsAppRuntime.Test.InstallFramework
 
 TargetDir=$(OutDir)$(TargetName)
-MSTestPfx=$(SolutionDir)..\..\..\temp\MSTest.pfx
+WinAppSDKTestPfx=$(SolutionDir)..\..\..\.user\winappsdk.certificate.test.pfx
```
aeloros, line 24: Do not use SolutionDir.  Use RepoRoot instead with full paths.

test/AppLifecycle/AppLifecycle.vcxproj
```diff
@@ -298,4 +298,11 @@
   <Target Name="CopyFiles" AfterTargets="AfterBuild">
     <Copy SourceFiles="$(OutDir)\..\WindowsAppRuntime_BootstrapDLL\Microsoft.WindowsAppRuntime.Bootstrap.dll" DestinationFolder="$(OutDir)" />
   </Target>
+  <PropertyGroup>
+<_SignMsix_UserDir>$(SolutionDir).user\</_SignMsix_UserDir>
```
aeloros, line 302: Do not use SolutionDir -- Throughout.

test/Deployment/data/WindowsAppRuntime.Test.Main/WindowsAppRuntime.Test.Main.vcxproj
```diff
@@ -87,95 +87,23 @@
...
-    <NMakeIncludeSearchPath>$(SolutionDir)tools</NMakeIncludeSearchPath>
-  </PropertyGroup>
-  <ItemDefinitionGroup>
-  </ItemDefinitionGroup>
   <ItemGroup>
-    <Xml Include="appxmanifest.xml" />
+    <MakeMsixInputAssets Include="$(MSBuildThisFileDirectory)\Assets\*" />
+    <MakeMsixInputs Include="$(MSBuildThisFileDirectory)appxmanifest.xml" />
   </ItemGroup>
+  <Import Project="$(RepoRoot)\MakeMsix.targets" />
```
aeloros, line 102: I love the use of RepoRoot here.  Should we consider moving the MakeMsix.targets file into the build directory to keep the root tidy?

test/TestApps/PushNotificationsDemoPackage/PushNotificationsDemoPackage.wapproj
```diff
@@ -42,7 +42,13 @@
     <EntryPointProjectUniqueName>..\PushNotificationsDemoApp\PushNotificationsDemoApp.vcxproj</EntryPointProjectUniqueName>
     <GenerateAppxPackageOnBuild>true</GenerateAppxPackageOnBuild>
     <GenerateAppInstallerFile>False</GenerateAppInstallerFile>
-    <PackageCertificateKeyFile>$(SolutionDir)temp\MSTest.pfx</PackageCertificateKeyFile>
+    <PackageCertificateKeyFile>$(SolutionDir).user\winappsdk.certificate.test.pfx</PackageCertificateKeyFile>
```
aeloros, line 45: Please remove the usage of this variable everywhere in your change.  SolutionDir is a terrible property especially in a tree that has more than one solution file (ours is one of those).  Use RepoRoot and full paths whereever possible.

test/PushNotifications/PushNotificationsLongRunningTask.Msix/PushNotificationsLongRunningTask.Msix.vcxproj
```diff
@@ -63,68 +63,37 @@
...
-    <OutDir>$(OutDir)</OutDir>
-    <IntDir>$(IntDir)</IntDir>
-    <NMakeIncludeSearchPath>$(SolutionDir)tools</NMakeIncludeSearchPath>
-  </PropertyGroup>
-  <ItemDefinitionGroup>
-  </ItemDefinitionGroup>
+  <ItemGroup>
+    <MakeMsixInputAssets Include="$(MSBuildThisFileDirectory)\Assets\*" />
+    <MakeMsixInputs Include="$(MSBuildThisFileDirectory)appxmanifest.xml" />
+    <MakeMsixInputs Include="$(OutDir)..\WindowsAppRuntime_DLL\Microsoft.Internal.FrameworkUdk.dll" />
```
aeloros, line 91: Avoid this pattern please every chance you get.

"$(OutDir)..\WindowsAppRuntime_DLL" == "$(BaseOutputPath)\WindowsAppRuntime_DLL"

Relative paths are terrible on build systems.  Not only does it make refactoring hard, but causes hidden bugs in the code base due to standard file names for headers.  When you move files in a refactor you can accidentally build clean, when you are including the wrong files.  It leads to very hard to diagnose problems.  If you can't find a variable that allows avoiding relative directory usage, let me know and I'll gladly help you find the right solution.

aeloros (changes requested):
Lot's of awesome work here.  Let me know once you've updated based on my feedback.

MakeMsix.targets
DrusTheAxe: Sounds like no, intentionally

https://social.msdn.microsoft.com/Forums/en-US/510f07b4-c5f7-43a8-b7cb-e3c398841725/getting-the-verbosity-level-from-a-task?forum=msbuild

>Is there a way for a Task (or, better, using a property) to determine the current verbosity level?
>
>You can't -- we didn't want tasks/build process to assume what loggers would want. Otherwise, eg., I wouldn't be able to walk up to any build process and attach my "super duper log everything and filter it later SQL database logger".

and https://social.msdn.microsoft.com/Forums/vstudio/en-US/0d4d53f9-ced2-4af9-a4b4-4a9694523492/condition-for-output-verbosity-level

>Tasks can't see logger verbosity. Verbosity is purely something the engine passes on to loggers, they can choose what they do with it. The reasoning is that if tasks understood verbosity, I couldn't write a logger that logged everything on low verbosity, for example. Also in future verbosity might be per-logger.
>
>I suggest that you have some other toggle for your exec task, using a property (you could set the property from the cmd line with /p).

and https://stackoverflow.com/questions/4341155/get-the-current-logging-verbosity-level-in-my-msbuild-script

aeloros: None of this is a Task.  Tasks are things like <Exec> which are exposed functionality from a Managed Assembly.  That being said, I'm sure those points are all still valid as they seem to point to a purposeful design here.  

I think this change kind of goes against that design though philosophy though.  This property makes all of the below output set to a certain level.  There are no varying levels of verbosity here for the logger to choose from.  This is because all of the logging is tied to this property.  So maybe the property is redundant, and the levels of each message should be set accordingly.  This allows the logger to actually decide the verbosity.

I only Unresolved this conversation because I don't think it's a done conversation.  I would not consider it blocking for your change or my signoff.  I'm just trying to help make your logging better.  

You should consider having more than one level of verbosity from your Target.  If you want a property similar to this (still beyond the built in one that works via the loggers), then you could have even more output exposed by the Target at the Info logging level.

tools/DevCheck.ps1
```diff
@@ -39,26 +54,41 @@
 #>
 
 Param(
-    [Switch]$CheckAll=$false,
+    [String]$CertPassword=$null,
```
bhardwajs, line 57: Does the `CertPassword` get used by the caller? If not, can it be a random string and then the codebase wouldn't have even a test password stored. Internally, we use a pattern like this for generating certificates to sign some test. I can share that with you over Teams/email.

MakeMsix.targets
DrusTheAxe: I defined the property consistent with other like implementations e.g. Microsoft.Windows.CppWinRT.targets in the CppWinRT package

```
    <PropertyGroup>
        <CppWinRTVerbosity Condition="'$(CppWinRTVerbosity)' == ''">normal</CppWinRTVerbosity>
        <CppWinRTCommandVerbosity Condition="'$(CppWinRTVerbosity)' == 'high'">-verbose</CppWinRTCommandVerbosity>
...
        <Message Text="GetResolvedWinMD: @(WinMDFullPath->'%(FullPath)')" Importance="$(CppWinRTVerbosity)"/>
```

Why's C++/WinRT do it? @kennykerr @BenJKuhn ?

MakeMsix.targets
DrusTheAxe: It's an override - when you create an MSIX package you can optionally also sign it. The default is true for the common case, but MSIX supports unsigned packages. This knob leaves an easy option to support that.

I thought about splitting this into 2 .targets but separating out signtool seemed unnecessary complexity.

MakeMsix.targets
DrusTheAxe: I see MrtCore.PriGen.targets does a whole bunch of local definition., but is that generic or specific to MRTCore? If generic shouldn't that be in a .props file imported as needed?

EnsureOutputLayout.props defines AppxPackageDir = $(BaseOutputPath)AppxPackages e.g. /BuildOutput/Debug/x64/AppxPackages. That's equivalent to what the various test projects were doing (MakeMsix.targets just centralizes that, didn't invent the path).

Q: Why plural AppxPackages?

Q: EnsureOutputLayout.props defines AppxPackageDir=...AppxPackages -- singular=plural. Seems odd. Intentional or accidental?

Q: This is the 'workdir' where we collect files to then run makeappx.exe "make a package from that dir and all its children". Should that be an 'Out' or 'Int' directory? IOW should that be under BuildOutput or obj? Seems like the latter. Isn't this just a temporary interim working directory to create the .msix? That once we have the .msix we don't need this workdir anymore? If so then obj would be the better place.

installer/test/CreateInstallerTestPackages/Makefile.mak
DrusTheAxe: That's going away. I haven't updated installer\** yet, but I didn't want to wait until then before getting feedback. That's just yet-more-of-the-same convert older to MakeMsix.targets.

MakeMsix.targets
kennykerr: @Scottj1s may know.

test/Deployment/data/WindowsAppRuntime.Test.Main/WindowsAppRuntime.Test.Main.vcxproj
DrusTheAxe: RepoRoot is pretty neat. I'd been using SolutionDir but RepoRoot is better (especially as we have multiple solutions, only 1 of them in the root dir). Will make some updates.

We've got a bunch other build files in the root. Yes, we probably should shuffle some things, but we have some matters to iron out before we go there.

`build` historically was used by the _build pipeline_ not the local VS content (i.e. not by WindowsAppRuntime.sln etc). I suspect it's useful to have a clear distinction between those things, though that could be separate dirs e.g. `build\pipeline` vs `build\vs`.

We've had some discussions of file layout and some planned improvements. I think it's good for now and will follow up on the more holistic layout / cleanup.

test/TestApps/PushNotificationsDemoPackage/PushNotificationsDemoPackage.wapproj
DrusTheAxe: Agreed. I only discovered RepoRoot recently. Updates coming...

test/PushNotifications/PushNotificationsLongRunningTask.Msix/PushNotificationsLongRunningTask.Msix.vcxproj
DrusTheAxe: What's the recommended macros? The important ones I know:

* RepoRoot = \
* OutDir = $(RepoRoot)\BuildOutput\$(Configuration)\$(PlatformTarget)\$(MSBuildProjectName)
* IntDir = $(RepoRoot)\$(Configuration)\$(PlatformTarget)\$(MSBuildProjectName)

At least, that's what I thought they are in usual VS projects. Am I mistaken?

Looking at EnsureOutputLayout.props...

* BaseOutputPath = $(RepoRoot)\BuildOutput\$(Configuration)\$(PlatformTarget)\
* OutDir = $(BaseOutputPath)
* OutputPath = $(OutDir)\$(MSBuildProjectName)
* TargetDir == $(OutDir)

Why do we have TargetDir? Is it a VS macro or one of our own invention?

OutDir is the common dir and OutputPath is a project's output dir?

* BaseIntermediateOutputPath = $(RepoRoot)\obj\$(Configuration)\$(PlatformTarget)\$(MSBuildProjectName)
* IntermediateOutputPath = $(BaseIntermediateOutputPath)

Why no IntDir?

Naming is confusing -- OutputBlah vs IntermediateOutputBlah. Took me a while to figure out the latter is a peer of the former and not a child / dependent. Would be better as OutputBlah and IntermediateBlah. Any reason we can't do that?



Also, what's the recommended syntax for one project o reference another project's OutDir? You suggest `$(OutDir)..\foo` is bad. Do what instead? `$(RepoRoot)\BuildOutput\$(Configuration)\$(PlatformTarget)\foo`? Is there a macro for `$(RepoRoot)\BuildOutput\$(Configuration)\$(PlatformTarget)\`?

According to EnsureOutputLayout.props that _is_ OutDir, but per actually building projects OutDir is what EnsureOutputLayout.props calls OutputPath. So I'm puzzled how it winds up in that state, and of course the $64,000 question - what should we change?

tools/DevCheck.ps1
DrusTheAxe: Yes. We need the password saved to disk for later reuse.

*.pfx are password protected. Using them (.e.g to make a signed .msix) requires the password. If by 'a random string' you mean 'a defined constant' that wouldn't be any better than no password. Also, some of this work requires admin privilege -- but normal development doesn't. So we do the work here to create the cert (.cer, .pfx) and save the password for later use.

The build pipeline specifies a value because it has to (after all, there's no human to do 1-time setup). See `build\AzurePipelinesTemplates\WindowsAppSDK-BuildProject-Steps.yml`. But that's done on the build machine during a build, gets stashed in the build environment, used during the build and then torched when the build's over, never copied off that machine.

And that's a constant as an interim thing (crawl, walk, run). I'd like to change it to a random number (for instance, `$CertPassword = New-Guid`) but I have questions about the build pipelines (e.g. build machine reuse) so best handled as a later change. @EricJohnson327 this also poses questions with the build work you're doing. I was planning to bring it up after we cleared some work from our plates.

MakeMsix.targets
DrusTheAxe: > How are MakeMsixInputs and MakeMsixInputsWithLocations different? Is it the same list but one does not have a path component?

Path and/or filename

Most uses copy a file to the package root with the same name. A few exception cases need to copy a file to a subdir, or to a different filename, or both.

I tried adding destination metadata to _some_ input files but couldn't figure out syntax to not choke on that. So I caved and made 2 different properties. A bit ugly but not that bad all things considered, and time's run well over for this. If there's a more elegant solution it can be revised in a future PR.

DrusTheAxe: > Also, why is the output an input?

Because...when it comes to msbuild I'm the one-eyed man (in the kingdom of the blind)? ;-)

Should be

```
  <Target Name="MakeMSIX"
    Inputs="@(MakeMsixInputs);@(MakeMsixInputsWithLocations);@(MakeMsixInputAssets)"
    Outputs="$(MakeMsixOutputFilename)"
    AfterTargets="AfterBuild">
```

?

MakeMsix.targets
DrusTheAxe: Human error?

test/AppLifecycle/AppLifecycle.vcxproj
```diff
@@ -298,4 +298,11 @@
...
+<_SignMsix_PFX>$(_SignMsix_UserDir)winappsdk.certificate.test.pfx</_SignMsix_PFX>
```
aeloros, line 303: Consider doing this only once for the entire tree instead of in each project.

MakeMsix.targets
DrusTheAxe: Updated MakeMsix.targets with specific log levels and removed MakeMsixVerbosity

DrusTheAxe:
@BenJKuhn @Scottj1s @aeloros @huichen123 @sachintaMSFT do we really need 5 different Win10 SDKs to build WinAppSDK? See below.

After resyncing with latest main I see there's a new RestartAgent project but it fails to build for me because

```
32>C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Microsoft\VC\v160\Microsoft.Cpp.WindowsSDK.targets(46,5): error MSB8036: The Windows SDK version 10.0.20348.0 was not found. Install the required version of Windows SDK or change the SDK version in the project property pages or by right-clicking the solution and selecting "Retarget solution".
```

That's due to `dev\RestartAgent\RestartAgent.vcxproj`
```
    <WindowsTargetPlatformVersion Condition=" '$(WindowsTargetPlatformVersion)' == '' ">10.0.20348.0</WindowsTargetPlatformVersion>
```

This gives RestartAgent access to APIs to up thru recent Windows releases but it requires that SDK be installed to build WinAppSDK, which hasn't been required of developers to date.

@aeloros Is that intentional or accidental?

Grep for `WindowsTargetPlatformVersion` across all *proj in the repo and everyone's got

```
    <WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>
```

except....

* `10.0.20348.0` in dev\RestartAgent\RestartAgent.vcxproj -- @aeloros is that intentional?
* `10.0.18362.0` in `dev\MRTCore\mrt\Microsoft.Windows.ApplicationModel.Resources\src\Microsoft.Windows.ApplicationModel.Resources.vcxproj` -- @huichen123 is that intentional?
* `10.0.19041.0` in installer\dev\WindowsAppRuntimeInstall.vcxproj -- @sachintaMSFT is that intentional?

Most projects don't specify `<WindowsTargetPlatformMinVersion>`. The few that do have 10.0.17763.0 in...

* `installer\dev\WindowsAppRuntimeInstall.vcxproj`
* `test\DynamicDependency\data\Framework.Math.Add\Framework.Math.Add.vcxproj`
* `test\DynamicDependency\data\Framework.Math.Multiply\Framework.Math.Multiply.vcxproj`
* `test\DynamicDependency\data\Framework.Widgets\Framework.Widgets.vcxproj`

and 10.0.17134.0 in

* `MRTCore\mrt\Microsoft.Windows.ApplicationModel.Resources\src\Microsoft.Windows.ApplicationModel.Resources.vcxproj`
* `dev\RestartAgent\RestartAgent.vcxproj`

Thanks to the [handy dandy decoder ring](https://en.wikipedia.org/wiki/Windows_10_version_history#Channels)...

* 17134 = RS4
* 17763 = RS5
* 18362 = 19H1
* 19041 = 20H1
* 20348 = recent Win11 build ([22000 = 21H1 Win11](https://en.wikipedia.org/wiki/Windows_11_version_history#Channels))

I've updated my local config to get around this for now. But I'm wondering if we really do need all 5 SDKs (RS4+). We need at least 2 which used to be min=17763, max=18362. Can there be 1 max and if so, should it be 20348?

riverar:
> But I'm wondering if we really do need all 5 SDKs (RS4+). We need at least 2 which used to be min=17763, max=18362. Can there be 1 max and if so, should it be 20348?

I would target the lowest versioned SDK that was released in conjunction with the minimum supported OS. (In other words, SDK=17763, OS=17763.) There is no guarantee that SDKs released after an OS release are compatible down-level.

MakeMsix.targets
DrusTheAxe: @aeloros Here's a weird one. In makemsix.targets add some printfery to the Clean target

```
  <Target Name="CleanMakeMSIX" AfterTargets="AfterClean">
<Message Text="aaa $(OutputPath)" Importance="high" />
<Message Text="bbb $(OutDir)" Importance="high" />
<Message Text="xxx $(MakeMsixOutputFilename)" Importance="high" />
<Message Text="yyy $(TargetName)" Importance="high" />
```

Then in VS do Clean Solution and you get inconsistent output:

```
27>------ Clean started: Project: Framework.Math.Add, Configuration: Debug x64 ------
27>aaa C:\source\repos\windowsappsdk\BuildOutput\Debug\x64\Framework.Math.Add\
27>bbb C:\source\repos\windowsappsdk\BuildOutput\Debug\x64\Framework.Math.Add\
27>xxx Framework.Math.Add.msix
27>yyy Framework.Math.Add
```

vs

```
36>------ Clean started: Project: Microsoft.WindowsAppRuntime.Framework, Configuration: Debug x64 ------
36>aaa C:\source\repos\windowsappsdk\BuildOutput\Debug\x64\
36>bbb C:\source\repos\windowsappsdk\BuildOutput\Debug\x64\Microsoft.WindowsAppRuntime.Framework\
36>xxx Microsoft.WindowsAppRuntime.Framework.msix
36>yyy Microsoft.WindowsAppRuntime.Framework
```

Observations:

* OutputPath=OutDir in 3 projects (Framework.Math.Add, Framework.Math.Multiply, Framework.Widgets)
* OutputPath!=OutDir in the other 10 projects
* OutputPath is the root-of-all-output vs OutDir is a project-specific output
* Per `build\EnsureOutputLayout.props` I'd expect the exact opposite

```
    <BaseOutputPath>$(RepoRoot)\BuildOutput\$(Configuration)\$(PlatformTarget)\</BaseOutputPath>
...
    <OutDir>$(BaseOutputPath)</OutDir>
    <OutputPath>$(OutDir)\$(MSBuildProjectName)\</OutputPath>
```

OutputPath should point to a subdir of OutDir, but I'm seeing the exact opposite. I don't understand why. Do you?

DrusTheAxe:
> > But I'm wondering if we really do need all 5 SDKs (RS4+). We need at least 2 which used to be min=17763, max=18362. Can there be 1 max and if so, should it be 20348?
> 
> I would target the lowest versioned SDK that was released in conjunction with the minimum supported OS. (In other words, SDK=17763, OS=17763.) There is no guarantee that SDKs released after an OS release are compatible down-level, nor is there typically a compelling reason to upgrade that SDK.

One reason is if you need to compile light-up functionality. You need the newer SDK for the newer headers/libs/winmd/etc to build, even though you implement to only conditionally use the functionality if available. We support RS5-to-latest so SDK=OS=17763=RS5 as the minimum sounds good (couple of violators pointing to RS4 should be updated?). What's unclear is (1) if we should always set the ceiling or only those projects that need it, and (2) when we set a ceiling why not always set it to the same value (e.g. 20348)?

.gitignore
```diff
@@ -381,6 +381,9 @@ installer/test/testpackages/
 # Local temp storage
 temp/
 
+# Local per-user storage
+.user/
```
BenJKuhn, line 385: (FYI for @EricJohnson327... I think you talked about having a similar need for package caching, etc.)

Directory.Build.props
```diff
@@ -47,4 +47,16 @@
 
   <!-- Build with the hybrid CRT (Universal CRT + Static VS CRT (for what little the Universal CRT doesn't cover) -->
   <Import Project="$(MSBuildThisFileDirectory)HybridCRT.props" />
+
```
aeloros, line 50: Should we have a <Error Text="Missing cert stuff, please run $(RepoRoot)\tools\DevCheck.cmd!" Condition="!Exists('$(RepoRoot)\.user')" /> or something like that here?

aeloros:
accidental


Sent from Outlook<http://aka.ms/weboutlook>

aeloros (approved):

DrusTheAxe:
I see the problem but I don't understand why (yet).

The build pipeline successfully ran DevCheck creating the test cert (.pfx etc under .user), then build WindowsAppRuntime.sln just fine. Then it built MRTCore and failed. That doesn't run DevCheck but relies on it having already been run - which it was. But it can't find `.user\winappsdk.certificate.test.pfx` despite looking for an absolute filename exactly the same as DevCheck created and WindowsAppRuntime.sln found just fine.

I suspect something's going on in the pipeline I don't understand. Will talk with ES folks in the morning

aeloros:
Maybe it's not including RepoRoot somehow?


Sent from Outlook<http://aka.ms/weboutlook>

Directory.Build.props
```diff
@@ -47,4 +50,16 @@
...
+  <!-- Load the test certificate's password (so we do it once for reuse across projects) -->
+  <PropertyGroup Condition="'$(RepoTestCertificatePFX)' == ''">
+    <RepoTestCertificatePFX>$(RepoRoot)\.user\winappsdk.certificate.test.pfx</RepoTestCertificatePFX>
```
kythant, line 56: I believe the dev\MRTCore\mrt\Core\src\MRM.vcxproj is pulling this property file and because this project is being built in a different job. It started failing.

kythant: DevCheck.ps1 will need to run in build\build-mrt.yml to fix this. build-mrt.yml is run in a different job than build\AzurePipelinesTemplates\WindowsAppSDK-BuildProject-Steps.yml
or
or this property be moved so it's not in directory.build.props

PR #1989, opened 2022-01-14
Add MRTCore interface name and id

huichen123:
Provide explicit interface name and id to ResourceLoader/ResourceManager/ResourceContext. They are the same as auto generated by MIDL so there is no change to the user.


DrusTheAxe:
> Provide explicit interface name and id to ResourceLoader/ResourceManager/ResourceContext

Why?

axelandrejs (approved):

jonwis:
I wouldn't have added the names. MIDL3 automatically generates the IID and interface name based on its shape -> ID algorithm and typename -> interface_name algorithms.

What motivated the change here?

PR #1995, opened 2022-01-18
MddBootstrapInitialize package enumeration if release Minor version >0

DrusTheAxe:
DDLM package names are `Microsoft.WinAppRuntime.DDLM.<minor><patch>.<msixminor>.<msixbuild>.<msixrevision>[-shortversiontag]`. But we only want to match down to `<minor>`. The code+spec(comment) was wrong to look for `<minor>.*` needs to be `<minor>*`

For want of a nail...

### DIFF
--- a/dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
+++ b/dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
@@ -500,7 +500,7 @@ void FindDDLMViaEnumeration(
     // We need to look for DDLM packages in the package family for release <major>.<minor> and <versiontag>
     // But we have no single (simple) enumeration to match that so our logic's more involved compared
     // to FindDDLMViaAppExtension():
-    // 1. Look for Framework packages with Name="microsoft.winappruntime.ddlm.<minorversion>.*[-shorttag]"
+    // 1. Look for Framework packages with Name="microsoft.winappruntime.ddlm.<minorversion>*[-shorttag]"
     // 1a. Enumerate all Framework packages registered to the user
     // 1b. Only consider packages whose Name starts with "microsoft.winappruntime.ddlm.<minorversion>."
     // 1c. If versiontag is specified, Only consider packages whose Name ends with [-shorttag]
@@ -519,7 +519,7 @@ void FindDDLMViaEnumeration(
     }
     else
     {
-        wsprintf(packageNamePrefix, L"microsoft.winappruntime.ddlm.%hu.", minorVersion);
+        wsprintf(packageNamePrefix, L"microsoft.winappruntime.ddlm.%hu", minorVersion);
     }
     const auto packageNamePrefixLength{ wcslen(packageNamePrefix) };
 


dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -500,7 +500,7 @@ void FindDDLMViaEnumeration(
     // We need to look for DDLM packages in the package family for release <major>.<minor> and <versiontag>
     // But we have no single (simple) enumeration to match that so our logic's more involved compared
     // to FindDDLMViaAppExtension():
-    // 1. Look for Framework packages with Name="microsoft.winappruntime.ddlm.<minorversion>.*[-shorttag]"
+    // 1. Look for Framework packages with Name="microsoft.winappruntime.ddlm.<minorversion>*[-shorttag]"
```
riverar, line 503: Shouldn't this be Major?

Example: `Microsoft.WinAppRuntime.DDLM.0.319.455.0-x8_0.319.455.0_x86__8wekyb3d8bbwe`
Major: 0
Minor: 319
Build: 455
Revision: 0

```suggestion
    // 1. Look for Framework packages with Name="microsoft.winappruntime.ddlm.<majorversion>*[-shorttag]"
```

riverar:
> DDLM package names are `Microsoft.WinAppRuntime.DDLM.<minor><patch>.<msixminor>.<msixbuild>.<msixrevision>[-shortversiontag]`. But we only want to match down to `<minor>`. The code+spec(comment) was wrong to look for `<minor>.*` needs to be `<minor>*`

Doesn't sound right. Where is the major version in the package name considered?

DrusTheAxe:
> > DDLM package names are `Microsoft.WinAppRuntime.DDLM.<minor><patch>.<msixminor>.<msixbuild>.<msixrevision>[-shortversiontag]`. But we only want to match down to `<minor>`. The code+spec(comment) was wrong to look for `<minor>.*` needs to be `<minor>*`
> 
> Doesn't sound right. Where is the major version in the package name considered?

There is no release major version in the DDLM's package Name.

DDLM [naming's weird](https://github.com/microsoft/WindowsAppSDK/blob/main/specs/Deployment/MSIXPackages.md#33-package-naming---ddlm)

* Name = Microsoft.WinAppRuntime.DDLM.<Version>-<ShortArchitecture>[-ShortVersionTag]

For a real world example, `Microsoft.WinAppRuntime.DDLM.1000.378.2151.0-x6` for a 1.1 Stable build.

That `<Version>` is the MSIX package version, and its `major` field is computed from the release's minor + patch. Thus for Stable release 1.2.3 the DDLM would be `Microsoft.WinAppRuntime.DDLM.2003.<epoch>.<build>.0`.

Odd, but it works because there's only ever 1 DDLM package in a package family (each a unique special snowflake). The epoch+build are date+time based so we never produce 2 builds with identical values so 1.2 + 2.2 + 3.2... will never collide.

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
DrusTheAxe: Nope. ByDesign. DDLM naming's weird. See other comment

riverar:
Ah right. Thanks for the reminder.

Nit: Maybe a tweak to the comment/variables would help? `<computed version>` perhaps? I think it's important for maintainability/code review that there is no confusion with other versioning schemes (e.g. semver, assemblyversion, etc). A link to the versioning spec would help too.

DrusTheAxe:
> Ah right. Thanks for the reminder.
> 
> Nit: Maybe a tweak to the comment/variables would help? `<computed version>` perhaps over `major` or `minor`? I think it's important for maintainability/code review that there is no confusion with other versioning schemes (e.g. semver, assemblyversion, etc). A link to the versioning spec would help too.

Yes it can get confusing given 'release version' and 'MSIX version'. Code/specs usually refer to <rmajor>, <rminor>, <rpatch> for release version and msixmajor/minor/build/revision meaning MSIX, if there's any confusion. Most contexts refer to one or the other so it's usually clear (or as clear as it gets).

I'll see about tweaking bits for more clarity when I'm next in there.

Thanks for the feedback

MikeHillberg (approved):

PR #2005, opened 2022-01-24
Feature for sharing security descriptors (take 2)

DefaultRyan:
This resurrects #209 , rather than attempt to re-merge an extremely old branch when so much has changed, I've created a new PR here.

The two major requests on the old PR, which are both addressed here, were to get the WinRT API coded up, and to make sure we hand out a [self-relative security descriptor](https://docs.microsoft.com/en-us/windows/win32/secauthz/absolute-and-self-relative-security-descriptors).

After coding up the tests for the WinRT API, I'm having some doubts about the WinRT version taking an array of name/mask structs. This pattern makes it much harder to avoid creating hstrings on the heap, and I wouldn't have thought it commonplace for callers to specify multiple app container names at a time. I'll probably change the WinRT API to take a single app container name and access mask, so as not to defeat helpers like C++/WinRT's `param::hstring` helper. I'd like to get some thoughts from people like @jonwis , @BenJKuhn , and @mikebattista, and @DrusTheAxe. Also, any thoughts on if this should be marked preview/experimental (with velocity)?


dev/AccessControl/GetSecurityDescriptorForAppContainerNames.cpp
```diff
@@ -0,0 +1,100 @@
...
+#include "SecurityDescriptorHelpers.h"
+#include <vector>
+
+using namespace std;
+
+STDAPI GetSecurityDescriptorForAppContainerNames(
+    uint32_t accessRequestCount,
+    _In_reads_(accessRequestCount)
+    const AppContainerNameAndAccess* accessRequests,
+    _In_opt_ PSID principal,
```
BenJKuhn, line 17: Should this also be const?

dev/AccessControl/GetSecurityDescriptorForAppContainerNames.cpp
```diff
@@ -0,0 +1,100 @@
...
+    _Outptr_ PSECURITY_DESCRIPTOR* securityDescriptor,
+    _Out_opt_ DWORD* securityDescriptorLength
+)
+{
+    *securityDescriptor = nullptr;
+    if (securityDescriptorLength)
+    {
+        *securityDescriptorLength = 0;
+    }
+    try
```
BenJKuhn, line 28: Consider moving to function scoped try catch (general goodness, no specific issue with code as written)

dev/AccessControl/Microsoft.Windows.Security.AccessControl.cpp
```diff
@@ -0,0 +1,50 @@
+﻿#include <pch.h>
+#include "Microsoft.Windows.Security.AccessControl.h"
+#include "Microsoft.Windows.Security.AccessControl.SecurityDescriptorHelpers.g.cpp"
+#include "Security.AccessControl.h"
+#include <sddl.h>
+
+namespace winrt::Microsoft::Windows::Security::AccessControl::implementation
+{
+    wil::unique_hlocal_security_descriptor GetSecurityDescriptor(array_view<winrt::Microsoft::Windows::Security::AccessControl::AppContainerNameAndAccess const> accessRequests, hstring const& principalStringSid, uint32_t principalAccessMask, DWORD* sdLength)
```
BenJKuhn, line 9: please wrap the super-long lines / declarations.

BenJKuhn (approved):
@DefaultRyan, a few minor nits, but mostly looks good. Please address before submitted. As to the question about an overload for getting for one app, I don't see the current code as problematic, but it would make things easier if that's a common usage. I'll defer to @jonwis on that one.

test/AccessControlTests/APITests.cpp
```diff
@@ -0,0 +1,171 @@
...
+            THROW_IF_FAILED(::GetSecurityDescriptorForAppContainerNames(1, ac, nullptr, EVENT_MODIFY_STATE | SYNCHRONIZE, &sd, nullptr));
+            WaitForEvent(sd.get());
+        }
+
+        TEST_METHOD(WinRTStringTest)
+        {
+            RunTestApp();
+            winrt::AppContainerNameAndAccess ac[1] = { {GetTestPackageFile(), EVENT_MODIFY_STATE | SYNCHRONIZE} };
+
+            auto sdString = winrt::SecurityDescriptorHelpers::GetSddlForAppContainerNames(ac, L"", EVENT_MODIFY_STATE | SYNCHRONIZE);
```
DefaultRyan, line 153: This is an instance where this call is going to generate an extra allocation for the hstring in the `ac` struct.

dev/AccessControl/GetSecurityDescriptorForAppContainerNames.cpp
jonwis: Sadly the way `PSID` is defined is as `typedef PVOID PSID`, and we do not have an equivalent `typedef PCVOID PCSID` in the platform SDK (... yet) ... so if you wanted to add a conditional typedef for `PCSID` that's probably fine.

(The `SID` struct type becomes `PISID`, but you have to reinterpret_cast from PSID to SID*.)

dev/AccessControl/GetSecurityDescriptorForAppContainerNames.cpp
```diff
@@ -0,0 +1,100 @@
...
+#include <userenv.h>
+#include "Security.AccessControl.h"
+#include "SecurityDescriptorHelpers.h"
+#include <vector>
+
+using namespace std;
+
+STDAPI GetSecurityDescriptorForAppContainerNames(
+    uint32_t accessRequestCount,
+    _In_reads_(accessRequestCount)
```
jonwis, line 15: Consider folding 15+16 into a single line.

dev/AccessControl/GetSecurityDescriptorForAppContainerNames.cpp
```diff
@@ -0,0 +1,100 @@
...
+        wil::unique_any<PACL, decltype(&::LocalFree), ::LocalFree> acl;
+        THROW_IF_WIN32_ERROR(::SetEntriesInAclW(static_cast<uint32_t>(ea.size()), ea.data(), nullptr, &acl));
+
+        auto absoluteSD = wil::make_unique_hlocal<SECURITY_DESCRIPTOR>();
+        THROW_IF_WIN32_BOOL_FALSE(::InitializeSecurityDescriptor(absoluteSD.get(), SECURITY_DESCRIPTOR_REVISION));
+        THROW_IF_WIN32_BOOL_FALSE(::SetSecurityDescriptorDacl(absoluteSD.get(), TRUE, acl.get(), FALSE));
+
+        DWORD sdLength = 0;
+
+        if (!::MakeSelfRelativeSD(absoluteSD.get(), nullptr, &sdLength))
```
jonwis, line 81: This little blobbet sounds like a good addition to wil's token_helpers.h if you're eventually inclined.

dev/AccessControl/Microsoft.Windows.Security.AccessControl.cpp
```diff
@@ -0,0 +1,50 @@
...
+        if (!principalStringSid.empty())
+        {
+            winrt::check_hresult(::ConvertStringSidToSidW(principalStringSid.data(), &sid));
+        }
+
+        std::vector<::AppContainerNameAndAccess> rawAccessRequests;
+        rawAccessRequests.reserve(accessRequests.size());
+        for (auto&& request : accessRequests)
+        {
+            rawAccessRequests.push_back({ request.appContainerName.data(), request.accessMask });
```
jonwis, line 21: Nit: `emplace_back` ?

dev/AccessControl/SecurityDescriptorHelpers.h
```diff
@@ -0,0 +1,43 @@
...
+#include <wil/result.h>
+#include <processthreadsapi.h>
+
+
+namespace Security::Descriptors
+{
+    wil::unique_any_psid GetCurrentProcessSid()
+    {
+        wil::unique_any_psid result;
+        wil::unique_process_heap_ptr<TOKEN_GROUPS> tg;
```
jonwis, line 14: Looks like `TOKEN_GROUPS` is missing from https://github.com/microsoft/wil/blob/f9284c19c9873664978b873b8858d7dfacc6af1e/include/wil/token_helpers.h#L39 - consider adding it there and using it when a new wil update is available.  But there is `TOKEN_GROUPS_AND_PRIVILEGES`, maybe that helps?

jonwis (approved):

DrusTheAxe:
> I'm having some doubts about the WinRT version taking an array of name/mask structs... I'll probably change the WinRT API to take a single app container name and access mask

If so please update the [API spec](https://github.com/microsoft/WindowsAppSDK/blob/main/specs/GetSecurityDescriptorForAppContainerNames/GetSecurityDescriptorForAppContainerNames.md) and code for review. If not, please remove this from the comment as it won't reflect the comment.

>any thoughts on if this should be marked preview/experimental (with velocity)?

Is this feature committed and expected in 1.1? If so then you don't need to tag anything.

A release cycle has Experimental release(s) (i.e. whatever's in main) and towards the latter part of the cycle sees Preview release(s) and then ultimately the Stable release. Experimental marking and Velocity'izing implementation is only relevant if a feature is experimental when we have Preview and Stable releases, as that marking+Velocity is how we 'strip' the experimental bits from a release.

If you're not committed for 1.1 then yes, you need to mark+Velocity.

If you're unsure you could wait a bit until Preview|Stable gets closer and worst case mark+Velocity for that. It's easy enough to add if/when needed. OTOH you could mark+Velocity now, and later when Preview|Stable rolls around if you're not experimental  you could remove the mark+Velocity bits. Deferred-optimization vs proactive-implementation, whichever fits best fits your case :-)

dev/AccessControl/GetSecurityDescriptorForAppContainerNames.cpp
```diff
@@ -0,0 +1,100 @@
+﻿#include <pch.h>
+#include <wil/resource.h>
+#include <wil/result.h>
```
DrusTheAxe, line 3: CONSIDER: moving WIL after the OS headers

General rule of thumb, you want to include headers from most generic to most specific e.g.

```
#include <windows.h>
#include <windows headers like aclapi.h>
#include <C/C++ language headers like vector>
#include <other libraries like WIL, WRL, etc>
#include "feature headers like Security.AccessControl.h"
```

dev/AccessControl/GetSecurityDescriptorForAppContainerNames.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+        auto absoluteSD = wil::make_unique_hlocal<SECURITY_DESCRIPTOR>();
+        THROW_IF_WIN32_BOOL_FALSE(::InitializeSecurityDescriptor(absoluteSD.get(), SECURITY_DESCRIPTOR_REVISION));
+        THROW_IF_WIN32_BOOL_FALSE(::SetSecurityDescriptorDacl(absoluteSD.get(), TRUE, acl.get(), FALSE));
+
+        DWORD sdLength = 0;
+
+        if (!::MakeSelfRelativeSD(absoluteSD.get(), nullptr, &sdLength))
+        {
+            auto err = ::GetLastError();
```
DrusTheAxe, line 83: NIT: `lastError`

Common convention for variable names:

* lastError = GetLastError()
* hr = HRESULT

dev/AccessControl/Microsoft.Windows.Security.AccessControl.cpp
```diff
@@ -0,0 +1,50 @@
+﻿#include <pch.h>
+#include "Microsoft.Windows.Security.AccessControl.h"
+#include "Microsoft.Windows.Security.AccessControl.SecurityDescriptorHelpers.g.cpp"
+#include "Security.AccessControl.h"
+#include <sddl.h>
```
DrusTheAxe, line 5: NIT: Move to line 1.5

Unless it's already included in pch.h (somewhat common functionality)?

dev/AccessControl/Microsoft.Windows.Security.AccessControl.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+namespace winrt::Microsoft::Windows::Security::AccessControl::implementation
+{
+    wil::unique_hlocal_security_descriptor GetSecurityDescriptor(array_view<winrt::Microsoft::Windows::Security::AccessControl::AppContainerNameAndAccess const> accessRequests, hstring const& principalStringSid, uint32_t principalAccessMask, DWORD* sdLength)
+    {
+        wil::unique_any_psid sid;
+        if (!principalStringSid.empty())
+        {
+            winrt::check_hresult(::ConvertStringSidToSidW(principalStringSid.data(), &sid));
```
DrusTheAxe, line 14: winrt::check_hresult() or THROW_IF_FAILED() ?

DrusTheAxe, line 14: THROW_IF_WIN32_BOOL_FALSE()

ConvertStringSidToSidW returns BOOL/LastError. Not HRESULT

https://docs.microsoft.com/en-us/windows/win32/api/sddl/nf-sddl-convertstringsidtosidw

dev/AccessControl/Microsoft.Windows.Security.AccessControl.cpp
```diff
@@ -0,0 +1,50 @@
...
+        return sd;
+    }
+
+    hstring SecurityDescriptorHelpers::GetSddlForAppContainerNames(array_view<winrt::Microsoft::Windows::Security::AccessControl::AppContainerNameAndAccess const> accessRequests, hstring const& principalStringSid, uint32_t principalAccessMask)
+    {
+        auto sd = GetSecurityDescriptor(accessRequests, principalStringSid, principalAccessMask, nullptr);
+
+        wil::unique_hlocal_string resultStr;
+        DWORD resultLen = 0;
+        winrt::check_bool(
```
DrusTheAxe, line 37: winrt::check_bool() or THROW_IF_WIN32_BOOL_FALSE() ?

dev/AccessControl/Microsoft.Windows.Security.AccessControl.cpp
```diff
@@ -0,0 +1,50 @@
...
+            ::ConvertSecurityDescriptorToStringSecurityDescriptorW(
+                sd.get(), SDDL_REVISION_1, OWNER_SECURITY_INFORMATION | GROUP_SECURITY_INFORMATION | DACL_SECURITY_INFORMATION | SACL_SECURITY_INFORMATION, &resultStr, &resultLen));
+        return { resultStr.get(), resultLen };
+    }
+
+    com_array<uint8_t> SecurityDescriptorHelpers::GetSecurityDescriptorBytesFromAppContainerNames(array_view<winrt::Microsoft::Windows::Security::AccessControl::AppContainerNameAndAccess const> accessRequests, hstring const& principalStringSid, uint32_t principalAccessMask)
+    {
+        DWORD sdLength = 0;
```
DrusTheAxe, line 45: `DWORD sdLength{};`

dev/AccessControl/Microsoft.Windows.Security.AccessControl.h
```diff
@@ -0,0 +1,19 @@
+﻿#pragma once
```
DrusTheAxe, line 1: Missing Copyright comment line

dev/AccessControl/Security.AccessControl.h
```diff
@@ -0,0 +1,19 @@
+﻿#pragma once
+
+#include <Windows.h>
+
+struct AppContainerNameAndAccess
+{
+    PCWSTR appContainerName;
```
DrusTheAxe, line 7: PCWSTR or PWSTR?

Is the const-ness in the attribute a problem? I've seen compile problems from this sort of thing in the past

dev/AccessControl/Security.AccessControl.h
```diff
@@ -0,0 +1,19 @@
...
+    UINT32 accessMask;
+};
+
+STDAPI GetSecurityDescriptorForAppContainerNames(
+    UINT32 accessRequestCount,
```
DrusTheAxe, line 12: UINT32 or uint32_t?

dev/AccessControl/Security.AccessControl.h
```diff
@@ -0,0 +1,19 @@
...
+    _In_reads_(accessRequestCount)
+    const AppContainerNameAndAccess* accessRequests,
```
DrusTheAxe, line 14: CONSIDER making line 13+14 one line

test/AccessControlTests/APITests.cpp
```diff
@@ -0,0 +1,171 @@
...
+            TEST_CLASS_PROPERTY(L"RunAs:Class", L"RestrictedUser")
+        END_TEST_CLASS()
+
+        TEST_CLASS_SETUP(ClassInit)
+        {
+            try
+            {
+                ::Test::Packages::RemovePackage(GetTestPackageFamilyName());
+                ::Test::Bootstrap::Setup();
+                ::Test::Packages::WapProj::AddPackage(TAEF::GetDeploymentDir(), GetTestPackageFile(), L".msix"); // Installs PushNotificationsTestApp.msix
```
DrusTheAxe, line 70: Do you need it?

test/AccessControlTests/AccessControlTests.vcxproj.filters
```diff
@@ -0,0 +1,40 @@
...
+  </ItemGroup>
+  <ItemGroup>
+    <None Include="packages.config" />
+  </ItemGroup>
+  <ItemGroup>
+    <CopyFileToFolders Include="$(TestPkgOutputPath)\Dependencies\$(PlatformTarget)\Microsoft.VCLibs.$(PlatformTarget)$(VCLibsDebugConfigName).14.00.appx" />
+    <CopyFileToFolders Include="$(TargetDir)..\AccessControlTestApp\AccessControlTestApp.exe" />
+    <CopyFileToFolders Include="$(TestPkgOutputFile)" />
+    <CopyFileToFolders Include="$(TestPkgOutputFile)" />
+    <CopyFileToFolders Include="$(TestPkgOutputFile)" />
```
DrusTheAxe, line 38: TestPkgOutputFile is faerie so you have to say its name 3 times to summon it? :P

dev/AccessControl/GetSecurityDescriptorForAppContainerNames.cpp
```diff
@@ -0,0 +1,100 @@
...
+        *securityDescriptorLength = 0;
+    }
+    try
+    {
+        vector<EXPLICIT_ACCESS_W> ea;
+        if (accessRequestCount == UINT32_MAX)
+        {
+            THROW_HR(E_OUTOFMEMORY);
+        }
+        ea.reserve(static_cast<size_t>(accessRequestCount) + 1);
```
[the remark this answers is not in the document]
DefaultRyan: Leftover code trying to be resilient to overflow. If `accessRequestCount == UINT32_MAX`, then adding one would overflow before the promotion to 64-bit `size_`t.

Ended up simply throwing out of the function to avoid this case, but this little nugget was left behind.

dev/AccessControl/GetSecurityDescriptorForAppContainerNames.cpp
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        {
+            EXPLICIT_ACCESS_W entry{};
+            entry.grfAccessMode = GRANT_ACCESS;
+            entry.grfAccessPermissions = principalAccessMask;
+            entry.grfInheritance = NO_INHERITANCE;
+            entry.Trustee.TrusteeForm = TRUSTEE_IS_SID;
+            entry.Trustee.TrusteeType = TRUSTEE_IS_USER;
+            entry.Trustee.ptstrName = static_cast<PWSTR>(principal);
```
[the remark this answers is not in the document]
DefaultRyan: `EXPLICIT_ACCESS_W` is a weird struct.

The [docs ](https://docs.microsoft.com/en-us/windows/win32/api/accctrl/ns-accctrl-trustee_w)say `ptstrName` is part of a union that also has a `SID*` field (for the `TRUSTEE_IS_SID` option), but they're mostly lying. A reading of `AccCtrl.h` shows that union is only in effect for MIDL, and only the `ptstrName` field exists. So we're stuck with casting to assign this field. :(

dev/AccessControl/Security.AccessControl.h
DefaultRyan: Definitely `PCWSTR`. You can change the value of the pointer, but not what it points to. The const-ness is important if you ever want to assign a const string or literal to that field, aka `L"AccessControlTestAppPackage_8wekyb3d8bbwe"`.

test/AccessControlTests/APITests.cpp
DefaultRyan: Is this about the out-of-date comment? I've updated it.

dev/AccessControl/GetSecurityDescriptorForAppContainerNames.cpp
```diff
@@ -0,0 +1,100 @@
...
+using namespace std;
+
+STDAPI GetSecurityDescriptorForAppContainerNames(
+    uint32_t accessRequestCount,
+    _In_reads_(accessRequestCount)
+    const AppContainerNameAndAccess* accessRequests,
+    _In_opt_ PSID principal,
+    uint32_t principalAccessMask,
+    _Outptr_ PSECURITY_DESCRIPTOR* securityDescriptor,
+    _Out_opt_ DWORD* securityDescriptorLength
```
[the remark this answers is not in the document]
DefaultRyan: Good call. I'll update it to use uint32 everywhere.

dev/AccessControl/Microsoft.Windows.Security.AccessControl.cpp
DefaultRyan: It's a plain old struct without a constructor, using aggregate initialization. So, it doesn't work with `emplace_back`. https://stackoverflow.com/questions/8782895/why-doesnt-emplace-back-use-uniform-initialization

dev/AccessControl/SecurityDescriptorHelpers.h
DefaultRyan: The helper for `TOKEN_GROUPS_AND_PRIVELEGES` doesn't work in this case. But I'll add a helper to WIL.

PR #2007, opened 2022-01-24
ShowToast implementation + unit tests

danielayala94:
This PR introduces ToastNotificationManager::ShowToast() implementation.

As part of this work, I implemented the NotificationProperties and NotificationTransientProperties ABI types to translate the Toast SDK class into objects that the UDK can understand.

To verify the API, I added a couple of unit tests, for both packaged and unpackaged scenarios. Also, I locally exercised the API using the PushNotificationsDemoApp. Ideally, we should add sample code once ToastNotificationsDemoApp is up.


dev/ToastNotifications/NotificationProperties.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+    *payloadSize = static_cast<unsigned int>(m_payload.size());
+    return S_OK;
+}
+
+STDMETHODIMP_(HRESULT __stdcall) NotificationProperties::get_Tag(_Out_ HSTRING* tag) noexcept
+{
+    auto lock = m_lock.lock_shared();
+
+    *tag = safe_make_unique_hstring(m_tag.c_str()).release();
```
danielayala94, line 85: Let's do noexcept try catch

hulumane: in general all the APIs right?

danielayala94: Only the ones that can throw, anything else can remain as noexcept.

dev/ToastNotifications/ToastNotificationManager.cpp
```diff
@@ -102,9 +107,21 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
...
-        throw hresult_not_implemented();
+        THROW_HR_IF(WPN_E_NOTIFICATION_POSTED, toast.ToastId() != 0);
+
+        winrt::com_ptr<::ABI::Microsoft::Internal::ToastNotifications::INotificationProperties> notificationProperties
+            = winrt::make_self<NotificationProperties>(toast);
+
+        winrt::com_ptr<::ABI::Microsoft::Internal::ToastNotifications::INotificationTransientProperties> notificationTransientProperties
+            = winrt::make_self<NotificationTransientProperties>(toast);
+
+        auto toastAppId{ RetrieveUnpackagedAppId() };
```
danielayala94, line 120: Packaged apps won't be happy with this.

test/TestApps/ToastNotificationsTestApp/main.cpp
```diff
@@ -378,6 +378,37 @@ bool VerifyToastExpiresOnReboot()
     return true;
 }
 
+bool VerifyShowToast()
+{
+    // Registration happens in main()
+    winrt::ToastNotification toast{ GetToastNotification() };
+
+    auto toastNotificationManager = winrt::ToastNotificationManager::Default();
+    toastNotificationManager.ShowToast(toast);
```
danielayala94, line 387: I think we can verify toast ID as well.

dev/ToastNotifications/NotificationProperties.cpp
```diff
@@ -0,0 +1,148 @@
...
+namespace Helpers
+{
+    using namespace winrt::Microsoft::Windows::PushNotifications::Helpers;
+}
+
+NotificationProperties::NotificationProperties(winrt::ToastNotification const& toastNotification)
+{
+    // Extract payload and convert it from XML to a byte array
+    auto payload = toastNotification.Payload();
+    auto payloadAsHstring = payload.GetXml();
```
hulumane, line 30: No need to declare a local. To get the equivalent raw representation, use wil::str_raw_ptr when needed.

dev/ToastNotifications/NotificationProperties.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+    std::wstring payloadAsStdWstring{ payloadAsHstring };
```
hulumane, line 32: No need to allocate more memory here. Just remove the local and use wil::str_raw_ptr below

dev/ToastNotifications/NotificationProperties.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+    m_notificationId = toastNotification.ToastId();
+
+    m_tag = toastNotification.Tag();
+    m_group = toastNotification.Group();
+
+    auto expiry = toastNotification.ExpirationTime();
+    m_expiry = winrt::clock::to_file_time(expiry);
+
+    auto arrivalTime = winrt::clock::now();
```
hulumane, line 47: local declaration not needed

dev/ToastNotifications/NotificationProperties.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+    m_payload = wil::unique_cotaskmem_array_ptr<byte>(static_cast<byte*>(CoTaskMemAlloc(payloadAsSimpleString.size())), payloadAsSimpleString.size());
+    CopyMemory(m_payload.data(), payloadAsSimpleString.c_str(), payloadAsSimpleString.size());
+
+    m_notificationId = toastNotification.ToastId();
+
+    m_tag = toastNotification.Tag();
+    m_group = toastNotification.Group();
+
+    auto expiry = toastNotification.ExpirationTime();
```
hulumane, line 44: local declaration not needed

dev/ToastNotifications/NotificationProperties.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+    *payloadSize = static_cast<unsigned int>(m_payload.size());
+    return S_OK;
+}
+
+STDMETHODIMP_(HRESULT __stdcall) NotificationProperties::get_Tag(_Out_ HSTRING* tag) noexcept try
+{
+    auto lock = m_lock.lock_shared();
+
+    *tag = safe_make_unique_hstring(m_tag.c_str()).release();
```
hulumane, line 91: Does safe_make_unique_string throw? If not, how would the caller know that we failed to allocate memory in this implementation?

dev/ToastNotifications/NotificationProperties.cpp
```diff
@@ -0,0 +1,148 @@
...
+{
+    auto lock = m_lock.lock_shared();
+
+    *arrivalTime = static_cast<ULONGLONG>(m_arrivalTime.dwHighDateTime) << 32 | static_cast<ULONGLONG>(m_arrivalTime.dwLowDateTime);
+    return S_OK;
+}
+
+STDMETHODIMP_(HRESULT __stdcall) NotificationProperties::get_BootId(_Out_ unsigned long long* bootId) noexcept
+{
+    *bootId = 0;
```
hulumane, line 123: Please declare a variable called m_bootId. Set that in the ctor. It's equal to 0 still but thats okay.

dev/ToastNotifications/NotificationProperties.cpp
```diff
@@ -0,0 +1,148 @@
...
+    // TODO: Uncomment once ToastProgressData is implemented
+    // auto lock = m_lock.lock_shared();
+    // *toastProgressData = m_toastProgressData.get();
+    *progressData = nullptr;
+    return S_OK;
+}
+
+STDMETHODIMP_(HRESULT __stdcall) NotificationProperties::get_ActivityId(_Out_ GUID* activityId) noexcept
+{
+    *activityId = GUID_NULL;
```
hulumane, line 146: Same feedback here. Declare m_activityId. Set that in the ctor. Its ok if it is GUID_NULL when being set.

dev/ToastNotifications/NotificationTransientProperties.cpp
```diff
@@ -0,0 +1,86 @@
...
+namespace ToastABI
+{
+    using namespace ::ABI::Microsoft::Internal::ToastNotifications;
+}
+
+NotificationTransientProperties::NotificationTransientProperties(winrt::ToastNotification const& toastNotification) noexcept
+{
+    m_suppressPopup = toastNotification.SuppressDisplay();
+
+    auto priority = toastNotification.Priority();
```
hulumane, line 25: No need for local declaration

dev/ToastNotifications/ToastNotificationManager.cpp
```diff
@@ -10,6 +10,11 @@
 #include "ToastNotificationUtility.h"
 #include <frameworkudk/pushnotifications.h>
 #include <frameworkudk/toastnotifications.h>
+#include <FrameworkUdk/toastnotificationsrt.h>
+
```
hulumane, line 14: Remove blank line please.

dev/ToastNotifications/ToastNotificationManager.cpp
```diff
@@ -102,9 +107,21 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
     {
         GetToastHandlers().remove(token);
     }
-    void ToastNotificationManager::ShowToast(winrt::Microsoft::Windows::ToastNotifications::ToastNotification const& /* toast */)
+    void ToastNotificationManager::ShowToast(winrt::Microsoft::Windows::ToastNotifications::ToastNotification const& toast)
     {
-        throw hresult_not_implemented();
+        THROW_HR_IF(WPN_E_NOTIFICATION_POSTED, toast.ToastId() != 0);
```
hulumane, line 112: This is great. It prevents misuse from the developer by preventing them from displaying the same toast again and again!

dev/ToastNotifications/ToastNotificationManager.cpp
```diff
@@ -102,9 +107,21 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
...
+            = winrt::make_self<NotificationProperties>(toast);
+
+        winrt::com_ptr<::ABI::Microsoft::Internal::ToastNotifications::INotificationTransientProperties> notificationTransientProperties
+            = winrt::make_self<NotificationTransientProperties>(toast);
+
+        auto toastAppId{ RetrieveAppId() };
+
+        DWORD notificationId = 0;
+        THROW_IF_FAILED(ToastNotifications_PostToast(toastAppId.c_str(), notificationProperties.get(), notificationTransientProperties.get(), &notificationId));
+        toast.ToastId(notificationId);
```
hulumane, line 124: Can we do a quick check here since this is critical code: THROW_HR_IF(E_UNEXPECTED, notificationId == 0);

test/TestApps/ToastNotificationsTestApp/main.cpp
hulumane: Yes. Verify that ToastId != 0

hulumane: Ideally, we want to use the History API to ensure that the Post succeeded here. Can you add that as a ToDo comment here?

@loneursid  FYI

test/TestApps/ToastNotificationsTestApp/main.cpp
```diff
@@ -378,6 +378,37 @@ bool VerifyToastExpiresOnReboot()
...
+
+    auto scope_exit = wil::scope_exit(
+        [&] {
+            toastNotificationManager.UnregisterActivator();
+        });
+
+    winrt::ToastNotification toast{ GetToastNotification() };
+
+    toastNotificationManager.ShowToast(toast);
+
```
hulumane, line 408: Verify notificationId. Also use ToastHistory to verify that it actually got posted. Put a comment for that please!

test/TestApps/ToastNotificationsTestApp/main.cpp
```diff
@@ -378,6 +378,37 @@ bool VerifyToastExpiresOnReboot()
     return true;
 }
 
+bool VerifyShowToast()
```
hulumane, line 381: VerifyShowToast_Packaged

test/ToastNotificationTests/APITests.cpp
```diff
@@ -323,5 +323,15 @@ namespace Test::ToastNotifications
         {
             RunTestUnpackaged(L"VerifyToastExpiresOnReboot", testWaitTime());
         }
+
+        TEST_METHOD(VerifyShowToast)
```
hulumane, line 327: VerifyShowToast_Packaged

dev/ToastNotifications/NotificationProperties.cpp
danielayala94: Yes, it can throw. If string length > 0, we do an ordinary `wil::make_unique_string`, which will throw in the scenario you have described.

dev/ToastNotifications/NotificationProperties.cpp
danielayala94: What are the benefits of doing this way or what are why trying to achieve?

dev/ToastNotifications/ToastNotificationManager.cpp
danielayala94: Good catch!

dev/ToastNotifications/NotificationProperties.cpp
hulumane: When we have a public API available to generate bootId in Reunion SDK, you can use that at the top of the stack without plumbing a bunch of changes here. It makes fixes less risky and more manageable.

hulumane: The same idea for CV too

test/TestApps/ToastNotificationsTestApp/main.cpp
danielayala94: IMHO there is no need to be that explicit. All unpackaged tests are marked as `_Unpackaged` and we call `RunTestUnpackaged` to run the test app as unpackaged. Anything else is a packaged-scenario test.

test/ToastNotificationTests/APITests.cpp
danielayala94: Similar resolution as in the other comment regarding test naming.

dev/ToastNotifications/NotificationProperties.cpp
danielayala94: There is CV in RS5. I don't think we need activityId for anything. Ideally, I should have noticed this and not even declare it as part of the interface declaration.

If we ever need this, we can take it separately. The UDK is ready to extract whatever GUID comes from this object.

dev/ToastNotifications/NotificationProperties.cpp
danielayala94: Fair enough for bootId.

For CV, I would imagine it will have its own datatype when introduced. Then, changes will be done anyway at the getter function to extract the CV as hstring. So, for now I'll leave it as is.

test/TestApps/ToastNotificationsTestApp/main.cpp
```diff
@@ -378,6 +378,51 @@ bool VerifyToastExpiresOnReboot()
...
+    toastNotificationManager.RegisterActivator(activationInfo);
+
+    auto scope_exit = wil::scope_exit(
+        [&] {
+            toastNotificationManager.UnregisterActivator();
+        });
+
+    winrt::ToastNotification toast{ GetToastNotification() };
+
+    toastNotificationManager.ShowToast(toast);
```
loneursid, line 414: Are we certain that this test works? When I run it on my devbox, I don't see the toast, yet the test passes. If I run the packaged version of the test, I see the toast.

dev/ToastNotifications/NotificationTransientProperties.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+    *priority = m_toastNotificationPriority;
+    return S_OK;
+}
+
+STDMETHODIMP_(HRESULT __stdcall) NotificationTransientProperties::get_CorrelationVector(_Out_ HSTRING* correlationVector) noexcept try
+{
+    // In the UDK, a null/empty CV will make the UDK to throw.
+    // Temporarily return a fixed CV until we address CV properly in the UDK.
+    *correlationVector = safe_make_unique_hstring(L"PmvzQKgYek6Sdk/T5sWaqw.0").release();
```
hulumane, line 81: I would return nullptr or empty rather than a hardcoded one here which serves no purpose. Let the UDK handle the generation of the CV for now. Atleast we can track some metrics that way.

test/TestApps/ToastNotificationsTestApp/main.cpp
danielayala94: Thanks for raising up this. We can't do more than this from a unit test perspective. We need E2E verification which I'm working on right now (manual verification).

dev/ToastNotifications/NotificationTransientProperties.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+    if (toastNotification.Priority() == winrt::Microsoft::Windows::ToastNotifications::ToastPriority::High)
+    {
+        m_toastNotificationPriority = ToastABI::ToastNotificationPriority::ToastNotificationPriority_High;
+    }
+}
+
+STDMETHODIMP_(HRESULT __stdcall) NotificationTransientProperties::get_OfflineCacheCount(_Out_ unsigned long long* offlineCacheCount) noexcept
+{
+    *offlineCacheCount = 0;
```
robertzhou-wpn, line 33: Is this property offlineCacheCount always be 0.  Are there any places that we will change its value? If not, why do we need this property?

dev/ToastNotifications/NotificationTransientProperties.cpp
danielayala94: At this point, with the UDK fix, we should be good to leave the SDK CV as nullptr or empty.

test/TestApps/ToastNotificationsTestApp/main.cpp
danielayala94: Actually, in the UDK tests we can have some form of E2E testing, if we implement the test toast sink. But this is not something we can get done before code complete. For now, we can rely on:

- Using History APIs after posting a toast
- Manual E2E verification

hulumane (approved):

test/TestApps/ToastNotificationsTestApp/main.cpp
danielayala94: I'm going to merge this change. Today I'll be tackling E2E testing. I will watch closely what happens with the unpackaged scenario. Any issues will be addressed in a future PR.

dev/ToastNotifications/ToastNotificationManager.cpp
```diff
@@ -103,9 +106,23 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
...
+        winrt::com_ptr<::ABI::Microsoft::Internal::ToastNotifications::INotificationTransientProperties> notificationTransientProperties
+            = winrt::make_self<NotificationTransientProperties>(toast);
+
+        auto toastAppId{ RetrieveToastAppId() };
+
+        DWORD notificationId = 0;
+        THROW_IF_FAILED(ToastNotifications_PostToast(toastAppId.c_str(), notificationProperties.get(), notificationTransientProperties.get(), &notificationId));
+
+        THROW_HR_IF(E_UNEXPECTED, notificationId == 0);
+        toast.ToastId(notificationId);
```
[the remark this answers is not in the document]
danielayala94: Thanks for the heads up! I can perform such change in an upcoming PR targeting L1.

dev/ToastNotifications/NotificationTransientProperties.cpp
danielayala94: It is useful for push notifications. For local toasts, this property is irrelevant.

While in the SDK layer it's not needed, we still want it in the UDK to translate internal types into the ABI types passed to the LRP.

PR #2028, opened 2022-01-26
Moving Intellisense files to single location

EricJohnson327:
Moving Intellisense to single location in Foundation repo and updated XMLs with latest information.


Scottj1s:
I think we should reconsider this.  It breaks encapsulation of transport packages.  And it introduces a risk of APIs and corresponding Intellisense being out of sync.

Scottj1s (approved):

PR #2030, opened 2022-01-27
Add CsWin32 to C# projects

sotteson1:
This adds the CsWin32 package to the C# templates. It also adds some code that uses Win32 interop via CsWin32 to the C# WinUI projects that allows them to set the window size. I built this using our internal pipeline and tested it on VS 2022 to make sure it worked correctly.


riverar:
Some questions, if you don't mind.

Are there any general consumer docs on CsWin32? Or a support lifecycle? Are there quality guarantees? Is it appropriate to include this in production templates at this stage?

DrusTheAxe:
Adding CsWin32 to the C# templates? Where's the spec for this feature?

DrusTheAxe (changes requested):
I don't see a spec re adding CsWin32 use to WinAppSDK's templates. What's the goal? What's the justifications? Caveats? Yadda yadda.

bpulliam:
I agree with the comments by @DrusTheAxe. This seems like a sample, not something that we would expect all projects to use. If you disagree, a spec and public discussion would be the next steps.

PR #2060, opened 2022-02-02
RS5 Fixes

pmpurifoy:
- Includes fixes to ChannelRequest for packaged applications in RS5
- Move from ShellExecuteEx to ComActivation for packaged applications in RS5 in background activation


dev/AppLifecycle/AppInstance.cpp
```diff
@@ -68,7 +68,7 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
 
         wil::unique_hlocal_ptr<PWSTR[]> argv{ CommandLineToArgvW(commandLine.c_str(), &argc) };
 
-        PCWSTR activationKinds[] = { c_msProtocolArgumentString, c_pushProtocolArgumentString };
+        PCWSTR activationKinds[] = { c_msProtocolArgumentString, c_pushProtocolArgumentString, c_toastProtocolArgumentString };
```
hulumane, line 71: c_appNotificationProtocolArgumentString

dev/PushNotifications/PushNotificationBackgroundTask.cpp
```diff
@@ -23,8 +29,17 @@ namespace winrt
 void PushNotificationBackgroundTask::Run(winrt::IBackgroundTaskInstance const& taskInstance)
 {
     auto appProperties = winrt::CoreApplication::Properties();
-    winrt::PushNotificationReceivedEventArgs activatedEventArgs = winrt::make<winrt::Microsoft::Windows::PushNotifications::implementation::PushNotificationReceivedEventArgs>(taskInstance);
-    appProperties.Insert(ACTIVATED_EVENT_ARGS_KEY, activatedEventArgs);
+    if (PushNotificationHelpers::IsPackagedAppScenario())
```
hulumane, line 32: None of the code in IBackgroundInstance::Run should be modified if this scenario is working correctly. 

The only difference is how the IBackgroundInstance is composed: In one scenario, BI composes it and sends it down. In the other, it is LRP. This snippet should be agnostic of that.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -89,22 +90,15 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
         THROW_IF_FAILED(notificationPlatform->RegisterFullTrustApplication(processName.get(), remoteId, &unpackagedAppUserModelId));
     }
 
-    winrt::hresult CreateChannelWithRemoteIdHelper(const winrt::guid& remoteId, ChannelDetails& channelInfo) noexcept try
+    winrt::hresult CreateChannelWithRemoteIdHelper(wil::unique_cotaskmem_string const& appUserModelId, const winrt::guid& remoteId, ChannelDetails& channelInfo) noexcept try
```
hulumane, line 93: Just appId. AppUserModelId infers to the format PRAID!PFN which is specifically for packages with identity. This appId isn't specifically for those with identity right?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -114,7 +108,7 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
         THROW_IF_FAILED(operationalCode);
 
         winrt::copy_from_abi(channelInfo.channelExpiryTime, &channelExpiryTime);
-        channelInfo.appUserModelId = winrt::hstring{ appUserModelId };
+        channelInfo.appUserModelId = winrt::hstring{ appUserModelId.get() };
```
hulumane, line 111: I would rename the property inside channelInfo to appId

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -157,8 +151,9 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
                 {
                     if (IsActivatorSupported(PushNotificationRegistrationActivators::PushTrigger))
                     {
+                        auto appUserModelId{ PushNotificationHelpers::GetAppUserModelId() };
                         ChannelDetails channelInfo{};
-                        winrt::hresult hr = CreateChannelWithRemoteIdHelper(remoteId, channelInfo);
+                        winrt::hresult hr{ CreateChannelWithRemoteIdHelper(appUserModelId, remoteId, channelInfo) };
```
hulumane, line 156: Is the else condition even needed below? We always expect this API call to succeed right?

hulumane: I would double check if the E_NOTIMPL case is valid anymore and would delete the code in the else block if not.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -198,22 +193,23 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
                     {
                         wil::unique_cotaskmem_string unpackagedAppUserModelId;
                         RegisterUnpackagedApplicationHelper(remoteId, unpackagedAppUserModelId);
```
hulumane, line 195: Add a comment here that unpackagedAppUserModelId is generated by the Long Running Process singleton.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -198,22 +193,23 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+
+                        ChannelDetails channelInfo{};
+                        winrt::hresult hr{ CreateChannelWithRemoteIdHelper(unpackagedAppUserModelId, remoteId, channelInfo) };
 
                         auto notificationPlatform{ PushNotificationHelpers::GetNotificationPlatform() };
 
                         wil::unique_cotaskmem_string processName;
                         THROW_IF_FAILED(GetCurrentProcessPath(processName));
-                        THROW_IF_FAILED(notificationPlatform->RegisterLongRunningActivator(processName.get()));
+                        THROW_IF_FAILED(notificationPlatform->RegisterLongRunningActivator(processName.get(), s_taskClsid));
```
hulumane, line 204: I have a couple of issues with the s_taskClsid:
1) It is a static which should be avoided.
2) I think it can potentially be GUID_NULL for unpackaged apps right? In that case, we need a comment here explaining that.

hulumane: Lets discuss offline

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -274,9 +270,20 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
                     THROW_HR_IF(E_INVALIDARG, s_protocolRegistration);
                 }
 
+                s_taskClsid = details.TaskClsid();
                 wil::unique_cotaskmem_string unpackagedAppUserModelId;
                 RegisterUnpackagedApplicationHelper(GUID_NULL, unpackagedAppUserModelId); // create default registration for app
```
hulumane, line 275: Add a comment here that unpackagedAppUserModelId is generated by the Long Running Process

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -274,9 +270,20 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
 
+                if (AppModel::Identity::IsPackagedProcess())
```
hulumane, line 277: We should talk about this. In general, Dveloper should be allowed to use the PushTrigger flag in conjunction with the COM activator flag for all packaged apps irrespective of whether we run in RS5 or not.

dev/PushNotifications/PushNotificationBackgroundTask.cpp
pmpurifoy: The constructor for PushNotificationReceivedEventArgs with an IBackgroundTaskInstance is only used if BI is available. Unpackaged scenarios use byte*+length | std::wstring. This is because we don't have a true IBackgroundTaskInstance and why it has to be separated.

dev/PushNotifications/PushNotificationManager.cpp
hulumane: And for unpackaged, the developer always uses ProtocolActivator flag.

dev/PushNotifications/PushNotificationsLongRunningTask.ProxyStub/NotificationsLongRunningProcess.idl
```diff
@@ -19,7 +19,7 @@ interface IWpnForegroundSink : IUnknown
 [pointer_default(unique)]
 interface INotificationsLongRunningPlatform : IUnknown
 {
-    HRESULT RegisterLongRunningActivator([in] LPCWSTR processName);
+    HRESULT RegisterLongRunningActivator([in] LPCWSTR processName, [in] GUID comServerClsid);
```
hulumane, line 22: Is this a breaking change that could impact older versions of Framework UDK?

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
```diff
@@ -32,7 +37,19 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationListener::OnRawNotificationReceived
...
-        THROW_IF_FAILED(winrt::Microsoft::Windows::PushNotifications::Helpers::ProtocolLaunchHelper(m_processName, payloadLength, payload));
+        if (m_comServerClsid == winrt::guid())
+        {
+            THROW_IF_FAILED(winrt::Microsoft::Windows::PushNotifications::Helpers::ProtocolLaunchHelper(m_processName, payloadLength, payload));
+        }
+        else
+        {
+            std::wstring payloadString(payloadArray.begin(), payloadArray.end());
+
+            auto localBackgroundTask = winrt::create_instance<winrt::Windows::ApplicationModel::Background::IBackgroundTask>(m_comServerClsid, CLSCTX_ALL);
```
hulumane, line 48: Nit: Always move next to its usage.

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
```diff
@@ -32,7 +37,19 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationListener::OnRawNotificationReceived
...
+            auto pushBackgroundTaskInstance{ winrt::make_self<PushBackgroundTaskInstance>() };
+            pushBackgroundTaskInstance->SetRawNotificationPayload(payloadString);
+            localBackgroundTask.Run(*pushBackgroundTaskInstance);
```
hulumane, line 51: Run returns HRESULT. So THROW_IF_FAILED here

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
```diff
@@ -32,7 +37,19 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationListener::OnRawNotificationReceived
...
+        {
+            THROW_IF_FAILED(winrt::Microsoft::Windows::PushNotifications::Helpers::ProtocolLaunchHelper(m_processName, payloadLength, payload));
+        }
+        else
+        {
+            std::wstring payloadString(payloadArray.begin(), payloadArray.end());
+
+            auto localBackgroundTask = winrt::create_instance<winrt::Windows::ApplicationModel::Background::IBackgroundTask>(m_comServerClsid, CLSCTX_ALL);
+            auto pushBackgroundTaskInstance{ winrt::make_self<PushBackgroundTaskInstance>() };
+            pushBackgroundTaskInstance->SetRawNotificationPayload(payloadString);
```
hulumane, line 50: You can pass the raw payload in the constructor. No need to have a setter.

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListenerManager.cpp
```diff
@@ -9,26 +9,24 @@ void NotificationListenerManager::Initialize(std::shared_ptr<ForegroundSinkManag
     m_toastRegistrationManager = toastRegistrationManager;
 }
 
-void NotificationListenerManager::SetAppIdMapping(std::map<std::wstring, std::wstring>& appIdList)
+void NotificationListenerManager::SetAppIdMapping(std::map<std::wstring, std::pair<std::wstring, winrt::guid>>& appIdList)
```
hulumane, line 12: This argument desperately needs a comment so that devs understand what exactly is being mapped.

Something along the lines of:
ProcessName->(appId, comclsid)

Please add comments when code isn't self explanatory

dev/PushNotifications/PushNotificationsLongRunningTask/PushBackgroundTaskInstance.h
```diff
@@ -0,0 +1,34 @@
+﻿#include <winrt/Windows.ApplicationModel.background.h>
+
+struct PushBackgroundTaskInstance : winrt::implements<PushBackgroundTaskInstance, winrt::Windows::ApplicationModel::Background::IBackgroundTaskInstance>
```
hulumane, line 3: Add a comment as to why this was introduced in LongRunningProcess

dev/PushNotifications/PushNotificationsLongRunningTask/PushBackgroundTaskInstance.cpp
```diff
@@ -0,0 +1,53 @@
...
+{
+    return nullptr;
+}
+
+winrt::event_token PushBackgroundTaskInstance::Canceled(winrt::Windows::ApplicationModel::Background::BackgroundTaskCanceledEventHandler const& handler)
+{
+    return winrt::event_token{};
+}
+
+void PushBackgroundTaskInstance::Canceled(winrt::event_token const& token) noexcept
```
hulumane, line 40: Remove or comment out the name of the arguments. Otherwise these result in compiler warnings since the arguments are unused.

dev/PushNotifications/PushNotificationsLongRunningTask/PushBackgroundTaskInstance.cpp
```diff
@@ -0,0 +1,53 @@
...
+}
+
+winrt::Windows::ApplicationModel::Background::BackgroundTaskDeferral PushBackgroundTaskInstance::GetDeferral()
+{
+    return nullptr;
+}
+
+winrt::Windows::Foundation::IInspectable PushBackgroundTaskInstance::TriggerDetails()
+{
+    return winrt::box_value(m_payload);
```
hulumane, line 52: Hacky workaround. Please see if you can mock this out properly or if we truly have technical limitations with mocking.

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -85,7 +88,8 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationsLongRunningPlatformImpl::RegisterL
         return S_OK;
     }
 
-    m_notificationListenerManager.AddListener(appId, processName);
+    AddComServerClsid(appId, comServerClsid);
```
hulumane, line 91: So for unpackaged apps that rely on Protocol, comServerCLSID is GUID_NULL correct?

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -240,6 +246,16 @@ const std::wstring NotificationsLongRunningPlatformImpl::BuildAppIdentifier(std:
     return guidStr.get();
 }
 
+void NotificationsLongRunningPlatformImpl::AddComServerClsid(std::wstring const& appId, winrt::guid const& comServerClsid)
+{
+    m_comServerClsidStorage.Values().Insert(appId, winrt::box_value(comServerClsid));
```
hulumane, line 251: No locks

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
```diff
@@ -240,6 +246,16 @@ const std::wstring NotificationsLongRunningPlatformImpl::BuildAppIdentifier(std:
...
+}
+
+void NotificationsLongRunningPlatformImpl::RemoveComServerClsid(std::wstring const& appId)
+{
+    m_comServerClsidStorage.Values().Remove(appId);
```
hulumane, line 256: No locks

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
hulumane: Actually is a lock needed at all? It seems to be called only from a locked context.

dev/PushNotifications/externs.h
```diff
@@ -28,5 +35,13 @@ inline winrt::Windows::Foundation::IInspectable GetArgsFromComStore()
     THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_TIMEOUT), !GetWaitHandleForArgs().wait(receiveArgsTimeoutInMSec));
 
     // If COM static store was uninit, let it throw
-    return winrt::Windows::ApplicationModel::Core::CoreApplication::Properties().Lookup(ACTIVATED_EVENT_ARGS_KEY);
+    if (PushNotificationHelpers::IsPackagedAppScenario())
```
hulumane, line 38: If (IsPackagedApp())
{
  use the COM activation key;
}
else
{
 use the protocol activator key for Push
}

hulumane: I believe the scenario is inconsequential now.

dev/PushNotifications/PushNotificationBackgroundTask.cpp
hulumane: AS discussed, we can create a mock BackgroundTaskInstance with the mock wrappers. We don't need to change this code if we do that correctly.

dev/PushNotifications/PushNotificationsLongRunningTask/platform.cpp
pmpurifoy: Don't need this function because it is only used once. Removing.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -157,8 +157,9 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
                 {
                     if (IsActivatorSupported(PushNotificationRegistrationActivators::PushTrigger))
                     {
+                        auto appUserModelId{ PushNotificationHelpers::GetAppUserModelId() };
                         ChannelDetails channelInfo{};
-                        winrt::hresult hr = CreateChannelWithRemoteIdHelper(remoteId, channelInfo);
+                        winrt::hresult hr{ CreateChannelWithRemoteIdHelper(appUserModelId, remoteId, channelInfo) };
```
hulumane, line 162: Just THROW_IF_FAILED here and remove the loop

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -157,8 +157,9 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
                 {
                     if (IsActivatorSupported(PushNotificationRegistrationActivators::PushTrigger))
                     {
+                        auto appUserModelId{ PushNotificationHelpers::GetAppUserModelId() };
```
hulumane, line 160: Make sure you test the BI flow: Foreground and background

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -173,47 +174,33 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-
-                            PushNotificationChannelManager channelManager{};
-                            winrt::PushNotificationChannel pushChannelReceived{ nullptr };
-
-                            pushChannelReceived = co_await channelManager.CreatePushNotificationChannelForApplicationAsync();
-
-                            PushNotificationTelemetry::ChannelRequestedByApi(S_OK, remoteId, usingLegacyImplementation);
-
-                            co_return winrt::make<PushNotificationCreateChannelResult>(
-                                winrt::make<PushNotificationChannel>(pushChannelReceived),
```
hulumane, line 188: We no longer hang of the wrapped PushNotificationChannel. All usages of it needs to be deleted.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -366,6 +354,7 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
 
                 GetWaitHandleForArgs().create();
 
+                s_taskClsid = taskClsid;
```
hulumane, line 357: This should be in a lock

hulumane: Also UnregisterActivator should remove the long running sink from LRP for Packaged apps

hulumane: fix unregister too to remove foreground sink with LRP for packaged apps

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
pmpurifoy: PushNotificationBackgroundTask::Run is a void function

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -274,9 +270,20 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
                     THROW_HR_IF(E_INVALIDARG, s_protocolRegistration);
                 }
 
+                s_taskClsid = details.TaskClsid();
```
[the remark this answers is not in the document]
pmpurifoy: Unpackaged apps will use the PushActivationInfo ctor with that only accepts ActivatorFlag. GUID_NULL will be implied.

pmpurifoy:
Having issues using PushRawNotification to hold payload. Will file bug to fix later on.

dev/PushNotifications/PushNotificationBackgroundTask.cpp
```diff
@@ -23,8 +29,17 @@ namespace winrt
...
+    {
+        winrt::PushNotificationReceivedEventArgs activatedEventArgs{ winrt::make<winrt::Microsoft::Windows::PushNotifications::implementation::PushNotificationReceivedEventArgs>(taskInstance) };
+        appProperties.Insert(ACTIVATED_EVENT_ARGS_KEY, activatedEventArgs);
+    }
+    else
+    {
+        winrt::hstring payload{ winrt::unbox_value<winrt::hstring>(taskInstance.TriggerDetails()) };
```
hulumane, line 39: Add a comment here that you are hacking the triggerDetails for now and opened bug *. The bug number should clearly be stated in the comments.

dev/PushNotifications/PushNotificationBackgroundTask.cpp
hulumane, line 32: Add a comment here that "BackgroundTask::Run can either be activated by the OS background Infrastructure component or the Long Running Process"

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -11,7 +11,6 @@
 #include <FrameworkUdk/PushNotifications.h>
 #include "externs.h"
 #include "PushNotificationTelemetry.h"
-#include <TerminalVelocityFeatures-PushNotifications.h>
 #include "PushNotificationUtility.h"
 
 namespace winrt::Windows
```
hulumane, line 16: Delete this namespace definitions here. We are no longer wrapping the Windows App sdk constructs anymore.

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -32,141 +31,89 @@ namespace PushNotificationHelpers
...
-                auto lock = m_lock.lock_exclusive();
-
-                if (!m_foregroundHandlerCount)
-                {
-                    auto appUserModelId{ winrt::Microsoft::Helpers::GetAppUserModelId() };
-
-                    THROW_IF_FAILED(PushNotifications_RegisterNotificationSinkForFullTrustApplication(appUserModelId.get(), this));
-                }
+            auto lock { m_lock.lock_shared() };
+            registeredEvent = bool(m_foregroundHandlers);
```
hulumane, line 70: I don't understand why we need this. Isn't this always guaranteed to be true? So why bother?

hulumane: You are just complicating the logic with this check.

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -32,141 +31,89 @@ namespace PushNotificationHelpers
...
-            else
-            {
-                auto lock = m_lock.lock_exclusive();
-
-                if (m_foregroundHandlerCount == 1)
-                {
-                    auto appUserModelId{ winrt::Microsoft::Helpers::GetAppUserModelId() };
+            auto lock { m_lock.lock_exclusive() };
+            m_foregroundHandlers.remove(token);
+            registeredEvent = bool(m_foregroundHandlers);
```
hulumane, line 104: Same comment as above. I don't really understand why this check is needed in the first place. It is always true right?

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -32,141 +31,89 @@ namespace PushNotificationHelpers
...
+        if(!registeredEvent)
+        {
+            if (PushNotificationHelpers::IsPackagedAppScenario())
+            {
+                auto appUserModelId{ winrt::Microsoft::Helpers::GetAppUserModelId() };
 
-                m_foregroundHandlers.remove(token);
-                --m_foregroundHandlerCount;
+                // Unregister foreground sink in UDK
+                THROW_IF_FAILED(PushNotifications_UnregisterNotificationSinkForFullTrustApplication(appUserModelId.get()));
```
hulumane, line 114: You cannot remove the foreground sink unless there are no registered handlers.

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -32,141 +31,89 @@ namespace PushNotificationHelpers
...
 
-                THROW_IF_FAILED(notificationsLongRunningPlatform->RegisterForegroundActivator(this, processName.get()));             
+                // Register foreground event through the LRP
+                THROW_IF_FAILED(notificationsLongRunningPlatform->RegisterForegroundActivator(this, processName.get()));
             }
-            return m_foregroundHandlers.add(handler);
         }
+
+        auto lock { m_lock.lock_exclusive() };
+        return m_foregroundHandlers.add(handler);
```
hulumane, line 95: How are handlers being ref counted? You removed the counter from earlier.

dev/PushNotifications/PushNotificationChannel.cpp
pmpurifoy: the bool operator checks if there are any registered handlers in the winrt::event. Removes the need for a handler count.

https://docs.microsoft.com/en-us/uwp/cpp-ref-for-winrt/event#eventoperator-bool

dev/PushNotifications/PushNotificationChannel.cpp
pmpurifoy: !registeredEvent checks if there are no handlers registered

dev/PushNotifications/PushNotificationChannel.cpp
hulumane: oh thats cool that they overloaded the bool operator this way. Would have expected an isEmpty or something similar.

dev/PushNotifications/PushNotificationChannel.cpp
hulumane: Addressed in previous comments

dev/PushNotifications/PushNotificationChannel.cpp
hulumane: Comment resolved above

dev/PushNotifications/PushNotificationChannel.cpp
hulumane: Comment resolved above

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -32,141 +31,89 @@ namespace PushNotificationHelpers
...
-                }
+        if(!registeredEvent)
+        {
+            if (PushNotificationHelpers::IsPackagedAppScenario())
+            {
+                auto appUserModelId{ winrt::Microsoft::Helpers::GetAppUserModelId() };
 
-                m_foregroundHandlers.remove(token);
-                --m_foregroundHandlerCount;
+                // Unregister foreground sink in UDK
```
hulumane, line 113: Self explanatory.

A more useful comment would be above the if else condition. Something along these lines,
"Apps that are treated as packaged will have their event handlers removed directly from the platform while unpackaged apps will have their event handlers removed from the Long Running Process"

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -32,141 +31,89 @@ namespace PushNotificationHelpers
...
+                THROW_IF_FAILED(PushNotifications_UnregisterNotificationSinkForFullTrustApplication(appUserModelId.get()));
             }
-        }
-        else
-        {
-            auto lock = m_lock.lock_exclusive();
-            m_foregroundHandlers.remove(token);
-            if (!--m_foregroundHandlerCount)
+            else
```
hulumane, line 116: OnRawNotificationReceived method below is broken for Packaged app scenario. I think you will need to call CoCreateInstance again and launch a new process with Payload? Test flow please :)

pmpurifoy: I've tested this on current build and downlevel build. This OnRawNotificationReceived is only for the foreground scenario.

hulumane: Correct. But it won't work for packaged apps the way it is currently implemented right?

pmpurifoy: RegisterForegroundActivator sends the channel object from the packaged app to the LRP. LRP then calls InvokeAll using the channel as a IWpnNotificationSink object.

hulumane: Lets talk offline. I was talking specifically about OnRawNotificationReceived since I cannot add a comment in that block.

pmpurifoy: Resolution: Packaged apps need to be CoCreateInstance'd instead of using ProtocolLaunchHelper.

danielayala94 (approved):
I see the ToastNotificationDemoApp is coming with this PR. Are you planning to check it in here or from the DemoApp PR?

Also, be careful with the renaming work before merging.

pmpurifoy:
> I see the ToastNotificationDemoApp is coming with this PR. Are you planning to check it in here or from the DemoApp PR?
> 
> Also, be careful with the renaming work before merging.

I'm checking in the DemoApp PR first, then this PR.

dev/AppLifecycle/ExtensionContract.h
```diff
@@ -26,7 +26,7 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
         { ExtendedActivationKind::Protocol, c_protocolContractId, &ProtocolActivatedEventArgs::Deserialize },
         { ExtendedActivationKind::StartupTask, c_startupTaskContractId, &StartupActivatedEventArgs::Deserialize },
         { ExtendedActivationKind::Push, c_pushContractId, &winrt::Microsoft::Windows::PushNotifications::Deserialize },
-        { ExtendedActivationKind::AppNotification, c_toastContractId, &winrt::Microsoft::Windows::PushNotifications::Deserialize },
+        { ExtendedActivationKind::AppNotification, c_appNotificationContractId, &winrt::Microsoft::Windows::PushNotifications::Deserialize },
```
hulumane, line 29: ITs on purpose I assume that we use the same function to deserialize push and toast?

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -32,143 +23,92 @@ namespace PushNotificationHelpers
...
 
-                ++m_foregroundHandlerCount;
+        if(!registeredEvent)
+        {
+            if(PushNotificationHelpers::IsPackagedAppScenario())
+            {
+                auto appUserModelId{ PushNotificationHelpers::GetAppUserModelId() };
 
-                return m_foregroundHandlers.add(handler);
+                // Register foreground event through the UDK
```
hulumane, line 71: Not a particularly useful comment. Instead, this would be more descriptive:
"Register a Sink with the platform initialized in the current process"

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -32,143 +23,92 @@ namespace PushNotificationHelpers
...
             {
-                auto notificationsLongRunningPlatform{ winrt::Microsoft::Helpers::GetNotificationPlatform() };
+                auto notificationsLongRunningPlatform{ PushNotificationHelpers::GetNotificationPlatform() };
 
                 wil::unique_cotaskmem_string processName;
                 THROW_IF_FAILED(GetCurrentProcessPath(processName));
 
-                THROW_IF_FAILED(notificationsLongRunningPlatform->RegisterForegroundActivator(this, processName.get()));             
+                // Register foreground event through the LRP
+                THROW_IF_FAILED(notificationsLongRunningPlatform->RegisterForegroundActivator(this, processName.get()));
```
hulumane, line 82: I would reword this comment:
"Register a Sink with platform brokered through the LongRunningProcess"

hulumane: Avoid use of LRP which doesn't mean much outside of our team.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -156,72 +160,38 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-
-                        PushNotificationTelemetry::ChannelRequestedByApi(S_OK, remoteId, usingLegacyImplementation);
-
-                        co_return winrt::make<PushNotificationCreateChannelResult>(
-                            winrt::make<PushNotificationChannel>(pushChannelReceived),
-                            S_OK,
-                            PushNotificationChannelStatus::CompletedSuccess);
                     }
+
+                    PushNotificationTelemetry::ChannelRequestedByApi(S_OK, remoteId, usingLegacyImplementation);
```
hulumane, line 186: I'm not sure how this compiles. This variable is no longer defined and is not apllicable to the flow either.

dev/PushNotifications/PushNotifications.idl
```diff
@@ -112,24 +113,27 @@ namespace Microsoft.Windows.PushNotifications
...
-        static void UnregisterAllActivators();
+        void UnregisterAllActivators();
 
         // Request a Push Channel with an encoded RemoteId from WNS. RemoteId is an AAD identifier GUID
-        static Windows.Foundation.IAsyncOperationWithProgress<PushNotificationCreateChannelResult, PushNotificationCreateChannelStatus> CreateChannelAsync(Guid remoteId);
+        Windows.Foundation.IAsyncOperationWithProgress<PushNotificationCreateChannelResult, PushNotificationCreateChannelStatus> CreateChannelAsync(Guid remoteId);
 
         // Applications will call this to check which flags are supported for RegisterActivator
-        static Boolean IsActivatorSupported(PushNotificationRegistrationActivators activators);
+        Boolean IsActivatorSupported(PushNotificationRegistrationActivators activators);
```
hulumane, line 137: This should still be static IMHO

hulumane (approved):

dev/AppLifecycle/ExtensionContract.h
pmpurifoy: Yup. We are pulling the args from the static store which can be reused by both Push/AppNotification. :)

PR #2062, opened 2022-02-03
Enable ToastProgressData Updates

sharath2727:
Description: Support Toast Progress Updates for a given tag and group

Validation: Implemented Unittests as part of the ToastNotificationTests

### DIFF
--- a/dev/ToastNotifications/NotificationProperties.cpp
+++ b/dev/ToastNotifications/NotificationProperties.cpp
@@ -7,6 +7,7 @@
 #include "NotificationProperties.h"
 #include "ToastNotificationUtility.h"
 #include "../PushNotifications/PushNotificationUtility.h"
+#include "ToastProgressDataABI.h"
 
 namespace winrt
 {
@@ -43,8 +44,7 @@ NotificationProperties::NotificationProperties(winrt::ToastNotification const& t
 
     m_expiresOnReboot = toastNotification.ExpiresOnReboot();
 
-    // TODO: Uncomment once ToastProgressData is implemented
-    // m_progressData = toastNotification.ProgressData();
+    m_toastProgressData = winrt::make_self<ToastProgressDataABI>(toastNotification.ProgressData());
 }
 
 STDMETHODIMP_(HRESULT __stdcall) NotificationProperties::get_NotificationId(_Out_ unsigned int* notificationId) noexcept
@@ -130,10 +130,10 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationProperties::get_ExpiresOnReboot(_Ou
 
 STDMETHODIMP_(HRESULT __stdcall) NotificationProperties::get_ToastProgressData(_Out_ ToastABI::IToastProgressData** progressData) noexcept
 {
-    // TODO: Uncomment once ToastProgressData is implemented
-    // auto lock = m_lock.lock_shared();
-    // *toastProgressData = m_toastProgressData.get();
-    *progressData = nullptr;
+    auto lock = m_lock.lock_shared();
+
+    m_toastProgressData.copy_to(progressData);
+
     return S_OK;
 }
 
--- a/dev/ToastNotifications/NotificationProperties.cpp
+++ b/dev/ToastNotifications/NotificationProperties.cpp
@@ -49,7 +49,7 @@ NotificationProperties::NotificationProperties(winrt::ToastNotification const& t
 
 STDMETHODIMP_(HRESULT __stdcall) NotificationProperties::get_NotificationId(_Out_ unsigned int* notificationId) noexcept
 {
-    auto lock = m_lock.lock_shared();
+    auto lock{ m_lock.lock_shared() };
 
     *notificationId = m_notificationId;
     return S_OK;
@@ -57,7 +57,7 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationProperties::get_NotificationId(_Out
 
 STDMETHODIMP_(HRESULT __stdcall) NotificationProperties::get_Payload(_Out_ unsigned int* payloadSize, _Out_ byte** payload) noexcept try
 {
-    auto lock = m_lock.lock_shared();
+    auto lock{ m_lock.lock_shared() };
 
     size_t tempPayloadSize = m_payload.size();
 
@@ -72,7 +72,7 @@ CATCH_RETURN()
 
 STDMETHODIMP_(HRESULT __stdcall) NotificationProperties::get_PayloadSize(_Out_ unsigned int* payloadSize) noexcept
 {
-    auto lock = m_lock.lock_shared();
+    auto lock{ m_lock.lock_shared() };
 
     *payloadSize = static_cast<unsigned int>(m_payload.size());
     return S_OK;
@@ -80,7 +80,7 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationProperties::get_PayloadSize(_Out_ u
 
 STDMETHODIMP_(HRESULT __stdcall) NotificationProperties::get_Tag(_Out_ HSTRING* tag) noexcept try
 {
-    auto lock = m_lock.lock_shared();
+    auto lock{ m_lock.lock_shared() };
 
     *tag = safe_make_unique_hstring(m_tag.c_str()).release();
     return S_OK;
@@ -89,7 +89,7 @@ CATCH_RETURN()
 
 STDMETHODIMP_(HRESULT __stdcall) NotificationProperties::get_Group(_Out_ HSTRING* group) noexcept try
 {
-    auto lock = m_lock.lock_shared();
+    auto lock{ m_lock.lock_shared() };
 
     *group = safe_make_unique_hstring(m_group.c_str()).release();
     return S_OK;
@@ -98,7 +98,7 @@ CATCH_RETURN()
 
 STDMETHODIMP_(HRESULT __stdcall) NotificationProperties::get_Expiry(_Out_ unsigned long long* expiry) noexcept
 {
-    auto lock = m_lock.lock_shared();
+    auto lock{ m_lock.lock_shared() };
 
     *expiry = static_cast<ULONGLONG>(m_expiry.dwHighDateTime) << 32 | static_cast<ULONGLONG>(m_expiry.dwLowDateTime);
     return S_OK;
@@ -106,7 +106,7 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationProperties::get_Expiry(_Out_ unsign
 
 STDMETHODIMP_(HRESULT __stdcall) NotificationProperties::get_ArrivalTime(_Out_ unsigned long long* arrivalTime) noexcept
 {
-    auto lock = m_lock.lock_shared();
+    auto lock{ m_lock.lock_shared() };
 
     *arrivalTime = static_cast<ULONGLONG>(m_arrivalTime.dwHighDateTime) << 32 | static_cast<ULONGLONG>(m_arrivalTime.dwLowDateTime);
     return S_OK;
@@ -114,7 +114,7 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationProperties::get_ArrivalTime(_Out_ u
 
 STDMETHODIMP_(HRESULT __stdcall) NotificationProperties::get_BootId(_Out_ unsigned long long* bootId) noexcept
 {
-    auto lock = m_lock.lock_shared();
+    auto lock{ m_lock.lock_shared() };
 
     *bootId = m_bootId;
     return S_OK;
@@ -122,7 +122,7 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationProperties::get_BootId(_Out_ unsign
 
 STDMETHODIMP_(HRESULT __stdcall) NotificationProperties::get_ExpiresOnReboot(_Out_ boolean* expiresOnReboot) noexcept
 {
-    auto lock = m_lock.lock_shared();
+    auto lock{ m_lock.lock_shared() };
 
     *expiresOnReboot = m_expiresOnReboot;
     return S_OK;
@@ -130,7 +130,7 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationProperties::get_ExpiresOnReboot(_Ou
 
 STDMETHODIMP_(HRESULT __stdcall) NotificationProperties::get_ToastProgressData(_Out_ ToastABI::IToastProgressData** progressData) noexcept
 {
-    auto lock = m_lock.lock_shared();
+    auto lock{ m_lock.lock_shared() };
 
     m_toastProgressData.copy_to(progressData);
 
--- a/dev/ToastNotifications/ToastNotificationManager.cpp
+++ b/dev/ToastNotifications/ToastNotificationManager.cpp
@@ -156,9 +156,7 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
 
     winrt::Windows::Foundation::IAsyncOperation<winrt::Microsoft::Windows::ToastNotifications::ToastProgressResult> ToastNotificationManager::UpdateToastProgressDataAsync(winrt::Microsoft::Windows::ToastNotifications::ToastProgressData const data, hstring const tag)
     {
-        auto toastProgressResult = co_await UpdateToastProgressDataAsync(data, tag, L"");
-
-        co_return toastProgressResult;
+        co_return co_await UpdateToastProgressDataAsync(data, tag, L"");
     }
 
     winrt::Microsoft::Windows::ToastNotifications::ToastNotificationSetting ToastNotificationManager::Setting()
--- a/dev/ToastNotifications/ToastNotificationManager.cpp
+++ b/dev/ToastNotifications/ToastNotificationManager.cpp
@@ -131,6 +131,11 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
     // For other scenarios the API throws
     winrt::Windows::Foundation::IAsyncOperation<winrt::Microsoft::Windows::ToastNotifications::ToastProgressResult> ToastNotificationManager::UpdateToastProgressDataAsync(winrt::Microsoft::Windows::ToastNotifications::ToastProgressData const& data, hstring const& tag, hstring const& group)
     {
+        if ((tag == winrt::hstring(L"") && group == winrt::hstring(L"")) || (tag == winrt::hstring(L"") && group != winrt::hstring(L"")))
+        {
+            co_return winrt::ToastProgressResult::NotificationNotFound;
+        }
+
         std::wstring appId{ RetrieveToastAppId() };
 
         winrt::com_ptr<::ABI::Microsoft::Internal::ToastNotifications::IToastProgressData> toastProgressData{ winrt::make_self<ToastProgressDataABI>(data) };
--- a/dev/ToastNotifications/ToastNotificationManager.cpp
+++ b/dev/ToastNotifications/ToastNotificationManager.cpp
@@ -130,22 +130,15 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
         toast.ToastId(notificationId);
     }
 
-    // API updates notification data when both tag and group are provided or when tag is provided.
-    // For other scenarios the API throws
-    winrt::Windows::Foundation::IAsyncOperation<winrt::Microsoft::Windows::ToastNotifications::ToastProgressResult> ToastNotificationManager::UpdateToastProgressDataAsync(winrt::Microsoft::Windows::ToastNotifications::ToastProgressData const& data, hstring const& tag, hstring const& group)
+    winrt::Windows::Foundation::IAsyncOperation<winrt::Microsoft::Windows::ToastNotifications::ToastProgressResult> ToastNotificationManager::UpdateToastProgressDataAsync(winrt::Microsoft::Windows::ToastNotifications::ToastProgressData const data, hstring const tag, hstring const group)
     {
-        if (tag == winrt::hstring(L""))
-        {
-            co_return winrt::ToastProgressResult::NotificationNotFound;
-        }
+        THROW_HR_IF_MSG(E_INVALIDARG, tag == winrt::hstring(L""), "Update operation isn't guarenteed to find a specific notification to replace correctly.");
 
         std::wstring appId{ RetrieveToastAppId() };
 
         winrt::com_ptr<ToastABI::IToastProgressData> toastProgressData{ winrt::make_self<ToastProgressDataABI>(data) };
 
         auto hr = ToastNotifications_UpdateNotificationData(appId.c_str(), tag.c_str(), group.c_str(), toastProgressData.get());
-
-        THROW_HR_IF(hr, (hr != S_OK) && (hr != E_NOT_SET));
 
         if (SUCCEEDED(hr))
         {
@@ -155,9 +148,13 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
         {
             co_return winrt::ToastProgressResult::NotificationNotFound;
         }
+        else
+        {
+            THROW_HR(hr);
+        }
     }
 
-    winrt::Windows::Foundation::IAsyncOperation<winrt::Microsoft::Windows::ToastNotifications::ToastProgressResult> ToastNotificationManager::UpdateToastProgressDataAsync(winrt::Microsoft::Windows::ToastNotifications::ToastProgressData const& data, hstring const& tag)
+    winrt::Windows::Foundation::IAsyncOperation<winrt::Microsoft::Windows::ToastNotifications::ToastProgressResult> ToastNotificationManager::UpdateToastProgressDataAsync(winrt::Microsoft::Windows::ToastNotifications::ToastProgressData const data, hstring const tag)
     {
         auto toastProgressResult = co_await UpdateToastProgressDataAsync(data, tag, L"");
 
--- a/dev/ToastNotifications/ToastNotificationManager.cpp
+++ b/dev/ToastNotifications/ToastNotificationManager.cpp
@@ -22,11 +22,6 @@ winrt::event<winrt::Windows::Foundation::EventHandler<winrt::Microsoft::Windows:
     return g_toastHandlers;
 }
 
-namespace abi
-{
-    using namespace ABI::Microsoft::Internal::ToastNotifications;
-}
-
 namespace winrt
 {
     using namespace winrt::Windows::UI;
@@ -132,17 +127,18 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
         toast.ToastId(notificationId);
     }
 
-    // API updates notification data when both tag and group are provided or when tag is provided but group is empty.
-    // For other scenarios the API returns ToastProgressResult::Failed.
-    winrt::Windows::Foundation::IAsyncOperation<winrt::Microsoft::Windows::ToastNotifications::ToastProgressResult> ToastNotificationManager::UpdateToastProgressDataAsync(winrt::Microsoft::Windows::ToastNotifications::ToastProgressData data, hstring tag, hstring group)
+    // API updates notification data when both tag and group are provided or when tag is provided.
+    // For other scenarios the API throws
+    winrt::Windows::Foundation::IAsyncOperation<winrt::Microsoft::Windows::ToastNotifications::ToastProgressResult> ToastNotificationManager::UpdateToastProgressDataAsync(winrt::Microsoft::Windows::ToastNotifications::ToastProgressData const& data, hstring const& tag, hstring const& group)
     {
         std::wstring appId{ RetrieveToastAppId() };
 
-        winrt::com_ptr<::ABI::Microsoft::Internal::ToastNotifications::IToastProgressData> toastProgressData
-            = winrt::make_self<ToastProgressDataABI>(data);
+        winrt::com_ptr<::ABI::Microsoft::Internal::ToastNotifications::IToastProgressData> toastProgressData{ winrt::make_self<ToastProgressDataABI>(data) };
 
         auto hr = ToastNotifications_UpdateNotificationData(appId.c_str(), tag.c_str(), group.c_str(), toastProgressData.get());
 
+        THROW_HR_IF(hr, (hr != S_OK) && (hr != E_NOT_SET));
+
         if (SUCCEEDED(hr))
         {
             co_return winrt::ToastProgressResult::Succeeded;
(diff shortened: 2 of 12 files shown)


dev/ToastNotifications/ToastNotificationManager.cpp
```diff
@@ -21,11 +22,17 @@ winrt::event<winrt::Windows::Foundation::EventHandler<winrt::Microsoft::Windows:
     return g_toastHandlers;
 }
 
+namespace abi
```
danielayala94, line 25: Either remove it or use it, I don't see the namespace being used at all.

danielayala94: I would call it ToastABI for consistency with similar usages.

dev/ToastNotifications/ToastNotificationManager.cpp
```diff
@@ -124,13 +131,49 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
         THROW_HR_IF(E_UNEXPECTED, notificationId == 0);
         toast.ToastId(notificationId);
     }
-    winrt::Windows::Foundation::IAsyncOperation<winrt::Microsoft::Windows::ToastNotifications::ToastProgressResult> ToastNotificationManager::UpdateToastProgressDataAsync(winrt::Microsoft::Windows::ToastNotifications::ToastProgressData /* data */, hstring /* tag */, hstring /* group */)
+
+    // API updates notification data when both tag and group are provided or when tag is provided but group is empty.
```
danielayala94, line 135: nit:
```suggestion
    // API updates notification data when both tag and group are provided or when tag is provided but the API fails if group is empty.
```

dev/ToastNotifications/ToastProgressDataABI.h
```diff
@@ -0,0 +1,27 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#pragma once
+
+#include <frameworkUdk/ToastNotificationsRT.h>
+
+struct ToastProgressDataABI : winrt::implements<ToastProgressDataABI, ::ABI::Microsoft::Internal::ToastNotifications::IToastProgressData>
```
danielayala94, line 8: Probably after @hulumane does the renaming work, this class can just be named `ToastProgressData` 😊.

test/TestApps/ToastNotificationsTestApp/main.cpp
```diff
@@ -435,6 +443,171 @@ std::string unitTestNameFromLaunchArguments(const winrt::ILaunchActivatedEventAr
...
+    auto activationInfo = winrt::ToastActivationInfo::CreateFromToastAssets(assets);
+
+    winrt::ToastNotificationManager::Default().RegisterActivator(activationInfo);
+
+    auto scope_exit = wil::scope_exit(
+        [&] {
+            winrt::ToastNotificationManager::Default().UnregisterActivator();
+        });
+
+    PostToastHelper(L"Tag", L"Group");
```
danielayala94, line 571: Any tests for empty tag but valid group?

test/TestApps/ToastNotificationsTestApp/main.cpp
```diff
@@ -435,6 +443,171 @@ std::string unitTestNameFromLaunchArguments(const winrt::ILaunchActivatedEventAr
     return unitTestName;
 }
 
+constexpr auto timeout{ std::chrono::seconds(5) };
```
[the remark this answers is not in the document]
danielayala94: Also, this is used only in the progress result helper. I think you are fine specifying the time directly in `wait_for`.

test/ToastNotificationTests/APITests.cpp
```diff
@@ -326,7 +326,47 @@ namespace Test::ToastNotifications
 
         TEST_METHOD(VerifyToastExpiresOnReboot_Unpackaged)
         {
-            RunTestUnpackaged(L"VerifyToastExpiresOnReboot", testWaitTime());
+            RunTest(L"VerifyToastExpiresOnReboot", testWaitTime());
```
danielayala94, line 329: Why did you modify this?

danielayala94 (changes requested):

dev/ToastNotifications/ToastNotificationManager.cpp
```diff
@@ -124,13 +131,49 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
...
+    // For other scenarios the API returns ToastProgressResult::Failed.
+    winrt::Windows::Foundation::IAsyncOperation<winrt::Microsoft::Windows::ToastNotifications::ToastProgressResult> ToastNotificationManager::UpdateToastProgressDataAsync(winrt::Microsoft::Windows::ToastNotifications::ToastProgressData data, hstring tag, hstring group)
     {
-        throw hresult_not_implemented();
+        std::wstring appId{ RetrieveToastAppId() };
+
+        winrt::com_ptr<::ABI::Microsoft::Internal::ToastNotifications::IToastProgressData> toastProgressData
+            = winrt::make_self<ToastProgressDataABI>(data);
+
```
hulumane, line 143: I was thinking about this a bit more. If only the group is specified and the tag is nullptr or empty, I believe we should return ProgressResult::NotificationNotFound.

Exceptions are only for fatal errors. Not programmer errors and shouldn't be thrown under winrt APIs in principle.

dev/ToastNotifications/ToastNotificationManager.cpp
```diff
@@ -124,13 +131,49 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
...
+        if (SUCCEEDED(hr))
+        {
+            co_return winrt::ToastProgressResult::Succeeded;
+        }
+        else if (hr == E_NOT_SET)
+        {
+            co_return winrt::ToastProgressResult::NotificationNotFound;
+        }
+
+        co_return winrt::ToastProgressResult::Failed;
```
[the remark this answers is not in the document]
hulumane: Yes do not return failed. Only return Succeeded and NotFound for now

dev/ToastNotifications/ToastNotificationManager.cpp
```diff
@@ -124,13 +131,49 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
...
     }
-    winrt::Windows::Foundation::IAsyncOperation<winrt::Microsoft::Windows::ToastNotifications::ToastProgressResult> ToastNotificationManager::UpdateToastProgressDataAsync(winrt::Microsoft::Windows::ToastNotifications::ToastProgressData /* data */, hstring /* tag */)
+
+    winrt::Windows::Foundation::IAsyncOperation<winrt::Microsoft::Windows::ToastNotifications::ToastProgressResult> ToastNotificationManager::UpdateToastProgressDataAsync(winrt::Microsoft::Windows::ToastNotifications::ToastProgressData data, hstring tag)
     {
-        throw hresult_not_implemented();
+        std::wstring appId{ RetrieveToastAppId() };
```
hulumane, line 160: Don't reimplement this function. Just call the function above with null group and non-empty tag.

dev/ToastNotifications/ToastProgressData.cpp
```diff
@@ -6,42 +6,42 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
 {
     uint32_t ToastProgressData::SequenceNumber()
     {
-        throw hresult_not_implemented();
+        return m_sequenceNumber;
```
hulumane, line 9: Missing locks everywhere

dev/ToastNotifications/ToastProgressData.h
```diff
@@ -17,6 +17,13 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
         void ValueStringOverride(hstring const& value);
         hstring Status();
         void Status(hstring const& value);
+
+    private:
+        uint32_t m_sequenceNumber;
+        hstring m_title;
+        double m_progressValue;
```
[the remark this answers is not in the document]
hulumane: I think its best rather than have undefined values.

dev/ToastNotifications/ToastProgressDataABI.cpp
```diff
@@ -0,0 +1,69 @@
...
+{
+    using namespace ::ABI::Microsoft::Internal::ToastNotifications;
+}
+
+ToastProgressDataABI::ToastProgressDataABI(winrt::ToastProgressData const& progressData)
+{
+    m_progressData = progressData;
+}
+
+STDMETHODIMP_(HRESULT __stdcall) ToastProgressDataABI::get_SequenceNumber(_Out_ unsigned int* value)
```
hulumane, line 25: noexcept for all methods returning hresult

dev/ToastNotifications/ToastProgressDataABI.h
hulumane: yep. Please file a bug Sharath. We can address after feature complete.

dev/ToastNotifications/ToastProgressDataABI.h
```diff
@@ -0,0 +1,27 @@
...
+{
+    ToastProgressDataABI(winrt::Microsoft::Windows::ToastNotifications::ToastProgressData const& progressData);
+
+    STDMETHOD(get_SequenceNumber)(_Out_ unsigned int* value);
```
hulumane, line 12: noexcept for all methods returning hresult

test/TestApps/ToastNotificationsTestApp/main.cpp
```diff
@@ -298,7 +307,6 @@ bool VerifyToastProgressDataFromToast()
     {
```
hulumane, line 307: Need to compare all the internal data in progressdata as well.

test/TestApps/ToastNotificationsTestApp/main.cpp
hulumane: That test should return Not_Found as part of enum because there is no toast to replace.

dev/ToastNotifications/NotificationProperties.cpp
```diff
@@ -130,10 +130,10 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationProperties::get_ExpiresOnReboot(_Ou
 
 STDMETHODIMP_(HRESULT __stdcall) NotificationProperties::get_ToastProgressData(_Out_ ToastABI::IToastProgressData** progressData) noexcept
 {
-    // TODO: Uncomment once ToastProgressData is implemented
-    // auto lock = m_lock.lock_shared();
-    // *toastProgressData = m_toastProgressData.get();
-    *progressData = nullptr;
+    auto lock = m_lock.lock_shared();
```
DrusTheAxe, line 133: NIT: `auto lock{ m_lock.lock_shared() };`

dev/ToastNotifications/ToastNotificationManager.cpp
```diff
@@ -124,13 +131,49 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
...
+
+    // API updates notification data when both tag and group are provided or when tag is provided but group is empty.
+    // For other scenarios the API returns ToastProgressResult::Failed.
+    winrt::Windows::Foundation::IAsyncOperation<winrt::Microsoft::Windows::ToastNotifications::ToastProgressResult> ToastNotificationManager::UpdateToastProgressDataAsync(winrt::Microsoft::Windows::ToastNotifications::ToastProgressData data, hstring tag, hstring group)
     {
-        throw hresult_not_implemented();
+        std::wstring appId{ RetrieveToastAppId() };
+
+        winrt::com_ptr<::ABI::Microsoft::Internal::ToastNotifications::IToastProgressData> toastProgressData
+            = winrt::make_self<ToastProgressDataABI>(data);
```
DrusTheAxe, line 142: Ohhhh so close. Got it right just 2 lives above ;-)

`winrt::com_ptr<...> toastProgressData{ winrt::make_self<...>(data); }`

dev/ToastNotifications/ToastNotificationManager.cpp
DrusTheAxe: Is NotFound due to bad input by the developer, that they could and should have known better? Or because the underlying data changed (perhaps by another thread) and the developer couldn't have reasonably known and prevented?

i.e. is it like 'file not found' or 'array out of bounds'? The latter is something the developer can and should know and have prevented. The former can happen beyond the devs control so it's something they should be prepared to encounter

dev/ToastNotifications/ToastNotificationManager.cpp
```diff
@@ -124,13 +131,49 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
...
+
+        auto hr = ToastNotifications_UpdateNotificationData(appId.c_str(), tag.c_str(), group.c_str(), toastProgressData.get());
```
DrusTheAxe, line 144: Perhaps `auto hr{ LOG_IF_FAILED(ToastNotification_UpdateNotificationdata(...)) };` ?

Below you swallow `hr` and map it to Succeeded | NotFound | Failed so you lose the actual error details if something was wrong. `LOG_IF_FAILED` would help in that case.

dev/ToastNotifications/ToastNotificationManager.cpp
```diff
@@ -124,13 +131,49 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
         THROW_HR_IF(E_UNEXPECTED, notificationId == 0);
         toast.ToastId(notificationId);
     }
-    winrt::Windows::Foundation::IAsyncOperation<winrt::Microsoft::Windows::ToastNotifications::ToastProgressResult> ToastNotificationManager::UpdateToastProgressDataAsync(winrt::Microsoft::Windows::ToastNotifications::ToastProgressData /* data */, hstring /* tag */, hstring /* group */)
+
+    // API updates notification data when both tag and group are provided or when tag is provided but group is empty.
+    // For other scenarios the API returns ToastProgressResult::Failed.
+    winrt::Windows::Foundation::IAsyncOperation<winrt::Microsoft::Windows::ToastNotifications::ToastProgressResult> ToastNotificationManager::UpdateToastProgressDataAsync(winrt::Microsoft::Windows::ToastNotifications::ToastProgressData data, hstring tag, hstring group)
```
DrusTheAxe, line 137: `...ToastProgressData data, hstring tag, hstring group)`

shouldn't that be

`...ToastProgressData& data, const hstring& tag, const hstring& group)`

?

dev/ToastNotifications/ToastProgressData.h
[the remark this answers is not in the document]
DrusTheAxe: Yes. Always initialize attributes and variables. Classes have ctors, everything else needs initialization. Thus `{}` is your friend

```
uint32_t m_sequenceNumber{};
hstring m_title;
double m_progressValue{};
```

dev/ToastNotifications/ToastProgressDataABI.cpp
DrusTheAxe: Isn't that implied by being WinRT methods?

I thought coding guidelines was to minimize use of noexcept, mainly boiling down to use noexcept for Flat-C APIs but not WinRT or implementation details.

@jonwis what's the latest guidance?

DrusTheAxe, line 25: Should be just `STDMETHODIMP`

`STDMETHODIMP_(HRESULT __stdcall)` has 2 problems:

1. `STDMETHOD_(HRESULT)` is redundant with `STDMETHODIMP`. STDMETHOD_(type) is for when you don't return an HRESULT e.g. `STDMETHOD_(void)`
2. STDMETHODIMP_() already includes __stdcall

See winnrt.h
```
#define STDAPICALLTYPE          __stdcall
...
#define STDMETHODIMP            HRESULT STDMETHODCALLTYPE
#define STDMETHODIMP_(type)     type STDMETHODCALLTYPE
```

Thus...

* `STDMETHODIMP` == `HRESULT __stdcall`
* `STDMETHODIMP_(x)` == `x __stdcall`

dev/ToastNotifications/ToastNotificationManager.cpp
sharath2727: Actually it might not be needed now. We will be throwing hr if Failed scenario. Thanks.

dev/ToastNotifications/ToastNotificationManager.cpp
sharath2727: Actually I haven't really modified it, it was what the tool generated and I am using the same syntax. I will modify it thanks.

dev/ToastNotifications/ToastNotificationManager.cpp
sharath2727: Generally when we throw Notfound is: The developer tried to update notification data for a given combination of tag and group. when developer hasn't even posted a notification but tried to update notification data using the above API then we return Notification not found error. 

"I was thinking about this a bit more. If only the group is specified and the tag is nullptr or empty, I believe we should return ProgressResult::NotificationNotFound." -  I accept that this is not fatal error but I don't think NotificationNotfound is the right result to return. Maybe a different error code like ProgressResult::Unsupported?

dev/ToastNotifications/ToastProgressDataABI.h
sharath2727: https://github.com/microsoft/WindowsAppSDK/issues/2067

dev/ToastNotifications/ToastNotificationManager.cpp
```diff
@@ -124,13 +131,49 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
...
+        else if (hr == E_NOT_SET)
+        {
+            co_return winrt::ToastProgressResult::NotificationNotFound;
+        }
+
+        co_return winrt::ToastProgressResult::Failed;
     }
-    winrt::Windows::Foundation::IAsyncOperation<winrt::Microsoft::Windows::ToastNotifications::ToastProgressResult> ToastNotificationManager::UpdateToastProgressDataAsync(winrt::Microsoft::Windows::ToastNotifications::ToastProgressData /* data */, hstring /* tag */)
+
+    winrt::Windows::Foundation::IAsyncOperation<winrt::Microsoft::Windows::ToastNotifications::ToastProgressResult> ToastNotificationManager::UpdateToastProgressDataAsync(winrt::Microsoft::Windows::ToastNotifications::ToastProgressData data, hstring tag)
```
robertzhou-wpn, line 158: As Howard pointed out in another method, the signature of this API might need to be 
...(winrt::Microsoft::Windows::ToastNotifications::ToastProgressData& data, hstring& tag). Do we know why the auto-generated API signature not use references here?

sharath2727: I updated it to use what howard pointed out. But I don't know why the auto generated API did not use the reference. It was strange even the .h file did not have references .

dev/ToastNotifications/ToastProgressDataABI.cpp
sharath2727: @hulumane @jonwis what is the latest guidance here?

dev/ToastNotifications/ToastNotificationManager.cpp
hulumane: If we really look into the platform code though, even if there are multiple notifications with the same group, we cannot select between them. ie. we cannot find the right notification to replace. So NotFound sounds correct to me.

@DrusTheAxe  Yes this is a programmer error ie. By Design. No exceptions right on principle?

DrusTheAxe: Exceptions should only occur on excecptional conditions.

Bad developer input they could have prevented (e.g. array-out-of-bounds) is fair game for exceptions.

The canonical example I always think of is FileExists(). That has **3** possible outcomes:

1. File found == Success
2. File not found == Failure
3. Something went wrong == Error

'Something went wrong' can range from out-of-memory to disk subsystem threw an error to developer specified illegal input to the function e.g. filename=null when the function's documented as requiring a non-null filename.

Success+Failure should communicate their results w/o exceptions. Only Errors should throw exceptions e.g. FileExists() throwing an exception when not-found is a bad idea.

sharath2727: ok.

dev/ToastNotifications/ToastNotificationManager.cpp
robertzhou-wpn: @hulumane and @DrusTheAxe , Any idea on this? Is this a bug in the auto generation tool?

dev/ToastNotifications/ToastNotificationManager.cpp
```diff
@@ -124,14 +126,41 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
...
         toast.ToastId(notificationId);
     }
-    winrt::Windows::Foundation::IAsyncOperation<winrt::Microsoft::Windows::ToastNotifications::ToastProgressResult> ToastNotificationManager::UpdateToastProgressDataAsync(winrt::Microsoft::Windows::ToastNotifications::ToastProgressData /* data */, hstring /* tag */, hstring /* group */)
+
+    // API updates notification data when both tag and group are provided or when tag is provided.
+    // For other scenarios the API throws
+    winrt::Windows::Foundation::IAsyncOperation<winrt::Microsoft::Windows::ToastNotifications::ToastProgressResult> ToastNotificationManager::UpdateToastProgressDataAsync(winrt::Microsoft::Windows::ToastNotifications::ToastProgressData const& data, hstring const& tag, hstring const& group)
     {
-        throw hresult_not_implemented();
+        if ((tag == winrt::hstring(L"") && group == winrt::hstring(L"")) || (tag == winrt::hstring(L"") && group != winrt::hstring(L"")))
```
danielayala94, line 134: We return `NotificationNotFound` if:

`tag == winrt::hstring(L"") && group == winrt::hstring(L"")`
OR
`tag == winrt::hstring(L"") && group != winrt::hstring(L"")`

which means you can get rid of the group check, it doesn't matter.

But, isn't a valid a combination an empty tag but valid group?

dev/ToastNotifications/ToastNotificationManager.cpp
```diff
@@ -124,14 +126,41 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
...
+        {
+            co_return winrt::ToastProgressResult::NotificationNotFound;
+        }
+
+        std::wstring appId{ RetrieveToastAppId() };
+
+        winrt::com_ptr<::ABI::Microsoft::Internal::ToastNotifications::IToastProgressData> toastProgressData{ winrt::make_self<ToastProgressDataABI>(data) };
```
danielayala94, line 141: Please use a namespace for the ABI types, this var declaration is unnecessarily long.

dev/ToastNotifications/ToastNotificationManager.cpp
```diff
@@ -124,14 +126,41 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
...
+
+        auto hr = ToastNotifications_UpdateNotificationData(appId.c_str(), tag.c_str(), group.c_str(), toastProgressData.get());
+
+        THROW_HR_IF(hr, (hr != S_OK) && (hr != E_NOT_SET));
+
+        if (SUCCEEDED(hr))
+        {
+            co_return winrt::ToastProgressResult::Succeeded;
+        }
+        else if (hr == E_NOT_SET)
```
danielayala94, line 151: You can get rid of the `else if` condition. We know the remaining scenario is `E_NOT_SET` given the THROW macro before.

dev/ToastNotifications/ToastNotificationManager.cpp
```diff
@@ -124,14 +126,41 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
...
+        {
+            co_return winrt::ToastProgressResult::NotificationNotFound;
+        }
     }
-    winrt::Windows::Foundation::IAsyncOperation<winrt::Microsoft::Windows::ToastNotifications::ToastProgressResult> ToastNotificationManager::UpdateToastProgressDataAsync(winrt::Microsoft::Windows::ToastNotifications::ToastProgressData /* data */, hstring /* tag */)
+
+    winrt::Windows::Foundation::IAsyncOperation<winrt::Microsoft::Windows::ToastNotifications::ToastProgressResult> ToastNotificationManager::UpdateToastProgressDataAsync(winrt::Microsoft::Windows::ToastNotifications::ToastProgressData const& data, hstring const& tag)
     {
-        throw hresult_not_implemented();
+        auto toastProgressResult = co_await UpdateToastProgressDataAsync(data, tag, L"");
```
danielayala94, line 159: Can you directly return `UpdateToastProgressDataAsync` in the `co_return`?

dev/ToastNotifications/NotificationProperties.cpp
danielayala94, line 133: Change initialization style to curly braces?

dev/ToastNotifications/ToastProgressData.cpp
```diff
@@ -6,42 +6,62 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
 {
     uint32_t ToastProgressData::SequenceNumber()
     {
-        throw hresult_not_implemented();
+        auto lock{ m_lock.lock_shared() };
+
+        return m_sequenceNumber;
     }
-    void ToastProgressData::SequenceNumber(uint32_t /* value */)
+    void ToastProgressData::SequenceNumber(uint32_t value)
```
danielayala94, line 13: Please change the parameter names to its actual proper names, i.e. `sequenceNumber`. Please apply to all functions.

dev/ToastNotifications/ToastProgressData.h
```diff
@@ -17,6 +17,14 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
         void ValueStringOverride(hstring const& value);
         hstring Status();
         void Status(hstring const& value);
+
+    private:
+        uint32_t m_sequenceNumber{};
+        hstring m_title;
+        double m_progressValue{};
+        hstring m_progressvalueString;
```
danielayala94, line 25: nit: 
```suggestion
        hstring m_progressValueString;
```

dev/ToastNotifications/ToastProgressDataABI.cpp
```diff
@@ -0,0 +1,69 @@
...
+STDMETHODIMP ToastProgressDataABI::get_SequenceNumber(_Out_ unsigned int* value) noexcept
+{
+    auto lock{ m_lock.lock_shared() };
+
+    *value = m_progressData.SequenceNumber();
+
+    return S_OK;
+}
+
+STDMETHODIMP ToastProgressDataABI::get_Title(_Out_ HSTRING* value) noexcept
```
danielayala94, line 34: Functions featuring `safe_make_unique_hstring` cannot be marked as `noexcept`. Unless you apply the noexcept try catch pattern.

test/TestApps/ToastNotificationsTestApp/main.cpp
```diff
@@ -435,6 +458,189 @@ std::string unitTestNameFromLaunchArguments(const winrt::ILaunchActivatedEventAr
     return unitTestName;
 }
 
+winrt::ToastProgressResult ProgressResultOperationHelper(winrt::IAsyncOperation<winrt::ToastProgressResult> progressResultOperation)
+{
+    if (progressResultOperation.wait_for(std::chrono::seconds(2)) != winrt::AsyncStatus::Completed)
+    {
+        progressResultOperation.Cancel();
+        return winrt::ToastProgressResult::Failed; // timed out or failed
```
danielayala94, line 466: When will we remove the `Failed` element? Do we have a bug for this or will be addressed in this PR?

test/TestApps/ToastNotificationsTestApp/main.cpp
```diff
@@ -435,6 +458,189 @@ std::string unitTestNameFromLaunchArguments(const winrt::ILaunchActivatedEventAr
...
+
+    return progressResult == winrt::ToastProgressResult::Succeeded;
+}
+
+bool VerifyUpdateToastProgressDataUsingValidTagAndValidGroup_Unpackaged()
+{
+    winrt::ToastAssets assets(L"ToastNotificationApp", winrt::Uri{ LR"(C:\Windows\System32\WindowsSecurityIcon.png)" });
+    auto activationInfo = winrt::ToastActivationInfo::CreateFromToastAssets(assets);
+
+    winrt::ToastNotificationManager::Default().RegisterActivator(activationInfo);
```
danielayala94, line 507: It's a little inefficient keep calling `Default()` again and again. Make it a local variable.

danielayala94: Please address it to all applicable tests.

danielayala94 (changes requested):

dev/ToastNotifications/ToastNotificationManager.cpp
sharath2727: No it doesn't like it that way.

dev/ToastNotifications/ToastNotificationManager.cpp
sharath2727: No this is needed what if the developer calls the UDK API without posting a notification, then it returns E_NOT_SET. We need to cover that scenario.

danielayala94: The condition, not the block. Of course, you need to return `NotificationNotFound`.

dev/ToastNotifications/ToastNotificationManager.cpp
```diff
@@ -131,14 +134,14 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
     // For other scenarios the API throws
     winrt::Windows::Foundation::IAsyncOperation<winrt::Microsoft::Windows::ToastNotifications::ToastProgressResult> ToastNotificationManager::UpdateToastProgressDataAsync(winrt::Microsoft::Windows::ToastNotifications::ToastProgressData const& data, hstring const& tag, hstring const& group)
     {
-        if ((tag == winrt::hstring(L"") && group == winrt::hstring(L"")) || (tag == winrt::hstring(L"") && group != winrt::hstring(L"")))
+        if (tag == winrt::hstring(L""))
         {
             co_return winrt::ToastProgressResult::NotificationNotFound;
```
danielayala94, line 139: I don't think this outcome is the right one. We can post toasts with no tag. But I do agree this scenario is not valid for updating progress data. This case seems more like a developer mistake. I would throw here as E_INVALIDARG.

danielayala94: I just noticed there was a mini thread discussion on it. Take my comment above as my two cents on it 😊.

dev/ToastNotifications/ToastNotificationManager.cpp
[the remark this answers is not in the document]
sharath2727: The API is expecting me to use co_await before I co_return.

dev/ToastNotifications/ToastNotificationManager.h
```diff
@@ -19,9 +19,9 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
 
         void ShowToast(winrt::Microsoft::Windows::ToastNotifications::ToastNotification const& toast);
 
-        winrt::Windows::Foundation::IAsyncOperation<winrt::Microsoft::Windows::ToastNotifications::ToastProgressResult> UpdateToastProgressDataAsync(winrt::Microsoft::Windows::ToastNotifications::ToastProgressData data, hstring tag, hstring group);
+        winrt::Windows::Foundation::IAsyncOperation<winrt::Microsoft::Windows::ToastNotifications::ToastProgressResult> UpdateToastProgressDataAsync(winrt::Microsoft::Windows::ToastNotifications::ToastProgressData const& data, hstring const& tag, hstring const& group);
```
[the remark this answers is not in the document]
sharath2727: I missed this completely. I experienced this with CreateChannelAsync API where remoteId was lost when we co_await as it was passed by reference. Thanks for pointing it.

dev/ToastNotifications/ToastNotificationManager.cpp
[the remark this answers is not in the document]
hulumane: Cant we just return  UpdateToastProgressDataAsync(data, tag, L""); then?

hulumane (approved):

dev/ToastNotifications/ToastNotificationManager.cpp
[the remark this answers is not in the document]
sharath2727: I started some research and landed on raymond's blog regarding the above: https://devblogs.microsoft.com/oldnewthing/?p=106160

I learnt more about this error and why we need to return the toastprogressresult. Sharing this for others too :)

loneursid (approved):

dev/ToastNotifications/ToastNotificationManager.cpp
```diff
@@ -130,23 +130,16 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
...
-    // API updates notification data when both tag and group are provided or when tag is provided.
-    // For other scenarios the API throws
-    winrt::Windows::Foundation::IAsyncOperation<winrt::Microsoft::Windows::ToastNotifications::ToastProgressResult> ToastNotificationManager::UpdateToastProgressDataAsync(winrt::Microsoft::Windows::ToastNotifications::ToastProgressData const& data, hstring const& tag, hstring const& group)
+    winrt::Windows::Foundation::IAsyncOperation<winrt::Microsoft::Windows::ToastNotifications::ToastProgressResult> ToastNotificationManager::UpdateToastProgressDataAsync(winrt::Microsoft::Windows::ToastNotifications::ToastProgressData const data, hstring const tag, hstring const group)
     {
-        if (tag == winrt::hstring(L""))
-        {
-            co_return winrt::ToastProgressResult::NotificationNotFound;
-        }
+        THROW_HR_IF_MSG(E_INVALIDARG, tag == winrt::hstring(L""), "Update operation isn't guarenteed to find a specific notification to replace correctly.");
```
danielayala94, line 135: nit: guaranteed

dev/ToastNotifications/ToastNotificationManager.cpp
```diff
@@ -155,13 +148,15 @@ namespace winrt::Microsoft::Windows::ToastNotifications::implementation
         {
             co_return winrt::ToastProgressResult::NotificationNotFound;
         }
+        else
```
danielayala94, line 151: Nit: no need for this else, just throw.

danielayala94 (approved):

PR #2063, opened 2022-02-03
Rename all instances of Toast in Public API layer to AppNotifications

hulumane:
We have received developer feedback that we need to infer Toasts as simply Notifications. This PR renames the Public API surface to reflect that change. More changes are incoming as this goes through the API review process.


WindowsAppRuntime.sln
```diff
@@ -289,6 +287,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Licensing", "Licensing", "{
 EndProject
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "Licensing", "dev\Licensing\Licensing.vcxitems", "{885A43FA-052D-4B0D-A2DC-13EE15796435}"
 EndProject
+<<<<<<< HEAD
```
loneursid, line 290: Remove this

loneursid:
Get a compile error in: test/TestApps/PushNotificationsDemoApp/pch.h. Can't find: winrt/Microsoft.Windows.ToastNotifications.h

loneursid (approved):
Re-ran all the tests for push and toast. Got a few failures for push on the first run but everything was green on the second one.

PR #2065, opened 2022-02-04
Revert "Update the Windows SDK version for RestartAgent"

loneursid:
Reverts microsoft/WindowsAppSDK#2043

I'm reverting this PR as it seems that it is causing the nightly build to fail. I am not sure I fully understand why that it the case as the other projects for the WindowsAppSDK seem to be happy with 10.0 (latest installed verion)

I find it strange that this project is specifically targeting v10.0.20348.0 but I assume that we're relying on a bug fix that wasn't present before. Then the question becomes why is WindowsTargetPlatformMinVersion set to 10.0.17134.0.


kythant (approved):

DrusTheAxe (approved):

loneursid:
On hold for now, as this may not be what broke the build originally

loneursid:
Closing for now as this doesn't seem to be the root cause of the build break.

PR #2074, opened 2022-02-08
Adding support for remove toasts async

loneursid:


test/TestApps/ToastNotificationsTestApp/main.cpp
```diff
@@ -510,9 +510,195 @@ bool VerifyShowToast_Unpackaged()
...
+    winrt::ToastNotification toast2{ CreateToastNotification() };
+    toast2.Tag(L"tag2");
+    toast2.Group(L"Shared group");
+    toastNotificationManager.ShowToast(toast2);
+
+    winrt::ToastNotification toast3{ CreateToastNotification() };
+    toast3.Tag(L"tag3");
+    toast3.Group(L"Shared group");
+    toastNotificationManager.ShowToast(toast3);
+
```
robertzhou-wpn, line 682: looks like all toasts are in the same group. maybe we can add another toast with a different group and verify it will not be removed.

test/TestApps/ToastNotificationsTestApp/main.cpp
```diff
@@ -510,9 +510,195 @@ bool VerifyShowToast_Unpackaged()
...
+    winrt::ToastNotification toast2{ CreateToastNotification() };
+    toast2.Tag(L"Shared tag");
+    toast2.Group(L"group2");
+    toastNotificationManager.ShowToast(toast2);
+
+    winrt::ToastNotification toast3{ CreateToastNotification() };
+    toast3.Tag(L"Shared tag");
+    toast3.Group(L"group3");
+    toastNotificationManager.ShowToast(toast3);
+
```
robertzhou-wpn, line 606: looks like all 3 test toasts have the same tag, maybe we can add one more toast with a different tag and verify that it will not be removed.

PR #2084, opened 2022-02-09
Move CSWinRT to Toolset Dependencies in Version.Details.xml

kythant:
This allows us to not have an exclude list in the ConvertToVersionDetails powershell script.

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -1,6 +1,12 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Dependencies>
   <ProductDependencies>
+    <Dependency Name="Microsoft.WindowsAppSDK.AppLicensingInternal.TransportPackage" Version="1.0.0-preview1.20211111.12">
+      <Uri>https://dev.azure.com/microsoft/ProjectReunion/_git/ProjectReunionClosed</Uri>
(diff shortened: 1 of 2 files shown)


eng/Version.Details.xml
```diff
@@ -1,6 +1,12 @@
...
   <ProductDependencies>
+    <Dependency Name="Microsoft.WindowsAppSDK.AppLicensingInternal.TransportPackage" Version="1.0.0-preview1.20211111.12">
+      <Uri>https://dev.azure.com/microsoft/ProjectReunion/_git/ProjectReunionClosed</Uri>
+      <Sha>a3b6cd2d041c342d4a5ac2ed0c90c30e9ed08f15</Sha>
+    </Dependency>
+  </ProductDependencies>
+  <ToolsetDependencies>
     <Dependency Name="Microsoft.Windows.CsWinRT" Version="1.5.0-prerelease.220124.4">
       <Uri>https://github.com/microsoft/CsWinRT</Uri>
       <Sha>c90d2e33c650b65377a3eb046f1f449d56e2ab7d</Sha>
```
manodasanW, line 12: Should CsWinRT.Dependency.DotNetCoreSdk**n** be CsWinRT.Dependency.DotNetCoreSdk below?

kythant: oh wow that's a typo. Good catch

manodasanW (approved):

PR #2085, opened 2022-02-09
Added apicontract

DrusTheAxe:
Adds APIcontract for WinAppSDK

Adds Microsoft.Foundation.winmd to the Transport Package


aeloros (approved):

dev/Microsoft.Foundation/Microsoft.Foundation.ApiContract.idl
```diff
@@ -0,0 +1,11 @@
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+
+namespace Microsoft.Foundation
+{
+    // Contract version corresponds to the Windows App SDK release version (1.0, 1.1, ...). MIDL
+    // interprets this as MAJOR.MINOR, not as a float value. For example, 1.10 is major version
+    // 1, minor version 10, which is _after_ 1.9. Note also that "1" and "1.0" are equivalent.
+
+    [contractversion(1.1)]
+    apicontract WindowsAppSDKContract{};
```
jonwis, line 10: The UniversalApiContract in the OS has grown too large to manage effectively. Are we thinking of a single contract per contributing repo (ie: WinUI3 has its own WinUIContract, DWrite has a DWriteContract, etc.)?

DrusTheAxe: > The UniversalApiContract in the OS has grown too large to manage effectively. Are we thinking of a single contract per contributing repo (ie: WinUI3 has its own WinUIContract, DWrite has a DWriteContract, etc.)?

I'd say 1 per _feature_. That may be 1:1 for some repos, or 1:N for others. Foundation comes to mind, but so too WinUI (a huge API, all of it's 1 contract?).

We're likely to have >1 over time when other existing code joins WinAppSDK if they've already defined apicontract(s).

But regardless of the fine print, I'm strongly in favor of N contracts (not 1 like UniversalApiContract).

We should identify someone to write up a proposal (Issue) to discuss further. Want to volunteer? :P I'll follow up offline to get that ball rolling...

DrusTheAxe: BTW this PR was created as an outgrowth of some conversations. It's paused until an owner's identified and proposal/spec is available and approved. This can be modified as needed once that's decided.

dev/Microsoft.Foundation/Microsoft.Foundation.vcxitems.filters
```diff
@@ -0,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup>
+    <Midl Include="$(MSBuildThisFileDirectory)Microsoft.Foundation.ApiContractidl" />
```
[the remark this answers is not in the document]
DrusTheAxe: See latest commits

dev/Microsoft.Foundation/Microsoft.Foundation.ApiContract.idl
DefaultRyan: I hadn't considered splitting up into multiple contracts. It will add more complexity both for consumers and for our own engineering. What's the benefit of doing this?

BenJKuhn: Contracts are essentially a unit of distribution and a unit of independence for components to evolve at different speeds. The universal contract was meant to be the "common base" for all universal APIs, though its over-inclusive and as a result some SKUs have inert APIs just to satisfy the contract's expected API surface, even though they're unused. With a small exception of the installer in the main repo, one repo == one transport package, including for dwrite, WinUI3, and closed source. Composition and UDK components are the one instance where one repo contains two components that are delivered in two different transport packages. This also is a bit of an anomaly that I hope we can sort out over time.

I've sometimes used the terms component as a representation of a repo that produces a transport package. Essentially, App SDK is reaching a critical mass where it would make sense to have a well-defined component list & layer map. At that point, one contract per component makes sense, and although a transport package might have more than one logical component, and one component might be comprised of more than one binary or file. (e.g. foundation transport package), the components in the package should be clearly delineated somehow.

```
   +-------------+
   | Aggregation |
   +-----+-------+
         |
         | +---------+
         +>| Package |
           +-+-------+
             |
             | +------------------------+
             +>| Component <> Contract  |
               +------------------------+
```

One advantage of splitting along components is that if, say, dwritecore ships through multiple channels, they can still keep their contracts moving forward as a linear series of supersets in each version across any or all of their distribution channels. Otherwise, they would need to introduce versioning complexity to ensure they didn't produce conflicting representations under the same contract & version. If dwrite has its own contract, they can bump it every time they ship, regardless of which channel it is, and not need to coordinate with all the other App SDK work.

Ryan, it's going to be a little bit of a pain, but I think it makes sense to invest a little time in getting this right up front. Let's spend a little time this week enumerating the components and seeing if we can apply contracts in all the right places. I don't think it'll be a huge amount of work but might be a bit tedious.

In terms of shipping the winmd's in files organized by namespace, that's a requirement of metadata lookup: the code references types by name, so for fast lookup we search filenames based on namespace. The contracts don't correlate to type names and are mutable (contract migration). The metadata, in turn, is used by various dev tools, and at runtime for metadata-based marshalling (in practice, mostly for delegates). For App SDK 1.0 we avoided the metadata merge process that happens in Windows by ensuring that each component had well-defined namespaces, though in practice namespaces can span contracts.

BenJKuhn: For tracking, https://task.ms/38115237

DrusTheAxe:
Obsolete and superceded by https://github.com/microsoft/WindowsAppSDK/pull/2110

PR #2087, opened 2022-02-10
Add changes for the new activation pattern

hulumane:


pmpurifoy:
No longer going with this pattern.

PR #2094, opened 2022-02-11
Retrieve Push/Toast Activator from registry

pmpurifoy:
- Retrieve clsid from the registry and verify clsids passed into Register functions for Push/AppNotifictionManager


dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -78,6 +85,13 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
             THROW_IF_FAILED(notificationPlatform->AddToastRegistrationMapping(processName.get(), toastAppId.c_str()));
         }
 
+        if (AppModel::Identity::IsPackagedProcess())
```
hulumane, line 88: REcommend: Create a new Register() API in the idl and reimplement this function in that. You don't have to delete this until that implementation is complete.

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -78,6 +85,13 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
+        {
+            winrt::guid registeredClsid{ GUID_NULL };
+            THROW_IF_FAILED(PushNotificationHelpers::CheckComServerClsid(registeredClsid, expectedAppServerArgs.data()));
+            THROW_HR_IF(E_INVALIDARG, registeredClsid != details.TaskClsid());
```
hulumane, line 92: This won't be needed if you have a pure Register() function.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -291,6 +292,10 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
                 GUID taskClsid = details.TaskClsid();
                 THROW_HR_IF(E_INVALIDARG, taskClsid == GUID_NULL);
 
+                winrt::guid registeredClsid{ GUID_NULL };
```
hulumane, line 295: Create a pure RegisterActivator() API. It will greatly simplify this logic!

dev/PushNotifications/PushNotificationUtility.h
```diff
@@ -123,4 +127,126 @@ namespace winrt::Microsoft::Windows::PushNotifications::Helpers
...
+
+        return S_OK;
+    }
+    CATCH_RETURN()
+
+    inline HRESULT GetExeNameFromPath(std::wstring& exePath) noexcept try
+    {
+        size_t pos{ exePath.rfind(L"\\") };
+        THROW_HR_IF(E_UNEXPECTED, pos == std::wstring::npos);
+        exePath = exePath.substr(pos + 1); // One after the delimiter
```
hulumane, line 148: Make sure you filter out .exe from this string.

dev/PushNotifications/PushNotificationUtility.h
```diff
@@ -123,4 +127,126 @@ namespace winrt::Microsoft::Windows::PushNotifications::Helpers
...
+        THROW_IF_FAILED(RegOpenKeyEx(HKEY_LOCAL_MACHINE, clsidPath.c_str(), 0, KEY_READ, &hKey));
+
+        DWORD comServerGuids{};
+        DWORD maxSubKeyLen;
+        THROW_IF_FAILED(::RegQueryInfoKey(
+            hKey.get(),
+            nullptr,    // No user-defined class
+            nullptr,    // No user-defined class size
+            nullptr,    // Reserved
+            &comServerGuids,
```
hulumane, line 173: rename to comServerGuidCount

dev/PushNotifications/PushNotificationUtility.h
```diff
@@ -123,4 +127,126 @@ namespace winrt::Microsoft::Windows::PushNotifications::Helpers
...
+            std::wstring exeString{ exeBuffer };
+            THROW_IF_FAILED(GetExeNameFromPath(exeString));
+
+            wil::unique_cotaskmem_string exePath;
+            THROW_IF_FAILED(wil::GetModuleFileNameExW(GetCurrentProcess(), nullptr, exePath));
+
+            std::wstring exeToCheck{ exePath.get() };
+            THROW_IF_FAILED(GetExeNameFromPath(exeToCheck));
+
+            if (argumentString == argumentToCheck && exeString == exeToCheck)
```
hulumane, line 242: Use THROW_HR_IF to filter out errors as a pattern instead of returning S_OK in a conditional prematurely.

As a general rule, return S_OK only at the end of a function call.

dev/PushNotifications/PushNotificationUtility.h
```diff
@@ -123,4 +127,126 @@ namespace winrt::Microsoft::Windows::PushNotifications::Helpers
...
+            nullptr,    // No subkey class length
+            nullptr, // No values we want to get
+            nullptr,
+            nullptr,    // No max value length
+            nullptr,    // No security descriptor
+            nullptr     // No last write time
+        ));
+
+        maxSubKeyLen++; // Account for the null character
+
```
hulumane, line 184: THROW_HR_IF_MSG(E_FAIL, comServerGuidCount < 1, "No COM servers are registered for this app);

dev/PushNotifications/PushNotificationUtility.h
hulumane: if (argumentString == argumentToCheck)
{
      THROW_HR_IF_MSG(E_FAIL, exeString != exeToCheck, "Caller RegistrationProcess is not the same as COM Server!");
      found = true;
      break;
}

dev/PushNotifications/PushNotificationUtility.h
```diff
@@ -123,4 +127,126 @@ namespace winrt::Microsoft::Windows::PushNotifications::Helpers
...
+            THROW_IF_FAILED(GetExeNameFromPath(exeToCheck));
+
+            if (argumentString == argumentToCheck && exeString == exeToCheck)
+            {
+                std::wstring storedComActivatorString{ comServerGuid.data() };
+                comServerClsid = winrt::guid(storedComActivatorString.substr(1, storedComActivatorString.size() - 2));
+                return S_OK;
+            }
+        }
+        return E_NOT_SET;
```
hulumane, line 249: End the API call like this to make things clearer:

THROW_HR_IF(E_NOT_SET, found != true);
return S_OK;

dev/AppNotifications/AppNotificationManager.cpp
sharath2727: Apologies I have already Modified the existing API and got rid of the arguments. Hopefully it is perfect :)

dev/PushNotifications/PushNotificationManager.cpp
sharath2727: As discussed the activators will be removed in the upcoming PR. Thanks.

dev/PushNotifications/PushNotificationUtility.h
```diff
@@ -135,4 +139,135 @@ namespace winrt::Microsoft::Windows::PushNotifications::Helpers
     {
         return AppModel::Identity::IsPackagedProcess() && IsBackgroundTaskBuilderAvailable();
     }
+
+    inline HRESULT GetPackageFullName(wil::unique_cotaskmem_string& packagedFullName) noexcept try
+    {
+        WCHAR packageFullName[PACKAGE_FULL_NAME_MAX_LENGTH + 1]{};
+        UINT32 packageFullNameLength = ARRAYSIZE(packageFullName);
```
pmpurifoy, line 146: {} ctor

dev/PushNotifications/PushNotificationUtility.h
```diff
@@ -135,4 +139,135 @@ namespace winrt::Microsoft::Windows::PushNotifications::Helpers
...
+                &serverIdSize));
+
+            // Path: HKLM\SOFTWARE\Classes\PackagedCom\Package\{PackageFullName}\Server\{serverId}
+            std::wstring serverPath{ packagedRegistryPath + LR"(\Server\)" + std::to_wstring(serverId) };
+            WCHAR argumentsBuffer[MAX_VALUE_NAME];
+            DWORD argumentsBufferLength = sizeof(argumentsBuffer);
+            THROW_IF_FAILED(RegGetValueW(
+                HKEY_LOCAL_MACHINE,
+                serverPath.c_str(),
+                L"Arguments",
```
pmpurifoy, line 231: @hulumane Should these be constants in this file?

dev/PushNotifications/PushNotificationUtility.h
```diff
@@ -135,4 +139,135 @@ namespace winrt::Microsoft::Windows::PushNotifications::Helpers
...
+                &argumentsBufferLength));
+
+            std::wstring argumentString{ argumentsBuffer };
+
+            WCHAR exeBuffer[MAX_VALUE_NAME];
+            DWORD exeBufferLength = sizeof(exeBuffer);
+            THROW_IF_FAILED(RegGetValueW(
+                HKEY_LOCAL_MACHINE,
+                serverPath.c_str(),
+                L"Executable",
```
pmpurifoy, line 244: Same question from above.

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -69,7 +74,12 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
             {
                 THROW_IF_FAILED(PushNotifications_RegisterFullTrustApplication(toastAppId.c_str(), GUID_NULL));
 
-                storedComActivatorString = RegisterComActivatorGuidAndAssets(details);
+
+                // TODO: Add permanent displayname and iconuri.
+                wil::unique_cotaskmem_string processName;
+                THROW_IF_FAILED(GetCurrentProcessPath(processName));
+
+                storedComActivatorString = RegisterComActivatorGuidAndAssets(processName.get(), winrt::Uri{ LR"(C:\Windows\System32\WindowsSecurityIcon.png)" });
```
hulumane, line 82: REmove hardcoded PNG please. No hardcoding anywhere in our code base

hulumane: Its likely null that you want to pass through since @danielayala94 is doing the work to retrieve that

hulumane: And when you register the asset, check for null and only store in registry if it is non-null or non-empty.

hulumane: Also @danielayala94  You may want to baseline against this PR

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -69,7 +74,12 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
             {
                 THROW_IF_FAILED(PushNotifications_RegisterFullTrustApplication(toastAppId.c_str(), GUID_NULL));
 
-                storedComActivatorString = RegisterComActivatorGuidAndAssets(details);
+
+                // TODO: Add permanent displayname and iconuri.
+                wil::unique_cotaskmem_string processName;
```
hulumane, line 79: Put this outside the if-condition. You are reunsing it below too and the calls are redundant.

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -80,8 +90,11 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
             THROW_IF_FAILED(notificationPlatform->AddToastRegistrationMapping(processName.get(), toastAppId.c_str()));
         }
 
+        winrt::guid registeredClsid{ GUID_NULL };
+        THROW_IF_FAILED(PushNotificationHelpers::GetComRegistrationFromRegistry(expectedAppServerArgs.data(), registeredClsid));
```
hulumane, line 94: Why is this code path getting executed for unpackaged apps? It wouldnt work right?

hulumane: The if checks aren't accurate imho.

dev/AppNotifications/AppNotificationManager.cpp
danielayala94: Removing the hardcoded properties was on me, I have fixed it as part of the displayName/icon inference work.

WindowsAppRuntime.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 17
-VisualStudioVersion = 17.0.31919.166
+# Visual Studio Version 16
```
robertzhou-wpn, line 3: why do we need to change the visual studio version?

danielayala94: I don't think we do this in purpose. I bet this depends on the version the latest person to commit to this PR is using.

dev/AppNotifications/AppNotificationActivationInfo.cpp
```diff
@@ -1,41 +0,0 @@
...
-namespace winrt::Microsoft::Windows::AppNotifications::implementation
-{
-    AppNotificationActivationInfo::AppNotificationActivationInfo(winrt::guid const& taskClsid)
-    {
-        THROW_HR_IF_MSG(E_ILLEGAL_METHOD_CALL, !AppModel::Identity::IsPackagedProcess(), "Not applicable for unpackaged applications");
-        THROW_HR_IF(E_INVALIDARG, (taskClsid == winrt::guid(GUID_NULL)));
-        m_taskClsid = taskClsid;
-    }
-
-    AppNotificationActivationInfo::AppNotificationActivationInfo(hstring const& displayName, winrt::Windows::Foundation::Uri const& iconUri)
```
hulumane, line 14: Awesome! Gone forever :)

dev/AppNotifications/AppNotificationActivationInfo.h
```diff
@@ -1,27 +0,0 @@
...
-namespace winrt::Microsoft::Windows::AppNotifications::implementation
-{
-    struct AppNotificationActivationInfo : AppNotificationActivationInfoT<AppNotificationActivationInfo>
-    {
-        AppNotificationActivationInfo() = default;
-
-        AppNotificationActivationInfo(winrt::guid const& taskClsid);
-        AppNotificationActivationInfo(hstring const& displayName, winrt::Windows::Foundation::Uri const& iconUri);
-        winrt::guid TaskClsid();
-        hstring DisplayName();
```
hulumane, line 13: Yay!

dev/AppNotifications/AppNotificationManager.cpp
danielayala94: This may be solved if GetComRegistrationFromRegistry returns a GUID instead of HRESULT. I would take this as a bug fix (if we change this, then we need to make changes in PushNotificationManager as well).

dev/PushNotifications/PushNotificationUtility.h
```diff
@@ -135,4 +143,135 @@ namespace winrt::Microsoft::Windows::PushNotifications::Helpers
...
+            THROW_IF_FAILED(RegGetValueW(
+                HKEY_LOCAL_MACHINE,
+                serverPath.c_str(),
+                c_argumentCommandString.c_str(),
+                RRF_RT_REG_SZ,
+                nullptr /* pdwType */,
+                &argumentsBuffer,
+                &argumentsBufferLength));
+
+            std::wstring argumentString{ argumentsBuffer };
```
danielayala94, line 241: Nit: Please move closer to its usage.

dev/PushNotifications/PushNotificationUtility.h
```diff
@@ -135,4 +143,135 @@ namespace winrt::Microsoft::Windows::PushNotifications::Helpers
...
+                &exeBufferLength));
+
+            std::wstring exeString{ exeBuffer };
+            THROW_IF_FAILED(GetExeNameFromPath(exeString));
+
+            wil::unique_cotaskmem_string exePath;
+            THROW_IF_FAILED(wil::GetModuleFileNameExW(GetCurrentProcess(), nullptr, exePath));
+
+            std::wstring exeToCheck{ exePath.get() };
+            THROW_IF_FAILED(GetExeNameFromPath(exeToCheck));
```
danielayala94, line 261: All of this should be inside the if block below.

dev/PushNotifications/PushNotificationUtility.h
```diff
@@ -135,4 +143,135 @@ namespace winrt::Microsoft::Windows::PushNotifications::Helpers
...
+            if (argumentString == argumentToCheck)
+            {
+                THROW_HR_IF_MSG(E_FAIL, exeString != exeToCheck, "Caller RegistrationProcess is not the same as COM Server!");
+                std::wstring storedComActivatorString{ comServerGuid.data() };
+                comServerClsid = winrt::guid(storedComActivatorString.substr(1, storedComActivatorString.size() - 2));
+                found = true;
+                break;
+            }
+        }
+        THROW_HR_IF(E_NOT_SET, found != true);
```
danielayala94, line 272: Nit:

```suggestion
        THROW_HR_IF(E_NOT_SET, !found);
```

test/TestApps/PushNotificationsTestApp/main.cpp
```diff
@@ -201,8 +162,7 @@ bool MultipleRegisterActivatorTest()
         if(PushNotificationManager::Default().IsActivatorSupported(PushNotificationRegistrationActivators::PushTrigger | PushNotificationRegistrationActivators::ComActivator))
         {
             PushNotificationActivationInfo info(
-                PushNotificationRegistrationActivators::PushTrigger | PushNotificationRegistrationActivators::ComActivator,
-                c_fakeComServerId); // Fake clsid to test multiple activators
+                PushNotificationRegistrationActivators::PushTrigger | PushNotificationRegistrationActivators::ComActivator); // Fake clsid to test multiple activators
```
danielayala94, line 165: `Fake clsid to test multiple activators`

What does this mean?

test/TestApps/ToastNotificationsDemoApp/main.cpp
```diff
@@ -196,11 +238,15 @@ int main()
...
+    std::wcout << L"Post a Toast..." << std::endl;
+    PostToastHelper(L"Tag", L"Group");
+    std::wcout << L"Done.\n\n";
+
     std::wcout << L"Press enter to exit the app.\n\n";
     std::cin.ignore();
 
     // If you want to stop receiving AppNotifications for the app
-    /* AppNotificationManager.UnregisterActivator(); */
+    appNotificationManager.Unregister();
```
danielayala94, line 249: Hmmm I think I forgot to comment this back. I'm going to revert this.

test/TestApps/ToastNotificationsTestApp/main.cpp
```diff
@@ -239,101 +239,45 @@ bool PostToastHelper(std::wstring const& tag, std::wstring const& group)
     return true;
 }
 
-bool VerifyFailedRegisterActivatorUsingNullClsid()
+bool VerifyRegisterActivatorandUnRegisterActivator()
```
danielayala94, line 242: Nit: test name

```suggestion
bool VerifyRegisterActivatorAndUnregisterActivator()
```

test/TestApps/ToastNotificationsTestApp/main.cpp
```diff
@@ -239,101 +239,45 @@ bool PostToastHelper(std::wstring const& tag, std::wstring const& group)
...
-        return winrt::to_hresult() == E_INVALIDARG;
+        return false;
     }
-
-    return false;
+    return true;
 }
 
-bool VerifyRegisterActivatorandUnRegisterActivatorUsingClsid()
+bool VerifyRegisterActivatorandUnRegisterActivatorUsingAssets_Unpackaged()
```
danielayala94, line 259: We need to thoroughly review the caps in test names.

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -204,31 +210,79 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
         &hKey,
         nullptr /* lpdwDisposition */));
 
+    // Retrieve DisplayName and IconUri
+    // 1. We call into Shell APIs, to get both from the app taskbar shortcut.
+    // 2. If 1 doesn't work (i.e. possibly because the developer didn't specify any DisplayName/Icon,
+    //    we take the file name as app name, and let Shell to put a default icon
```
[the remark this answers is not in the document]
danielayala94: I'll address this myself.

dev/AppNotifications/AppNotificationManager.cpp
hulumane: Don't hide functionality in an API please as discussed. Do the IsPackaged check here. Not in the API

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -83,11 +91,6 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
         GetWaitHandleForArgs().create();
```
hulumane, line 91: Race condition here. This needs to be done before calling into CoRegisterClassObject. Otherwise we can get an incoming payload before we create the event.

dev/AppNotifications/AppNotificationManager.cpp
hulumane: @sharath2727 said that he will fix this now. If it is really a big change, then sure we can move it to a bug.

dev/AppNotifications/AppNotificationUtility.cpp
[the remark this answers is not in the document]
hulumane: I would throw as per our offline conversation if icon isn't specified by dev in either the shortcut or in the Shell API call.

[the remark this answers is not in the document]
hulumane: For DisplayNmame, there isn't a need to throw since we can fall back on the processName instead.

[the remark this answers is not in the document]
hulumane: Also check for icon filetype below and throw if the fileformat isn't supported.

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -204,31 +210,79 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
...
+        displayName = displayName.substr(lastBackslashPosition + 1); // One after the delimiter
+
+        displayName.erase(displayName.find_first_of(L".")); // Remove file extension
+    }
+
+    // Retrieve the icon
+    std::wstring iconFilePath{};
+
+    wil::unique_prop_variant propVariantIcon;
+    THROW_IF_FAILED(propertyStore->GetValue(PKEY_AppUserModel_RelaunchIconResource, &propVariantIcon));
```
hulumane, line 247: THROW_IF_FAILED_MSG(propertyStore->GetValue(PKEY_AppUserModel_RelaunchIconResource, &propVariantIcon), "Couldn't find the Icon Uri Path! Please specifiy it in API blah blah")

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -204,31 +210,79 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
...
+    //    we take the file name as app name, and let Shell to put a default icon
+    winrt::com_ptr<IPropertyStore> propertyStore;
+    wil::unique_hwnd hWindow{ GetConsoleWindow() };
+    THROW_IF_FAILED(SHGetPropertyStoreForWindow(hWindow.get(), IID_PPV_ARGS(propertyStore.put())));
+
+    // Retrieve the display name
+    std::wstring displayName{};
+
+    wil::unique_prop_variant propVariantDisplayName;
+    THROW_IF_FAILED(propertyStore->GetValue(PKEY_AppUserModel_RelaunchDisplayNameResource, &propVariantDisplayName));
```
hulumane, line 225: Don't throw if this isn't found. Instead default to the ProcessName.

hulumane: The only manadatory property is IconUri.

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -204,31 +210,79 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
...
+
+    wil::unique_prop_variant propVariantIcon;
+    THROW_IF_FAILED(propertyStore->GetValue(PKEY_AppUserModel_RelaunchIconResource, &propVariantIcon));
+    iconFilePath = propVariantIcon.pwszVal != nullptr ? propVariantIcon.pwszVal : L"";
+
+    // Icon filepaths from Shell APIs have this format: <filepath>,-<index>,
+    // since .ico files can have multiple icons in the same file.
+    // NotificationController doesn't seem to support such format, so let it take the first icon by default.
+    auto iterator = iconFilePath.find_first_of(L",");
+    if (iterator != std::wstring::npos)
```
hulumane, line 254: THROWHR_IF(E_UNEXPECTED, iterator == std::wstring::npos)

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -204,31 +210,79 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
...
+
+        displayName.erase(displayName.find_first_of(L".")); // Remove file extension
+    }
+
+    // Retrieve the icon
+    std::wstring iconFilePath{};
+
+    wil::unique_prop_variant propVariantIcon;
+    THROW_IF_FAILED(propertyStore->GetValue(PKEY_AppUserModel_RelaunchIconResource, &propVariantIcon));
+    iconFilePath = propVariantIcon.pwszVal != nullptr ? propVariantIcon.pwszVal : L"";
```
hulumane, line 248: Do not assign empty string in any case. Please throw if it is null

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -204,31 +210,79 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
...
         THROW_IF_FAILED(CoCreateGuid(&comActivatorGuid));
 
-        // StringFromCLSID returns GUID String with braces.
+        // StringFromCLSID returns GUID String with braces
         wil::unique_cotaskmem_string comActivatorGuidString;
         THROW_IF_FAILED(StringFromCLSID(comActivatorGuid, &comActivatorGuidString));
 
         std::wstring notificationAppId{ RetrieveNotificationAppId() };
-        RegisterAssets(notificationAppId, details.DisplayName(), details.IconUri(), comActivatorGuidString);
+        RegisterAssets(notificationAppId, comActivatorGuidString);
```
hulumane, line 285: This code has a serious bug. If the Display assets are set once, subsequent calls to Register()_ for unpackaged apps can never update the display assets again!

dev/AppNotifications/AppNotifications.idl
```diff
@@ -5,26 +5,6 @@ import "..\AppLifecycle\AppLifecycle.idl";
...
 namespace Microsoft.Windows.AppNotifications
 {
-    // The Registration Info for Packaged and Unpackaged Notification Activations
-    [experimental]
-    runtimeclass AppNotificationActivationInfo
-    {
-        // Initialize using a manifest defined COM Activator Id. Only applicable to Packaged Win32 applications
-        AppNotificationActivationInfo(Guid taskClsid);
-
-        // Initialize using Notification Assets. Only applicable to Unpackaged Win32 applications which need to specify their own assets like DisplayName and Icon.
```
hulumane, line 15: Excellent! :)

dev/AppNotifications/AppNotificationUtility.cpp
danielayala94: Ah, that's true. Good catch!

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -337,7 +337,7 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
 
                 GetWaitHandleForArgs().create();
                 GetComServerClsid() = taskClsid;
-
```
hulumane, line 340: Undo all changes to this file please

dev/AppNotifications/AppNotificationUtility.cpp
danielayala94: `GetValue` returns S_OK even if there is no element there. But we can check the propVariant to verify there is no element: https://docs.microsoft.com/en-us/previous-versions/windows/desktop/legacy/bb761473(v=vs.85)#remarks

dev/PushNotifications/PushNotificationUtility.h
```diff
@@ -135,4 +143,135 @@ namespace winrt::Microsoft::Windows::PushNotifications::Helpers
...
+        THROW_IF_FAILED(GetCurrentPackageFullName(&packageFullNameLength, packageFullName));
+
+        packagedFullName = wil::make_cotaskmem_string(packageFullName);
+        THROW_IF_NULL_ALLOC(packagedFullName);
+
+        return S_OK;
+    }
+    CATCH_RETURN()
+
+    inline HRESULT GetExeNameFromPath(std::wstring& exePath) noexcept try
```
hulumane, line 160: This is a weird API for 2 reasons: It takes an already existing construct in a path format as an argument and modifies it to a pure filename. Its a bad pattern.

Instead try
std::string GetExeNameFromPath(std::wstring& const exepath)

dev/PushNotifications/PushNotificationUtility.h
```diff
@@ -135,4 +143,135 @@ namespace winrt::Microsoft::Windows::PushNotifications::Helpers
...
+                std::wstring exeString{ exeBuffer };
+                THROW_IF_FAILED(GetExeNameFromPath(exeString));
+
+                wil::unique_cotaskmem_string exePath;
+                THROW_IF_FAILED(wil::GetModuleFileNameExW(GetCurrentProcess(), nullptr, exePath));
+
+                std::wstring exeToCheck{ exePath.get() };
+                THROW_IF_FAILED(GetExeNameFromPath(exeToCheck));
+
+                THROW_HR_IF_MSG(E_FAIL, exeString != exeToCheck, "Caller RegistrationProcess is not the same as COM Server!");
```
hulumane, line 263: Registration Caller Process is not the same as manifest defined COM server!

dev/AppNotifications/AppNotificationUtility.cpp
hulumane: Thats a weird API. But yes, please check propVariant in that case using recommended pattern in msdn.

dev/AppNotifications/AppNotificationUtility.cpp
danielayala94: Not necessarily. If the dev provides a plain png filepath without an index, it is still a valid case.

test/TestApps/ToastNotificationsDemoApp/main.cpp
```diff
@@ -196,11 +238,15 @@ int main()
...
     winrt::PushNotificationChannel channel{ RequestChannel() };
 
+    std::wcout << L"Post a Toast..." << std::endl;
+    PostToastHelper(L"Tag", L"Group");
+    std::wcout << L"Done.\n\n";
+
     std::wcout << L"Press enter to exit the app.\n\n";
     std::cin.ignore();
 
     // If you want to stop receiving AppNotifications for the app
```
hulumane, line 248: Update comment here: "Call Unregister so that COM can launch a new process for ToastInvokes after we terminate this process.

test/TestApps/ToastNotificationsDemoApp/main.cpp
```diff
@@ -136,24 +193,9 @@ int main()
...
-    {
-        std::wcout << L"Calling AppNotificationActivationInfo with app assets...\n";
-        activationInfo = winrt::AppNotificationActivationInfo(L"AppNotificationApp", winrt::Uri{ LR"(C:\Windows\System32\WindowsSecurityIcon.png)" });
-        std::wcout << L"Done.\n\n";
-    }
-
     // Registering app for activation
-    std::wcout << L"Calling AppNotificationActivationInfo::Register(activationInfo)...\n";
-    appNotificationManager.Register(activationInfo);
+    std::wcout << L"Calling AppNotificationManager::Register()...\n";
```
hulumane, line 197: Shouldn't FG activator be called before REgister @pmpurifoy ?

pmpurifoy: Yes

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -204,37 +210,95 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
...
+
+    // Retrieve the icon
+    std::wstring iconFilePath{};
 
-    std::wstring wideStringClsid{ clsid.get() };
-    RegisterValue(hKey, L"CustomActivator", reinterpret_cast<const BYTE*>(wideStringClsid.c_str()), REG_SZ, wideStringClsid.size() * sizeof(wchar_t));
+    wil::unique_prop_variant propVariantIcon;
+    // Throw in case of failure, since Icon is mandatory and we don't have a fallback!
+    THROW_IF_FAILED(propertyStore->GetValue(PKEY_AppUserModel_RelaunchIconResource, &propVariantIcon));
+    THROW_HR_IF_MSG(E_UNEXPECTED, propVariantDisplayName.vt == VT_EMPTY, "You must specify an app icon before calling Register().");
```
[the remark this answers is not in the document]
danielayala94: Yes, it is a mistake. Thanks for catching this!

hulumane (approved):

PR #2104, opened 2022-02-15
Update MSIX package names' spec

DrusTheAxe:
Update MSIX package names' spec to match actual 1.0 behavior (missed updating spec when we updated code for 1.0-stable)

### DIFF
--- a/specs/Deployment/MSIXPackages.md
+++ b/specs/Deployment/MSIXPackages.md
@@ -7,8 +7,9 @@
     - [2.1.4. Dynamic Dependency Lifetime Manager (DDLM)](#214-dynamic-dependency-lifetime-manager-ddlm)
 - [3. Package Naming](#3-package-naming)
   - [3.1. Package Naming - SubName](#31-package-naming---subname)
-  - [3.2. Package Naming - Singleton](#32-package-naming---singleton)
-  - [3.3. Package Naming - DDLM](#33-package-naming---ddlm)
+  - [3.2. Package Naming - Singleton](#32-package-naming---main)
+  - [3.3. Package Naming - Singleton](#33-package-naming---singleton)
+  - [3.4. Package Naming - DDLM](#34-package-naming---ddlm)
 - [4. Package Versioning](#4-package-versioning)
 
 # 1. Background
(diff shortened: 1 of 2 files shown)


specs/Deployment/MSIXPackages.md
```diff
@@ -159,11 +160,24 @@ The following SubName values are used:
...
+The Main package follows a different naming scheme
+
+* Name = MicrosoftCcorporationII.WinAppRuntime.Main.\<ReleaseMajorMinor\>[-ShortVersionTag]
+
+where
+
+* ReleaseMajorMinor = project release major.minor version number. See the [MSIX Package Versioning](https://github.com/microsoft/WindowsAppSDK/blob/main/specs/deployment/MSIXPackageVersioning.md) for more details.
+* ShortVersionTag = short form of the VersionTag
+
+ShortVersionTag is derived from a VersionTag by combining the 1st letter and the last digit (if any) for non-Stable channels (ShortVeresionTag is blank for the Stable channel, just like VersionTag).
```
BenJKuhn, line 178: last digit? This is guaranteed to collide if we hit preview10, right? (which we should never do). Can we just say number (is there room for 2 digits if we ever got that crazy for some reason?)

BenJKuhn (approved):

specs/Deployment/MSIXPackages.md
DrusTheAxe: >This is guaranteed to collide if we hit preview10, right?

Yup. Ditto experimental10. There's games we could play, but we're off the edge of the map into the wild and wooly lands...

We're crunching pretty close on package Name max length (50) e.g. release 1.1.0 Preview1
* `MicrosoftCorporationII.WinAppRuntime.Main.1.1-p1` = 48
* `Microsoft.WinAppRuntime.DDLM.1000.619.455.0-x6-p1` = 49

The MSIX revision is .0. If we need fixes and that hits 2 digits (10-99) then we're @ 50. If we hit some odd policy where, say, it must be multiples of 10 or 16 then our 1st such fix would be 2 digits and boom! We're @ the 50char max.

The MSIX minor version is # days since Jan'1, 2021. Release 1.0.0 Stable had a value of 319. That'll hit 4 digits in Sep'2023. and boom - 50 chars max. that's only ~18 months out.

We could find ways around this if necessary. For instance, change "DDLM" to "D". But names are encoded pretty widely (product code, test code, build processes, documentation, samples...) so it would have costs. We'd really want (or need) to go there.

Can't swear we won't need to get crazy someday. But devil's very much in the details here. So I'd rather play it conservative for now and, if the day comes to figure out where to bend the rules with least impact, we'll have as much flexibility as possible.

PR #2110, opened 2022-02-16
User/dahoehna/api contracts

dhoehna:
Adding documentation got API contracts.


docs/WinRTAPIContracts.md
```diff
@@ -0,0 +1,76 @@
...
+2.  Stable
+3.  Customizable so developers can include only the parts they want
+
+Contracts, and contract versions should be added to WASDK.
+
+## Architecture Decisions
+1. Contract boundaries
+Contracts will be split among the winmd boundaries.  This is a good medium between having 1 contract for everything, and 1 contract per class.
+2. Contract numbering.
+Contracts will be numbered with MAJOR.MINOR.  Any update to the API is a MAJOR change.  Any servicing update is a MINOR change.
```
BenJKuhn, line 16: Contract versions already have semantics, but since they were originally designed for platform use, they have no notion of a "breaking" change semantic. For now, I do like the idea of using the minor version number to indicate contract increments within the 1.* Windows App SDK series, but I think at this time we should leave it as an open issue as to how we might handle breaking changes. The contract numbering is primarily to indicate ordering of when things are introduced, but the numbers themselves are mostly meaningless outside of the inner workings of the type system. There is an "is contract present" API but the type / API checks are preferred over contract checks (IsContractPresent further suggests that within a contract there's never breaking changes, breaking change might imply new contract name).

docs/WinRTAPIContracts.md
```diff
@@ -0,0 +1,76 @@
...
+
+This data will allow us to better direct our development efforts to help the most developers.
+
+## Gate check-ins
+Because the contracts are used for API versions there will need to be work done in the build system to make sure
+1. All API's have contracts
+2. Each contract is correct.  This means version and name.
+3. All Contracts and dependencies work with each other.
+
+If there is an issue then the build should stop and no release should be made.
```
BenJKuhn, line 71: ❤️

docs/WinRTAPIContracts.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Open issues
+1. Telemetry needed for the data science team. The Data Science team should be involved with our contracts so we can make sure they have the correct data to answer business questions.
```
BenJKuhn, line 74: I didn't think we had telemetry at the contract level. Because contracts are largely a versioning construct for binary compatibility, I don't think it's the right granularity for telemetry, either. We should discuss this one more.

docs/WinRTAPIContracts.md
```diff
@@ -0,0 +1,76 @@
...
+2. Contract boundary.  1 contract for WASDK is too big.  Yet, 1 contract per API is too many.  1 contract per winmd is a good middle.  This is an open issue because there might be a better option.
+3. Should ProjectReunionClosed use contracts?
```
BenJKuhn, line 76: if there are any winrt apis in there, the answer is yes.

docs/WinRTAPIContracts.md
```diff
@@ -0,0 +1,76 @@
+# Windows App SDK API Contracts
+
+## Summary
+Customers trust Microsoft that the products we make are sound, and will work.  Windows App SDK (WASDK) has recently released version 1.0.  As WASDK evolvs and more developers use WASDK in their own projects WASDK needs a way to guarantee that the API’s that ship are stable and that API versions are well documented.
```
riverar, line 4:
```suggestion
Customers trust Microsoft that the products we make are sound, and will work.  Windows App SDK (WASDK) has recently released version 1.0.  As WASDK evolves and more developers use WASDK in their own projects, WASDK needs a way to guarantee that the APIs that ship are stable and that API versions are well documented.
```

Document throughout needs a few spelling/grammar passes. Some nits that apply globally:
* Style nit: Remove double-space after full stop `.` (L14)
* Consistency nit: Remove `.` (L15) or add `.` to bulleted items (create global consistency)

DrusTheAxe (dismissed):

docs/WinRTAPIContracts.md
DrusTheAxe, line 4: TYPO: evolves

docs/WinRTAPIContracts.md
```diff
@@ -0,0 +1,76 @@
...
+            public String GetEnvironmentVariable(String name);
+        }
+    }
+
+## 1.1 ship blocker
+Any API's that changed between 1.0 and 1.1 need to add a contract or 1.1 will not ship.
+
+## Telemetry
+Because all API's have a contract version data can be used to see
+ 1. How many people are using each API
```
DrusTheAxe, line 60: I'm not seeing how API contract versions propogate through telemetry. If we add telemetry to include component=`Microsoft.UI.XAML.Control.Button` AND apicontract=`1.1` but I expect winappsdkversion=`1.1.8` (or whatever) is the more useful superset of information (given version you can determine contracts, but not vice versa).

Please elaborate how contracts can be used to via telemetry to provide richer information than available via the alternatives (like winappsdk version).

docs/WinRTAPIContracts.md
```diff
@@ -0,0 +1,76 @@
...
+ 2. How many people are using each specific version
+
+This data will allow us to better direct our development efforts to help the most developers.
+
+## Gate check-ins
+Because the contracts are used for API versions there will need to be work done in the build system to make sure
+1. All API's have contracts
+2. Each contract is correct.  This means version and name.
+3. All Contracts and dependencies work with each other.
```
DrusTheAxe, line 69: What about compatibility? Don't we need reports (detection) of **changes**? maybe even enforcement to reject changes that break compat?

docs/WinRTAPIContracts.md
```diff
@@ -0,0 +1,76 @@
...
+
+If there is an issue then the build should stop and no release should be made.
+
+## Open issues
+1. Telemetry needed for the data science team. The Data Science team should be involved with our contracts so we can make sure they have the correct data to answer business questions.
+2. Contract boundary.  1 contract for WASDK is too big.  Yet, 1 contract per API is too many.  1 contract per winmd is a good middle.  This is an open issue because there might be a better option.
```
DrusTheAxe, line 75: Depends on the granularity of .winmd. What's the relationship of namespace to winmd, 1:1 or 1:N?

docs/WinRTAPIContracts.md
```diff
@@ -0,0 +1,76 @@
+# Windows App SDK API Contracts
```
DrusTheAxe, line 1: Reads more like a spec than coding guidelines. Mostly it's WHY and little on the HOWTO guide for developers writing code.

SUGGESTION: Move this to `specs\WinRT\WinRTAPIContracts.md` and once mostly baked out add a `docs/Coding-Guidelines/WinRTAPIContracts.md` focused on the HOWTO guidance for devs writing code with a link to the spec for those who want to know WHY

docs/Coding-Guidelines.md
```diff
@@ -11,3 +11,4 @@ languages used in implementation or testing.
   - [Markdown](Coding-Guidelines/Languages-Markdown.md)
 - [Experimental](Coding-Guidelines/Experimental.md)-only features
   - Use [TerminalVelocity](Coding-Guidelines/TerminalVelocity.md) tag and disable features
+- [Api Contracts](WinRTAPIContracts.md)-make a stable API.
```
DrusTheAxe, line 14: `Coding-Guidelines/WinRTAPIContracts`

This list will grow so we're keeping all coding guidelines docs (other than the top level one) in the Coding-Guidelines subdir. To prevent docs\* from becoming a huge mosh pit of many topics.

docs/WinRTAPIContracts.md
```diff
@@ -0,0 +1,76 @@
...
+To keep customers’ trust in WASDK, and to make sure the WASDK API’s are
+1.  Binary compatible
+2.  Stable
+3.  Customizable so developers can include only the parts they want
+
+Contracts, and contract versions should be added to WASDK.
+
+## Architecture Decisions
+1. Contract boundaries
+Contracts will be split among the winmd boundaries.  This is a good medium between having 1 contract for everything, and 1 contract per class.
```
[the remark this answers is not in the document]
DefaultRyan: Currently, `Microsoft.UI.Xaml.winmd` contains the `Microsoft.UI.Xaml.Controls` namespace. However, there's no reason that this always needs to be the case. The level of granularity that is required is: "contracts won't span across feeder repos", aka "at least one contract per feeder repo".

Because functionality is typically grouped by namespace, and API contracts will also follow functionality groups, we should expect a lot of overlap between API contracts and namespaces. In other words, the mapping between API contracts and namespaces won't be prescribed, but will be a common natural occurence.

docs/WinRTAPIContracts.md
DefaultRyan: I think that each release of the Windows App SDK should increment the *major* version, but not because we're using API contracts to imply a degree of change or compat (aka semantic versioning).

API contracts should increment the major version because, like @BenJKuhn indicated, the contract versioning indicates a strict ordering of API changes, but the API contract format in WinRT winmds has only major+minor, not major+minor+patch+build - we've already released multiple 0.8.* versions of Windows App SDK, and shouldn't be surprised to see a 1.0.1 release. For this reason alone, it's a **non-goal** for API contract versions to attempt to match Windows App SDK versions.

Incrementing just the major version removes any ambiguity about "which version is next". When deciding what to number the successor to contract 1.0, is it 1.1 or 2.0? Having a clear simple rule is a virtue here.

Another reason is that someday we may need to modify a contract version in a servicing branch, and we need to be able to do it in a way that maintains the strict ordering in contract versions. For example, say we've released Windows App SDK 1.1 and 1.2 containing `FooContract` 1.1 and 1.2, respectively. If we needed to make a change to `FooContract` 1.1, we wouldn't be able to version it properly because we already used up all of our precision/granularity when we released `FooContract` 1.2. We avoid that dilemma by incrementing only major versions of API contracts in stable releases, and keeping the minor version in our pocket as an emergency pressure relief.

docs/WinRTAPIContracts.md
DefaultRyan: That's a good point about the 1-n mapping between winappsdk and api contracts. Capturing the winappsdk version for telemetry is the more valuable data point.

docs/WinRTAPIContracts.md
DefaultRyan: Yes. We need to enforce these semantics on contract versioning. Once a contract version has gone into a stable release, any changes to types/APIs in that contract must go into a new version, and compat-breaking changes are forbidden. Specifically, after a release
- Method signatures must not change.
- Struct layouts must not change.
- Interfaces must not add, remove, or reorder methods
- All new properties/methods/events added to a runtimeclass must be put on a new interface, which is on a new contract version.
  - When using MIDL3, the new versioned interface is automatically created by putting the contract version on the added property/method/event.
- Ditto for adding static methods/properties/events to a runtimeclass.
- Enum fields can be added under a new contract version. All other enum changes like reordering, changing, or removing fields, is forbidden.

docs/WinRTAPIContracts.md
dhoehna: I'm open to discussion.  I would like for us to see what versions of API's are being used.

docs/WinRTAPIContracts.md
dhoehna: Nope.

dhoehna: There are.

docs/WinRTAPIContracts.md
dhoehna: Never have I ever thought I'd be called out for adding two spaces after a period.

docs/WinRTAPIContracts.md
dhoehna: Can a check like this happen at build time?  In both a developers environment and the internal build system?  Also, should this list be added to a "Forbidden changes" section?

docs/WinRTAPIContracts.md
dhoehna: Oooohhh.  Yeah.  I think I was missing a step in my logic.  I thought developers could choose what versions of APIs to use.  If that isn't the case and developers get whatever version of the API is in their version of Windows App SDK then it does not make sense to include telemetry for API versions.

docs/WinRTAPIContracts.md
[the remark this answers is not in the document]
dhoehna: We should all have a conversation about this.  Microsoft.Windows.System contains only EM.  Microsoft.Windows.Power contains only power notifications.

Some winmd's have one namespace have one API.

docs/WinRTAPIContracts.md
dhoehna: >  add . to bulleted items (create global consistency)

Can you please explain?  All the lists are numbered.  Is there a list somewhere else that I can't see?

docs/WinRTAPIContracts.md
dhoehna: I see.  I was referring to the internal-only APIs.  I'll update the document.

specs/WinRT/WinRTAPIContracts.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Example
+
+    namespace Microsoft.Windows.System
+    {
+        [contractversion(2)]
+        apicontract MicrosoftWindowsSystem
+        
+        // version 2
+        [Contract(MicrosoftWindowsSystem(2)]
```
jevansaks, line 36: i believe contract attribute is commas (as you have on line 50 below but here and on line 43 it's different.
```suggestion
        [Contract(MicrosoftWindowsSystem, 2)]
```

specs/WinRT/WinRTAPIContracts.md
```diff
@@ -0,0 +1,78 @@
...
+        public enum Scope
+        {
+            user,
+            machine
+        }
+
+        [Contract(MicrosoftWindowsSystem(2)]
```
jevansaks, line 43:
```suggestion
        [Contract(MicrosoftWindowsSystem, 2)]
```

specs/WinRT/WinRTAPIContracts.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Adding a new contract
+Developers are encouraged to use their best judgment when deciding where to place a new contract.  If in doubt you can reach out to the Windows App SDK team.
+
+## Example
+
+    namespace Microsoft.Windows.System
+    {
+        [contractversion(2)]
+        apicontract MicrosoftWindowsSystem
```
jevansaks, line 33: An unfortunate side effect of contract attributes is that they are a public thing that ends up on in the metadata and documentation site. Should we recommend these contract types go in a separate namespace so they don't clutter the docs?

specs/WinRT/WinRTAPIContracts.md
```diff
@@ -0,0 +1,78 @@
...
+        {
+            public String GetEnvironmentVariable(String name);
+        }
+    }
+
+## Forbidden interface changes
+To help keep ABI and binary compatibility between different versions of Windows App SDK here is a list of changes that should not happen between interface versions.
+1. Method order in the interface.
+2. Method signatures.
+3. Structs.
```
jevansaks, line 61: Structs cannot be modified. Interfaces cannot be modified. Enums can only be added to, but existing values and names cannot change. 

We should also make a decision about runtimeclass derivation. If I have runtimeclass A derives from B and then I introduce runtimeclass C that derives from B, can I change A to derive from C later? In theory this is safe but we have avoided doing it to date.

jevansaks: I'll also say that many people don't grok that a runtimeclass is an aggregation of interfaces, so if you want this to be a doc that folks reference when they are trying to figure out how to manage their contract versioning, please include examples like the internal wiki page does. Most additive things are possible but the syntax for doing it isn't always obvious, like the contract block syntax or adding a setter to a property that previously only had a getter.

specs/WinRT/WinRTAPIContracts.md
```diff
@@ -0,0 +1,78 @@
+
+
+
```
DrusTheAxe, line 3: Remove lines 1-3 (dead whitespace)

specs/WinRT/WinRTAPIContracts.md
```diff
@@ -0,0 +1,78 @@
...
+# Windows App SDK API Contracts
+
+## Summary
+Customers trust that Microsoft products are sound.  As Windows App SDK evolves and more developers use Windows App SDK in their own projects Windows App SDK needs a way, at build time and run time, to allow apps to light up API features and to gracefully degrade.  This can be achieved by looking at the deltas between APIs.
```
DrusTheAxe, line 7: Don't make markdown lines >100 chars long

https://github.com/microsoft/WindowsAppSDK/blob/main/docs/Coding-Guidelines/Languages-Markdown.md

VSCode has some nice extensions that automagically format paragraphs or do so at a keystroke (alt-q seems popular).

DrusTheAxe, line 7: Single space between sentences

specs/WinRT/WinRTAPIContracts.md
```diff
@@ -0,0 +1,78 @@
...
+
+In order to allow apps to depend on API versions Windows App SDK will add contracts to all their WinRT APIs.
+
+
+## Architecture Decisions
+1. Contract boundaries
+   1. One contract per feature/component/sub-system
+   2. Contracts will not span across a transport package/repo.
+2. Contract numbering.
+Contracts will be versioned with one number, starting with 1 and is incremented for any interface change.  This version does not correlate with the Windows App SDK version.
```
DrusTheAxe, line 17: Compare the markdown layout vs the formatted display
image
image

You forgot Markdown happily flows your paragraph instead of treating each newline as significant.

SUGGESTION: Add some minor formatting tweaks e.g.

1. **Contract boundaries**
    1. One contract per feature/component/sub-system
    2. Contracts will not span across a transport package/repo.
2. **Contract numbering** -- Contracts will be versioned...
3. **Enforcement** -- The build systems...

Don't rely on newlines to format your bulleted list

specs/WinRT/WinRTAPIContracts.md
```diff
@@ -0,0 +1,78 @@
...
+3. Enforcement
+The build systems will have enforcement gates to make sure that contracts are present, up to date, and there are no conflicts.
```
DrusTheAxe, line 19: Assume present tense and these things exist (as they do, just not as extensively as they will with work under way)

`The build system's enforcement gates make sure contracts are present, ...```

specs/WinRT/WinRTAPIContracts.md
```diff
@@ -0,0 +1,78 @@
...
+4. Contract Migration.
+Contract Migration will not be a part of this spec.  Windows App SDK will add a spec on contract migration if, and when, there is a situation where it is needed.
```
DrusTheAxe, line 21: NIT: `...if/when it is needed.`

specs/WinRT/WinRTAPIContracts.md
```diff
@@ -0,0 +1,78 @@
...
+5. Internal contracts
+All WinRT internal APIs will have an contract and the version will always be 1 regardless of any interface changes.
```
DrusTheAxe, line 23: Use present tense:

Internal WinRT APIs must have a contract. Their version is always 1 regardless of any interface changes e.g.

```
[contractversion(1)]
enum FooInternal
...
```

(or a more extensive example)

DrusTheAxe, line 23: All WinRT types are contracted, not just interfaces (enum, runtimeclass, other types of types &lt;sic&gt; need contracts).

specs/WinRT/WinRTAPIContracts.md
DrusTheAxe: You say clutter I say... ;-)

I don't see the merit of splitting them out, and not outweighed by the drawbacks

specs/WinRT/WinRTAPIContracts.md
DrusTheAxe, line 43: `[Contract(x,1)]` or `[contract(x,1)]`

Must be lower case?

Even if not, should be for consistency reasons?

specs/WinRT/WinRTAPIContracts.md
```diff
@@ -0,0 +1,78 @@
...
+            machine
+        }
+
+        [Contract(MicrosoftWindowsSystem(2)]
+        public class EnvironmentManager
+        {
+            public String GetEnvironmentVariable(String name, Scope scope);
+        }
+
+        // base version
```
DrusTheAxe, line 49: Purpose of the comment? Seems redundant and not meaningful to me.

Remove

specs/WinRT/WinRTAPIContracts.md
```diff
@@ -0,0 +1,78 @@
...
+        [contract(MicrosoftWindowsSystem, 1)]
+        public class EnvironmentManager
+        {
+            public String GetEnvironmentVariable(String name);
+        }
+    }
+
+## Forbidden interface changes
+To help keep ABI and binary compatibility between different versions of Windows App SDK here is a list of changes that should not happen between interface versions.
```
DrusTheAxe, line 58: NIT: Remove 'help'

This is necessary-but-insufficient, but that's a rather pedantic distinction.

DrusTheAxe, line 58: remove "ABI and"

ABI = Application **Binary** Interface. No need to say both ABI and binary compat, and we talk elsewhere about "binary compatibility" so it's the preferred term here

DrusTheAxe, line 58: I'd be a bit terser and stern about the guidance phrase

SUGGESTION: DO NOT break binary compatibility except across major release boundaries. The following changes break binary compatibility:

specs/WinRT/WinRTAPIContracts.md
dhoehna: From looking at other examples of the contract attribute I don't see any commas. @DrusTheAxe and @BenJKuhn should commas be included, or not in contract versions?

specs/WinRT/WinRTAPIContracts.md
dhoehna: Contract attributes should stay in the file where they are needed.

specs/WinRT/WinRTAPIContracts.md
dhoehna: > want this to be a doc that folks reference

Thank you for bringing this issue up.  This is a document for specs, not coding guidelines.  I'll make a separate PR with contract guidelines, including forbidden changes.  For this PR This section will be removed because "Don't do these things" is more "Coding guidelines" and not "spec"

specs/WinRT/WinRTAPIContracts.md
DrusTheAxe: Docs say commas

https://docs.microsoft.com/en-us/uwp/midl-3/predefined-attributes#the-contract-attribute
image

@BenJKuhn does MIDL also support a Super Secret Secret Squirrel alternative notation of `[contract(Foo(2))]` in addition to the doc'd `[contact(Foo,2)]`?

DrusTheAxe: Docs say commas

https://docs.microsoft.com/en-us/uwp/midl-3/predefined-attributes#the-contract-attribute
image

specs/WinRT/WinRTAPIContracts.md
```diff
@@ -0,0 +1,78 @@
...
+Contracts will be versioned with one number, starting with 1 and is incremented for any interface change.  This version does not correlate with the Windows App SDK version.
+3. Enforcement
+The build systems will have enforcement gates to make sure that contracts are present, up to date, and there are no conflicts.
+4. Contract Migration.
+Contract Migration will not be a part of this spec.  Windows App SDK will add a spec on contract migration if, and when, there is a situation where it is needed.
+5. Internal contracts
+All WinRT internal APIs will have an contract and the version will always be 1 regardless of any interface changes.
+
+## Adding a new contract
+Developers are encouraged to use their best judgment when deciding where to place a new contract.  If in doubt you can reach out to the Windows App SDK team.
```
jonwis, line 26: Contract association is part of API review.  So like:

"API review will determine whether an API should join an existing contract or be part of a new one. Contracts are groups of associated functionality. A contract containing just one feature is too small, while a contract containing all functionality is too large."

specs/WinRT/WinRTAPIContracts.md
jonwis: Lower case.

specs/WinRT/WinRTAPIContracts.md
jonwis, line 33: >MicrosoftWindowsSystem [](http://example.com/codeflow?start=20&length=22)

Contracts follow a `FooBarContract` pattern, so this would be `WindowsSystemContract`

specs/WinRT/WinRTAPIContracts.md
```diff
@@ -0,0 +1,78 @@
...
+    namespace Microsoft.Windows.System
+    {
+        [contractversion(2)]
+        apicontract MicrosoftWindowsSystem
+        
+        // version 2
+        [Contract(MicrosoftWindowsSystem(2)]
+        public enum Scope
+        {
+            user,
```
jonwis, line 39: >user [](http://example.com/codeflow?start=12&length=4)

Consider cloning this from the existing IDL and applying the contract markers, rather than rewriting here.

specs/WinRT/WinRTAPIContracts.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Summary
+Customers trust that Microsoft products are sound. As Windows App SDK evolves and more developers
+use Windows App SDK in their own projects Windows App SDK needs a way, at build time and run time,
+to allow apps to light up API features and to gracefully degrade. This can be achieved by looking
+at the deltas between APIs.
+
+In order to allow apps to depend on API versions Windows App SDK will add contracts to all their
+WinRT APIs.
+
```
DrusTheAxe, line 11: NIT: Remove blank line

You only need 1 blank line between sections/paragraphs

specs/WinRT/WinRTAPIContracts.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Architecture Decisions
+1. Contract boundaries
+   1. One contract per feature/component/sub-system
+   2. Contracts will not span across a transport package/repo.
+3. Contract numbering.
```
DrusTheAxe, line 17: Numbering is weird.

```
1. Contract Boundaries
    1. Blah
    2. Blah
3. Contract numbering
4. Yadda yadda

Shouldn't that be 2. Contract numbering and renumber the rest to match?

Comment (DrusTheAxe) [inline specs/WinRT/WinRTAPIContracts.md:17]: NIT: Inconsistent formatting

3+5 are "<number> Blah blah." whereas 1+4+6 are "<number> Blah blah" (no dot)

You're also doing formatting that Markdown doesn't respect.

```
4. Enforcement
The build system has...
```

which you'd think is 2 lines but Markdown (like HTML) is flow based. Take a look at the "rich diff" for the formatted document
image
Very different from the raw layout.

File: specs/WinRT/WinRTAPIContracts.md [hunk 28]
```diff
@@ -0,0 +1,80 @@
...
+change. This version does not correlate with the Windows App SDK version.
+4. Enforcement
+The build system has gates that enforce contract versioning to verify that contract is present,
+ up to date, and there are no conflicts between APIs.
+5. Contract Migration.
+This spec does not deal with contract migration. A separate spec will be available for contract
+migration if, and when, it is needed.
+6. Internal contracts
+All WinRT internal APIs have a contract, and the version is 1.  The version is never updated.
+
```
DrusTheAxe, line 28: I don't see a "Contract Naming" item in the list. Don't all contracts need to end with `...Contract`? e.g. `BlahBlahContract`

specs/WinRT/WinRTAPIContracts.md
```diff
@@ -0,0 +1,80 @@
...
+## Adding a new contract
+Contract placement is part of an API review. When adding a new API, the API team will
+talk about what contract the API will go under.
+
+## Example
+
+    namespace Microsoft.Windows.System
```
DrusTheAxe, line 35: NIT: You don't need to indent the code here.

You also forgot the code markers (this is just prose)

Should be

```c# (really midl3)
namespace Microsoft.Windows.System
{
    [contractversion(1)]
    apicontract WindowsSystemContract {}
    ...
}
```

specs/WinRT/WinRTAPIContracts.md
```diff
@@ -0,0 +1,80 @@
...
+## Telemetry
+API specific telemetry will not be collected because it does not provide any additional information
+compared to the Windows App SDK version.
+
+## Gate check-ins
+Because the contracts are used for API versions there will need to be work done in the build system
+to make sure
+1. All APIs have contracts
+2. Each contract is correct. This means version and name.
+3. All Contracts and dependencies work with each other.
```
DrusTheAxe, line 75: 4. Contracts/Types in releases honor the source/binary/behavior compatibility rules e.g. contracts/types in a stable release doesn't change in future releases of the same major version (e.g. contracts/types in 1.1 don't change in 1.2, 1.3, ...)

specs/WinRT/WinRTAPIContracts.md
```diff
@@ -0,0 +1,80 @@
...
+
+The build will stop if an issue is found and there will be no nightly release.
```
DrusTheAxe, line 77: "and" connects 2 distinct issues
* If issues arise the build stops i.e. it's a build break
* Nightly releases are affected

Do these gates only run on nightly builds? What about CI and other builds?

specs/WinRT/WinRTAPIContracts.md
dhoehna: Interesting.  Many of your comments have been formatting related.  I've been using stackedit for md.

dhoehna: seems like stackedit and github's viewers don't agree.

specs/WinRT/WinRTAPIContracts.md
```diff
@@ -0,0 +1,88 @@
...
+
+In order to allow apps to depend on API versions Windows App SDK will add contracts to all their
+WinRT APIs.
+
+## Architecture Decisions
+1. Contract boundaries
+   1. One contract per feature/component/sub-system.
+   2. Contracts will not span across a transport package/repo.
+2. Contract numbering  
+Contracts versions are one number, starting with 1 and is incremented for any interface
```
DefaultRyan, line 18: This could use some clarification. A reader new to the topic might think every single change within a release requires a new contract version. Perhaps something like this:

> Contract version numbers are a single number, with no minor "x.y" versions. The first version of a contract will be version 1. Contract versions are frozen upon a stable release of Windows App SDK, any changes within a contract after that stable release will require incrementing the version number.

DefaultRyan:
Couple of minor nits, but I believe this works. Let's check it in.

Issue #2075, opened 2022-02-08
Fix payload sending from PushNotificationsLongRunningProcess 


pmpurifoy:
### Describe the bug

Currently sending the payload by winrt::box_value in the LRP and unboxing it in the SDK. Need to change the process to mock the RawNotification object and pass that through BackgroundTaskInstance. 

### Steps to reproduce the bug

Not a bug. More an implementation improvement.

PR #2119, opened 2022-02-17
Refactor AppNotificationManager and add firstNotificationReceived

pmpurifoy:


dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -74,7 +74,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
             THROW_HR_IF_MSG(E_INVALIDARG, m_notificationComActivatorRegistration, "Already Registered for App Notifications!");
             THROW_IF_FAILED(::CoRegisterClassObject(
                 AppModel::Identity::IsPackagedProcess() ? details.TaskClsid() : winrt::guid(storedComActivatorString),
-                winrt::make<AppNotificationActivationCallbackFactory>().get(),
+                winrt::make<AppNotificationManagerFactory>().get(),
```
hulumane, line 77: This means that we have two instances of AppNotificationManager. One through Default() and the other created specifically for COM. Can we have just one instance for both?

pmpurifoy: CreateInstance in AppNotificationManagerFactory is not creating a new object but it is returning the same static manager from Default() every time.

hulumane: Excellent!. I had to read forward to understand how this was done. Thanks for this!

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -132,6 +132,47 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
+        winrt::IMap<winrt::hstring, winrt::hstring> userInput{ winrt::single_threaded_map<winrt::hstring, winrt::hstring>() };
+        for (unsigned long i = 0; i < dataCount; i++)
+        {
+            userInput.Insert(data[i].Key, data[i].Value);
+        }
+
+        winrt::AppNotificationActivatedEventArgs activatedEventArgs = winrt::make<winrt::Microsoft::Windows::AppNotifications::implementation::AppNotificationActivatedEventArgs>(invokedArgs, userInput);
+
+        // Need to store the first notification in the case of ToastActivation
+        if (!m_firstNotificationReceived)
```
hulumane, line 150: Lock Lock Lock!

pmpurifoy: Isn't this already synchronized by ComActivation?

hulumane: Otherwise 2 activations can race through the if condition and cause problems. This if is in essence atomic.

hulumane: We can have two threads that call CoCreateInstance at the same time and the processor can context switch!

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -132,6 +132,47 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
+        {
+            m_firstNotificationReceived = true;
+
+            std::wstring commandLine{ GetCommandLine() };
+            auto pos{ commandLine.find(c_notificationActivatedArgument) };
+            if (pos == std::wstring::npos)
```
hulumane, line 156: Add a comment here to give context. 
// If the app was not launched due to a toast invoked event, we will invoke the event handler. Otherwise, we populate the EventArgs and signal the Main thread that we are ready to access the EventArgs.

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -132,6 +132,47 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
+            {
+                m_notificationHandlers(Default(), activatedEventArgs);
```
hulumane, line 158: Won't this crash if there are no event handlers registered with the Manager?

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -132,6 +132,47 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
+            else
+            {
+                auto appProperties = winrt::CoreApplication::Properties();
+                appProperties.Insert(ACTIVATED_EVENT_ARGS_KEY, activatedEventArgs);
+                SetEvent(GetWaitHandleForArgs().get());
+            }
+        }
+        else
+        {
+            m_notificationHandlers(Default(), activatedEventArgs);
```
hulumane, line 169: Won't this crash if there are no event handlers associated with the manager?

dev/AppNotifications/AppNotificationManager.h
```diff
@@ -36,14 +31,40 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
         winrt::Windows::Foundation::IAsyncAction RemoveAllAsync();
         winrt::Windows::Foundation::IAsyncOperation<winrt::Windows::Foundation::Collections::IVector<winrt::Microsoft::Windows::AppNotifications::AppNotification>> GetAllAsync();
 
-        // IAppNotificationManagerInternal
-        bool ContainsInvokeHandler();
-        void InvokeHandler(const winrt::Microsoft::Windows::AppNotifications::AppNotificationActivatedEventArgs& args);
```
hulumane, line 41: // INotificationActivationCallback

dev/AppNotifications/AppNotificationManager.h
```diff
@@ -36,14 +31,40 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
+    struct AppNotificationManagerFactory : winrt::implements<AppNotificationManagerFactory, IClassFactory>
+    {
+        STDMETHODIMP CreateInstance(_In_opt_ IUnknown* aggregateInterface, _In_ REFIID interfaceId, _Outptr_ VOID** object) noexcept final try
+        {
+            RETURN_HR_IF(CLASS_E_NOAGGREGATION, aggregateInterface != nullptr);
+            return AppNotificationManager::Default().as(interfaceId, object);
+        }
+        CATCH_RETURN()
+
+            STDMETHODIMP LockServer(BOOL fLock) noexcept final
```
hulumane, line 56: Bad indents here for this function.

dev/AppNotifications/AppNotificationManager.cpp
pmpurifoy: I believe it would just do nothing.

dev/PushNotifications/GetNotificationEventArgs.h
```diff
@@ -26,6 +26,26 @@ namespace winrt::Microsoft::Windows::PushNotifications
...
+        THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_TIMEOUT), !GetWaitHandleForArgs().wait(receiveArgsTimeoutInMSec));
+
+        // If COM static store was uninit, let it throw
+        if (PushNotificationHelpers::IsPackagedAppScenario())
+        {
+            return winrt::Windows::ApplicationModel::Core::CoreApplication::Properties().Lookup(ACTIVATED_EVENT_ARGS_KEY);
+        }
+        else
+        {
+            return winrt::Windows::ApplicationModel::Core::CoreApplication::Properties().Lookup(LRP_ACTIVATED_EVENT_ARGS_KEY);
```
hulumane, line 40: This key is true for the COM activation case and ShellExecute case from LRP right?

test/TestApps/ToastNotificationsDemoApp/main.cpp
```diff
@@ -133,6 +133,23 @@ int main()
     // Grab an instance of AppNotificationManager
     auto appNotificationManager{ winrt::AppNotificationManager::Default() };
 
+    // Setting up foreground handler for AppNotification
+    std::wcout << L"Registering foreground handler to receive AppNotificationActivatedEventArgs...\n";
+    winrt::event_token token = appNotificationManager.NotificationInvoked([](const auto&, winrt::AppNotificationActivatedEventArgs const& toastArgs)
+        {
```
hulumane, line 139: indenting issue

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -132,6 +132,47 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
         m_notificationHandlers.remove(token);
     }
 
+    HRESULT __stdcall AppNotificationManager::Activate(
```
hulumane, line 135: If the developer subscribes to the first even handler after the REgister API call we should THROW.

hulumane: There was no other place to put this comment, so i put it here :)

dev/PushNotifications/GetNotificationEventArgs.h
pmpurifoy: Will add a comment pointing to open bug. https://github.com/microsoft/WindowsAppSDK/issues/2075

dev/AppNotifications/AppNotificationManager.cpp
hulumane: Test and verify please

hulumane: At the least, I would expect it to throw

dev/AppNotifications/AppNotificationManager.cpp
pmpurifoy: Tested that nothing happens :)

dev/AppNotifications/AppNotificationManager.cpp
pmpurifoy: Update: Does nothing

hulumane (approved):

PR #2122, opened 2022-02-17
Refactor PushNotificationManager to hold event handlers

pmpurifoy:


pmpurifoy:
Still need to fix the PushNotifications spec to account for these changes.

pmpurifoy:
Need to throw exception when no handlers are registered.

dev/PushNotifications/PushNotificationChannel.h
```diff
@@ -3,19 +3,14 @@
...
 #include <TerminalVelocityFeatures-PushNotifications.h>
 #include "externs.h"
 
 namespace winrt::Microsoft::Windows::PushNotifications::implementation
 {
-    typedef winrt::Windows::Foundation::TypedEventHandler<
-        winrt::Microsoft::Windows::PushNotifications::PushNotificationChannel,
-        winrt::Microsoft::Windows::PushNotifications::PushNotificationReceivedEventArgs> PushNotificationEventHandler;
-
-    struct PushNotificationChannel : PushNotificationChannelT<PushNotificationChannel, IWpnForegroundSink, ABI::Microsoft::Internal::PushNotifications::INotificationListener>
```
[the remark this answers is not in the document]
pmpurifoy: Yup :)

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -52,15 +53,10 @@ namespace PushNotificationHelpers
...
-    static winrt::Windows::ApplicationModel::Background::IBackgroundTaskRegistration s_pushTriggerRegistration{ nullptr };
-    static wil::unique_com_class_object_cookie s_comActivatorRegistration;
-    static bool s_protocolRegistration{ false };
-    static wil::srwlock s_activatorInfoLock;
-
     inline constexpr auto c_maxBackoff{ 5min };
     inline constexpr auto c_initialBackoff{ 60s };
     inline constexpr auto c_backoffIncrement{ 60s };
-
+
```
hulumane, line 59: nit:extra spaces here.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -240,17 +268,17 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
             if (isProtocolActivatorSet)
```
hulumane, line 268: It would be nice when all this stuff is cleaned away by @sharath2727 . It is way too complex in its current form and I'm looking forward to the cleanup.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -182,6 +199,17 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
                         THROW_IF_FAILED(notificationPlatform->AddToastRegistrationMapping(processName.get(), toastAppId.c_str()));
                     }
 
+                    // Need to register sink on new channel creation if handlers are registered
+                    bool registeredEvent{ false };
```
hulumane, line 203: I'm having a hard time understanding why this is here as opposed to REgister() call. I mean if a prior channel already exists, shouldnt the foreground handler stuff just work after Register is called?

This puts Channel request as a forcing function that seems incorrect.

pmpurifoy: If a Channel is requested, the old sink is torn down. So what here is an example of what could happen.
1. App requests channel 
2. App closes and is launched by raw notification
3. Register foreground handlers and handle first notification in background
4. Request channel (old sink is now dead)
5. Second notification goes to background because foreground activation failed

hulumane: So 
Main()
1) Register event Handler
2) Manager::Register() (At this point, the event handlers should be active if we have a prior channel along with the Push Kind extension)
3) Channel request succeeds and we get a new channel which replaces old one
4) All new channel payloads are still routed to event handler

pmpurifoy: Essentially every time a Channel is requested, a new foreground sink has to be registered.

pmpurifoy: Step 3 removes the old notification sink and since PushReceived is never called again, there is no longer a foreground sink

hulumane: I need to think about this a bit more. The issue for me is what happens with a valid channel (from prior launch) and we get a new notification.

hulumane: Or lets say, we get 2 new notifications

hulumane: Can you explain what happens for packaged and unpackaged scenarios when channel request is not done, but REgister is called and we have an active channel?

pmpurifoy: Everything is fine since the old sink still exists. But if the 2nd launch requests a channel, the old sink is torn down and the 3rd notification is sent to background.

pmpurifoy: Valid channel (both scenarios):
1. PushNotification launches the app
2. Register FG handlers 
3. Register() and receive the first notification via background
4. Any notification after this is sent to foreground

5. Request Channel (old foreground sink is torn down)
6. All notifications will route to background because no foreground sink exists

dev/AppLifecycle/ExtensionContract.h
```diff
@@ -25,8 +25,8 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
         { ExtendedActivationKind::File, c_fileContractId, &FileActivatedEventArgs::Deserialize },
         { ExtendedActivationKind::Protocol, c_protocolContractId, &ProtocolActivatedEventArgs::Deserialize },
         { ExtendedActivationKind::StartupTask, c_startupTaskContractId, &StartupActivatedEventArgs::Deserialize },
-        { ExtendedActivationKind::Push, c_pushContractId, &winrt::Microsoft::Windows::PushNotifications::Deserialize },
-        { ExtendedActivationKind::AppNotification, c_appNotificationContractId, &winrt::Microsoft::Windows::PushNotifications::Deserialize },
+        { ExtendedActivationKind::Push, c_pushContractId, &winrt::Microsoft::Windows::PushNotifications::PushDeserialize },
+        { ExtendedActivationKind::AppNotification, c_appNotificationContractId, &winrt::Microsoft::Windows::PushNotifications::ToastDeserialize },
```
hulumane, line 29: Its weird to see ToastDeserialize in the PushNotifications namespace rather than AppNotifications

dev/PushNotifications/PushNotificationChannel.h
```diff
@@ -3,19 +3,14 @@
...
+    struct PushNotificationChannel : PushNotificationChannelT<PushNotificationChannel>
     {
         PushNotificationChannel(struct ChannelDetails channelInfo)
```
hulumane, line 15: Move implementation to cpp file please from header.

dev/PushNotifications/PushNotificationManager.h
```diff
@@ -4,22 +4,42 @@
...
+        void PushReceived(winrt::event_token const& token) noexcept;
+
+        // IWpnForegroundSink
+        HRESULT __stdcall InvokeAll(_In_ ULONG length, _In_ byte* payload, _Out_ BOOL* foregroundHandled) noexcept;
+
+        // INotificationHandler
+        HRESULT __stdcall OnRawNotificationReceived(unsigned int payloadLength, _In_ byte* payload, _In_ HSTRING /*correlationVector */) noexcept;
+    private:
+
+        bool IsBackgroundTaskRegistered(winrt::hstring const& backgroundTaskFullName);
```
hulumane, line 37: Keep private functions sperate from private variables

dev/AppLifecycle/ExtensionContract.h
pmpurifoy: Since both were using the same Deserialize function, it was PushNotifications. Since this will be reverted soon I decided to leave it in PushNotifications.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -174,14 +187,33 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                    {
+                        auto lock{ m_lock.lock_shared() };
+                        registeredEvent = bool(m_foregroundHandlers);
+                    }
+
+                    // Channel creation removes old sink, need to add sink if foreground handlers.
```
hulumane, line 211: nit: fix this comment
"Need to recreate the Foreground sinks since channel creation removes the old sink"

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +253,282 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
         }
     }
 
-    void PushNotificationManager::RegisterActivator(PushNotificationActivationInfo const& details)
+    bool PushNotificationManager::IsBackgroundTaskRegistered(winrt::hstring const& backgroundTaskFullName)
```
hulumane, line 256: Why pass an argument into a function call when it is essentially a well defined const?
Remove the argument please, define it as a const and reuse the same const everywhere.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +253,282 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        auto tasks{ BackgroundTaskRegistration::AllTasks() };
+        bool isTaskRegistered{ std::any_of(std::begin(tasks), std::end(tasks),
+            [&](auto&& task)
+            {
+                auto name{ task.Value().Name() };
 
-        try
-        {
-            THROW_HR_IF_NULL(E_INVALIDARG, details);
+                if (std::wstring_view(name).substr(0, backgroundTaskName.size()) != backgroundTaskName)
```
hulumane, line 264: Why have this condition at all? The next condition should suffice right?

hulumane: Infact I don't understand why this is even there.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +253,282 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+
+        try
+        {
+            if (!AppModel::Identity::IsPackagedProcess())
             {
                 {
-                    auto lock = s_activatorInfoLock.lock_shared();
-                    THROW_HR_IF(E_INVALIDARG, s_protocolRegistration);
+                    auto lock = m_lock.lock_shared();
+                    THROW_HR_IF(E_INVALIDARG, m_protocolRegistration);
```
hulumane, line 291: THROW_HR_IF_MSG(E_INVALIDARG, m_protocolRegistration, "The app has already been registered once!");

hulumane, line 291: Make sure Unregister() sets this bool to false.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +253,282 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
                 RegisterUnpackagedApplicationHelper(GUID_NULL, unpackagedAppUserModelId); // create default registration for app
 
                 {
-                    auto lock = s_activatorInfoLock.lock_exclusive();
-                    s_protocolRegistration = true;
+                    auto lock = m_lock.lock_exclusive();
+                    // Flag protects from multiple registrations
+                    m_protocolRegistration = true;
                 }
             }
```
hulumane, line 303: Use an else to improve readability. You have another if statement below stating the exact opposite conditional.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +253,282 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-                        }
+            /* Every packaged app is already getting the benefit of COM Registration to handle notifications.
+               However, we would benefit from power management improvements in Background Infrastructure.
+               So we are continuing to register a Push trigger post 19041 builds */
+               
+            if (PushNotificationHelpers::IsPackagedAppScenario())
+            {
+                {
+                    auto lock = m_lock.lock_shared();
+                    THROW_HR_IF(E_INVALIDARG, m_pushTriggerRegistration);
```
hulumane, line 335: THROW_HR_IF_MSG(E_INVALIDARG, m_pushTriggerRegistration, "The app has already Registered!");

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +253,282 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                }
 
-                        throw winrt::hresult_invalid_argument(L"RegisterActivator has different clsid registered.");
-                    }) };
+                winrt::hstring registeredClsidStr{ winrt::to_hstring(registeredClsid) };
```
hulumane, line 338: No need of this extra local: registeredClsidStr. REmove and use directly below.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +253,282 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                winrt::hstring backgroundTaskFullName{ backgroundTaskName + registeredClsidStr };
 
+                bool isTaskRegistered = IsBackgroundTaskRegistered(backgroundTaskFullName);
+               
                 if (!isTaskRegistered)
                 {
+                    BackgroundTaskBuilder builder{ nullptr };
+
                     builder = BackgroundTaskBuilder();
```
hulumane, line 347: Do this in one line please.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +253,282 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
                     builder.Name(backgroundTaskFullName);
 
                     PushNotificationTrigger trigger{};
                     builder.SetTrigger(trigger);
 
-                    THROW_HR_IF(E_NOTIMPL, !AppModel::Identity::IsPackagedProcess());
-
                     // In case the interface is not supported, let it throw.
```
hulumane, line 353: This comment is not needed. This is a PackagedAppScenario which by default indicates that the Cast below should pass. If it doesnt, it is unexpected and fatal.

Nothing to do with it not being supported.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +253,282 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                    auto lock{ m_lock.lock_exclusive() };
+                    m_pushTriggerRegistration = builder.Register();
                 }
             }
 
-            BackgroundTaskRegistration registeredTaskFromBuilder{ nullptr };
+            if (AppModel::Identity::IsPackagedProcess())
+            {
+                bool registeredEvent{ false };
+                THROW_HR_IF(E_INVALIDARG, registeredClsid == winrt::guid());
```
hulumane, line 365: registeredClsid is not an arg anymore. So this should be:
THROW_HR_IF(E_UNEXPECTED, registeredClsid == winrt::guid());

dev/PushNotifications/PushNotificationManager.cpp
hulumane, line 291: You need to move THROW_HR_IF_MSG(E_FAIL, m_protocolRegistration, "The app has already been registered once!"); to the exclusive lock below. Otherwise you have introduced a race on this bool.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +253,282 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-                    auto lock{ s_activatorInfoLock.lock_shared() };
-                    THROW_HR_IF_MSG(E_INVALIDARG, s_comActivatorRegistration, "ComActivator already registered.");
+                    // Need to register foreground sink if foreground handler exists.
+                    RegisterSinkHelper();
                 }
 
-                GetWaitHandleForArgs().create();
-                GetComServerClsid() = taskClsid;
-                
                 THROW_IF_FAILED(::CoRegisterClassObject(
```
hulumane, line 381: Modifying member variable. So this needs to be in an exclusive lock.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +253,282 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                    if (m_pushTriggerRegistration)
                     {
-                        auto name{ task.Value().Name() };
-
-                        if (std::wstring_view(name).substr(0, backgroundTaskName.size()) != backgroundTaskName)
-                        {
-                            return false;
-                        }
+                        m_pushTriggerRegistration.Unregister(true);
+                        m_pushTriggerRegistration = nullptr;
```
hulumane, line 322: You also need 
if (comRegistered)
{
    CoUnregister(registerClsid);

hulumane: And comRegistered = false;

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +253,282 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
 
-                winrt::hstring taskClsidStr{ winrt::to_hstring(taskClsid) };
-                winrt::hstring backgroundTaskFullName{ backgroundTaskName + taskClsidStr };
+                    m_comActivatorRegistration.reset();
 
-                auto tasks{ BackgroundTaskRegistration::AllTasks() };
-                bool isTaskRegistered{ std::any_of(std::begin(tasks), std::end(tasks),
-                    [&](auto&& task)
+                    // Clean the task registration only if it was created during this call
+                    if (m_pushTriggerRegistration)
```
hulumane, line 319: What about cleanup of the Sink with the platform or LongRunning singleton if registration fails?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +253,282 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
         }
 
-        PushNotificationTelemetry::ActivatorUnregisteredByApi(S_OK, activators);
+        PushNotificationTelemetry::ActivatorUnregisteredByApi(S_OK);
     }
 
-    void PushNotificationManager::UnregisterAllActivators()
+    void PushNotificationManager::UnregisterAll()
     {
         THROW_HR_IF(E_NOTIMPL, !::Microsoft::Windows::PushNotifications::Feature_PushNotifications::IsEnabled());
```
hulumane, line 426: REmove and have one check in Default()

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +253,282 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-                auto notificationPlatform{ PushNotificationHelpers::GetNotificationPlatform() };
-
-                wil::unique_cotaskmem_string processName;
-                THROW_IF_FAILED(GetCurrentProcessPath(processName));
-                LOG_IF_FAILED(notificationPlatform->UnregisterLongRunningActivator(processName.get()));
-
-                s_protocolRegistration = false;
+                THROW_HR_IF_MSG(HRESULT_FROM_WIN32(ERROR_NOT_FOUND), !m_comActivatorRegistration, "ComActivator not registered.");
+                m_comActivatorRegistration.reset();
             }
```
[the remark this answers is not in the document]
hulumane: CAre should be taken to only sever the connection with LRP. Not the connection between LRP and platform.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +253,282 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        PushNotificationTelemetry::ActivatorUnregisteredByApi(S_OK);
     }
 
-    bool PushNotificationManager::IsActivatorSupported(PushNotificationRegistrationActivators const& activators)
+    winrt::event_token PushNotificationManager::PushReceived(TypedEventHandler<winrt::Microsoft::Windows::PushNotifications::PushNotificationManager, winrt::Microsoft::Windows::PushNotifications::PushNotificationReceivedEventArgs> handler)
     {
-        THROW_HR_IF(E_NOTIMPL, !::Microsoft::Windows::PushNotifications::Feature_PushNotifications::IsEnabled());
+        {
+            auto lock{ m_lock.lock_shared() };
+            THROW_HR_IF_MSG(HRESULT_FROM_WIN32(ERROR_NOT_FOUND), m_comActivatorRegistration, "Must register event handlers before calling Register().");
```
hulumane, line 468: What about protocol? Why didn't we check that?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +253,282 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+
+                THROW_IF_FAILED(notificationsLongRunningPlatform->UnregisterForegroundActivator(processName.get()));
+            }
         }
-        return false;
     }
+
+    HRESULT __stdcall PushNotificationManager::InvokeAll(_In_ ULONG length, _In_ byte* payload, _Out_ BOOL* foregroundHandled) noexcept try
+    {
+        auto args = winrt::make<winrt::Microsoft::Windows::PushNotifications::implementation::PushNotificationReceivedEventArgs>(payload, length);
```
hulumane, line 509: Please add locks!

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +253,282 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        m_foregroundHandlers(*this, args);
```
hulumane, line 510: @pmpurifoy should this have a first notification check to decide whether it needs to be sent to args or the event handler?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +253,282 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        return S_OK;
+    }
+    CATCH_RETURN()
+
+    HRESULT __stdcall PushNotificationManager::OnRawNotificationReceived(unsigned int payloadLength, _In_ byte* payload, _In_ HSTRING /*correlationVector */) noexcept try
+    {
+        BOOL foregroundHandled = true;
+        THROW_IF_FAILED(InvokeAll(payloadLength, payload, &foregroundHandled));
+
+        if (!foregroundHandled)
```
hulumane, line 521: @David-Quero Should clean this up

dev/PushNotifications/PushNotificationManager.cpp
hulumane: Where is that logic for Push BTW? It needs to be similar to AppNotification's implementation.

dev/PushNotifications/PushNotifications.idl
```diff
@@ -118,22 +83,22 @@ namespace Microsoft.Windows.PushNotifications
...
+        // For packaged applications with Background Infrastructure support, registers Push Trigger with the Background Infrastructure
+        // For packaged applications without Background Infrastructure support, registers a COM Service
+        void Register();
 
-        // Unregister any activator given PushNotificationRegistrationActivators
-        // 1) If kind = PushTrigger is specified, the trigger itself will be removed
-        // 2) If kind = ComActivator is specified, the Project Reunion Background Task component will no longer act as an InProc COM Server
-        // 3) If kind = ProtocolActivator is specified, the application will be unregistered from the long running process that handles activation
-        void UnregisterActivator(PushNotificationRegistrationActivators activators);
+        // Unregisters the COM Service
```
hulumane, line 92: // Packaged Apps: Force COM to launch a new process
// Unpackaged apps: Remove the Foreground Sink to force Long Running Process Singleton to launch a new process instance.

dev/PushNotifications/PushNotifications.idl
```diff
@@ -118,22 +83,22 @@ namespace Microsoft.Windows.PushNotifications
...
+        void Unregister();
 
-        // Unregister all activators registered for the application
-        void UnregisterAllActivators();
+        // Unregister all services registered for the application
```
hulumane, line 95: Wrong comment. Correct comment:

// Disables Push completely for Packaged and Unpackaged apps. Register() needs to be called again for Push to work.

dev/PushNotifications/PushNotificationManager.cpp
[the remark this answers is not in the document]
sharath2727: Yes pavan this was not done yet. I mentioned paul that I couldn't complete testing for unpackaged as I couldn't complete it on friday. I am working on this issue today and will address it appropriately.

dev/PushNotifications/PushNotificationManager.cpp
sharath2727: @pmpurifoy

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +253,282 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-            }
+                THROW_HR_IF_NULL_MSG(HRESULT_FROM_WIN32(ERROR_NOT_FOUND), m_pushTriggerRegistration, "PushTrigger not registered.");
 
-            s_comActivatorRegistration.reset();
+                m_pushTriggerRegistration.Unregister(true);
+                m_pushTriggerRegistration = nullptr;
+            }
 
-            if (s_protocolRegistration)
+            if (!AppModel::Identity::IsPackagedProcess())
```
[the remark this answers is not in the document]
sharath2727: UnregisterLongRunningActivator -  As part of this API we clean up the sinks in the LRP. So we don't need to call UnregisterForegroundActivator again.

dev/PushNotifications/PushNotificationManager.cpp
sharath2727: @pmpurifoy

pmpurifoy: Unpackaged scenarios will load their background payload in the command line which wouldn't conflict with the foreground scenario. BUT for consistency with packaged scenarios we should add the check.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +253,282 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        bool registeredEvent{ false };
         {
-            return true;
+            auto lock{ m_lock.lock_exclusive() };
+            m_foregroundHandlers.remove(token);
+            registeredEvent = bool(m_foregroundHandlers);
         }
-        else if(!AppModel::Identity::IsPackagedProcess() && isProtocolActivatorSet)
+
+        if (!registeredEvent)
```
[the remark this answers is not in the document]
pmpurifoy: Should all sink management be done by ChannelSuccess/Register/Unregister*? I thought we wanted to make the optimization to remove sinks if there were no longer event handlers?

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: InvokeAll is only used by foreground handlers. InvokeAll isn't the entry point for the app so there doesn't need to be a first notification check. LRP will launch via protocol/cocreate if the foreground handlers fail.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +241,260 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        try
+        {
+            if (PushNotificationHelpers::IsPackagedAppScenario())
             {
+                auto scopeExitToCleanRegistrations{ wil::scope_exit([&]()
                 {
-                    auto lock = s_activatorInfoLock.lock_shared();
-                    THROW_HR_IF(E_INVALIDARG, s_protocolRegistration);
-                }
+                    m_comActivatorRegistration.reset();
```
hulumane, line 279: Missing locks everywhere here.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +241,260 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                    {
+                        m_pushTriggerRegistration.Unregister(true);
+                        m_pushTriggerRegistration = nullptr;
+                    }
 
-                {
-                    auto lock = s_activatorInfoLock.lock_exclusive();
-                    s_protocolRegistration = true;
+                    auto appUserModelId{ PushNotificationHelpers::GetAppUserModelId() };
+                    PushNotifications_UnregisterNotificationSinkForFullTrustApplication(appUserModelId.get());
```
hulumane, line 289: LOG_IF_FAILED_EXPECTED

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +241,260 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+            THROW_HR_IF_MSG(E_FAIL, registering, "Registration is in progress!");
+            registering = true;
+        }
 
-            if (isProtocolActivatorSet)
+        try
+        {
+            if (PushNotificationHelpers::IsPackagedAppScenario())
             {
+                auto scopeExitToCleanRegistrations{ wil::scope_exit([&]()
```
hulumane, line 277: You need a check here: If the app has already registered, simply throw an error. We shouldn't allow multiple registrations.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +241,260 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-            BackgroundTaskBuilder builder{ nullptr };
-
-            if (WI_IsFlagSet(registrationActivators, PushNotificationRegistrationActivators::PushTrigger) && PushNotificationHelpers::IsPackagedAppScenario())
-            {
-                GUID taskClsid = details.TaskClsid();
-                THROW_HR_IF(E_INVALIDARG, taskClsid == GUID_NULL);
+                )};
 
+                // Register a PushTrigger for BI to activate the application through COM
+                if (!m_pushTriggerRegistration)
```
hulumane, line 294: Remove this if check. There should be a check earlier in IsPackagedScenario() to ensure that this isn't already set and throw if it is.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +241,260 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-                }
-
-                winrt::hstring taskClsidStr{ winrt::to_hstring(taskClsid) };
-                winrt::hstring backgroundTaskFullName{ backgroundTaskName + taskClsidStr };
-
-                auto tasks{ BackgroundTaskRegistration::AllTasks() };
-                bool isTaskRegistered{ std::any_of(std::begin(tasks), std::end(tasks),
-                    [&](auto&& task)
+                    winrt::hstring backgroundTaskFullName{ backgroundTaskName + winrt::to_hstring(m_registeredClsid) };
+                    if (!IsBackgroundTaskRegistered(backgroundTaskFullName))
```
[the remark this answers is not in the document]
pmpurifoy: That's done in IsBackgroundTaskRegistered

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +241,260 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                }
 
-                    THROW_HR_IF(E_NOTIMPL, !AppModel::Identity::IsPackagedProcess());
+                scopeExitToCleanRegistrations.release();
+            }
+            else
+            {
+                auto scopeExitToCleanRegistrations{ wil::scope_exit([&]()
+                {
+                    m_comActivatorRegistration.reset();
```
hulumane, line 333: no locks anywhere

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +241,260 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                        REGCLS_MULTIPLEUSE,
+                        &m_comActivatorRegistration));
+                }
 
-                    THROW_HR_IF(E_NOTIMPL, !AppModel::Identity::IsPackagedProcess());
+                scopeExitToCleanRegistrations.release();
+            }
+            else
+            {
+                auto scopeExitToCleanRegistrations{ wil::scope_exit([&]()
```
hulumane, line 331: You need to throw here if Register is already done. We wont allow 2 registrations

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +241,260 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
 
-                // AppId generated by PushNotificationsLongRunningTask singleton
-                wil::unique_cotaskmem_string unpackagedAppUserModelId;
-                RegisterUnpackagedApplicationHelper(GUID_NULL, unpackagedAppUserModelId); // create default registration for app
+                    // Clean the task registration only if it was created during this call
+                    if (m_pushTriggerRegistration)
+                    {
+                        m_pushTriggerRegistration.Unregister(true);
+                        m_pushTriggerRegistration = nullptr;
+                    }
```
hulumane, line 286: Both the scope cleanups need to set registering to false.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +241,260 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                        auto appUserModelId{ PushNotificationHelpers::GetAppUserModelId() };
+                        PushNotifications_UnregisterNotificationSinkForFullTrustApplication(appUserModelId.get());
                     }
-                }
-            }
+                    else
+                    {
+                        notificationsLongRunningPlatform->UnregisterForegroundActivator(m_processName.get());
+                    }
+                    notificationsLongRunningPlatform->UnregisterFullTrustApplication(m_processName.get());
```
hulumane, line 346: What does this do? Does it Unregister the app from the platform?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +241,260 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-                    auto builder5{ builder.as<winrt::IBackgroundTaskBuilder5>() };
-                    builder5.SetTaskEntryPointClsid(taskClsid);
-                    winrt::com_array<winrt::IBackgroundCondition> conditions = details.GetConditions();
-                    for (auto condition : conditions)
+                    auto notificationsLongRunningPlatform{ PushNotificationHelpers::GetNotificationPlatform() };
+                    if (PushNotificationHelpers::IsPackagedAppScenario())
                     {
-                        builder.AddCondition(condition);
+                        auto appUserModelId{ PushNotificationHelpers::GetAppUserModelId() };
+                        PushNotifications_UnregisterNotificationSinkForFullTrustApplication(appUserModelId.get());
```
hulumane, line 340: If the method returns HRESULT, we need a LOG_IF_FAILED at the minimal.

hulumane: Here and everywhere else below

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +241,260 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                    m_comActivatorRegistration.reset();
+                    m_protocolRegistration = false;
 
-                    // In case the interface is not supported, let it throw.
-                    auto builder5{ builder.as<winrt::IBackgroundTaskBuilder5>() };
-                    builder5.SetTaskEntryPointClsid(taskClsid);
-                    winrt::com_array<winrt::IBackgroundCondition> conditions = details.GetConditions();
-                    for (auto condition : conditions)
+                    auto notificationsLongRunningPlatform{ PushNotificationHelpers::GetNotificationPlatform() };
+                    if (PushNotificationHelpers::IsPackagedAppScenario())
```
hulumane, line 337: This check doesn't line up with the if condition which presumes that this is UnpackagedAppScenario. What am i missing here?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +241,260 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                    }
+                    notificationsLongRunningPlatform->UnregisterFullTrustApplication(m_processName.get());
 
-            BackgroundTaskRegistration registeredTaskFromBuilder{ nullptr };
+                }
+                ) };
 
-            auto scopeExitToCleanRegistrations{ wil::scope_exit(
-                [&]()
+                if (!m_protocolRegistration)
```
hulumane, line 351: This is the wrong check. m_protocolRegistration is true for unpackaged apps only.

hulumane, line 351: Please remove this. For unpackagedAppScenario, you will have 2 blocks. IsPackaged() and IsUnpackaged(). I'm not sure what this is.

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: Ooops. Changing this to IsPackagedProcess

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -261,155 +256,221 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-            
-            // Check for COM flag, a valid cookie
-            if (AppModel::Identity::IsPackagedProcess())
+            if (PushNotificationHelpers::IsPackagedAppScenario())
             {
-                THROW_HR_IF_MSG(HRESULT_FROM_WIN32(ERROR_NOT_FOUND), !m_comActivatorRegistration, "ComActivator not registered.");
-                m_comActivatorRegistration.reset();
+                {
+                    auto lock{ m_lock.lock_exclusive() };
+                    THROW_HR_IF_MSG(E_FAIL, !m_comActivatorRegistration, "No active COM registration");
```
danielayala94, line 444: E_UNEXPECTED may be a better error code here.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -261,155 +256,221 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                }
+
+                {
+                    auto lock{ m_lock.lock_exclusive() };
+                    THROW_HR_IF_MSG(E_FAIL, !m_singletonForegroundRegistration, "No foreground sink registered");
+                }
+
+                auto notificationsLongRunningPlatform{ PushNotificationHelpers::GetNotificationPlatform() };
+                // Unregister foreground sink with the Long Running Process Singelton
+                THROW_IF_FAILED(notificationsLongRunningPlatform->UnregisterForegroundActivator(m_processName.get()));
```
danielayala94, line 468: There is no guard for `m_processName`. Probably you can have a local processName copy.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -261,155 +256,221 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+
+                auto appUserModelId{ PushNotificationHelpers::GetAppUserModelId() };
+                THROW_IF_FAILED(PushNotifications_UnregisterNotificationSinkForFullTrustApplication(appUserModelId.get()));
+            }
+            else
+            {
+                if (AppModel::Identity::IsPackagedProcess())
+                {
+                    auto lock{ m_lock.lock_exclusive() };
+                    THROW_HR_IF_MSG(E_FAIL, !m_comActivatorRegistration, "No active COM registration");
```
danielayala94, line 457: Are developers going to see these messages somewhere? If yes, we should try to see something more generic or understandable. The developer will not have any idea of what "COM registration" or "foreground sink" means, especially after having removed the registration flags.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -423,49 +484,89 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
 
-        try
+        if (!unregisterCompleted)
         {
             Unregister();
+        }
 
+        {
             auto lock{ m_lock.lock_exclusive() };
+            THROW_HR_IF_MSG(E_FAIL, registering, "Register or Unregister currently in progress!");
```
danielayala94, line 500: Consider ERROR_OPERATION_IN_PROGRESS instead (or any error code trying to tell the developer that unregistration is in progress, this is the best I found so far 😊).

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -423,49 +484,89 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+            unregisterCompleted = (!m_singletonForegroundRegistration || !m_comActivatorRegistration);
+        }
 
-        try
+        if (!unregisterCompleted)
         {
             Unregister();
+        }
 
+        {
```
danielayala94, line 498: Should this block be the very first thing we do in this function?

danielayala94 (changes requested):

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: m_processName is set in the PushNotificationManager constructor. This is the only time the member variable is touched. Essentially it is a constant for the class.

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: Yup. All these error messages will show up for the developer.

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: If that is set at the beginning of the functions, Unregister will throw because m_registering will be true :(

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: Changing message to "Need to call Register() before calling Unregister()"

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -167,24 +185,32 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-                        THROW_IF_FAILED(CreateChannelWithRemoteIdHelper(unpackagedAppUserModelId, remoteId, channelInfo));
+                        THROW_IF_FAILED(notificationPlatform->RegisterFullTrustApplication(m_processName.get(), remoteId, &unpackagedAppUserModelId));
 
-                        auto notificationPlatform{ PushNotificationHelpers::GetNotificationPlatform() };
+                        THROW_IF_FAILED(CreateChannelWithRemoteIdHelper(unpackagedAppUserModelId, remoteId, channelInfo));
 
-                        wil::unique_cotaskmem_string processName;
-                        THROW_IF_FAILED(GetCurrentProcessPath(processName));
-                        THROW_IF_FAILED(notificationPlatform->RegisterLongRunningActivatorWithClsid(processName.get(), GetComServerClsid()));
+                        THROW_IF_FAILED(notificationPlatform->RegisterLongRunningActivatorWithClsid(m_processName.get(), m_registeredClsid));
```
hulumane, line 196: Add a comment here to explain why we need to do this after channel creation. There is a legit reason which folks reading the code for the first time won't understand.

hulumane: Also can this be moved to RegisterForegroundSinkHelper?

hulumane, line 196: no locks for m_processName :(

hulumane: Here and everywhere above

pmpurifoy: yes we can move this into RegisterForegroundSinkHelper

pmpurifoy: https://github.com/microsoft/WindowsAppSDK/pull/2122#discussion_r814358341 

Here the discussion between Daniel and me. What do you think here?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +247,365 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-                        {
-                            s_pushTriggerRegistration = task.Value();
-                            return true;
-                        }
+                        m_registering = false;
+                    }
 
-                        throw winrt::hresult_invalid_argument(L"RegisterActivator has different clsid registered.");
-                    }) };
+                    auto appUserModelId{ PushNotificationHelpers::GetAppUserModelId() };
```
hulumane, line 306: This should be done for only PackagedAppScenario

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +247,365 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-                if (!isTaskRegistered)
+                winrt::hstring backgroundTaskFullName;
                 {
-                    builder = BackgroundTaskBuilder();
+                    auto lock{ m_lock.lock_shared() };
+                    backgroundTaskFullName = backgroundTaskName + winrt::to_hstring(m_registeredClsid);
+                }
+
+                // Register a PushTrigger for BI to activate the application through COM
+                if (!IsBackgroundTaskRegistered(backgroundTaskFullName))
```
hulumane, line 318: If the background task is already registered by a previous instance of the process, then what? m_pushTriggerRegistration would be nullptr due to this check

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +247,365 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-                }
-            ) };
 
-            if (WI_IsFlagSet(registrationActivators, PushNotificationRegistrationActivators::ComActivator))
-            {
-                GUID taskClsid = details.TaskClsid();
-                THROW_HR_IF(E_INVALIDARG, taskClsid == GUID_NULL);
+                    auto notificationsLongRunningPlatform{ PushNotificationHelpers::GetNotificationPlatform() };
+                  
+                    LOG_IF_FAILED(notificationsLongRunningPlatform->UnregisterForegroundActivator(m_processName.get()));
```
hulumane, line 372: locks?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +247,365 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-            if (builder)
-            {
-                registeredTaskFromBuilder = builder.Register();
-            }
+                // Register a COM object for the PushNotificationLongRunningProcess to CoCreate in background activation scenarios
+                if (AppModel::Identity::IsPackagedProcess())
+                {
+                    auto lock{ m_lock.lock_exclusive() };
+                    THROW_IF_FAILED(::CoRegisterClassObject(
+                        m_registeredClsid,
```
hulumane, line 401: lock?

pmpurifoy: There is a lock right before CoRegisterClassObject

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: https://github.com/microsoft/WindowsAppSDK/pull/2122#discussion_r814358341

Let me know what you think.

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: m_pushTriggerRegistration was being set in IsBackgroundTaskRegistered. Making another helper function to retrieve the backgroundTaskRegistration for more transparency.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +272,371 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
     }
 
-    void PushNotificationManager::UnregisterAllActivators()
+    void PushNotificationManager::UnregisterAll()
     {
-        THROW_HR_IF(E_NOTIMPL, !::Microsoft::Windows::PushNotifications::Feature_PushNotifications::IsEnabled());
+        bool unregisterCompleted{ false };
+        {
+            auto lock{ m_lock.lock_exclusive() };
+            unregisterCompleted = (!m_singletonForegroundRegistration || !m_comActivatorRegistration);
```
robertzhou-wpn, line 520: Can you please help me to understand this OR condition? for  PackagedAppScenario, in unregister(), looks like we do not change m_singletonForegroundRegistration at all, if  m_singletonForegroundRegistration is always false for packaged AppScenario, looks like unregisterCompleted will always be true here.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +272,371 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-            PushNotificationTelemetry::ActivatorRegisteredByApi(S_OK, details.Activators());
         }
-
-        catch(...)
+        catch (...)
         {
-            PushNotificationTelemetry::ActivatorRegisteredByApi(wil::ResultFromCaughtException(),
-                details == nullptr ? PushNotificationRegistrationActivators::Undefined : details.Activators());
+            PushNotificationTelemetry::ActivatorRegisteredByApi(wil::ResultFromCaughtException());
             throw;
```
robertzhou-wpn, line 446: Do we need to reset m_registering to false before throw?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +272,371 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        }
     }
 
-    void PushNotificationManager::UnregisterActivator(PushNotificationRegistrationActivators const& activators)
+    void PushNotificationManager::Unregister()
     {
-        THROW_HR_IF(E_NOTIMPL, !::Microsoft::Windows::PushNotifications::Feature_PushNotifications::IsEnabled());
+        {
+            auto lock{ m_lock.lock_exclusive() };
+            THROW_HR_IF_MSG(E_FAIL, m_registering, "Register or Unregister currently in progress!");
```
robertzhou-wpn, line 459: rename m_registering?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -34,11 +36,6 @@ wil::unique_event& GetWaitHandleForArgs()
     return g_waitHandleForArgs;
```
hulumane, line 36: Why do we have this static? Can we move it to be a member variable?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -20,11 +20,13 @@
...
 #include "AppNotificationUtility.h"
+#include "PushNotificationReceivedEventArgs.h"
 
 using namespace std::literals;
 using namespace Microsoft::Windows::AppNotifications::Helpers;
 
 constexpr std::wstring_view backgroundTaskName = L"PushBackgroundTaskName"sv;
+constexpr std::wstring_view expectedPushServerArgs = L"----WindowsAppRuntimePushServer:"sv;
 
 static wil::unique_event g_waitHandleForArgs;
```
hulumane, line 31: Please get rid of these statics

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -164,32 +205,42 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
                     PushNotificationTelemetry::ChannelRequestedByApi(S_OK, remoteId);
                     co_return winrt::make<PushNotificationCreateChannelResult>(
-                        winrt::make<PushNotificationChannel>(channelInfo),
+                        channel,
                         S_OK,
                         PushNotificationChannelStatus::CompletedSuccess);
                 }
                 catch (...)
                 {
+                    UnregisterSinkHelper(); // Need to clean up sinks if failure occurs.
```
hulumane, line 243: This should not be here. LEts not do this at all please

hulumane: It needs a scope exit and a dismiss pattern

hulumane: And probably outside the try catch. I would have a bool and use that to determine if channel request succeeded. If something failed, then cleanup after the catch

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +272,371 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-            THROW_HR_IF(E_INVALIDARG, isProtocolActivatorSet && !IsActivatorSupported(registrationActivators));
-
-            if (isProtocolActivatorSet)
+        auto tasks{ BackgroundTaskRegistration::AllTasks() };
+        bool isTaskRegistered{ std::any_of(std::begin(tasks), std::end(tasks),
+            [&](auto&& task)
             {
+                auto name{ task.Value().Name() };
+
+                if (name == backgroundTaskFullName)
```
hulumane, line 283: return name == backgroundTaskFullName;

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +272,371 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-                        {
-                            s_pushTriggerRegistration = task.Value();
-                            return true;
-                        }
+                        m_registering = false;
+                    }
 
-                        throw winrt::hresult_invalid_argument(L"RegisterActivator has different clsid registered.");
-                    }) };
+                    UnregisterSinkHelper();
```
hulumane, line 340: This will throw if no sinks are registered. Better off having a seperate scope for this right before you call RegisterSinkHelper.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +272,371 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                        m_registeredClsid,
+                        winrt::make<PushNotificationBackgroundTaskFactory>().get(),
+                        CLSCTX_LOCAL_SERVER,
+                        REGCLS_MULTIPLEUSE,
+                        &m_comActivatorRegistration));
+                }
 
-            auto scopeExitToCleanRegistrations{ wil::scope_exit(
-                [&]()
+                scopeExitToCleanRegistrations.release();
```
hulumane, line 389: dismiss or release?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +272,371 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                    UnregisterSinkHelper();
+
+                    auto notificationsLongRunningPlatform{ PushNotificationHelpers::GetNotificationPlatform() };
+                    LOG_IF_FAILED(notificationsLongRunningPlatform->UnregisterFullTrustApplication(m_processName.get()));
+                }) };
+
+                auto notificationPlatform{ PushNotificationHelpers::GetNotificationPlatform() };
+                wil::unique_cotaskmem_string unpackagedAppUserModelId;
+                // Apps treated as unpackaged need to call RegisterFullTrustApplication and register with the LRP
+                THROW_IF_FAILED(notificationPlatform->RegisterFullTrustApplication(m_processName.get(), GUID_NULL, &unpackagedAppUserModelId));
```
hulumane, line 418: Bug: You are passing GUID_NULL here instead of the GUID activator

dev/PushNotifications/PushNotificationManager.cpp
hulumane: In general, the patern of throwing in a catch should be discouraged. 

Instead we should catch the hresult here and throw outside the try - catch

dev/PushNotifications/PushNotificationManager.h
```diff
@@ -4,22 +4,49 @@
...
+        winrt::Microsoft::Windows::PushNotifications::PushNotificationManager,
+        winrt::Microsoft::Windows::PushNotifications::PushNotificationReceivedEventArgs> PushNotificationEventHandler;
 
-    struct PushNotificationManager : PushNotificationManagerT<PushNotificationManager>
+    struct PushNotificationManager : PushNotificationManagerT<PushNotificationManager, IWpnForegroundSink, ABI::Microsoft::Internal::PushNotifications::INotificationListener>
     {
-        PushNotificationManager() = default;
+        PushNotificationManager();
 
         static winrt::Microsoft::Windows::PushNotifications::PushNotificationManager Default();
```
BenJKuhn, line 20: Be mindful when putting COM / WinRT types into static variables. Unless you control for it, there may be situations where COM forcibly running down (e.g. at process exit) can lead to issues with the DLL that implements held objects (think transitive, not just direct reference) being unloaded before CRT cleanup occurs, leading to dereferencing unloaded code.

dev/PushNotifications/PushNotificationManager.cpp
sharath2727: I think in the meeting we decided it is fine for now. :)

dev/PushNotifications/PushNotificationManager.cpp
sharath2727: @pmpurifoy

dev/PushNotifications/PushNotificationManager.cpp
sharath2727: @pmpurifoy

dev/PushNotifications/PushNotificationManager.cpp
sharath2727: @pmpurifoy

dev/PushNotifications/PushNotificationManager.cpp
sharath2727: @pmpurifoy  or you could also use the m_singletonForegroundRegistration flags?

dev/PushNotifications/PushNotificationManager.cpp
sharath2727: it is release there is no member variable called dismiss. Thanks.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +272,371 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-                    auto lock = s_activatorInfoLock.lock_exclusive();
-                    s_protocolRegistration = true;
-                }
+        return isTaskRegistered;
+    }
+
+    IBackgroundTaskRegistration RetreiveBackgroundTaskRegistration(winrt::hstring const& backgroundTaskFullName)
+    {
+        auto tasks{ BackgroundTaskRegistration::AllTasks() };
+        for (auto task : tasks)
```
hulumane, line 297: Can you use the std libraries to iterate through this based on the conditional?

dev/PushNotifications/PushNotificationManager.cpp
hulumane: I wouldnt recommend the same flag to override. If you put the new scope next to the actual RegisterSink call, it would guarantee that cleanup is safer

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +272,371 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                        m_registering = false;
                     }
-                }
-            ) };
 
-            if (WI_IsFlagSet(registrationActivators, PushNotificationRegistrationActivators::ComActivator))
-            {
-                GUID taskClsid = details.TaskClsid();
-                THROW_HR_IF(E_INVALIDARG, taskClsid == GUID_NULL);
+                    UnregisterSinkHelper();
```
hulumane, line 409: same issue as above.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +272,371 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-            auto lock{ s_activatorInfoLock.lock_exclusive() };
-            if (WI_IsFlagSet(activators, PushNotificationRegistrationActivators::PushTrigger))
+            if (PushNotificationHelpers::IsPackagedAppScenario())
             {
-                THROW_HR_IF_NULL_MSG(HRESULT_FROM_WIN32(ERROR_NOT_FOUND), s_pushTriggerRegistration, "PushTrigger not registered.");
-                s_pushTriggerRegistration.Unregister(true);
-                s_pushTriggerRegistration = nullptr;
-            }
+                {
+                    auto lock{ m_lock.lock_exclusive() };
```
hulumane, line 474: This is common code for all packaged apps. So you don't have to repeat twice.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +272,371 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-                s_comActivatorRegistration.reset();
+                UnregisterSinkHelper();
             }
-
-            if (WI_IsFlagSet(activators, PushNotificationRegistrationActivators::ProtocolActivator))
+            else
             {
-                THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_NOT_FOUND), !s_protocolRegistration);
-                auto coInitialize = wil::CoInitializeEx();
+                if (AppModel::Identity::IsPackagedProcess())
```
hulumane, line 484: See comment above

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +272,371 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                    // m_comActivatorRegistration handles multiple unregistrations
+                    m_comActivatorRegistration.reset();
+                }
 
-            // Check for COM flag, a valid cookie
-            if (WI_IsFlagSet(activators, PushNotificationRegistrationActivators::ComActivator))
-            {
-                THROW_HR_IF_MSG(HRESULT_FROM_WIN32(ERROR_NOT_FOUND), !s_comActivatorRegistration, "ComActivator not registered.");
-                s_comActivatorRegistration.reset();
+                UnregisterSinkHelper();
```
hulumane, line 480: You just destroyed Push permanently.

hulumane: Instead we need to just remove the foreground activator

hulumane: And the bool needs to be set after we do that

dev/PushNotifications/PushNotificationManager.cpp
hulumane: A lot of this logic needs to be simplified. Unregister seems to destroy state in its current form due to a bug. This check isn't clear at all. Lets not do this. Prevent races between REgister, Unregister and UnregisterAll by using a single bool.

And have simple bool checks to ensure that Unregister is done via THROW_IF checks

hulumane: It should be
if (m_singletonForegroundregistration || m_comActivationREgistration)
{
    Unregister();
}

hulumane: Since the presence of one of them implies that Unregister wasn't called

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +272,371 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
 
         try
         {
-            auto lock{ s_activatorInfoLock.lock_exclusive() };
-            if (s_pushTriggerRegistration)
+            try
             {
-                s_pushTriggerRegistration.Unregister(true);
-                s_pushTriggerRegistration = nullptr;
+                auto lock{ m_lock.lock_exclusive() };
```
hulumane, line 544: Why is this exclusive lock?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +272,371 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+            [&] {
+                auto lock{ m_lock.lock_exclusive() };
+                m_registering = false;
+            });
 
         try
         {
-            auto lock{ s_activatorInfoLock.lock_exclusive() };
-            if (s_pushTriggerRegistration)
+            try
```
hulumane, line 542: Remove try catch. Looks like the implementation of close() swallows the error for E_NOT_SET

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +272,371 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-                LOG_IF_FAILED(notificationPlatform->UnregisterLongRunningActivator(processName.get()));
+                // Removes the Long Running Singleton sink registered with platform for both packaged and unpackaged applications 
+                THROW_IF_FAILED(notificationsLongRunningPlatform->UnregisterLongRunningActivator(m_processName.get()));
+
+                THROW_IF_FAILED(notificationsLongRunningPlatform->UnregisterFullTrustApplication(m_processName.get()));
 
-                s_protocolRegistration = false;
+                {
+                    auto lock{ m_lock.lock_exclusive() };
+                    m_singletonBackgroundRegistration = false;
```
hulumane, line 582: Set this to false after you call UnregisterLongRunningActivator

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +272,371 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                }
+            }
+            else
             {
-                auto coInitialize = wil::CoInitializeEx();
+                auto notificationsLongRunningPlatform{ PushNotificationHelpers::GetNotificationPlatform() };
 
-                auto notificationPlatform{ PushNotificationHelpers::GetNotificationPlatform() };
+                {
+                    auto lock{ m_lock.lock_exclusive() };
```
hulumane, line 571: Why is this exclusive?

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: waitHandle is used in GetNotificationEventArgs.h to receive the args from background. I have an idea which we move PushDeserialize in PushNotificationManager which then we could turn waitHandle that into a member variable

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: We could LOG_IF_FAILED here. I'm just worried about all these scope_exits not really making our code more simple.

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: This sets up a default registration to be overrode by CreateChannelAsync who actually has the remoteId.

dev/PushNotifications/PushNotificationManager.cpp
sharath2727: if we remove handle set the foregroundhandled = bool(foregroundhandlers)

dev/PushNotifications/PushNotificationManager.cpp
danielayala94: While I understand the reasoning behind not locking, as a good habit I would still scope it.

danielayala94: Taking an anti-pattern approach should be avoided unless really needed.

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: Added scope_exit after sink registrations

dev/PushNotifications/PushNotificationManager.cpp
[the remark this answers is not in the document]
pmpurifoy: Fixed sink cleanup

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: We can make this a bug for improvement.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +232,412 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+    IBackgroundTaskRegistration RetreiveBackgroundTaskRegistration(winrt::hstring const& backgroundTaskFullName)
+    {
+        auto tasks{ BackgroundTaskRegistration::AllTasks() };
+        IBackgroundTaskRegistration registration{};
+
+        // Custom version of std::find since we are looking for the name
+        // of the task not the registration.
+        auto first{ std::begin(tasks) };
+        auto last{ std::end(tasks) };
+        while (first != last)
```
hulumane, line 258: So there is no expression you can use to achieve the same without introducing a loop?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +232,412 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                    THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_ALREADY_REGISTERED), m_comActivatorRegistration || m_pushTriggerRegistration);
                 }
 
-                winrt::hstring taskClsidStr{ winrt::to_hstring(taskClsid) };
-                winrt::hstring backgroundTaskFullName{ backgroundTaskName + taskClsidStr };
-
-                auto tasks{ BackgroundTaskRegistration::AllTasks() };
-                bool isTaskRegistered{ std::any_of(std::begin(tasks), std::end(tasks),
-                    [&](auto&& task)
+                auto scopeExitToCleanRegistrations{ wil::scope_exit([&]()
```
hulumane, line 294: name collision with scope exit above. It will compile but please rename to avoid confusion.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +232,412 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-                    // Clean the task registration only if it was created during this call
-                    if (registeredTaskFromBuilder)
+                scopeExitToCleanSink.release();
+                scopeExitToCleanRegistrations.release();
+            }
+            else
+            {
+                {
+                    auto lock{ m_lock.lock_shared() };
+                    THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_ALREADY_REGISTERED), m_singletonForegroundRegistration || m_singletonLongRunningSinkRegistration);
```
hulumane, line 368: what about m_comRegistration?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +232,412 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                scopeExitToCleanRegistrations.release();
+            }
+        }
+        catch (...)
         {
-            PushNotificationTelemetry::ActivatorRegisteredByApi(wil::ResultFromCaughtException(),
-                details == nullptr ? PushNotificationRegistrationActivators::Undefined : details.Activators());
-            throw;
+            hr = wil::ResultFromCaughtException();
+            PushNotificationTelemetry::ActivatorRegisteredByApi(hr);
```
hulumane, line 451: you are logging twice. Here and below

hulumane: instead log once before you throw

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +232,412 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        THROW_IF_FAILED(hr);
+        PushNotificationTelemetry::ActivatorRegisteredByApi(hr);
     }
 
-    void PushNotificationManager::UnregisterActivator(PushNotificationRegistrationActivators const& activators)
+    void PushNotificationManager::Unregister()
     {
-        THROW_HR_IF(E_NOTIMPL, !::Microsoft::Windows::PushNotifications::Feature_PushNotifications::IsEnabled());
-
+        hresult hr = S_OK;
```
hulumane, line 460: winrt::hresult hr{ S_OK };

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: If it fails, it will log and throw outside the try/catch. Otherwise it will log success.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +232,412 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                    m_singletonForegroundRegistration = false;
+                }
             }
         }
         catch (...)
         {
-            PushNotificationTelemetry::ActivatorUnregisteredByApi(wil::ResultFromCaughtException(), activators);
-            throw;
+            hr = wil::ResultFromCaughtException();
+            PushNotificationTelemetry::ActivatorUnregisteredByApi(hr);
```
hulumane, line 509: calling twice again

pmpurifoy: If it fails, it will log and throw outside the try/catch. Otherwise it will log success.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +232,412 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
 
-                wil::unique_cotaskmem_string processName;
-                THROW_IF_FAILED(GetCurrentProcessPath(processName));
-                LOG_IF_FAILED(notificationPlatform->UnregisterLongRunningActivator(processName.get()));
+                m_pushTriggerRegistration.Unregister(true);
+                m_pushTriggerRegistration = nullptr;
+            }
+            else
+            {
+                auto notificationsLongRunningPlatform{ PushNotificationHelpers::GetNotificationPlatform() };
```
hulumane, line 565: Move close to usage

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: Replacing with std::find_if

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -221,249 +232,412 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
             }
         }
-        catch(...)
+        catch (...)
         {
-            PushNotificationTelemetry::ActivatorUnregisteredByApi(wil::ResultFromCaughtException(),
-                PushNotificationRegistrationActivators::PushTrigger | PushNotificationRegistrationActivators::ComActivator);
-            throw;
+            hr = wil::ResultFromCaughtException();
+            PushNotificationTelemetry::ActivatorUnregisteredByApi(hr);
```
hulumane, line 588: Remove this and use it once only please

hulumane (approved):

dev/PushNotifications/PushNotificationManager.h
pmpurifoy: Filed a bug to address this issue: https://github.com/microsoft/WindowsAppSDK/issues/2183

danielayala94 (approved):

PR #2125, opened 2022-02-18
Modify the CoRegister for Com Activation so if there are not notification handlers registered a new process will be launched every time

David-Quero:
When registering an activator for Toast check if there are handlers registered with the Toast Notification Manager, if so then use the multi use flag in CoRegister otherwise use the single use flag to launch a new process each time.


dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -90,7 +90,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
                 AppModel::Identity::IsPackagedProcess() ? registeredClsid : winrt::guid(storedComActivatorString),
                 winrt::make<AppNotificationActivationCallbackFactory>().get(),
                 CLSCTX_LOCAL_SERVER,
-                REGCLS_MULTIPLEUSE,
+                m_notificationHandlers ? REGCLS_MULTIPLEUSE : REGCLS_SINGLEUSE,
```
hulumane, line 93: Add a comment here along the lines of:
// If the developer has registered event handlers, COM will ensure that all activators are routed inproc else COM will launch a new process of the win32 app for subsequent invocations

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -90,7 +90,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
                 AppModel::Identity::IsPackagedProcess() ? registeredClsid : winrt::guid(storedComActivatorString),
                 winrt::make<AppNotificationActivationCallbackFactory>().get(),
                 CLSCTX_LOCAL_SERVER,
-                REGCLS_MULTIPLEUSE,
```
hulumane, line 93: Please work with @pmpurifoy  to test both the cases using the demo app.

PR #2128, opened 2022-02-18
Modify the CoRegister for Com Activation so if there are not notification handlers registered a new process will be launched every time

David-Quero:
When registering an activator for Toast check if there are handlers registered with the Toast Notification Manager, if so then use the multi use flag in CoRegister otherwise use the single use flag to launch a new process each time.

### DIFF
--- a/dev/AppNotifications/AppNotificationManager.cpp
+++ b/dev/AppNotifications/AppNotificationManager.cpp
@@ -86,6 +86,9 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
         {
             auto lock{ m_lock.lock_exclusive() };
             THROW_HR_IF_MSG(E_INVALIDARG, m_notificationComActivatorRegistration, "Already Registered for App Notifications!");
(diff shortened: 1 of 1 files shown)


dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -86,11 +86,14 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
         {
             auto lock{ m_lock.lock_exclusive() };
             THROW_HR_IF_MSG(E_INVALIDARG, m_notificationComActivatorRegistration, "Already Registered for App Notifications!");
+
+            // Check if the caller has registered event handlers, if so the REGCLS_MULTIPLEUSE flag will cause COM to ensure that all activators
+            // are routed improc, otherwise with REGCLS_MULTIPLEUSE COM will launch a new process of the Win32 app for each invocation.
```
hulumane, line 91: typo: inproc.

PR #2135, opened 2022-02-19
IsSelfContained()

DrusTheAxe:
Function for implementation to detect if WinAppSDK is in use via Self-Contained (i.e. if WinAppSDK is deployed via MSIX or Self-Contained).

https://task.ms/37999189


dev/Common/Microsoft.WindowsAppSDK.SelfContained.h
```diff
@@ -0,0 +1,41 @@
...
+namespace Microsoft::WindowsAppSDK::SelfContained
+{
+inline bool IsSelfContained()
+{
+    const UINT32 flags{ PACKAGE_FILTER_HEAD | PACKAGE_FILTER_DIRECT | PACKAGE_FILTER_STATIC | PACKAGE_FILTER_DYNAMIC | PACKAGE_INFORMATION_BASIC };
+    const PackagePathType packagePathType{ PackagePathType_Install }; // Don't need the path so use the leanest possible answer
+    uint32_t packageInfoCount{};
+    const PACKAGE_INFO* packageInfo{};
+    wil::unique_cotaskmem_ptr<BYTE[]> buffer;
+    THROW_IF_FAILED(GetCurrentPackageInfo(flags, packagePathType, packageInfoCount, packageInfo, buffer));
```
riverar, line 25: I believe this will throw if a package graph doesn't exist (e.g. running my custom app during oobe). I believe Task Manager is broken today due to a similar assumption.

Given self-contained apps may not have identity or have a graph, and that's considered valid, I don't think failures from `GetCurrentPackageInfo` should be considered a failure. Instead, perhaps you should swallow the error here and return `false`.

hulumane: Sounds like a good default for the THROW case then is to assume "Not Self Contained"?

Instead, can we fix the API "GetCurrentPackageInfo" to not throw and simply return an empty list? That way, this code would work without complexities.

riverar: Yep. `GetCurrentPackageInfo` doesn't throw. It returns Win32 error codes like `APPMODEL_ERROR_NO_PACKAGE`.

BenJKuhn: @hulumane, I would think lack of a package graph (failure case)  is more likely to mean that App SDK *is* self-contained (since self-contained can run against a classic win32-style manifest without any package references, whereas using the package inherently requires a package graph.

hulumane: Ah. In that case, e could just special case on this error code and return false?

hulumane: Or true depending on what we believe is the right default.

DrusTheAxe: This is an internal function. We don't throw across major boundaries (e.g. WinRT/COM/Flat-C APIs) but this isn't one.

And as a general rule, 'APIs' like this should NOT swallow errors. As a general rule one shouldn't swallow errors, and when an 'error' is merely a 'failure'  (vs 'success') is a _contextual_ judgement. Something we lack here. This can be used by many contexts, it's their call to decide how to handle errors (which could include swallowing)

As for `APPMODEL_ERROR_NO_PACKAGE` yes. I intentionally didn't special case that out yet as I need to add tests. Left it out here so they initially fail, as a reminder-to-self.

>Sounds like a good default for the THROW case then is to assume "Not Self Contained"?

That's not recommended. If you call IsSelfContained() and it errors, it throws an exception. You need to decide how to deal with that. If you're using it in, say, `foo.IsSupported() { return IsSelfContained() }` and `foo` is a WinRT object then the caller will get an error (or exception, depending if they're using ABI, C++/WinRT projection etc). That's probably the appropriate answer.

If you don't like the classic FileExists=...

* exists(success)
* not-exists(failure)
* badpath/diskerror/outofmemory/yaddayadda(error)

Then try the telephone game. When you make a call the possible outcomes are...

* Connection! Someone picked up the phone. Hi Mom! (success)
* No connection! Ring ring ring... no one picks up the phone (failure)
* Problem! Busy signal, "This number is no longer in service", ... (error)

APIs should avoid swallowing errors as they rarely know their caller's context. If I called someone I'd much prefer "This number is no longer in service" than have the phone company silently swallow the error and just let it ring forever (mapping 'error' to 'failure').

>can we fix the API "GetCurrentPackageInfo" to not throw and simply return an empty list?

`GetCurrentPackageInfo` is a public API since Win8. Changing behavior's not an option.

But that's OK. This is an implementation detail and as noted above, intentionally not _yet_ handled. It will be.

Longer-term we can invent new APIs with more modern sensibilities e.g. `GetCurrentPackageInfo` predates WIL and certain coding standards so a more modern `GetPackageGraph` returning HRESULT for errors, count=0 for no package graph, etc would be welcome. New APIs via WinAppSDK of course ;-) Something we've talked about. I thought there was an Issue for this but will double check and if not, create one, for tracking purposes.

DrusTheAxe: > @hulumane, I would think lack of a package graph (failure case) is more likely to mean that App SDK _is_ self-contained (since self-contained can run against a classic win32-style manifest without any package references, whereas using the package inherently requires a package graph.

`APPMODEL_ERROR_NO_PACKAGE` == `Count=0`. The implementation isn't trapping for that _yet_. I wanted to stand up tests first (GetCurrentPackageInfo has some sharp edge cases depending how you call it).

riverar: > That's not recommended. If you call IsSelfContained() and it errors, it throws an exception. [...]

So you're saying:
* If your app is self-contained, this will return true
* If your app is not self-contained, this will either throw or return false

Mixing boolean retval with SEH semantics seems less than ideal.

DrusTheAxe: This isn't an API. It's an internal implementation detail.

If this were an API then yes the footprint should and would differ e.g.

```
STDAPI WindowsAppSDK_IsSelfContained(BOOL*isSelfContained);
```

AKA


```
extern "C" HRESULT __stdcall WindowsAppSDK_IsSelfContained(BOOL*isSelfContained);
```


>but rather swallow "no identity"

Ignore the `APPMODEL_ERROR_NO_PACKAGE` case. That's not an error for this function, it just hasn't been dealt with yet. `IsSelfContained()` will return `false` when that occurs.

riverar: Understood, not concerned about external usage. I am more concerned that this helper function will get used by your folks not realizing it both throws and returns error codes and it's going to be a bug factory.

But sounds like you're addressing the no-identity case, thanks!

DrusTheAxe: Not a problem. Standard practices in our implementation is heavily biased for Modern C++. Throwing is _common_ so it's routinely handled.

Problematic in other codebases where that's not the case, e.g. where return codes like HRESULT or Win32 Error Codes reign supreme. The upside to (relatively) new code, we could lay down new(er) ground rules :-)

hulumane (approved):

docs/Coding-Guidelines/SelfContained.md
```diff
@@ -0,0 +1,46 @@
...
+
+void JustDoIt()
+{
+    if (::Microsoft::WindowsAppSDK::SelfContained::IsSelfContained())
+    {
+        printf("WindowsAppSDK/SelfContained");
+    }
+    else
+    {
+        printf("WindowsAppSDK/MSIX");
```
Scottj1s, line 28: "WindowsAppSDK/FrameworkPackage" would be clearer, as you can have a SelfContained MSIX

dev/Common/Microsoft.WindowsAppSDK.SelfContained.h
```diff
@@ -0,0 +1,48 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#ifndef __MICROSFT_WINDOWSAPPSDK_SELFCONTAINED_H
+#define __MICROSFT_WINDOWSAPPSDK_SELFCONTAINED_H
+
+//TODO:37999189 Support 3rd scenario (Yes=DevInnerLoop, Yes=PipelineBuild product, No=PipelineBuild Unit/FunctionalTests)
+#if defined(WINDOWSAPPSDK_SELFCONTAINED_RUNTIME_PACKAGE_FRAMEWORK_PACKAGEFAMILYNAME_W)
```
Scottj1s, line 8: why is SELFCONTAINED in the symbol?

test/Common/pch.h
```diff
@@ -0,0 +1,45 @@
...
+#include <MsixDynamicDependency.h>
+
+#include <MddBootstrap.h>
+#include <MddBootstrapTest.h>
+
+#include <appmodel_msixdynamicdependency.h>
+
+#include <WexTestClass.h>
+
+#define WINDOWSAPPSDK_SELFCONTAINED_RUNTIME_PACKAGE_FRAMEWORK_PACKAGEFAMILYNAME_W L"Microsoft.WindowsAppRuntime.Framework_8wekyb3d8bbwe"
```
Scottj1s, line 37: SELFCONTAINED?

riverar:
Nit: Commit summaries should be concise and consistent and ~50 characters or less, leaving the other words for subsequent lines.

dev/Common/Microsoft.WindowsAppSDK.SelfContained.h
```diff
@@ -0,0 +1,48 @@
...
+
+namespace Microsoft::WindowsAppSDK::SelfContained
+{
+inline bool IsSelfContained()
+{
+    const UINT32 flags{ PACKAGE_FILTER_HEAD | PACKAGE_FILTER_DIRECT | PACKAGE_FILTER_STATIC | PACKAGE_FILTER_DYNAMIC | PACKAGE_INFORMATION_BASIC };
+    uint32_t packageInfoCount{};
+    const PACKAGE_INFO* packageInfo{};
+    wil::unique_cotaskmem_ptr<BYTE[]> buffer;
+    THROW_IF_FAILED(::AppModel::PackageGraph::GetCurrentPackageGraph(flags, packageInfoCount, packageInfo, buffer));
```
riverar, line 32: Still throws. Will this PR have the changes mentioned earlier?

docs/Coding-Guidelines/SelfContained.md
```diff
@@ -0,0 +1,46 @@
+# Self-Contained
+
+## What is 'Self-Contained'?
+
+Windows App SDK can be deployed via MSIX or self-contained, aka WinAppSDK/MSIX and WinAppSDK/SelfContained.
```
riverar, line 5: Nit: Doesn't answer the question -- what is self-contained?

docs/Coding-Guidelines/SelfContained.md
```diff
@@ -0,0 +1,46 @@
...
+
+See the [Self-Contained spec](https://github.com/microsoft/WindowsAppSDK/pull/1202) for more details.
+
+## How to detect if Windows App SDK is deployed Self-Contained?
+
+If necessary, use `Microsoft::WindowsAppSDK::SelfContained::IsSelfContained()` to detect if the
+current process is using WindowsAppSDK deployed via SelfContained (vs MSIX).
```
riverar, line 12: But you can have an app that's MSIX packaged and self-contained, right? Perhaps drop the parenthetical to avoid having to explain.

```suggestion
current process is using WindowsAppSDK deployed via SelfContained.
```

docs/Coding-Guidelines/SelfContained.md
DrusTheAxe: This is specifically referring to how WinAppSDK is deployed - via MSIX or Self-Contained.
How the caller/user/app is deployed is a different question

Phrasing folks have thrown around various, but WinAppSDK/MSIX vs WinAppSDK/SelfContained has been one of the more common and unambiguous ones.

[Initial draft of AppLocal Deployment spec.](https://github.com/microsoft/WindowsAppSDK/pull/1202) (#1202) should define the appropos terminology but that's a draft and a bit dated (still uses AppLocal terminology). So until that's updated I'll stick with the terms here

DrusTheAxe: +@mikebattista for visibility

dev/Common/Microsoft.WindowsAppSDK.SelfContained.h
DrusTheAxe: Don't sweat it this is going away. It was a quick hack to unblock some testing but needs to change from inline to an exported function to support all 3 scenarios.

test/Common/pch.h
DrusTheAxe: See previous comcment

docs/Coding-Guidelines/SelfContained.md
DrusTheAxe: "Self-Contained" is the phrase for what was formerly referred to as "AppLocal" or "In-App".

In short, Windows App SDK deployed as Self-Contained means "here's the files, add them to your add". No Microsoft.WindowsAppSDK MSIX packages.

The Self-Contained spec is in-progress at [Initial draft of AppLocal Deployment spec. #1202](https://github.com/microsoft/WindowsAppSDK/pull/1202)

+@mikebattista

docs/Coding-Guidelines/SelfContained.md
DrusTheAxe: > But you can have an app that's MSIX packaged and self-contained, right?

Yes.

There are 2 actors involved -- (1) WinAppSDK bits and (2) an app using WinAppSDK. This IsSelfContained() determines how (1) WinAppSDK is deployed, whether that's via MSIX packages or as files with/in/part-of your app.

That's independent of how (2) an app is deployed. You absolutely can make MyApp.msix using WinAppSDK/SelfContained, in which case MyApp.msix contains MyApp.exe etc AND microsoft.windowsappruntime.dll and other files making up WinAppSDK's runtime.

@mikebattista Terminology seems to be a bit confusing. I look forward to [the Self-Contained spec](https://github.com/microsoft/WindowsAppSDK/pull/1202) defining the terminology we should use.

PR #2145, opened 2022-02-23
Add basic telemetry to toast notification APIs

loneursid:
Initial draft for adding telemetry to the toast notifications public APIs. Two options are presented, option 1, which uses a single point for logging failures and successes and option 2, which handles failures separately from successes.

The interesting code is in AppNotificationManager.cpp starting at line 309.

Here's what the data is logged, the first two entries are from sucessful calls and the last two from failures.
| Event Name | OperationResult (Field 3) | AppId (Field 4) | IsAppPackaged (Field 5) | AppName (Field 6) |
|---|---|---|---|---|
| GetAllAsyncByAPI | 0x00000000 | ToastNotificationsTestAppPackage_8wekyb3d8bbwe!App | true  | ToastNotificationsTestAppPackage_8wekyb3d8bbwe!App |
| GetAllAsyncByAPI | 0x00000000 | TestAppId | false | ToastNotificationsTestApp.exe |
| GetAllAsyncByAPI | 0x80004005 | ToastNotificationsTestAppPackage_8wekyb3d8bbwe!App | true  | ToastNotificationsTestAppPackage_8wekyb3d8bbwe!App |
| GetAllAsyncByAPI | 0x80004005 | TestAppId | false | ToastNotificationsTestApp.exe |


dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -296,37 +302,102 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
         co_await winrt::resume_background();
 
         std::wstring appId{ RetrieveNotificationAppId() };
+
         THROW_IF_FAILED(ToastNotifications_RemoveAllToastsForApp(appId.c_str()));
     }
 
+#if 1 //  Option 1
```
loneursid, line 309: First option, using scope_exit() to provide a single point for handling all telemetry.

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -296,37 +302,102 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
+        }
+        catch (...)
         {
-            THROW_IF_FAILED(toastPropertiesCollection->get_Size(&count));
+            eptr = std::current_exception();
+            telemetryResult = wil::ResultFromCaughtException();
         }
+    }
+#else // Option 2
+    winrt::Windows::Foundation::IAsyncOperation<winrt::Windows::Foundation::Collections::IVector<winrt::Microsoft::Windows::AppNotifications::AppNotification>> AppNotificationManager::GetAllAsync()
```
loneursid, line 359: Second option, which handles successes and failures in separate path. Mostly offered for reference and comparison, I'm slowly warming up to the first option.

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -299,37 +305,102 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
+                toastNotifications.Append(toastNotification);
+            }
+
+            co_return toastNotifications;
+        }
+        catch (...)
         {
-            THROW_IF_FAILED(toastPropertiesCollection->get_Size(&count));
+            eptr = std::current_exception();
+            telemetryResult = wil::ResultFromCaughtException();
```
loneursid, line 358: This is the reason I prefer Option 2 over Option 1. Capturing the exception in the catch block so it can be rethrown later outside the catch block makes the code more complex to maintain and understand. I much prefer the Option 1 approach which lets the language handle rethrowing the exception from the catch block.

dev/AppNotifications/AppNotificationManager.cpp
danielayala94: I agree with the second option as well 😊.

dev/AppNotifications/AppNotificationTelemetry.h
```diff
@@ -0,0 +1,319 @@
...
+        std::wstring const& appId) noexcept try
+    {
+        if (c_maxEventLimit >= UpdateLogEventCount())
+        {
+            TraceLoggingClassWriteMeasure(
+                "RegisterByAPI",
+                TelemetryPrivacyDataTag(PDT_ProductAndServicePerformance),
+                _GENERIC_PARTB_FIELDS_ENABLED,
+                TraceLoggingHexUInt32(hr, "OperationResult"),
+                TraceLoggingWideString(appId.c_str(), "AppId"),
```
danielayala94, line 31: I would imagine `AppId` is sufficient. What's the reason of logging `AppName` as well?

dev/AppNotifications/AppNotificationTelemetry.h
```diff
@@ -0,0 +1,319 @@
...
+    // {7c1b07ef-a7c0-56d1-5456-385ebd4412b2}
+    (0x7c1b07ef, 0xa7c0, 0x56d1, 0x54, 0x56, 0x38, 0x5e, 0xbd, 0x44, 0x12, 0xb2)); //ELx need a new guid here
+
+
+class AppNotificationTelemetry : public wil::TraceLoggingProvider
+{
+    IMPLEMENT_TELEMETRY_CLASS(AppNotificationTelemetry, AppNotificationTelemetryProvider);
+
+public:
+    DEFINE_EVENT_METHOD(RegisterByAPI)(
```
danielayala94, line 20: What does ByAPI suffix mean in all event names?

dev/AppNotifications/AppNotificationTelemetry.h
danielayala94: Or maybe AppName is the DisplayName for unpackaged apps?

loneursid: If you look at the table in the description, you'll see that for packaged app appId and appName are the same but for unpackaged, they are **TestAppId** and **ToastNotificationsTestApp.exe** respectively.

dev/AppNotifications/AppNotificationTelemetry.h
loneursid: I replicated the same pattern we used for PushNotification telemetry. We added **byApi** to the events to indicate that these were discreet API calls made by an app.

dev/AppNotifications/AppNotificationTelemetry.h
loneursid: PushNotification telemetry is only logging the appName. That's something we'll likely have to revisit later when we go through the comprehensive notification tasks.

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -322,37 +328,102 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
+            }
+
+            AppNotificationTelemetry::GetAllAsyncByAPI(S_OK, appId);
+
+            co_return toastNotifications;
+        }
+        catch (...)
+        {
+            AppNotificationTelemetry::GetAllAsyncByAPI(wil::ResultFromCaughtException(), appId);
+            throw;
```
danielayala94, line 425: It doesn't seem like this `throw;` propagate the error code? Maybe we can use `THROW_HR(wil::Result...())`.

dev/AppNotifications/AppNotificationTelemetry.h
danielayala94: If all events are marked as ByApi, then it isn't needed at all, right?

dev/AppNotifications/AppNotificationTelemetry.h
```diff
@@ -0,0 +1,319 @@
...
+    }
+    CATCH_LOG()
+
+    DEFINE_EVENT_METHOD(UnregisterByAPI)(
+        winrt::hresult hr,
+        std::wstring const& appId) noexcept try
+    {
+        if (c_maxEventLimit >= UpdateLogEventCount())
+        {
+            TraceLoggingClassWriteMeasure(
```
danielayala94, line 44: I remember @brflynn demonstrated that we can get file and line number of failure origin being logged as part of the event. Is this something we can have as part of WinAppSDK telemetry?

dev/AppNotifications/AppNotificationTelemetry.h
```diff
@@ -0,0 +1,319 @@
...
+                TraceLoggingHexUInt32(hr, "OperationResult"),
+                TraceLoggingWideString(appId.c_str(), "AppId"),
+                TraceLoggingBool(IsPackagedApp(), "IsAppPackaged"),
+                TraceLoggingWideString(GetAppName(), "AppName"));
+        }
+    }
+    CATCH_LOG()
+
+private:
+    Microsoft::WRL::Wrappers::CriticalSection m_lock;
```
danielayala94, line 237: Nit: It is encouraged to use wil wrappers or winrt APIs instead of WRL.

dev/AppNotifications/AppNotificationUtility.h
```diff
@@ -41,6 +41,16 @@ namespace Microsoft::Windows::AppNotifications::Helpers
         return (str == nullptr || wcslen(str) == 0) ? wil::unique_hstring() : wil::make_unique_string<wil::unique_hstring>(str);
     }
 
+    inline void RethrowIf(std::exception_ptr eptr) try
+    {
+        if (eptr) std::rethrow_exception(eptr);
```
danielayala94, line 46: nit:

```
if (...)
{
// some code
}
```

danielayala94: I know it's just one line, but that's the format we follow.

dev/AppNotifications/AppNotificationManager.cpp
danielayala94: Are you waiting for more thoughts on the approach?

dev/AppNotifications/AppNotificationTelemetry.h
loneursid: I don't disagree, but we might add more events in the future when we do the comprehensive version and the push notifications event also all have byApi and may also get new events soon. I would think that whatever reasoning was applied when we decided to add ByApI for all push notification events would apply for this situation too.

dev/AppNotifications/AppNotificationManager.cpp
loneursid: Throw propagates the current caught exception. The error code is part of the exception, there is no point in extracting the error code from the exception and creating a new exception based on that same error code.

dev/AppNotifications/AppNotificationUtility.h
loneursid: Hopefully this code will go away once we've decided between option 1 and 2. Otherwise I'll fix it

loneursid:
@danielayala94 - github won;t let me reply to some of the comments you left. About _"Are you waiting for more thoughts on the approach?"_ I'm waiting for Pavan's call. It was his request that we use scope_exit() and I'd like him to see what it looks like and confirm that this is really what he wants before I implement everywhere.

dev/AppNotifications/AppNotificationManager.cpp
loneursid: Think about it this way: we only catch the exception because we want to get some telemetry, otherwise we wouldn't catch it. That's exactly what the code does: catch the exception, extract and log the error code, then let the exception propagate without additional intervention.

dev/AppNotifications/AppNotificationTelemetry.h
loneursid: Do you have an example? We'd have to pass the file name and line number at the call site or wrapped the call into a macro that does it, no?

dev/AppNotifications/AppNotificationManager.cpp
danielayala94: I see. Thanks for explaining!

dev/AppNotifications/AppNotificationTelemetry.h
```diff
@@ -0,0 +1,319 @@
...
+    }
+    CATCH_LOG()
+
+    DEFINE_EVENT_METHOD(GetAllAsyncByAPI)(
+        winrt::hresult hr,
+        std::wstring const& appId) noexcept try
+    {
+        if (c_maxEventLimit >= UpdateLogEventCount())
+        {
+            TraceLoggingClassWriteMeasure(
```
danielayala94, line 224: Is WinAppSDK version being logged for free? Like in Part A?

Either way, I think version is an important field to have.

dev/AppNotifications/AppNotificationTelemetry.h
```diff
@@ -0,0 +1,319 @@
...
+                "GetAllAsyncByAPI",
+                TelemetryPrivacyDataTag(PDT_ProductAndServicePerformance),
+                _GENERIC_PARTB_FIELDS_ENABLED,
+                TraceLoggingHexUInt32(hr, "OperationResult"),
+                TraceLoggingWideString(appId.c_str(), "AppId"),
+                TraceLoggingBool(IsPackagedApp(), "IsAppPackaged"),
```
danielayala94, line 230: Once @hulumane has added the self-contained checks, we may also want to add an isSelfContained field in all these events.

dev/AppNotifications/AppNotificationTelemetry.h
danielayala94: I think wil can do part of the work for you. I'll share with you the PR offline.

danielayala94:
> @danielayala94 - github won't let me reply to some of the comments you left. About _"Are you waiting for more thoughts on the approach?"_ I'm waiting for Pavan's call. It was his request that we use scope_exit() and I'd like him to see what it looks like and confirm that this is really what he wants before I implement everywhere.

@loneursid Thanks for responding! It seems like we can't respond directly to comments that were added from previous conversations, but you can respond in the conversation.

dev/AppNotifications/AppNotificationTelemetry.h
pmpurifoy: wil::srwlock m_lock in most of our classes :)

dev/AppNotifications/AppNotificationTelemetry.h
loneursid: Thanks for the doc. I'll add that info to the task for adding comprehensive telemetry for toast and push. This is beyond the goal of this PR, which is to provide basic telemetry to measure feature adoption rate.

dev/AppNotifications/AppNotificationTelemetry.h
loneursid: It's in part C: (for example)
"COMMON_WINDOWSAPPSDK_PARAMS": {
           "WindowsAppSDKVersion": "0.9",
           "WindowsAppSDKExperimentationLevel": "",
           "IsDebugging": false,
           "AppSessionGuid": "00000D80-0001-0002-B425-C8C4E284D701"
}

dev/AppNotifications/AppNotificationTelemetry.h
loneursid: and go back to push, I guess and add it there too.

dev/AppNotifications/AppNotificationTelemetry.h
danielayala94: Great! What is WindowsAppSDKExperimentationLevel? And what is the scope of AppSessionGuid?

loneursid: WindowsAppSDKExperimentationLevel whether this is an experimental build, I think. Don;'t know what the AppSessionGuid comes from

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -322,37 +328,102 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
+
         THROW_IF_FAILED(ToastNotifications_RemoveAllToastsForApp(appId.c_str()));
     }
 
+#if 1 //  Option 1
     winrt::Windows::Foundation::IAsyncOperation<winrt::Windows::Foundation::Collections::IVector<winrt::Microsoft::Windows::AppNotifications::AppNotification>> AppNotificationManager::GetAllAsync()
     {
-        co_await winrt::resume_background();
+        HRESULT telemetryResult{ S_OK };
+        std::exception_ptr eptr{};
```
hulumane, line 339: You don't need this. Just capture the HRESULT in the catch block.

Then outside the try catch, use the HRESULT with:
LogTelemetry(API, hr);
THROW_IF_FAILED(hr);

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -322,37 +328,102 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
         std::wstring appId{ RetrieveNotificationAppId() };
+
         THROW_IF_FAILED(ToastNotifications_RemoveAllToastsForApp(appId.c_str()));
     }
 
+#if 1 //  Option 1
     winrt::Windows::Foundation::IAsyncOperation<winrt::Windows::Foundation::Collections::IVector<winrt::Microsoft::Windows::AppNotifications::AppNotification>> AppNotificationManager::GetAllAsync()
     {
-        co_await winrt::resume_background();
+        HRESULT telemetryResult{ S_OK };
```
hulumane, line 338: Just call it hr.

winrt::hresult hr{S_OK};

dev/AppNotifications/AppNotificationManager.cpp
hulumane: Using std::exception is overkill and not needed imho.

dev/AppNotifications/AppNotificationManager.cpp
hulumane: Take a look at my comments above. I think that simplifies option 1.

dev/AppNotifications/AppNotificationTelemetry.h
```diff
@@ -0,0 +1,319 @@
...
+class AppNotificationTelemetry : public wil::TraceLoggingProvider
+{
+    IMPLEMENT_TELEMETRY_CLASS(AppNotificationTelemetry, AppNotificationTelemetryProvider);
+
+public:
+    DEFINE_EVENT_METHOD(RegisterByAPI)(
+        winrt::hresult hr,
+        std::wstring const& appId) noexcept try
+    {
+        if (c_maxEventLimit >= UpdateLogEventCount())
```
hulumane, line 24: @brflynn Throttling being implemented here using a count pattern. Thoughts on this Vs the approach you implemented recently?

dev/AppNotifications/AppNotificationTelemetry.h
hulumane: Yep. Shouldn't be using WRL. 
wil == good!

dev/AppNotifications/AppNotificationTelemetry.h
hulumane: I added @brflynn to see if he has comments on this. He suggested a completely different pattern to use in a recent PR for aggregation.

dev/AppNotifications/AppNotificationManager.cpp
loneursid: We're going with the first option.

dev/AppNotifications/AppNotificationManager.cpp
loneursid: We're going with the first option.

dev/AppNotifications/AppNotificationTelemetry.h
loneursid: Done, fixed in PushNotificationsTelemetry also

dev/AppNotifications/AppNotificationUtility.h
loneursid: Code is gone

dev/AppNotifications/AppNotificationManager.cpp
loneursid: Done.

dev/AppNotifications/AppNotificationManager.cpp
loneursid: done

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -55,76 +56,121 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
+            }
 
-        // Create event handle before COM Registration otherwise if a notification arrives will lead to race condition
-        GetWaitHandleForArgs().create();
+            // Create event handle before COM Registration otherwise if a notification arrives will lead to race condition
+            GetWaitHandleForArgs().create();
 
+            {
+                auto lock{ m_lock.lock_exclusive() };
+                THROW_HR_IF_MSG(E_INVALIDARG, m_notificationComActivatorRegistration, "Already Registered for App Notifications!");
```
loneursid, line 97: This is an example of why I'm not a fan of throwing away the existing exception and creating a new one solely based on the hr in scope_exit(). You loose context, like the message in this case.

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -212,145 +258,297 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
+        HRESULT hr{ S_OK };
+        std::wstring appId{};
+
         auto strong = get_strong();
         co_await resume_background();
 
-        std::wstring appId{ RetrieveNotificationAppId() };
+        auto logTelemetry{ wil::scope_exit([&]() {
+            AppNotificationTelemetry::UpdateAsyncByAPI(hr, appId);
+            THROW_HR_IF(hr, !(hr == S_OK || hr == E_NOT_SET));
```
loneursid, line 304: This is an example of why I'm a proponent of keeping success and failed exit points separate.

loneursid:
What I really would like is that, instead of using scope_exit and function level vars, we use a class to hide implementation details and manage the vars and the actual call to telemetry.

Instead of this:
```
        HRESULT hr{ S_OK };
        std::wstring appId{};
        auto logTelemetry{ wil::scope_exit([&]() {
            AppNotificationTelemetry::RemoveByGroupAsyncByAPI(hr, appId);
        }) };

        try
        {
                do some stuff
                appid = RetrieveNotificationAppId();
                do more stuff

                maybe return something...
        }
        catch
        {
                hr = wil::ResultFromCaughtException();
        }

```
we could have:
```
         logTelemetry{ AppNotificationsTelementry(:RemoveByGroupAsyncByAP) };

        try
        {
                do some stuff
                appid = RetrieveNotificationAppId();
                logTelemetry.setAppId(appid);
                do more stuff

                maybe return something...
        }
        catch
        {
                logTelemetry.setHResult(wil::ResultFromCaughtException());
        }
```
note that there are no function level variables (besides the logTelemetry object). The function level variables now encapsulated in the class and default initialize. The body still has the try / catch and just add whatever data it wants to log as it becomes available.

The class would log when its destructor is called. i,e.: at scope exit.

I think this would keep everything neat and hides some of the implementation details better.

dev/AppNotifications/AppNotificationManager.cpp
loneursid: I've fixed this. we don't have to create a new exception with the hr. Actually, creating a new expectation would lead to a crash

dev/AppNotifications/AppNotificationManager.cpp
loneursid: Also resolved.

dev/AppNotifications/AppNotificationTelemetry.h
loneursid: That should be good enough for basic telemetry. That's the same pattern that push utilizes. We can revisit when we take on the tasks for comprehensive telemetry

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -57,72 +58,117 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
 
     void AppNotificationManager::Register()
     {
-        winrt::guid storedComActivatorGuid{ GUID_NULL };
-        if (!PushNotificationHelpers::IsPackagedAppScenario())
+        HRESULT hr{ S_OK };
+        std::wstring appId{};
+
+        auto logTelemetry{ wil::scope_exit([&]() {
+            AppNotificationTelemetry::RegisterByAPI(hr, appId);
```
sharath2727, line 65: should we log packaged or unpackaged? Also, where is the appId being logged?

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -57,72 +58,117 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
     void AppNotificationManager::Unregister()
     {
-        auto lock{ m_lock.lock_exclusive() };
-        THROW_HR_IF_MSG(HRESULT_FROM_WIN32(ERROR_NOT_FOUND), !m_notificationComActivatorRegistration, "Not Registered for App Notifications!");
-        m_notificationComActivatorRegistration.reset();
+        HRESULT hr{ S_OK };
+        std::wstring appId{};
+
+        auto logTelemetry{ wil::scope_exit([&]() {
+            AppNotificationTelemetry::UnregisterByAPI(hr, appId);
```
sharath2727, line 121: where is the appId being assigned?

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -57,72 +58,117 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
     }
 
     void AppNotificationManager::UnregisterAll()
     {
-        Unregister();
+        HRESULT hr{ S_OK };
+        std::wstring appId{};
+
+        auto logTelemetry{ wil::scope_exit([&]() {
+            AppNotificationTelemetry::UnregisterAllByAPI(hr, appId);
```
sharath2727, line 143: same comment as above?

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -210,145 +256,297 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
+        std::wstring appId{};
+
+        auto logTelemetry{ wil::scope_exit([&]() {
+            AppNotificationTelemetry::ShowByAPI(hr, appId);
+        }) };
 
-        winrt::com_ptr<::ABI::Microsoft::Internal::ToastNotifications::INotificationTransientProperties> notificationTransientProperties = winrt::make_self<NotificationTransientProperties>(notification);
+        try
+        {
+            winrt::com_ptr<::ABI::Microsoft::Internal::ToastNotifications::INotificationProperties> notificationProperties = winrt::make_self<NotificationProperties>(notification);
```
sharath2727, line 268: do we want to log any of the attributes in NotificationProperties here?

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -210,145 +256,297 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
 
+        HRESULT hr{ S_OK };
+        std::wstring appId{};
+
         auto strong = get_strong();
         co_await resume_background();
 
-        std::wstring appId{ RetrieveNotificationAppId() };
+        auto logTelemetry{ wil::scope_exit([&]() {
+            AppNotificationTelemetry::UpdateAsyncByAPI(hr, appId);
```
sharath2727, line 301: can we also log the tag and group here?

loneursid: yes we can

dev/AppNotifications/AppNotificationTelemetry.h
```diff
@@ -0,0 +1,318 @@
...
+        {
+            TraceLoggingClassWriteMeasure(
+                "UnregisterAllByAPI",
+                TelemetryPrivacyDataTag(PDT_ProductAndServicePerformance),
+                _GENERIC_PARTB_FIELDS_ENABLED,
+                TraceLoggingHexUInt32(hr, "OperationResult"),
+                TraceLoggingWideString(appId.c_str(), "AppId"),
```
sharath2727, line 66: check for appId is nullptr?

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -57,72 +58,117 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
             {
-                THROW_IF_FAILED(PushNotifications_RegisterFullTrustApplication(notificationId.c_str(), GUID_NULL));
+                std::wstring notificationId{ RetrieveNotificationAppId() };
+                if (!AppModel::Identity::IsPackagedProcess())
+                {
+                    THROW_IF_FAILED(PushNotifications_RegisterFullTrustApplication(notificationId.c_str(), GUID_NULL));
 
-                storedComActivatorGuid = RegisterComActivatorGuidAndAssets();
-            }
+                    storedComActivatorGuid = RegisterComActivatorGuidAndAssets();
```
hulumane, line 78: Incorrect spacing\indenting everywhere here

hulumane: Please look into indenting in all the files and below as well.

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -57,72 +58,117 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
+        auto logTelemetry{ wil::scope_exit([&]() {
+            AppNotificationTelemetry::RegisterByAPI(hr, appId);
+        }) };
+
+        try
         {
-            std::wstring notificationId{ RetrieveNotificationAppId() };
-            if (!AppModel::Identity::IsPackagedProcess())
+            winrt::guid storedComActivatorGuid{ GUID_NULL };
+            if (!PushNotificationHelpers::IsPackagedAppScenario())
```
hulumane, line 71: Why is this PR changing the core logic here?

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -14,6 +14,7 @@
 #include "NotificationTransientProperties.h"
 #include "AppNotification.h"
 #include <NotificationProgressData.h>
+#include "AppNotificationTelemetry.h"
```
hulumane, line 17: Please do not baseline against the L1. The L1 is strictly for PR comments and P0 work.

hulumane: I would baseline this against Pauls bug fix branch which has the latest changes

dev/AppNotifications/AppNotificationManager.cpp
hulumane, line 65: Unless I'm missing something critical here, appId is always empty.

dev/AppNotifications/AppNotificationManager.cpp
sharath2727: so I can either remove it from the corresponding telemetry events or fetch it in the function so it's part of the telemetry event -  Can you fetch it in the function so it is part of the telemetry event. Maybe you could do that as part of the scope exit that weren't using the appId internally? Also, I think it would be great to have a bool for isPackagedAppScenario?

dev/AppNotifications/AppNotificationManager.cpp
loneursid: It's the best base branch to use for my work and I want to keep the option of merging directly into the L1 open in case it slips since this work needs to be in for Friday. Either the branch will merge early as planned and I'll re-target my work to main, or it will slip and I'll merge directly into the L1

dev/AppNotifications/AppNotificationManager.cpp
loneursid: It's possible that there are a few functions that weren't using the appid internally, so I can either remove it from the corresponding telemetry events or fetch it in the function so it's part of the telemetry event. The telemetry events already log the app name.

hulumane: Logging appName based on what? ProcessName?

dev/AppNotifications/AppNotificationTelemetry.h
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    inline const wchar_t* GetAppName()
+    {
+        static const std::wstring appName = IsPackagedApp() ? GetAppNamePackaged() : GetAppNameUnpackaged();
+
+        return appName.c_str();
+    }
+
+    std::wstring GetAppNamePackaged() noexcept
```
hulumane, line 275: I think this should throw

dev/AppNotifications/AppNotificationTelemetry.h
```diff
@@ -0,0 +1,318 @@
...
+    {
+        wchar_t appUserModelId[APPLICATION_USER_MODEL_ID_MAX_LENGTH]{};
+
+        UINT32 appUserModelIdSize = ARRAYSIZE(appUserModelId);
+        auto result = GetCurrentApplicationUserModelId(&appUserModelIdSize, appUserModelId);
+        if (result != ERROR_SUCCESS)
```
hulumane, line 281: Should throw if not ERROR_SUCCESS

dev/AppNotifications/AppNotificationTelemetry.h
```diff
@@ -0,0 +1,318 @@
...
+        {
+            wcscpy_s(appUserModelId, L"AppUserModelId not found");
+            LOG_WIN32(result);
+        }
+
+        return appUserModelId;
+    }
+
+    PCWSTR CensorFilePath(PCWSTR path) noexcept
```
hulumane, line 290: Should throw

dev/AppNotifications/AppNotificationTelemetry.h
```diff
@@ -0,0 +1,318 @@
...
+        wil::unique_cotaskmem_string processName;
+        auto result = wil::GetModuleFileNameExW(GetCurrentProcess(), nullptr, processName);
+        if (result == ERROR_SUCCESS)
+        {
+            appName = CensorFilePath(processName.get());
+        }
+        else
+        {
+            appName = L"ModuleFileName not found";
+            LOG_WIN32(result);
```
hulumane, line 313: Should throw

hulumane: THROW_HR_WIN32_MSG

dev/AppNotifications/AppNotificationTelemetry.h
hulumane, line 290: This method should return const std::wstring. Don't return raw pointers as a rule.

dev/AppNotifications/AppNotificationTelemetry.h
```diff
@@ -0,0 +1,318 @@
...
+            wcscpy_s(appUserModelId, L"AppUserModelId not found");
+            LOG_WIN32(result);
+        }
+
+        return appUserModelId;
+    }
+
+    PCWSTR CensorFilePath(PCWSTR path) noexcept
+    {
+        if (path)
```
hulumane, line 292: If Path is empty or null, this should throw.

dev/AppNotifications/AppNotificationTelemetry.h
```diff
@@ -0,0 +1,318 @@
...
+        {
+            path = !PathIsFileSpecW(path) ? PathFindFileNameW(path) : path;
+        }
+
+        return path;
+    }
+
+    std::wstring GetAppNameUnpackaged()
```
hulumane, line 300: const std::wstring

dev/AppNotifications/AppNotificationTelemetry.h
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    inline bool IsPackagedApp()
+    {
+        static const bool isPackagedApp = AppModel::Identity::IsPackagedProcess();
+
+        return isPackagedApp;
+    }
+
+    inline const wchar_t* GetAppName()
```
hulumane, line 268: return const std::wstring. No raw pointers please.

dev/AppNotifications/AppNotificationTelemetry.h
```diff
@@ -0,0 +1,318 @@
...
+    {
+        static const std::wstring appName = IsPackagedApp() ? GetAppNamePackaged() : GetAppNameUnpackaged();
+
+        return appName.c_str();
```
hulumane, line 272: memory corruption since appName loses scope after we bail this function.

dev/AppNotifications/AppNotificationTelemetry.h
hulumane: So if someone does this outside:
auto name = GetAppName;
Foo(name); // name is undefined at this point

dev/AppNotifications/AppNotificationTelemetry.h
hulumane: appId is an std::wstring. It is never nullptr.

dev/AppNotifications/AppNotificationTelemetry.h
```diff
@@ -0,0 +1,318 @@
...
+                _GENERIC_PARTB_FIELDS_ENABLED,
+                TraceLoggingHexUInt32(hr, "OperationResult"),
+                TraceLoggingWideString(appId.c_str(), "AppId"),
+                TraceLoggingBool(IsPackagedApp(), "IsAppPackaged"),
+                TraceLoggingWideString(GetAppName(), "AppName"));
+        }
+    }
+    CATCH_LOG()
+
+    DEFINE_EVENT_METHOD(UnregisterByAPI)(
```
hulumane, line 37: Strange name. Better naming would be LogUnregister or something similar

dev/AppNotifications/AppNotificationTelemetry.h
```diff
@@ -0,0 +1,318 @@
...
+                TraceLoggingHexUInt32(hr, "OperationResult"),
+                TraceLoggingWideString(appId.c_str(), "AppId"),
+                TraceLoggingBool(IsPackagedApp(), "IsAppPackaged"),
+                TraceLoggingWideString(GetAppName(), "AppName"));
+        }
+    }
+    CATCH_LOG()
+
+    DEFINE_EVENT_METHOD(UpdateAsyncByAPI)(
+        winrt::hresult hr,
```
hulumane, line 92: Suggest renaming all these events to something like this:
LogUpdateAsync()

so Log{API Name}

dev/AppNotifications/AppNotificationManager.cpp
hulumane: Please remove appId if you believe it isn't critical to logging. Totally fine with that as part of V1 logging story since we use ProcessName/AppUserModelId as AppName instead.

dev/AppNotifications/AppNotificationManager.cpp
loneursid: Also, I think it would be great to have a bool for isPackagedAppScenario? As I said all events already have a bool for package / unpackaged scenarios.

dev/AppNotifications/AppNotificationTelemetry.h
loneursid: done

loneursid: done

dev/AppNotifications/AppNotificationTelemetry.h
loneursid: done

dev/AppNotifications/AppNotificationTelemetry.h
loneursid: not a good idea because telemetry events are all no throw, so they can be used in situation where exceptions are already in flight or in destructors.

This is a private function, even if it throws, I'll have to catch it before it exits the telemetry log function and there's nothing more that I'll be able to do at that point.

The goal here is to do a best effort of logging as mush details as possible but nothing more.

loneursid: not a good idea because telemetry events are all no throw, so they can be used in situation where exceptions are already in flight or in destructors.

This is a private function, even if it throws, I'll have to catch it before it exits the telemetry log function and there's nothing more that I'll be able to do at that point.

The goal here is to do a best effort of logging as mush details as possible but nothing more.

dev/AppNotifications/AppNotificationManager.cpp
loneursid: All usage of appid have been addressed

dev/AppNotifications/AppNotificationTelemetry.h
loneursid: I don't think we want to throw from telemetry, that would lead to host of issues

dev/AppNotifications/AppNotificationTelemetry.h
loneursid: I don't think we want to throw from telemetry, that would lead to host of issues

dev/AppNotifications/AppNotificationManager.cpp
loneursid: Indenting looks good to me now. Let me know if you find other issues

dev/AppNotifications/AppNotificationManager.cpp
loneursid: done

dev/AppNotifications/AppNotificationManager.cpp
loneursid: Spacing should be fixed now

dev/AppNotifications/AppNotificationTelemetry.h
hulumane: Seen Internal developers misuse APIs several times. Using raw pointers as returns should be avoided at all costs because it can lead to lifetime management issues.

dev/AppNotifications/AppNotificationTelemetry.h
hulumane: Best effort is not an option at this layer. The Logging wrappers do the best effort like LogRegister() which is wrapped in a try catch_log block. So we would never throw at the SDK layer even if something bad happens during logging.
The helpers that it uses to achieve its goal should throw. Otherwise, we end up with non-determinist behavior in the Logging wrapper layers.

dev/AppNotifications/AppNotificationTelemetry.h
hulumane: See explanation above as to why lower layers need to throw. 
It won't lead to host of issues because the logging wrappers essentially swallow any errors today: they are wrapped in try catch_log

dev/AppNotifications/AppNotificationTelemetry.h
hulumane: Sure

dev/AppNotifications/AppNotificationManager.cpp
hulumane: Hmm not sure if a new iteration was sent with these changes but the indentation issues can still be seen above and below

dev/AppNotifications/AppNotificationTelemetry.h
hulumane: I would be very concerned if this starts throwing. I would rather go with correctness at this layer and not log it if there is a bug here.

Watson should help us find and resolve bugs at this layer.

dev/AppNotifications/AppNotificationManager.cpp
loneursid: I think that it's because you're looking at the code through github. If you look at in in VS, then it's fine. Keep in mind that most the code was indented an extra tab when I introduced the try / catch

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -57,72 +58,124 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
+
+        try
         {
-            std::wstring notificationId{ RetrieveNotificationAppId() };
-            if (!AppModel::Identity::IsPackagedProcess())
+            winrt::guid storedComActivatorGuid{ GUID_NULL };
+            if (!PushNotificationHelpers::IsPackagedAppScenario())
             {
-                THROW_IF_FAILED(PushNotifications_RegisterFullTrustApplication(notificationId.c_str(), GUID_NULL));
+                appId = RetrieveNotificationAppId();
```
hulumane, line 73: This call needs to be moved before the IsPacakgedAppScenario check. Otherwise we won't populate appId when we hit the else case right?

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -57,72 +58,124 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
 
     void AppNotificationManager::Register()
     {
-        winrt::guid storedComActivatorGuid{ GUID_NULL };
-        if (!PushNotificationHelpers::IsPackagedAppScenario())
+        HRESULT hr{ S_OK };
+        std::wstring appId{};
+
+        auto logTelemetry{ wil::scope_exit([&]() {
```
hulumane, line 64: It's great that we use scopes now : )

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -57,72 +58,124 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
-        THROW_HR_IF_MSG(HRESULT_FROM_WIN32(ERROR_NOT_FOUND), !m_notificationComActivatorRegistration, "Not Registered for App Notifications!");
-        m_notificationComActivatorRegistration.reset();
+        HRESULT hr{ S_OK };
+
+        auto logTelemetry{ wil::scope_exit([&]() {
+            std::wstring appId{};
+
+            try
+            {
+                appId = RetrieveNotificationAppId();
```
hulumane, line 124: Rather than add a try catch in scope exit, I would prefer that we call

appId = RetrieveNotificationAppId();
in the try catch below and use the appId directly here.

This would make the pattern consistent with the above code.

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -57,72 +58,124 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
+            // Remove any Registrations from the Long Running Process that are necessary for Cloud toasts
+            if (!PushNotificationHelpers::IsPackagedAppScenario())
+            {
+                auto notificationPlatform{ PushNotificationHelpers::GetNotificationPlatform() };
+                THROW_IF_FAILED(notificationPlatform->RemoveToastRegistrationMapping(m_processName.c_str()));
+            }
+
+            if (!AppModel::Identity::IsPackagedProcess())
+            {
+                appId = RetrieveNotificationAppId();
```
hulumane, line 166: Move this to outside: probably should be the first API call in the try block.

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -57,72 +58,124 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
 
     void AppNotificationManager::Register()
     {
-        winrt::guid storedComActivatorGuid{ GUID_NULL };
-        if (!PushNotificationHelpers::IsPackagedAppScenario())
+        HRESULT hr{ S_OK };
+        std::wstring appId{};
```
hulumane, line 62: See appId being used a lot everywhere. Maybe just introduce a member variable m_appId and initialize that in the ctor? Then we can delete all this redundant code and reuse m_appId everywhere.

dev/AppNotifications/AppNotificationTelemetry.h
```diff
@@ -0,0 +1,330 @@
...
+                TraceLoggingBool(IsPackagedApp(), "IsAppPackaged"),
+                TraceLoggingWideString(GetAppName(), "AppName"));
+        }
+    }
+    CATCH_LOG()
+
+    DEFINE_EVENT_METHOD(LogRemoveByTagAsync)(
+        winrt::hresult hr,
+        std::wstring const& appId,
+        winrt::hstring const& tag) noexcept try
```
hulumane, line 154: Tag is PII. Strongly recommend that you don't log it. Log length of tag instead.

dev/AppNotifications/AppNotificationTelemetry.h
```diff
@@ -0,0 +1,330 @@
...
+                TraceLoggingBool(IsPackagedApp(), "IsAppPackaged"),
+                TraceLoggingWideString(GetAppName(), "AppName"));
+        }
+    }
+    CATCH_LOG()
+
+    DEFINE_EVENT_METHOD(LogRemoveByTagAndGroupAsync)(
+        winrt::hresult hr,
+        std::wstring const& appId,
+        winrt::hstring const& tag,
```
hulumane, line 174: Tag and group are PII. REcommend logging only length of these properties.

dev/AppNotifications/AppNotificationManager.cpp
loneursid: maybe not everyone agrees, though :)

dev/AppNotifications/AppNotificationManager.cpp
loneursid: good call, done

dev/AppNotifications/AppNotificationTelemetry.h
loneursid: done

dev/AppNotifications/AppNotificationTelemetry.h
loneursid: done

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -210,46 +251,70 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
     {
         THROW_HR_IF(WPN_E_NOTIFICATION_POSTED, notification.Id() != 0);
 
-        winrt::com_ptr<::ABI::Microsoft::Internal::ToastNotifications::INotificationProperties> notificationProperties = winrt::make_self<NotificationProperties>(notification);
+        HRESULT hr{ S_OK };
+
+        auto logTelemetry{ wil::scope_exit([&]() {
+            AppNotificationTelemetry::LogShow(hr, m_appId);
```
hulumane, line 257: Recommend: Log length of payload, tag and group here as well.

dev/AppNotifications/AppNotificationTelemetry.h
loneursid: good catch, never thought of these as PII. 

done.

dev/AppNotifications/AppNotificationTelemetry.h
loneursid: good catch, never thought of these as PII.

done.

dev/AppNotifications/AppNotificationTelemetry.h
```diff
@@ -0,0 +1,330 @@
...
+                TraceLoggingBool(IsPackagedApp(), "IsAppPackaged"),
+                TraceLoggingWideString(GetAppName(), "AppName"));
+        }
+    }
+    CATCH_LOG()
+
+    DEFINE_EVENT_METHOD(LogUpdateAsync)(
+            winrt::hresult hr,
+            std::wstring const& appId,
+            winrt::hstring const& tag,
```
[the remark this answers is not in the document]
loneursid: good catch, never thought of these as PII.

done.

dev/AppNotifications/AppNotificationManager.cpp
loneursid: done

dev/AppNotifications/AppNotificationManager.cpp
loneursid: appId is now initialized at construction.

dev/AppNotifications/AppNotificationManager.cpp
loneursid: appId is now initialized at construction

dev/AppNotifications/AppNotificationManager.cpp
loneursid: appId is now initialized at construction now.

dev/AppNotifications/AppNotificationTelemetry.h
```diff
@@ -0,0 +1,331 @@
...
+        wil::unique_cotaskmem_string processName;
+        auto result = wil::GetModuleFileNameExW(GetCurrentProcess(), nullptr, processName);
+        if (result == ERROR_SUCCESS)
+        {
+            appName = CensorFilePath(processName.get());
+        }
+        else
+        {
+            appName = L"ModuleFileName not found";
+            LOG_WIN32(result);
```
hulumane, line 326: Still only logging instead of throwing. Was this intended?

loneursid: No, that's something I missed. thanks for checking.

It's fixed now

hulumane (approved):

PR #2147, opened 2022-02-23
Fix AppNotification activation when no foreground handlers defined

pmpurifoy:
There was a bug with using REG_SINGLEUSE on app launch for AppNotification activation. The issue was that the com object was not activated the first time and only on the second activation would another process be created.

Now if the app has no foreground handlers, was launched by any ExtendedActivationKind that is not AppNotification, and receives a AppNotification activation, we CoCreateInstance the app and propagate the arguments through Activate.

### DIFF
--- a/dev/AppNotifications/AppNotificationManager.cpp
+++ b/dev/AppNotifications/AppNotificationManager.cpp
@@ -167,11 +167,13 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
 
             std::wstring commandLine{ GetCommandLine() };
 
-            // If the app was not launched due to ToastActivation, we will invoke the foreground handler.
+            // If the app was not launched due to ToastActivation, we will launch a new instance or invoke the foreground handlers.
             // Otherwise we store the EventArgs and signal to the Main thread
             auto pos{ commandLine.find(c_notificationActivatedArgument) };
-            if (pos == std::wstring::npos) // !AppNotification
+            if (pos == std::wstring::npos) // Any launch kind that is not AppNotification
             {
+                // If the Process was launched due to other Activation Kinds, we will need to
+                // re-route the payload to a new process if there are no registered event handlers.
                 if (!m_notificationHandlers)
                 {
                     winrt::guid registeredClsid{ GUID_NULL };
--- a/dev/AppNotifications/AppNotificationManager.cpp
+++ b/dev/AppNotifications/AppNotificationManager.cpp
@@ -149,6 +149,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
         [[maybe_unused]] NOTIFICATION_USER_INPUT_DATA const* data,
         [[maybe_unused]] ULONG dataCount) noexcept try
     {
+
         winrt::IMap<winrt::hstring, winrt::hstring> userInput{ winrt::single_threaded_map<winrt::hstring, winrt::hstring>() };
         for (unsigned long i = 0; i < dataCount; i++)
         {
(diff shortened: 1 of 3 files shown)


David-Quero:
Can we add a prompt in the demo app to let the user choose to register or not foreground handlers?

test/TestApps/ToastNotificationsDemoApp/main.cpp
```diff
@@ -113,15 +113,17 @@ std::wstring GetEnumString(winrt::AppNotificationSetting const& setting)
...
 
-winrt::AppNotification CreateToastNotification(winrt::hstring message)
-{
-    winrt::hstring xmlPayload{ L"<toast>" + message + L"</toast>" };
-    return winrt::AppNotification(xmlPayload);
-}
-
 winrt::AppNotification CreateToastNotification()
 {
-    return CreateToastNotification(L"intrepidToast");
```
David-Quero, line 124: Shell was having a pretty hard time handling this toast. We need to figure out what went wrong with the schema validation, I am concerned this could an issue on the Toast History.

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -169,9 +170,29 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
             // If the app was not launched due to ToastActivation, we will invoke the foreground handler.
             // Otherwise we store the EventArgs and signal to the Main thread
             auto pos{ commandLine.find(c_notificationActivatedArgument) };
-            if (pos == std::wstring::npos)
+            if (pos == std::wstring::npos) // !AppNotification
```
hulumane, line 173: Fix the comment please along the lines:
// If the Process was launched due to other Activation Kinds, we will need to re-route the payload to a new process if there are no registered event handlers.

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -169,9 +170,29 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
+                    }
+                    else
+                    {
+                        std::wstring storedComActivatorString{ RegisterComActivatorGuidAndAssets() };
+                        // Remove braces around the guid string
+                        registeredClsid = winrt::guid(storedComActivatorString.substr(1, storedComActivatorString.size() - 2));
+                    }
+
+                    auto notificationCallback = winrt::create_instance<INotificationActivationCallback>(registeredClsid, CLSCTX_ALL);
+                    notificationCallback->Activate(L"", invokedArgs, data, dataCount);
```
hulumane, line 192: Please don't pass empty AppId. Instead call RetrieveAppIdentifier and pass that in.

hulumane (approved):

dev/AppNotifications/AppNotificationManager.cpp
pmpurifoy: Actually, we can use the appUserModelId in the parameters that is currently unused :).

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -166,12 +167,34 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
+                    }
+                    else
+                    {
+                        std::wstring storedComActivatorString{ RegisterComActivatorGuidAndAssets() };
+                        // Remove braces around the guid string
+                        registeredClsid = winrt::guid(storedComActivatorString.substr(1, storedComActivatorString.size() - 2));
+                    }
+
+                    auto notificationCallback{ winrt::create_instance<INotificationActivationCallback>(registeredClsid, CLSCTX_ALL) };
+                    notificationCallback->Activate(appUserModelId, invokedArgs, data, dataCount);
```
danielayala94, line 192: This returns HRESULT. Do we need to throw if this fails, or how do we want to handle this?

hulumane: we should throw

hulumane (approved):

pmpurifoy:
> Can we add a prompt in the demo app to let the user choose to register or not foreground handlers?

I think this is a good idea. Let's have a PR that contains demo app improvements

PR #2148, opened 2022-02-24
Modify constructor for AppNotificationProgressData per API review

hulumane:
Per API review feedback, we need a ctor that takes in a sequence number. Sequence number needs to be non-zero and enforced, otherwise we throw.

### DIFF
--- a/dev/AppNotifications/AppNotificationProgressData.cpp
+++ b/dev/AppNotifications/AppNotificationProgressData.cpp
@@ -6,7 +6,6 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
 {
     AppNotificationProgressData::AppNotificationProgressData(uint32_t sequenceNumber)
     {
-        auto lock{ m_lock.lock_exclusive() };
         THROW_HR_IF(E_INVALIDARG, sequenceNumber == 0); // The sequence number is always greater than 0
         m_sequenceNumber = sequenceNumber;
     }
@@ -18,8 +17,10 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
     }
     void AppNotificationProgressData::SequenceNumber(uint32_t sequenceNumber)
     {
-        auto lock{ m_lock.lock_exclusive() };
         THROW_HR_IF(E_INVALIDARG, sequenceNumber == 0); // The sequence number is always greater than 0
+
+        auto lock{ m_lock.lock_exclusive() };
+
         m_sequenceNumber = sequenceNumber;
     }
     hstring AppNotificationProgressData::Title()
--- a/dev/AppNotifications/AppNotificationProgressData.cpp
+++ b/dev/AppNotifications/AppNotificationProgressData.cpp
@@ -4,6 +4,12 @@
 
 namespace winrt::Microsoft::Windows::AppNotifications::implementation
 {
+    AppNotificationProgressData::AppNotificationProgressData(uint32_t sequenceNumber)
+    {
+        auto lock{ m_lock.lock_exclusive() };
+        THROW_HR_IF(E_INVALIDARG, sequenceNumber == 0); // The sequence number is always greater than 0
+        m_sequenceNumber = sequenceNumber;
+    }
     uint32_t AppNotificationProgressData::SequenceNumber()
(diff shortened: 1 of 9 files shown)


dev/AppNotifications/AppNotificationProgressData.h
```diff
@@ -19,7 +19,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
         void Status(hstring const& progressStatus);
 
     private:
-        uint32_t m_sequenceNumber{};
+        uint32_t m_sequenceNumber = 1;
```
loneursid, line 22: I thought brace initializers were prefered

dev/AppNotifications/AppNotificationProgressData.cpp
```diff
@@ -4,6 +4,12 @@
 
 namespace winrt::Microsoft::Windows::AppNotifications::implementation
 {
+    AppNotificationProgressData::AppNotificationProgressData(uint32_t sequenceNumber)
+    {
+        auto lock{ m_lock.lock_exclusive() };
+        THROW_HR_IF(E_INVALIDARG, sequenceNumber == 0); // The sequence number is always greater than 0
```
loneursid, line 10: This could be moved before the lock

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -368,7 +367,7 @@ winrt::Microsoft::Windows::AppNotifications::AppNotification Microsoft::Windows:
         progressData.Title(wil::str_raw_ptr(title));
 
         double progressValue{};
-        toastProgressData->get_Value(&progressValue);
+        THROW_IF_FAILED(toastProgressData->get_Value(&progressValue));
```
loneursid, line 370: good catch

dev/AppNotifications/AppNotificationProgressData.h
hulumane: Its a primitive type. So = should be fine. It's mainly preference at the end of the day.

dev/AppNotifications/AppNotificationProgressData.cpp
hulumane: Actually there is no point in having a lock in consructors. So removing it there. However, I will fix this in the getters/setters

dev/AppNotifications/AppNotificationProgressData.h
```diff
@@ -6,7 +6,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
     struct AppNotificationProgressData : AppNotificationProgressDataT<AppNotificationProgressData>
     {
         AppNotificationProgressData() = default;
-
+        AppNotificationProgressData(uint32_t sequenceNumber);
```
robertzhou-wpn, line 9: Nit: can we use default parameter here like: AppNotificationProgressData(uint32_t sequenceNumber =1)?

dev/AppNotifications/NotificationProperties.cpp
```diff
@@ -131,8 +134,11 @@ STDMETHODIMP_(HRESULT __stdcall) NotificationProperties::get_ExpiresOnReboot(_Ou
 STDMETHODIMP_(HRESULT __stdcall) NotificationProperties::get_ToastProgressData(_Out_ ToastABI::IToastProgressData** progressData) noexcept
 {
     auto lock{ m_lock.lock_shared() };
-
-    m_toastProgressData.copy_to(progressData);
+    *progressData = nullptr;
```
danielayala94, line 137: Nit: this is not needed to happen inside the lock.

danielayala94 (approved):

dev/AppNotifications/NotificationProperties.cpp
hulumane: Fixing!

dev/AppNotifications/AppNotificationProgressData.h
hulumane: This would assume that we provide a default constructor in our projections. However, there is no default projection per API review feedback. We only have an overload ctor.

PR #2149, opened 2022-02-24
Publish all BuildOutput as a pipeline artifact, to enable running additional tests in the future.

jefgen:
This change modifies the build pipeline to publish the entire `BuildOutput` folder as an artifact, so that we can download it later  in the test stage of the pipeline, and then push it over to Helix. (This is make enabling more tests in Helix easier in the future.)

This also removes the previous build artifact that was published containing the MRT Core test binaries, and uses the `BuildOutput` artifact instead for running the MRT cores. 
(So this is doing both Step 1 & 2 together at the same time.)

### DIFF
--- a/build/AzurePipelinesTemplates/WindowsAppSDK-RunHelixTests-Job.yml
+++ b/build/AzurePipelinesTemplates/WindowsAppSDK-RunHelixTests-Job.yml
@@ -145,7 +145,7 @@ jobs:
   - task: CopyFiles@2
     displayName: 'Copy Microsoft.Windows.ApplicationModel.Resources.winmd to test root'
     inputs:
-      sourceFolder: $(testPayloadDir)\MrtCoreUnpackagedTests\MrtCoreUnpackagedTests
+      sourceFolder: $(testPayloadDir)\BuildOutput\$(buildConfiguration)\$(buildPlatform)\MrtCoreUnpackagedTests
       contents: |
         Microsoft.Windows.ApplicationModel.Resources.winmd
       targetFolder: $(testPayloadDir)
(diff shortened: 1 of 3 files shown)


loneursid (approved):
Confirmed that the build output by adding and running some non-MRT unit tests with helix. The tests were found and executed has expected.

Thanks for doing that work.

build/AzurePipelinesTemplates/WindowsAppSDK-RunHelixTests-Job.yml
```diff
@@ -93,21 +96,20 @@ jobs:
...
-    displayName: 'Publish testPayloadDir'
-    condition: always()
-    inputs:
-      PathtoPublish: $(testPayloadDir)
-      artifactName: HelixTestPayloadDir_$(buildPlatform)_$(buildConfiguration)
+  - powershell: |
+      Write-Host ""
+      Write-Host "$(Build.SourcesDirectory)"
+      Tree /F /A $(Build.SourcesDirectory)
+    displayName: 'DIAG: dir'
```
DrusTheAxe, line 112: ?

DrusTheAxe (approved):
Some debuggery good present but otherwise looks sweet. Thanks for doing this!

build/AzurePipelinesTemplates/WindowsAppSDK-RunHelixTests-Job.yml
jefgen: Ah, sorry about this. I had added this to double-check that the full BuildOutput was being downloaded. I'll remove it.

PR #2152, opened 2022-02-24
Implement Unregister APIs for PushNotificationManager

sharath2727:


dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -30,6 +30,7 @@ constexpr std::wstring_view expectedPushServerArgs = L"----WindowsAppRuntimePush
 
 static wil::unique_event g_waitHandleForArgs;
 static winrt::guid g_comServerClsid{ GUID_NULL };
+static bool registering{ false };
```
pmpurifoy, line 33: Can this be a member variable instead?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -202,9 +203,17 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
                     // Need to recreate the Foreground sinks since channel creation removes the old sink
                     RegisterForegroundSinkHelper();
 
+                    auto channel{ winrt::make<PushNotificationChannel>(channelInfo) };
+                 //   {  };
```
pmpurifoy, line 207: Remove this line.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -447,44 +486,95 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        {
+            try
+            {
+                bool channelExists{ false };
+                {
+                    auto lock{ m_lock.lock_exclusive() };
+                    channelExists = (m_channel != nullptr) ? true : false;
+                }
+
+                if (channelExists)
```
pmpurifoy, line 522: This can move into the lock exclusive block above

pmpurifoy (approved):

dev/PushNotifications/PushNotificationManager.cpp
sharath2727: No I think it is fine let this be static.

PR #2146, opened 2022-02-23
Add push notification tests to Helix

loneursid:

PR #2153, opened 2022-02-24
Try adding push notification unit tests to Helix

loneursid:


danielayala94:
Can you tell me what is the difference between #2146 and this PR? I know the changes are different, but both are suggesting onboarding the push notification tests into Helix 😊

loneursid:
@danielayala94 - They're just different experiment based on work done by Howard and Jeff Genovy. They're all experiments and not intended to be merged anytime soon. They need to be open as PR (even draft) as it's the only way to run the pipelines.

PR #2166, opened 2022-02-25
test

kythant:


build/ProjectReunion-BuildFoundation.yml
```diff
@@ -9,7 +9,7 @@ parameters:
     default: False
   - name: "IgnoreFailures"
     type: boolean
-    default: False
+    default: True
```
DrusTheAxe, line 12: Dupe of #2173 ?

kythant: Was testing the build validation policies. Not meant to be a real PR

PR #2173, opened 2022-02-25
Nothing makes sense

kythant:


kythant:
image
Going to just live with this pipeline triggering two builds

build/ProjectReunion-BuildFoundation.yml
```diff
@@ -9,7 +9,7 @@ parameters:
     default: False
   - name: "IgnoreFailures"
     type: boolean
-    default: False
+    default: True
```
DrusTheAxe, line 12: Why?

kythant: Was testing the build validation policies. Not meant to be a real PR

PR #2184, opened 2022-02-28
Remove the Handled property from PushNotificationReceivedEventArgs

David-Quero:
Update the idl and all the callers on the WindowsAppSDK and the demo and test apps.
Verified that the Push Notification tests keep passing.

### DIFF
--- a/dev/PushNotifications/PushNotificationManager.cpp
+++ b/dev/PushNotifications/PushNotificationManager.cpp
@@ -600,8 +600,16 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
         auto args = winrt::make<winrt::Microsoft::Windows::PushNotifications::implementation::PushNotificationReceivedEventArgs>(payload, length);
 
         auto lock{ m_lock.lock_exclusive() };
-        m_foregroundHandlers(*this, args);
-        *foregroundHandled = args.Handled();
+        if (m_foregroundHandlers)
+        {
+            m_foregroundHandlers(*this, args);
+            *foregroundHandled = true;
+        }
+        else
+        {
+            *foregroundHandled = false;
+        }   
+        
         return S_OK;
     }
     CATCH_RETURN()
--- a/dev/PushNotifications/PushNotificationManager.cpp
+++ b/dev/PushNotifications/PushNotificationManager.cpp
@@ -599,7 +599,7 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
     {
         auto args = winrt::make<winrt::Microsoft::Windows::PushNotifications::implementation::PushNotificationReceivedEventArgs>(payload, length);
 
-        auto lock{ m_lock.lock_exclusive() };
+        auto lock{ m_lock.lock_shared() };
         if (m_foregroundHandlers)
         {
             m_foregroundHandlers(*this, args);
--- a/dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
+++ b/dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
@@ -35,7 +35,6 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
     }
 
     PushNotificationReceivedEventArgs::PushNotificationReceivedEventArgs(winrt::PushNotificationReceivedEventArgs const& args):
-        m_args(args),
         m_rawNotificationPayload(BuildPayload(args.RawNotification().ContentBytes())),
         m_unpackagedAppScenario(false)
     {
--- a/dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
+++ b/dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
@@ -115,33 +115,5 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
             m_backgroundTaskInstance.Canceled(token);
         }
     }
-
-    bool PushNotificationReceivedEventArgs::Handled()
-    {
-        if (!m_unpackagedAppScenario)
-        {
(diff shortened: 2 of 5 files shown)


danielayala94 (approved):

dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
```diff
@@ -115,33 +115,5 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-    {
-        if (!m_unpackagedAppScenario)
-        {
-            THROW_HR_IF_NULL_MSG(E_ILLEGAL_METHOD_CALL, m_args, "Background activation cannot call this.");
-
-            return m_args.Cancel();
-        }
-        else
-        {
-            return m_handledUnpackaged;
```
danielayala94, line 129: m_handledUnpackaged isn't used anywhere now, please remove it in the header.

loneursid:
re-ran the pipelines as some processes failed on x64. The failure doesn't seem related to this PR.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -600,8 +600,16 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
         auto args = winrt::make<winrt::Microsoft::Windows::PushNotifications::implementation::PushNotificationReceivedEventArgs>(payload, length);
 
         auto lock{ m_lock.lock_exclusive() };
-        m_foregroundHandlers(*this, args);
-        *foregroundHandled = args.Handled();
+        if (m_foregroundHandlers)
```
pmpurifoy, line 603: Do we need a lock_shared to get grab the bool from m_foregroundHandlers?

David-Quero: What difference would it make from the current lock_esclusive?

hulumane: Paul is correct. This should be a shared lock. You aren't modifying what m_foreGroundHandlers is pointing to. So shared since it is equivalent to a read operation.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -600,8 +600,16 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
         auto args = winrt::make<winrt::Microsoft::Windows::PushNotifications::implementation::PushNotificationReceivedEventArgs>(payload, length);
 
         auto lock{ m_lock.lock_exclusive() };
```
hulumane, line 602: Curious but why do we have the argument foregroundHandled at all in this function? Shouldn't it be removed?

dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
hulumane: Yes. Please delete all unised variables, function arguments etc. that are no longer being used.

dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
```diff
@@ -115,33 +115,5 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
             m_backgroundTaskInstance.Canceled(token);
         }
     }
-
-    bool PushNotificationReceivedEventArgs::Handled()
-    {
-        if (!m_unpackagedAppScenario)
-        {
-            THROW_HR_IF_NULL_MSG(E_ILLEGAL_METHOD_CALL, m_args, "Background activation cannot call this.");
```
hulumane, line 123: Delete m_args since it isn't being used anywhere else.

dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
```diff
@@ -115,33 +115,5 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-
-            return m_args.Cancel();
```
hulumane, line 125: m_args not used anywhere else. Delete member variable.

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: This is for unpackaged scenarios attempting to activate a foreground handler. If no handlers are registered, it will protocol launch a new instance.

pmpurifoy: Or in the packaged treated as unpackaged, CoCreateInstance.

hulumane: Can you point me to where in our code is *foregroundHandled being used to do this?

hulumane: Is it this?
PushNotificationManager::OnRawNotificationReceived

hulumane: I don't see any CoCreateInstance for packaged applications being done here.

pmpurifoy: NotificationListener::OnRawNotificationReceived in the Long Running Process

pmpurifoy: ForegroundSinkManager::InvokeForegroundHandlers

hulumane: What about packaged apps that are getting activated through COM? 
PushNotificationManager::OnRawNotificationReceived doesnt do  anything to account for them and launching a new process.

pmpurifoy: OnRawNotificationReceived is only activated if a NotificationSink is registered. Sink registration should be wrapped in a foreground handler check. I added this in my PR to force BI to activate the packaged app.

https://github.com/microsoft/WindowsAppSDK/pull/2187

hulumane (approved):

hulumane (approved):

Issue #2183, opened 2022-02-26
Ensure all objects are destroyed before unloading dll in PushNotificationManager


pmpurifoy:
### Describe the bug

Currently PushNotificationManager is a static variable inside the SDK and need to make sure that everything is run down correctly even if the dll is unload.

### Steps to reproduce the bug

PushNotificationManager.cpp/h

pmpurifoy:
PR merged.

PR #2187, opened 2022-02-28
Move wait handles to statics in class, add singleuse in the case of no handlers

pmpurifoy:


dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -610,18 +646,80 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-        if (!foregroundHandled)
+    void PushNotificationManager::Run(winrt::IBackgroundTaskInstance const& taskInstance)
+    {
+        auto appProperties = winrt::CoreApplication::Properties();
+        // This function can be triggered by either OS background infrastructure
+        // or by the PushNotificationsLongRunningProcess.
+        if (!m_firstNotificationReceived)
         {
-            if (!AppModel::Identity::IsPackagedProcess())
+            m_firstNotificationReceived = true;
```
danielayala94, line 670: Should we guard this member variable?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -610,18 +646,80 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                std::wstring payloadAsWstring{ winrt::Windows::Foundation::Uri::UnescapeComponent(payloadAsEscapedWstring) };
+                return winrt::make<winrt::Microsoft::Windows::PushNotifications::implementation::PushNotificationReceivedEventArgs>(payloadAsWstring);
+            }
+        }
+
+        const DWORD receiveArgsTimeoutInMSec{ 2000 };
+        
+        THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_TIMEOUT), !s_waitHandleForArgs.wait(receiveArgsTimeoutInMSec));
+
+        // If COM static store was uninit, let it throw
```
danielayala94, line 722: nit: 
```suggestion
        // If the COM static store was uninitialized, let it throw
```

dev/PushNotifications/PushNotificationManager.h
```diff
@@ -32,20 +36,56 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+    struct PushNotificationManagerFactory : winrt::implements<PushNotificationManagerFactory, IClassFactory>
+    {
+        STDMETHODIMP CreateInstance(_In_opt_ IUnknown* aggregateInterface, _In_ REFIID interfaceId, _Outptr_ VOID** object) noexcept final try
+        {
+            RETURN_HR_IF(CLASS_E_NOAGGREGATION, aggregateInterface != nullptr);
+            return PushNotificationManager::Default().as(interfaceId, object);
+        }
+        CATCH_RETURN()
+
+            STDMETHODIMP LockServer(BOOL fLock) noexcept final
```
danielayala94, line 76: nit: fix indentation.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -330,7 +340,21 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                    registeredEvent = (bool)m_foregroundHandlers;
+                }
+
+                if (registeredEvent)
+                {
+                    {
+                        auto lock{ m_lock.lock_exclusive() };
+                        m_sinkRegisteredWithPlatform = true;
+                    }
+                    THROW_IF_FAILED(PushNotifications_RegisterNotificationSinkForFullTrustApplication(appUserModelId.get(), PushNotificationManager::Default().as<ABI::Microsoft::Internal::PushNotifications::INotificationListener>().get()));
```
danielayala94, line 355: Can you introduce a namespace for `ABI::Microsoft::Internal::PushNotifications`?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -421,12 +446,14 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
                     THROW_IF_FAILED(::CoRegisterClassObject(
-                        registeredClsid,
-                        winrt::make<PushNotificationBackgroundTaskFactory>().get(),
+                        m_registeredClsid,
+                        winrt::make<PushNotificationManagerFactory>().get(),
                         CLSCTX_LOCAL_SERVER,
-                        REGCLS_MULTIPLEUSE,
-                        &m_comActivatorRegistration));
+                        m_foregroundHandlers ? REGCLS_MULTIPLEUSE : REGCLS_SINGLEUSE,
+                        & m_comActivatorRegistration));
```
danielayala94, line 460: nit: remove space on `& m_comActivatorRegistration`

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -610,18 +646,80 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        THROW_HR_IF(E_NOT_SET, !foregroundHandled);
+
+        return S_OK;
+    }
+    CATCH_RETURN();
 
-        if (!foregroundHandled)
+    void PushNotificationManager::Run(winrt::IBackgroundTaskInstance const& taskInstance)
+    {
+        auto appProperties = winrt::CoreApplication::Properties();
```
danielayala94, line 665: Not a big fan of declaring the variable here. It isn't used at all outside of the if-else blocks. I think it's fine to declare it once in each block.

danielayala94 (commented):
We really need some visual content representing how activation flows look like. It is getting increasingly harder to keep doing meaningful code reviews on this area.

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: Yup

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: This should be moved after the firstNotificationReceived check.

danielayala94 (approved):

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -47,6 +47,9 @@ using namespace Microsoft::Windows::AppNotifications::Helpers;
 
 namespace winrt::Microsoft::Windows::AppNotifications::implementation
 {
+    // Must be static for AppNotificationManager::Deserialize
+    static wil::unique_event s_waitHandleForArgs;
```
hulumane, line 51: Why static? Can you make it a member variable?

dev/AppNotifications/AppNotificationManager.h
```diff
@@ -37,6 +39,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
             [[maybe_unused]] NOTIFICATION_USER_INPUT_DATA const* data,
             [[maybe_unused]] ULONG dataCount) noexcept;
 
+        static winrt::Windows::Foundation::IInspectable Deserialize(winrt::Windows::Foundation::Uri const& uri);
```
hulumane, line 42: Why static?

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -355,4 +358,12 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
 
         co_return toastNotifications;
     }
+
+    winrt::Windows::Foundation::IInspectable AppNotificationManager::Deserialize(winrt::Windows::Foundation::Uri const& /* uri */)
```
hulumane, line 362: Why take a uri that isn't being used?

pmpurifoy: ExtensionMap forces a specific type winrt::Windows::Foundation::IInspectable(*factory)(winrt::Windows::Foundation::Uri const& uri) as Deserialize. We must at least have the param

dev/AppNotifications/AppNotificationManager.cpp
pmpurifoy: Other Deserialize functions like LaunchActivatedEventArgs, FileActivatedEvent args use the static keyword with Deserialize. The ExtensionMap in ExtensionContract.h is expecting a winrt::Windows::Foundation::IInspectable(*factory)(winrt::Windows::Foundation::Uri const& uri). I guess by using the & operator on a static function you get the factory. Compile errors occur when removing the static tag.

dev/AppNotifications/AppNotificationManager.h
pmpurifoy: Other Deserialize functions like LaunchActivatedEventArgs, FileActivatedEvent args use the static keyword with Deserialize. The ExtensionMap in ExtensionContract.h is expecting a winrt::Windows::Foundation::IInspectable(*factory)(winrt::Windows::Foundation::Uri const& uri). I guess by using the & operator on a static function you get the factory. Compile errors occur when removing the static tag.

dev/AppNotifications/AppNotificationManager.cpp
hulumane: Can you do this and get rid of the static in the mapping? 
&winrt::Microsoft::Windows::AppNotifications::implementation::AppNotificationManager::GetDefault()::Deserialize

hulumane: I guess you answered the question : )

pmpurifoy: @aeloros We would like to reference a member variable in our Deserialize function but we are having issues since it can't be static. What should we do in this scenario? Is there a limitation with Deserialize that can only be a static function? My guess is that the & operator returns the factory for a static function.

If you can help us get to winrt::Windows::Foundation::IInspectable(*factory)(winrt::Windows::Foundation::Uri const& uri), that would be great :)

danielayala94: Is this a blocker for checking in? It seems like we can afford to address this as a bug in the big PR.

hulumane: I talked to @pmpurifoy . I believe we have a solution that will work with these limitations. He will give it a shot and reply back.

hulumane, line 51: We need to move the Manager classes to the COM static store to ensure that we have no issues with cleanup if COM rundown occurs. Please file a separate bug for that.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -156,7 +156,20 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                        bool registeredEvent{ false };
+                        {
+                            auto lock{ m_lock.lock_shared() };
+                            registeredEvent = (bool) m_foregroundHandlers;
+                        }
+
+                        if (registeredEvent)
+                        {
+                            {
+                                auto lock{ m_lock.lock_exclusive() };
```
hulumane, line 168: Recommend setting this to true only after the RegisterSinkWithPlatform operation below succeeded.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -156,7 +156,20 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                                m_sinkRegisteredWithPlatform = true;
```
hulumane, line 169: What is the rationale for this bool?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -330,21 +343,40 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                {
+                    auto lock{ m_lock.lock_shared() };
+                    registeredEvent = (bool)m_foregroundHandlers;
+                }
+
+                if (registeredEvent)
+                {
+                    {
+                        auto lock{ m_lock.lock_exclusive() };
+                        m_sinkRegisteredWithPlatform = true;
```
hulumane, line 356: Set this to true only after the sink has actually registered.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -330,21 +343,40 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                    }
+                    THROW_IF_FAILED(PushNotifications_RegisterNotificationSinkForFullTrustApplication(appUserModelId.get(), PushNotificationManager::Default().as<PushNotificationsABI::INotificationListener>().get()));
+                }
+
                 auto scopeExitToCleanSink{ wil::scope_exit([&]()
                 {
+                    {
+                        auto lock{ m_lock.lock_exclusive() };
+                        m_sinkRegisteredWithPlatform = false;
```
hulumane, line 365: Set this to false only after the unregister operation succeeds

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -330,21 +343,40 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                    }
                     THROW_IF_FAILED(PushNotifications_UnregisterNotificationSinkForFullTrustApplication(appUserModelId.get()));
```
hulumane, line 367: Only Unregister if this is set to true.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -601,7 +648,7 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
 
         auto lock{ m_lock.lock_exclusive() };
         m_foregroundHandlers(*this, args);
-        *foregroundHandled = args.Handled();
+        *foregroundHandled
```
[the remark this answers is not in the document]
pmpurifoy: Just rebased and fixed

dev/PushNotifications/PushNotificationManager.cpp
hulumane: Ideally the way this is structure, it shouldnt matter, bt still better for correctness sake IMHO

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -618,18 +665,89 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        }
+        if (!firstNotificationReceived)
         {
-            if (!AppModel::Identity::IsPackagedProcess())
             {
-                wil::unique_cotaskmem_string processName;
-                THROW_IF_FAILED(GetCurrentProcessPath(processName));
-                THROW_IF_FAILED(PushNotificationHelpers::ProtocolLaunchHelper(processName.get(), payloadLength, payload));
+                auto lock{ m_lock.lock_exclusive() };
+                m_firstNotificationReceived = true;
```
hulumane, line 687: No lock?? :(

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -618,18 +665,89 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+    void PushNotificationManager::Run(winrt::IBackgroundTaskInstance const& taskInstance)
+    {
+        // This function can be triggered by either OS background infrastructure
+        // or by the PushNotificationsLongRunningProcess.
+        bool firstNotificationReceived{ false };
+        {
+            auto lock{ m_lock.lock_shared() };
+            firstNotificationReceived = m_firstNotificationReceived;
+        }
+        if (!firstNotificationReceived)
```
hulumane, line 683: Race condition

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -618,18 +665,89 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+            if (pos == std::wstring::npos) // Any launch kind that is not PushNotification
+            {
+                THROW_HR_IF(E_UNEXPECTED, m_foregroundHandlers);
+
+                winrt::guid registeredClsid{ GUID_NULL };
+                THROW_IF_FAILED(PushNotificationHelpers::GetComRegistrationFromRegistry(c_expectedPushServerArgs.data(), registeredClsid));
+
+                auto backgroundTask{ winrt::create_instance<winrt::IBackgroundTask>(registeredClsid, CLSCTX_ALL) };
+                backgroundTask.Run(taskInstance);
+                return;
```
hulumane, line 701: bad pattern. Suggest that you structure the code in such a way that we don't return in the middle of a function.

Rationale: If we need to add things later to this, the premature returning could be problematic although it is technically correct now.

hulumane: One way to fix this is to use an if {{} else based on the condition above.

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: So that we only register a sink if the user defined a foreground handler. Any other activation should trigger background activation.

dev/AppNotifications/AppNotificationManager.cpp
pmpurifoy: https://github.com/microsoft/WindowsAppSDK/issues/2183

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: We only want to register a sink if there are foreground handlers registered. Otherwise we want background activation to handle delivering payload.

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -47,12 +47,29 @@ using namespace Microsoft::Windows::AppNotifications::Helpers;
 
 namespace winrt::Microsoft::Windows::AppNotifications::implementation
 {
+    // Must be static for AppNotificationManager::Deserialize
```
hulumane, line 50: Stale comment. Remove.

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -47,12 +47,29 @@ using namespace Microsoft::Windows::AppNotifications::Helpers;
...
     winrt::Microsoft::Windows::AppNotifications::AppNotificationManager AppNotificationManager::Default()
     {
         static auto appNotificationManager{winrt::make<AppNotificationManager>()};
         return appNotificationManager;
     }
 
+    winrt::Windows::Foundation::IInspectable AppNotificationManager::Deserialize(winrt::Windows::Foundation::Uri const& /* uri */)
+    {
+        auto appNotificationManager{ Default() };
+        auto deserializer{ appNotificationManager.as<INotificationDeserializer>() };
```
hulumane, line 61: Please call it INotificationManagerDeserializer since this is NotificationManager

dev/AppNotifications/AppNotificationManager.h
```diff
@@ -2,18 +2,22 @@
...
+constexpr PCWSTR c_appNotificationContractId = L"Windows.Toast";
 
 typedef winrt::Windows::Foundation::TypedEventHandler<
     winrt::Microsoft::Windows::AppNotifications::AppNotificationManager,
     winrt::Microsoft::Windows::AppNotifications::AppNotificationActivatedEventArgs> NotificationActivationEventHandler;
 
 namespace winrt::Microsoft::Windows::AppNotifications::implementation
 {
-    struct AppNotificationManager : AppNotificationManagerT<AppNotificationManager, INotificationActivationCallback>
+    struct AppNotificationManager : AppNotificationManagerT<AppNotificationManager, INotificationActivationCallback, INotificationDeserializer>
```
hulumane, line 15: INotificationManagerDeserializer

dev/AppNotifications/AppNotificationManager.h
```diff
@@ -37,11 +41,14 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
             [[maybe_unused]] NOTIFICATION_USER_INPUT_DATA const* data,
             [[maybe_unused]] ULONG dataCount) noexcept;
 
+        // INotificationDeserializer
```
hulumane, line 44: INotificationManagerDeserializer

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -55,7 +52,7 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
     const HRESULT WNP_E_NOT_CONNECTED = static_cast<HRESULT>(0x880403E8L);
     const HRESULT WNP_E_RECONNECTING = static_cast<HRESULT>(0x880403E9L);
     const HRESULT WNP_E_BIND_USER_BUSY = static_cast<HRESULT>(0x880403FEL);
-
+
```
hulumane, line 55: Remove spaces

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -117,6 +114,31 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        else
+        {
+            // Need to mock a RawNotification object instead of winrt boxing: https://github.com/microsoft/WindowsAppSDK/issues/2075
+            return winrt::Windows::ApplicationModel::Core::CoreApplication::Properties().Lookup(LRP_ACTIVATED_EVENT_ARGS_KEY);
+        }
+    }
+
     winrt::IAsyncOperationWithProgress<winrt::Microsoft::Windows::PushNotifications::PushNotificationCreateChannelResult, winrt::Microsoft::Windows::PushNotifications::PushNotificationCreateChannelStatus> PushNotificationManager::CreateChannelAsync(const winrt::guid remoteId)
     {
         THROW_HR_IF(E_NOTIMPL, !::Microsoft::Windows::PushNotifications::Feature_PushNotifications::IsEnabled());
```
hulumane, line 144: Delete everywhere that is not default. Already guarded in Default()

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -156,7 +178,20 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
 
                         // Register a sink with platform which is initialized in the current process
-                        THROW_IF_FAILED(PushNotifications_RegisterNotificationSinkForFullTrustApplication(appUserModelId.get(), PushNotificationManager::Default().as<ABI::Microsoft::Internal::PushNotifications::INotificationListener>().get()));
+                        bool registeredEvent{ false };
+                        {
+                            auto lock{ m_lock.lock_shared() };
+                            registeredEvent = (bool) m_foregroundHandlers;
+                        }
+
+                        if (registeredEvent)
```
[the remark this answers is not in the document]
pmpurifoy: m_sinkRegisteredWithPlatform only applies to PackagedAppScenario. Foreground delegate is still registered for unpackaged app scenarios

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: Should retrieving firstNotificationReceived be lock_exclusive()?

hulumane: What do we do in AppNotificationManager for races?

pmpurifoy: Looks like we lock_exclusive() the entire block.

hulumane: Yeah. You have your answer then :)

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -471,7 +534,20 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
             if (PushNotificationHelpers::IsPackagedAppScenario())
             {
                 auto appUserModelId{ PushNotificationHelpers::GetAppUserModelId() };
-                THROW_IF_FAILED(PushNotifications_UnregisterNotificationSinkForFullTrustApplication(appUserModelId.get()));
+                bool sinkRegisteredWithPlatform{ false };
+                {
+                    auto lock{ m_lock.lock_exclusive() };
+                    sinkRegisteredWithPlatform = m_sinkRegisteredWithPlatform;
```
hulumane, line 540: What is this local used for?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -471,7 +534,20 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
             if (PushNotificationHelpers::IsPackagedAppScenario())
             {
                 auto appUserModelId{ PushNotificationHelpers::GetAppUserModelId() };
-                THROW_IF_FAILED(PushNotifications_UnregisterNotificationSinkForFullTrustApplication(appUserModelId.get()));
+                bool sinkRegisteredWithPlatform{ false };
+                {
+                    auto lock{ m_lock.lock_exclusive() };
```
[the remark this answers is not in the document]
hulumane: Honestly, don't see the point of having 2 locks. Just clump this in with the lock below and get rid of the local.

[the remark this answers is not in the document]
pmpurifoy: We need to read the value before we start setting the value which is the reason for the lock

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: See my fix for this.

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -49,8 +49,32 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
 {
     winrt::Microsoft::Windows::AppNotifications::AppNotificationManager AppNotificationManager::Default()
     {
-        static auto appNotificationManager{winrt::make<AppNotificationManager>()};
-        return appNotificationManager;
+        auto appProperties{ winrt::CoreApplication::Properties() };
+        // Store the PushNotificationManager in the COM static store
+        auto storedAppNotificationManager{ appProperties.TryLookup(STORED_APPNOTIFICATION_MANAGER_KEY) };
```
hulumane, line 54: No lock :(

hulumane: And you have read-write races too because of the absence of a lock around this whole operation.

hulumane (approved):

PR #2191, opened 2022-03-01
Add C++savvy API over Bootstrap's Flat-C API

DrusTheAxe:
C++ API over the Bootstrap Flat-C API

Provides equivalent goodness as the C# API does for .NET developers -- overloading, automatic call to Shutdown, yadda yadda. Notably:

* All symbols in the `Microsoft::Windows::ApplicationModel::DynamicDependency::Bootstrap` namespace
* `InitializeNoThrow()` returns HRESULT vs `Initialize()` throws exceptions vs `InitializeFailFast()` aborts on error
* `Shutdown()` for completeness
* `Initialize()` and `InitializeNoThrow()` return an scoped cleanup object whose destructor calls `Shutdown()`

For example:
```
#include <WindowsAppSDK-VersionInfo.h>
#include <MddBootstrap.h>

int main()
{
    const auto hr{ ::Microsoft::Windows::ApplicationModel::DynamicDependency::Bootstrap::Initialize_nothrow() };
    if (FAILED(hr))
    {
        wprintf(L"Error 0x%X in Bootstrap initialization\n", hr);
        return hr;
    }
    ...
    ::Microsoft::Windows::ApplicationModel::DynamicDependency::Bootstrap::Shutdown();
}
```

or for those who prefer exceptions

```
#include <WindowsAppSDK-VersionInfo.h>
#include <MddBootstrap.h>

int main()
{
    try
    {
        auto mddBootstrapCleanup = ::Microsoft::Windows::ApplicationModel::DynamicDependency::Bootstrap::Initialize());
        ...
    }
    catch (const winrt::hresult_error& ex)
    {
        const auto hr{ ex.code() };
        wprintf(L"Error 0x%X in Bootstrap initialization\n", hr);
        return hr;
    }
    return 0;
}
```

and of course the abort/fail-fast model

```
#include <WindowsAppSDK-VersionInfo.h>
#include <MddBootstrap.h>

int main()
{
    auto mddBootstrapCleanup = ::Microsoft::Windows::ApplicationModel::DynamicDependency::Bootstrap::InitializeFailFast());
    ...
    return 0;
}
```


specs/dynamicdependencies/DynamicDependencies.md
```diff
@@ -1281,6 +1281,101 @@ STDAPI MddBootstrapInitialize(
...
+    /// @see Initialize(uint32_t, PCWSTR, PackageVersion)
+    /// @see Initialize_nothrow(uint32_t, PCWSTR, PackageVersion)
+    /// @see Shutdown()
+    WI_NODISCARD inline unique_mddbootstrapshutdown_call Initialize_failfast(
+        uint32_t majorMinorVersion = WINDOWSAPPSDK_RELEASE_MAJORMINOR,
+        PCWSTR versionTag = WINDOWSAPPSDK_RELEASE_VERSION_TAG_W,
+        PackageVersion minVersion = WINDOWSAPPSDK_RUNTIME_VERSION_UINT64);
+
+#if defined(WIL_ENABLE_EXCEPTIONS)
+    /// Call MddBootstrapInitialize and throws an exception if it fails; returns an RAII objecct that reverts
```
oldnewthing, line 1337:
```suggestion
    /// Call MddBootstrapInitialize and throws a wil::ResultException exception if it fails; returns an RAII object that reverts
```
Should clarify what kind of exception, so people can catch it.

DrusTheAxe: Should we? Or is that an implementation detail?

If we document it we're declaring the exception as part of the interface. Changing that in the future would be a breaking change.

C++ coding guidelines say (E.30: Don’t use exception specifications](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Re-specifications). Are you saying we should document exceptions but not declare them?

Do we have precedent or is this new territory we should bring up at API Review?

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.h
```diff
@@ -1,12 +1,12 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
```
jonwis, line 2: >Licensed [](http://example.com/codeflow?start=3&length=8)

Do you still need `#pragma once` here?

DrusTheAxe: No. VS got smarter some years back recognizing the decades old pattern

```
#if !defined(FOO)
#define FOO
...
#endif
```

as "this file has zero compilation content don't reprocess" just like `#pragma once`.

A popular enough topic it actually has its own Wikipedia page https://en.wikipedia.org/wiki/Pragma_once

specs/dynamicdependencies/DynamicDependencies.md
oldnewthing: Document the exceptions, but don't put them in the function prototype. (Starting in C++17, if you try to put them in the prototype, you get a compiler error.)

This is consistent with the C++ standard library, which documents what exceptions can be thrown by each method in the specification, but does not use `throw(blah)` annotations on the prototype.

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.h
```diff
@@ -26,4 +26,136 @@ STDAPI MddBootstrapInitialize(
...
+    /// for use by the current process. If multiple packages meet the criteria the best
+    /// candidate is selected.
+    ///
+    /// @param majorMinorVersion major and minor version of Windows App SDK's framework package, encoded as `0xMMMMNNNN` where M=Major, N=Minor (e.g. 1.2 == 0x00010002).
+    /// @param versionTag version tag (if any), e.g. "preview1".
+    /// @param minVersion the minimum version to use
+    /// @see Initialize(uint32_t, PCWSTR, PackageVersion)
+    /// @see Initialize_nothrow(uint32_t, PCWSTR, PackageVersion)
+    /// @see Shutdown()
+    WI_NODISCARD inline unique_mddbootstrapshutdown_call Initialize_failfast(uint32_t majorMinorVersion = WINDOWSAPPSDK_RELEASE_MAJORMINOR,
```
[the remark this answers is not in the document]
DrusTheAxe: API Design feedback: use MixedCase (aka PascalCase) for C++ APIs

specs/dynamicdependencies/DynamicDependencies.md
DrusTheAxe: Per API Design:

* DO use C++/WinRT for error handling and exceptions/throwing
* DO NOT require WIL to use C++ APIs
* DO use std::unique_ptr<> with custom deleter for the cleanup method type (instead of wil::unique_call)
* DO document throwing a C++/WinRT exception (not just a std::exception)
  * the thrown C++/WinRT exception is part of the API contract. It would be a Breaking Change if it changes in the future

DrusTheAxe:
> C++ API over the Bootstrap Flat-C API

NOTE: The [user/drustheaxe/bootstrap-cpp-experimental PR](https://github.com/microsoft/WindowsAppSDK/pull/1586) aged out too badly so created a new branch incorporating those changes and feedback.

specs/dynamicdependencies/DynamicDependencies.md
```diff
@@ -1281,6 +1281,101 @@ STDAPI MddBootstrapInitialize(
 /// @warning Packages made available via MddBootstrapInitialize() and
 ///          the Dynamic Dependencies API should not be used after this call.
 STDAPI_(void) MddBootstrapShutdown() noexcept;
+
+// C++ friendly APIs
+#if defined(__cplusplus)
+#if defined(WINDOWSAPPSDK_RELEASE_MAJORMINOR) && defined(WINDOWSAPPSDK_RELEASE_VERSION_TAG_W) && defined(WINDOWSAPPSDK_RUNTIME_VERSION_UINT64)
```
jonwis, line 1287: Can you add a sample use of the types? The test is instructive, but providing that canonical `WinMain` instance would help.

jonwis (approved):

specs/dynamicdependencies/DynamicDependencies.md
DrusTheAxe: Thanks for the reminder! I thought about that and got distracted

PR #2192, opened 2022-03-01
RI feature branch WNP_ToastNotifications_L1 into main

loneursid:
Toast notifications is feature complete and ready to be added to the WindowsAppSDK v1.1


build/NuSpecs/AppxManifest.xml
```diff
@@ -55,6 +55,11 @@
 
           <!-- PowerNotifications -->
           <ActivatableClass ActivatableClassId="Microsoft.Windows.System.Power.PowerManager" ThreadingModel="both" />
+		  
+		  <!-- ToastNotifications -->
+		  <ActivatableClass ActivatableClassId="Microsoft.Windows.ToastNotifications.ToastActivationInfo" ThreadingModel="both" />
+		  <ActivatableClass ActivatableClassId="Microsoft.Windows.ToastNotifications.ToastAssets" ThreadingModel="both" />
+		  <ActivatableClass ActivatableClassId="Microsoft.Windows.ToastNotifications.ToastNotificationManager" ThreadingModel="both" />
```
sharath2727, line 62: should be AppNotifications?

build/build-mrt.yml
```diff
@@ -13,6 +13,8 @@ steps:
     arguments: /envonly $(buildPlatform)fre
     modifyEnvironment: true
 
+- template: AzurePipelinesTemplates\WindowsAppSDK-InstallWindowsSDK-Steps.yml
+
```
loneursid, line 17: This change is coming from main (through a merge) and thus, it shouldn't show up here.

dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/UnpackagedTests/MrtCoreUnpackagedTests.csproj
```diff
@@ -7,7 +7,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>MrtCoreUnpackagedTests</RootNamespace>
     <AssemblyName>MrtCoreUnpackagedTests</AssemblyName>
-    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.7.2</TargetFrameworkVersion>
```
loneursid, line 10: This change is coming from main (through a merge) and thus, it shouldn't show up here.

dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/UnpackagedTests/TestClass.cs
```diff
@@ -110,36 +110,52 @@ public class TestClass
...
+                        System.Threading.Thread.Sleep(200);
+                        if (i == 9)
+                        {
+                            throw e;
+                        }
+                    }
+                }
+            }
+        }
+
```
loneursid, line 135: This change is coming from main (through a merge) and thus, it shouldn't show up here. 

This applies to the other changes to that file also

build/build-mrt.yml
loneursid: Solved

dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/UnpackagedTests/MrtCoreUnpackagedTests.csproj
loneursid: Solved

dev/AppNotifications/NotificationProgressData.h
```diff
@@ -0,0 +1,27 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#pragma once
+
+#include <frameworkUdk/ToastNotificationsRT.h>
+
+struct NotificationProgressData : winrt::implements<NotificationProgressData, ::ABI::Microsoft::Internal::ToastNotifications::IToastProgressData>
```
brflynn, line 8: Why did we use HSTRING here instead of the new types?

dev/AppNotifications/NotificationProperties.cpp
```diff
@@ -0,0 +1,150 @@
...
+}
+
+NotificationProperties::NotificationProperties(winrt::AppNotification const& toastNotification)
+{
+    // Extract payload and convert it from XML to a byte array
+    auto payload = toastNotification.Payload();
+
+    auto payloadAsSimpleString = Helpers::WideStringToUtf8String(payload.c_str());
+
+    m_payload = wil::unique_cotaskmem_array_ptr<byte>(static_cast<byte*>(CoTaskMemAlloc(payloadAsSimpleString.size())), payloadAsSimpleString.size());
```
brflynn, line 34: We should check the return value here.

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -0,0 +1,395 @@
...
+
+            wil::unique_cotaskmem_string newNotificationGuidString;
+            THROW_IF_FAILED(StringFromCLSID(newNotificationGuid, &newNotificationGuidString));
+
+            std::wstring guidWideStr{ newNotificationGuidString.get() };
+            RegisterValue(hKey, L"NotificationGUID", reinterpret_cast<const BYTE*>(guidWideStr.c_str()), REG_SZ, guidWideStr.size() * sizeof(wchar_t));
+            return guidWideStr;
+        }
+
+        THROW_HR_IF(status, FAILED_WIN32(status));
```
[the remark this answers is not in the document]
hulumane: Thanks for pointing this out. Chris Guzak spotted this earlier too and let me know. There is a bug where a lot of the Registry access code is using the wrong wil macros to check for error status. And there was other fallout from this that I just discovered. I am fixing this and some registry corruption bugs in a seperate PR. I will move it up here once completed!

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -0,0 +1,395 @@
...
+    {
+        return;
+    }
+    std::wstring notificationAppId{ RetrieveNotificationAppId() };
+
+    wil::unique_hkey hKey;
+    //subKey: \Software\Classes\AppUserModelId\{AppGUID}
+    std::wstring subKey{ c_appIdentifierPath + notificationAppId };
+
+    THROW_IF_FAILED(RegDeleteKeyEx(
```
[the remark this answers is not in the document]
hulumane: Yup. See my answer above.

build/NuSpecs/AppxManifest.xml
hulumane: I'll fix this

dev/AppNotifications/AppNotificationUtility.cpp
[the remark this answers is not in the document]
pmpurifoy: Yup. All registry functions are being corrected in here.

https://github.com/microsoft/WindowsAppSDK/pull/2203

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -0,0 +1,395 @@
...
+        registeredGuid = comActivatorGuidString.get();
+    }
+
+    std::wstring notificationAppId{ RetrieveNotificationAppId() };
+    RegisterAssets(notificationAppId, registeredGuid);
+
+    return registeredGuid;
+}
+
+wil::unique_cotaskmem_string Microsoft::Windows::AppNotifications::Helpers::ConvertUtf8StringToWideString(unsigned long length, const byte* utf8String)
```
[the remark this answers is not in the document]
jefgen: Thanks @brflynn.
Good point. It looks like there is already a common helper for this in the file here `dev\Common\Microsoft.Utf8.h`. See the various `ToUtf16` methods.
(They look to also ensure that the string is always null terminated as well, which is good.)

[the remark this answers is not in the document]
hulumane: @danielayala94 or @pmpurifoy Can you look into fixing this?

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -0,0 +1,395 @@
...
+
+        notification.Progress(progressData);
+    }
+
+    unsigned long long expiry{};
+    THROW_IF_FAILED(properties->get_Expiry(&expiry));
+    FILETIME expiryFileTime{};
+    expiryFileTime.dwHighDateTime = expiry >> 32;
+    expiryFileTime.dwLowDateTime = static_cast<DWORD>(expiry);
+    auto expiryClock{ winrt::clock::from_file_time(expiryFileTime) };
```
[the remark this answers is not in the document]
hulumane: @danielayala94

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -0,0 +1,395 @@
...
+    THROW_IF_FAILED(SHGetPropertyStoreForWindow(hWindow.get(), IID_PPV_ARGS(propertyStore.put())));
+
+    // Retrieve the display name
+    std::wstring displayName{};
+
+    wil::unique_prop_variant propVariantDisplayName;
+    // Do not throw in case of failure. We can use the filepath approach below as fallback to set a DisplayName.
+    LOG_IF_FAILED(propertyStore->GetValue(PKEY_AppUserModel_RelaunchDisplayNameResource, &propVariantDisplayName));
+
+    displayName = propVariantDisplayName.vt != VT_EMPTY ? propVariantDisplayName.pwszVal : L"";
```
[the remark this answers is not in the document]
hulumane: @danielayala94

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -0,0 +1,395 @@
...
+        HRESULT status = RegGetValueW(
+            hKey.get(),
+            nullptr /* lpValue */,
+            L"NotificationGUID",
+            RRF_RT_REG_SZ,
+            nullptr /* pdwType */,
+            &registeredGuidBuffer,
+            &bufferLength);
+
+        if (status == ERROR_FILE_NOT_FOUND)
```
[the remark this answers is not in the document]
hulumane: Yeah another fix coming up for all these registry bugs with HRESULT

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -0,0 +1,395 @@
...
+}
+
+std::wstring Microsoft::Windows::AppNotifications::Helpers::RegisterComActivatorGuidAndAssets()
+{
+    std::wstring registeredGuid;
+    HRESULT status = GetActivatorGuid(registeredGuid);
+
+    THROW_HR_IF(status, FAILED(status) && status != HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND));
+
+    if (status == ERROR_FILE_NOT_FOUND)
```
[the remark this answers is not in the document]
hulumane: yep

dev/AppNotifications/NotificationProgressData.h
hulumane: @sharath2727 or @danielayala94

dev/AppNotifications/AppNotificationUtility.cpp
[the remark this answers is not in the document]
pmpurifoy: @danielayala94 Do you remember why we used our own compared to this one?

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -0,0 +1,358 @@
...
+        static auto appNotificationManager{winrt::make<AppNotificationManager>()};
+        return appNotificationManager;
+    }
+
+    void AppNotificationManager::Register()
+    {
+        std::wstring storedComActivatorString;
+        if (!PushNotificationHelpers::IsPackagedAppScenario())
+        {
+            std::wstring toastAppId{ RetrieveNotificationAppId() };
```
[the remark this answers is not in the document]
hulumane: I'll fix this in my other change coming up

dev/AppNotifications/AppNotificationUtility.cpp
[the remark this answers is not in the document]
loneursid: This is code that was copied from the FrameworkUDK. In the UDK, NotificationProperties' ctor takes a PCWSTR and internally converts it to utf8 but, the corresponding Get() operation return a utf8 string. It's dangerous to have to seprate places where the conversion code resides. It's bound to diverge in the future.

This better fix for that would be fixing the API at the UDK level to do the conversion internally before returning the value. The problem is the turnaround time with anything touching the UDk. 

Maybe we should take this a s a bug and fix it for preview 2?

dev/AppNotifications/NotificationProgressData.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+    return S_OK;
+}
+CATCH_RETURN()
+
+STDMETHODIMP NotificationProgressData::get_Status(_Out_ HSTRING* value) noexcept try
+{
+    auto lock{ m_lock.lock_shared() };
+
+    *value = Microsoft::Windows::AppNotifications::Helpers::safe_make_unique_hstring(m_progressData.Status().c_str()).release();
```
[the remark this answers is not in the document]
loneursid: Since there is nothing fundamentally wrong with the code (no bug or security risk, etc.), should we take things like these as bugs to be addressed after the merge? My concern is that moving this to a new file, means that we'll have to add at least 1 additional file (if it involves only modifying a .h) to this already very large PR.

dev/AppNotifications/AppNotificationManager.h
```diff
@@ -0,0 +1,75 @@
...
+        wil::srwlock m_lock;
+        winrt::event<NotificationActivationEventHandler> m_notificationHandlers;
+        bool m_firstNotificationReceived{ false };
+    };
+
+    struct AppNotificationManagerFactory : winrt::implements<AppNotificationManagerFactory, IClassFactory>
+    {
+        STDMETHODIMP CreateInstance(_In_opt_ IUnknown* aggregateInterface, _In_ REFIID interfaceId, _Outptr_ VOID** object) noexcept final try
+        {
+            RETURN_HR_IF(CLASS_E_NOAGGREGATION, aggregateInterface != nullptr);
```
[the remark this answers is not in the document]
pmpurifoy: https://github.com/microsoft/WindowsAppSDK/pull/1875#discussion_r779132171

"Callers that pass in a aggregateInterface are expecting an aggregate object (providing access to various inner and outer interfaces/methods). I think it's important you keep this in place to fulfill the contract, indicating to callers that aggregation is not supported."

See also: https://docs.microsoft.com/en-us/windows/win32/api/unknwn/nf-unknwn-iclassfactory-createinstance

@riverar

dev/AppNotifications/NotificationProgressData.cpp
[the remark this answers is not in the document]
robertzhou-wpn: Looks to me that we can address this as a bug fix and get it done after this PR.

dev/AppNotifications/AppNotificationUtility.cpp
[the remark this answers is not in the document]
danielayala94: Agree that we should use common APIs if already existing. But I also agree that we can take this as a bug fix.

dev/AppNotifications/NotificationProgressData.h
danielayala94: Which new types? I guess the question here is "why this is not implemented in the winrt style like the SDK types?"

This is an ABI interface, declared in the FrameworkUdk. There is no projection (winmd) giving us winrt-styled functions. @sharath2727 actually looked into it and there was no straightforward solution (if any) to generate a winmd for these UDK interfaces. We decided to go this way instead.

test/TestApps/PushNotificationsDemoApp/main.cpp
```diff
@@ -147,15 +143,15 @@ int main()
...
-    if (PushNotificationManager::IsActivatorSupported(PushNotificationRegistrationActivators::ComActivator))
-    {
-        // Don't unregister PushTrigger because we still want to receive push notifications from background infrastructure.
-        PushNotificationManager::UnregisterActivator(PushNotificationRegistrationActivators::ComActivator);
-    }
+    PushNotificationManager::Default().Unregister();
 
+    AppNotificationManager::Default().Unregister();
+
+    AppNotificationManager::Default().Unregister();
```
robertzhou-wpn, line 150: I don't think we need to unregister it twice

dev/AppNotifications/NotificationProgressData.cpp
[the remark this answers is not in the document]
loneursid: New issue created to track the work: https://github.com/microsoft/WindowsAppSDK/issues/2207

dev/AppNotifications/AppNotificationUtility.cpp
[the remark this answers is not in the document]
loneursid: Find the bug here: https://github.com/microsoft/WindowsAppSDK/issues/2208

test/TestApps/PushNotificationsDemoApp/main.cpp
danielayala94: Maybe they meant UnregisterAll in the second call? But yeah, a second call isn't needed.

dev/AppNotifications/NotificationProperties.cpp
hulumane: @danielayala94

test/TestApps/PushNotificationsDemoApp/main.cpp
hulumane: Nah. That would hose notifications completely. I'm sure that wasn't the intention :) @pmpurifoy

hulumane: More likely a typo here. This actually would end up throwing the second time around.

dev/AppNotifications/AppNotificationManager.cpp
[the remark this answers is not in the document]
hulumane: Fixed with the registry changes yesterday!

dev/AppNotifications/AppNotificationUtility.cpp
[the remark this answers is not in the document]
danielayala94: Resolving since we have a bug now. If we need further discussion on how to approach the bug, we can discuss it once it's the time to fix it.

dev/AppNotifications/NotificationProgressData.cpp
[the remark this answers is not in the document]
danielayala94: Resolving since we have a bug to fix it after code complete.

dev/AppNotifications/NotificationProperties.cpp
danielayala94: Will look into it.

dev/AppNotifications/AppNotification.cpp
```diff
@@ -0,0 +1,102 @@
...
+        xmlDocument.LoadXml(payload);
+        m_payload = payload;
+    }
+
+    hstring AppNotification::Tag()
+    {
+        return m_tag;
+    }
+
+    void AppNotification::Tag(hstring const& tag)
```
jonwis, line 23: >AppNotification [](http://example.com/codeflow?start=9&length=15)

This type is marked as "both" meaning "agile" but there's no read/write locks anywhere in this type. Should there be?

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -0,0 +1,358 @@
...
+    {
+        std::wstring storedComActivatorString;
+        if (!PushNotificationHelpers::IsPackagedAppScenario())
+        {
+            std::wstring notificationId{ RetrieveNotificationAppId() };
+            if (!AppModel::Identity::IsPackagedProcess())
+            {
+                THROW_IF_FAILED(PushNotifications_RegisterFullTrustApplication(notificationId.c_str(), GUID_NULL));
+
+                storedComActivatorString = RegisterComActivatorGuidAndAssets();
```
jonwis, line 66: >RegisterComActivatorGuidAndAssets [](http://example.com/codeflow?start=43&length=33)

Maybe this method should just return the proper format?

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -0,0 +1,358 @@
...
+            wil::unique_cotaskmem_string processName;
+            THROW_IF_FAILED(GetCurrentProcessPath(processName));
+            auto notificationPlatform{ PushNotificationHelpers::GetNotificationPlatform() };
+            THROW_IF_FAILED(notificationPlatform->AddToastRegistrationMapping(processName.get(), notificationId.c_str()));
+        }
+
+        winrt::guid registeredClsid{ GUID_NULL };
+        if (AppModel::Identity::IsPackagedProcess())
+        {
+            THROW_IF_FAILED(PushNotificationHelpers::GetComRegistrationFromRegistry(expectedAppServerArgs.data(), registeredClsid));
```
jonwis, line 80: >GetComRegistrationFromRegistry [](http://example.com/codeflow?start=53&length=30)

This method could return the `winrt::guid` rather than emitting it as an out-parameter.

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -0,0 +1,358 @@
...
+
+    void AppNotificationManager::UnregisterAll()
+    {
+        Unregister();
+
+        // Remove any Registrations from the Long Running Process that are necessary for Cloud toasts
+        if (!PushNotificationHelpers::IsPackagedAppScenario())
+        {
+            wil::unique_cotaskmem_string processName;
+            THROW_IF_FAILED(GetCurrentProcessPath(processName));
```
jonwis, line 116: Nit: maybe add a future WIL helper for `wil::GetCurrentProcessPath()` that returns this value?

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -0,0 +1,358 @@
...
+        [[maybe_unused]] ULONG dataCount) noexcept try
+    {
+
+        winrt::IMap<winrt::hstring, winrt::hstring> userInput{ winrt::single_threaded_map<winrt::hstring, winrt::hstring>() };
+        for (unsigned long i = 0; i < dataCount; i++)
+        {
+            userInput.Insert(data[i].Key, data[i].Value);
+        }
+
+        winrt::AppNotificationActivatedEventArgs activatedEventArgs = winrt::make<winrt::Microsoft::Windows::AppNotifications::implementation::AppNotificationActivatedEventArgs>(invokedArgs, userInput);
```
jonwis, line 159: >winrt::Microsoft::Windows::AppNotifications::implementation:: [](http://example.com/codeflow?start=82&length=61)

Within the namespace, you can say `implementation::AppNotificationActivatedEventArgs` to get the right type.

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -0,0 +1,358 @@
...
+}
+
+using namespace Microsoft::Windows::AppNotifications::Helpers;
+
+namespace winrt::Microsoft::Windows::AppNotifications::implementation
+{
+    winrt::Microsoft::Windows::AppNotifications::AppNotificationManager AppNotificationManager::Default()
+    {
+        static auto appNotificationManager{winrt::make<AppNotificationManager>()};
+        return appNotificationManager;
```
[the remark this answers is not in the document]
pmpurifoy: This PR addresses this issue:  https://github.com/microsoft/WindowsAppSDK/pull/2187

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -0,0 +1,358 @@
...
+        DWORD notificationId = 0;
+        THROW_IF_FAILED(ToastNotifications_PostToast(notificationAppId.c_str(), notificationProperties.get(), notificationTransientProperties.get(), &notificationId));
+
+        THROW_HR_IF(E_UNEXPECTED, notificationId == 0);
+
+        winrt::Microsoft::Windows::AppNotifications::implementation::AppNotification* notificationImpl = get_self< winrt::Microsoft::Windows::AppNotifications::implementation::AppNotification>(notification);
+        notificationImpl->SetNotificationId(notificationId);
+    }
+
+    winrt::Windows::Foundation::IAsyncOperation<winrt::Microsoft::Windows::AppNotifications::AppNotificationProgressResult> AppNotificationManager::UpdateAsync(winrt::Microsoft::Windows::AppNotifications::AppNotificationProgressData const data, hstring const tag, hstring const group)
```
jonwis, line 234: >UpdateAsync [](http://example.com/codeflow?start=148&length=11)

Must fix: This method does not "co_await resume_backgound" and thus executes synchronously with the caller. Maybe line 240 should have:

```
auto strong = get_strong();
co_await resume_background();
```

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -0,0 +1,358 @@
...
+        DWORD appNotificationSetting{ 0 };
+        ToastNotifications_QuerySettings(appId.c_str(), &appNotificationSetting);
+        return static_cast<winrt::Microsoft::Windows::AppNotifications::AppNotificationSetting>(appNotificationSetting);
+    }
+
+    winrt::Windows::Foundation::IAsyncAction AppNotificationManager::RemoveByIdAsync(uint32_t notificationId)
+    {
+        THROW_HR_IF(E_INVALIDARG, notificationId == 0);
+
+        co_await winrt::resume_background();
```
jonwis, line 276: >co_await [](http://example.com/codeflow?start=8&length=8)

Must fix: Anywhere a method does `co_await` and then uses methods/data on itself it must get a reference to itself, like `auto strong = get_strong()` - otherwise this is a race condition waiting to happen.

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -0,0 +1,358 @@
...
+
+        std::wstring appId{ RetrieveNotificationAppId() };
+        THROW_IF_FAILED(ToastNotifications_RemoveToast(appId.c_str(), notificationId));
+    }
+
+    winrt::Windows::Foundation::IAsyncAction AppNotificationManager::RemoveByTagAsync(hstring const tag)
+    {
+        THROW_HR_IF(E_INVALIDARG, tag == winrt::hstring(L""));
+		
+        std::wstring _tag{ tag.c_str() };
```
jonwis, line 286: >std::wstring _tag{ tag.c_str() }; [](http://example.com/codeflow?start=8&length=33)

No need for this - you can say `.c_str()` on an `hstring` just fine.

dev/AppNotifications/AppNotificationManager.h
```diff
@@ -0,0 +1,75 @@
...
+        winrt::Microsoft::Windows::AppNotifications::AppNotificationSetting Setting();
+        winrt::Windows::Foundation::IAsyncAction RemoveByIdAsync(uint32_t notificationId);
+        winrt::Windows::Foundation::IAsyncAction RemoveByTagAsync(hstring const tag);
+        winrt::Windows::Foundation::IAsyncAction RemoveByTagAndGroupAsync(hstring const tag, hstring const group);
+        winrt::Windows::Foundation::IAsyncAction RemoveByGroupAsync(hstring const group);
+        winrt::Windows::Foundation::IAsyncAction RemoveAllAsync();
+        winrt::Windows::Foundation::IAsyncOperation<winrt::Windows::Foundation::Collections::IVector<winrt::Microsoft::Windows::AppNotifications::AppNotification>> GetAllAsync();
+
+        // INotificationActivationCallback
+        HRESULT __stdcall Activate(
```
jonwis, line 34: >HRESULT __stdcall  [](http://example.com/codeflow?start=8&length=18)

Should this use IFACEMETHODIMP instead?

dev/AppNotifications/AppNotificationManager.cpp
pmpurifoy: Could we also use wil::winrt_module_reference?

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -0,0 +1,421 @@
...
+    }
+
+    auto iteratorForFileExtension = iconFilePath.find_first_of(L".");
+    THROW_HR_IF_MSG(E_UNEXPECTED, iteratorForFileExtension == std::wstring::npos, "You must provide a valid filepath as the app icon.");
+
+    std::wstring iconFileExtension = iconFilePath.substr(iteratorForFileExtension);
+    THROW_HR_IF_MSG(E_UNEXPECTED, iconFileExtension != L".ico" && iconFileExtension != L".png",
+        "You must provide a supported file extension as the icon (.ico or .png).");
+
+    RegisterValue(hKey, L"DisplayName", reinterpret_cast<const BYTE*>(displayName.c_str()), REG_EXPAND_SZ, displayName.size() * sizeof(wchar_t));
```
jonwis, line 293: >RegisterValue [](http://example.com/codeflow?start=4&length=13)

Future: The signature of this method seems strange. I would have expected:

RegisterValue(HKEY key, const wchar_t* name, std::wstring_view value, bool expandString);

... and then it's up to RegisterValue to compute the lengths of the incoming thing.

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -0,0 +1,421 @@
...
+            nullptr /* lpClass */,
+            REG_OPTION_NON_VOLATILE,
+            KEY_ALL_ACCESS,
+            nullptr /* lpSecurityAttributes */,
+            &hKey,
+            nullptr /* lpdwDisposition */));
+
+        WCHAR registeredGuidBuffer[GUID_LENGTH];
+        DWORD bufferLength = sizeof(registeredGuidBuffer);
+        auto status = RegGetValueW(
```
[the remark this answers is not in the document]
pmpurifoy: Should we write a helper that omits the nullptr varibles?

dev/AppNotifications/NotificationProgressData.cpp
```diff
@@ -0,0 +1,76 @@
...
+    *value = Microsoft::Windows::AppNotifications::Helpers::safe_make_unique_hstring(m_progressData.Title().c_str()).release();
+
+    return S_OK;
+}
+CATCH_RETURN()
+
+STDMETHODIMP NotificationProgressData::get_Value(_Out_ double* value) noexcept
+{
+    auto lock{ m_lock.lock_shared() };
+
```
jonwis, line 51: Nit - are these locks required? There are no "update" methods here, only read. The data is set in the constructor and then effectively read-only throughout.

dev/PushNotifications/PushBackgroundTaskInstance.h
```diff
@@ -0,0 +1,35 @@
...
+    HRESULT __stdcall CreateInstance(_In_opt_ IUnknown* aggregateInterface, _In_ REFIID interfaceId, _Outptr_ VOID** object) noexcept final
+    {
+        RETURN_HR_IF(CLASS_E_NOAGGREGATION, aggregateInterface != nullptr);
+        return winrt::make<PushBackgroundTaskInstance>().as(interfaceId, object);
+    }
+
+    HRESULT __stdcall LockServer(BOOL) noexcept final
+    {
+        return S_OK;
+    }
```
jonwis, line 34: Should this be like the other LockServer implementation in the AppNotifications directory?

dev/AppNotifications/AppNotificationUtility.cpp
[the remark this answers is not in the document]
danielayala94: Addressing

dev/AppNotifications/AppNotificationUtility.cpp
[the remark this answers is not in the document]
danielayala94: Addressing

dev/PushNotifications/PushNotifications.idl
```diff
@@ -112,24 +74,28 @@ namespace Microsoft.Windows.PushNotifications
...
+        // Registers an application for Notifications
+        // For Unpackaged applications, registers support for protocol activations
+        // For packaged applications with Background Infrastructure support, registers Push Trigger with the Background Infrastructure
+        // For packaged applications without Background Infrastructure support, registers a COM Service
+        void Register();
 
-        // Unregister all activators registered for the application
-        static void UnregisterAllActivators();
+        // Packaged Apps: Force COM to launch a new process
+//      // Unpackaged apps: Remove the Foreground Sink to force Long Running Process Singleton to launch a new process instance.
```
jonwis, line 89: >// [](http://example.com/codeflow?start=0&length=2)

Nit: Extra comment

dev/PushNotifications/PushNotificationUtility.h
```diff
@@ -4,9 +4,22 @@
...
+#include "wil/win32_helpers.h"
+#include "PushBackgroundTaskInstance.h"
 
+namespace winrt
+{
+    using namespace Windows::Foundation;
+}
 namespace winrt::Microsoft::Windows::PushNotifications::Helpers
 {
+    const std::wstring c_serverIdCommandString = L"ServerId";
```
jonwis, line 19: >const [](http://example.com/codeflow?start=4&length=5)

Future: Make these constexpr

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.h
```diff
@@ -1,19 +1,29 @@
 ﻿#pragma once
 
 #include <FrameworkUdk/PushNotificationsRT.h>
+#include <FrameworkUdk/ToastNotificationsRT.h>
+#include "ToastRegistrationManager.h"
 
-class NotificationListener : public Microsoft::WRL::RuntimeClass<::ABI::Microsoft::Internal::PushNotifications::INotificationListener>
+class NotificationListener : public Microsoft::WRL::RuntimeClass<::ABI::Microsoft::Internal::PushNotifications::INotificationListener,
+                                                                    ::ABI::Microsoft::Internal::PushNotifications::INotificationListener2>
```
jonwis, line 8: Future: This should be a cppwinrt type; avoid using WRL unless required.

dev/PushNotifications/PushNotificationsLongRunningTask/ToastRegistrationManager.h
```diff
@@ -0,0 +1,24 @@
...
+    void Add(std::wstring const& processName, std::wstring const& toastAppId);
+
+    void Remove(std::wstring const& processName);
+
+    std::wstring GetToastRegistration(std::wstring const& processName);
+
+private:
+    winrt::Windows::Storage::ApplicationDataContainer m_toastStorage{ nullptr };
+
+    std::unordered_map<std::wstring, std::wstring> m_registrationMap = {};
```
jonwis, line 22: >p<std::wstring,  [](http://example.com/codeflow?start=21&length=16)

Are the process names' casing important? map's default hash/comparator is case-sensitive.

jonwis (approved):
Saw a few threading items under the "must fix" that should be addressed. Otherwise seems ready to go.  Thank you for adding all the test coverage!

dev/PushNotifications/PushNotificationsLongRunningTask/ToastRegistrationManager.h
pmpurifoy: Yes, processName is case-sensitive. We assume the processName we get from GetCurrentProcessPath is consistent.

hulumane: Should we always change processName that is returned as part of our Process APIs to lower case so that we don't hit compare issues?

pmpurifoy: I guess my question would be how would the casing change from run-to-run? It's probably better to be defensive and use all lowercase.

dev/AppNotifications/AppNotificationManager.cpp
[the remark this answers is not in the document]
hulumane: We need to use the COM static store and compare the singleton against that. Otherwise I believe we run into cleanup problems in the termination stage. Pauls change addresses that.

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -0,0 +1,358 @@
...
+        }
+        else
+        {
+            THROW_HR(hr);
+        }
+    }
+
+    winrt::Windows::Foundation::IAsyncOperation<winrt::Microsoft::Windows::AppNotifications::AppNotificationProgressResult> AppNotificationManager::UpdateAsync(winrt::Microsoft::Windows::AppNotifications::AppNotificationProgressData const data, hstring const tag)
+    {
+        co_return co_await UpdateAsync(data, tag, L"");
```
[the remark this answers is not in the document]
hulumane: @sharath2727 ?

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.h
pmpurifoy: https://github.com/microsoft/WindowsAppSDK/issues/2221

Going to look into this in the future.

dev/PushNotifications/PushNotifications.idl
pmpurifoy: Fixing.

dev/PushNotifications/PushNotificationUtility.h
pmpurifoy: Fixing.

dev/PushNotifications/PushBackgroundTaskInstance.h
```diff
@@ -0,0 +1,35 @@
...
+    std::wstring m_payload;
+};
+
+struct PushBackgroundTaskInstanceFactory : winrt::implements<PushBackgroundTaskInstanceFactory, IClassFactory>
+{
+    HRESULT __stdcall CreateInstance(_In_opt_ IUnknown* aggregateInterface, _In_ REFIID interfaceId, _Outptr_ VOID** object) noexcept final
+    {
+        RETURN_HR_IF(CLASS_E_NOAGGREGATION, aggregateInterface != nullptr);
+        return winrt::make<PushBackgroundTaskInstance>().as(interfaceId, object);
+    }
```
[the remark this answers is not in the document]
pmpurifoy: Fixing.

dev/AppNotifications/NotificationProgressData.cpp
pmpurifoy: @sharath2727

dev/AppNotifications/AppNotificationUtility.cpp
pmpurifoy: This is good improvement to add :)

Filing issue: https://github.com/microsoft/WindowsAppSDK/issues/2222

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -0,0 +1,421 @@
...
+{
+    if (AppModel::Identity::IsPackagedProcess())
+    {
+        wchar_t appUserModelId[APPLICATION_USER_MODEL_ID_MAX_LENGTH] = {};
+        UINT32 appUserModelIdSize{ APPLICATION_USER_MODEL_ID_MAX_LENGTH };
+
+        THROW_IF_FAILED(GetCurrentApplicationUserModelId(&appUserModelIdSize, appUserModelId));
+        return appUserModelId;
+    }
+
```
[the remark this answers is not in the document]
pmpurifoy: Fixing.

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -0,0 +1,421 @@
...
+        if (status == ERROR_FILE_NOT_FOUND)
+        {
+            GUID newNotificationGuid;
+            THROW_IF_FAILED(CoCreateGuid(&newNotificationGuid));
+
+            wil::unique_cotaskmem_string newNotificationGuidString;
+            THROW_IF_FAILED(StringFromCLSID(newNotificationGuid, &newNotificationGuidString));
+
+            std::wstring guidWideStr{ newNotificationGuidString.get() };
```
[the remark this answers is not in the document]
pmpurifoy: Fixing.

dev/AppNotifications/NotificationProgressData.cpp
pmpurifoy: As a principle, we are locking all read/write operations. But in this case since it's a one-time write for all member variables, we don't need to lock this class.

dev/AppNotifications/AppNotificationUtility.cpp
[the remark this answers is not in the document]
pmpurifoy: Will fix as a part of this issue: https://github.com/microsoft/WindowsAppSDK/issues/2222

dev/AppNotifications/AppNotificationManager.h
pmpurifoy: Fixing

dev/AppNotifications/NotificationProgressData.cpp
pmpurifoy: Fixing.

dev/AppNotifications/AppNotification.cpp
pmpurifoy: This needs locks.

dev/AppNotifications/AppNotificationManager.cpp
pmpurifoy: Who should we @ for this?

pmpurifoy: https://github.com/microsoft/WindowsAppSDK/issues/2223

dev/AppNotifications/AppNotificationManager.cpp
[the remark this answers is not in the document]
sharath2727: I have followed suggestions based on raymond's blog: https://devblogs.microsoft.com/oldnewthing/?p=106160

dev/AppNotifications/AppNotificationManager.cpp
sharath2727: @loneursid

dev/PushNotifications/PushNotificationsLongRunningTask/ToastRegistrationManager.h
pmpurifoy: Fixed but should look into adding comparator to map.

Issue to track: https://github.com/microsoft/WindowsAppSDK/issues/2225

dev/AppNotifications/AppNotification.h
```diff
@@ -0,0 +1,63 @@
...
+
+        winrt::Microsoft::Windows::AppNotifications::AppNotificationProgressData m_progressData{ nullptr };
+
+        winrt::Windows::Foundation::DateTime m_expirationTime{};
+
+        bool m_expiresOnReboot{ false };
+
+        winrt::Microsoft::Windows::AppNotifications::AppNotificationPriority m_priority
+        { winrt::Microsoft::Windows::AppNotifications::AppNotificationPriority::Default };
+
```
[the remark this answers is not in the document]
pmpurifoy: Fixing.

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -265,27 +265,46 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
     std::wstring iconFilePath{};
 
     wil::unique_prop_variant propVariantIcon;
-    // Throw in case of failure, since Icon is mandatory and we don't have a fallback!
-    THROW_IF_FAILED(propertyStore->GetValue(PKEY_AppUserModel_RelaunchIconResource, &propVariantIcon));
-    THROW_HR_IF_MSG(E_UNEXPECTED, propVariantIcon.vt == VT_EMPTY, "You must specify an app icon before calling Register().");
+    // In case there is no icon asset, do not throw. Set a default icon instead.
+    LOG_IF_FAILED(propertyStore->GetValue(PKEY_AppUserModel_RelaunchIconResource, &propVariantIcon));
```
hulumane, line 269: I'm not sure I understand this. Get nervous when i see code that swallows errors. Why are we not throwing? @danielayala94

hulumane: Same concern for DisplayNameResource above.

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -265,27 +265,46 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
...
-    // since .ico files can have multiple icons in the same file.
-    // NotificationController doesn't seem to support such format, so let it take the first icon by default.
-    auto iteratorForCommaDelimiter = iconFilePath.find_first_of(L",");
-    if (iteratorForCommaDelimiter != std::wstring::npos) // It may or may not have an index, which is fine.
+    if (propVariantIcon.vt == VT_LPWSTR && propVariantIcon.pwszVal != nullptr)
     {
-        iconFilePath.erase(iteratorForCommaDelimiter);
-    }
+        iconFilePath = propVariantIcon.pwszVal;
+
```
hulumane, line 274: Check for empty string
THROW_HR_IF(E_FAIL, iconFilePath.empty());

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -265,27 +265,46 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
...
+        // NotificationController doesn't seem to support such format, so let it take the first icon by default.
+        auto iteratorForCommaDelimiter = iconFilePath.find_first_of(L",");
+        if (iteratorForCommaDelimiter != std::wstring::npos) // It may or may not have an index, which is fine.
+        {
+            iconFilePath.erase(iteratorForCommaDelimiter);
+        }
+
+        auto iteratorForLastBackSlash = iconFilePath.find_last_of(L"\\");
+        auto iteratorForForwardSlash = iconFilePath.find_last_of(L"/");
+        THROW_HR_IF_MSG(ERROR_BAD_PATHNAME, iteratorForLastBackSlash != std::wstring::npos && iteratorForForwardSlash != std::wstring::npos,
```
hulumane, line 286: Break this up please to improve readablility. For example, have the first check after 
auto iteratorForLastBackSlash = iconFilePath.find_last_of(L"\\");
And have the second check after:
auto iteratorForForwardSlash = iconFilePath.find_last_of(L"/");

hulumane: Honestly, I'm not sure I understand what this check is about. What path are you considering as invalid and planning on throwing for? Add a comment please.

hulumane: Why can't we simply prevent forward slash and keep it consistent instead of this complex check.

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -265,27 +265,46 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
...
+            "You must provide a valid icon filepath (do not combine forward and back slashes).");
+        auto iteratorForLastSlash = min(iteratorForLastBackSlash, iteratorForForwardSlash);
```
hulumane, line 288: More complexity that can simply be avoided if we say that we support back slash only in the path. That seems pretty standard and reasonable to me

dev/AppNotifications/AppNotificationUtility.cpp
hulumane: To me c:\test\test.png is valid. It only supports backslash. So fail only if forward slash is part of the path.

hulumane, line 286: What if the path is not a local path?

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -265,27 +265,46 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
...
+
+        std::wstring fileName{ iconFilePath };
 
-    auto iteratorForFileExtension = iconFilePath.find_first_of(L".");
-    THROW_HR_IF_MSG(E_UNEXPECTED, iteratorForFileExtension == std::wstring::npos, "You must provide a valid filepath as the app icon.");
+        if (iteratorForLastSlash != std::wstring::npos)
```
hulumane, line 292: Not sure I understand this. We should chat offline on how to simplify everything here. Its way too complex IMHO

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -265,27 +265,46 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
...
+        auto iteratorForFileExtension = fileName.find_first_of(L".");
+        THROW_HR_IF_MSG(E_UNEXPECTED, iteratorForFileExtension == std::wstring::npos, "You must provide a valid filepath as the app icon.");
+
+        std::wstring iconFileExtension = fileName.substr(iteratorForFileExtension);
+        THROW_HR_IF_MSG(E_UNEXPECTED, iconFileExtension != L".ico" && iconFileExtension != L".png",
+            "You must provide a supported file extension as the icon (.ico or .png).");
+    }
+    else
+    {
+        iconFilePath = std::wstring(LR"(ms-resource://Windows.UI.ShellCommon/Files/Images/DefaultSystemNotification.png)");
```
hulumane, line 306: Isn't ms-resource only valid for packaged apps?

dev/AppNotifications/AppNotificationUtility.cpp
danielayala94: Agree that we have room to simplify. Let's discuss at standup all your questions.

dev/PushNotifications/PushNotificationManager.h
```diff
@@ -28,10 +29,14 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        IFACEMETHODIMP InvokeAll(_In_ ULONG length, _In_ byte* payload, _Out_ BOOL* foregroundHandled) noexcept;
 
-        // INotificationHandler
-        HRESULT __stdcall OnRawNotificationReceived(unsigned int payloadLength, _In_ byte* payload, _In_ HSTRING /*correlationVector */) noexcept;
+        // INotificationListener
+        IFACEMETHODIMP OnRawNotificationReceived(unsigned int payloadLength, _In_ byte* payload, _In_ HSTRING /*correlationVector */) noexcept;
+
+        // INotificationListener2
+        IFACEMETHODIMP OnToastNotificationReceived(ABI::Microsoft::Internal::ToastNotifications::INotificationProperties* notificationProperties,
+            ABI::Microsoft::Internal::ToastNotifications::INotificationTransientProperties*) noexcept;
```
sharath2727, line 39: alignment?

dev/AppNotifications/AppNotificationUtility.cpp
danielayala94: After discussing offline, this isn't the right time to address this. I'll proceed to revert.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -629,4 +635,14 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
         return S_OK;
     }
     CATCH_RETURN();
+
+    IFACEMETHODIMP PushNotificationManager::OnToastNotificationReceived(
+        ToastNotifications::INotificationProperties* notificationProperties,
+        ToastNotifications::INotificationTransientProperties* notificationTransientProperties) noexcept try
+    {
+        DWORD notificationId{ 0 };
+        ToastNotifications_PostToast(PushNotificationHelpers::GetAppUserModelId().get(), notificationProperties, notificationTransientProperties, &notificationId);
```
hulumane, line 644: Is this call only applicable for Packaged Apps? It seems to be the case. If so, I would recommend having a check in the beginning of the function:
THROW_HR_IF(E_UNEXPECTED, !IsPackaged());

hulumane: Actually even better:
THROW_HR_IF(E_UNEXPECTED, !IsPackagedAppScenario());

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -245,8 +245,10 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
     wil::unique_prop_variant propVariantDisplayName;
     // Do not throw in case of failure. We can use the filepath approach below as fallback to set a DisplayName.
     LOG_IF_FAILED(propertyStore->GetValue(PKEY_AppUserModel_RelaunchDisplayNameResource, &propVariantDisplayName));
-
```
hulumane: @danielayala94  Should the above call be a THROW_IF_FAILED?

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -245,8 +245,10 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
...
-    displayName = propVariantDisplayName.vt != VT_EMPTY ? propVariantDisplayName.pwszVal : L"";
+    if (propVariantDisplayName.vt == VT_LPWSTR && propVariantDisplayName.pwszVal != nullptr)
+    {
+        displayName = propVariantDisplayName.pwszVal;
```
hulumane, line 250: After this, check to make sure displayName is not empty
THROW_HR_IF(E_UNEXPECTED, displayName.isEmpty());

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -245,8 +245,10 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
...
+    }
 
     if (displayName.length() == 0)
```
hulumane, line 253: use .empty() instead.

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -66,10 +68,8 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
                 storedComActivatorGuid = RegisterComActivatorGuidAndAssets();
             }
 
-            wil::unique_cotaskmem_string processName;
-            THROW_IF_FAILED(GetCurrentProcessPath(processName));
             auto notificationPlatform{ PushNotificationHelpers::GetNotificationPlatform() };
-            THROW_IF_FAILED(notificationPlatform->AddToastRegistrationMapping(processName.get(), notificationId.c_str()));
+            THROW_IF_FAILED(notificationPlatform->AddToastRegistrationMapping(m_processName.data(), notificationId.c_str()));
```
hulumane, line 72: m_processName.c_str().

What does .data() do?

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -110,10 +110,8 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
         // Remove any Registrations from the Long Running Process that are necessary for Cloud toasts
         if (!PushNotificationHelpers::IsPackagedAppScenario())
         {
-            wil::unique_cotaskmem_string processName;
-            THROW_IF_FAILED(GetCurrentProcessPath(processName));
             auto notificationPlatform{ PushNotificationHelpers::GetNotificationPlatform() };
-            THROW_IF_FAILED(notificationPlatform->RemoveToastRegistrationMapping(processName.get()));
+            THROW_IF_FAILED(notificationPlatform->RemoveToastRegistrationMapping(m_processName.data()));
```
hulumane, line 114: c_str()?

dev/PushNotifications/externs.h
```diff
@@ -25,7 +25,10 @@ struct ChannelDetails
     winrt::Windows::Foundation::DateTime channelExpiryTime;
 };
 
-inline HRESULT GetCurrentProcessPath(wil::unique_cotaskmem_string& processName)
+inline std::wstring GetCurrentProcessPath()
 {
-    return wil::GetModuleFileNameExW(GetCurrentProcess(), nullptr, processName);
+    std::wstring processPath{};
+    THROW_IF_FAILED(wil::GetModuleFileNameExW(GetCurrentProcess(), nullptr, processPath));
```
hulumane, line 31: How is this even working? What is the third argument for this api? a wstring?? I would be very surprised if so.

hulumane: I think it needs to be wil::unique_cotaskmem_string. Otherwise it may not allocate\deallocate memory correctly

pmpurifoy: 3rd argument is string_type.

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -254,10 +250,7 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
...
     {
-        wil::unique_cotaskmem_string processName;
-        THROW_IF_FAILED(GetCurrentProcessPath(processName));
-
-        displayName = processName.get();
+        THROW_IF_FAILED(wil::GetModuleFileNameExW(GetCurrentProcess(), nullptr, displayName));
```
hulumane, line 255: @danielayala94 Should we guarantee that the displayname here starts with uppercase?

hulumane, line 255: Should this be a wil construct? Not wstring?

dev/AppNotifications/AppNotificationManager.cpp
pmpurifoy: data gets a byte pointer by c_str is better. Fixing.

dev/AppNotifications/AppNotificationUtility.cpp
hulumane: For example, it should be Tiktok. Not tiktok.

pmpurifoy: This returns the original path name. So it will be uppercase (original-casing).

pmpurifoy: Looks like GetModuleFileName just takes in a generic string_type. I can change this to wil::unique_cotaskmem_string which is the default string_type used.

danielayala94: While for the demo app this is going to retrieve a good-looking display name, we should consider this as a fix while working on introducing the default icon.

danielayala94: I have created a work item for it -> #2228

dev/AppNotifications/AppNotificationUtility.cpp
danielayala94: Agree, though it's a minor change. I will address it as part of #2228.

dev/AppNotifications/AppNotificationUtility.cpp
danielayala94: Yes. My logic was to prevent throwing if there was no element in the property store. But such assumption is wrong, because it returns S_OK anyway.

I will address this as part of #2228.

dev/PushNotifications/externs.h
pmpurifoy: https://github.com/microsoft/wil/blob/23fca640eae099f9120545483ae7d5d0a98a6fa5/include/wil/win32_helpers.h#L200

maker.release() here returns a std::wstring  so we should be good :)

hulumane (approved):

PR #2196, opened 2022-03-02
Update PushNotifications-spec.md

pmpurifoy:


specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -233,20 +230,10 @@ Process A (Registration of the Push Trigger only):
...
-            winrt::guid("BACCFA91-F1DE-4CA2-xxxx-90BE66934EC6"));
-    }
-    else
-    {
-        return E_NOTIMPL;
-    }
-
     // Registers a Push Trigger with the Background Infra component
-    PushNotificationManager::RegisterActivator(info);
+    PushNotificationManager::Default().Register();
```
brflynn, line 234: nit; blank line below

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -411,58 +389,25 @@ would need to be pre-set to a well-known string that defines Push Triggers.
...
-        static PushNotificationReceivedEventArgs CreateFromBackgroundTaskInstance(Windows.ApplicationModel.Background.IBackgroundTaskInstance backgroundTask);
-
-        // Initialize using the PushNotificationEventArgs from Windows: used specifically for in-memory event handling when app is already in foreground
-        static PushNotificationReceivedEventArgs CreateFromPushNotificationReceivedEventArgs(Windows.Networking.PushNotifications.PushNotificationReceivedEventArgs args);
-
         // The Push payload
         byte[] Payload { get; };
 
-        // Gets a deferral to run under specific modes like low power mode
+        // Gets a deferral to allow push notification processing even if the system goes into low power mode.
```
brflynn, line 399: This is **only** needed for allowing notification processing in low power-mode right? 

If so, let's re-word as "// Gets a deferral to allow notification processing when the system is in low-power mode"

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -472,71 +417,75 @@ namespace Microsoft.Windows.PushNotifications
...
+    [feature(Feature_PushNotifications)]
     struct PushNotificationCreateChannelStatus
     {
-        // The last extended error we failed Channel requests on that caused the inprogress retry status. S_OK if this is just progress status.
-        HRESULT extendedError;
-
         // Either InProgress or InProgressRetry status
         PushNotificationChannelStatus status;
 
+        // The last extended error we failed Channel requests on that caused the inprogress retry status. E_PENDING if this is just progress status.
```
brflynn, line 426: Is this really the "last extended error"? If so maybe rename to lastExtendedError so it's clear that other errors are overwritten. 

Can we also re-word to  something a little more clear such as - "// The last extended error seen when creating a channel request. This error causes InProgressRetry status. If there is no retry and just InProgress status this defaults to E_PENDING"

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -472,71 +417,75 @@ namespace Microsoft.Windows.PushNotifications
...
+    [feature(Feature_PushNotifications)]
     runtimeclass PushNotificationCreateChannelResult
     {
         PushNotificationCreateChannelResult(
             PushNotificationChannel channel,
             HRESULT extendedError,
             PushNotificationChannelStatus status);
 
-        // The Push channel associated with the Result. Null if InProgress or completion failed
+        // The Push channel associated with the Result. Valid only if status is CompletedSuccess.
```
brflynn, line 454: What is the value of this if the status is not CompletedSuccess? I am assuming null? 

If this is null or an object, maybe "A Channel only exists when status is CompletedSuccess"?

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -472,71 +417,75 @@ namespace Microsoft.Windows.PushNotifications
...
-        // 2) If kind = ComActivator is specified, the Project Reunion Background Task component will no longer act as an InProc COM Server
-        // 3) If kind = ProtocolActivator is specified, the application will be unregistered from the long running process that handles activation
-        static void UnregisterActivator(PushNotificationRegistrationActivators activators);
+        // Packaged Apps: Force COM to launch a new process
+        // Unpackaged apps: Remove the Foreground Sink to force Long Running Process Singleton to launch a new process instance.
+        void Unregister();
 
-        // Unregister all activators registered for the application
-        static void UnregisterAllActivators();
+        // Disables Push completely for Packaged and Unpackaged apps. Register() needs to be called again for Push to work
```
brflynn, line 480: // Disables Push Notifications completely for Packaged and Unpackaged apps. Register() needs to be called again afterwards for push notifications to work.

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -472,71 +417,75 @@ namespace Microsoft.Windows.PushNotifications
...
+        void UnregisterAll();
 
         // Request a Push Channel with an encoded RemoteId from WNS. RemoteId is an AAD identifier GUID
-        static Windows.Foundation.IAsyncOperationWithProgress<PushNotificationCreateChannelResult, PushNotificationCreateChannelStatus> CreateChannelAsync(Guid remoteId);
+        Windows.Foundation.IAsyncOperationWithProgress<PushNotificationCreateChannelResult, PushNotificationCreateChannelStatus> CreateChannelAsync(Guid remoteId);
 
-        // Applications will call this to check which flags are supported for RegisterActivator
-        static Boolean IsActivatorSupported(PushNotificationRegistrationActivators activators);
+        // Event handler for Push Notifications
+        // NOTE: You must register your foreground handlers before calling Register() or else it will throw.
```
brflynn, line 487: Is it really only "foreground" handlers here? Background handlers do not have this requirement?

brflynn (changes requested):

specs/PushNotifications/PushNotifications-spec.md
pmpurifoy: GetActivatedEventArgs is the equivalent of background handlers. Handling background activation is a one-time event and to account for synchronization issues we are forcing developers to register handlers first.

specs/PushNotifications/PushNotifications-spec.md
pmpurifoy: It will be nullptr.

specs/PushNotifications/PushNotifications-spec.md
jonwis: Deferrals are not always related to low power. They're a way for the app to say "I will need to continue working even after the event handler that sent me this returns, so please don't consider me done until I complete the deferral."  I'd remove the "into low power mode" text and say:

```c++
// Gets a deferral for when the app needs to continue processing after the event handler returns.
// Complete the deferral when the app has finished handling the notification.
```

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -81,9 +81,14 @@ Link to the official Windows App SDK timeline can be found
 
 The code in Main would follow the pattern below:
 
--   The Registration will take in a CLSID of the COM server as part of the Activator setup. The
-    registration API will simply abstract away the COM details from the developer including
-    registration of the in-process Windows App SDK component.
+-   All apps will call into the registration API.
```
jonwis, line 84: Editorial note - consider a stronger form of statements like this, basically eliminating the word "will":

> Apps start by calling into the registration API

Similarly below:

> For packaged applications, registration with the notification services is handled automatically.

And:

> Use the `CoreApp` activation event system to determine the `ActivationKind` - for `PushNotification`, the activation arguments object is an instance of `PushActivatedEventArgs`.

... no change required for now.

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -95,48 +100,50 @@ The code in Main would follow the pattern below:
...
-    for information on how to build and deploy unpackaged applications with Windows App SDK.
 
 ```cpp
 int main()
 {
-    PushNotificationActivationInfo info {};
-    bool isComActivationSupported { PushNotificationManager::IsActivatorSupported(PushNotificationRegistrationActivators::PushTrigger |
-                                        PushNotificationRegistrationActivators::ComActivator)};
-    if (isComActivationSupported)
+    if (!IsPackagedProcess())
```
jonwis, line 111: Nit on sample: Consider removing this and simply having a one-line:

```c++
// See documentation on how to ensure your app has access to the Windows App SDK
```

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -95,48 +100,50 @@ The code in Main would follow the pattern below:
...
+    // Register Push Event for Foreground
+    winrt::event_token token = pushNotificationManager.PushReceived([](const auto&, PushNotificationReceivedEventArgs const& args)
     {
-        info = PushNotificationActivationInfo(PushNotificationRegistrationActivators::ProtocolActivator);
-    }
+        auto payload = args.Payload();
+
+        // Do stuff to process the raw payload
+        std::string payloadString(payload.begin(), payload.end());
+        std::cout << "Push notification content received from FOREGROUND: " << payloadString << std::endl << std::endl;
```
jonwis, line 128: This "foreground" comment is a little confusing. Consider removing it, and instead have a sample like:

```c++
// App logic to handle the incoming payload string. Maybe pop a toast, maybe
// kick your sync engine to fetch updates, maybe request an update from your
// service. See our full samples for building an end-to-end push notification system
```

specs/PushNotifications/PushNotifications-spec.md
hulumane: This uses the same pattern as AppNotificationManager and should be easy for devs to comprehend since it follows similar patterns.

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -95,48 +100,50 @@ The code in Main would follow the pattern below:
...
-    else
+
+    auto pushNotificationManager { PushNotificationManager::Default() };
+
+    // Register Push Event for Foreground
+    winrt::event_token token = pushNotificationManager.PushReceived([](const auto&, PushNotificationReceivedEventArgs const& args)
     {
-        info = PushNotificationActivationInfo(PushNotificationRegistrationActivators::ProtocolActivator);
-    }
+        auto payload = args.Payload();
```
jonwis, line 124: API design note: In samples, unless the type is obvious, spell it out:

```c++
std::vector<byte> payload = args.Payload(); // OK
auto moop = MyType{a, b, c}; // OK, "moop" is of MyType
auto boop = foo.GetBar(); // What type is "boop?"
```

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -95,48 +100,50 @@ The code in Main would follow the pattern below:
...
+
+        // Do stuff to process the raw payload
+        std::string payloadString(payload.begin(), payload.end());
+        std::cout << "Push notification content received from FOREGROUND: " << payloadString << std::endl << std::endl;
+        args.Handled(true);
```
hulumane, line 129: Handled has gone away. It should be removed here and in the idl. 

New pattern accounts for a new process instance bringup if the event handler doesnt exist.

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -205,20 +201,21 @@ int main()
...
         if (result.Status() == PushNotificationChannelStatus::CompletedSuccess)
         {
-            result.Channel().PushReceived(pushToken);
+            // Maintain the channel object to retrieve Uri and ExpirationTime
+            auto channel { result.Channel() };
         }
     }
 
-    if(isComActivationSupported)
+    // If packaged: Unregisters foreground sink and the app as the ComServer
```
jonwis, line 212: These are implementation details for us, you can just say:

```
// Unregistering this process makes sure that other incoming pushes launch a new process
// to handle them.
```

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -472,71 +417,75 @@ namespace Microsoft.Windows.PushNotifications
...
+    [feature(Feature_PushNotifications)]
+    runtimeclass PushNotificationManager
     {
-        // Register an activator using an ActivationInfo context and cache the flag for unregister
-        static void RegisterActivator(PushNotificationActivationInfo details);
+        // Gets a Default instance of a PushNotificationManager
+        static PushNotificationManager Default{ get; };
+
+        // Registers an application for Notifications
+        // For Unpackaged applications, registers support for protocol activations
```
jonwis, line 471: I'd leave out these implementation details here. Just say:

```cpp
// Registers an application for activation. After registration completes, incoming
// notifications are delivered to the PushRecieved event.
```

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -472,71 +417,75 @@ namespace Microsoft.Windows.PushNotifications
...
+        // For packaged applications with Background Infrastructure support, registers Push Trigger with the Background Infrastructure
+        // For packaged applications without Background Infrastructure support, registers a COM Service
+        void Register();
 
-        // Unregister any activator given PushNotificationRegistrationActivators
-        // 1) If kind = PushTrigger is specified, the trigger itself will be removed
-        // 2) If kind = ComActivator is specified, the Project Reunion Background Task component will no longer act as an InProc COM Server
-        // 3) If kind = ProtocolActivator is specified, the application will be unregistered from the long running process that handles activation
-        static void UnregisterActivator(PushNotificationRegistrationActivators activators);
+        // Packaged Apps: Force COM to launch a new process
```
jonwis, line 476: I'm sure you don't mean that calling `Unregister` will cause a new foo.exe to be launched just because of unregistration.  Maybe:

```cpp
// Stops the flow of notifications to this process and manager object. Additional incoming
// notifications may cause a new process to be launched to handle them. To stop recieving
// all notifications, use <this method name>.
```

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -472,71 +417,75 @@ namespace Microsoft.Windows.PushNotifications
...
         HRESULT ExtendedError{ get; };
 
-        // The Status of the ChannelComplete operation
+        // The Status of the CreateChannel operation
         PushNotificationChannelStatus Status { get; };
     };
 
-    static runtimeclass PushNotificationManager
+    [feature(Feature_PushNotifications)]
+    runtimeclass PushNotificationManager
```
jonwis, line 465: I wonder if the model should be more like Register/Start/Stop/Unregister, where:

* Register configures all the registration goo, but does not start the flow of notifications yet
* Start begins really listening and handing incoming pushes to the handler
* Stop ends the flow of notifications to this manager instance so the app can shut down
* Unregister removes all channels that were configured and is basically the reverse of Register

No change required, something to think about.

pmpurifoy: https://github.com/microsoft/WindowsAppSDK/issues/2227

pmpurifoy: Tracking this.

specs/PushNotifications/PushNotifications-spec.md
pmpurifoy: Fixing the comment and tracking bug to rename the param. https://github.com/microsoft/WindowsAppSDK/issues/2229

brflynn (approved):

jonwis (approved):

PR #2198, opened 2022-03-02
Added contracts to PowerNotifications.idl

huzaifa-d:
Adding contract to the IDL to facilitate adding new APIs and more!

### DIFF
--- a/dev/PowerNotifications/PowerNotifications.idl
+++ b/dev/PowerNotifications/PowerNotifications.idl
@@ -3,7 +3,11 @@
 
 namespace Microsoft.Windows.System.Power
 {
+    [contractversion(1.1)]
+    apicontract MicrosoftWindowsSystemPower{};
+
     // Enums duplicated from IPowerManagerStatics
+    [contract(MicrosoftWindowsSystemPower, 1.0)]
     enum EnergySaverStatus
     {
         Uninitialized = 0,
@@ -12,6 +16,7 @@ namespace Microsoft.Windows.System.Power
         On
     };
 
+    [contract(MicrosoftWindowsSystemPower, 1.0)]
     enum BatteryStatus
     {
         NotPresent = 0,
@@ -20,6 +25,7 @@ namespace Microsoft.Windows.System.Power
         Charging
     };
 
+    [contract(MicrosoftWindowsSystemPower, 1.0)]
     enum PowerSupplyStatus
     {
         NotPresent = 0,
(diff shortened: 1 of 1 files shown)


brialmsft (approved):
There are some conventions used here that are interesting. My understanding is that @dhoehna has been more involved in formalizing contract versioning conventions for WinAppSdk, so I'll let him verify that this _matches_ the intended conventions. But I do have some thoughts on them if so:
1. Using 1.1 as the next version number instead of 2. If the intention is to support both minor (non-breaking) and major (breaking) changes with this version scheme, it's worth noting that some non-additive breaking changes to COM interfaces are *really* bad. If 2 pieces of code disagree on the shape of an interface, it can cause AVs, which could potentially be a security risk depending on exactly where the mismatch is. So even within breaking changes, you *must* treat existing interfaces as immutable. But other there are other breaking changes you can get away with.
We do have existing tooling that the OS uses to confirm that all changes to the system metadata from version to version are purely additive and non-breaking. I don't think we have any tooling that would currently differentiate between "breaking without interface mutation" and "breaking with interface mutation".
2. The fully namespace-qualified contract name is Microsoft.Windows.System.Power.MicrosoftWindowsSystemPower. Probably not an issue since you wouldn't fully qualify it unless you're trying to refer to the contract from somewhere outside of this file (and I don't think anyone necessarily has reason to do so). But it is a little bit silly looking in that form.

dev/PowerNotifications/PowerNotifications.idl
```diff
@@ -3,7 +3,11 @@
 
 namespace Microsoft.Windows.System.Power
 {
+    [contractversion(1.1)]
```
dhoehna, line 6: No MAJOR.MINOR.  Only 1 number.  Start with 1.  Any changes to the interface will increment the number.

dhoehna (changes requested):
Start with contract version 1.  No MAJOR.MINOR versions.

dev/PowerNotifications/PowerNotifications.idl
jevansaks: Where is that requirement coming from? Is that an API review guideline?

dhoehna: @DrusTheAxe, @BenJKuhn, @DefaultRyan, and I came up with it.

Here is the spec currently for contracts: https://github.com/microsoft/WindowsAppSDK/pull/2110

It isn't an API review guideline.  The version gets messy when upgrading the MAJOR version that includes a patch.  IF you believe that having a MAJOR.MINOR version is better can you please add the comment to the PR I mentioned?

DrusTheAxe (approved):
LGTM

PR #2205, opened 2022-03-03
Rename unique_package_dependency_context

DrusTheAxe:
Rename `unique_package_dependency_context` to `unique_mdd_package_dependency_context`

This is a specialization of WIL's `unique_any` for WinAppSDK's `MDD_PACKAGEDEPENDENCY_CONTEXT`.

This rename is to prevent conflicts with a WIL specialization for Windows' `PACKAGEDEPENDENCY_CONTEXT` (i.e. the OS Dynamic Dependnecy API).

### DIFF
--- a/dev/DynamicDependency/API/wil_msixdynamicdependency.h
+++ b/dev/DynamicDependency/API/wil_msixdynamicdependency.h
@@ -10,7 +10,7 @@ typedef unique_any<DLL_DIRECTORY_COOKIE, decltype(&::RemoveDllDirectory), ::Remo
 
 #if defined(MSIXDYNAMICDEPENDENCY_H) && !defined(__WIL_MSIXDYNAMICDEPENDENCY_H)
 #define __WIL_MSIXDYNAMICDEPENDENCY_H
-typedef unique_any<MDD_PACKAGEDEPENDENCY_CONTEXT, decltype(&::MddRemovePackageDependency), ::MddRemovePackageDependency> unique_package_dependency_context;
+typedef unique_any<MDD_PACKAGEDEPENDENCY_CONTEXT, decltype(&::MddRemovePackageDependency), ::MddRemovePackageDependency> unique_mdd_package_dependency_context;
 #endif // __WIL_MSIXDYNAMICDEPENDENCY_H
 
 #if defined(_APPMODEL_H_) && !defined(__WIL_APPMODEL_H_) && WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP)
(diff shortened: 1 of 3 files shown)


DrusTheAxe:
@jefgen this fails due to

```
Invalid file path 'D:\a\1\s\build\Helix\packages\Microsoft.Internal.WinUI.Helix.0.5.4\scripts\pipeline\ProcessHelixFiles.ps1'. A path to a .ps1 file is required.
```

but I didn't change of that here. Did you change the backend in some way for the other PR but it affects ALL PRs? This (and others?) will fail until that PR's completed?

jefgen:
> Did you change the backend in some way for the other PR but it affects ALL PRs?

Nope, I haven't touched anything. :)

This other one looks like x64 passed, but the x86 part failed.

It looks like there might have been some network flakiness. Looks like the `NuGetToolInstallertimed` timed out, which might explain why the "ProcessHelixFiles.ps1" script is missing (it comes from a NuGet package).
`ERR:connect ETIMEDOUT 152.199.4.184:443`

I'm going to try to re-queue the failed x86 portion.

DrusTheAxe:
> might have been some network flakiness

Not the first time I've seen that. Can we add some magic-resiliency in here, e.g. if we fail due to network flakiness try again for, say, 3 bites at the apple before failing?

Can we determine errors that are infrastructure/network-flakiness vs actual errors?

Not for this PR, just as a general infrastructure improvement follow up

jefgen:
Yeah, that's a fair point.
Though I'm not sure how tricky it might be to try and re-run all of the test logic in something fails.
We might be able to retry parts of it though. I'll file a [bug](https://microsoft.visualstudio.com/OS/_workitems/edit/38392024) so there is something tracking for now. (It can always be dupe'd to an existing task if there is one, for example).

dev/DynamicDependency/API/wil_msixdynamicdependency.h
```diff
@@ -10,7 +10,7 @@ typedef unique_any<DLL_DIRECTORY_COOKIE, decltype(&::RemoveDllDirectory), ::Remo
 
 #if defined(MSIXDYNAMICDEPENDENCY_H) && !defined(__WIL_MSIXDYNAMICDEPENDENCY_H)
 #define __WIL_MSIXDYNAMICDEPENDENCY_H
-typedef unique_any<MDD_PACKAGEDEPENDENCY_CONTEXT, decltype(&::MddRemovePackageDependency), ::MddRemovePackageDependency> unique_package_dependency_context;
```
jonwis, line 13: Is this new as of 1.0? If so the rename is OK. If not it's a little dangerous since it's a source-breaking change, but I bet nobody uses it directly so the risk is very low.

jonwis (approved):

sachintaMSFT (approved):

dev/DynamicDependency/API/wil_msixdynamicdependency.h
DrusTheAxe: This was new in 1.0

Seemed low enough risk and easily fixed if developers encounter the change and mention in release notes seemed reasonable. Has zero impact on binary or behavior compat.

PR #2224, opened 2022-03-08
Removing [experimental] tags for preview1

loneursid:


loneursid:
@DrusTheAxe - I've removed the [experimental] tags from the idl. Is there anything else I'm supposed to do to ensure our APIs will show up in the preview1 build?

I've noticed that in our PushNotifications idl, we have these tags [feature(Feature_PushNotifications)], instead of [experimental], are we supposed to remove these too? Or, are we supposed to add similar tagsto the toast notifications APIs? Along with the checks in the implementation that throws unless the feature is enabled?

DrusTheAxe:
> Is there anything else I'm supposed to do to ensure our APIs will show up in the preview1 build?

i.e. remove the experimental-ness?

`dev\Common\TerminalVelocityFeatures-PushNotifications.xml`

```
        <alwaysDisabledChannelTokens>
            <channelToken>Preview</channelToken>
            <channelToken>Stable</channelToken>
        </alwaysDisabledChannelTokens>
```

Delete that and run the generator to update the .h.

I'm not sure about the `[feature(Feature_PushNotifications)]` attribute. @MikeHillberg should know more

DrusTheAxe (approved):
LGTM as far as it goes. Also needs to update the TerminalVelocity files

Not sure about the 'feature' attribute. See comment

loneursid:
@DrusTheAxe  - I removed the <alwaysDisabledChannelTokens> entries from dev\Common\TerminalVelocityFeatures-PushNotifications.xml and ran the terminal velocity generator with "Stable" instead of "Experimental". All it did was update the comments in the .h.

 @MikeHillberg - Should we have something similar to [feature(Feature_PushNotifications)] in the idl for AppNotifications? (AppNotifications are new APIs that we are introducing in v1.1.)

dev/AppNotifications/AppNotifications.idl
```diff
@@ -6,7 +6,6 @@ import "..\AppLifecycle\AppLifecycle.idl";
 namespace Microsoft.Windows.AppNotifications
 {
     // Event args for the Notification Activation
-    [experimental]
     runtimeclass AppNotificationActivatedEventArgs
```
jonwis, line 9: Can this PR also be used to apply the appropriate contract information?  @DrusTheAxe or @dhoehna might have guidance for you on the specific one to use.

dhoehna: I agree with @jonwis.  This PR can be used to add contracts.  

The questions is this "Should a contract be added to this runtime class, or something higher up?"

loneursid: Happy to do that. @dhoehna - What do we need to decide if we apply the contract to this class or to a higher  up?

What do contracts look like? Can you point me to an example?

dhoehna: An example is here: https://github.com/microsoft/WindowsAppSDK/pull/2110#pullrequestreview-885233578

@jonwis @DrusTheAxe what level should this new contract be placed?

DrusTheAxe: Sorry, what do you mean by "level"?

What is "higher up" than a (runtime)class? A namespace? I didn't think you could apply a contract to a namespace, and even if you _could_ I'm not sure you _should_. Would that mean everything in namespace A.B gets it...unless specifically overridden per type? Would child namespaces inherit it too? MIDL doesn't tend to like varying definition of a type across...an .idl? Compilation? e.g.

```
[contract(FooContract, 1)]
namespace Foo
{
    [contractversion(1)]
    apicontract FooContract;

    [contract(FooContract, 1)]
    runtimeclass Bar {...}
}
```

MIDL historically doesn't like undefined type references that are later defined, as in this example **line 1**. I'm doubtful if this compiles.

Namespaces aren't types as I understand it. If so then I wouldn't expect contracts to apply to namespaces.

runtimeclass, interface, enum, more are types and be have a contract applied.

Do the MIDL docs clearly define what `[contract(...)]` can be applied to? If not that would be a doc bug.

Do the MIDL(?) docs provide _guidance_ onw how you could _and should_ use them? As we've seen contracts are a useful versioning mechanism for more than OS types. @BenJKuhn guidance on when/how to use them may be thin but if so, we should improve that. Need a doc bug?

DrusTheAxe:
> ran the terminal velocity generator with "Stable"

You should always run the generator with `-Channel experimental` (or omit `-Channel`, the default is `experimental`) when generating files to checkin.

The build pipeline reruns it for the pipelines particular configuration so doesn't matter there. The checked in files are for use by devs during inner loop. In _this_ case there's no difference across channels but if there were the generated code would differ, e.g. `-Channel stable` would generate code that thinks it's a stable branch and thus omit/disable any experimental functionality guarded by the velocity feature check. But since main is expected to have experimental bits you won't get what you expect.

I suggest rerunning the generator without `-Channel` as it's no different _today_ but to avoid any potential future surprises.

dev/AppNotifications/AppNotifications.idl
dhoehna: > What is "higher up" than a (runtime)class? 

I don't know actually.  I thought that there was something higher up than a runtime class.  My mistake.

@loneursid Please add a contract to AppNotifications and call it WindowsAppNotificationsContract.  Start with version 1.

loneursid: @dhoehna - I'm getting this error: **error MIDL2025: [msg]syntax error [context]: expecting { near "["**

Here's what I've tired: (based on your example)
```
namespace Microsoft.Windows.AppNotifications
{
    [contractversion(1)]
    apicontract WindowsAppNotificationsContract

    // Event args for the Notification Activation
    [Contract(MicrosoftWindowsAppNotifications, 1)]
    runtimeclass AppNotificationActivatedEventArgs

...
```

riverar: Should have braces, not a semi-colon.

```
[contractversion(1)]
apicontract WindowsAppNotificationsContract { }
```

dhoehna: Sorry about that.  The example has been updated.

loneursid: @dhoehna  - I have tagged the 3 runtime classes in this namespace with the contract. I haven't tagged the enums? Should I?

@DrusTheAxe - I've kept the initial name for the contract: WindowsAppNotificationsContract, let me know if I should change it.

MikeHillberg:
> I'm not sure about the [feature(Feature_PushNotifications)] attribute

The [feature] attribute is a way to put _on_ [experimental].

dev/AppNotifications/AppNotifications.idl
DrusTheAxe: > WindowsAppNotificationsContract, let me know if I should change it.

IMO yes, as the 'Windows' part refers to the inbox Windows product types and not the appnotifications functionality. thus `namespace Microsoft.AppNotifications { apicontract AppNotificationsContract {} }`.

But it's an API Review question. You should ask your API Design representative.

DrusTheAxe: > I haven't tagged the enums

Yes

DrusTheAxe:
> Should I create these files?

Yes.

Velocity definitions (*.xml) should be feature based (for our sanity :-). Better multiple small ones than a few big mosh pits

dev/AppNotifications/AppNotification.cpp
```diff
@@ -6,8 +6,15 @@ using namespace winrt::Windows::Data::Xml::Dom;
 
 namespace winrt::Microsoft::Windows::AppNotifications::implementation
 {
+    AppNotification::AppNotification()
+    {
+        THROW_HR_IF(E_NOTIMPL, !::Microsoft::Windows::PushNotifications::Feature_AppNotifications::IsEnabled());
```
hulumane, line 11: Not needed here. Just put this in the beginning of AppNotificationManager::GetDEfault() which is the real entry point to everything AppNotification related.

dev/AppNotifications/AppNotification.cpp
```diff
@@ -6,8 +6,15 @@ using namespace winrt::Windows::Data::Xml::Dom;
...
+    }
+
     AppNotification::AppNotification(hstring const& payload)
     {
+        THROW_HR_IF(E_NOTIMPL, !::Microsoft::Windows::PushNotifications::Feature_AppNotifications::IsEnabled());
```
hulumane, line 16: Not needed. See comment above.

hulumane: Actually I see this being put everywhere else? Why not gate all the entire feature via AppManager::Default() instead of these individual checks?

dev/Common/TerminalVelocityFeatures-AppNotifications.xml
```diff
@@ -0,0 +1,16 @@
...
+<!-- Copyright (c) Microsoft Corporation. All rights reserved.
+     Licensed under the MIT License. See LICENSE in the project root for license information.
+-->
+
+    <!-- See docs/TerminalVelocity.md for more info.  -->
+
+    <!-- Enabled variants -->
+    <feature>
+        <name>Feature_AppNotifications</name>
+        <description>App Notifications for the WindowsAppRuntime</description>
```
hulumane, line 13: Nit: Better description would be "Enables Local App sourced Notifications on the device"

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -47,7 +47,10 @@ using namespace Microsoft::Windows::AppNotifications::Helpers;
...
 namespace winrt::Microsoft::Windows::AppNotifications::implementation
 {
-    AppNotificationManager::AppNotificationManager() : m_processName(GetCurrentProcessPath()) {}
+    AppNotificationManager::AppNotificationManager() : m_processName(GetCurrentProcessPath())
+    {
+        THROW_HR_IF(E_NOTIMPL, !::Microsoft::Windows::PushNotifications::Feature_AppNotifications::IsEnabled());
+    }
 
     winrt::Microsoft::Windows::AppNotifications::AppNotificationManager AppNotificationManager::Default()
     {
```
hulumane, line 56: Probably need the feature gate here since this is the entry point of the AppNotification feature set.

dev/Common/TerminalVelocityFeatures-AppNotifications.xml
loneursid: done

dev/AppNotifications/AppNotification.cpp
loneursid: I had added this into the ctor of each public runtime class, that's what push seems to be doing. 

I've removed it from all ctor (for toast) and only added one single check in AppNotificationManager::Default().

dev/AppNotifications/AppNotification.cpp
loneursid: I had added this into the ctor of each public runtime class, that's what push seems to be doing. 

I've removed it from all ctor (for toast) and only added one single check in AppNotificationManager::Default().

hulumane (approved):

loneursid:
@DrusTheAxe and @MikeHillberg 

Last question before I pull the trigger on this PR.

I've added [feature(Feature_AppNotifications)] to the IDL and I've gated it in AppNotificationManager::Default() as this is the entry point to the app notifications subsystem.

My question is: is this sufficient or should I also gate all the ctors from any object a dev can possibly construct but is otherwise useless anyways.

I assume that the APIs get stripped from the IDL and that a devs has to go through some hoops in order to cann the functions and that the gate  is only there to enforce the policy, in case someone tries to go around, right?

loneursid:
@DrusTheAxe and @MikeHillberg 

Last question before I pull the trigger on this PR.

I've added [feature(Feature_AppNotifications)] to the IDL and I've gated it in AppNotificationManager::Default() as this is the entry point to the app notifications subsystem.

My question is: is this sufficient or should I also gate all the ctors from any object a dev can possibly construct but is otherwise useless anyways.

I assume that the APIs get stripped from the IDL and that a devs has to go through some hoops in order to cann the functions and that the gate  is only there to enforce the policy, in case someone tries to go around, right?

hulumane:
> @DrusTheAxe and @MikeHillberg
> 
> Last question before I pull the trigger on this PR.
> 
> I've added [feature(Feature_AppNotifications)] to the IDL and I've gated it in AppNotificationManager::Default() as this is the entry point to the app notifications subsystem.
> 
> My question is: is this sufficient or should I also gate all the ctors from any object a dev can possibly construct but is otherwise useless anyways.
> 
> I assume that the APIs get stripped from the IDL and that a devs has to go through some hoops in order to cann the functions and that the gate is only there to enforce the policy, in case someone tries to go around, right?

I'm guessing that gating the ctors just implies completeness of the feature gate but in practicality doesn't really achieve much since we gate the choke point in Manager::GetDefault() anyways.

DrusTheAxe:
> is this sufficient or should I also gate all the ctors

Can a dev instantiate an object with the ctors? If so then you should gate them too

If AppNotifications is experimental then any use of AppNotifications types should fail

Issue #1898, opened 2021-12-11
MRT Core does not support VS's FastUpToDate check


evelynwu-msft:
### Describe the bug

PRI files are not integrated into the FastUpToDate logic, so if only a `.resw` is touched in a project A that project B depends upon, VS will incorrectly believe that project B is still up-to-date relative to project A (project A does not have any up-to-date markers corresponding to its PRI file, and the PRI file is the only output that changed) and thus doesn't need to be incrementally built.

This can be addressed by making the following changes to `touch` the [`CopyUpToDateMarker` file](https://github.com/dotnet/project-system/blob/main/docs/repo/up-to-date-check-implementation.md#copyuptodatemarker-and-resolvedcompilationreference), remove `.resw` and `.resjson` from the default set of up-to-date check inputs, and explicitly registering them ourselves with the fast up-to-date check infrastructure:

`MrtCore.PriGen.targets`:
```xml
  <ItemGroup>
    <PropertyPageSchema Include="$(MSBuildThisFileDirectory)ProjectItemsSchema.xaml" />
  </ItemGroup>

  <ItemGroup>
    <UpToDateCheckInput Include="@(PRIResource)" Set="PRIGeneration" />
    <UpToDateCheckBuilt Include="$(ProjectPriFullPath)" Set="PRIGeneration" />
  </ItemGroup>

  <Target Name="_GenerateProjectPriFileCore"
          Inputs="$(MSBuildAllProjects);@(_GenerateProjectPriFileCoreInput)"
          Outputs="$(ProjectPriFullPath)">

    <MakeDir Condition="'$(InsertReverseMap)' == 'true'"
             Directories="$(_ReverseMapProjectPriDirectory)"
                 />

    <GenerateProjectPriFile MakePriExeFullPath="$(MakePriExeFullPath)"
                            MakePriExtensionPath="$(OutOfProcessMakePriExtensionPath)"
                            PriConfigXmlPath="$(_PriConfigXmlPath)"
                            IndexFilesForQualifiersCollection="$(_FilteredLayoutResfilesPath);$(_ResourcesResfilesPath)"
                            ProjectPriIndexName="$(ProjectPriIndexName)"
                            InsertReverseMap="$(InsertReverseMap)"
                            ProjectDirectory="$(ProjectDir)"
                            OutputFileName="$(ProjectPriFullPath)"
                            QualifiersPath="$(_QualifiersPath)"
[...]

PR #2226, opened 2022-03-08
[MRT Core] Add support for VS's FastUpToDate check in MRT Core.

rohanp-msft:
**Description**
Please see issue: https://github.com/microsoft/WindowsAppSDK/issues/1898

**Credit**
The fix here is from @evelynwu-msft (thanks Evelyn!) with two modifications from me.

**How verified**
Used https://github.com/microsoft/WindowsAppSDK/issues/1898#issuecomment-1058629425 with the following modifications:
- Upgraded the NuGet package to WAS 1.0.
- Commented out the contents of Directory.Build.targets.
- Moved fix into MrtCore.PriGen.targets in the NuGet pkg with the None ItemGroup removal bit removed (it's not needed, and if kept, will introduce an issue).
- Removed the change to the DisplayName (changes the name displayed in the Build Action drop-down list where there is already a "Resource").

Without and with fix:
- Delete string in RESW in the StringResources project.
- Build the solution.
- Add string in RESW in the StringResources project.
- Build the solution.

With the fix, all 4 projects are built, and without, only 2 are built (StringResources and the WAPPROJ project).

### DIFF
--- a/build/publish-mrt.yml
+++ b/build/publish-mrt.yml
@@ -112,7 +112,6 @@ steps:
       native\MrtCore.C.props
       native\MrtCore.props
       native\MrtCore.targets
-      ProjectItemsSchema.xaml
       README.md
     TargetFolder: '$(Build.ArtifactStagingDirectory)\fullnuget\build'
     flattenFolders: false
--- a/build/publish-mrt.yml
+++ b/build/publish-mrt.yml
@@ -112,6 +112,7 @@ steps:
       native\MrtCore.C.props
       native\MrtCore.props
       native\MrtCore.targets
+      ProjectItemsSchema.xaml
       README.md
     TargetFolder: '$(Build.ArtifactStagingDirectory)\fullnuget\build'
     flattenFolders: false
--- a/dev/MRTCore/packaging/MrtCore.PriGen.targets
+++ b/dev/MRTCore/packaging/MrtCore.PriGen.targets
@@ -1585,7 +1585,7 @@
     </ValidateConfiguration>
   </Target>
 
-  <!-- Ensure Fast Up To Date picks up changes to PRI related files. -->
+  <!-- Configure Fast Up To Date for PRI generation. Please see the comment in ProjectItemsSchema.xaml for more info. -->
 
   <ItemGroup>
     <PropertyPageSchema Include="$(MSBuildThisFileDirectory)ProjectItemsSchema.xaml" />
@@ -1599,19 +1599,18 @@
     </_MrtCoreUpToDateCheckInputDependsOn>
   </PropertyGroup>
 
-  <!-- Inform the fast up-to-date check of the inputs for PRI generation -->
+  <!-- Inform the fast up-to-date check of the inputs for PRI generation. -->
   <Target Name="_MrtCoreUpToDateCheckInput" DependsOnTargets="$(_MrtCoreUpToDateCheckInputDependsOn)" BeforeTargets="CollectUpToDateCheckInputDesignTime">
-    <!-- 
-        Copied from the MRT Core target _GenerateProjectPriConfigurationFiles.
-        @(_LayoutFile),  @(_EmbedFile), and @(_PRIResourceFiltered) are ultimately inputs to the MRT Core indexer, 
-        but they're populated by _GenerateProjectPriConfigurationFiles which does work (runs a tool to generate the indexer configuration files)
-        that we don't need during a design-time build, so we hoist them out to avoid calling the target.
+        <!-- 
+        Copied from the target _GenerateProjectPriConfigurationFiles. @(_LayoutFile),  @(_EmbedFile), and @(_PRIResourceFiltered) are ultimately inputs to
+        the MRT Core indexer, but they're populated by _GenerateProjectPriConfigurationFiles which does work (runs a tool to generate the indexer
+        configuration files) that we don't need during a design-time build, so we hoist them out to avoid calling the target.
         -->
     <ItemGroup>
-    <!--
-        First, build out the complete list of files we want to consider for the layout.
-        Then exclude anything that matches any pattern or filename listed in _AppxLayoutAssetPackageFiles.
-        We could do this as a 'Remove' operation, but by building an oracle we don't modify, we simplify future manipulations of this data set.
+        <!--
+        First, build out the complete list of files we want to consider for the layout. Then exclude anything that matches any pattern or filename listed in
+        _AppxLayoutAssetPackageFiles. We could do this as a 'Remove' operation, but by building an oracle we don't modify, we simplify future manipulations of
+        this data set.
         -->
       <_LayoutFileSource Include="@(PackagingOutputs)" Condition="'%(OutputGroup)' == 'ContentFilesProjectOutputGroup' and '%(ProjectName)' == '$(ProjectName)'" />
       <_LayoutFileSource Include="@(PackagingOutputs)" Condition="'%(OutputGroup)' == 'CustomOutputGroupForPackaging' and '%(ProjectName)' == '$(ProjectName)'" />
--- a/dev/MRTCore/packaging/MrtCore.PriGen.targets
+++ b/dev/MRTCore/packaging/MrtCore.PriGen.targets
@@ -1584,4 +1584,70 @@
       Platform="$(Platform)">
     </ValidateConfiguration>
   </Target>
+
+  <!-- Ensure Fast Up To Date picks up changes to PRI related files. -->
+
+  <ItemGroup>
+    <PropertyPageSchema Include="$(MSBuildThisFileDirectory)ProjectItemsSchema.xaml" />
+  </ItemGroup>
+
+  <PropertyGroup>
+    <_MrtCoreUpToDateCheckInputDependsOn>
+      $(_MrtCoreUpToDateCheckInputDependsOn);
+      GetMrtPackagingOutputs;
+      _GetPriFilesFromPayload;
+    </_MrtCoreUpToDateCheckInputDependsOn>
+  </PropertyGroup>
+
+  <!-- Inform the fast up-to-date check of the inputs for PRI generation -->
+  <Target Name="_MrtCoreUpToDateCheckInput" DependsOnTargets="$(_MrtCoreUpToDateCheckInputDependsOn)" BeforeTargets="CollectUpToDateCheckInputDesignTime">
+    <!-- 
+        Copied from the MRT Core target _GenerateProjectPriConfigurationFiles.
+        @(_LayoutFile),  @(_EmbedFile), and @(_PRIResourceFiltered) are ultimately inputs to the MRT Core indexer, 
+        but they're populated by _GenerateProjectPriConfigurationFiles which does work (runs a tool to generate the indexer configuration files)
+        that we don't need during a design-time build, so we hoist them out to avoid calling the target.
+        -->
+    <ItemGroup>
+    <!--
+        First, build out the complete list of files we want to consider for the layout.
+        Then exclude anything that matches any pattern or filename listed in _AppxLayoutAssetPackageFiles.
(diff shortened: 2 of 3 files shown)


axelandrejs (approved):

dev/MRTCore/packaging/MrtCore.PriGen.targets
```diff
@@ -1584,4 +1584,69 @@
...
+      <_UpToDateCheckInputTemp Include="@(_PRIResourceFiltered)" Set="PRIGeneration" />
+      <_UpToDateCheckInputTemp Include="@(_PriFilesFromPayload)" Set="PRIGeneration" />
+    </ItemGroup>
+
+    <RemoveDuplicates Inputs="@(_UpToDateCheckInputTemp)">
+      <Output TaskParameter="Filtered" ItemName="_UpToDateCheckInputDeduplicated" />
+    </RemoveDuplicates>
+
+    <ItemGroup>
+      <UpToDateCheckInput Include="@(_UpToDateCheckInputDeduplicated)" />
```
evelynwu-msft, line 1642: This is missing the `Set="PRIGeneration"` attribute.

evelynwu-msft (approved):
:shipit:

build/publish-mrt.yml
```diff
@@ -112,6 +112,7 @@ steps:
       native\MrtCore.C.props
       native\MrtCore.props
       native\MrtCore.targets
+      ProjectItemsSchema.xaml
       README.md
```
huichen123, line 116: >README.md [](http://example.com/codeflow?start=6&length=9)

this is probably a mistake

rohanp-msft: Seems fine:
image

I'll also verify the change with the NuGet package from the pipeline build.

huichen123: this is fine. we shouldn't have this file in the final package. but not your change. we can remove later.

huichen123:
> <?xml version="1.0" encoding="utf-8" ?>

winui doesn't use this. is that a problem too?

---
Refers to: dev/MRTCore/packaging/MrtCore.PriGen.targets:1 in ee096d6. [](commit_id = ee096d66070775c43ed79c94f1a068ee6e935cb1, deletion_comment = False)

dev/MRTCore/packaging/MrtCore.PriGen.targets
```diff
@@ -1584,4 +1584,69 @@
...
+      <_LayoutFileXbfXaml Include="$([System.IO.Path]::ChangeExtension('%(_LayoutFile.Identity)','.xaml'))" Condition="'%(Extension)' == '.xbf'" />
+      <_LayoutFile Remove="@(_LayoutFileXbfXaml)" />
+
+      <!-- Filter out PRIResource files which are marked by C++ project system as ExcludedFromBuild. -->
+      <_PRIResourceFiltered Include="@(PRIResource)" Condition="'%(PRIResource.ExcludedFromBuild)' != 'true'" />
+    </ItemGroup>
+
+    <ItemGroup>
+      <_UpToDateCheckInputTemp Include="@(SourceAppxManifest)" Set="PRIGeneration" />
+      <_UpToDateCheckInputTemp Include="@(_LayoutFile)" Condition="'%(Extension)'!='.xbf'" Set="PRIGeneration" />
```
huichen123, line 1631: >'%(Extension)'!='.xbf' [](http://example.com/codeflow?start=67&length=22)

is xbf always embedded?

rohanp-msft: I believe so. @evelynwu-msft to confirm.

rohanp-msft:
> > 
> 
> winui doesn't use this. is that a problem too?
> 
> Refers to: dev/MRTCore/packaging/MrtCore.PriGen.targets:1 in [ee096d6](https://github.com/microsoft/WindowsAppSDK/commit/ee096d66070775c43ed79c94f1a068ee6e935cb1). [](commit_id = [ee096d6](https://github.com/microsoft/WindowsAppSDK/commit/ee096d66070775c43ed79c94f1a068ee6e935cb1), deletion_comment = False)

WinUI does use this for WAP proj based apps.

evelynwu-msft:
> <?xml version="1.0" encoding="utf-8" ?>

The XML declaration is optional. UTF-8 is the default encoding when no declaration is present unless there's a Unicode BOM (in which case it's UTF-16). [https://docs.microsoft.com/en-us/previous-versions/troubleshoot/msxml/xml-encoding-dom-interface-methods](https://docs.microsoft.com/en-us/previous-versions/troubleshoot/msxml/xml-encoding-dom-interface-methods)

Including it in this case is making things extra-explicit, but there's no advantage or disadvantage to doing so. Personally, I would bias to including it in new XML files (if I remembered to) but I wouldn't go out of my way to add it to existing files.

---
In reply to: [1064466833](https://github.com/microsoft/WindowsAppSDK/pull/2226#issuecomment-1064466833) [](http://example.com/codeflow?ancestors=1064466833,1064466833)

---
Refers to: dev/MRTCore/packaging/MrtCore.PriGen.targets:1 in ee096d6. [](commit_id = ee096d66070775c43ed79c94f1a068ee6e935cb1, deletion_comment = False)

dev/MRTCore/packaging/MrtCore.PriGen.targets
evelynwu-msft: There are cases where they are not, but by definition those are consequently not inputs to resources.pri.

huichen123: if xbf is not embedded, wouldn't changing of xbf require a build to update the file?

evelynwu-msft: (to be honest, I don't recall why I added this condition)

evelynwu-msft: .xbf files will only change if source .xaml or code-behind are modified, and this will already trigger a build independent of what MRT Core does/doesn't do.

huichen123: then do we need to do the extra remove from line 1621-1623?

rohanp-msft: If we have a loose XBF, then we're not going to embed it (I believe this is our intended design). We exclude XBFs in the input set for that reason, so excluding the corresponding XAMLs too makes sense to me. @evelynwu-msft is all that correct?

evelynwu-msft: I could imagine a weird edge cases where a developer embeds a `.xbf` that was externally generated by including it in the package as part of `@(Content)` In that case, you'd want a change to it to be recognized as making `resources.pri` out of date.

rohanp-msft: @huichen123 Spoke with Evelyn offline. I'll remove the XBF filtering in `<_UpToDateCheckInputTemp Include="@(_LayoutFile)" Condition="'%(Extension)'!='.xbf'" Set="PRIGeneration" />` and leave the _conditional_ XAML removal as it is. Any objections?

huichen123: sounds good to me.

huichen123:
> <?xml version="1.0" encoding="utf-8" ?>

I meant WinUI doesn't use MrtCore.PriGen.targets, does it?

---
In reply to: [1064476702](https://github.com/microsoft/WindowsAppSDK/pull/2226#issuecomment-1064476702) [](http://example.com/codeflow?ancestors=1064476702,1064466833)

---
Refers to: dev/MRTCore/packaging/MrtCore.PriGen.targets:1 in ee096d6. [](commit_id = ee096d66070775c43ed79c94f1a068ee6e935cb1, deletion_comment = False)

rohanp-msft:
> > 
> 
> I meant WinUI doesn't use MrtCore.PriGen.targets, does it?
> 
> In reply to: [1064476702](https://github.com/microsoft/WindowsAppSDK/pull/2226#issuecomment-1064476702)
> 
> Refers to: dev/MRTCore/packaging/MrtCore.PriGen.targets:1 in [ee096d6](https://github.com/microsoft/WindowsAppSDK/commit/ee096d66070775c43ed79c94f1a068ee6e935cb1). [](commit_id = [ee096d6](https://github.com/microsoft/WindowsAppSDK/commit/ee096d66070775c43ed79c94f1a068ee6e935cb1), deletion_comment = False)

I think the WAPPROJ based solutions use it.

huichen123:
WinUI has "<EnableCoreMrtTooling>false</EnableCoreMrtTooling>". It's no longer the case?

---
In reply to: [1068504985](https://github.com/microsoft/WindowsAppSDK/pull/2226#issuecomment-1068504985) [](http://example.com/codeflow?ancestors=1068504985,1068504985)

rohanp-msft:
> WinUI has "false". It's no longer the case?
> 
> In reply to: [1068504985](https://github.com/microsoft/WindowsAppSDK/pull/2226#issuecomment-1068504985)

Can you clarify which property you're referring to?

huichen123:
uhh, xml doesn't display well. it's EnableCoreMrtTooling

---
In reply to: [1068524766](https://github.com/microsoft/WindowsAppSDK/pull/2226#issuecomment-1068524766) [](http://example.com/codeflow?ancestors=1068524766,1068524766)

rohanp-msft:
> uhh, xml doesn't display well. it's EnableCoreMrtTooling
> 
> In reply to: [1068524766](https://github.com/microsoft/WindowsAppSDK/pull/2226#issuecomment-1068524766)

EnableCoreMrtTooling is our control switch. WinUI doesn't set it or reference it.

huichen123:
isn't that the switch controls importing mrtcore.prigen.targets? (WinUI has a separate path to import, but that should also be turned off in normal case)

---
In reply to: [1069401717](https://github.com/microsoft/WindowsAppSDK/pull/2226#issuecomment-1069401717) [](http://example.com/codeflow?ancestors=1069401717,1069401717)

huichen123:
what you quoted is for WAP or UWP which is of no interest to me here. My point is if mrtcore.prigen.targets is not imported, your change would make no effect.

---
In reply to: [1069561942](https://github.com/microsoft/WindowsAppSDK/pull/2226#issuecomment-1069561942) [](http://example.com/codeflow?ancestors=1069561942,1069561942)

rohanp-msft:
> what you quoted is for WAP or UWP which is of no interest to me here. My point is if mrtcore.prigen.targets is not imported, your change would make no effect.
> 
> In reply to: [1069561942](https://github.com/microsoft/WindowsAppSDK/pull/2226#issuecomment-1069561942)

WinUI does use this in the non single project MSIX case.

huichen123:
it's not clear to me how WinUI uses this based on code search. but i could miss something. How about single project MSIX or unpackaged?

---
In reply to: [1069585917](https://github.com/microsoft/WindowsAppSDK/pull/2226#issuecomment-1069585917) [](http://example.com/codeflow?ancestors=1069585917,1069585917)

rohanp-msft:
Single project MSIX uses Microsoft.Build.Msix.Pri.targets. Not sure about unpackaged (I suspect WinUI doesn't support it). I suggest asking Evelyn or Tommy.

huichen123:
do you need to make sure your fix covers all cases?

---
In reply to: [1069747148](https://github.com/microsoft/WindowsAppSDK/pull/2226#issuecomment-1069747148) [](http://example.com/codeflow?ancestors=1069747148,1069747148)

rohanp-msft:
> do you need to make sure your fix covers all cases?
> 
> In reply to: [1069747148](https://github.com/microsoft/WindowsAppSDK/pull/2226#issuecomment-1069747148)

I think this is sufficient. The verification is whether the project's changes are picked up by FUTD or not and whether the parent projects are rebuilt. That's what was done.

PR #2234, opened 2022-03-10
Add checks for SelfContained in AppNotificationManager and PushNotificationManager

hulumane:
Issue:
Self-Contained apps may not support notifications in some cases. Please refer to the matrix below:
Screenshot 2022-03-10 104221

In general, All local notification scenarios will be supported in self-contained mode. We will not support a subset of Push scenarios for Self-Contained apps.

Fix:
For local notifications, we gate the calls to the Long Running Singleton via a Self-Contained checks so that API calls don't blow up.

For Push notifications, we introduce an IsSupported gatekeeping API that the developer could use to infer Push functionality.

Also added some checks to prevent races in AppNotificationManager::Register() APIs.


test/TestApps/PushNotificationsTestApp/PushNotificationsTestApp.vcxproj
```diff
@@ -153,7 +153,7 @@
       <PrecompiledHeader>Use</PrecompiledHeader>
       <WarningLevel>Level3</WarningLevel>
       <SDLCheck>true</SDLCheck>
-      <AdditionalIncludeDirectories>%(AdditionalIncludeDirectories);$(OutDir)\..\WindowsAppRuntime_DLL;$(OutDir)\..\WindowsAppRuntime_BootstrapDLL;..\..\inc</AdditionalIncludeDirectories>
+      <AdditionalIncludeDirectories>$(RepoRoot)\Dev\Common;%(AdditionalIncludeDirectories);$(OutDir)\..\WindowsAppRuntime_DLL;$(OutDir)\..\WindowsAppRuntime_BootstrapDLL;..\..\inc</AdditionalIncludeDirectories>
```
DrusTheAxe, line 156: NIT: `dev` (not `Dev`)

Ultra-NITy. Is my inner OCD showing? :P

DrusTheAxe (approved):

dev/PushNotifications/PushNotifications.idl
```diff
@@ -76,6 +71,10 @@ namespace Microsoft.Windows.PushNotifications
     [feature(Feature_PushNotifications)]
     runtimeclass PushNotificationManager
     {
+        // Checks to see if the APIs are supported for the Desktop app
```
hulumane, line 74: @jonwis @brflynn There are some minor idl changes here. Just wanted to check that we are good here.

dev/PushNotifications/PushNotifications.idl
```diff
@@ -58,11 +58,6 @@ namespace Microsoft.Windows.PushNotifications
     [feature(Feature_PushNotifications)]
     runtimeclass PushNotificationCreateChannelResult
     {
-        PushNotificationCreateChannelResult(
```
hulumane, line 61: This should have never been exposed to developers as a projection. The ctor is internal.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -105,10 +106,21 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
     }
     CATCH_RETURN()
 
+    bool PushNotificationManager::IsSupported()
+    {
+        // Only scenarios that use the Background Infrastructure component of the OS support Push in the SelfContained case
+        if (WindowsAppRuntime::SelfContained::IsSelfContained() && !PushNotificationHelpers::IsPackagedAppScenario())
```
pmpurifoy, line 112: Suggest:
`static bool isSelfContained { WindowsAppRuntime::SelfContained::IsSelfContained() && !PushNotificationHelpers::IsPackagedAppScenario() }`

So we only calculate this one time.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -136,23 +148,31 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+                    return winrt::make<winrt::Microsoft::Windows::PushNotifications::implementation::PushNotificationReceivedEventArgs>(payloadAsWstring);
+                }
             }
-        }
 
-        THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_TIMEOUT), !m_waitHandleForArgs.wait(c_receiveArgsTimeoutInMSec));
-        THROW_HR_IF(E_UNEXPECTED, !m_backgroundTaskArgs);
-        return m_backgroundTaskArgs;
+            THROW_HR_MSG(E_UNEXPECTED, "Could not serialize payload from command line Uri!");
+            return nullptr;
```
pmpurifoy, line 174: Is there a way we can restructure this so we don't have unreachable code? 

Suggestion: Set a result variable PushNotificationEventArgs at the front of the if/else and assign it as needed. In the for loop, break on success and if !found throw E_UNEXPECTED.

dev/PushNotifications/PushNotificationManager.cpp
hulumane: done

dev/PushNotifications/PushNotificationManager.cpp
hulumane: Cleaned this up.

test/TestApps/PushNotificationsTestApp/PushNotificationsTestApp.vcxproj
hulumane: Ok. Will fix to lower cased

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -136,23 +145,36 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-                std::wstring payloadAsWstring{ winrt::Windows::Foundation::Uri::UnescapeComponent(payloadAsEscapedWstring) };
-                return winrt::make<winrt::Microsoft::Windows::PushNotifications::implementation::PushNotificationReceivedEventArgs>(payloadAsWstring);
+                if (pair.Name() == L"payload")
+                {
+                    // Convert escaped components to its normal content
+                    // from the conversion done in the LRP (see NotificationListener.cpp)
+                    std::wstring payloadAsEscapedWstring{ pair.Value() };
+                    std::wstring payloadAsWstring{ winrt::Windows::Foundation::Uri::UnescapeComponent(payloadAsEscapedWstring) };
+                    eventArgs = winrt::make<winrt::Microsoft::Windows::PushNotifications::implementation::PushNotificationReceivedEventArgs>(payloadAsWstring);
+                    found = true;
```
hulumane, line 170: No need of bool. Use eventArgs as the truth.

jonwis:
What is "vibranium"? What Windows release number is that?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -111,10 +112,18 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+        static bool isSupported{ !WindowsAppRuntime::SelfContained::IsSelfContained() || PushNotificationHelpers::IsPackagedAppScenario() };
+        return isSupported;
+       
+    }
+
     winrt::Microsoft::Windows::PushNotifications::PushNotificationManager PushNotificationManager::Default()
     {
         THROW_HR_IF(E_NOTIMPL, !::Microsoft::Windows::PushNotifications::Feature_PushNotifications::IsEnabled());
-
+        THROW_HR_IF_MSG(E_FAIL, !IsSupported(), "Push APIs are not supported for this Application Type!");
```
jonwis, line 126: The `IsSupported` pattern for WinRT APIs does *not* require the caller check IsSupported before calling. If the thing isn't supported, then future uses of it are "inert" - meaning they do nothing, and provide a reasonable state as a result.  So for instance, calling Register or Unregister[All] has no side-effect, the app never gets an incoming notification, and any requests to open a channel get back the "oops, couldn't do that" state.  But _not_ an exception here.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -142,23 +151,34 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-                // Convert escaped components to its normal content
-                // from the conversion done in the LRP (see NotificationListener.cpp)
-                std::wstring payloadAsEscapedWstring{ pair.Value() };
-                std::wstring payloadAsWstring{ winrt::Windows::Foundation::Uri::UnescapeComponent(payloadAsEscapedWstring) };
-                return winrt::make<winrt::Microsoft::Windows::PushNotifications::implementation::PushNotificationReceivedEventArgs>(payloadAsWstring);
+                if (pair.Name() == L"payload")
+                {
+                    // Convert escaped components to its normal content
+                    // from the conversion done in the LRP (see NotificationListener.cpp)
+                    std::wstring payloadAsEscapedWstring{ pair.Value() };
```
jonwis, line 172: Shouldn't need this temporary.  `pair.Value()` is of type `winrt::hstring` and `UnescapeComponent` takes hstring.  When possible, avoid copying between string types - and avoid copying strings.

Consider:

```c++
auto unescapedString = winrt::Windows::Foundation::Uri::UnescapeComponent(pair.Value());
```

... and then make sure PushNotificationReceivedEventArgs's ctor accepts `winrt::hstring`

dev/PushNotifications/PushNotifications.idl
jonwis: I don't _mind_ exposing the ctor - it would make testing of app code easier, as the ISV can just create a new Args and hand it to its Args-handling method, rather than having to invent a wrapper layer.

jonwis (approved):

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -126,17 +141,50 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
 
     void AppNotificationManager::Unregister()
     {
+        {
+            auto lock{ m_lock.lock_exclusive() };
+            THROW_HR_IF_MSG(E_FAIL, m_registering, "Register or Unregister currently in progress!");
```
danielayala94, line 146: This error code should be `ERROR_OPERATION_IN_PROGRESS`.

hulumane: Will update

dev/PushNotifications/PushNotificationManager.cpp
hulumane: Thanks for pointing this out. I have removed the THROW here and gated the remaining APIs in PushManager to have no side effects for the !IsSupported() case.

specs/PushNotifications/PushNotifications-spec.md
```diff
@@ -453,6 +448,10 @@ namespace Microsoft.Windows.PushNotifications
     [feature(Feature_PushNotifications)]
     runtimeclass PushNotificationManager
     {
+        // Checks to see if the APIs are supported for the Desktop app
+        // Certain self-contained apps may not support Push Notification scenarios by design
+        static Boolean IsSupported();
```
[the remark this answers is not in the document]
hulumane: I'll update the sample in the spec to use this correctly

dev/PushNotifications/PushNotificationManager.cpp
hulumane: Done. I updated all the methods to use hstring. There is one API helper at the bottom of the stack which still requires a change to wstring but changing that would impact code elsewhere so I decided to leave as is.

dev/PushNotifications/PushNotifications.idl
hulumane: Right. I was just going with correctness here since it wasn't really needed from the projection perspective. In the past API review always erred on the side of exposing projections only if absolutely needed. Unless you do feel super strongly here, I will go ahead and remove since the test scenario can still be exercised by the developer using an E2E test.

specs/PushNotifications/PushNotifications-spec.md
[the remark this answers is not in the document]
hulumane: @jonwis @danielayala94 please refer to the new updates to see how the developer can use IsSupported() API in a useful way. Please let me know if you have any comments on this.

jonwis (approved):

danielayala94 (approved):

raghavthind2005 (approved):

raghavthind2005 (approved):

pmpurifoy:
Done in another PR. Closing

PR #2238, opened 2022-03-10
Default display name when no window handle available in Notifications

sharath2727:
Description: GetConsoleWindow in AppNotificationUtility returns a Null pointer when we are working with WinUI apps.

Fix: Fallback to default displayname and icon when console window is not available for use.

### DIFF
--- a/dev/AppNotifications/AppNotificationUtility.cpp
+++ b/dev/AppNotifications/AppNotificationUtility.cpp
@@ -276,7 +276,7 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring 
 
         wil::unique_prop_variant propVariantIcon;
         THROW_IF_FAILED(propertyStore->GetValue(PKEY_AppUserModel_RelaunchIconResource, &propVariantIcon));
-        THROW_HR_IF_MSG(E_UNEXPECTED, propVariantIcon.vt == VT_EMPTY, "You must specify a valid image filepath before calling Register().");
+        THROW_HR_IF_MSG(E_UNEXPECTED, (propVariantIcon.vt != VT_LPWSTR || propVariantIcon.pwszVal == nullptr), "You must specify a valid image filepath before calling Register().");
 
         iconFilePath = propVariantIcon.pwszVal;
 
--- a/dev/AppNotifications/AppNotificationUtility.cpp
+++ b/dev/AppNotifications/AppNotificationUtility.cpp
@@ -278,11 +278,11 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring 
         wil::unique_prop_variant propVariantIcon;
         THROW_IF_FAILED(propertyStore->GetValue(PKEY_AppUserModel_RelaunchIconResource, &propVariantIcon));
 
-        THROW_HR_IF_MSG(E_UNEXPECTED, (propVariantIcon.vt == VT_EMPTY || propVariantIcon.pwszVal == nullptr), "You must specify a valid image filepath before calling Register().");
+        THROW_HR_IF_MSG(E_UNEXPECTED, (propVariantIcon.vt == VT_EMPTY || propVariantIcon.pwszVal == nullptr), "Icon is not specified");
 
-        THROW_HR_IF_MSG(E_UNEXPECTED, propVariantIcon.vt != VT_LPWSTR, "Provide a valid zero-terminated string of UNICODE characters");
+        THROW_HR_IF_MSG(E_UNEXPECTED, propVariantIcon.vt != VT_LPWSTR, "Icon should be a valid Unicode string");
 
-        THROW_HR_IF_MSG(E_UNEXPECTED, wcslen(propVariantIcon.pwszVal) == 0, "Provide a non empty icon string");
+        THROW_HR_IF_MSG(E_UNEXPECTED, wcslen(propVariantIcon.pwszVal) == 0, "Icon is an empty string");
 
         iconFilePath = propVariantIcon.pwszVal;
 
--- a/dev/AppNotifications/AppNotificationUtility.cpp
+++ b/dev/AppNotifications/AppNotificationUtility.cpp
@@ -215,7 +215,7 @@ HRESULT Microsoft::Windows::AppNotifications::Helpers::GetActivatorGuid(std::wst
 }
 CATCH_RETURN()
 
-std::wstring Microsoft::Windows::AppNotifications::Helpers::SetDisplayName()
+std::wstring Microsoft::Windows::AppNotifications::Helpers::SetDisplayNameBasedOnProcessName()
 {
     std::wstring displayName{};
     THROW_IF_FAILED(wil::GetModuleFileNameExW(GetCurrentProcess(), nullptr, displayName));
@@ -246,11 +246,6 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring 
         &hKey,
         nullptr /* lpdwDisposition */));
 
-    // Retrieve DisplayName and IconUri
-    // - DisplayName: Retrieve from Shell. If not specified, fall back to filename.
-    // - Icon: Retrieve from Shell. If it's not the case or the file extension is unsupported, then throw.
-    winrt::com_ptr<IPropertyStore> propertyStore;
-
     // Retrieve the display name
     std::wstring displayName{};
 
@@ -261,6 +256,10 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring 
 
     if (hWindow)
     {
+        // Retrieve DisplayName and IconUri
+        // - DisplayName: Retrieve from Shell. If not specified, fall back to filename.
+        // - Icon: Retrieve from Shell. If it's not the case or the file extension is unsupported, then throw.
+        winrt::com_ptr<IPropertyStore> propertyStore;
         THROW_IF_FAILED(SHGetPropertyStoreForWindow(hWindow.get(), IID_PPV_ARGS(propertyStore.put())));
 
         wil::unique_prop_variant propVariantDisplayName;
@@ -270,37 +269,36 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring 
         {
             displayName = propVariantDisplayName.pwszVal;
         }
-        else if (displayName.length() == 0)
+        else if (displayName.empty())
         {
-            displayName = SetDisplayName();
+            displayName = SetDisplayNameBasedOnProcessName();
         }
 
         wil::unique_prop_variant propVariantIcon;
-        // Throw in case of failure, since Icon is mandatory and we don't have a fallback!
         THROW_IF_FAILED(propertyStore->GetValue(PKEY_AppUserModel_RelaunchIconResource, &propVariantIcon));
-        THROW_HR_IF_MSG(E_UNEXPECTED, propVariantIcon.vt == VT_EMPTY, "You must specify an app icon before calling Register().");
+        THROW_HR_IF_MSG(E_UNEXPECTED, propVariantIcon.vt == VT_EMPTY, "You must specify a valid image filepath before calling Register().");
 
         iconFilePath = propVariantIcon.pwszVal;
 
         // Icon filepaths from Shell APIs have this format: <filepath>,-<index>,
         // since .ico files can have multiple icons in the same file.
         // NotificationController doesn't seem to support such format, so let it take the first icon by default.
-        auto iteratorForCommaDelimiter = iconFilePath.find_first_of(L",");
+        auto iteratorForCommaDelimiter{ iconFilePath.find_first_of(L",") };
         if (iteratorForCommaDelimiter != std::wstring::npos) // It may or may not have an index, which is fine.
         {
             iconFilePath.erase(iteratorForCommaDelimiter);
         }
 
-        auto iteratorForFileExtension = iconFilePath.find_first_of(L".");
+        auto iteratorForFileExtension{ iconFilePath.find_first_of(L".") };
         THROW_HR_IF_MSG(E_UNEXPECTED, iteratorForFileExtension == std::wstring::npos, "You must provide a valid filepath as the app icon.");
 
-        std::wstring iconFileExtension = iconFilePath.substr(iteratorForFileExtension);
+        std::wstring iconFileExtension{ iconFilePath.substr(iteratorForFileExtension) };
         THROW_HR_IF_MSG(E_UNEXPECTED, iconFileExtension != L".ico" && iconFileExtension != L".png",
             "You must provide a supported file extension as the icon (.ico or .png).");
     }
     else
     {
-        displayName = SetDisplayName();
+        displayName = SetDisplayNameBasedOnProcessName();
     }
 
     RegisterValue(hKey, L"DisplayName", reinterpret_cast<const BYTE*>(displayName.c_str()), REG_EXPAND_SZ, displayName.size() * sizeof(wchar_t));
--- a/dev/AppNotifications/AppNotificationUtility.cpp
+++ b/dev/AppNotifications/AppNotificationUtility.cpp
@@ -3,6 +3,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 #include "pch.h"
+#include <cwctype>
 #include "AppNotificationUtility.h"
 #include <winrt/Windows.ApplicationModel.Core.h>
 #include <winrt/Windows.Foundation.h>
@@ -226,6 +227,8 @@ std::wstring Microsoft::Windows::AppNotifications::Helpers::SetDisplayNameBasedO
 
     displayName.erase(displayName.find_first_of(L".")); // Remove file extension
 
+    displayName[0] = std::towupper(displayName[0]);
+
     return displayName;
 }
 
@@ -248,6 +251,7 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring 
 
     // Retrieve the display name
     std::wstring displayName{};
+    displayName = SetDisplayNameBasedOnProcessName();
 
     // Retrieve the icon
     std::wstring iconFilePath{};
@@ -264,14 +268,10 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring 
 
         wil::unique_prop_variant propVariantDisplayName;
         // Do not throw in case of failure. We can use the filepath approach below as fallback to set a DisplayName.
-        LOG_IF_FAILED(propertyStore->GetValue(PKEY_AppUserModel_RelaunchDisplayNameResource, &propVariantDisplayName));
+        THROW_IF_FAILED(propertyStore->GetValue(PKEY_AppUserModel_RelaunchDisplayNameResource, &propVariantDisplayName));
         if (propVariantDisplayName.vt == VT_LPWSTR && propVariantDisplayName.pwszVal != nullptr)
         {
             displayName = propVariantDisplayName.pwszVal;
-        }
-        else if (displayName.empty())
-        {
-            displayName = SetDisplayNameBasedOnProcessName();
         }
 
         wil::unique_prop_variant propVariantIcon;
@@ -295,10 +295,6 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring 
         std::wstring iconFileExtension{ iconFilePath.substr(iteratorForFileExtension) };
         THROW_HR_IF_MSG(E_UNEXPECTED, iconFileExtension != L".ico" && iconFileExtension != L".png",
             "You must provide a supported file extension as the icon (.ico or .png).");
-    }
-    else
-    {
-        displayName = SetDisplayNameBasedOnProcessName();
     }
 
     RegisterValue(hKey, L"DisplayName", reinterpret_cast<const BYTE*>(displayName.c_str()), REG_EXPAND_SZ, displayName.size() * sizeof(wchar_t));
--- a/dev/AppNotifications/AppNotificationUtility.cpp
+++ b/dev/AppNotifications/AppNotificationUtility.cpp
@@ -215,6 +215,20 @@ HRESULT Microsoft::Windows::AppNotifications::Helpers::GetActivatorGuid(std::wst
 }
 CATCH_RETURN()
 
+std::wstring Microsoft::Windows::AppNotifications::Helpers::SetDisplayName()
+{
+    std::wstring displayName{};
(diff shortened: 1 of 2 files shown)


dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -236,57 +250,59 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
     // - DisplayName: Retrieve from Shell. If not specified, fall back to filename.
     // - Icon: Retrieve from Shell. If it's not the case or the file extension is unsupported, then throw.
     winrt::com_ptr<IPropertyStore> propertyStore;
```
pmpurifoy, line 252: Nit: Move closer to usage

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -236,57 +250,59 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
...
-        THROW_HR_IF(E_UNEXPECTED, lastBackslashPosition == std::wstring::npos);
-        displayName = displayName.substr(lastBackslashPosition + 1); // One after the delimiter
+        wil::unique_prop_variant propVariantDisplayName;
+        // Do not throw in case of failure. We can use the filepath approach below as fallback to set a DisplayName.
+        LOG_IF_FAILED(propertyStore->GetValue(PKEY_AppUserModel_RelaunchDisplayNameResource, &propVariantDisplayName));
+        if (propVariantDisplayName.vt == VT_LPWSTR && propVariantDisplayName.pwszVal != nullptr)
+        {
+            displayName = propVariantDisplayName.pwszVal;
+        }
+        else if (displayName.length() == 0)
```
pmpurifoy, line 273: Use displayName.empty() instead.

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -236,57 +250,59 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
...
+        iconFilePath = propVariantIcon.pwszVal;
 
-    wil::unique_prop_variant propVariantIcon;
-    // Throw in case of failure, since Icon is mandatory and we don't have a fallback!
-    THROW_IF_FAILED(propertyStore->GetValue(PKEY_AppUserModel_RelaunchIconResource, &propVariantIcon));
-    THROW_HR_IF_MSG(E_UNEXPECTED, propVariantIcon.vt == VT_EMPTY, "You must specify an app icon before calling Register().");
+        // Icon filepaths from Shell APIs have this format: <filepath>,-<index>,
+        // since .ico files can have multiple icons in the same file.
+        // NotificationController doesn't seem to support such format, so let it take the first icon by default.
+        auto iteratorForCommaDelimiter = iconFilePath.find_first_of(L",");
```
pmpurifoy, line 288: Nit: {}-init

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -236,57 +250,59 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
...
+        if (iteratorForCommaDelimiter != std::wstring::npos) // It may or may not have an index, which is fine.
+        {
+            iconFilePath.erase(iteratorForCommaDelimiter);
+        }
 
-    iconFilePath = propVariantIcon.pwszVal;
+        auto iteratorForFileExtension = iconFilePath.find_first_of(L".");
```
pmpurifoy, line 294: Nit: {}-init

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -236,57 +250,59 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
...
+        THROW_HR_IF_MSG(E_UNEXPECTED, iteratorForFileExtension == std::wstring::npos, "You must provide a valid filepath as the app icon.");
 
-    // Icon filepaths from Shell APIs have this format: <filepath>,-<index>,
-    // since .ico files can have multiple icons in the same file.
-    // NotificationController doesn't seem to support such format, so let it take the first icon by default.
-    auto iteratorForCommaDelimiter = iconFilePath.find_first_of(L",");
-    if (iteratorForCommaDelimiter != std::wstring::npos) // It may or may not have an index, which is fine.
+        std::wstring iconFileExtension = iconFilePath.substr(iteratorForFileExtension);
```
pmpurifoy, line 297: Nit: {}-init Last one :)

pmpurifoy (approved):

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -236,57 +250,59 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
...
+        }
+        else if (displayName.length() == 0)
+        {
+            displayName = SetDisplayName();
+        }
 
-        displayName.erase(displayName.find_first_of(L".")); // Remove file extension
-    }
+        wil::unique_prop_variant propVariantIcon;
+        // Throw in case of failure, since Icon is mandatory and we don't have a fallback!
```
danielayala94, line 279: This comment is stale now, please remove.

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -236,57 +250,59 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
...
+        THROW_IF_FAILED(propertyStore->GetValue(PKEY_AppUserModel_RelaunchIconResource, &propVariantIcon));
+        THROW_HR_IF_MSG(E_UNEXPECTED, propVariantIcon.vt == VT_EMPTY, "You must specify an app icon before calling Register().");
```
danielayala94, line 281:
```suggestion
        THROW_HR_IF_MSG(E_UNEXPECTED, propVariantIcon.vt == VT_EMPTY, "You must specify a valid image filepath before calling Register().");
```

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -215,6 +215,20 @@ HRESULT Microsoft::Windows::AppNotifications::Helpers::GetActivatorGuid(std::wst
 }
 CATCH_RETURN()
 
+std::wstring Microsoft::Windows::AppNotifications::Helpers::SetDisplayName()
```
danielayala94, line 218: SetDisplayNameBasedOnProcessName

danielayala94 (approved):

danielayala94 (approved):

pmpurifoy (approved):

danielayala94 (approved):

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -215,6 +215,20 @@ HRESULT Microsoft::Windows::AppNotifications::Helpers::GetActivatorGuid(std::wst
...
+    size_t lastBackslashPosition{ displayName.rfind(L"\\") };
+    THROW_HR_IF(E_UNEXPECTED, lastBackslashPosition == std::wstring::npos);
+    displayName = displayName.substr(lastBackslashPosition + 1); // One after the delimiter
+
+    displayName.erase(displayName.find_first_of(L".")); // Remove file extension
+
+    return displayName;
+}
+
 void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring const& appId, std::wstring const& clsid)
```
hulumane, line 232: This change is too big. REcommend we chat quickly.

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -232,61 +246,61 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
...
-    auto iteratorForCommaDelimiter = iconFilePath.find_first_of(L",");
-    if (iteratorForCommaDelimiter != std::wstring::npos) // It may or may not have an index, which is fine.
+        std::wstring iconFileExtension{ iconFilePath.substr(iteratorForFileExtension) };
+        THROW_HR_IF_MSG(E_UNEXPECTED, iconFileExtension != L".ico" && iconFileExtension != L".png",
+            "You must provide a supported file extension as the icon (.ico or .png).");
+    }
+    else
     {
-        iconFilePath.erase(iteratorForCommaDelimiter);
+        displayName = SetDisplayNameBasedOnProcessName();
```
hulumane, line 301: Bug: Set the displayname from processName at the beginning itself and remove this else clause.

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -232,61 +246,61 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
...
+    {
+        // Retrieve DisplayName and IconUri
+        // - DisplayName: Retrieve from Shell. If not specified, fall back to filename.
+        // - Icon: Retrieve from Shell. If it's not the case or the file extension is unsupported, then throw.
+        winrt::com_ptr<IPropertyStore> propertyStore;
+        THROW_IF_FAILED(SHGetPropertyStoreForWindow(hWindow.get(), IID_PPV_ARGS(propertyStore.put())));
+
+        wil::unique_prop_variant propVariantDisplayName;
+        // Do not throw in case of failure. We can use the filepath approach below as fallback to set a DisplayName.
+        LOG_IF_FAILED(propertyStore->GetValue(PKEY_AppUserModel_RelaunchDisplayNameResource, &propVariantDisplayName));
```
hulumane, line 267: THROW_IF_FAILED

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -232,61 +246,61 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
...
+        if (propVariantDisplayName.vt == VT_LPWSTR && propVariantDisplayName.pwszVal != nullptr)
+        {
+            displayName = propVariantDisplayName.pwszVal;
+        }
+        else if (displayName.empty())
+        {
+            displayName = SetDisplayNameBasedOnProcessName();
```
hulumane, line 274: Overcomplicaed. Set this in the beginning before the if clause. This is the baseline.

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -215,6 +215,20 @@ HRESULT Microsoft::Windows::AppNotifications::Helpers::GetActivatorGuid(std::wst
...
+{
+    std::wstring displayName{};
+    THROW_IF_FAILED(wil::GetModuleFileNameExW(GetCurrentProcess(), nullptr, displayName));
+
+    size_t lastBackslashPosition{ displayName.rfind(L"\\") };
+    THROW_HR_IF(E_UNEXPECTED, lastBackslashPosition == std::wstring::npos);
+    displayName = displayName.substr(lastBackslashPosition + 1); // One after the delimiter
+
+    displayName.erase(displayName.find_first_of(L".")); // Remove file extension
+
```
hulumane, line 228: Make sure that the display name starts with caps: For example Tiktok and not tiktok.

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -232,61 +249,54 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
...
+        // - Icon: Retrieve from Shell. If it's not the case or the file extension is unsupported, then throw.
+        winrt::com_ptr<IPropertyStore> propertyStore;
+        THROW_IF_FAILED(SHGetPropertyStoreForWindow(hWindow.get(), IID_PPV_ARGS(propertyStore.put())));
+
+        wil::unique_prop_variant propVariantDisplayName;
+        // Do not throw in case of failure. We can use the filepath approach below as fallback to set a DisplayName.
+        THROW_IF_FAILED(propertyStore->GetValue(PKEY_AppUserModel_RelaunchDisplayNameResource, &propVariantDisplayName));
+        if (propVariantDisplayName.vt == VT_LPWSTR && propVariantDisplayName.pwszVal != nullptr)
+        {
+            displayName = propVariantDisplayName.pwszVal;
```
hulumane, line 274: check the length of pwszval to make sure it isnt zero else throw e_unexpected

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -232,61 +249,54 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
...
+        // Retrieve DisplayName and IconUri
+        // - DisplayName: Retrieve from Shell. If not specified, fall back to filename.
+        // - Icon: Retrieve from Shell. If it's not the case or the file extension is unsupported, then throw.
+        winrt::com_ptr<IPropertyStore> propertyStore;
+        THROW_IF_FAILED(SHGetPropertyStoreForWindow(hWindow.get(), IID_PPV_ARGS(propertyStore.put())));
+
+        wil::unique_prop_variant propVariantDisplayName;
+        // Do not throw in case of failure. We can use the filepath approach below as fallback to set a DisplayName.
+        THROW_IF_FAILED(propertyStore->GetValue(PKEY_AppUserModel_RelaunchDisplayNameResource, &propVariantDisplayName));
+        if (propVariantDisplayName.vt == VT_LPWSTR && propVariantDisplayName.pwszVal != nullptr)
```
hulumane, line 272: Please check out the documentation for propvariant->Get()
https://docs.microsoft.com/en-us/previous-versions/windows/desktop/legacy/bb761473(v=vs.85)

You always check for VT_EMPTY after GetValue succeeds

hulumane: Not that displayName check for VT_EMPTY is best effort. If it isn't present, then we default to processname

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -232,61 +249,54 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
...
-    LOG_IF_FAILED(propertyStore->GetValue(PKEY_AppUserModel_RelaunchDisplayNameResource, &propVariantDisplayName));
-    if (propVariantDisplayName.vt == VT_LPWSTR && propVariantDisplayName.pwszVal != nullptr)
-    {
-        displayName = propVariantDisplayName.pwszVal;
-    }
+    // Retrieve the icon
+    std::wstring iconFilePath{};
 
-    if (displayName.length() == 0)
+    wil::unique_hwnd hWindow{ GetConsoleWindow() };
```
hulumane, line 259: This doesn't throw right if ConSoleWindow is not present? That would be really bad.

sharath2727: https://docs.microsoft.com/en-us/windows/console/getconsolewindow - doesn't say anything about throwing. It returns null if there is no handle.

hulumane: Documentation indicates that it would return null: https://docs.microsoft.com/en-us/windows/console/getconsolewindow

hulumane: Ok looks like our responses crossed. This should be good then.

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -232,61 +249,54 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
...
-        size_t lastBackslashPosition{ displayName.rfind(L"\\") };
-        THROW_HR_IF(E_UNEXPECTED, lastBackslashPosition == std::wstring::npos);
-        displayName = displayName.substr(lastBackslashPosition + 1); // One after the delimiter
+        wil::unique_prop_variant propVariantIcon;
+        THROW_IF_FAILED(propertyStore->GetValue(PKEY_AppUserModel_RelaunchIconResource, &propVariantIcon));
+        THROW_HR_IF_MSG(E_UNEXPECTED, propVariantIcon.vt == VT_EMPTY, "You must specify a valid image filepath before calling Register().");
 
-        displayName.erase(displayName.find_first_of(L".")); // Remove file extension
-    }
+        iconFilePath = propVariantIcon.pwszVal;
```
hulumane, line 281: Inconsistency in pattern: You check for this above but not here:
propVariantDisplayName.vt == VT_LPWSTR && propVariantDisplayName.pwszVal != nullptr

dev/AppNotifications/AppNotificationUtility.cpp
sharath2727: pavan why check for VT_EMPTY? We know in our success case it has to be VT_LPWSTR if vt isn't set to VT_LPWSTR we just default to filepath name. So I think we don't need any checks here I guess. Let me know.

dev/AppNotifications/AppNotificationUtility.cpp
danielayala94: It can be either VT_LPWSTR or VT_EMPTY. The current check is sufficient.

dev/AppNotifications/AppNotificationUtility.cpp
sharath2727: @danielayala94 why just check for VT_EMPTY in filepath we should be doing this instead:

 THROW_HR_IF_MSG(E_UNEXPECTED, propVariantDisplayName.vt != VT_LPWSTR", You must specify a valid image filepath before calling Register().");

Let me know what you think?

dev/AppNotifications/AppNotificationUtility.cpp
danielayala94: Setting an empty VT_LPWSTR results in getting a VT_EMPTY. I think such check will be redundant.

sharath2727: Yeah the only thing here is, that the document isn't clear that if user set an empty string does it return a VT_EMPTY. It only says that:

If the [PROPERTYKEY](https://msdn.microsoft.com/en-us/library/bb773381(v=vs.85)) referenced in key is not present in the property store, this method returns S_OK and the vt member of the structure pointed to by pv is set to VT_EMPTY."

I think pavan might be right here that: propVariantDisplayName.vt == VT_LPWSTR && propVariantDisplayName.pwszVal != nullptr && propVariantDisplayName.pwszVal != L""

@danielayala94 any thoughts????

dev/AppNotifications/AppNotificationUtility.cpp
hulumane: Oh I thought it was a flag?

dev/AppNotifications/AppNotificationUtility.cpp
pmpurifoy: The only way the app could reach this is if they explicitly set the propVariantDisplayName.vt to VT_LPWSTR. Unless I am missing something like default values for hWindow property store, the user is trying to set the property store. 

If in this flow they decide to use L"" for their displayName, I think it is a valid case. It seems intentional since pwszVal is nullptr by default and if they set it to empty string, well, that's what they want.

But is it really what they want? I don't know. :) but my thoughts are allowing the developer to do this since it looks like a conscious decision.

dev/AppNotifications/AppNotificationUtility.cpp
danielayala94: It is mostly a code saying what is inside the propVariant:
https://docs.microsoft.com/en-us/windows/win32/api/propidlbase/ns-propidlbase-propvariant?redirectedfrom=MSDN

Mostly, because there are a few flags indeed, but VT_EMPTY and VT_LPWSTR are mutually exclusive, for instance.

dev/AppNotifications/AppNotificationUtility.cpp
danielayala94: I tried setting an empty VT_LPWSTR in a dummy Win32 app. That's how I came up with my statement above.

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -267,16 +267,22 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
...
             displayName = propVariantDisplayName.pwszVal;
         }
 
         wil::unique_prop_variant propVariantIcon;
         THROW_IF_FAILED(propertyStore->GetValue(PKEY_AppUserModel_RelaunchIconResource, &propVariantIcon));
-        THROW_HR_IF_MSG(E_UNEXPECTED, (propVariantIcon.vt != VT_LPWSTR || propVariantIcon.pwszVal == nullptr), "You must specify a valid image filepath before calling Register().");
+
+        THROW_HR_IF_MSG(E_UNEXPECTED, (propVariantIcon.vt == VT_EMPTY || propVariantIcon.pwszVal == nullptr), "You must specify a valid image filepath before calling Register().");
+
+        THROW_HR_IF_MSG(E_UNEXPECTED, propVariantIcon.vt != VT_LPWSTR, "Provide a valid zero-terminated string of UNICODE characters");
```
hulumane, line 283: Msg:Icon should be a valid Uncode string

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -267,16 +267,22 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
...
+        if (propVariantDisplayName.vt == VT_LPWSTR && propVariantDisplayName.pwszVal != nullptr && wcslen(propVariantDisplayName.pwszVal) > 0)
         {
             displayName = propVariantDisplayName.pwszVal;
         }
 
         wil::unique_prop_variant propVariantIcon;
         THROW_IF_FAILED(propertyStore->GetValue(PKEY_AppUserModel_RelaunchIconResource, &propVariantIcon));
-        THROW_HR_IF_MSG(E_UNEXPECTED, (propVariantIcon.vt != VT_LPWSTR || propVariantIcon.pwszVal == nullptr), "You must specify a valid image filepath before calling Register().");
+
+        THROW_HR_IF_MSG(E_UNEXPECTED, (propVariantIcon.vt == VT_EMPTY || propVariantIcon.pwszVal == nullptr), "You must specify a valid image filepath before calling Register().");
```
hulumane, line 281: Icon is not specified!

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -267,16 +267,22 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
...
+
+        THROW_HR_IF_MSG(E_UNEXPECTED, propVariantIcon.vt != VT_LPWSTR, "Provide a valid zero-terminated string of UNICODE characters");
+
+        THROW_HR_IF_MSG(E_UNEXPECTED, wcslen(propVariantIcon.pwszVal) == 0, "Provide a non empty icon string");
```
hulumane, line 285: Icon is an empty string!

hulumane (approved):

PR #2245, opened 2022-03-11
Revert try_create use as it breaks single instancing.

jlaanstra:
This PR reverts a suggestion made in #2211 to use try_create. try_create returns true even if the named mutex already exists since it calls CreateMutexExW and so we have to manually check GetLastError(). This behaviour is documented at: https://docs.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-createmutexexw


jonwis (approved):

riverar (approved):

jlaanstra:
> @jlaanstra I take the blame for that suggestion, missed that pitfall! 😢

No worries. I learned something new and the wil is getting fixed.

PR #2249, opened 2022-03-12
Add API contract version to Deployment interface

sachintaMSFT:
Added API contract version to Deployment interface

### DIFF
--- a/specs/Deployment/DeploymentAPI.md
+++ b/specs/Deployment/DeploymentAPI.md
@@ -184,7 +184,11 @@ they shut down, to refer to the updated framework package.
 ```C# (but really MIDL3)
 namespace Microsoft.Windows.ApplicationModel.WindowsAppRuntime
 {
+    [contractversion(2)]
+    apicontract Deployment{};
+
     /// Represents the current Deployment status of the WindowsAppRuntime
+    [contract(Deployment, 1)]
     enum DeploymentStatus
     {
         Unknown = 0,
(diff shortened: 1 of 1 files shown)


specs/Deployment/DeploymentAPI.md
```diff
@@ -184,7 +184,11 @@ they shut down, to refer to the updated framework package.
 ```C# (but really MIDL3)
 {
+    [contractversion(2)]
+    apicontract Deployment{};
```
DrusTheAxe, line 188: `DeploymentContract`

Contract names end with `Contract`

DrusTheAxe (commented):
LGTM. We'll need to do an API Review

@jonwis perhaps we should do 1 all-up review for the repo? Most WinRT APIs are v1 from 1.0, a small % are adding WinRT APIs in 1.1. All existing APIs will be tagged with a v1 contract, but their contract names should be reviewed. How do you feel about 1 list of all the apicontracts with the namespaces they're applied to? Should expedite matters

specs/Deployment/DeploymentAPI.md
sachintaMSFT: @DrusTheAxe, would DynamicDependency contract name also need to be updated likewise - https://github.com/microsoft/WindowsAppSDK/blob/main/dev/DynamicDependency/API/M.AM.DynamicDependency.idl ?

DrusTheAxe: Yes. At a minimum it'll need to change from `DynamicDependency` to `DynamicDependencyContract` comply with [specs/WinRT/WinRTAPIContracts.md](https://github.com/microsoft/WindowsAppSDK/pull/2293) section `2.5 Contract naming`

I wouldn't worry about code right now. For anything checked in let it sit. For anything pending let it wait (if it's just adding contracts). Once we sort out #2293 and get API Review signoff we can update code.

DrusTheAxe (approved):

PR #2260, opened 2022-03-14
User/sachinta/add microsoft internal telemetry nuget package config

sachintaMSFT:
* Add microsoft internal telemetry nuget package config
* Remove unnecessary inclusion of microsofttelemetry.h as <wil/traceloggingconfig.h> provides the same stub functionality from public WIL nuget package
* Update search for WindowsAppSDK-VersionInfo.h header includes current directory as the build pipeline is updated to copy the file to installer project folder for it's inclusion.

Testing:
These changes have been tested by verifying the installer from the build pipeline is generating the telemetry events as expected.

NOTE: These changes will need to be pulled into 1.1-preview1 branch after it's merged to main.


installer/dev/tracelogging.h
```diff
@@ -4,10 +4,9 @@
 #pragma once
 
 #include "pch.h"
-#include "MicrosoftTelemetry.h"
 
-#if __has_include(<WindowsAppSDK-VersionInfo.h>)
-#include <WindowsAppSDK-VersionInfo.h>
+#if __has_include("WindowsAppSDK-VersionInfo.h")
```
DrusTheAxe, line 8: Why "" instead of <>?

"" means "first look relative to current/project directory" but this file lives in the root, not under root\installer, right? Found via $(AdditionalIncludeDirectories)

DrusTheAxe (approved):
LGTM modulo comments

installer/dev/tracelogging.h
sachintaMSFT: No. CreateInstaller stage doesn't share the same file system as the Aggregates. Hence, I am having to extract the WindowsAppSDK  nuget package build artifact from Aggregates stage and copy the version info header file to current directory.

PR #2267, opened 2022-03-15
Add Selfcontained checks and misc fixes for race condition in AppNoti…

hulumane:
[(https://user-images.githubusercontent.com/26147119/157732742-8d74ffb6-21ad-4a24-919a-9b525a93dd68.png)Issue:
Self-Contained apps may not support notifications in some cases. Please refer to the matrix below:
Screenshot 2022-03-10 104221

In general, All local notification scenarios will be supported in self-contained mode. We will not support a subset of Push scenarios for Self-Contained apps.

Fix:
For local notifications, we gate the calls to the Long Running Singleton via a Self-Contained checks so that API calls don't blow up.

For Push notifications, we introduce an IsSupported gatekeeping API that the developer could use to infer Push functionality.

Also added some checks to prevent races in AppNotificationManager::Register() APIs.


dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -95,6 +96,18 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
```
jonwis, line 96: Does the public WIL Tracelogging have "Activity" support? In that, you'd say:

```c++
auto registerActivity = AppNotificationTelemetry::RegisterStart(m_appId);
```

... and then the activity manages the start/stop.

hulumane: @loneursid Do you know? Added this telemetry point recently. Not aware if this support exists publicly.

dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
```diff
@@ -65,9 +65,10 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
         return { payload, payload + (length * sizeof(uint8_t)) };
     }
 
-    std::vector<uint8_t> PushNotificationReceivedEventArgs::BuildPayload(std::wstring const& payload)
+    std::vector<uint8_t> PushNotificationReceivedEventArgs::BuildPayload(winrt::hstring const& payload)
     {
-        std::string payloadToSimpleString{ ::winrt::Microsoft::Windows::PushNotifications::Helpers::WideStringToUtf8String(payload) };
+        const std::wstring payloadWideString(payload);
```
jonwis, line 70: >const std::wstring payloadWideString(payload) [](http://example.com/codeflow?start=8&length=45)

Maybe WideStringToUtf8String should accept std::wstring_view, which both std::wstring and winrt::hstring convert to.

jonwis (approved):
:shipit:

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -170,10 +193,31 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
+                Unregister();
+            }
+
+            {
+                auto lock{ m_lock.lock_exclusive() };
+                THROW_HR_IF_MSG(HRESULT_FROM_WIN32(ERROR_OPERATION_IN_PROGRESS), m_registering, "Register or Unregister currently in progress!");
+                m_registering = true;
+            }
+
+            auto scope_exit = wil::scope_exit(
```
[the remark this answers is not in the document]
hulumane: m_registering is always true when Register(), Unregister() or UnregisterAll() is being called by the API developer.

If this API is called twice for some reason back to back, one of the API calls is expected to fail due to the check above.

dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
hulumane: Thats a good point. Let me give that a shot.

hulumane: It turns out that there was just one other reference to this API. I went ahead and changed it to accept hstring only and avoid all conversions altogether. Its much better now!

dev/AppNotifications/AppNotificationManager.cpp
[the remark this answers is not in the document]
hulumane: There is a small race with Unregister where it takes a dependency on the same m_registering condition. I have refactored to use an UnregisterHelper() whichshould always be called in the scope of the exclusive lock.

UnregisterHelper is called from both Unregister as well as UnregisterAll. If any of these operations are already in progress, the API call is guaranteed to fail with ERROR_OPERATION_IN_PROGRESS (code is structured that way).

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -95,6 +96,18 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
+            {
+                auto lock{ m_lock.lock_exclusive() };
+                THROW_HR_IF_MSG(HRESULT_FROM_WIN32(ERROR_OPERATION_IN_PROGRESS), m_registering, "Registration is in progress!");
+                m_registering = true;
+            }
+
+            auto registeringScopeExit{ wil::scope_exit([&]()
+            {
+                auto lock { m_lock.lock_exclusive() };
+                m_registering = false;
```
pmpurifoy, line 108: Should we be worried about unregistering on failures?

If a notificationPlatform call fails, registry will still have all the entries (I believe we have a failsafe for this). If AddToastRegistrationMapping succeeds and we throw later, the entry will be in the Long Running Process. I'm not super worried about this since the Long Running Process has its own cleanup functions, but how bad is it really? 
And we have called PushNotification_RegisterFullTrustApplication. Do we want to unregister if anything in Register() fails?

Something to think about.

hulumane: Yeah i think we can go all the way for cleanup. But at the end of the day, all this is a best effort. If Register fails, there is a good chance that Unregister fails as well (that typically happens if the plaform is in a bad state or has crashed.). So I don't believe we have a perfect solution.

hulumane: Talked offline. We will file a bug to do best effort cleanup similar to PushManager::Register @pmpurifoy

pmpurifoy: https://github.com/microsoft/WindowsAppSDK/issues/2269

pmpurifoy (approved):

PR #2268, opened 2022-03-15
Fix bug in GetExeNameFromFilePath

pmpurifoy:
There is a case where the exe is not in a filepath in the registry but is a standalone exe. This fixes that scenario.

### DIFF
--- a/dev/PushNotifications/PushNotificationUtility.h
+++ b/dev/PushNotifications/PushNotificationUtility.h
@@ -158,8 +158,14 @@ namespace winrt::Microsoft::Windows::PushNotifications::Helpers
     inline std::wstring GetExeNameFromPath(std::wstring const& exePath)
     {
         size_t pos{ exePath.rfind(L"\\") };
-        THROW_HR_IF(E_UNEXPECTED, pos == std::wstring::npos);
-        return exePath.substr(pos + 1); // One after the delimiter
+        if (pos != std::wstring::npos)
+        {
+            return exePath.substr(pos + 1); // One after the delimiter
+        }
+        else
+        {
+            return exePath;
+        }
     }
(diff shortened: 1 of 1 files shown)


jonwis (approved):

dev/PushNotifications/PushNotificationUtility.h
```diff
@@ -158,8 +158,14 @@ namespace winrt::Microsoft::Windows::PushNotifications::Helpers
     inline std::wstring GetExeNameFromPath(std::wstring const& exePath)
```
hulumane, line 158: When were we sending a relative path as opposed to absolute path? Just curious :)

pmpurifoy: For unpackaged, we send absolute path. For packaged, it's assigned in the manifest which for MSIX existed in a subfolder, but for Self-contained it was in the same BuildOutput.

dev/PushNotifications/PushNotificationUtility.h
```diff
@@ -158,8 +158,14 @@ namespace winrt::Microsoft::Windows::PushNotifications::Helpers
...
     {
         size_t pos{ exePath.rfind(L"\\") };
-        THROW_HR_IF(E_UNEXPECTED, pos == std::wstring::npos);
-        return exePath.substr(pos + 1); // One after the delimiter
+        if (pos != std::wstring::npos)
```
DrusTheAxe, line 161: Why not

```
return std::filesystem::path(exePath).filename();
```

pmpurifoy: The main reason is because I forgot that there is a std::filesystem library...

PR #2270, opened 2022-03-15
Update Windows App SDK Roadmap

gabbybilka:

Issue #58, opened 2020-06-02
Proposal: Address Project Reunion goals by addressing Visual Studio restrictions 


Scottj1s:
# Proposal: Address Project Reunion goals by addressing Visual Studio restrictions 

## Summary
Many of the barriers to code sharing and interop that Project Reunion seeks to remove are either active restrictions, or passive limitations, of Visual Studio.  These have been surfaced with the release of new WinRT projections like C++/WinRT and C#/WinRT.  

## Rationale
Project Reunion is likely to highlight these Visual Studio restrictions even more, and so should address their elimination.  These restrictions fall into build time and run time categories, detailed below.  This proposal is something of a grab bag of known issues, and seeks to prioritize their resolution so that the end to end developer experience is more streamlined.  

#### C++ PackageReferences
C++ projects (*.vcxproj files) do not support Nuget PackageReferences, but only packages.config references, which have a number of limitations:
- They must target a specific package version, even with allowedVersions
- They don't participate with the project system, so can't import other artifacts or use replaceable parameters
- They don't support 'msbuild /t:restore', so a separate 'nuget restore' is necessary in build pipelines, etc

There is a longstanding open request to add this support:
[Use PackageReference in vcxproj](https://developercommunity.visualstudio.com/idea/351636/use-packagereference-in-vcxproj.html
)
And a related longstanding open PR to implement it:
[add C++ PackageReference support](https://github.com/NuGet/NuGet.Client/pull/3145)

#### Desktop <--> UWP Project References
Project references cannot be added in Visual Studio between Desktop and UWP (Universal Windows) projects.  Attempts to do so result in an error:  [A reference cannot be added because the two projects target different platforms.](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/854753).  This can be worked around with a project file edit, or introduction of a directory.build.* file, to manually add the project reference.  In other words, this is strictly a tooling restriction, which prevents interop unless developers are aware of the workarounds.

#### Unified CRT 
[...]

PR #2273, opened 2022-03-16
Roadmap edits & refinements

gabbybilka:
Edits & refinements to WinAppSDK roadmap based on comments from @riverar, thank you!


gabbybilka:
There was an additional comment from @riverar in the previous PR (#2270) around whether WinAppSDK should be referred to as a 'platform' in this document, line #58. @ryandemopoulos what are your thoughts?

PR #2274, opened 2022-03-16
Use internal nuget feed for develop branch

kythant:
Use internal nuget feed for develop branch

Also updated FrameworkUDK to use the internal version.

WIP or Coming later:
Use version.details.xml to bring in the FrameworkUDK and flow the versions down to the dependencies so we don't have version numbers repeated in different projects.


ionvasilian (approved):
Moving forward, we should figure out a better way to update the well-known package dependencies.

kythant:
> Moving forward, we should figure out a better way to update the well-known package dependencies.

Definitely. This is my next change.

PR #2278, opened 2022-03-16
Changing contract name to be spec compliant.

dhoehna:
Contracts were added while the spec was being fleshed out.  Changing to a spec compliant name.

### DIFF
--- a/dev/EnvironmentManager/API/Microsoft.Windows.System.EnvironmentManager.idl
+++ b/dev/EnvironmentManager/API/Microsoft.Windows.System.EnvironmentManager.idl
@@ -5,6 +5,10 @@
 
 namespace Microsoft.Windows.System
 {
+    [contractversion(1)]
+    apicontract WindowsSystemContract{};
+
+    [contract(WindowsSystemContract, 1)]
     runtimeclass EnvironmentManager
     {
         static EnvironmentManager GetForProcess();
--- a/dev/PowerNotifications/PowerNotifications.idl
+++ b/dev/PowerNotifications/PowerNotifications.idl
@@ -4,10 +4,10 @@
 namespace Microsoft.Windows.System.Power
 {
     [contractversion(2)]
-    apicontract Notifications{};
+    apicontract WindowsSystemPowerContract{};
 
     // Enums duplicated from IPowerManagerStatics
-    [contract(Notifications, 1)]
+    [contract(WindowsSystemPowerContract, 1)]
     enum EnergySaverStatus
     {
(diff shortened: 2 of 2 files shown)


dev/EnvironmentManager/API/Microsoft.Windows.System.EnvironmentManager.idl
```diff
@@ -5,6 +5,10 @@
 
 namespace Microsoft.Windows.System
 {
+    [contractversion(1)]
+    apicontract WindowsSystemContract{};
```
DrusTheAxe, line 9: This is rather broad. I expected `EnvironmentManagerContract`

There was a (tactical/concrete) question about WinAppSDK contract names and API Review. Let me check on that

dhoehna: https://github.com/microsoft/WindowsAppSDK/pull/2110#discussion_r821085536

@jonwis said WindowsSystemContract.

Should the naming convention be "Something that describes the type", or "Namespace"?

DefaultRyan: I thought the naming convention was by "feature" or "feature set". Namespace is redundant, because the contract is already in a namespace, so this contract is currently named:
``` Microsoft.Windows.System.WindowsSystemContract ```

That's no good at all. I'm putting in another vote for `EnvironmentManagerContract`.

dhoehna: The only item in the spec that talks about contract naming is the rule "Contract names need to end with the word `Contract`."

I like `EnvironmentManagerContract` as well.  Looks like there needs to be a talk about this.

jonwis: Contracts are units of versioning.  I'm looking for a squishy middle here - `UniversalApiContract` is much too large. `FooTypeContract` is much too small.  Other discussions suggest that a contract of 1-2 types is not interesting.

I think I'd go with `WindowsSystemContract` for now.  Ideally we get one contract per "block of functionality" ... I know that's not a satisfying answer.

bpulliam: @dhoehna Is this still applicable?

bpulliam:
Closing PRs that appear to have been abandoned.

PR #2279, opened 2022-03-17
Add ForceDeploy option to Installer and to Deployment API via the new DeploymentInitializeOptions runtime class

sachintaMSFT:
1. Add DeploymentInitializeOptions runtime class (containing ForceDeployment Boolean), DeploymentManager Initialize method overload to take DeploymentInitializeOptions object as parameter and add API Contract to Deployment IDL.
2. Add -f command line option and implement ForceDeployment functionality in WindowsAppRuntimeInstall.exe.
3. Add -f command line option and implement ForceDeployment functionality in WindowsAppRuntime_MSIXInstallFromPath.exe.
4. In all 3 cases above, add attempt to restart push notifications LRP when ForceDeployment is specified.


dev/Deployment/DeploymentManager.cpp
```diff
@@ -8,8 +8,12 @@
 #include <TerminalVelocityFeatures-DeploymentAPI.h>
 #include <Microsoft.Windows.ApplicationModel.WindowsAppRuntime.DeploymentManager.g.cpp>
 
+#include <WinBase.h>
```
DrusTheAxe, line 11: Well that's weird. Why? Isn't that already included by stuff in pch.h?

winbase.h is included by windows.h (or unknwn.h)

Or is this a same named but different file from something else? C++/WinRT ?

dev/Deployment/DeploymentManager.cpp
```diff
@@ -8,8 +8,12 @@
...
 #include "WindowsAppRuntime-License.h"
 
+using namespace winrt;
+using namespace Windows::Foundation;
```
DrusTheAxe, line 15: added? This was a C++/WinRT implementation in 1.0, you just added the options object and initialize overload. These weren't declared before?

dev/Deployment/DeploymentManager.cpp
```diff
@@ -25,6 +29,12 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
         return Initialize(GetCurrentFrameworkPackageFullName());
     }
 
+    winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::DeploymentResult DeploymentManager::Initialize(winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::DeploymentInitializeOptions const& deploymentInitializeOptions)
+    {
+        FAIL_FAST_HR_IF(HRESULT_FROM_WIN32(APPMODEL_ERROR_NO_PACKAGE), !AppModel::Identity::IsPackagedProcess());
```
DrusTheAxe, line 34: BTW DeploymentManager's documented as not supporting unpackaged?

We should lift that restriction. Packaged Store-distributed apps _need_ this API but (20-20 hindsight) it's useful (and viable) for _everyone_ for _repair_ scenarios if nothing else. Also, given DynDep changes for elevation as long as the DDLM package(s) are registered this could be used by unpackaged apps as a way to install Main+Singleton.

Please file a vFuture issue to lift the current restriction.

dev/WindowsAppRuntime_MSIXInstallFromPath/main.cpp
```diff
@@ -118,18 +118,18 @@ class PackageId
...
-            L"  -?, --help = Display help\n"
-            L"          -- = End of options\n"
-            L"where:\n"
-            L"  path = Path where Windows App SDK's MSIX packages can be found\n"
-            L"           => Microsoft.WindowsAppRuntime.*.msix\n"
-            L"           => Microsoft.WindowsAppRuntime.DDLM.*.msix\n"
-            L"           => Microsoft.WindowsAppRuntime.Main.*.msix\n"
-            L"           => Microsoft.WindowsAppRuntime.Singleton.*.msix\n");
+        L"options:\n"
+        L"  -f, --force = Force shutdown the target applications if they are in use, to install the MSIX packages\n"
```
DrusTheAxe, line 122: Phrasing's a bit odd (applications? which? but this is messing with winappsdk's packages...)

SUGGESTION: "Force shutdown WinAppSDK's processes if necessary to update WinAppSDK's MSIX packages"

Not entirely sure I love that phrasing but closer. You could go with this (or something like it) and follow up offline with @angelazhangmsft @cwruss to vet final phrasing and a future PR for any changes that come of it

dev/WindowsAppRuntime_MSIXInstallFromPath/main.cpp
```diff
@@ -161,29 +161,30 @@ HRESULT JustDoIt(PCWSTR path) noexcept try
...
     }
 
-    auto hr{ AddPackage(path, filename) };
+    auto hr{ AddPackage(path, filename, forceDeployment) };
     if (FAILED(hr))
     {
-        wprintf(L"AddPackage(): 0x%X Path:%ls Filename:%ls PackageFullName:%ls", hr, path, filename.c_str(), packageFullName.c_str());
-        THROW_HR_MSG(hr, "Path:%ls Filename:%ls PackageFullName:%ls", path, filename.c_str(), packageFullName.c_str());
+        wprintf(L"AddPackage(): 0x%X Path:%ls Filename:%ls PackageFullName:%ls forceDeployment:%d", hr, path, filename.c_str(), packageFullName.c_str(), forceDeployment);
+        THROW_HR_MSG(hr, "Path:%ls Filename:%ls PackageFullName:%ls forceDeployment:%d", path, filename.c_str(), packageFullName.c_str(), forceDeployment);
```
DrusTheAxe, line 187: %d = int == int32 = 4byte
bool == 1byte
(at least, the way our compiler implements these)

Suggestion: `...forceDeployment:%s", ..., forceDeployment ? "true" : "false");`

installer/dev/main.cpp
```diff
@@ -103,6 +108,11 @@ int wmain(int argc, wchar_t *argv[])
 
             std::wcerr << "One or more install operations failed. Result: 0x" << std::hex << deployPackagesResult << std::endl;
         }
+
+        if (!Security::IntegrityLevel::IsElevated())
+        {
+            std::wcout << std::endl <<  "WARNING: Since WindowsAppRuntimeInstaller.exe is not run elevated, WindowsAppSDK packages are not provisioned for all users on your machine." << std::endl;
```
DrusTheAxe, line 114: WARNING is a bit strong

What's the normal text when provisioning packages? If "Provisioing: OK" or "Provisioning: ERROR 0x12345678 blahblah"  then you could make this "Provisioning: Skipped (requires elevation)".

Or simply say nothing. Only mention Provisioning if _DOING_ elevation

dev/WindowsAppRuntime_MSIXInstallFromPath/main.cpp
DrusTheAxe, line 122: Do we log anything that we're doing work with the Forced option? e.g. "Installing Package [Options=Force]: OK" or something? I don't see it but perhaps I'm just overlooking it

If not, we should. Here, or a follow up PR.

sachintaMSFT: Yes, we do already log Options in Install activity start event which captures Force or not.

installer/dev/main.cpp
sachintaMSFT: Agree. Changed it no to say "Provisioning Skipped as it requires elevation"

dev/Deployment/DeploymentManager.cpp
sachintaMSFT: https://task.ms/38653726

dev/Deployment/DeploymentManager.cpp
```diff
@@ -206,21 +219,29 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
...
     {
-        auto packagePathUri = winrt::Windows::Foundation::Uri(packagePath.c_str());
+        const auto packagePathUri = winrt::Windows::Foundation::Uri(packagePath.c_str());
         winrt::Windows::Management::Deployment::PackageManager packageManager;
-        auto options{ winrt::Windows::Management::Deployment::DeploymentOptions::None };
-        auto deploymentResult{ packageManager.AddPackageAsync(packagePathUri, nullptr, options).get() };
+
+        auto GetDeploymentOptions = [](bool forceDeployment)
+        {
+            return (forceDeployment == true ?
```
hulumane, line 229: Nit: ForceDeployment ? winrt::Windows::Management::Deployment::DeploymentOptions::ForceTargetApplicationShutdown:winrt::Windows::Management::Deployment::DeploymentOptions::None);

dev/Deployment/Deployment.idl
```diff
@@ -24,14 +29,36 @@ namespace Microsoft.Windows.ApplicationModel.WindowsAppRuntime
         HRESULT ExtendedError{ get; };
     };
 
+    /// This object is used to specify deployment options to apply when using DeploymentManager's
+    /// Initialize method
+    [contract(DeploymentContract, 2)]
+    runtimeclass DeploymentInitializeOptions
```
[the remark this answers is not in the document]
hulumane: My understanding was that the APIs are private and the SDK handles all of this underneath it all.

dev/Deployment/DeploymentInitializeOptions.cpp
```diff
@@ -0,0 +1,18 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+#include <pch.h>
+#include <DeploymentInitializeOptions.h>
+#include <Microsoft.Windows.ApplicationModel.WindowsAppRuntime.DeploymentInitializeOptions.g.cpp>
+
+namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implementation
+{
+    bool DeploymentInitializeOptions::ForceDeployment()
+    {
+        return m_ForceDeployment;
```
hulumane, line 11: Should there be a lock for this?

hulumane: I think it needs a lock because there is a multi-directional setter and a getter.

Interesting article on stack overflow regarding locks for bools:
https://stackoverflow.com/questions/3225223/do-i-need-to-protect-this-variable-with-a-lock#:~:text=With%20a%20boolean%2C%20you%20may%20be%20ok%20most,an%20actual%20mutex%20and%20are%20a%20lighter-weight%20solution.?msclkid=870ac6c1ac5a11eca331424d09d58709

dev/Deployment/DeploymentManager.cpp
```diff
@@ -206,21 +219,29 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
 
     // Adds the package at the path using PackageManager.
     // This requires the 'packageManagement' or 'runFullTrust' capabilities.
-    HRESULT DeploymentManager::AddPackage(const std::filesystem::path& packagePath) try
+    HRESULT DeploymentManager::AddPackage(const std::filesystem::path& packagePath, bool forceDeployment) try
     {
-        auto packagePathUri = winrt::Windows::Foundation::Uri(packagePath.c_str());
+        const auto packagePathUri = winrt::Windows::Foundation::Uri(packagePath.c_str());
         winrt::Windows::Management::Deployment::PackageManager packageManager;
```
hulumane, line 225: Nit: Move declaration of this variable next to it's usage.

dev/Deployment/DeploymentManager.cpp
```diff
@@ -274,7 +295,14 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
...
 
             // Deploy package.
-            RETURN_IF_FAILED(AddPackage(packagePath));
+            RETURN_IF_FAILED(AddPackage(packagePath, forceDeployment));
+        }
+
+        // Restart Push Notifications Long Running Platform when ForceDeployment option is applied.
+        if (forceDeployment)
+        {
+            wil::com_ptr_nothrow<INotificationsLongRunningPlatform> longRunningProcessPlatform{
```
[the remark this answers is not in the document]
hulumane: It would be good data for us as well from a reliability perspective.

hulumane, line 304: We talked about a 3 time retry policy here as well if it fails

dev/WindowsAppRuntime_MSIXInstallFromPath/main.cpp
```diff
@@ -161,29 +169,37 @@ HRESULT JustDoIt(PCWSTR path) noexcept try
...
 
-        AddPackageIfNecessary(path, filename, packageFullName);
+        AddPackageIfNecessary(path, filename, packageFullName, forceDeployment);
+    }
+
+    // Restart Push Notifications Long Running Platform when ForceDeployment option is applied.
+    if (forceDeployment)
+    {
+        wil::com_ptr_nothrow<INotificationsLongRunningPlatform> longRunningProcessPlatform{
+            wil::CoCreateInstance<NotificationsLongRunningPlatform, INotificationsLongRunningPlatform>(CLSCTX_LOCAL_SERVER) };
```
hulumane, line 179: Same concerns here. I think it would be nice to log and track errors here if Activation fails for any reason.

dev/WindowsAppRuntime_MSIXInstallFromPath/main.cpp
```diff
@@ -231,12 +247,20 @@ bool NeedToRegisterPackage(PCWSTR packageFullName)
...
+HRESULT AddPackage(PCWSTR path, const std::wstring& filename, bool forceDeployment)
 {
     const auto packagePath{ std::filesystem::path(path) / filename };
     const auto packagePathUri{ winrt::Windows::Foundation::Uri(packagePath.c_str()) };
     winrt::Windows::Management::Deployment::PackageManager packageManager;
-    const auto options{ winrt::Windows::Management::Deployment::DeploymentOptions::None };
+
+    auto GetDeploymentOptions = [](bool forceDeployment)
+    {
+        return (forceDeployment == true ?
```
hulumane, line 258: Nit: just forceDeployment

installer/dev/install.cpp
```diff
@@ -289,6 +300,40 @@ namespace WindowsAppRuntimeInstaller
...
 
+    // RestartPushNotificationsLRP is best effort and non-blocking to Installer functionality
+    void RestartPushNotificationsLRP()
+    {
+        WindowsAppRuntimeInstaller::InstallActivity::Context::Get().SetInstallStage(WindowsAppRuntimeInstaller::InstallActivity::InstallStage::RestartPushNotificationsLRP);
+
+        IID pushNotificationsIMPL_CLSID;
+        HRESULT hClsIdFromString = CLSIDFromString(PUSHNOTIFICATIONS_IMPL_CLSID_WSTRING, &pushNotificationsIMPL_CLSID);
+
+        if (SUCCEEDED(hClsIdFromString))
```
hulumane, line 311: Why the if check? Above call isn't expected to fail in any way right?

installer/dev/install.cpp
```diff
@@ -289,6 +300,40 @@ namespace WindowsAppRuntimeInstaller
...
+        {
+            IID pushNotificationsLRP_IID;
+            hClsIdFromString = CLSIDFromString(PUSHNOTIFICATIONS_LRP_CLSID_WSTRING, &pushNotificationsLRP_IID);
+
+            if (SUCCEEDED(hClsIdFromString))
```
hulumane, line 316: Same issue here. Above call isn't expected to fail. IF it does, its likely a bug and we want to be aware of it.

installer/dev/install.cpp
```diff
@@ -289,6 +300,40 @@ namespace WindowsAppRuntimeInstaller
     }
     CATCH_RETURN()
 
+    // RestartPushNotificationsLRP is best effort and non-blocking to Installer functionality
+    void RestartPushNotificationsLRP()
```
hulumane, line 304: Probably needs to be a helper method somewhere? You ssem to be initializing LRP from several other components in your code base.

installer/dev/install.cpp
```diff
@@ -289,6 +300,40 @@ namespace WindowsAppRuntimeInstaller
...
+
+        if (SUCCEEDED(hClsIdFromString))
+        {
+            IID pushNotificationsLRP_IID;
+            hClsIdFromString = CLSIDFromString(PUSHNOTIFICATIONS_LRP_CLSID_WSTRING, &pushNotificationsLRP_IID);
+
+            if (SUCCEEDED(hClsIdFromString))
+            {
+                wil::com_ptr<::IUnknown> pNotificationsLRP{};
+                LOG_IF_FAILED(CoCreateInstance(
```
hulumane, line 319: Why not use the wil equivalent? WRL::CoCreateInstanc is old style right?

dev/Deployment/DeploymentManager.cpp
sachintaMSFT: Yes, that is still the plan for the next commit update for the PR. First, I am looking to verify LRP restart change is working. Installer exe is not part of WAR dll and is not built in this repo's build pipelines but in internal pipelines due to dependencies. So, to test it from Installer, I have to run the internal pipeline to refer to this private branch.

installer/dev/install.cpp
sachintaMSFT: Just following good coding practice.

installer/dev/install.cpp
sachintaMSFT: Yes, the else condition logs it and we get telemetry for it.

installer/dev/InstallActivityContext.cpp
```diff
@@ -52,4 +52,13 @@ void WindowsAppRuntimeInstaller::InstallActivity::Context::SetLastFailure(const
     {
         m_lastFailure.message.clear();
     }
+
+    if (failure.pszCode)
+    {
+        m_lastFailure.code = *failure.pszCode;
```
hulumane, line 58: No lock protecting member variable access

installer/dev/install.cpp
hulumane: The CLSIDFromString call is never expected to fail. So being defensing and logging that seems overkill. I believe we should just THROW if that call fails because it is unexpected and should never happen.

installer/dev/install.cpp
sachintaMSFT: Installer exe is built outside of the WAR dll build pipeline and hence, it doesn't have access to runtime generated header files. Hence, I had to switch to using compile time header and to using WRL::CoCreateInstance due to having no access to NotificationsLRP Interface header file at compile time. Or is there one that I can use ? I will be very happy, if that's the case.

installer/dev/InstallActivityContext.cpp
sachintaMSFT: Installer is an executable and a process on it's own. Hence, there is no concern of simultaneous multiple entrance to this code block that we need to protect against.

installer/dev/install.cpp
sachintaMSFT: Fine. I will do that.

installer/dev/install.cpp
sachintaMSFT: LRP restart code is slightly different for Installer vs the other two locations because Installer is not built in the same WAR build. More details explained in other comments.

Even otherwise, there is quite a bit of overlap of code between components and inherited it from 1.0. After 1.1 code complete, I plan to pull out common parts for easy code maintenance over time. For this PR, this is it.

dev/Deployment/DeploymentManager.cpp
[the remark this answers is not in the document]
sachintaMSFT: Sure. That's the plan. Coming in my next update to the PR. Currently, just verifying that the LRP restart works as expected.

installer/dev/InstallActivityContext.cpp
hulumane: Gotcha. Understood! @DrusTheAxe Any concerns with not having locks here? My only concern is if the APIs are somehow misused outside the installer space (Not sure if that is possible today)

dev/Deployment/Deployment.idl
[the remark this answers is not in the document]
DrusTheAxe: `DeploymentManager.GetStatus()` tells you if WinAppSDK's runtime is properly/fully installed and ready to rock - or not.

`DeploymentManager.Initialize()` checcks status and, if not OK, installs packages. The driving motivator behind this is Store-distributed packaged apps. Store will install framework package dependencies so WARfwk is covered...but that's it. MSIX today has no way for a package to dependency on a main package and thus Store has no way of knowing you also need WARmain+singleton. So Store-distributed packaged apps can call this (e.g. in main()) to ensure WinAppSDK's installed and ready to rock, and thus things like Push Notifications are functional :-)

It can also make for a convenient "repair" option (for any type of app), but the critical raison raison d'etre is the Store-distributed packaged app, as noted above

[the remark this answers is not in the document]
DrusTheAxe: > My understanding was that the APIs are private and the SDK handles all of this underneath it all.

DeploymentManager is a new WinAppSDK invention, and it only calls public APIs. Just like the installer, not special sauce anyone couldn't do for themselves - just conveniently shrinkwrapped we provide to DoTheRightThing(TM) so devs don't have to sweat the details :-)

dev/Deployment/DeploymentInitializeOptions.cpp
DrusTheAxe: This is an inproc object accessing instance data. It's not designed for concurrent use of an instance, and if you foolishly did you'd only be horking yourself.

So, no. No need for a lock.

If this was an OOP object it'd be an entirely different matter.

dev/Deployment/DeploymentManager.cpp
DrusTheAxe: Never explicitly check for == or != for `true` or `false`.

dev/PushNotifications/PushNotifications-Constants.h
```diff
@@ -7,12 +7,19 @@
 #define _STRINGIZE(x)   _STRINGIZEX(x)
 #endif
 
+#define _WSTRINGIZEXX(x) L ## #x
+#define _WSTRINGIZEX(x) _WSTRINGIZEXX(x)
+#define _WSTRINGIZE(x) _WSTRINGIZEX({ ## x ## })
```
DrusTheAxe, line 12: Why this?

Why does this differ from _STRINGIZE[X] above? That comes from yvals_core.h (as noted), and only of necessity.

dev/WindowsAppRuntime_MSIXInstallFromPath/main.cpp
```diff
@@ -138,6 +140,12 @@ HRESULT JustDoIt(PCWSTR path) noexcept try
     auto inventory{ fpath / L"MSIX.inventory" };
     std::string lineUtf8;
     std::ifstream f{ inventory };
+
+    if (!std::filesystem::exists(inventory))
+    {
+        THROW_HR_MSG(HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND), "inventory file doesn't exist: %s", inventory);
```
DrusTheAxe, line 146: 1. Message text here is redundant with the FILE_NOT_FOUND error code
2. WIL *_MSG(....%s...) == narrow string but `inventory` is a wide string
3. inventory is a `std::filesystem::path` but %s expects C string

`THROW_HR_MSG(HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND), "%ls", inventory.c_str());`

dev/WindowsAppRuntime_MSIXInstallFromPath/main.cpp
DrusTheAxe, line 179: Does PushNotifications have an API? Would be preferable to assuming "restart push == CCI(pushserver)"

SUGGESTION:

1. `dev\PushNotificationsServicing.h` declares namespace Microsoft::blahblah::PushNotifications { inline HRESULT StartService() { return CCI(pushserver); }`
2. `devPushNotifications.vcxitems publishes `PushNotificationsServicing.h` (much like DynDep does for msixdynamicdependency.h)
3. Installer `#include <PushNotificationsServicing.h>` and calls `StartService()`

This way if/when Push' implementation evolves Push has direct control and visibility to the implementation and installer can be oblivious to the fine print

dev/PushNotifications/PushNotifications-Constants.h
sachintaMSFT: WideString vs String. That's the difference. It was needed for CLSIDFromString API usage.

dev/WindowsAppRuntime_MSIXInstallFromPath/main.cpp
sachintaMSFT: Check my PR update for this which I coded before seeing your comment. If there is strong resistance to the method I used, I will add another commit. Looking to code complete.

dev/WindowsAppRuntime_MSIXInstallFromPath/main.cpp
```diff
@@ -138,6 +139,9 @@ HRESULT JustDoIt(PCWSTR path) noexcept try
     auto inventory{ fpath / L"MSIX.inventory" };
     std::string lineUtf8;
     std::ifstream f{ inventory };
+
+    THROW_HR_MSG(HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND), "%ls", inventory.c_str());
```
hulumane, line 143: Looks like a bug. Did you intend to have this always throw?

hulumane: Likely you meant THROW_HR_IF_MSG?

sachintaMSFT: aargh!! copy pasted Howard's feedback as is. my bad. thanks for catching it.

installer/dev/InstallActivityContext.cpp
sachintaMSFT: Installer doesn't expose any APIs. It's a self-contained executable.

hulumane (approved):

dev/WindowsAppRuntime_MSIXInstallFromPath/main.cpp
DrusTheAxe: I wouldn't say strong or resistance. Merely a suggestion for refactoring - sooner or later. Not blocking here, just a good idea for long term hygiene

DrusTheAxe (approved):

PR #2285, opened 2022-03-17
User/shasnayak/unpackaged iou

ShashankNay:
Addresses some concerns with[ the original PR ](https://github.com/microsoft/WindowsAppSDK/pull/1582), contains some fixes based on suggestions in the comments.


DrusTheAxe:
Superceded by more recent work accomplishing same but better

PR #2289, opened 2022-03-17
Prevent WinUI item templates from showing up in other project types

japarson:
### Problem
WinUI item templates are appearing in projects where they do not belong. For example, here they are appearing in a C# console application:
image

### Solution
In the .vstemplate files for each project item, there is a [TemplateGroupID property](https://docs.microsoft.com/en-us/visualstudio/extensibility/templategroupid-element-visual-studio-templates?view=vs-2022) that specifies what kind of project an item templates will show up in. The WinUI item templates correctly set this property. However, there is a [ShowByDefault property](https://docs.microsoft.com/en-us/visualstudio/extensibility/showbydefault-visual-studio-templates?view=vs-2022) which specifies that the template will only be displayed under the specified `TemplateGroupID` when set to `false`.

The current WinUI item templates have `ShowByDefault` set to `true` and the solution here it to set them to `false`.

### Demo
The item templates no longer appear in a C# console application:
image

And they still appear in a WinUI app:
image

### Work Item
[Bug 1433670](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1433670): WinUI C# templates are available to incompatible projects

### DIFF
--- a/dev/VSIX/ItemTemplates/Desktop/CSharp/BlankWindow/WinUI.Desktop.Cs.BlankWindow.vstemplate
+++ b/dev/VSIX/ItemTemplates/Desktop/CSharp/BlankWindow/WinUI.Desktop.Cs.BlankWindow.vstemplate
@@ -10,7 +10,7 @@
     <TemplateGroupID>WinRT-Managed-UAP</TemplateGroupID>
     <AppliesTo>CSharp + SharedAssetsProject</AppliesTo>
     <NumberOfParentCategoriesToRollUp>2</NumberOfParentCategoriesToRollUp>
-    <ShowByDefault>true</ShowByDefault>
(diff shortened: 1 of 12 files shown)


RealTommyKlein:
Adding Ryan and Evelyn since I don't own the templates and they've worked on them in the past, but LGTM 👍

RealTommyKlein (approved):

japarson:
@evelynwu-msft @ShashankNay I tested the VSIX that has my changes. The results can be found in the PR description. It appears to be working correctly now. How can I run the status checks for this PR?

evelynwu-msft (approved):
:shipit:

ShashankNay (approved):

PR #2292, opened 2022-03-18
Add AppLifecycle contract

DefaultRyan:
The AppLifecycle types were not on an api contract in the 1.0 release. And in the meantime, there have been two additions to these types:
- The enum `ExtendedActivationKind` added `AppNotification`
- The runtimeclass `AppInstance` added the static method `Restart`.

Without adding a new contract version, these changes would create an ABI breaking change. These two additions are now in the v2 contract version.

(Note: this will need to be cherry-picked into the 1.1 release branch)

### DIFF
--- a/dev/AppLifecycle/AppLifecycle.idl
+++ b/dev/AppLifecycle/AppLifecycle.idl
@@ -3,6 +3,10 @@
 
 namespace Microsoft.Windows.AppLifecycle
 {
+    [contractversion(2)]
+    apicontract AppLifecycleContract {};
+
+    [contract(AppLifecycleContract, 1)]
     enum ExtendedActivationKind
     {
         Launch = 0,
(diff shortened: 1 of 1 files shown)


dev/AppLifecycle/AppLifecycle.idl
```diff
@@ -53,20 +57,26 @@ namespace Microsoft.Windows.AppLifecycle
...
     };
 
+    [contract(AppLifecycleContract, 1)]
     runtimeclass AppInstance
     {
         static AppInstance GetCurrent();
         static Windows.Foundation.Collections.IVector<AppInstance> GetInstances();
         static AppInstance FindOrRegisterForKey(String key);
+
+        [contract(AppLifecycleContract, 2)]
```
jonwis, line 79: Did @dhoehna cover whether we're using major or major.minor contract versions? There's no _technical_ problem with using just major, except that we might eventually want to make breaking changes clearly called out with the "2.0" version number.

jonwis (approved):

dev/AppLifecycle/AppLifecycle.idl
dhoehna: It was agreed that there would be a MAJOR version only.

dhoehna (approved):

PR #2293, opened 2022-03-18
List of contracts across WinApSDK

DrusTheAxe:
Added list of contacts acrocss WinAppSDK. NOTE: The list is meant to start the dialogue. There's undoubtedly corrections and additions needed.

Added TOC and section numbering. Minor formatting corrections

NOTE: This is a work-in-progress. Goal is to have a master list of the various contract names across features, then bring the list through API Review in 1 pass, then update code to match if/as necessary.

### DIFF
--- a/specs/WinRT/WinRTAPIContracts.md
+++ b/specs/WinRT/WinRTAPIContracts.md
@@ -120,5 +120,5 @@ TODO:Start of the list. Needs review and undoubtedly correction and additions (e
 | ---                | ---           | WindowsAppSDKContract           | Microsoft.Foundation                                 |1,2,3    |
 
 <sup>1</sup> Only for use by IXP.<BR>
-<sup>2</sup> This contract is misnamed. It should be called `InteractiveExperiencesContract` in the `Windows.UI.Composition` namespace.<BR>
+<sup>2</sup> This contract is misnamed. It should have been named `InteractiveExperiencesContract` in the `Windows.UI.Composition` namespace.<BR>
 <sup>3</sup> This contract is misversioned. It should have been `[contract(...,1)]` instead of `[contract(...,1.1)]`. The next contract version will be `[contract(...,2)]`.<BR>
--- a/specs/WinRT/WinRTAPIContracts.md
+++ b/specs/WinRT/WinRTAPIContracts.md
@@ -22,18 +22,22 @@ WinRT APIs.
    1. One contract per feature/component/sub-system
    2. Contracts will not span across a transport package/repo
 2. **Contract Version**<BR>
-Contracts versions are a single number with no minor version, no "x.y".  The contract version is
-frozen on a stable release.  Any changes after a stable release requires the version to be incremented.
+Contracts versions are a single number with no minor version, no "x.y". The contract version is
+frozen on a stable release. Any changes after a stable release requires the version to be incremented.
 3. **Enforcement**<BR>
 The build system has gates that enforce contract versioning to verify that contract is present,
  up to date, and there are no conflicts between APIs.
 4. **Contract Migration**<BR>
 This spec does not deal with contract migration. A separate spec will be available for contract
 migration if, and when, it is needed.
-5. **Internal contracts**<BR>
-All internal WinRT APIs have a contract, and the version is 1. The version is never updated.
-6. **Contract naming**<BR>
-Contract names must end with the word `Contract`.
+5. **Contract naming**<BR>
+Contract names must end with `Contract`.
+6. **Internal contracts**<BR>
+Internal contracts follow additional rules:
+    1. All internal WinRT APIs have an internal contract.
+    2. Internal contracts are tagged with `[internal]`.
+    3. Internal contracts are always version 1. The version is never updated.
+    4. Internal contract names must end with `InternalContract`
 
 # 3. Adding a new contract
 Contract placement is part of an API review. When adding a new API, the API team will
@@ -98,16 +102,19 @@ The list of all contracts defined across Windows App SDK
 
 TODO:Start of the list. Needs review and undoubtedly correction and additions (e.g. IXP)
 
-| Feature            | Repository    | Contract                        | NamespaceMicrosoft.Windows.System.Power              | Comment |
-|--------------------|---------------|---------------------------------|------------------------------------------------------|---------|
-| AccessControl      | windowsappsdk | AccessControlContract           | Microsoft.Windows.Security.AccessControl             |         |
-| AppLifecycle       | windowsappsdk | AppLifecycleContract            | Microsoft.Windows.AppLifecycle                       |         |
-| AppNotifications   | windowsappsdk | WindowsAppNotificationsContract | Microsoft.Windows.AppNotifications                   |         |
-| Deployment         | windowsappsdk | DeploymentManagerContract       | Microsoft.Windows.ApplicationModel.WindowsAppRuntime |         |
-| DynamicDependency  | windowsappsdk | DynamicDependencyContract       | Microsoft.Windows.ApplicationModel.DynamicDependency |         |
-| EnvironmentManager | windowsappsdk | EnvironmentManagerContracct     | Microsoft.Windows.System                             |         |
-| MRTCore            | windowsappsdk | MrtContract                     | Microsoft.Windows.ApplicationModel.Resources         |         |
-| PowerNotifications | windowsappsdk | PowerNotificationsContract      | Microsoft.Windows.System.Power                       |         |
-| PushNotifications  | windowsappsdk | PushNotificationsContract       | Microsoft.Windows.PushNotifications                  |         |
-| WinUI              | winui         | WinUIContract                   | Microsoft.UI.Xaml                                    |         |
-| WinUI              | winui         | WinUIControlsContract           | Microsoft.UI.Xaml.Controls                           |         |
+| Feature            | Repository    | Contract                        | NamespaceMicrosoft.Windows.System.Power              | Comment  |
+|--------------------|---------------|---------------------------------|------------------------------------------------------|----------|
+| AccessControl      | windowsappsdk | AccessControlContract           | Microsoft.Windows.Security.AccessControl             |          |
+| AppLifecycle       | windowsappsdk | AppLifecycleContract            | Microsoft.Windows.AppLifecycle                       |          |
+| AppNotifications   | windowsappsdk | WindowsAppNotificationsContract | Microsoft.Windows.AppNotifications                   |          |
+| Deployment         | windowsappsdk | DeploymentManagerContract       | Microsoft.Windows.ApplicationModel.WindowsAppRuntime |          |
+| DynamicDependency  | windowsappsdk | DynamicDependencyContract       | Microsoft.Windows.ApplicationModel.DynamicDependency |          |
+| EnvironmentManager | windowsappsdk | EnvironmentManagerContracct     | Microsoft.Windows.System                             |          |
+| MRTCore            | windowsappsdk | MrtContract                     | Microsoft.Windows.ApplicationModel.Resources         |          |
+| PowerNotifications | windowsappsdk | PowerNotificationsContract      | Microsoft.Windows.System.Power                       |          |
+| PushNotifications  | windowsappsdk | PushNotificationsContract       | Microsoft.Windows.PushNotifications                  |          |
+| WinUI              | winui         | HostingContract                 | Microsoft.UI.Xaml.Hosting                            |          |
+| WinUI              | winui         | WinUIContract                   | Microsoft.UI.Xaml                                    |          |
+| WinUI              | winui         | WinUIControlsContract           | Microsoft.UI.Xaml.Controls                           |          |
+| WinUI              | winui         | XamlDirectContract              | Microsoft.UI.Xaml.Core.Direct                        |          |
+| ---                | ---           | WindowsAppSDKContract           | Microsoft.Foundation                                 |Do not use|
--- a/specs/WinRT/WinRTAPIContracts.md
+++ b/specs/WinRT/WinRTAPIContracts.md
@@ -102,7 +102,7 @@ The list of all contracts defined across Windows App SDK
 
 TODO:Start of the list. Needs review and undoubtedly correction and additions (e.g. IXP)
 
-| Feature            | Repository    | Contract                        | NamespaceMicrosoft.Windows.System.Power              | Comment  |
+| Feature            | Repository    | Contract                        | Namespace                                            | Comment  |
 |--------------------|---------------|---------------------------------|------------------------------------------------------|----------|
 | AccessControl      | windowsappsdk | AccessControlContract           | Microsoft.Windows.Security.AccessControl             |          |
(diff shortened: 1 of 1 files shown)


specs/WinRT/WinRTAPIContracts.md
```diff
@@ -62,27 +70,44 @@ namespace Microsoft.Windows.System
...
-## Open issues
-None.
+# 5. Telemetry
+API contract version specific telemetry will not be collected because it does not provide any
+additional information compared to the Windows App SDK version.
+
+# 6. All Contracts in Windows App SDK
+The list of all contracts defined across Windows App SDK
+
+TODO:Start of the list. Needs review and undoubtedly correction and additions (e.g. IXP)
```
MikeHillberg, line 99: Also `WindowsAppSDKContract` and `XamlContract`

specs/WinRT/WinRTAPIContracts.md
```diff
@@ -22,18 +22,22 @@ WinRT APIs.
...
 The build system has gates that enforce contract versioning to verify that contract is present,
  up to date, and there are no conflicts between APIs.
 4. **Contract Migration**<BR>
 This spec does not deal with contract migration. A separate spec will be available for contract
 migration if, and when, it is needed.
-5. **Internal contracts**<BR>
-All internal WinRT APIs have a contract, and the version is 1. The version is never updated.
-6. **Contract naming**<BR>
-Contract names must end with the word `Contract`.
+5. **Contract naming**<BR>
```
MikeHillberg, line 33: Please no HTML (BR tag). To get a soft return in MD end the line with two spaces

specs/WinRT/WinRTAPIContracts.md
```diff
@@ -22,18 +22,22 @@ WinRT APIs.
    1. One contract per feature/component/sub-system
```
MikeHillberg, line 22: Why? What's the motivation/scenario?

MikeHillberg, line 22: Should add a note that [ApiInformation.IsApiContractPresent](https://docs.microsoft.com/uwp/api/Windows.Foundation.Metadata.ApiInformation.IsApiContractPresent) doesn't currently work against WinAppSDK contracts.

specs/WinRT/WinRTAPIContracts.md
```diff
@@ -22,18 +22,22 @@ WinRT APIs.
...
    2. Contracts will not span across a transport package/repo
 2. **Contract Version**<BR>
-Contracts versions are a single number with no minor version, no "x.y".  The contract version is
-frozen on a stable release.  Any changes after a stable release requires the version to be incremented.
+Contracts versions are a single number with no minor version, no "x.y". The contract version is
+frozen on a stable release. Any changes after a stable release requires the version to be incremented.
 3. **Enforcement**<BR>
 The build system has gates that enforce contract versioning to verify that contract is present,
  up to date, and there are no conflicts between APIs.
 4. **Contract Migration**<BR>
```
MikeHillberg, line 30: I'd like to update the contracts to whatever new plan comes out of this spec. Since IsApiContractPresent isn't yet supported, this isn't a breaking change.

specs/WinRT/WinRTAPIContracts.md
```diff
@@ -98,16 +102,19 @@ The list of all contracts defined across Windows App SDK
...
+| DynamicDependency  | windowsappsdk | DynamicDependencyContract       | Microsoft.Windows.ApplicationModel.DynamicDependency |          |
+| EnvironmentManager | windowsappsdk | EnvironmentManagerContracct     | Microsoft.Windows.System                             |          |
+| MRTCore            | windowsappsdk | MrtContract                     | Microsoft.Windows.ApplicationModel.Resources         |          |
+| PowerNotifications | windowsappsdk | PowerNotificationsContract      | Microsoft.Windows.System.Power                       |          |
+| PushNotifications  | windowsappsdk | PushNotificationsContract       | Microsoft.Windows.PushNotifications                  |          |
+| WinUI              | winui         | HostingContract                 | Microsoft.UI.Xaml.Hosting                            |          |
+| WinUI              | winui         | WinUIContract                   | Microsoft.UI.Xaml                                    |          |
+| WinUI              | winui         | WinUIControlsContract           | Microsoft.UI.Xaml.Controls                           |          |
```
MikeHillberg, line 118: I don't actually see all of these contracts in metadata, at least not in 1.0. For example this one, and for example the XamlDirect feature below is in Windows but not WinAppSDK. All I see in metadata are
`MrtContract`
`WindowsAppSDKContract`
`WinUIContract`
`XamlContract`
?

specs/WinRT/WinRTAPIContracts.md
DrusTheAxe: That plays poorly with editors that strip trailing whitespace

specs/WinRT/WinRTAPIContracts.md
DrusTheAxe: Because https://task.ms/36717508 or something else?

DrusTheAxe: Past experience with `Windows.Foundation.UniversalApiContract` being too monolithic

Back when contracts were created the competing options were contracts per feature/component/subsystem vs 1 contract per Windows release. We chose the latter, 'cause, reasons, but WinAppSDK != Windows so our needs and policies differ. We discussed this in detail recently and agreed the per feature/component/subsystem modularity is a better fit for our needs than Windows' 1 per release.

specs/WinRT/WinRTAPIContracts.md
DrusTheAxe: > I'd like to update the contracts

Which contracts? Existing contracts defined in 1.0? That would be a breaking change.

If `IsApiContractPresent` is broken _today_ and fixed tomorrow, callers would expect it to work...but if they run into 1.0 artifacts it would produce different answers than probing 1.1+.

We can discuss further offline if you like

specs/WinRT/WinRTAPIContracts.md
```diff
@@ -62,27 +74,47 @@ namespace Microsoft.Windows.System
...
+# 5. Telemetry
+API contract version specific telemetry will not be collected because it does not provide any
+additional information compared to the Windows App SDK version.
+
+# 6. All Contracts in Windows App SDK
+The list of all contracts defined across Windows App SDK
+
+TODO:Start of the list. Needs review and undoubtedly correction and additions (e.g. IXP)
+
+| Feature            | Repository    | Contract                        | NamespaceMicrosoft.Windows.System.Power              | Comment  |
```
dhoehna, line 105: NamespaceMicrosoft.Windows.System.Power?

specs/WinRT/WinRTAPIContracts.md
DrusTheAxe: `MrtContract` was defined in `dev\MRTcore\mrt\Microsoft.Windows.AppliccationModel.Resources\src\Miccrosoft.Windows.ApplicationModel.Resources.idl` [2 years ago](https://github.com/microsoft/WindowsAppSDK/blame/31b83ed3e72512d37ebbab8d7d37f20f7cf9b44b/dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/src/Microsoft.Windows.ApplicationModel.Resources.idl)

`WindowsAppSDKContract` was defined recently which directly led to further discussion and then this spec. It may not be present in any build yet

specs/WinRT/WinRTAPIContracts.md
DrusTheAxe: Are you looking at latest? I don't see that

image

dhoehna (approved):

specs/WinRT/WinRTAPIContracts.md
BenJKuhn: Just keep in mind that renaming contracts will impact our ability to detect interface breaking changes using the winrt comparison tooling. If they're under different contracts, the tooling will likely treat them as unrelated APIs and not flag changes that it should.

specs/WinRT/WinRTAPIContracts.md
```diff
@@ -62,27 +74,49 @@ namespace Microsoft.Windows.System
...
+additional information compared to the Windows App SDK version.
+
+# 6. All Contracts in Windows App SDK
+The list of all contracts defined across Windows App SDK
+
+| Feature            | Repository    | Contract                        | Namespace                                            | Comment |
+|--------------------|---------------|---------------------------------|------------------------------------------------------|---------|
+| AccessControl      | windowsappsdk | AccessControlContract           | Microsoft.Windows.Security.AccessControl             |         |
+| AppLifecycle       | windowsappsdk | AppLifecycleContract            | Microsoft.Windows.AppLifecycle                       |         |
+| AppNotifications   | windowsappsdk | WindowsAppNotificationsContract | Microsoft.Windows.AppNotifications                   |         |
```
loneursid, line 107: I think @hulumane requested that the "Windows" prefix be removed from WindowsAppNotificationsContract as no other contract has it. Doing so would require that the contract name be change in the corresponding idl. Since the contract has only been released as part of preview1, now (before preview2 is probably the best time to do it,

hulumane: Yes please : )

hulumane: Unless there was a specific reason to use Windows* in the prefix, i don't see any other contracts using the same.

DrusTheAxe: > I think @hulumane requested that the "Windows" prefix be removed from WindowsAppNotificationsContract

Super! That was one I questioned. Done.

>Since the contract has only been released

Not a problem. We have until 1.1 Stable, though sooner is better (and once 1.1 ships we're pretty much locked until 2.0).

Don't worry about what's currently checked in. We have a wide mix - some types have the desired contract (per above), some have a different one (e.g. DynamicDependency lacks the 'Contract' suffix) and some (most) have none. Goal is to get this list through API Review, then we can update whatever code necessary.

loneursid (approved):

hulumane (approved):

specs/WinRT/WinRTAPIContracts.md
jonwis: The comparison tooling compares namespace+typename, and then all the attributes on the typename. Contract is an attribute on the typename. Moving a type across namespaces is detected as "unexpected attribute change; ContractAttribute" ... but it _is_ a breaking change to rename a contract and move a type across contracts, as `IsApiContractPresent` on the old name will fail, even though the resolved-by-namespace-and-name type is present.

specs/WinRT/WinRTAPIContracts.md
jonwis: That's a surprise; IsApiContractPresent should be using the `Ro...` APIs that everything else uses.

specs/WinRT/WinRTAPIContracts.md
MikeHillberg: Renaming or rearranging contracts isn't a breaking change in WinAppSDK yet because IsApiContractPresent isn't supported for non-OS types.

MikeHillberg: Sorry, repeating an earlier comment, I can't find the page that says not to use contracts on non-OS types, and I must have had a typo when I tested it earlier.

specs/WinRT/WinRTAPIContracts.md
DrusTheAxe: [Predefined attributes (MIDL 3.0) - Windows UWP applications | Microsoft Docs](https://docs.microsoft.com/en-us/uwp/midl-3/predefined-attributes#the-contract-attribute)
image

specs/WinRT/WinRTAPIContracts.md
```diff
@@ -62,27 +74,49 @@ namespace Microsoft.Windows.System
...
 }
 ```
-## 1.1 ship blocker
-Any APIs that changed between 1.0 and 1.1 need to add a contract or 1.1 will not ship.
 
-## Telemetry
-API specific telemetry will not be collected because it does not provide any additional information
-compared to the Windows App SDK version.
+# 4. Missing contracts is a ship blocker
+WinRT APIs without a contract are a shipblocking bug.
```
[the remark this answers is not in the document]
DrusTheAxe: Policy decision for WinAppSDK. We don't want to ship any WinRT types w/o contracts

specs/WinRT/WinRTAPIContracts.md
```diff
@@ -62,27 +74,49 @@ namespace Microsoft.Windows.System
...
+| AppNotifications   | windowsappsdk | AppNotificationsContract        | Microsoft.Windows.AppNotifications                   |         |
+| Deployment         | windowsappsdk | DeploymentManagerContract       | Microsoft.Windows.ApplicationModel.WindowsAppRuntime |         |
+| DynamicDependency  | windowsappsdk | DynamicDependencyContract       | Microsoft.Windows.ApplicationModel.DynamicDependency |         |
+| EnvironmentManager | windowsappsdk | EnvironmentManagerContracct     | Microsoft.Windows.System                             |         |
+| MRTCore            | windowsappsdk | MrtContract                     | Microsoft.Windows.ApplicationModel.Resources         |         |
+| PowerNotifications | windowsappsdk | PowerNotificationsContract      | Microsoft.Windows.System.Power                       |         |
+| PushNotifications  | windowsappsdk | PushNotificationsContract       | Microsoft.Windows.PushNotifications                  |         |
+| WinUI              | winui         | HostingContract                 | Microsoft.UI.Xaml.Hosting                            |         |
```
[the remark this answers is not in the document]
DrusTheAxe: @MikeHillberg any objections?

specs/WinRT/WinRTAPIContracts.md
```diff
@@ -62,27 +74,49 @@ namespace Microsoft.Windows.System
...
+| WinUI              | winui         | WinUIContract                   | Microsoft.UI.Xaml                                    |         |
+| WinUI              | winui         | WinUIControlsContract           | Microsoft.UI.Xaml.Controls                           |         |
+| WinUI              | winui         | XamlDirectContract              | Microsoft.UI.Xaml.Core.Direct                        |         |
+| ---                | ---           | WindowsAppSDKContract           | Microsoft.Foundation                                 |1,2,3    |
+
+<sup>1</sup> Only for use by IXP.<BR>
```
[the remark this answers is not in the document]
DrusTheAxe: It's not a secret. "ixp" appears in multiple places

WinAppSDK's nuget `build\Microsoft.InteractiveExperiences.targets`

```xml
     <CsWinRTInputs Include="$(MSBuildThisFileDirectory)..\lib\$(Ixp-UAPTargetVersion)\*.winmd" />
```

Our nuspec
```xml
    <tags>...,Interactive Experiences,ixp,...</tags>
```

But here it's only referenced once in the footnote. That can easily be changed to 'InteractiveExperiences' (and should be, since the abbreviation's not defined anywhere, no matter how commonly informally used)

jonwis (approved):

sachintaMSFT (approved):

PR #2295, opened 2022-03-18
Notifications assets full support

danielayala94:


dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -232,79 +234,124 @@ std::wstring Microsoft::Windows::AppNotifications::Helpers::SetDisplayNameBasedO
...
+    *displayName = nullptr;
+    *iconFilePath = nullptr;
 
-    // Retrieve the icon
-    std::wstring iconFilePath{};
+    wil::unique_cotaskmem_string localDisplayName;
+    wil::unique_cotaskmem_string localIconFilePath;
 
-    wil::unique_hwnd hWindow{ GetConsoleWindow() };
+    HWND hWindow = GetForegroundWindow();
```
hulumane, line 263: Will this work if app window is minimized?

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -232,79 +234,124 @@ std::wstring Microsoft::Windows::AppNotifications::Helpers::SetDisplayNameBasedO
...
 
     if (hWindow)
```
hulumane, line 265: Why have this if check? Return ERROR_NOT_FOUND if this is null.

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -232,79 +234,124 @@ std::wstring Microsoft::Windows::AppNotifications::Helpers::SetDisplayNameBasedO
...
+    // 2. Based on the best app shortcut, using the FrameworkUdk.
+    // 3. From the foreground window.
+    // 4. Use the default assets.
+    if (FAILED(RetrieveAssetsFromProcess(&displayName, &iconFilePath)) &&
+        FAILED(ToastNotifications_RetrieveAssets_Stub(&displayName, &iconFilePath)) &&
+        FAILED(RetrieveAssetsFromWindow(&displayName, &iconFilePath)))
+    {
+        displayName = wil::make_unique_string<wil::unique_cotaskmem_string>(SetDisplayNameBasedOnProcessName().c_str());
+        iconFilePath = wil::make_unique_string<wil::unique_cotaskmem_string>(defaultAppNotificationIcon);
+        useDefaultAssets = true;
```
hulumane, line 336: Please remove the bool here. All the methods above need to check for format and return accordingly.

hulumane: So introduce a helper called: bool CheckIconFormat(PCWSTR iconPath). Call that form RetrieveAssetsFromProcess, ToastNotifications_RetrieveAssets_Stub and RetrieveAssetsFromWindow and only return S_OK if the format is supported.

hulumane: Or alternatively, if you have to use the bool, then put the Icon check below in a helper.

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -429,3 +476,8 @@ winrt::Microsoft::Windows::AppNotifications::AppNotification Microsoft::Windows:
 
     return notification;
 }
+
+bool Microsoft::Windows::AppNotifications::Helpers::IsWideStringEmptyOrNull(PCWSTR wideString)
+{
+    return wideString == nullptr || wcslen(wideString) > 0;
```
hulumane, line 482: This is incorrect. 
It should return (wideString == nullptr) || (wcslen(wideString) == 0)

dev/AppNotifications/AppNotificationUtility.cpp
hulumane: Actually there is a bug here. If one of these succeed but the icon is incompatible, we won't use the default icon anymore.

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -31,6 +31,8 @@ namespace ToastABI
     using namespace ::ABI::Microsoft::Internal::ToastNotifications;
 }
 
+constexpr PCWSTR defaultAppNotificationIcon = LR"(ms-resource://Windows.UI.ShellCommon/Files/Images/DefaultSystemNotification.png)";
```
hulumane, line 34: Is this resource guaranteed in all downlevel OS builds?

dev/AppNotifications/AppNotificationUtility.cpp
hulumane: Lets talk about this offline.

danielayala94: This makes more sense 😊

dev/AppNotifications/AppNotificationUtility.cpp
danielayala94: Yes, Lester confirmed it.

dev/AppNotifications/AppNotificationUtility.cpp
hulumane, line 263: HWND: Looks like a leak and you may need to use the wil construct instead.

dev/AppNotifications/AppNotificationUtility.cpp
jonwis: ... but it puts now a reverse dependency on the `Windows.UI.Shellcommon` component to host this forever. Is this a publicly documented value? Can you link to the `docs.microsoft.com` topic about how to use it?

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -232,79 +234,124 @@ std::wstring Microsoft::Windows::AppNotifications::Helpers::SetDisplayNameBasedO
...
+// Do nothing. This is just a placeholder while the UDK is ingested with the proper API.
+HRESULT ToastNotifications_RetrieveAssets_Stub(_Out_ PWSTR* displayName, _Out_ PWSTR* iconFilePath)
+{
+    *displayName = nullptr;
+    *iconFilePath = nullptr;
+
+    return E_NOTIMPL;
+}
+
+HRESULT RetrieveAssetsFromWindow(_Out_ PWSTR* displayName, _Out_ PWSTR* iconFilePath)
```
jonwis, line 255: Maybe this method should return a struct containing two things?  Like:

```c++
struct WindowAssets {
    std::wstring displayName;
    std::wstring iconFilePath;
}
WindowAssets RetrieveAssetsFromWindow();
```

dev/AppNotifications/AppNotificationUtility.cpp
jonwis: This doesn't seem right. What are you doing with this information?  The foreground window is any window in the system that has foreground, not just the one for this app.  Did you mean GetActiveWindow? Even then that might not be right.

jonwis: HWNDs don't really "leak" per se.

jonwis: Can you have the caller pass you context somehow?  Coming soon we'll have a `WindowId` construct that the caller can provide that is a reference to an HWND you can then use the HWNDy functions on directly.

jonwis (changes requested):

dev/AppNotifications/AppNotificationUtility.cpp
danielayala94: @jonwis Given that WinUI apps are not running under a console window, we still need to get its window. What would be the right API to use here?

Also, when the `WindowId` functionality will be available for use?

dev/AppNotifications/AppNotificationUtility.cpp
danielayala94: I don't see this icon being publicly documented.

hulumane: @DrusTheAxe Can we get this icon installed as part of the Framework Package? That might solve the reverse dependency issue ?

DrusTheAxe (changes requested):

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -232,79 +234,124 @@ std::wstring Microsoft::Windows::AppNotifications::Helpers::SetDisplayNameBasedO
...
+        REG_OPTION_NON_VOLATILE,
+        KEY_ALL_ACCESS,
+        nullptr /* lpSecurityAttributes */,
+        &hKey,
+        nullptr /* lpdwDisposition */));
+
+    bool useDefaultAssets{ false };
+
+    wil::unique_cotaskmem_string displayName;
+    wil::unique_cotaskmem_string iconFilePath;
```
DrusTheAxe, line 323: NIT: Move lines 322-323 to line 329.5

Declare variables at use

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -232,79 +234,124 @@ std::wstring Microsoft::Windows::AppNotifications::Helpers::SetDisplayNameBasedO
...
+        subKey.c_str(),
+        0,
+        nullptr /* lpClass */,
+        REG_OPTION_NON_VOLATILE,
+        KEY_ALL_ACCESS,
+        nullptr /* lpSecurityAttributes */,
+        &hKey,
+        nullptr /* lpdwDisposition */));
+
+    bool useDefaultAssets{ false };
```
DrusTheAxe, line 320: `bool useDefaultAssets{};`

no need for the explicit `false`

DrusTheAxe, line 320: NIT: Move to line 329.5

Declare variables at use

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -232,79 +234,124 @@ std::wstring Microsoft::Windows::AppNotifications::Helpers::SetDisplayNameBasedO
...
+    wil::unique_cotaskmem_string iconFilePath;
+
+    // Try the following techniques to retrieve display name and icon:
+    // 1. From the current process.
+    // 2. Based on the best app shortcut, using the FrameworkUdk.
+    // 3. From the foreground window.
+    // 4. Use the default assets.
+    if (FAILED(RetrieveAssetsFromProcess(&displayName, &iconFilePath)) &&
+        FAILED(ToastNotifications_RetrieveAssets_Stub(&displayName, &iconFilePath)) &&
+        FAILED(RetrieveAssetsFromWindow(&displayName, &iconFilePath)))
```
DrusTheAxe, line 332: FAILED(...)? Not FAILED(LOG_IF_FAILED(...))?

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -232,79 +234,124 @@ std::wstring Microsoft::Windows::AppNotifications::Helpers::SetDisplayNameBasedO
...
+        auto iteratorForFileExtension{ iconFilePathAsWstring.find_first_of(L".") };
         THROW_HR_IF_MSG(E_UNEXPECTED, iteratorForFileExtension == std::wstring::npos, "You must provide a valid filepath as the app icon.");
 
-        std::wstring iconFileExtension{ iconFilePath.substr(iteratorForFileExtension) };
-        THROW_HR_IF_MSG(E_UNEXPECTED, iconFileExtension != L".ico" && iconFileExtension != L".png",
-            "You must provide a supported file extension as the icon (.ico or .png).");
+        std::wstring iconFileExtension = iconFilePathAsWstring.substr(iteratorForFileExtension);
+        THROW_HR_IF_MSG(E_UNEXPECTED,
+            iconFileExtension != L".ico" && iconFileExtension != L".png" &&
+            iconFileExtension != L".jpg" && iconFileExtension != L".bmp",
```
DrusTheAxe, line 349: These won't match ".ICO", ".JPG", ".PNG", ".BMP". Is that expected?

(usually filenames/extensions are case-insensitive in Windows)

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -232,79 +234,124 @@ std::wstring Microsoft::Windows::AppNotifications::Helpers::SetDisplayNameBasedO
...
+    if (!useDefaultAssets)
+    {
+        std::wstring iconFilePathAsWstring{ iconFilePath.get() };
+        auto iteratorForFileExtension{ iconFilePathAsWstring.find_first_of(L".") };
         THROW_HR_IF_MSG(E_UNEXPECTED, iteratorForFileExtension == std::wstring::npos, "You must provide a valid filepath as the app icon.");
 
-        std::wstring iconFileExtension{ iconFilePath.substr(iteratorForFileExtension) };
-        THROW_HR_IF_MSG(E_UNEXPECTED, iconFileExtension != L".ico" && iconFileExtension != L".png",
-            "You must provide a supported file extension as the icon (.ico or .png).");
+        std::wstring iconFileExtension = iconFilePathAsWstring.substr(iteratorForFileExtension);
```
DrusTheAxe, line 346: Lines 342-346 are trying to get the extension of the file?

```
#include <filesystem>
...
if (!useDefaultAssets)
{
    const auto filename{ std::filesystem::path(iconFilePath.get() };
    const auto extension{ filename.extension() };
    THROW_HR_IF_MSG(...)
```

if iconFilePath="foo\bar.ico"
then extension=".ico"

std::filesystem::path is your friend :-)
https://en.cppreference.com/w/cpp/filesystem/path/extension

DrusTheAxe (changes requested):

dev/AppNotifications/AppNotificationUtility.h
```diff
@@ -62,4 +62,6 @@ namespace Microsoft::Windows::AppNotifications::Helpers
     winrt::Microsoft::Windows::AppNotifications::AppNotification ToastNotificationFromToastProperties(ABI::Microsoft::Internal::ToastNotifications::INotificationProperties* properties);
```
DrusTheAxe, line 63: github won't let me comment on the lines for `ConvertUtf8StringToWideString` so here you go

Are you aware of `dev\common\Microsoft.Utf8.h`?

`ConvertUtf8StringToWideString` can be replaced with a `ToUtf16()` in there. There's functions to give you a `std::unique_ptr<WCHAR[]>`, `std::wstring()` or `winrt::hstring`.

No `wil::unique_cotaskmem_string` overload but if you really need that it's not hard to add

DrusTheAxe (changes requested):

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -232,79 +234,124 @@ std::wstring Microsoft::Windows::AppNotifications::Helpers::SetDisplayNameBasedO
...
         winrt::com_ptr<IPropertyStore> propertyStore;
-        THROW_IF_FAILED(SHGetPropertyStoreForWindow(hWindow.get(), IID_PPV_ARGS(propertyStore.put())));
+        THROW_IF_FAILED(SHGetPropertyStoreForWindow(hWindow, IID_PPV_ARGS(propertyStore.put())));
 
         wil::unique_prop_variant propVariantDisplayName;
-        // Do not throw in case of failure, default to the filepath approach below as fallback to set a DisplayName.
         THROW_IF_FAILED(propertyStore->GetValue(PKEY_AppUserModel_RelaunchDisplayNameResource, &propVariantDisplayName));
 
-        if (propVariantDisplayName.vt == VT_LPWSTR && propVariantDisplayName.pwszVal != nullptr && wcslen(propVariantDisplayName.pwszVal) > 0)
+        if (propVariantDisplayName.vt == VT_LPWSTR && Microsoft::Windows::AppNotifications::Helpers::IsWideStringEmptyOrNull(propVariantDisplayName.pwszVal))
```
DrusTheAxe, line 273: Wow. There's actually a ...LPWSTR constant. I never knew that. How very retro ;-)

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -232,79 +234,124 @@ std::wstring Microsoft::Windows::AppNotifications::Helpers::SetDisplayNameBasedO
...
         wil::unique_prop_variant propVariantIcon;
         THROW_IF_FAILED(propertyStore->GetValue(PKEY_AppUserModel_RelaunchIconResource, &propVariantIcon));
 
-        THROW_HR_IF_MSG(E_UNEXPECTED, (propVariantIcon.vt == VT_EMPTY || propVariantIcon.pwszVal == nullptr), "Icon is not specified");
-
-        THROW_HR_IF_MSG(E_UNEXPECTED, propVariantIcon.vt != VT_LPWSTR, "Icon should be a valid Unicode string");
+        THROW_HR_IF(E_UNEXPECTED, propVariantIcon.vt != VT_LPWSTR || Microsoft::Windows::AppNotifications::Helpers::IsWideStringEmptyOrNull(propVariantIcon.pwszVal));
 
-        THROW_HR_IF_MSG(E_UNEXPECTED, wcslen(propVariantIcon.pwszVal) == 0, "Icon is an empty string");
+        std::wstring localIconFilePathAsWstring = propVariantIcon.pwszVal;
```
DrusTheAxe, line 283: NIT: `std::wstring localIconFilePathString{ propVariantIcon.pwszVal };`

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -232,79 +234,124 @@ std::wstring Microsoft::Windows::AppNotifications::Helpers::SetDisplayNameBasedO
...
+    return HRESULT_FROM_WIN32(ERROR_NOT_FOUND);
+}
+
+void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring const& appId, std::wstring const& clsid)
+{
+    wil::unique_hkey hKey;
+    // subKey: \Software\Classes\AppUserModelId\{AppGUID}
+    std::wstring subKey{ c_appIdentifierPath + appId };
+
+    THROW_IF_WIN32_ERROR(RegCreateKeyEx(
```
DrusTheAxe, line 309: Does WIL have registry functions?

dev/AppNotifications/AppNotificationUtility.cpp
DrusTheAxe: Better yet `return !wideString || (*wideString == L'\0')`

See my other comment

DrusTheAxe (changes requested):

riverar:
I can't attach comments to unchanged lines so am putting it here instead.

---

https://github.com/microsoft/WindowsAppSDK/blob/ce1563a38385b82543eba38f68462cbab8d3f6e0/dev/AppNotifications/AppNotificationUtility.cpp#L75-L81

This chunk of code generates a GUID with a known length then measures the length again, which isn't efficient. Instead you can perhaps do one of a few things (consult your nearest FTE for final guidance):
* Switch out the `StringFromCLSID` for `StringFromGUID2`, which returns the length
* Switch out the `wcslen` and use a [WIL guid length constant](https://github.com/microsoft/wil/wiki/Win32-helpers#string-length-constants) (i.e. `wil::guid_string_length`)

---

https://github.com/microsoft/WindowsAppSDK/blob/ce1563a38385b82543eba38f68462cbab8d3f6e0/dev/AppNotifications/AppNotificationUtility.cpp#L334

https://github.com/microsoft/WindowsAppSDK/blob/ce1563a38385b82543eba38f68462cbab8d3f6e0/dev/AppNotifications/AppNotificationUtility.cpp#L353-L355

`SetDisplayNameBasedOnProcessName` returns a `std::wstring` here (which tracks string length!), pulls out a raw pointer, then measures the string length again.

dev/AppNotifications/AppNotificationUtility.cpp
jonwis: You should probably talk with @rkarman or @jeffstall with a comprehensive "what are you really trying to do" question.  

My main issue here is the _implicit_ context being used. Instead, consider an API where the app passes you the asset/text information necessary.

dev/AppNotifications/AppNotificationUtility.cpp
danielayala94: There are valid concerns in this thread. I have started an email thread to address this offline.

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -31,6 +31,8 @@ namespace ToastABI
     using namespace ::ABI::Microsoft::Internal::ToastNotifications;
 }
```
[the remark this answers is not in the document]
danielayala94: I filed a bug for this -> #2307

dev/AppNotifications/AppNotificationUtility.cpp
danielayala94: Is it to log the error? Also, does LOG_IF_FAILED return the corresponding HRESULT?

dev/AppNotifications/AppNotificationUtility.h
danielayala94: Added to this bug -> #2307

dev/AppNotifications/AppNotificationUtility.cpp
danielayala94: After discussing with the team, for now I'm going to remove the window check and re-introduce it at a later release.

hulumane: Agree with this decision. We should focus on the Process Icon, Shortcut icon and default icon. Everything else is a lower priority IMHO.

dev/AppNotifications/AppNotificationUtility.cpp
hulumane: 2 options here:
1) Add a UDK API since this is closed source and can change over time. Have a contract in place with Shell to handle name changes in the future or prevent name changes in the future.
2) Include the icons in Framework SDK and infer to those icons instead.

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -232,6 +237,34 @@ std::wstring Microsoft::Windows::AppNotifications::Helpers::SetDisplayNameBasedO
...
+    // THROW_HR_IF_MSG(E_UNEXPECTED, VerifyIconFileExtension(iconFilePath));
+
+    return E_NOTIMPL;
+}
+
+HRESULT VerifyIconFileExtension(std::filesystem::path& iconFilePath)
+{
+    const auto fileExtension = iconFilePath.extension();
+
+    const bool isFileExtensionSupported =
```
hulumane, line 260: easier solution. Convert the extension to lower case. That takes care of .Bmp, .Png and other cases that aren't considered here. Then compare against the lower case variant. 

Or alternatively use wcscicmp or equivalent std:: API which ignores case

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -249,63 +282,23 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
...
-            iconFilePath.erase(iteratorForCommaDelimiter);
-        }
-
-        auto iteratorForFileExtension{ iconFilePath.find_first_of(L".") };
-        THROW_HR_IF_MSG(E_UNEXPECTED, iteratorForFileExtension == std::wstring::npos, "You must provide a valid filepath as the app icon.");
-
-        std::wstring iconFileExtension{ iconFilePath.substr(iteratorForFileExtension) };
-        THROW_HR_IF_MSG(E_UNEXPECTED, iconFileExtension != L".ico" && iconFileExtension != L".png",
-            "You must provide a supported file extension as the icon (.ico or .png).");
+        assets.displayName = wil::make_unique_string<wil::unique_cotaskmem_string>(SetDisplayNameBasedOnProcessName().c_str());
```
hulumane, line 295: This should be called GetDisplayNameFromProcess.
SetDisplayName* seems incorrect.

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -249,63 +282,23 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
...
-    displayName = SetDisplayNameBasedOnProcessName();
-
-    // Retrieve the icon
-    std::wstring iconFilePath{};
-
-    wil::unique_hwnd hWindow{ GetConsoleWindow() };
+    // Try the following techniques to retrieve display name and icon:
+    // 1. From the current process.
+    // 2. Based on the best app shortcut, using the FrameworkUdk.
+    // 3. From the foreground window.
```
hulumane, line 288: Remove #3 from comments please.

dev/AppNotifications/AppNotificationUtility.cpp
hulumane: Or even better. Keep the pattern consistent with above. Call it RetrieveDefaultAssets(assets);

Then do THROW__IF_FAILED(RetrieveDEfaultAssets(assets));

dev/AppNotifications/AppNotificationUtility.h
```diff
@@ -16,6 +16,11 @@ namespace Microsoft::Windows::AppNotifications::Helpers
     const std::wstring c_quote{ LR"(")" };
     const std::wstring c_notificationActivatedArgument{ L" ----AppNotificationActivated:" };
 
+    struct AppNotificationAssets {
+        wil::unique_cotaskmem_string displayName;
```
hulumane, line 20: Why not std::wstring? Its easier right and more standard than wil?

dev/AppNotifications/AppNotificationUtility.h
```diff
@@ -62,4 +67,9 @@ namespace Microsoft::Windows::AppNotifications::Helpers
     winrt::Microsoft::Windows::AppNotifications::AppNotification ToastNotificationFromToastProperties(ABI::Microsoft::Internal::ToastNotifications::INotificationProperties* properties);
 
     std::wstring SetDisplayNameBasedOnProcessName();
+
+    inline bool IsNullOrEmpty(PCWSTR string)
```
hulumane, line 71: This was move to common. So why is it still here?

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -232,6 +237,46 @@ std::wstring Microsoft::Windows::AppNotifications::Helpers::SetDisplayNameBasedO
...
+
+HRESULT RetrieveDefaultAssets(_Out_ Microsoft::Windows::AppNotifications::Helpers::AppNotificationAssets& assets)
+{
+    assets.displayName = Microsoft::Windows::AppNotifications::Helpers::GetDisplayNameBasedOnProcessName();
+    assets.iconFilePath = defaultAppNotificationIcon;
+
+    return S_OK;
+}
+
+HRESULT VerifyIconFileExtension(std::filesystem::path const& iconFilePath)
```
hulumane, line 264: This should return bool. It should be called IsIconFileExtensionSupported()

hulumane: bool IsIconFileExtensionSupported()

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -232,6 +237,46 @@ std::wstring Microsoft::Windows::AppNotifications::Helpers::SetDisplayNameBasedO
...
+
+// Do nothing. This is just a placeholder while the UDK is ingested with the proper API.
+HRESULT ToastNotifications_RetrieveAssets_Stub(_Out_ Microsoft::Windows::AppNotifications::Helpers::AppNotificationAssets& /*assets*/)
+{
+    // THROW_HR_IF_MSG(E_UNEXPECTED, VerifyIconFileExtension(iconFilePath));
+
+    return E_NOTIMPL;
+}
+
+HRESULT RetrieveDefaultAssets(_Out_ Microsoft::Windows::AppNotifications::Helpers::AppNotificationAssets& assets)
```
hulumane, line 256: try catch return

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -237,14 +240,57 @@ std::wstring Microsoft::Windows::AppNotifications::Helpers::GetDisplayNameBasedO
     return displayName;
 }
 
-// Placeholder
-HRESULT RetrieveAssetsFromProcess(_Out_ Microsoft::Windows::AppNotifications::Helpers::AppNotificationAssets& /*assets*/)
+inline wil::unique_hicon RetrieveIconFromProcess()
```
hulumane, line 243: Recommend moving this all to a seperate class called ShellUtility.cpp, .h

dev/AppNotifications/WICUtility.h
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#pragma once
+
+#include "pch.h"
+
+namespace Microsoft::Windows::AppNotifications::WICHelpers
+{
+    void WriteHIconToPngFile(wil::unique_hicon const& hIcon, _In_ PCWSTR pszFileName);
```
hulumane, line 10: All Shell helpers should be moved here. This function should be private. So only include its definition and implementation in the cpp.

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -237,14 +240,57 @@ std::wstring Microsoft::Windows::AppNotifications::Helpers::GetDisplayNameBasedO
...
 {
-    // THROW_HR_IF_MSG(E_UNEXPECTED, VerifyIconFileExtension(iconFilePath));
+    std::wstring processPath{};
+    THROW_IF_FAILED(wil::GetModuleFileNameExW(GetCurrentProcess(), nullptr, processPath));
 
-    return E_NOTIMPL;
+    // Extract the small icon from the first .ico, if failed extract the large icon.
```
hulumane, line 248: Add a comment: Small icon is sufficient for Toast icons since higher pixel count doesn't impact the visual in any substantial way

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -237,14 +240,57 @@ std::wstring Microsoft::Windows::AppNotifications::Helpers::GetDisplayNameBasedO
...
 }
 
+inline std::wstring RetrieveLocalFolderPath()
+{
+    wil::unique_cotaskmem_string localAppDataPath;
+    THROW_IF_FAILED(SHGetKnownFolderPath(FOLDERID_LocalAppData, 0 /* flags */, nullptr /* access token handle */, &localAppDataPath));
+
+    // path: C:\Users\<currentUser>\AppData\Local\Microsoft
+    std::path localFolderPath{ std::wstring(localAppDataPath.get()) + Microsoft::Windows::AppNotifications::Helpers::c_localMicrosoftFolder };
+    THROW_HR_IF(ERROR_FILE_NOT_FOUND, !std::exists(localFolderPath));
```
hulumane, line 265: Not needed. This path is standardized and shouldnt be missing. And even if it doesn't exist, we should create the file in the folder structure.

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -237,14 +240,57 @@ std::wstring Microsoft::Windows::AppNotifications::Helpers::GetDisplayNameBasedO
...
+}
+
+HRESULT Microsoft::Windows::AppNotifications::Helpers::RetrieveAssetsFromProcess(_Out_ Microsoft::Windows::AppNotifications::Helpers::AppNotificationAssets& assets) noexcept try
+{
+    wil::unique_hicon hIcon{ RetrieveIconFromProcess() };
+
+    std::wstring notificationAppId{ RetrieveNotificationAppId() };
+
+    // path: C:\Users\<currentUser>\AppData\Local\Microsoft\WindowsAppSDK\{AppGUID}.png
+    std::wstring writeToFile{ RetrieveLocalFolderPath().c_str() + c_backSlash + notificationAppId + c_pngExtension };
```
hulumane, line 284: Rename to iconFilePath.

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -300,8 +346,8 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring
     // 3. Use the default assets.
     Microsoft::Windows::AppNotifications::Helpers::AppNotificationAssets assets{};
 
-    if (FAILED(RetrieveAssetsFromProcess(assets)) &&
-        FAILED(ToastNotifications_RetrieveAssets_Stub(assets)))
+    if (FAILED(ToastNotifications_RetrieveAssets_Stub(assets)) &&
```
hulumane, line 349: Rename to RetrieveAssetsFromShortcut

hulumane: That keeps naming consistent with the *FromProcess function.

dev/AppNotifications/WICUtility.cpp
```diff
@@ -0,0 +1,165 @@
...
+    winrt::com_ptr<IWICBitmapEncoder> const& wicEncoder,
+    winrt::com_ptr<IWICBitmapSource> const& wicBitmapSource,
+    BITMAP_VERSION bmpv)
+{
+    winrt::com_ptr<IWICBitmapFrameEncode> wicFrameEncoder;
+    winrt::com_ptr<IPropertyBag2> wicEncoderOptions;
+
+    THROW_IF_FAILED(wicEncoder->CreateNewFrame(wicFrameEncoder.put(), wicEncoderOptions.put()));
+
+    GUID containerGuid;
```
hulumane, line 52: winrt::guid?

dev/AppNotifications/AppNotificationUtility.cpp
DrusTheAxe: Recommend not naming anything `Utility`. That plus `Helper` are trigger words for me :P

Functions here all seem related to localizable data - icons, etc. `ShellLocalization.h`?

hulumane: That sounds good to me!

dev/AppNotifications/WICUtility.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+    THROW_IF_FAILED(wicFrameEncoder->Initialize(wicEncoderOptions.get()));
+
+    UINT uWidth, uHeight;
+    THROW_IF_FAILED(wicBitmapSource->GetSize(&uWidth, &uHeight));
+
+    THROW_IF_FAILED(wicFrameEncoder->SetSize(uWidth, uHeight));
+
+    winrt::com_ptr<IWICBitmapSource> wicbitmapSourceConverted;
+    wicbitmapSourceConverted = ConvertWICBitmapPixelFormat(wicImagingFactory, wicBitmapSource, GUID_WICPixelFormat32bppBGRA, WICBitmapDitherTypeNone);
```
hulumane, line 80: Nit: Initialize and declare above please.

dev/AppNotifications/WICUtility.cpp
```diff
@@ -0,0 +1,165 @@
...
+}
+
+void SaveImageWithWIC(
+    winrt::com_ptr<IWICImagingFactory> const& wicImagingFactory,
+    winrt::com_ptr<IWICBitmapSource> const& wicBitmapSource,
+    _In_ REFGUID guidContainerFormat,
+    winrt::com_ptr<IStream>& pStream)
+{
+    winrt::com_ptr<IStream> spImageStream;
+    spImageStream = GetStreamOfWICBitmapSource(wicImagingFactory, wicBitmapSource, guidContainerFormat, BMPV_1);
```
hulumane, line 124: Nit: Declare and init in same line

dev/AppNotifications/WICUtility.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+    // Seek the stream to the beginning and transfer
+    static LARGE_INTEGER const lnBeginning{ 0 };
```
hulumane, line 127: Can you explain why this is static? Is it because we are calling this in some loop somehow and need to maintain state?

dev/AppNotifications/WICUtility.h
hulumane: Per @DrusTheAxe comment above,  this should be called ShellLocalization.h

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -232,6 +240,89 @@ std::wstring Microsoft::Windows::AppNotifications::Helpers::SetDisplayNameBasedO
...
+{
+    wil::unique_hicon hIcon{ RetrieveIconFromProcess() };
+
+    std::wstring notificationAppId{ RetrieveNotificationAppId() };
+
+    // path: C:\Users\<currentUser>\AppData\Local\Microsoft\WindowsAppSDK\{AppGUID}.png
+    std::wstring writeToFile{ RetrieveLocalFolderPath().c_str() + c_backSlash + notificationAppId + c_pngExtension };
+    Microsoft::Windows::AppNotifications::WICHelpers::WriteHIconToPngFile(hIcon, writeToFile.c_str());
+
+    assets.displayName = Microsoft::Windows::AppNotifications::Helpers::GetDisplayNameBasedOnProcessName();
```
hulumane, line 287: AppNotificationManager::UnregisterAll() should also delete this file if it exists. Can you please add the relevant code to cleanup there as well? Otherwise we have a leak.

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -232,6 +240,89 @@ std::wstring Microsoft::Windows::AppNotifications::Helpers::SetDisplayNameBasedO
...
+    assets.iconFilePath = writeToFile;
+
+    return S_OK;
```
hulumane, line 290: @danielayala94 A comment is needed here with a link to the task for cleaning up the file in case UnregisterAll is not called.

danielayala94: Sure thing!

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -237,14 +240,57 @@ std::wstring Microsoft::Windows::AppNotifications::Helpers::GetDisplayNameBasedO
...
+}
+CATCH_RETURN()
+
 // Do nothing. This is just a placeholder while the UDK is ingested with the proper API.
 HRESULT ToastNotifications_RetrieveAssets_Stub(_Out_ Microsoft::Windows::AppNotifications::Helpers::AppNotificationAssets& /*assets*/)
```
danielayala94, line 295: I'll change the name of this function

dev/AppNotifications/AppNotificationUtility.cpp
danielayala94: #2343

dev/AppNotifications/WICUtility.cpp
danielayala94: Unfortunately, a `winrt::guid` cannot be passed in the place of a GUID*, so will leave it as is.

dev/AppNotifications/WICUtility.cpp
```diff
@@ -0,0 +1,165 @@
...
+    static const LARGE_INTEGER lnBeginning = {};
+    THROW_IF_FAILED(spImageStream->Seek(lnBeginning, STREAM_SEEK_SET, nullptr /* new seek pointer */));
+
+    return spImageStream;
+}
+
+void SaveImageWithWIC(
+    winrt::com_ptr<IWICImagingFactory> const& wicImagingFactory,
+    winrt::com_ptr<IWICBitmapSource> const& wicBitmapSource,
+    _In_ REFGUID guidContainerFormat,
```
DrusTheAxe, line 120: NIT: Replace `_In_ REFGUID` with `const GUID&`

REFGUID is old school way to say `const GUID&`

And const& is a readonly input parameter so `_In_` is redundant with language semantics

dev/AppNotifications/WICUtility.cpp
DrusTheAxe, line 124: Merge lines 123+124

```
winrt::com_ptr<IStream> spImageStream{ GetStreamOfWICBitmapSource(wicImagingFactory... };
```

dev/AppNotifications/WICUtility.cpp
```diff
@@ -0,0 +1,165 @@
...
+    winrt::com_ptr<IWICBitmapSource> wicBitmapSource;
+    wicBitmapSource = wicBitmap;
+
+    SaveImageWithWIC(wicImagingFactory, wicBitmapSource, GUID_ContainerFormatPng, spStream);
+
+    // Write out the stream to a file.
+    winrt::com_ptr<IStream> spStreamOut;
+    THROW_IF_FAILED(SHCreateStreamOnFileEx(pszFileName, STGM_CREATE | STGM_WRITE | STGM_SHARE_DENY_WRITE, FILE_ATTRIBUTE_NORMAL, TRUE, nullptr, spStreamOut.put()));
+
+    STATSTG statstg;
```
DrusTheAxe, line 154: Always ensure variables are initialized. Any not-class (i.e. anything without a default constructor) isn't initialized unless you help

```c++
STATSTG statstg{};
```

dev/AppNotifications/WICUtility.cpp
```diff
@@ -0,0 +1,165 @@
...
+    THROW_IF_FAILED(spStream->Stat(&statstg, STATFLAG_NONAME));
+
+    THROW_IF_FAILED(IStream_Reset(spStream.get()));
+
+    THROW_IF_FAILED(spStreamOut->SetSize(statstg.cbSize));
+
+    // TODO: Comments need to be added
```
DrusTheAxe, line 161: ?

test/TestApps/ToastNotificationsDemoApp/ToastNotificationsDemoApp.vcxproj
```diff
@@ -285,4 +292,4 @@
   <Target Name="CopyFiles" AfterTargets="AfterBuild">
     <Copy SourceFiles="$(OutDir)\..\WindowsAppRuntime_BootstrapDLL\Microsoft.WindowsAppRuntime.Bootstrap.dll" DestinationFolder="$(OutDir)" />
   </Target>
-</Project> 
\ No newline at end of file
+</Project>
```
DrusTheAxe, line 295: NIT: GitHub doesn't line files ending without a newline

test/TestApps/ToastNotificationsDemoApp/ToastNotificationsDemoApp.vcxproj.filters
```diff
@@ -14,4 +14,33 @@
...
+  <ItemGroup>
+    <ResourceCompile Include="ToastNotificationsDemoApp.rc">
+      <Filter>Resource Files</Filter>
+    </ResourceCompile>
+  </ItemGroup>
+  <ItemGroup>
+    <Image Include="icon1.ico">
+      <Filter>Resource Files</Filter>
+    </Image>
+  </ItemGroup>
```
DrusTheAxe, line 45: NIT: GitHub diff doesn't like files ending without a newline

dev/AppNotifications/WICUtility.cpp
```diff
@@ -0,0 +1,165 @@
...
+        VARIANT varValue{};
+
+        varValue.vt = VT_BOOL;
+        varValue.boolVal = VARIANT_TRUE;
+
+        // Options to enable the v5 header support for 32bppBGRA.
+        PROPBAG2 v5HeaderOption{};
+
+        std::wstring str = L"EnableV5Header32bppBGRA";
+        v5HeaderOption.pstrName = (LPOLESTR)str.c_str();
```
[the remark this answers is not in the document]
danielayala94: I tried, but the compiler doesn't like using any of the modern cast approaches.

dev/AppNotifications/ShellLocalization.cpp
```diff
@@ -0,0 +1,265 @@
...
+CATCH_RETURN()
+
+void Microsoft::Windows::AppNotifications::ShellLocalization::DeleteIconFromCache() noexcept try
+{
+    std::wstring notificationAppId{ Microsoft::Windows::AppNotifications::Helpers::RetrieveNotificationAppId() };
+
+    // path: C:\Users\<currentUser>\AppData\Local\Microsoft\WindowsAppSDK\{AppGUID}.png
+    std::wstring iconFilePath{ RetrieveLocalFolderPath().c_str() + c_backSlash + notificationAppId + c_pngExtension };
+
+    THROW_IF_WIN32_BOOL_FALSE(DeleteFile(iconFilePath.c_str()));
```
hulumane, line 263: For the first time, I am going to have to say that this needs to be best effort. ie. If the file is in use or of the file is absent for some reason, please ignore. Thoughts?

hulumane: But we should LOG_IF_FALSE for sure.

hulumane: Also it may be better to log why the delete failed. You will need to look for the win32 error:https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-deletefilea

dev/AppNotifications/ShellLocalization.h
```diff
@@ -0,0 +1,31 @@
...
+        std::wstring iconFilePath;
+    };
+
+    HRESULT RetrieveAssetsFromProcess(_Out_ Microsoft::Windows::AppNotifications::ShellLocalization::AppNotificationAssets& assets) noexcept;
+
+    HRESULT RetrieveAssetsFromShortcut(_Out_ Microsoft::Windows::AppNotifications::ShellLocalization::AppNotificationAssets& /*assets*/);
+
+    HRESULT RetrieveDefaultAssets(_Out_ Microsoft::Windows::AppNotifications::ShellLocalization::AppNotificationAssets& assets) noexcept;
+
+    void DeleteIconFromCache() noexcept;
```
hulumane, line 30: This should throw IMHO. In general HRESULT APIs don't throw. Everything else does.

dev/AppNotifications/ShellLocalization.cpp
```diff
@@ -0,0 +1,265 @@
...
+HRESULT Microsoft::Windows::AppNotifications::ShellLocalization::RetrieveDefaultAssets(_Out_ Microsoft::Windows::AppNotifications::ShellLocalization::AppNotificationAssets& assets) noexcept try
+{
+    assets.displayName = Microsoft::Windows::AppNotifications::Helpers::GetDisplayNameBasedOnProcessName();
+    assets.iconFilePath = L""; // No icon filepath for now. This is coming in a future FrameworkUdk ingestion.
+
+    return S_OK;
+}
+CATCH_RETURN()
+
+void Microsoft::Windows::AppNotifications::ShellLocalization::DeleteIconFromCache() noexcept try
```
hulumane, line 256: Don't swallow exceptions in methods that don't return HRESULT. Preferrably handle the exception at the caller by ignoring it, logging it or rethrow.

dev/AppNotifications/ShellLocalization.h
```diff
@@ -0,0 +1,31 @@
...
+    const std::wstring c_pngExtension{ LR"(.png)" };
+
+    struct AppNotificationAssets {
+        std::wstring displayName;
+        std::wstring iconFilePath;
+    };
+
+    HRESULT RetrieveAssetsFromProcess(_Out_ Microsoft::Windows::AppNotifications::ShellLocalization::AppNotificationAssets& assets) noexcept;
+
+    HRESULT RetrieveAssetsFromShortcut(_Out_ Microsoft::Windows::AppNotifications::ShellLocalization::AppNotificationAssets& /*assets*/);
```
hulumane, line 26: noexcept

dev/AppNotifications/ShellLocalization.cpp
```diff
@@ -0,0 +1,265 @@
...
+    std::wstring notificationAppId{ Microsoft::Windows::AppNotifications::Helpers::RetrieveNotificationAppId() };
+
+    // path: C:\Users\<currentUser>\AppData\Local\Microsoft\WindowsAppSDK\{AppGUID}.png
+    std::wstring iconFilePath{ RetrieveLocalFolderPath().c_str() + c_backSlash + notificationAppId + c_pngExtension };
+    WriteHIconToPngFile(hIcon, iconFilePath.c_str());
+
+    assets.displayName = Microsoft::Windows::AppNotifications::Helpers::GetDisplayNameBasedOnProcessName();
+    assets.iconFilePath = iconFilePath;
+
+    // TODO: Clean up the icon file in case UnregisterAll wasn't called.
```
hulumane, line 233: Please change this comment. Cleanup isn't really done here. Some mapping needs to be done here that will clean it up based on a schedule.

dev/AppNotifications/ShellLocalization.cpp
danielayala94: Actually, I'm using the API wrongly. It returns non-zero if successful, otherwise it's a failure and should use GetLastError to know what the issue is: https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-deletefilew

danielayala94: I would say, let DeleteFile to throw and let the caller decide if propagate the error or log it.

hulumane: Then why don't you check if file exists before you call delete? Then caller can probably treat this as a genuine error.

dev/AppNotifications/ShellLocalization.h
danielayala94: I changed the return type to HRESULT.

dev/AppNotifications/ShellLocalization.cpp
```diff
@@ -0,0 +1,272 @@
...
+}
+
+HRESULT Microsoft::Windows::AppNotifications::ShellLocalization::RetrieveAssetsFromProcess(_Out_ Microsoft::Windows::AppNotifications::ShellLocalization::AppNotificationAssets& assets) noexcept try
+{
+    wil::unique_hicon hIcon{ RetrieveIconFromProcess() };
+
+    std::wstring notificationAppId{ Microsoft::Windows::AppNotifications::Helpers::RetrieveNotificationAppId() };
+
+    // path: C:\Users\<currentUser>\AppData\Local\Microsoft\WindowsAppSDK\{AppGUID}.png
+    std::wstring iconFilePath{ RetrieveLocalFolderPath().c_str() + c_backSlash + notificationAppId + c_pngExtension };
```
hulumane, line 227: c_str() +? That works?

dev/AppNotifications/ShellLocalization.cpp
```diff
@@ -0,0 +1,272 @@
...
+
+HRESULT Microsoft::Windows::AppNotifications::ShellLocalization::DeleteIconFromCache() noexcept try
+{
+    std::wstring notificationAppId{ Microsoft::Windows::AppNotifications::Helpers::RetrieveNotificationAppId() };
+
+    // path: C:\Users\<currentUser>\AppData\Local\Microsoft\WindowsAppSDK\{AppGUID}.png
+    std::wstring iconFilePath{ RetrieveLocalFolderPath().c_str() + c_backSlash + notificationAppId + c_pngExtension };
+
+    // If DeleteFile returned FALSE, then deletion failed and we should return the corresponding error code.
+    if (DeleteFile(iconFilePath.c_str()) == FALSE)
```
hulumane, line 265: This does not return TRUE or FALSE. It returns 0 or non-zero

hulumane: if(!DeleteFile(iconFilePath.c_str())
{
}

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -331,12 +262,45 @@ winrt::guid Microsoft::Windows::AppNotifications::Helpers::RegisterComActivatorG
...
+    Microsoft::Windows::AppNotifications::ShellLocalization::AppNotificationAssets assets{};
+
+    if (FAILED(Microsoft::Windows::AppNotifications::ShellLocalization::RetrieveAssetsFromShortcut(assets)) &&
+        FAILED(Microsoft::Windows::AppNotifications::ShellLocalization::RetrieveAssetsFromProcess(assets)))
+    {
+        THROW_IF_FAILED(Microsoft::Windows::AppNotifications::ShellLocalization::RetrieveDefaultAssets(assets));
+    }
+
+    RegisterValue(hKey, L"DisplayName", reinterpret_cast<const BYTE*>(assets.displayName.c_str()), REG_EXPAND_SZ, assets.displayName.size() * sizeof(wchar_t));
+    RegisterValue(hKey, L"IconUri", reinterpret_cast<const BYTE*>(assets.iconFilePath.c_str()), REG_EXPAND_SZ, assets.iconFilePath.size() * sizeof(wchar_t));
```
hulumane, line 302: We should check if the icon is empty. In the default case it is, so we probably shouldnt populate this registry.

dev/AppNotifications/ShellLocalization.cpp
```diff
@@ -0,0 +1,272 @@
...
+
+    THROW_IF_FAILED(wicEncoder->Initialize(spImageStream.get(), WICBitmapEncoderCacheOption::WICBitmapEncoderNoCache));
+
+    // Add a single frame to the encoder with the Bitmap
+    AddFrameToWICBitmap(wicImagingFactory, wicEncoder, wicBitmapSource, bitmapVersion);
+
+    THROW_IF_FAILED(wicEncoder->Commit());
+
+    // Seek the stream to the beginning and transfer
+    static const LARGE_INTEGER lnBeginning = {};
```
hulumane, line 116: Why static?

dev/AppNotifications/ShellLocalization.cpp
danielayala94: The function returns BOOL. It's a bit odd that in the doc, they say things like "If the function fails, the return value is zero (0)" instead of saying TRUE/FALSE.

dev/AppNotifications/ShellLocalization.cpp
```diff
@@ -0,0 +1,272 @@
...
+{
+    // THROW_HR_IF_MSG(E_UNEXPECTED, IsIconFileExtensionSupported(iconFilePath));
+
+    return E_NOTIMPL;
+}
+
+HRESULT Microsoft::Windows::AppNotifications::ShellLocalization::RetrieveDefaultAssets(_Out_ Microsoft::Windows::AppNotifications::ShellLocalization::AppNotificationAssets& assets) noexcept try
+{
+    assets.displayName = Microsoft::Windows::AppNotifications::Helpers::GetDisplayNameBasedOnProcessName();
+    assets.iconFilePath = L""; // No icon filepath for now. This is coming in a future FrameworkUdk ingestion.
```
hulumane, line 251: Not needed. Strings are empty by default.

dev/AppNotifications/ShellLocalization.cpp
danielayala94: @sharath2727 can you help with this?

test/TestApps/ToastNotificationsDemoApp/main.cpp
```diff
@@ -175,7 +148,8 @@ int main()
         const PACKAGE_VERSION minVersion{};
         RETURN_IF_FAILED(MddBootstrapInitialize(c_Version_MajorMinor, nullptr, minVersion));
 
-        SetDisplayNameAndIcon();
+        // Not mandatory, but it's highly recommended to specify AppUserModelId
+        THROW_IF_FAILED(SetCurrentProcessExplicitAppUserModelID(L"TestAppId"));
```
hulumane, line 152: @vaheeshta This is needed if we want apps to maintain their toast states even after a process name update later in the future. For example, if unpackaged app foo.exe has a name change to bar.exe in the future, this ensures that the apps toasts remain in the actioncentre. We should document.

dev/AppNotifications/ShellLocalization.cpp
danielayala94: Yes. But we don't really need to call c_str().

test/TestApps/ToastNotificationsDemoApp/main.cpp
hulumane: It also is needed if apps change their path from c:\foo\name.exe to d:\bar\name.exe.

dev/AppNotifications/ShellLocalization.cpp
```diff
@@ -0,0 +1,272 @@
...
+    const auto fileExtension = iconFilePath.extension();
+
+    std::string lowercaseFileExtension{ fileExtension.u8string() };
+
+    std::transform(lowercaseFileExtension.begin(), lowercaseFileExtension.end(), lowercaseFileExtension.begin(),
+        [](unsigned char c) { return static_cast<char>(std::tolower(c)); });
+
+    const bool isFileExtensionSupported =
+        lowercaseFileExtension == ".ico" || lowercaseFileExtension == ".bmp" || lowercaseFileExtension == ".jpg" || lowercaseFileExtension == ".png";
+
```
DrusTheAxe, line 181: Instead of converting ( copying) the string to UTF8 and then lowercase (and that's doing a lexical conversion) better to use CompareStringOrdinal

```c++
static PCWSTR c_supportedExtensions[]{ L".bmp", L".ico", L".jpg", L".png" };
...
const auto fileExtension{ iconFilePath.extension() };
const auto extension{ fileExtension.c_str() };
for (var supportedExtension : c_supportedExtensions)
{
    if (CompareStringOrdinal(extension, -1, supportedExtension, -1, TRUE) == CSTR_EQUAL)
    {
        return true;
    }
}
return true;
```

dev/AppNotifications/AppNotificationUtility.cpp
danielayala94: I have verified there are no issues if we write an empty string as the IconUri. There won't be an icon in the toast.

Let me verify if I can see the same behavior if we don't set IconUri at all.

dev/AppNotifications/ShellLocalization.cpp
```diff
@@ -0,0 +1,272 @@
...
+    THROW_IF_FAILED(IStream_Reset(spStream.get()));
+
+    THROW_IF_FAILED(spStreamOut->SetSize(statstg.cbSize));
+
+    THROW_IF_FAILED(spStream->CopyTo(spStreamOut.get(), statstg.cbSize, nullptr /* pointer to number of bytes read */, nullptr /* pointer to number of bytes written */));
+
+    THROW_IF_FAILED(spStreamOut->Commit(STGC_DANGEROUSLYCOMMITMERELYTODISKCACHE));
+}
+
+HRESULT IsIconFileExtensionSupported(std::filesystem::path const& iconFilePath)
```
hulumane, line 170: noexcept try catch return

dev/AppNotifications/ShellLocalization.cpp
```diff
@@ -0,0 +1,272 @@
...
+    std::path localFolderPath{ std::wstring(localAppDataPath.get()) +
+        Microsoft::Windows::AppNotifications::Helpers::c_localMicrosoftFolder +
+        Microsoft::Windows::AppNotifications::Helpers::c_localWindowsAppSDKFolder };
+
+    if (!std::exists(localFolderPath))
+    {
+        std::create_directory(localFolderPath);
+    }
+
+    return std::wstring(localFolderPath.c_str());
```
DrusTheAxe, line 201: `return localFolderPath.wstring();`

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -220,6 +222,11 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
 
             if (!AppModel::Identity::IsPackagedProcess())
             {
+                // If the app icon was inferred from process, then we should clean it up.
+                // Do not fail this function if such a file doesn't exist.
```
hulumane, line 226: I think we should mention that the file may not exist in cache if the icon was retrieved from Shortcut or was the default icon. So this is more of best-effort check

dev/AppNotifications/ShellLocalization.cpp
```diff
@@ -0,0 +1,272 @@
...
+{
+    wil::unique_cotaskmem_string localAppDataPath;
+    THROW_IF_FAILED(SHGetKnownFolderPath(FOLDERID_LocalAppData, 0 /* flags */, nullptr /* access token handle */, &localAppDataPath));
+
+    // path: C:\Users\<currentUser>\AppData\Local\Microsoft\WindowsAppSDK
+    std::path localFolderPath{ std::wstring(localAppDataPath.get()) +
+        Microsoft::Windows::AppNotifications::Helpers::c_localMicrosoftFolder +
+        Microsoft::Windows::AppNotifications::Helpers::c_localWindowsAppSDKFolder };
+
+    if (!std::exists(localFolderPath))
```
hulumane, line 196: I'm curious why we need to create a directory at all here. IF the file is saved, wouldn't the directory automatically be created? So this looks like overhead.

dev/AppNotifications/ShellLocalization.cpp
DrusTheAxe, line 227: lots of string concatenation makes me sad
Each + is a new copy so lots of malloc+free+memcpy
std::format is your friend...if you've got C++20. We're currently C++17 so wil::str_printf is your friend

BUT...you've building a path. `std::filesystem::path` is actually your friend here

```
auto iconFilePath{ std::filesystem::path(RetrieveLocalFolderPath()) / (notificationAppId + c_pngExtension) };
```

Though it'd be even better if `RetrieveLocalFolderPath` returned `std::filesystem::path`. After all it's called RetrieveLocalFolder**Path** not RetrieveLocalFolder**String** :-)

dev/AppNotifications/ShellLocalization.cpp
hulumane, line 201: just return localfolderPath.c_str(). The string conversion is automatically done!

hulumane: Or maybe you can get away with simply return localFolderPath if string has been overloaded for std::path types.

dev/AppNotifications/ShellLocalization.cpp
```diff
@@ -0,0 +1,272 @@
...
+    assets.iconFilePath = iconFilePath;
+
+    // TODO: In case we are caching icons from uninstalled apps who didn't call UnregisterAll,
+    // we should make an effort to clean them up in an opportunistic way.
+
+    return S_OK;
+}
+CATCH_RETURN()
+
+// Do nothing. This is just a placeholder while the UDK is ingested with the proper API.
```
DrusTheAxe, line 240: So this will be updated when the UDK's updated?

Move the comment inside the function and change it to `//TODO: This is a placeholder until the UDK is ingested with the proper API`

Even better if there's an ADO item for that work make it `//TODO:12345 Use UDK API to retrieve icon` or whatever the ADO item # and title are

dev/AppNotifications/ShellLocalization.cpp
DrusTheAxe: BTW if it were needed assets.icconFilePath.clear() would be better

dev/AppNotifications/ShellLocalization.cpp
```diff
@@ -0,0 +1,272 @@
...
+    return S_OK;
+}
+CATCH_RETURN()
+
+HRESULT Microsoft::Windows::AppNotifications::ShellLocalization::DeleteIconFromCache() noexcept try
+{
+    std::wstring notificationAppId{ Microsoft::Windows::AppNotifications::Helpers::RetrieveNotificationAppId() };
+
+    // path: C:\Users\<currentUser>\AppData\Local\Microsoft\WindowsAppSDK\{AppGUID}.png
+    std::wstring iconFilePath{ RetrieveLocalFolderPath().c_str() + c_backSlash + notificationAppId + c_pngExtension };
```
DrusTheAxe, line 262: See other comment re std::filesystem::path

dev/AppNotifications/AppNotificationUtility.cpp
hulumane: Well if the behavior is the same, I would keep it simple then and just have this as empty.

dev/AppNotifications/ShellLocalization.h
```diff
@@ -0,0 +1,31 @@
...
+
+#include "pch.h"
+
+// In this file we define methods that write/return Shell assets, like DisplayName and icon URI.
+
+namespace Microsoft::Windows::AppNotifications::ShellLocalization
+{
+    const std::wstring c_appIdentifierPath{ LR"(Software\Classes\AppUserModelId\)" };
+    const std::wstring c_quote{ LR"(")" };
+    const std::wstring c_backSlash{ LR"(\)" };
```
DrusTheAxe, line 14: Why? If this is just used to do string concatenation of paths you should use instead std::filesystem::path

dev/AppNotifications/ShellLocalization.h
```diff
@@ -0,0 +1,31 @@
...
+    const std::wstring c_localMicrosoftFolder{ LR"(\Microsoft)" };
+    const std::wstring c_localWindowsAppSDKFolder{ LR"(\WindowsAppSDK)" };
+    const std::wstring c_pngExtension{ LR"(.png)" };
```
DrusTheAxe, line 17: Why `wstring`? `wstring` is a deep copy. Malloc+memcpy+free, when it's a compile-time known constant.

For string constants like these you should use `PCWSTR`


You can use `wstring_view` in some cases but I can never remember the right syntax to get equally zero-runtime compile-time construction. `constexpr std::wstring_view foo{ L"bar" };` ?

DrusTheAxe (dismissed):

dev/AppNotifications/ShellLocalization.cpp
danielayala94: > So this will be updated when the UDK's updated?

Yes. But it's not blocking the PR.

The change has been committed, now we are waiting for a fresh UDK that contains the API.

dev/AppNotifications/AppNotificationUtility.cpp
danielayala94: No IconUri means that Shell will render the default icon 😊.

dev/AppNotifications/ShellLocalization.cpp
danielayala94: @DrusTheAxe there is much to learn from you! (or at least read the C++ docs 😅).

dev/AppNotifications/ShellLocalization.cpp
danielayala94: @sharath2727 ?

dev/AppNotifications/ShellLocalization.cpp
hulumane, line 251: Add a comment here that we need to set this to empty string because the OS will use this hint to display a default icon.

dev/AppNotifications/AppNotificationUtility.cpp
hulumane: Then you need to add a comment for that here. It isn't obvious.

danielayala94: Yeah, of course. I added a comment right before writing to the registry.

dev/AppNotifications/ShellLocalization.cpp
sharath2727: We don't need the static. We are only dealing with one frame to convert to png. Got rid of it. Thanks.

dev/AppNotifications/ShellLocalization.cpp
sharath2727: The API WriteHICONtoPng when it takes in a filepath as an argument doesn't create windowsappsdk folder (if it doesn't exist already) and throws if the path doesn't exist. So that is why we have this overhead of making sure the windowsappsdk folder exists in the local path.

test/TestApps/ToastNotificationsDemoApp/main.cpp
vaheeshta: Got it. I'll document this once I have the unpackaged sample code.

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -331,12 +264,51 @@ winrt::guid Microsoft::Windows::AppNotifications::Helpers::RegisterComActivatorG
...
+
+    if (FAILED(Microsoft::Windows::AppNotifications::ShellLocalization::RetrieveAssetsFromShortcut(assets)) &&
+        FAILED(Microsoft::Windows::AppNotifications::ShellLocalization::RetrieveAssetsFromProcess(assets)))
+    {
+        assets.displayName = GetDisplayNameBasedOnProcessName();
+    }
+
+    RegisterValue(hKey, L"DisplayName", reinterpret_cast<const BYTE*>(assets.displayName.c_str()), REG_EXPAND_SZ, assets.displayName.size() * sizeof(wchar_t));
+
+    // If IconUri is empty, then Shell will render a default icon in AppNotifications.
```
hulumane, line 305: Can we change this statement please? IT should be:
If Icon is not specified in the Registry, OS will render a default icon for the app

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -331,12 +264,51 @@ winrt::guid Microsoft::Windows::AppNotifications::Helpers::RegisterComActivatorG
...
+    if (!assets.iconFilePath.empty())
+    {
+        RegisterValue(hKey, L"IconUri", reinterpret_cast<const BYTE*>(assets.iconFilePath.c_str()), REG_EXPAND_SZ, assets.iconFilePath.size() * sizeof(wchar_t));
```
hulumane, line 308: What would happen the second time if the App Registers but there is no icon specified?
I think we need to clear the registry of all these assets before we repopulate it.

hulumane: Otherwise we run the risk of having a stale icon from a prior registration.

hulumane: One solution is to delete the icon from registry if the iconFilePath is empty

hulumane (approved):

dev/AppNotifications/ShellLocalization.cpp
```diff
@@ -0,0 +1,272 @@
...
+    WICBitmapDitherType bitmapDitherType)
+{
+    winrt::com_ptr<IWICFormatConverter> wicFormatConverter;
+    THROW_IF_FAILED(wicImagingFactory->CreateFormatConverter(wicFormatConverter.put()));
+
+    THROW_IF_FAILED(wicFormatConverter->Initialize(wicBitmapSource.get(), guidPixelFormatSource, bitmapDitherType, nullptr, 0.f, WICBitmapPaletteTypeCustom));
+
+    // Store the converted bitmap as ppToRenderBitmapSource
+    winrt::com_ptr<IWICBitmapSource> wicBitmapSourceConverted;
+    THROW_IF_FAILED(wicFormatConverter->QueryInterface(_uuidof(IWICBitmapSource), wicBitmapSourceConverted.put_void()));
```
jonwis, line 44: (No change required)

I _think_ this can be `return wicFormatConverter.as<IWICBitmapSource>()` ?

dev/AppNotifications/ShellLocalization.cpp
```diff
@@ -0,0 +1,272 @@
...
+    winrt::com_ptr<IWICBitmapSource> const& wicBitmapSource,
+    WICPixelFormatGUID guidPixelFormatSource,
+    WICBitmapDitherType bitmapDitherType)
+{
+    winrt::com_ptr<IWICFormatConverter> wicFormatConverter;
+    THROW_IF_FAILED(wicImagingFactory->CreateFormatConverter(wicFormatConverter.put()));
+
+    THROW_IF_FAILED(wicFormatConverter->Initialize(wicBitmapSource.get(), guidPixelFormatSource, bitmapDitherType, nullptr, 0.f, WICBitmapPaletteTypeCustom));
+
+    // Store the converted bitmap as ppToRenderBitmapSource
```
jonwis, line 42: Dead comment, please remove

jonwis (approved):

PR #2299, opened 2022-03-18
Add script for install/uninstall test sdk packages

pmpurifoy:
Script for inner loop testing to make install/uninstalling the main packages for E2E testing faster instead of individually adding/removing packages and having to kill active SDK processes in the task manager.

Packages installed/uninstalled:
- Microsoft.WindowsAppRuntime.Framework.msix
- DynamicDependency.DataStore.msix
- DynamicDependencyLifetimeManager.msix
- PushNotificationsLongRunningTask.msix


SDKPackageManager.ps1
```diff
@@ -0,0 +1,31 @@
+param([Parameter(Mandatory=$true)][string]$arch, [Parameter(Mandatory=$true)][string]$config, [Parameter(Mandatory=$true)][string]$task)
+
+$currentLocation = Get-Location
+$packagePath = "$currentLocation\BuildOutput\$config\$arch\"
+
+if(!$task.CompareTo("i"))
+{
+	write-host "installing packages"
+	add-appxpackage "$packagePath\Microsoft.WindowsAppRuntime.Framework\Microsoft.WindowsAppRuntime.Framework.msix"
```
danielayala94, line 9: I think this can be useful for all devs working in any WinAppSDK feature, not only Notifications.

If anyone works in PowerNotifications, why would they want to install the LRP?

Is it LRP the only other extra package that is installed to test any of the other WinAppSDK features?

SDKPackageManager.ps1
```diff
@@ -0,0 +1,31 @@
+param([Parameter(Mandatory=$true)][string]$arch, [Parameter(Mandatory=$true)][string]$config, [Parameter(Mandatory=$true)][string]$task)
+
+$currentLocation = Get-Location
```
danielayala94, line 3: I would move this to Tools, instead of hosting the script in the root folder.

SDKPackageManager.ps1
```diff
@@ -0,0 +1,31 @@
...
+$packagePath = "$currentLocation\BuildOutput\$config\$arch\"
+
+if(!$task.CompareTo("i"))
+{
+	write-host "installing packages"
+	add-appxpackage "$packagePath\Microsoft.WindowsAppRuntime.Framework\Microsoft.WindowsAppRuntime.Framework.msix"
+	add-appxpackage "$packagePath\DynamicDependency.DataStore.Msix\DynamicDependency.DataStore.msix"
```
danielayala94, line 10: Maybe print that the script installed this and that package?

pmpurifoy:
Need to parameter for demo app package and move to tools folder

SDKPackageManager.ps1
```diff
@@ -0,0 +1,31 @@
...
+	add-appxpackage "$packagePath\PushNotificationsLongRunningTask.Msix\PushNotificationsLongRunningTask.msix"
+}
+elseif(!$task.CompareTo("u"))
+{
+	taskkill /IM PushNotificationsLongRunningTask.exe /F
+	taskkill /IM DynamicDependency.DataStore.exe /F
+	taskkill /IM DynamicDependencyLifetimeManager.exe /F
+
+	write-host "uninstalling packages"
+	remove-appxpackage "WindowsAppRuntime.Test.DDLM-4.1.1967.333-$($arch)_4.1.1967.333_$($arch)__8wekyb3d8bbwe"
```
danielayala94, line 21: Is there a way to have a unit test that verifies the script runs successfully? The minute any package name is changed, the script will be broken.

SDKPackageManager.ps1
```diff
@@ -0,0 +1,31 @@
+param([Parameter(Mandatory=$true)][string]$arch, [Parameter(Mandatory=$true)][string]$config, [Parameter(Mandatory=$true)][string]$task)
+
+$currentLocation = Get-Location
+$packagePath = "$currentLocation\BuildOutput\$config\$arch\"
+
+if(!$task.CompareTo("i"))
```
danielayala94, line 6: We need some welcome message saying what can we do with this script.

danielayala94 (changes requested):
Great to see this script coming! Thanks @pmpurifoy for bringing this on! 😊

I encourage you to reach out to other devs and see how they can leverage on this script.

SDKPackageManager.ps1
danielayala94: Adding options to install the demo packages would be wonderful!

SDKPackageManager.ps1
```diff
@@ -0,0 +1,31 @@
+param([Parameter(Mandatory=$true)][string]$arch, [Parameter(Mandatory=$true)][string]$config, [Parameter(Mandatory=$true)][string]$task)
```
DrusTheAxe, line 1: Please don't put this in the root directory
1. we'd like to _reduce_ what we have in the root, not increase it (some are expected to move, just hasn't happened yet)
2. this is related to the test packages, specific to the `\test` tree

`\test\tools` seems appropos

SDKPackageManager.ps1
DrusTheAxe, line 6: "i" and "u"? Not very powershell-esque

`--Install` and `--Uninstall` are more appropos.

```
Param(
    [Switch]$Install=$false,
    [Switch]$Unnstall=$false
)

if (($Install -eq $false) -And $($Uninstall -eq $false)) 
{
    Write-Output "Invalid arguments, aborting..."
    Exit 1
}
```

You could make 1 parameter as an enum e.g. 

```
Param(
    [ValidateSet("Install", "Uninstall")]
    [String]$Task
)
```

though this "-foo bar" feels a bit more irksome than the style above "-bar"

See `tools\TerminalVelocity\Generator-TerminalVelocityFeatures.ps1` for more ValidateSet examples

DrusTheAxe, line 6: NIT: `if (...`

space between `if` and parenthesis

SDKPackageManager.ps1
```diff
@@ -0,0 +1,31 @@
...
+$currentLocation = Get-Location
+$packagePath = "$currentLocation\BuildOutput\$config\$arch\"
+
+if(!$task.CompareTo("i"))
+{
+	write-host "installing packages"
+	add-appxpackage "$packagePath\Microsoft.WindowsAppRuntime.Framework\Microsoft.WindowsAppRuntime.Framework.msix"
+	add-appxpackage "$packagePath\DynamicDependency.DataStore.Msix\DynamicDependency.DataStore.msix"
+	add-appxpackage "$packagePath\DynamicDependencyLifetimeManager.Msix\DynamicDependencyLifetimeManager.msix"
+	add-appxpackage "$packagePath\PushNotificationsLongRunningTask.Msix\PushNotificationsLongRunningTask.msix"
```
DrusTheAxe, line 12: As-is this is a utility for testing Push Notifications. It's not general purpose e.g. if I'm testing AppLifecycle I don't need (or want) PushNotificationsLongRunningTask.msix installed. And there are others I'm interested in.

As-is this should be somewhere Push-specific e.g. `\test\PushNotifications\PushNotificationsTestPackages.ps1`

SDKPackageManager.ps1
```diff
@@ -0,0 +1,31 @@
...
+}
+elseif(!$task.CompareTo("u"))
+{
+	taskkill /IM PushNotificationsLongRunningTask.exe /F
+	taskkill /IM DynamicDependency.DataStore.exe /F
+	taskkill /IM DynamicDependencyLifetimeManager.exe /F
```
DrusTheAxe, line 18: Why?

This only lives if the `MddBootstrapInitialize()` caller is running in an AppContainer. Otherwise the DDLMShadow.exe is used. And that kills itself if `MddBootstrapShutdown()` is called or the process that called `MddBootstrapInitalize()` terminates (DDLMShadow watches a process handler for the init'ing caller).

SDKPackageManager.ps1
```diff
@@ -0,0 +1,31 @@
...
+	write-host "installing packages"
+	add-appxpackage "$packagePath\Microsoft.WindowsAppRuntime.Framework\Microsoft.WindowsAppRuntime.Framework.msix"
+	add-appxpackage "$packagePath\DynamicDependency.DataStore.Msix\DynamicDependency.DataStore.msix"
+	add-appxpackage "$packagePath\DynamicDependencyLifetimeManager.Msix\DynamicDependencyLifetimeManager.msix"
+	add-appxpackage "$packagePath\PushNotificationsLongRunningTask.Msix\PushNotificationsLongRunningTask.msix"
+}
+elseif(!$task.CompareTo("u"))
+{
+	taskkill /IM PushNotificationsLongRunningTask.exe /F
+	taskkill /IM DynamicDependency.DataStore.exe /F
```
DrusTheAxe, line 17: Why?

This only runs if DynDep APIs are called in an AppContainer. Otherwise DynDep uses `ApplicationDataManager.CreateForPackageFamily()` which is an inproc API (no helper process needed).

SDKPackageManager.ps1
DrusTheAxe: > Is there a way to have a unit test that verifies the script runs successfully?

I'm not following the use case.

UTs/FTs are written as TAEF tests. If they call ::Test::Bootstrap::Setup() that's canned code to correctly install the fake-WinAppSDK packages (WindowsAppRuntime.Test.DDLM... etc) and rigorously verify correctness else fail as expected.

Who's running this script? An interactive user for manual/monkey testing? Or TAEF UT/FT test code?

WindowsAppRuntime.sln
```diff
@@ -301,7 +301,11 @@ EndProject
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "AppNotifications", "dev\AppNotifications\AppNotifications.vcxitems", "{B4824897-88E0-4927-8FB9-E60106F01ED9}"
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Windows.AppNotifications.Projection", "dev\Projections\CS\Microsoft.Windows.AppNotifications.Projection\Microsoft.Windows.AppNotifications.Projection.csproj", "{2385A443-A1C2-4562-8D28-D0AD239EE5E7}"
+EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Common", "Common", "{82A73181-EA4A-431A-B82B-BE6734604CC9}"
+	ProjectSection(SolutionItems) = preProject
+		SDKPackageManager.ps1 = SDKPackageManager.ps1
```
DrusTheAxe, line 307: Pre-Project?

For Common?

So you want this to run for everyone using WindowsAppRuntime.sln, regardless the feature they're working on?

When errors occur how are they detected and handled?

DrusTheAxe, line 307: Relying on the .sln to perform test infrastructure work is strongly _not_ recommended.

We're currently using _VS_ as our project management process, but that has multiple problems today. We're identified we need to do something but haven't decided exactly what yet, and we've had more urgent work to focus on that we haven't address this yet. But as they say, it's not the end, but I can see it from here. Whether we stick with VS and define processes to make it manageable (e.g. projects we design for "I need a DLL" or "I need a test DLL" or... rather than the project templates VS has which are poor fits for our needs and time consuming + error prone to alter for our work). Or switch to msbuild (where *proj may be similar but *.sln doesn't exist). Or CMAKE. Or... We have thoughts but no specific decision other than _not what we're doing today_.

Also, the .sln is used in the build pipeline to compile our bits. We then send them over to Helix test VMs to run -- no .sln exists there.

Making our TAEF UT/FT self-contained and doing setup+cleanup in TAEF fixtures and using things like Test::Bootstrap::Setup() and like aids in `\test\inc\WindowsAppRuntime.Test.*` isa more reliable, performant and flexible solution.

I'm not quite following the intent here. What are you trying to accomplish? What is it about the test\inc etc support that you find inadequate?

DrusTheAxe (changes requested):

SDKPackageManager.ps1
pmpurifoy: Use case is a manual testing where I am installing all the test SDK packages manually. There are some E2E activation flows we want to test which require us to install all the packages. This is a tedious process when you are trying multiple iterations where remove-appxpackage has to be called since these packages have a dependency on the fwk package, so it's not easily replaceable. This script helps improve that process.

WindowsAppRuntime.sln
pmpurifoy: Notifications have their own apps where we test what a developer app may look like. In fact, Notifications app going into the samples repo is basically 1-1 with this inner loop dev app. And while unit testing gets us close, we can't be fully confident since we have multiple activation scenarios where we have to receive/activate a notification (raw/toast) from WNS, which we currently don't have automaton for. This E2E testing gives us confidence that if a user follows our API patterns, everything should be working properly when it makes it to the NuGet.

SDKPackageManager.ps1
hulumane, line 9: It's just a little strange to me that i and u deal with different packages. ie. There is no 1 to 1 mapping between the 2. Why is that?

SDKPackageManager.ps1
hulumane: I feel there should be a little option prompt in the beginning of the script to select which packages to install rather than hardcode these packages:

1) Install Packages for Push
2) Install Packages for Deployment
3) Uninstall packages for Push
3) Uninstall packages for deployment etc.

pmpurifoy: Looks like PushNotifications is the only component with manual demo testing. It seems appropriate to move this to test/PushNotifications. If another component wants to follow this pattern, then they can select only the packages they need.

PR #2306, opened 2022-03-21
Remove [experimental] tags take 2

loneursid:
Turns out that activating APIs for preview and stable releases, requires more than just updating the terminal velocity xml files, it requires eradicating all references to the feature from the code. (feature_) from the IDL and gates from the code.

For reference: https://github.com/microsoft/WindowsAppSDK/pull/2224.

### DIFF
--- a/dev/AppNotifications/AppNotification.cpp
+++ b/dev/AppNotifications/AppNotification.cpp
@@ -9,8 +9,6 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
 {
     AppNotification::AppNotification(hstring const& payload)
     {
-        THROW_HR_IF(E_NOTIMPL, !::Microsoft::Windows::AppNotifications::Feature_AppNotifications::IsEnabled());
-
         XmlDocument xmlDocument{};
 
         // We call LoadXml to verify the payload is xml
--- a/dev/AppNotifications/AppNotification.cpp
+++ b/dev/AppNotifications/AppNotification.cpp
@@ -1,7 +1,6 @@
 ﻿#include "pch.h"
 #include "AppNotification.h"
 #include "Microsoft.Windows.AppNotifications.AppNotification.g.cpp"
-#include <TerminalVelocityFeatures-AppNotifications.h>
 
 using namespace winrt::Windows::Data::Xml::Dom;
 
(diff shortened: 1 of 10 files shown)


kythant (approved):

danielayala94 (approved):

DrusTheAxe (approved):

dev/PushNotifications/PushNotificationChannel.h
```diff
@@ -3,9 +3,7 @@
 
 #pragma once
 #include "Microsoft.Windows.PushNotifications.PushNotificationChannel.g.h"
-
 #include "winrt/Windows.Networking.PushNotifications.h"
-#include <TerminalVelocityFeatures-PushNotifications.h>
```
hulumane, line 8: Should we remove this header from the project?

dev/AppNotifications/AppNotification.cpp
```diff
@@ -1,16 +1,13 @@
 ﻿#include "pch.h"
 #include "AppNotification.h"
 #include "Microsoft.Windows.AppNotifications.AppNotification.g.cpp"
-#include <TerminalVelocityFeatures-AppNotifications.h>
```
hulumane, line 4: Should we remove this header from the project?

DefaultRyan: Depends.

If you can think of another feature coming soon that's going to need this header, it might be nice to avoid the churn of deleting and then re-adding it later.

On the other hand, if there's nothing on the horizon, then get rid of it so we don't accumulate cruft.

dev/PushNotifications/PushNotificationChannel.h
DefaultRyan: Depends.

If you can think of another feature coming soon that's going to need this header, it might be nice to avoid the churn of deleting and then re-adding it later.

On the other hand, if there's nothing on the horizon, then get rid of it so we don't accumulate cruft.

DefaultRyan (approved):

dev/AppNotifications/AppNotification.cpp
loneursid: It's a generated file and although it isn't included anywhere at the moment, it will become handy next time we want to add a few experimental APIs.

Removing the xml and the generation code is beyond the minimum work required for this fix.

dev/PushNotifications/PushNotificationChannel.h
loneursid: It's a generated file and although it isn't included anywhere at the moment, it will become handy next time we want to add a few experimental APIs.

Removing the xml and the generation code is beyond the minimum work required for this fix.

PR #2308, opened 2022-03-22
Remove [experimental] tags take 2 (#2306)

loneursid:
* removing experimental requires removing everything

* Code cleanup

Co-authored-by: Eric Langlois <erlangl@microsoft.com>


hulumane (approved):

jonwis (commented):
Seems OK.  Is this the right time to also add the `[contract]` tagging?

loneursid:
> Seems OK. Is this the right time to also add the [contract] tagging?

No! This is targeting the preview1 branch and still hasn't cleared shiproom. I don't we want to add contracts. We already have contracts for AppNotifications, I'm happy to create a separate PR to add contracts for PushNotifications but targeting main

kythant (approved):

PR #2315, opened 2022-03-24
Removing Silent Failures from Bootstrap Autoinitialization

ShashankNay:
The existing `MddBootstrapAutoInitializer.cs` and `MddBootstrapAutoInitializer.cpp` files would either `FailFast` or `exit` without giving any signal to the user. For example, if an unpackaged app's .exe was clicked without having the prerequisite Framework packages installed, it would fail silently and not let the user know why. This change adds a `MessageBox` that lets the user know that they may be missing the framework package. It will also create a debuggable moment if launched from Visual Studio.

### DIFF
--- a/dev/WindowsAppRuntime_BootstrapDLL/MddBootstrapAutoInitializer.cpp
+++ b/dev/WindowsAppRuntime_BootstrapDLL/MddBootstrapAutoInitializer.cpp
@@ -29,6 +29,12 @@ namespace Microsoft::Windows::ApplicationModel::DynamicDependency::Bootstrap
             const HRESULT hr{ ::MddBootstrapInitialize(c_majorMinorVersion, c_versionTag, c_minVersion) };
             if (FAILED(hr))
             {
+                if (IsDebuggerPresent()) {
+                    DebugBreak();
+                }
+                else {
+                    MessageBox(NULL,L"You need to run the Windows App Runtime Installer and its MSIX components", L"MddBootstrap", 0x00000010L);
+                }
                 exit(hr);
             }
         }
--- a/dev/WindowsAppRuntime_BootstrapDLL/MddBootstrapAutoInitializer.cs
+++ b/dev/WindowsAppRuntime_BootstrapDLL/MddBootstrapAutoInitializer.cs
@@ -5,6 +5,9 @@ namespace Microsoft.Windows.ApplicationModel.DynamicDependency.BootstrapCS
 {
     class AutoInitialize
     {
+        [global::System.Runtime.InteropServices.DllImport("user32.dll", CharSet = global::System.Runtime.InteropServices.CharSet.Unicode, SetLastError = true)]
+        private static extern int MessageBox(global::System.IntPtr hWnd, string lpText, string lpCaption, uint uType);
+
         [global::System.Runtime.CompilerServices.ModuleInitializer]
         internal static void AccessWindowsAppSDK()
         {
@@ -17,6 +20,14 @@ namespace Microsoft.Windows.ApplicationModel.DynamicDependency.BootstrapCS
             }
             catch (global::System.Exception e)
             {
+                if (global::System.Diagnostics.Debugger.IsAttached)
+                {
+                    global::System.Diagnostics.Debugger.Break();
+                }
+                else
+                {
+                    MessageBox(global::System.IntPtr.Zero, "You need to run the Windows App Runtime Installer and its MSIX components", "MddBootstrap", (uint)0x00000010L);
+                }
                 global::System.Environment.FailFast(e.Message);
             }
         }


dev/WindowsAppRuntime_BootstrapDLL/MddBootstrapAutoInitializer.cpp
```diff
@@ -29,6 +29,12 @@ namespace Microsoft::Windows::ApplicationModel::DynamicDependency::Bootstrap
             const HRESULT hr{ ::MddBootstrapInitialize(c_majorMinorVersion, c_versionTag, c_minVersion) };
             if (FAILED(hr))
             {
+                if (IsDebuggerPresent()) {
+                    DebugBreak();
+                }
+                else {
+                    MessageBox(NULL,L"You need to run the Windows App Runtime Installer and its MSIX components", L"MddBootstrap", 0x00000010L);
```
riverar, line 36: This is consistent with what I suggested during initial planning with @DrusTheAxe @zaryaf. This would align bootstrap-required app failures with common .NET runtime-required app failures.

I think the message should be reworded to use as much prior art as possible, so power users+ instinctively know what to do next.

Example:
image

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrapAutoInitializer.cpp
```diff
@@ -29,6 +29,12 @@ namespace Microsoft::Windows::ApplicationModel::DynamicDependency::Bootstrap
             const HRESULT hr{ ::MddBootstrapInitialize(c_majorMinorVersion, c_versionTag, c_minVersion) };
             if (FAILED(hr))
             {
+                if (IsDebuggerPresent()) {
+                    DebugBreak();
+                }
```
riverar, line 34: Opinion: I travel around a lot with the debugger and little is more annoying than hitting breakpoints I didn't set. I think an interested party would set a bp on `MddBootstrapInitialize` anyway, making this of no value.

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrapAutoInitializer.cpp
riverar: Beware without a fusion manifest, the message box will be unthemed/ugly. It may or may not get dark mode treatment angering users.

marb2000: Agreed about rewording it, we can say something like: "This application requires installing the Windows App Runtime and its MSIX components."  
At least, can we add a link to the page to install it? 
https://docs.microsoft.com/en-us/windows/apps/windows-app-sdk/downloads

> Beware without a fusion manifest, the message box will be unthemed/ugly. It may or may not get dark mode treatment angering users.

At this point, there is neither WinUI3 nor OS XAML. So we can only delegate on Win32 dialogs, so I hope that at some points all these dialogs support dark more too.

Scottj1s: @riverar yes, that was my opinion too - model this as closely as possible on the precedent set for the CLR and CRT message boxes

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrapAutoInitializer.cpp
Scottj1s: The reasoning here is that many developers are not even aware of MddBootstrapInitialize living in generated code.  So knowing to set a BP on that in case of failure is a catch-22, and the problem we're trying to solve here - drawing the developer's attention to that generated call.  Without the DebugBreak, we could pop the MessageBox unconditionally, but that's even more annoying to debug (break on all thread, find the one with a message pump, find the call frame ...).

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrapAutoInitializer.cpp
Scottj1s: Let's replace "you need to" with "please verify" language.  We actually don't have certainty that the WAR isn't installed here, and misdiagnosing can be even more frustrating to users.

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrapAutoInitializer.cpp
```diff
@@ -29,6 +29,12 @@ namespace Microsoft::Windows::ApplicationModel::DynamicDependency::Bootstrap
             const HRESULT hr{ ::MddBootstrapInitialize(c_majorMinorVersion, c_versionTag, c_minVersion) };
             if (FAILED(hr))
             {
+                if (IsDebuggerPresent()) {
```
Scottj1s, line 32: use "chameleon coding" - stick with original style/conventions, like brace placement

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrapAutoInitializer.cpp
Scottj1s: The title should be "[application_exe] - This application could not be started" versus "MddBootstrap" which doesn't point to the app and has little meaning for end users.

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrapAutoInitializer.cs
```diff
@@ -17,6 +20,14 @@ internal static void AccessWindowsAppSDK()
             }
             catch (global::System.Exception e)
             {
+                if (global::System.Diagnostics.Debugger.IsAttached)
+                {
+                    global::System.Diagnostics.Debugger.Break();
+                }
+                else
+                {
+                    MessageBox(global::System.IntPtr.Zero, "You need to run the Windows App Runtime Installer and its MSIX components", "MddBootstrap", (uint)0x00000010L);
```
DrusTheAxe, line 29: Why is it either or?

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrapAutoInitializer.cs
```diff
@@ -17,6 +20,14 @@ internal static void AccessWindowsAppSDK()
...
+                }
                 global::System.Environment.FailFast(e.Message);
             }
```
DrusTheAxe, line 32: We could implement the C++ goo in the bootstrap DLL and then none of this C# is needed.

I'd started working on a MddBootstrapInitialize2(). Let me steal parts of this PR over into it and share

DrusTheAxe (changes requested):

DrusTheAxe:
This PR is superseded by [Rich information on Bootstrap initalization failure #2316](https://github.com/microsoft/WindowsAppSDK/pull/2316)

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrapAutoInitializer.cpp
riverar: I just realized this is the consumer side, not the Microsoft impl side, so agree @scottj1s!

Scottj1s:
@ShashankNay  thanks for starting this conversation. Let's run with Howard's work, which addresses all concerns with this PR.

ShashankNay:
Abandoning this PR in favour of Howard's work.

Issue #2216, opened 2022-03-07
If an 'unpackaged' app is packaged, it silently fails


fredemmott:
### Describe the bug

Apps built as 'unpackaged' use MddBootstrapAutoInitializer; at present, if the app is being ran from a package, this makes the app instantly exit with:
- no crash dump, even if enabled
- no messages in the windows event viewer
- run with no messages in Visual Studio -> Debug -> Other Debug Targets -> Debug Installed Package...

### Steps to reproduce the bug

1. build an app as unpackaged
2. build and install an appx/msix with this binary
3. try to run it
4. silent exit
5. check logs

An alternative to steps 1 and 2 is to download the msix from https://github.com/fredemmott/OpenKneeboard/actions/runs/1942621414

### Expected behavior

Ideally, it should work. Unpackaged runs can be useful for debugging, and it's also useful to be able to distribute the same binaries in multiple forms, have one set of PDB files, etc.

Alternatively, it should - by default - log a clear error to the application log in event viewer.

Alternatively, throwing an exception would be a big improvement.

### NuGet package version

1.0.0

### Packaging type

Packaged (MSIX), Unpackaged

### Windows version

Windows 10 version 21H2 (19044, November 2021 Update)

### IDE

Other

fredemmott:
As a workaround for this issue (and a solution for #2215):

- Add `DISABLE_XAML_GENERATED_MAIN=1` to your project's preprocessor definitions
- set `WindowsAppSdkBootstrapInitialize` to `false` in your project's global properties
- copy the default generated `wwinmain()` to your code, removing the `#ifndef DISABLE_XAML_GENERATED_MAIN`
- add something like the following:

```C++
  UINT32 packageNameLen = MAX_PATH;
  wchar_t packageName[MAX_PATH];
  const bool isPackaged
    = GetCurrentPackageFullName(&packageNameLen, packageName) == ERROR_SUCCESS;
  if (!isPackaged) {
    PACKAGE_VERSION minimumVersion {WINDOWSAPPSDK_RUNTIME_VERSION_UINT64};
    winrt::check_hresult(MddBootstrapInitialize(
      WINDOWSAPPSDK_RELEASE_MAJORMINOR,
      WINDOWSAPPSDK_RELEASE_VERSION_TAG_W,
      minimumVersion));
  }

  // ORIGINAL GENERATED wWinMain CODE STARTS HERE
  {
    void(WINAPI * pfnXamlCheckProcessRequirements)();
[...]

PR #2316, opened 2022-03-25
Rich information on Bootstrap initalization failure

DrusTheAxe:
Add new `MddBootstrapInitialize2(...options)`
If Bootstrap Initialize fails...

...Write to the eventlog
image
image

...Optionally
* DebugBreak
* DebugBreak if a debugger is attached to the process
* FailFast
* MessageBox (modeled after .NET's) asking if you'd like to install the WindowsAppRuntime and if so, ShellExecute(url) to the download page

image

The optional behavior is triggered by the new `MddBootstrapInitialize(...options)` parameter or by environment variables
```
SET MICROSOFT_WINDOWSAPPRUNTIME_BOOTSTRAP_INITIALIZE_DEBUGBREAK=1
SET MICROSOFT_WINDOWSAPPRUNTIME_BOOTSTRAP_INITIALIZE_SHOWUI=1
SET MICROSOFT_WINDOWSAPPRUNTIME_BOOTSTRAP_INITIALIZE_FAILFAST=1
```

Tested via manual testing (`HelloWorldAdvancedCPP.exe`) under debugger with memory fiddling and running after setting environment variables.

*.targets updated to provide build-time control (instead of AutoInitialize always unconditionally enabled like in 1.0).

* **To disable the auto-initializer** the auto-initialize set the property `WindowsAppSDKBootstrapAutoInitialize`=`false`
* **If the auto-initalizer is enabled** you can define properties to control the options passed to `MddBootstrapInitialize2()`
  * **WindowsAppSDKBootstrapAutoInitializeOptions_Default=true** -- use the default options (OnNoMatch_ShowUI)
  * **WindowsAppSDKBootstrapAutoInitializeOptions_None=true** -- use no options (None)
  * **WindowsAppSDKBootstrapAutoInitializeOptions_*=true** -- enable the option where * is one or more of the following
    * OnError_DebugBreak
    * OnError_DebugBreak_IfDebuggerAttached
    * OnError_FailFast
    * OnNoMatch_ShowUI
    * OnPackageIdentity_NoOp

For example, to disable the auto-initializer add to your *proj file:

```xml
<PropertyGroup>
    <WindowsAppSDKBootstrapAutoInitialize>false</WindowsAppSDKBootstrapAutoInitialize>
</PropertyGroup>
```

or to enable the auto-initializer (default) but use the DebugBreak+FailFast options (instead of the default)

```xml
<PropertyGroup>
    <WindowsAppSDKBootstrapAutoInitializeOptions_OnError_DebugBreak>true</WindowsAppSDKBootstrapAutoInitializeOptions_OnError_DebugBreak>
    <WindowsAppSDKBootstrapAutoInitializeOptions_OnNoMatch_ShowUI>true</WindowsAppSDKBootstrapAutoInitializeOptions_OnNoMatch_ShowUI>
</PropertyGroup>
```

https://task.ms/38570577 Improve Bootstrap error discoverability+diagnostics
https://task.ms/38497503 Improve Bootstrap AutoInitialize discoverability+diagnostics
https://task.ms/38570355 `MddBootstrapInitializeOptions_OnPackageIdentity_NOOP`


dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.h
```diff
@@ -25,6 +25,48 @@ STDAPI MddBootstrapInitialize(
     PCWSTR versionTag,
     PACKAGE_VERSION minVersion) noexcept;
 
+/// Options for MddBootstrapInitialize()
+typedef enum MddBootstrapInitializeOptions
```
Scottj1s, line 29: we should add another enum to silently ignore failure in packaged apps (for Fred)

Scottj1s:
@DrusTheAxe, this is great - thanks for the fit and finish work!

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -815,3 +876,129 @@ CLSID GetClsid(const winrt::Windows::ApplicationModel::AppExtensions::AppExtensi
...
+    PCWSTR versionTag,
+    PACKAGE_VERSION minVersion) noexcept try
+{
+    HANDLE hEventLog{ RegisterEventSourceW(nullptr, L"Windows App Runtime") };
+    RETURN_LAST_ERROR_IF_NULL(hEventLog);
+
+    const DWORD c_eventId{ static_cast<DWORD>(hrInitialize) };
+    PCWSTR message1{ L"Windows App Runtime" };
+    WCHAR message2[1024]{};
+    PCWSTR message2Format{ L"ERROR 0x%08X: Bootstrapper initialization failed while looking for version %hu.%hu%s (MSIX packages version >= %hu.%hu.%hu.%hu)" };
```
Scottj1s, line 909: packages -> package (or package's)

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -815,3 +876,129 @@ CLSID GetClsid(const winrt::Windows::ApplicationModel::AppExtensions::AppExtensi
...
+                                       L"<Process %d> - This application could not be started",
+                                       GetCurrentProcessId()));
+        caption = captionOnError;
+    }
+
+    // Get the message body
+    WCHAR text[1024]{};
+    PCWSTR textFormat{ L"This application requires the Windows App Runtime\n"
+                       L"    Version %hu.%hu%s\n"
+                       L"    (MSIX packages version >= %hu.%hu.%hu.%hu)\n"
```
Scottj1s, line 976: packages here too

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -74,11 +88,61 @@ static std::wstring g_test_ddlmPackagePublisherId;
...
+        LOG_IF_FAILED(MddBootstrapInitialize_Log(hr, majorMinorVersion, versionTag, minVersion));
+
+        if (WI_IsFlagSet(options, MddBootstrapInitializeOptions_OnError_DebugBreak) ||
+            (WI_IsFlagSet(options, MddBootstrapInitializeOptions_OnError_DebugBreak_IfDebuggerAttached) && IsDebuggerPresent()) ||
+            IsOptionEnabled(L"MICROSOFT_WINDOWSAPPRUNTIME_BOOTSTRAP_INITIALIZE_DEBUGBREAK"))
+        {
+            DebugBreak();
+        }
+
+        if (WI_IsFlagSet(options, MddBootstrapInitializeOptions_OnNoMatch_ShowUI) ||
```
Scottj1s, line 118: are there other HRESULTs that we can expect, besides ERROR_NO_MATCH?  if so, rename enum to MddBootstrapInitializeOptions_OnError_ShowUI

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -815,3 +876,129 @@ CLSID GetClsid(const winrt::Windows::ApplicationModel::AppExtensions::AppExtensi
...
+    return S_OK;
+}
+CATCH_RETURN()
+
+HRESULT MddBootstrapInitialize_ShowUI(
+    UINT32 majorMinorVersion,
+    PCWSTR versionTag,
+    PACKAGE_VERSION minVersion)
+{
+    //TODO: Show this if ERROR_NO_MATCH vs simpler MessageBox("ERROR 0xN in Bootstrapper initialize", MB_OK) ?
```
Scottj1s, line 934: cruft comment?

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -815,3 +876,129 @@ CLSID GetClsid(const winrt::Windows::ApplicationModel::AppExtensions::AppExtensi
...
+
+    // Show the prompt
+    const auto yesno{ MessageBoxW(nullptr, text, caption, MB_YESNO | MB_ICONERROR) };
+    if (yesno == IDYES)
+    {
+        SHELLEXECUTEINFOW sei{};
+        sei.cbSize = sizeof(sei);
+        sei.lpVerb = L"open";
+        sei.lpFile = L"https://docs.microsoft.com/en-us/windows/apps/windows-app-sdk/downloads";
+        //TODO:Replace with https://aka.ms/windowsappsdk/<major>.<minor>/latest/windowsappruntimeinstall-<architecture>.exe
```
Scottj1s, line 999: could create an aka.ms and pass version info as query params for now, then implement url support later

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -74,11 +88,61 @@ static std::wstring g_test_ddlmPackagePublisherId;
...
+    const auto hr{ _MddBootstrapInitialize(majorMinorVersion, versionTag, minVersion) };
+    if (FAILED(hr))
+    {
+        LOG_IF_FAILED(MddBootstrapInitialize_Log(hr, majorMinorVersion, versionTag, minVersion));
+
+        if (WI_IsFlagSet(options, MddBootstrapInitializeOptions_OnError_DebugBreak) ||
+            (WI_IsFlagSet(options, MddBootstrapInitializeOptions_OnError_DebugBreak_IfDebuggerAttached) && IsDebuggerPresent()) ||
+            IsOptionEnabled(L"MICROSOFT_WINDOWSAPPRUNTIME_BOOTSTRAP_INITIALIZE_DEBUGBREAK"))
+        {
+            DebugBreak();
```
Scottj1s, line 115: this will int3 if only MddBootstrapInitializeOptions_OnError_DebugBreak is set, and invoke Watson, which is more or less what the FailFast support provides.  is that useful?

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
DrusTheAxe: This is actually intentional to draw comment about how we should treat NoMatch vs other errors. See threat above regarding line 118

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
DrusTheAxe: We've got the URLs I just need to lookup the exact syntax. The TODO .../latest/... URL is what we ~should~ will invoke ;-)

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
DrusTheAxe: You're suggesting the DebugBreak + FailFast are the same? If a debugger's attached they're not. If no debugger doesn't fail fast do more?

Scottj1s:
Per offline discussion, this is helpful but won't accrue to apps already deployed until/unless the store provides support for fwp updates in a future version.  Meantime, should developers be alerted to the existence of MddBootstrapInitialize2 at build time (e.g., deprecated attribute on MddBootstrapInitialize)?

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
Scottj1s: It feels brittle to tie the behavior to specific HRESULTs, which would be all-too-easy to change in future.  So I'd opt for 2, with the somewhat weasily language.

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.h
Scottj1s: works for me

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
Scottj1s: I'm questioning the usefulness of MddBootstrapInitializeOptions_OnError_DebugBreak, when running without debugger.  That's a hardcoded int3 that takes the process down, presumably with a JIT debugger (watson) to clean things up.  Doesn't seem like a useful scenario beyond the FailFast option.

I would suggest folding the two DebugBreak options into just MddBootstrapInitializeOptions_OnError_DebugBreak, with IsDebuggerPresent().

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -815,3 +876,129 @@ CLSID GetClsid(const winrt::Windows::ApplicationModel::AppExtensions::AppExtensi
...
+        }
+    }
+    return false;
+}
+
+HRESULT MddBootstrapInitialize_Log(
+    HRESULT hrInitialize,
+    UINT32 majorMinorVersion,
+    PCWSTR versionTag,
+    PACKAGE_VERSION minVersion) noexcept try
```
riverar, line 901: Am curious what's driving the need to write to event log. Seems like a lot of additional complexity for zero value.

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -815,3 +876,129 @@ CLSID GetClsid(const winrt::Windows::ApplicationModel::AppExtensions::AppExtensi
...
+                                         minVersion.Major, minVersion.Minor, minVersion.Build, minVersion.Revision));
+
+    // Show the prompt
+    const auto yesno{ MessageBoxW(nullptr, text, caption, MB_YESNO | MB_ICONERROR) };
+    if (yesno == IDYES)
+    {
+        SHELLEXECUTEINFOW sei{};
+        sei.cbSize = sizeof(sei);
+        sei.lpVerb = L"open";
+        sei.lpFile = L"https://docs.microsoft.com/en-us/windows/apps/windows-app-sdk/downloads";
```
riverar, line 998: Remove hardcoded language tag in URL
```suggestion
        sei.lpFile = L"https://docs.microsoft.com/windows/apps/windows-app-sdk/downloads";
```

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -815,3 +876,129 @@ CLSID GetClsid(const winrt::Windows::ApplicationModel::AppExtensions::AppExtensi
...
+    {
+        LOG_IF_FAILED(StringCchPrintfW(captionOnError, ARRAYSIZE(captionOnError),
+                                       L"<Process %d> - This application could not be started",
+                                       GetCurrentProcessId()));
+        caption = captionOnError;
+    }
+
+    // Get the message body
+    WCHAR text[1024]{};
+    PCWSTR textFormat{ L"This application requires the Windows App Runtime\n"
```
riverar, line 974: Should this be in a localizable resource file? (e.g. `.rc`) Or is that for a later round w/ loc lens?

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
angelazhangmsft: Or: https://aka.ms/windowsappsdk/downloads
In case there are ever any redirects needed.

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
Scottj1s: add an option to make it explicit and use delayload?

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
DrusTheAxe: > Should this be in a localizable resource file? (e.g. `.rc`) Or is that for a later round w/ loc lens?

Expected later, as a subsequent PR.

Text is segregated to facilitate eventual move to STRINGTABLE resources. Needs a bit more than simple STRINGTABLE lookup e.g. sprintf and %s, %hd, etc will need to change to FormatMessage() and %1 etc to properly support localization.

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
DrusTheAxe: Anyone can write to the EventLog. You need admin privilege to define a new channel or source (which we don't); if you `ReportEvent` to a non-existent source the event goes to the Application channel. No special powers needed.

This is unconditional as there's no persisted note if problems occur, and this should be a rare event and only logged if something went wrong. Having that persisted log event becomes critical to troubleshoot real world problems.

Notice we don't log anything on success. That's not problematic or helpful, and it _would_ be a ton of data for no value. Check out the Application channel on your local PC and you'll find a ton of events (especially Informational) for routine operations.

>Seems like a lot of additional complexity for little value

The complexity is to handle the "if something goes wrong make sure _something_ is logged." We have too much 1st hand experience troubleshooting problems (even just WinAppSDK issues the past 12 months) not to value this logging.

And it's not really a lot of complexity. More than a dumb sprintf statement but not really that much. Reliability and supportability often require more than simplstic sprintf or string+string 'golden path' code entails.

DrusTheAxe:
> Per offline discussion, this is helpful but won't accrue to apps already deployed until/unless

Anyone directly calling `MddBootstrapInitialize()` will need to make a change. Anyone using C# in 1.0 should be using the [C# API](https://github.com/microsoft/WindowsAppSDK/blob/main/dev/Bootstrap/CS/Microsoft.WindowsAppRuntime.Bootstrap.Net/Bootstrap.cs) and will get benefits (I just haven't updated the C# API yet. Note the TODO list :-). Likewise, anyone using 1.1's C++ app will get similar benefits. And the AutoIntializer will be updated to use Init2 and support project properties for non-default control. Things not yet in this PR but coming RSN.

The only 'problem' is existing apps directly calling `MddBootstrapInitialize()` will need to make a change to their source to get the benefits here.

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
riverar: > Nope. Anyone can write to the EventLog.

That may hold true in the commercial sector, but it isn't true in DoD/IC world. And because it's not universally true, that means my app's runtime suddenly writing to the Event Log has now introduced an additional dependency I need to worry about. In this case, developers can't even opt-out. Just my thinking, not posing this as a problem.

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
riverar: 💧Water cooler gossip: Those aka.ms links are handy (as you pointed out), but also scare me. It reminds me of the go.microsoft.com that expired and still plague docs/Windows today. And now we're betting all our eggs on the stability of the `.ms` ccTLD owned by the UK. Lets just hope Montserrat/the UK doesn't wake up one morning super cranky... 🙃.

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
DrusTheAxe: `DebugBreak if IsDebuggerPresent()=FALSE` isn't the same as `FailFast`. IsDebuggerPresent() tells you if the process is currently under a debugger; it doesn't return `TRUE` if AeDebug is defined.

There's also some logging and telemetry differences for DebugBreak vs FailFast.

Given you have no ability to alter flow here other than these options I'm inclined to keep both `DebugBreak` and `DebugBreak_IfDebuggerPresent`.




More details on how FailFast and DebugBreak are equivalent in concept, but differ in detail...

Both cause the process to terminate if no debugger and AeDebug isn't set. EventLog's slightly different
* ERROR: Exception code = 0xC00000409 vs 0x80000003
* INFORMATION: Event Name = BEX64 vs APPCRASH (this was an x64 exe on an x64 system; BEX64 probably varies for x86 or arm64)
* INFORMATION: P7=address(?), P8=exception code, p9=??? vs P7=exception code, P8=(a different) address(?)

If AeDebug is enabled then the debugger pops up but with very different state
* Fatal process is terminating vs continues on after DebugBreak


Detailed information...

**FailFast**
image

Error:
There was no match for the specified key in the index.

Error:
Faulting application name: HelloWorldAdvancedCPP.exe, version: 0.0.0.0, time stamp: 0x62313c9d
Faulting module name: KERNELBASE.dll, version: 10.0.19041.1566, time stamp: 0x0833f2d4
Exception code: 0xc0000409
Fault offset: 0x000000000010b362
Faulting process id: 0x19888
Faulting application start time: 0x01d842d22bfca769
Faulting application path: C:\source\repos\projectreunionInternal\BuildOutput\Debug\x64\HelloWorldAdvancedCPP\HelloWorldAdvancedCPP.exe
Faulting module path: C:\Windows\System32\KERNELBASE.dll
Report Id: 423f7996-9b19-4a12-95b0-d7b6fad5c24f
Faulting package full name: 
Faulting package-relative application ID: 

Information:
Fault bucket 1462433526648390129, type 5
Event Name: BEX64
Response: Not available
Cab Id: 0

Problem signature:
P1: HelloWorldAdvancedCPP.exe
P2: 0.0.0.0
P3: 62313c9d
P4: KERNELBASE.dll
P5: 10.0.19041.1566
P6: 0833f2d4
P7: 000000000010b362
P8: c0000409
P9: 0000000000000007
P10: 

Attached files:
\\?\C:\ProgramData\Microsoft\Windows\WER\Temp\WERBCFA.tmp.dmp
\\?\C:\ProgramData\Microsoft\Windows\WER\Temp\WERBD49.tmp.WERInternalMetadata.xml
\\?\C:\ProgramData\Microsoft\Windows\WER\Temp\WERBD69.tmp.xml
\\?\C:\ProgramData\Microsoft\Windows\WER\Temp\WERBD67.tmp.csv
\\?\C:\ProgramData\Microsoft\Windows\WER\Temp\WERBD97.tmp.txt

These files may be available here:
\\?\C:\ProgramData\Microsoft\Windows\WER\ReportArchive\AppCrash_HelloWorldAdvanc_5dadd263235e2195da207dcc877eacdbf82a21f_ce7060d9_5b59924f-5f13-44ed-bcae-556e117bc5f1

Analysis symbol: 
Rechecking for solution: 0
Report Id: 423f7996-9b19-4a12-95b0-d7b6fad5c24f
Report Status: 268435456
Hashed bucket: 9d3a75106516a09a744b9b8b4f7fc1f1
Cab Guid: 0

**DebugBreak**
image

Error:
There was no match for the specified key in the index.

Error:
Faulting application name: HelloWorldAdvancedCPP.exe, version: 0.0.0.0, time stamp: 0x62313c9d
Faulting module name: KERNELBASE.dll, version: 10.0.19041.1566, time stamp: 0x0833f2d4
Exception code: 0x80000003
Fault offset: 0x00000000000c90d2
Faulting process id: 0xb6c0
Faulting application start time: 0x01d842d293fc852e
Faulting application path: C:\source\repos\projectreunionInternal\BuildOutput\Debug\x64\HelloWorldAdvancedCPP\HelloWorldAdvancedCPP.exe
Faulting module path: C:\Windows\System32\KERNELBASE.dll
Report Id: 0970faf8-13bc-4a83-8680-f34ba58454fc
Faulting package full name: 
Faulting package-relative application ID: 

Information:
Fault bucket 1789497600844845430, type 4
Event Name: APPCRASH
Response: Not available
Cab Id: 0

Problem signature:
P1: HelloWorldAdvancedCPP.exe
P2: 0.0.0.0
P3: 62313c9d
P4: KERNELBASE.dll
P5: 10.0.19041.1566
P6: 0833f2d4
P7: 80000003
P8: 00000000000c90d2
P9: 
P10: 

Attached files:
\\?\C:\ProgramData\Microsoft\Windows\WER\Temp\WER666F.tmp.dmp
\\?\C:\ProgramData\Microsoft\Windows\WER\Temp\WER66AE.tmp.WERInternalMetadata.xml
\\?\C:\ProgramData\Microsoft\Windows\WER\Temp\WER66CE.tmp.xml
\\?\C:\ProgramData\Microsoft\Windows\WER\Temp\WER66DC.tmp.csv
\\?\C:\ProgramData\Microsoft\Windows\WER\Temp\WER66FC.tmp.txt

These files may be available here:
\\?\C:\ProgramData\Microsoft\Windows\WER\ReportArchive\AppCrash_HelloWorldAdvanc_9f369cebc7cd67252bb2e93dcf8fe3370961c5_ce7060d9_9dea9c30-8594-454c-abb4-92cf7401ffc5

Analysis symbol: 
Rechecking for solution: 0
Report Id: 0970faf8-13bc-4a83-8680-f34ba58454fc
Report Status: 268435456
Hashed bucket: f93bd8b151199ec258d5929638c55176
Cab Guid: 0

FailFast AEDebug 
image

DebugBreak
image

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.h
DrusTheAxe: `MddBootstrapInitializeOptions_OnPackageIdentity_NOP` for clarity 

This avoids the whole 'packaged'/'unpackaged' app lingo, 'is a sparse package a packaged app' and like questions. Merely the statement of fact - if the package has package identity do nothing.

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
DrusTheAxe: >If it gracefully fails, then no problem. Just thinking out loud, thanks.

It does :-)

`MddBootstrapInitialize_Log()` is best effort so worst case no harm no foul

DrusTheAxe:
See https://github.com/microsoft/WindowsAppSDK/issues/2216#issuecomment-1081372678 for additional information

build/NuSpecs/Microsoft.WindowsAppSDK.Bootstrap.CS.targets
```diff
@@ -1,7 +1,23 @@
-<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003"> 
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
-    <Target Name="GenerateBootstrapCS" 
-            BeforeTargets="BeforeCompile">
+    <PropertyGroup Condition="'$(WindowsAppSDKBootstrapAutoInitialize)'==''">
```
Scottj1s, line 3: why are we introducing another opt-out property?  we already have WindowsAppSdkBootstrapInitialize for conditionally importing the C#/C++ autoinit logic

build/NuSpecs/Microsoft.WindowsAppSDK.Bootstrap.CS.targets
```diff
@@ -1,7 +1,23 @@
...
+        <WindowsAppSDKBootstrapAutoInitialize>true</WindowsAppSDKBootstrapAutoInitialize>
+    </PropertyGroup>
+
+    <PropertyGroup Condition="'$(WindowsAppSDKBootstrapAutoInitialize)'=='true'">
+        <DefineConstants
```
Scottj1s, line 8: this looks ill-formed, and the defines below should be appending, not replacing (this is a property, not an itemgroup)

build/NuSpecs/Microsoft.WindowsAppSDK.Bootstrap.CS.targets
```diff
@@ -1,7 +1,23 @@
...
+        <DefineConstants Condition="'$(WindowsAppSDKBootstrapAutoInitializeOptions_Default)'=='true'">MICROSOFT_WINDOWSAPPSDK_BOOTSTRAP_AUTO_INITIALIZE_OPTIONS_DEFAULT</DefineConstants>
+        <DefineConstants Condition="'$(WindowsAppSDKBootstrapAutoInitializeOptions_None)'=='true'">MICROSOFT_WINDOWSAPPSDK_BOOTSTRAP_AUTO_INITIALIZE_OPTIONS_DEFAULT</DefineConstants>
```
Scottj1s, line 10: None also maps to Default?  typo?

build/NuSpecs/Microsoft.WindowsAppSDK.Bootstrap.CS.targets
```diff
@@ -1,7 +1,23 @@
...
+        <DefineConstants Condition="'$(WindowsAppSDKBootstrapAutoInitializeOptions_OnError_DebugBreak)'=='true'">MICROSOFT_WINDOWSAPPSDK_BOOTSTRAP_AUTO_INITIALIZE_OPTIONS_ONERROR_DEBUGBREAK</DefineConstants>
```
Scottj1s, line 11: I still don't see the usefulness of a DebugBreak without the debugger being attached

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
Scottj1s: I'm not persuaded that DebugBreak (without debugger present) is a useful production option.  A hardcoded int3 (aka 80000003) is typically an indication that code has shipped with an ASSERT - a mistake.  I suppose we can leave it up to developers to decide.

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrapAutoInitializer.cpp
```diff
@@ -21,12 +33,42 @@ namespace Microsoft::Windows::ApplicationModel::DynamicDependency::Bootstrap
             ::MddBootstrapShutdown();
         }
 
+        constexpr static MddBootstrapInitializeOptions Options()
+        {
+#if defined(MICROSOFT_WINDOWSAPPSDK_BOOTSTRAP_AUTO_INITIALIZE_OPTIONS_DEFAULT)
+            // Use the default options
+            return MddBootstrapInitializeOptions_OnNoMatch_ShowUI;
+#elif defined(MICROSOFT_WINDOWSAPPSDK_BOOTSTRAP_AUTO_INITIALIZE_OPTIONS_NONE)
```
Scottj1s, line 41: nothing is setting this from msbuild

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrapAutoInitializer.cs
```diff
@@ -11,13 +23,44 @@ internal static void AccessWindowsAppSDK()
...
+
+        internal static global::Microsoft.Windows.ApplicationModel.DynamicDependency.Bootstrap.InitializeOptions Options
+        {
+            get
             {
-                global::System.Environment.FailFast(e.Message);
+#if MICROSOFT_WINDOWSAPPSDK_BOOTSTRAP_AUTO_INITIALIZE_OPTIONS_DEFAULT
+                // Use the default options
+                return global::Microsoft.Windows.ApplicationModel.DynamicDependency.Bootstrap.InitializeOptions.OnNoMatch_ShowUI;
+#elif MICROSOFT_WINDOWSAPPSDK_BOOTSTRAP_AUTO_INITIALIZE_OPTIONS_NONE
```
Scottj1s, line 41: nothing is setting this from msbuild

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
Scottj1s: enum name still doesn't match behavior, which is to pop the UI for all HRESULTs

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrapAutoInitializer.cpp
DrusTheAxe: See other comment. Fixed

build/NuSpecs/Microsoft.WindowsAppSDK.Bootstrap.CS.targets
DrusTheAxe: You don't use AEDebug, do you?

build/NuSpecs/Microsoft.WindowsAppSDK.Bootstrap.CS.targets
DrusTheAxe: Yep copy/paste error. See other comment. Fixed. Thanks.

build/NuSpecs/Microsoft.WindowsAppSDK.Bootstrap.CS.targets
DrusTheAxe: Doh! Only samples I found didn't show any appending but I [just found one](https://stackoverflow.com/questions/36793499/defineconstants-in-addition-to-existing-ones-in-csproj#36793713)

```
<PropertyGroup>
  <DefineConstants>FOO</DefineConstants>
</PropertyGroup>
<PropertyGroup>
  <DefineConstants>$(DefineConstants);BAR</DefineConstants>
</PropertyGroup>
```

That makes sense. Fixing. Thanks

build/NuSpecs/Microsoft.WindowsAppSDK.Bootstrap.CS.targets
DrusTheAxe: Ah. Didn't know we already had that. Distasteful (implies all bootstrap, not just autoinitialization) but OK, will use that.

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
DrusTheAxe: Good point. Mulled it over for a bit and I think the enum is right / fix the code, for 1.1, and consider the other-errors case for 1.2 different message, different URL (https://aka.ms/windowsappsdk/1.1/troubleshooting ?)

Code updated. Thanks!

build/NuSpecs/Microsoft.WindowsAppSDK.Bootstrap.CS.targets
```diff
@@ -1,23 +1,23 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
-    <PropertyGroup Condition="'$(WindowsAppSDKBootstrapAutoInitialize)'==''">
-        <WindowsAppSDKBootstrapAutoInitialize>true</WindowsAppSDKBootstrapAutoInitialize>
+    <PropertyGroup Condition="'$(WindowsAppSDKBootstrapInitialize)'==''">
```
Scottj1s, line 3: this is unnecessary.

https://github.com/microsoft/WindowsAppSDK/blob/48e6b8f1766210977b9c32f1b35641ae71dce551/build/NuSpecs/Microsoft.WindowsAppSDK.Foundation.targets#L6:
  <Import Project="$(MSBuildThisFileDirectory)Microsoft.WindowsAppSDK.Bootstrap.CS.targets" Condition="'$(WindowsAppSdkBootstrapInitialize)' == 'true'"/>

build/NuSpecs/WindowsAppSDK-Nuget-Native.Bootstrap.targets
```diff
@@ -1,17 +1,17 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
-    <PropertyGroup Condition="'$(WindowsAppSDKBootstrapAutoInitialize)'==''">
-        <WindowsAppSDKBootstrapAutoInitialize>true</WindowsAppSDKBootstrapAutoInitialize>
+    <PropertyGroup Condition="'$(WindowsAppSDKBootstrapInitialize)'==''">
```
Scottj1s, line 3: also unnecessary

build/NuSpecs/Microsoft.WindowsAppSDK.Bootstrap.CS.targets
DrusTheAxe: Because `build\NuSpecs\Microsoft.WindowsAppSDK.MddCommon.targets` handles it?

I overlooked that file as it didn't mention bootstrap. Mdd=DynamicDependencies. Generic facility for any package. Bootstrap=DynDep-specialization for WinAppSDK's runtime. 

Any objection to renaming `...MddCommon...` to `...BootstrapCommon...` ?

Scottj1s (approved):
:shipit:

Issue #1788, opened 2021-11-12
C++ project templates should specify variable toolset


Scottj1s:
### Describe the bug

C++ templates should specify default toolset based on the VS version, [like C++/WinRT does](https://github.com/microsoft/cppwinrt/blob/bddd2e1303ecdf4c26e4721cc6b66a58b37d9949/vsix/ProjectTemplates/VC/Windows%20Desktop/ConsoleApplication/ConsoleApplication.vcxproj#L36-L37):
    <PlatformToolset>v143</PlatformToolset>
    <PlatformToolset Condition="'$(VisualStudioVersion)' == '16.0'">v142</PlatformToolset>

The alternative is to add a static dependency to the VSIX, but that's less appealing:
Microsoft.VisualStudio.ComponentGroup.VC.Tools.142.x86.x64

### Steps to reproduce the bug

Install VS 2022 Current and select no workloads
Install WAS C++ Standalone VS 2022 VSIX 
Create new single-project C++ app from template
VS prompts to install UWP prereq and restarts itself
(After restarting, my solution has 0 projects, requiring a new project)
Create a second project and build 

Expected:
Solution builds

Actual:
Error MSB8020 The build tools for Visual Studio 2019 (Platform Toolset = 'v142') cannot be found. To build using the v142 build tools, please install Visual Studio 2019 build tools. Alternatively, you may upgrade to the current Visual Studio tools by selecting the Project menu or right-click the solution, and then selecting "Retarget solution".

riverar:
Just hit this bug myself today, thanks for filing a bug @Scottj1s. Definitely not great out of the box UX.

angelazhangmsft:
I hit this bug today as well. This is being tracked internally at [task.ms/38198157](task.ms/38198157)

PR #2318, opened 2022-03-25
Project template updates (Single-project MSIX Packaging and C++ toolset)

evelynwu-msft:
This change:
* Renames the `EnablePreviewMsixTooling` property (used to opt into Single-project MSIX Packaging) to `EnablePreviewTooling`
* C++ project templates now select a C++ toolset based on the currently used version of Visual Studio (#1788)
* Bake the `HasPackageAndPublishMenu` property (used by Visual Studio to determine if it should enable the `Package and Publish` Solution Explorer context menu) into the project templates that use Single-project MSIX Packaging so that it is available even if the WinAppSdk Nuget package hasn't been restored yet


RealTommyKlein (approved):
LGTM 👍

PR #2321, opened 2022-03-25
MddBootstrap Telemetry & TraceLogging

sachintaMSFT:
Telemetry and TraceLogging for MddBootstrap


riverar:
Is the intention here to start adding telemetry to the bootstrap code/library? Will my apps start sending telemetry on initialize?

dev/WindowsAppRuntime_BootstrapDLL/WindowsAppRuntime_BootstrapDLL.vcxproj.filters
```diff
@@ -44,11 +50,14 @@
...
+    </ClCompile>
   </ItemGroup>
   <ItemGroup>
     <None Include="WindowsAppRuntime.Bootstrap.def">
       <Filter>Source Files</Filter>
     </None>
     <None Include="packages.config" />
   </ItemGroup>
-</Project>
+</Project>
```
DrusTheAxe, line 63: NIT: GitHub complaining about last EOL

DrusTheAxe (commented):
I think I'm missing something. This looks like telemetry support but I don't see where it's actively invoked. Or not these aren't WIL acctivities or remotely like them, it's WIL's fallback provider hook handling everything flowing through that?

sachintaMSFT:
> I think I'm missing something. This looks like telemetry support but I don't see where it's actively invoked. Or not these aren't WIL acctivities or remotely like them, it's WIL's fallback provider hook handling everything flowing through that?

Yes, the PR is half complete. Something weird happened when I pushed the PR (I changed the branch name before pushing and not sure if that had any weird impact to the push) and only half changes went in. Getting all up now.

dev/WindowsAppRuntime_BootstrapDLL/tracelogging.h
```diff
@@ -0,0 +1,136 @@
...
+        HRESULT hresult,
+        UINT32 majorMinorVersion,
+        PCWSTR versionTag,
+        PACKAGE_VERSION &resolvedMinFrameworkPackageVersion,
+        UINT32 initializationCount)
+    {
+        SetStopResult(hresult);
+
+        wil::FailureInfo* pFailure{};
+        if (pFailure)
```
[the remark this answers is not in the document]
sachintaMSFT: Ignore this. Some weirdness in the PR push. Wait for next commit.

dev/WindowsAppRuntime_BootstrapDLL/tracelogging.cpp
```diff
@@ -0,0 +1,61 @@
...
+                initializeActivity.StopWithResult(
+                    failure.hr,
+                    g_initializationMajorMinorVersion,
+                    g_initializationVersionTag.c_str(),
+                    g_initializationFrameworkPackageVersion,
+                    static_cast<UINT32>(g_initializationCount));
+            }
+            else
+            {
+                // BootStrap uses CATCH_RETURN and hence, there is no need to log anything for wil::FailureType::Return to avoid redundanct logging.
```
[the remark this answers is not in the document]
sachintaMSFT: for THROW_**.. this call back will be called.. and for CATCH_RETURN also, this call back is called. So for one throw, this call back is called twice. Hence, although the call back is called twice for same exception, logging it only once to avoid two redundant logs with same info.

dev/WindowsAppRuntime_BootstrapDLL/tracelogging.cpp
```diff
@@ -0,0 +1,61 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#pragma once
+
+#include "pch.h"
+#include "tracelogging.h"
+#include "BootstrapActivity.h"
```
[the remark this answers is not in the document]
sachintaMSFT: None. will add them to pch.h

dev/WindowsAppRuntime_BootstrapDLL/tracelogging.cpp
```diff
@@ -0,0 +1,61 @@
...
+                initializeActivity.StopWithResult(
+                    failure.hr,
+                    g_initializationMajorMinorVersion,
+                    g_initializationVersionTag.c_str(),
+                    g_initializationFrameworkPackageVersion,
+                    static_cast<UINT32>(g_initializationCount));
+            }
+            else if (failure.type == wil::FailureType::FailFast)
+            {
+                WindowsAppRuntimeBootStrapInitialize_WriteEventWithActivity("FailFast",
```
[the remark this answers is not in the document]
sachintaMSFT: I did take care of it generally but missed on few instances. thanks for catching it.

dev/WindowsAppRuntime_BootstrapDLL/tracelogging.h
```diff
@@ -0,0 +1,136 @@
...
+
+        // Set a process-wide callback function for WIL to call each time it logs a failure.
+        wil::SetResultLoggingCallback(initialize_wilResultLoggingCallback);
+
+        TraceLoggingClassWriteStart(Initialize,
+            _GENERIC_PARTB_FIELDS_ENABLED,
+            TraceLoggingValue(majorMinorVersion, "majorMinorVersion"),
+            TraceLoggingValue(versionTag, "majorMinorVersion"),
+            TraceLoggingValue(
+                ConvertPackagVersionToString(requestedMinFrameworkPackageVersion).c_str(),
```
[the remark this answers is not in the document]
sachintaMSFT: Isn't it easy to read a version as a string than as a UINT64 ? in logs and in device drill ?

dev/WindowsAppRuntime_BootstrapDLL/tracelogging.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include "pch.h"
+#include "tracelogging.h"
+#include "BootstrapActivity.h"
+
+extern PACKAGE_VERSION g_initializationFrameworkPackageVersion{};
+extern std::atomic<uint32_t> g_initializationCount{};
+extern UINT32 g_initializationMajorMinorVersion{};
+std::wstring g_initializationVersionTag;
+
+void __stdcall initialize_wilResultLoggingCallback(const wil::FailureInfo& failure) noexcept
```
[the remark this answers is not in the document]
sachintaMSFT: This callback is applied as a process-wide call back. So, I am being very specific to what code block this callback is supporting. Also, notice that I am resetting the call backs appropriately in initialize activity's Start and StopWithResult methods. I am looking into making it a thread-wide callback instead - I don't see any new threads being created to worry about it - what do you think ?

dev/WindowsAppRuntime_BootstrapDLL/tracelogging.h
```diff
@@ -0,0 +1,159 @@
...
+            TraceLoggingValue(!versionTag ? L"": versionTag, "versionTag"),
+            TraceLoggingValue(
+                ConvertPackagVersionToString(minFrameworkPackageVersion).c_str(),
+                "minFrameworkPackageVersion"),
+            TraceLoggingValue(initializationCount, "initializationCount"));
+    }
+    void StopWithResult(
+        const HRESULT hresult,
+        UINT32 initializationCount,
+        UINT16 ddlmFindMethodUsed,
```
DrusTheAxe, line 57: I would drop the ddlmFindMethodUsed

If anything it's interesting to know the IntegrityLevel and IsAppContainer, but as the former is multiple int values perhaps just a `UINT32 flags` and a telemetry enum

```
enum class BootstrapInitializeFlags : uint32_t
{
    None = 0,
    IsElevated = 0x0001,
    IsMediumIL = 0x0002,
    IsLowIL = 0x0003,
    IsAppContainer = 0x0004,
};
DEFINE_ENUM_FLAG_OPERATORS(BootstrapInitializeFlags)
```

and you can compute it inside the telemetry code via

```
#include <Security.IntegrityLevel.h>

BootstrapInitializeFlags ComputeFlags(HANDLE token = nullptr)
{
    BootstrapInitializeFlags flags{};
    const auto integrityLevel{ Security::IntegrityLevel::GetIntegrityLevel(token) };
    if (integrityLevel >= SECURITY_MANDATORY_HIGH_RID)
    {
        flags |= BootstrapInitializeFlags::IsHighIL;
    }
    else if (integrityLevel == SECURITY_MANDATORY_MEDIUM_RID)
    {
        flags |= BootstrapInitializeFlags::IsMediumIL;
    }
    else if (integrityLevel == SECURITY_MANDATORY_LOW_RID)
    {
        flags |= BootstrapInitializeFlags::IsLowIL;
        if (SUCCCEEDED(LOG_IF_FAILED(wil::get_token_is_app_container_nothrow(token, isAppContainer)))
        {
            flags |= BootstrapInitializeFlags::IsAppContainer;
        }
    }
    return flags;
}
```

and in here automagically compute flags and log it

```
const auto flags{ ComputeFlags() };
...
TraceLoggingValue(static_cast<uint32_t>(flags), "Flags")
```


That's equivalent but more useful than knowing AppExtension vs Enumeration

DrusTheAxe (changes requested):

dev/WindowsAppRuntime_BootstrapDLL/tracelogging.h
```diff
@@ -0,0 +1,159 @@
...
+    return std::wstring(formattedVersion);
+}
+#else
+static std::wstring ConvertPackagVersionToString(PACKAGE_VERSION packageVersion) //uint16_t major, uint16_t minor, uint16_t build, uint16_t revision)
+{
+    std::wstring dummy{ L"0.0.0.0" }; return dummy;
+}
+#endif
+
+bool __stdcall wilResultLoggingThreadCallback(const wil::FailureInfo& failure) noexcept;
```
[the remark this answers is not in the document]
sachintaMSFT: No. This is custom definition for custom handling

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -131,16 +162,29 @@ STDAPI_(void) MddBootstrapShutdown() noexcept
         g_endTheLifetimeManagerEvent.reset();
     }
 
+    HRESULT hrLifetimeManagerShutdown = S_OK;
     if (g_lifetimeManager)
     {
-        (void)LOG_IF_FAILED(g_lifetimeManager->Shutdown());
+        hrLifetimeManagerShutdown = g_lifetimeManager->Shutdown();
+        (void)LOG_IF_FAILED(hrLifetimeManagerShutdown);
```
DrusTheAxe, line 169: Why this change?

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -167,9 +211,9 @@ void VerifyInitializationIsCompatible(
     // Sanity check we're already initialized
     // g_lifetimeManager is optional. Don't check it
     // g_endTheLifetimeManagerEvent is optional. Don't check it
-    FAIL_FAST_HR_IF(E_UNEXPECTED, g_windowsAppRuntimeDll == nullptr);
-    FAIL_FAST_HR_IF(E_UNEXPECTED, g_packageDependencyId == nullptr);
-    FAIL_FAST_HR_IF(E_UNEXPECTED, g_packageDependencyContext == nullptr);
+    FAIL_FAST_HR_IF_MSG(E_UNEXPECTED, g_windowsAppRuntimeDll == nullptr, "g_windowsAppRuntimeDll is null");
+    FAIL_FAST_HR_IF_MSG(E_UNEXPECTED, g_packageDependencyId == nullptr, "g_packageDependencyId is null");
+    FAIL_FAST_HR_IF_MSG(E_UNEXPECTED, g_packageDependencyContext == nullptr, "g_packageDependencyContext is null");
```
DrusTheAxe, line 216: Revert.

If this fails we get line# so we'd know, and this is more to catch dev-mistake anyway (we should never have reached this state) so the message doesn't have any real value.

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
sachintaMSFT: to capture the error, if any, in the Stop event down under. This is because Shutdown is best effort and we don't stop Shutdown activity on error.

DrusTheAxe: Ahhh. That wasn't clear. Then make it

`hrLifetimeManagerShutdown = LOG_IF_FAILED(g_lifetimeManager->Shutdown());`

`LOG_*()` return their parameter, they just happen to log as it passes through

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrapActivity.cpp
```diff
@@ -45,3 +45,27 @@ void WindowsAppRuntime::MddBootstrap::Activity::Context::SetLastFailure(const wi
...
+        flags |= WindowsAppRuntime::MddBootstrap::Activity::IntegrityFlags::IsHighIL;
+    }
+    else if (integrityLevel == SECURITY_MANDATORY_MEDIUM_RID)
+    {
+        flags |= WindowsAppRuntime::MddBootstrap::Activity::IntegrityFlags::IsMediumIL;
+    }
+    else if (integrityLevel == SECURITY_MANDATORY_LOW_RID)
+    {
+        flags |= WindowsAppRuntime::MddBootstrap::Activity::IntegrityFlags::IsLowIL;
+        bool isAppContainer;
```
DrusTheAxe, line 64: NIT: `bool isAppContainer{};`

Always initialize variables

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrapActivity.h
```diff
@@ -73,17 +79,45 @@ namespace WindowsAppRuntime::MddBootstrap::Activity
             return m_lastFailure;
         }
 
-        void SetMddBootStrapAPI(MddBootStrapAPI mddBootStrapAPI)
+        void SetMddBootstrapAPI(MddBootstrapAPI mddBootstrapAPI)
+        {
+            m_mddBootstrapAPI = mddBootstrapAPI;
+        }
+
+        void SetInitializeData()
```
DrusTheAxe, line 87: Odd name as 'set' usually implies 'assign' and 'InitializeData' sounds like there's multiple values being set. But it's just the counter and it's increment not assign.

Rename to `IncrementInitializeCount()` and return the new value

```
uint32_t IncrementInitializeCount()
{
    return ++g_initializationCount;
}
```

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrapActivity.h
```diff
@@ -73,17 +79,45 @@ namespace WindowsAppRuntime::MddBootstrap::Activity
...
         {
-            m_mddBootStrapAPI = mddBootStrapAPI;
+            ++g_initializationCount;
         }
 
-        void SetDDLMFindMethodUsed(DDLMFindMethod ddlmFindMethodUsed)
+        // Returns a boolean to indicate to the caller if MddBootstrap should be shutdown
+        bool ShouldShutdownNow()
```
DrusTheAxe, line 93: Technically we're doing shutdown, it's just a question of the work involved

SUGGESTION: Rename to `DecrementInitializationCount()` and add a comment about returning true if the caller needs to do the cleanup work for shutdown to un-initialize our state

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -103,45 +103,38 @@ STDAPI MddBootstrapInitialize(
...
+    if (!WindowsAppRuntime::MddBootstrap::Activity::Context::Get().ShouldShutdownNow())
     {
-        // Multiply initialized. Just decrement our count
-        --g_initializationCount;
-        return;
-    }
-    else if (initializationCount == 0)
-    {
-        // Not initialized. Nothing to do
         return;
```
DrusTheAxe, line 138: This means we only reset state when transitioning count from 1->0 (yay) and only produce telemetry when transitioning count from 1->0 (boo) ?

This skips the reset-state in lines 144-166 and 177-179 AND the shutdownActivity.StopWithResult() on lines 168-175?

The mixed logic doesn't thrill me. Things were already complicated before adding telemetry.

We should report telemetry for every Shutdown call (regardless if count was 1, >1 or 0), and we should avoid mixing up the logic like this. We should...

* decrement the count
* stash data in locals if needed for later use
* reset state if count just transitioned 1->0
* report telemetry
* return

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrapActivity.h
```diff
@@ -33,29 +36,32 @@ namespace WindowsAppRuntime::MddBootstrap::Activity
 
     class Context
     {
-        MddBootStrapAPI m_mddBootStrapAPI{ MddBootStrapAPI::None };
-        DDLMFindMethod m_ddlmFindMethodUsed{ DDLMFindMethod::None };
-
+        MddBootstrapAPI m_mddBootstrapAPI{};
         WindowsAppRuntimeBootstrap_TraceLogger::Initialize m_bootstrapInitializeActivity;
+        std::atomic<uint32_t> g_initializationCount;
+        wil::unique_cotaskmem_string g_initializationPackageFullName;
```
DrusTheAxe, line 42: Do we ever need to know the value is null (vs empty)? At a glance it seems not so this could be `std::wstring`.

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrapActivity.h
```diff
@@ -73,17 +79,45 @@ namespace WindowsAppRuntime::MddBootstrap::Activity
...
+            }
+            return true;
         }
 
         void SetLastFailure(const wil::FailureInfo& failure);
+
+        void SetShutdownData()
+        {
+            g_initializationPackageFullName = {};
+            --g_initializationCount;
```
DrusTheAxe, line 114: This always decrements, but what if count was already zero? Does that caller trap that?

This always clears the pkgfullname, but what if count was >1? Wouldn't we lose the value?

And why do we have both `SetShutdownData()` and `ShouldShutdownNow()`?

Confusing.

DrusTheAxe (changes requested):

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrapActivity.h
sachintaMSFT: Agree. Removed SetShutdownData.

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrapActivity.h
sachintaMSFT: I think you had it as wil::unique_cotaskmem_string (which I maintianed) because IDynamicDependencyLifetimeManager's GetPackageFullName accepts LPWSTR and std::wstring return LPCWSTR. Do you have suggestion on elegant way to handle this ?

DrusTheAxe: Ach. OK, so mote it be.

Nothing to see here. Move along. Nothing to see ;-)

DrusTheAxe (approved):

raghavthind2005 (approved):

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrapActivity.cpp
```diff
@@ -0,0 +1,88 @@
...
+    }
+    else if (initializationCount == 0)
+    {
+        // Not initialized.
+        SetInitializationPackageFullName(PWSTR{});
+    }
+    return m_initializationCount;
+}
+
+const WindowsAppRuntime::MddBootstrap::Activity::IntegrityFlags& WindowsAppRuntime::MddBootstrap::Activity::Context::GetIntegrityFlags(HANDLE token)
```
DefaultRyan, line 66: You are returning a local variable by reference, which is undefined behavior. This has also triggered a build warning in the pipelines. Please fix.

Issue #2324, opened 2022-03-28
PRIResource build action is absent in WPF projects


rohanp-msft:
### Describe the bug

PRIResource build action is absent in WPF projects. As a result, our default inclusion logic to add RESW files to the PRIResource ItemGroup is reversed through automatic injection in the project file. Specifically, the RESW file is removed from PRIResource and added to None.

The solution to this is:
https://docs.microsoft.com/en-us/visualstudio/msbuild/visual-studio-integration-msbuild?view=vs-2022#additional-build-actions

Note: https://github.com/microsoft/WindowsAppSDK/issues/1674 may seem like a dupe but that issue covers two cases: the case described here and the case where something similar happens even for non-WPF projects. The latter is being addressed via this .NET SDK change: https://github.com/dotnet/sdk/pull/24139. We will use https://github.com/microsoft/WindowsAppSDK/issues/1674 to track case 2 and this issue to track case 1.

### Steps to reproduce the bug

1. Use the .NET SDK change https://github.com/dotnet/sdk/pull/24139 (the change isn't merged yet, so you can do this manually by copying over the text in the change to the existing .NET 6 installation).
2. Create a WPF project targeting .NET 6.0. Add a WAP project to it to make it packaged (you'll need to make the WAP project the startup project and add a project reference to the WPF project).
3. Install the Windows App SDK NuGet package in the WPF project.
4. Add a pre-existing RESW file to the project via the VS UI.

### Expected behavior

The project file should not be modified. The Build Action file property should be PRIResource.

### Screenshots

![image](https://user-images.githubusercontent.com/35987549/160479174-3d9bbe45-f1e6-46a0-bbdb-8b95344f64b6.png)
![image](https://user-images.githubusercontent.com/35987549/160479236-146b99a1-cee3-45d0-8e15-22e15afa74ff.png)

### NuGet package version

1.0.0

### Packaging type

Packaged (MSIX)

### Windows version

Insider Build (xxxxx)

### IDE

Visual Studio 2022

rohanp-msft:
Fixed via https://github.com/microsoft/WindowsAppSDK/pull/2327

PR #2327, opened 2022-03-28
Add PRIResource build action to WPF projects.

rohanp-msft:
Please see issue https://github.com/microsoft/WindowsAppSDK/issues/2324 for a description.

I confirmed that for both .NET 5 and .NET 6 WPF projects created in VS 2022, the property is automatically added and set to true.

Note: We don't restrict this to WPF projects because there is no harm in including the item in other project types.

Thanks to @drewnoakes for the help.

**How verified**
1. Use the .NET SDK change https://github.com/dotnet/sdk/pull/24139 (the change isn't merged yet, so you can do this manually by copying over the text in the change to the existing .NET 6 installation).
2. Create a WPF project targeting .NET 6.0. Add a WAP project to it to make it packaged (you'll need to make the WAP project the startup project and add a project reference to the WPF project).
3. Install the Windows App SDK NuGet package in the WPF project.
4. Modify the NuGet install to use this change.
5. Add a pre-existing RESW file to the project via the VS UI. Is the Build Action correct? Yes. Were there project file modifications? Nope.

I verified that nothing is broken by doing this include in non-WPF apps. For that, I did the same as above with a WinForms packaged app, with some additional verification:
1. Does the app build? Yes.
2. Were there dupe resw files in the resources.resfiles file? Nope.
3. Were there dupe strings in the PRI file? Nope.

### DIFF
--- a/dev/MRTCore/packaging/MrtCore.PriGen.targets
+++ b/dev/MRTCore/packaging/MrtCore.PriGen.targets
@@ -22,6 +22,15 @@
     <DefaultLanguage Condition="'$(DefaultLanguage)' == ''">en-US</DefaultLanguage>
   </PropertyGroup>
 
+  <!--
+    PRIResource is not by default in the list of Build Actions for WPF apps. So add it!
+    If we don't add it, our attempt to by default include RESW files in PRIResource will
+    fail (VS will inject MSBuild code in the project file to undo that include).
+  -->
+  <ItemGroup>
+    <AvailableItemName Include="PRIResource" Condition="'$(UseWPF)' == 'true'"/>
+  </ItemGroup>
+
   <PropertyGroup>
     <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
   </PropertyGroup>
--- a/dev/MRTCore/packaging/MrtCore.PriGen.targets
+++ b/dev/MRTCore/packaging/MrtCore.PriGen.targets
@@ -25,10 +25,12 @@
   <!--
     PRIResource is not by default in the list of Build Actions for WPF apps. So add it!
     If we don't add it, our attempt to by default include RESW files in PRIResource will
-    fail (VS will inject MSBuild code in the project file to undo that include).
+    fail (VS will inject MSBuild code in the project file to undo that include). We don't
+    restrict this to WPF projects because there is no harm in including the item in other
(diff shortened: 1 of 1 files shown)


dev/MRTCore/packaging/MrtCore.PriGen.targets
```diff
@@ -22,6 +22,15 @@
     <DefaultLanguage Condition="'$(DefaultLanguage)' == ''">en-US</DefaultLanguage>
   </PropertyGroup>
 
+  <!--
+    PRIResource is not by default in the list of Build Actions for WPF apps. So add it!
+    If we don't add it, our attempt to by default include RESW files in PRIResource will
+    fail (VS will inject MSBuild code in the project file to undo that include).
+  -->
+  <ItemGroup>
+    <AvailableItemName Include="PRIResource" Condition="'$(UseWPF)' == 'true'"/>
```
huichen123, line 31: should we always include PRIResource if it's not? WPF is what we found to have this issue, but not something we need to be limited to.

rohanp-msft: let's always include. i'll test with non-WPF also to verify it doesn't break anything.

rohanp-msft: done

dev/MRTCore/packaging/MrtCore.PriGen.targets
```diff
@@ -22,6 +22,17 @@
...
 
+  <!--
+    PRIResource is not by default in the list of Build Actions for WPF apps. So add it!
+    If we don't add it, our attempt to by default include RESW files in PRIResource will
+    fail (VS will inject MSBuild code in the project file to undo that include). We don't
+    restrict this to WPF projects because there is no harm in including the item in other
+    project types.
+  -->
+  <ItemGroup>
+    <AvailableItemName Include="PRIResource"/>
```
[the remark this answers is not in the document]
huichen123: also the description needs update.

huichen123 (approved):

dev/MRTCore/packaging/MrtCore.PriGen.targets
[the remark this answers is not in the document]
rohanp-msft: yup, doesn't hurt. also, updated description.

axelandrejs (approved):

PR #2344, opened 2022-04-01
Add Agile flag

pmpurifoy:
Fixed issue with activation on STA thread by adding REGCLS_AGILE.

### DIFF
--- a/dev/AppNotifications/AppNotificationManager.cpp
+++ b/dev/AppNotifications/AppNotificationManager.cpp
@@ -138,8 +138,12 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
                 THROW_HR_IF_MSG(E_INVALIDARG, m_notificationComActivatorRegistration, "Already Registered for App Notifications!");
 
                 // Check if the caller has registered event handlers, if so the REGCLS_MULTIPLEUSE flag will cause COM to ensure that all activators
-                // are routed inproc, otherwise with REGCLS_MULTIPLEUSE COM will launch a new process of the Win32 app for each invocation.
+                // are routed inproc, otherwise with REGCLS_SINGLEUSE COM will launch a new process of the Win32 app for each invocation.
                 auto activationFlag{ m_notificationHandlers ? REGCLS_MULTIPLEUSE : REGCLS_SINGLEUSE };
+
+                // Register an INotificationActivationCallback to receive background activations from AppNotification.
+                // Also, STA threads that call CoRegisterClassObject need to use the REGCLS_AGILE flag so that the object is
+                // associated with the neutral apartment. This allows other threads to activate the STA registered thread.
                 THROW_IF_FAILED(::CoRegisterClassObject(
                     AppModel::Identity::IsPackagedProcess() ? registeredClsid : storedComActivatorGuid,
                     winrt::make<AppNotificationManagerFactory>().get(),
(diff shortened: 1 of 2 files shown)


David-Quero:
Looks good to me, how did you validate it?

loneursid:
Did you run it against the winui sample or did you use something else?

pmpurifoy:
Validation was done using the inner loop demo app for console. WinUI verification was done by creating a nuget packagewith this change and verifying GetActivatedEventArgs succeeds in OnLaunched for AppNotification activation scenario.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -468,12 +468,15 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
                 {
                     auto lock{ m_lock.lock_exclusive() };
                     // Register a PushNotificationBackgroundTask to handle background activation scenarios
+
```
David-Quero, line 471: I think this blank line is in the wrong place

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -139,11 +139,12 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
 
                 // Check if the caller has registered event handlers, if so the REGCLS_MULTIPLEUSE flag will cause COM to ensure that all activators
                 // are routed inproc, otherwise with REGCLS_MULTIPLEUSE COM will launch a new process of the Win32 app for each invocation.
+                auto activationFlag{ m_notificationHandlers ? REGCLS_MULTIPLEUSE : REGCLS_SINGLEUSE };
                 THROW_IF_FAILED(::CoRegisterClassObject(
                     AppModel::Identity::IsPackagedProcess() ? registeredClsid : storedComActivatorGuid,
                     winrt::make<AppNotificationManagerFactory>().get(),
                     CLSCTX_LOCAL_SERVER,
-                    m_notificationHandlers ? REGCLS_MULTIPLEUSE : REGCLS_SINGLEUSE,
+                    activationFlag | REGCLS_AGILE,
```
hulumane, line 147: Needs a comment as to why the AGILE flag is added. It's non standard.

Something like:
// In WinUI Application::OnLaunch, the thread context in which CoRegisterClassObject can be an STA. So a subsequent invoke on that object from an MTA can lead to an error unless the object is declared as agile (neutral apartment)

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -468,12 +468,15 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
                     m_waitHandleForArgs.create();
+
+                    auto activationFlags{ m_foregroundHandlers ? REGCLS_MULTIPLEUSE : REGCLS_SINGLEUSE };
                     THROW_IF_FAILED(::CoRegisterClassObject(
                         m_registeredClsid,
                         winrt::make<PushNotificationManagerFactory>().get(),
                         CLSCTX_LOCAL_SERVER,
-                        m_foregroundHandlers ? REGCLS_MULTIPLEUSE : REGCLS_SINGLEUSE,
+                        activationFlags | REGCLS_AGILE,
```
hulumane, line 479: Comment needed here too

loneursid (approved):
Verified using the C++ WinUI Unpackaged sample and this does indeed fix the issue

hulumane (approved):

PR #2350, opened 2022-04-01
Return the payload if it was processed already.

pmpurifoy:
No need to wait for the payload if it has already been delivered. Push/AppNotificationManager no longer throw E_TIMEOUT if you call GetActivatedEventArgs twice.


hulumane (approved):

loneursid (approved):
Thanks for fixing this!

PR #2355, opened 2022-04-04
Add PushNotificationsLongRunningTask.ProxyStub pdb to symbols dir

pmpurifoy:
The PushNotificationsLongRunningTask.ProxyStub.pdb was not being sent to the symbols folder in the pipeline. This PR addresses the issue.


danielayala94:
Seems like this is something good, but please explain in the description what is the problem being addressed here.

pmpurifoy:
> Seems like this is something good, but please explain in the description what is the problem being addressed here.

Sorry. I was testing pipeline to make sure the change was correct 😄 . Will update description.

pmpurifoy:
The foundation pipeline run is blocking this from merging. Unrelated to change but cannot move to main because of it

PR #2358, opened 2022-04-04
Add missing '>' to Microsoft.WindowsAppSDK.Bootstrap.CS.targets

kythant:


jonwis:
How do we get test coverage for this kind of thing in the future?

jonwis (approved):

build/NuSpecs/Microsoft.WindowsAppSDK.Bootstrap.CS.targets
```diff
@@ -1,7 +1,7 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
     <PropertyGroup>
-        <DefineConstants
+        <DefineConstants>
```
riverar, line 4: Shouldn't this entire line be deleted?

kythant:
@jonwis 
I caught this error because the nightly Foundation build which is internal ran into this error in it's IntegrationTest stage which builds a WindowsAppSDK Nuget, builds the WindowsAppSDK tests from the internal repo, and run against it. However this isn't run as part of a PR validation because this stage needs access to the internal feed due the dependencies on other transport packages which are internal. External contributors wouldn't be able to see this type of validation. 

To catch this during the pull request validation/checks, we would need to build a test app that makes use of this target that will directly make use of the Foundation transport package.

build/NuSpecs/Microsoft.WindowsAppSDK.Bootstrap.CS.targets
kythant: Good question. Taking a closer look, seems like we don't even need this line.

kythant: @DrusTheAxe Want to double check with you to see if we can remove it.

riverar:
> To catch this during the pull request validation/checks, we would need to build a test app that makes use of this target that will directly make use of the Foundation transport package.

Maybe in the short term, PR validation could run with something like:

```yaml
- name: smoke test targets
  shell: pwsh
  run: |
     Get-ChildItem *.targets | ForEach-Object { $_.Name; [xml](Get-Content $_) | Out-Null }
```

build/NuSpecs/Microsoft.WindowsAppSDK.Bootstrap.CS.targets
DrusTheAxe: Correct -- delete line 4. That's a typo

For valid sample syntax see https://stackoverflow.com/questions/36793499/defineconstants-in-addition-to-existing-ones-in-csproj#36793713
image

DrusTheAxe (approved):
LGTM if you delete line 4

Rafael's verify-XML-is-WellFormed is a good (and cheap :-) suggestion. Should be applied to all XML files under `/build/NuSpecs`

kythant:
https://dev.azure.com/ms/ProjectReunion/_build/results?buildId=299752&view=logs&j=1f3a93c1-6a00-552a-efc0-7a088e8b5e99&t=963d0d2d-e502-590b-95a9-5ef016693b68

Smoke check test for props and targets tested here

kythant:
https://dev.azure.com/ms/ProjectReunion/_build/results?buildId=299752&view=logs&j=1f3a93c1-6a00-552a-efc0-7a088e8b5e99&t=963d0d2d-e502-590b-95a9-5ef016693b68

Smoke check test for props and targets tested here

PR #2362, opened 2022-04-04
Resolve issues where AppInstance.FindOrRegisterForKey("key") can return null.

jlaanstra:
This PR fixes additional issues with the AppInstance feature:

- Resolves an issue where calling FindOrRegisterForKey("key") will always duplicate the named Mutex, preventing new instances from being created once the actual named instance has closed.
- Resolves an issue where calling FindOrRegisterForKey("key") followed by FindOrRegisterForKey("key2"), will cause the first instance to be named "key2" and all subsequent instances will share the name "key", because FindOrRegisterForKey("key2") will override the m_keyCreationMutex making it looks like no instance is named "key" yet.

We now acquire the lock on the mutex to identify what process is the owner. I added extra comments to make sure the order of release the lock and releasing the mutex is handled correctly.

### DIFF
--- a/dev/AppLifecycle/AppInstance.cpp
+++ b/dev/AppLifecycle/AppInstance.cpp
@@ -582,11 +582,13 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
         // to get an 'open' instead of a 'create' due to it already existing.
         // try_create returns true for a named mutex that already exists
         // and so we can't rely on try_create and have to explicitly check GetLastError().
-        m_keyCreationMutex.create(mutexName.c_str(), 0, MUTEX_ALL_ACCESS);
+        wil::unique_mutex keyCreationMutex;
+        keyCreationMutex.create(mutexName.c_str(), 0, MUTEX_ALL_ACCESS);
 
         bool currentIsKeyOwner = (GetLastError() != ERROR_ALREADY_EXISTS);
         if (currentIsKeyOwner)
         {
+            m_keyCreationMutex = std::move(keyCreationMutex);
             m_key.Resize((key.length() + 1) * sizeof(key.data()[0]));
             THROW_IF_FAILED(StringCchCopy(m_key.Get(), (m_key.Size() / sizeof(wchar_t)), key.c_str()));
         }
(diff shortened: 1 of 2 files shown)


jonwis (approved):

dev/AppLifecycle/AppInstance.cpp
```diff
@@ -582,11 +582,13 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
         // to get an 'open' instead of a 'create' due to it already existing.
         // try_create returns true for a named mutex that already exists
         // and so we can't rely on try_create and have to explicitly check GetLastError().
-        m_keyCreationMutex.create(mutexName.c_str(), 0, MUTEX_ALL_ACCESS);
+        wil::unique_mutex keyCreationMutex;
+        keyCreationMutex.create(mutexName.c_str(), 0, MUTEX_ALL_ACCESS);
 
         bool currentIsKeyOwner = (GetLastError() != ERROR_ALREADY_EXISTS);
```
jlaanstra, line 588: We could probably make this even better by also acquiring the mutex.

dev/AppLifecycle/AppInstance.cpp
```diff
@@ -578,19 +579,31 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
...
-        // and so we can't rely on try_create and have to explicitly check GetLastError().
-        m_keyCreationMutex.create(mutexName.c_str(), 0, MUTEX_ALL_ACCESS);
+        wil::unique_mutex keyCreationMutex;
+        keyCreationMutex.create(mutexName.c_str(), 0, MUTEX_ALL_ACCESS);
 
-        bool currentIsKeyOwner = (GetLastError() != ERROR_ALREADY_EXISTS);
-        if (currentIsKeyOwner)
+        DWORD waitResult = 0;
+        // Wait for 0 seconds. If the mutex is taken we don't own the mutex
+        auto lock = keyCreationMutex.acquire(&waitResult, 0);
```
jonwis, line 589: Ok, so if the process crashes while owning the mutex, the ownership is reset.  Anybody waiting on the mutex gets a "wait abandoned" message. So is there some super slim possibility that some attempts to acquire will get "wait abandoned" meaning that the other person owning the mutex crashed, and it's OK for you to take ownership?

jonwis (approved):

aeloros (approved):

Issue #1674, opened 2021-10-28
resource/image file added from VS UI will not be built into PRI file for WPF project


huichen123:
**Describe the bug**
In a WPF project, adding resource image file from VS UI, the project is automatically modified to exclude those files. Thus they are not built into final PRI file.

**Steps to reproduce the bug**
1. Create a new WPF project
2. Reference to WindowsAppSDK 1.0-preview3
3. add a new resources.resw file or an existing image file to the project. The csproj file automatically adds something like:
~~~
  <ItemGroup>
    <Content Remove="SplashScreen.scale-200.png" />
  </ItemGroup>

  <ItemGroup>
    <PRIResource Remove="Resources.resw" />
  </ItemGroup>
~~~
4. build
5. The image and resource strings are not built into PRI file

**Expected behavior**
The image and resource strings are built into PRI file

**Screenshots**

**Version Info**

NuGet package version: 
[Microsoft.WindowsAppSDK 1.0.0-preview3] 

| Windows version                    | Saw the problem? |
| :--------------------------------- | :-------------------- |
| Insider Build (xxxxx)              |    |
| May 2020 Update (19041)            |    |
| November 2019 Update (18363)       |    |
| May 2019 Update (18362)            |    |
| October 2018 Update (17763)        |    |

**Additional context**
This seems to be a regression caused by https://github.com/microsoft/WindowsAppSDK/pull/1389. Workaround is to manually remove those ItemGroup added automatically.

soIsWinUI3PackorUnpack:
> The image and resource strings are not built into PRI file

**so is WinUI 3** (both Packaged and Unpackaged are affected --> xbfs files are not built into pri file). 
see https://github.com/microsoft/WindowsAppSDK/discussions/1248#discussioncomment-2195280

rohanp-msft:
This issue cannot be properly addressed in the Windows App SDK. It requires .NET SDK support. Will be filing feature request.

rohanp-msft:
> > The image and resource strings are not built into PRI file
> 
> **so is WinUI 3** (both Packaged and Unpackaged are affected --> xbfs files are not built into pri file). see [#1248 (comment)](https://github.com/microsoft/WindowsAppSDK/discussions/1248#discussioncomment-2195280)

[...]

PR #2367, opened 2022-04-05
Use .NET SDK's default inclusion support for Windows App SDK and remove our own broken solution.

rohanp-msft:
Fixes https://github.com/microsoft/WindowsAppSDK/issues/1674

### Default inclusion desired behavior
1. By default, add the .RESW files in the project to PRIResource and the image files in the project to Content. 
2. The app dev can opt out a file by changing the Build Action file property to None in Visual Studio, which will automatically modify the project file accordingly, or by manually modifying the CSPROJ file. 

### Reason for desired behavior
It’s needed for the strings in the .RESW files and the image files to be indexed during PRI generation. 

### Existing “solution” in the Windows App SDK is incomplete 
Existing “solution”: [link](https://github.com/microsoft/WindowsAppSDK/pull/1389) 
Incomplete because: adding resource files form VS UI sometimes results in automatic modification of the project file to exclude those files. See [issue](https://github.com/microsoft/WindowsAppSDK/issues/1674).

### Why the desired behavior couldn't be achieved in the Windows App SDK NuGet package
I tried to follow the [.NET SDK’s implementation](https://docs.microsoft.com/en-us/dotnet/core/project-sdk/overview#default-includes-and-excludes) for default file inclusion from the NuGet package. We don't remove the RESW and image files from the None ItemGroup in [our NuGet PROPS file](https://github.com/microsoft/WindowsAppSDK/blob/main/dev/MRTCore/packaging/MrtCore.props) unlike in [the .NET SDK impl]( https://github.com/dotnet/sdk/blob/main/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.DefaultItems.props#:~:text=%3CNone%20Remove%3D%22**/*%24(DefaultLanguageSourceExtension,None%20Remove%3D%22**/*.resx%22%20/%3E ) because None is populated [in Microsoft.NET.Sdk.DefaultItems.props](https://github.com/dotnet/sdk/blob/main/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.DefaultItems.props#:~:text=%3CNone%20Include%3D%22**/*%22%20Exclude%3D%22%24(DefaultItemExcludes)%3B%24(DefaultExcludesInProjectFolder)%22%20/%3E), which is evaluated after our NuGet PROPS file. Doing that from a NuGet targets file doesn't work because it prevents the app dev from changing the Build Action from PRIResource to None or Content to None (they see: "the path is explicitly excluded from the project").

### .NET SDK feature request and changes
Feature request: https://github.com/dotnet/sdk/issues/24093
Changes: https://github.com/dotnet/sdk/pull/24139 and https://github.com/dotnet/sdk/pull/24704

### How verified
Used Microsoft.WindowsAppSDK.1.1.0-dev.20220405.1.nupkg and dotnet-sdk-6.0.300-preview.22204.3-win-x64.

Note, for the NuGet package: requires a manual modification of the PriGen.targets file because of a infra issue in NuGet generation with the Foundation pipeline; won't be the case for a "real" NuGet package.

Test cases (all in a WinUI WAP app):
- I tested the effect of toggling the four properties added here. It was as expected.
- I tested adding a RESW file and an image file via the VS UI. I examined the PRI.
- I confirmed there were no duplicate images.

Didn't test with a WPF app, WinForms app, etc., because that was already done with the .NET change.


axelandrejs (approved):

dev/MRTCore/packaging/MrtCore.PriGen.targets
```diff
@@ -7,6 +7,16 @@
     <TargetPlatformIdentifierAdjusted>UAP</TargetPlatformIdentifierAdjusted>
   </PropertyGroup>
 
+  <!--
```
huichen123, line 10: my impression is WinUI app doesn't import this file. Does everything still work for WinUI app?

build/publish-mrt.yml
```diff
@@ -105,7 +105,6 @@ steps:
   inputs:
     SourceFolder: '${{ parameters.MRTSourcesDirectory }}\packaging'
     Contents: |
-      MrtCore.props
```
huichen123, line 108: also need to remove the native one.

dev/MRTCore/packaging/MrtCore.PriGen.targets
rohanp-msft: WinUI with WAP does import it:

image

image

WinUI single project MSIX doesn't:

image

image

These properties should be added to Microsoft.Build.Msix.Pri.targets for default file inclusion to work in single project MSIX projects (I just tested - it doesn't).

Default file inclusion works as expected in a WinUI WAP app, the PRI is correct too.

Unless you have objections, I'll also update Microsoft.Build.Msix.Pri.targets in this change.

build/publish-mrt.yml
rohanp-msft: the native one does something native specific via MrtCore.C.props, so I think we still need it

dev/MRTCore/packaging/MrtCore.PriGen.targets
rohanp-msft: I will also add detailed notes to How Verified so it's clear what I tested

rohanp-msft: Actually, Microsoft.Build.Msix.Pri.targets isn't in this repo. So that'll have to be a separate change.

PR #2369, opened 2022-04-06
Deleting SetDisplayNameAndIcon from test and demo apps

danielayala94:
During AppNotification registration for unpackaged apps, we don't retrieve icon/displayName from window anymore. Hence, in this PR we remove the sample code used to set icon and display name using Shell APIs.

### DIFF
--- a/test/TestApps/PushNotificationsDemoApp/main.cpp
+++ b/test/TestApps/PushNotificationsDemoApp/main.cpp
@@ -87,33 +87,6 @@ winrt::PushNotificationChannel RequestChannel()
     return result;
 }
 
-// This function is intended to be called in the unpackaged scenario.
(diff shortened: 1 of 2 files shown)


sharath2727 (approved):

loneursid (approved):
I presume this isn't needed anymore because the functionality is now part of the WIndowsAppSDK or the UDK, right? What about the Preview2 fork, does it still need this code?

danielayala94:
> I presume this isn't needed anymore because the functionality is now part of the WIndowsAppSDK or the UDK, right? What about the Preview2 fork, does it still need this code?

It's not needed because the functionality to support icons using Shell property store is gone, and the code we have now in place will provide at least the default icon.

PR #2373, opened 2022-04-07
Rename ToastNotificationsDemoApp to AppNotificationsDemoApp

sharath2727:
Rename references to ToastNotifications to AppNotifications


loneursid:
How was this tested? I've rebuilt the solution and I've run all the tests and didn't notice anything wrong.

How is the demo app used? always as a stand-alone app or does it get called by other components?

loneursid (commented):
Left a comment but otherwise looks good

danielayala94 (approved):

raghavthind2005 (approved):

PR #2389, opened 2022-04-11
Refactoring PushNotification unit tests 

pmpurifoy:
This PR refactors the PushNotification unit tests to fully run inside TAEF instead of a separate test app exe. Failures are now diagnosable from TAEF output.


test/PowerNotifications/APITests.cpp
```diff
@@ -51,8 +53,10 @@ namespace Test::PowerNotifications
 
         TEST_METHOD(GetBatteryStatus)
         {
-            auto value = PowerManager::BatteryStatus();
-            VERIFY_ARE_EQUAL(value, BatteryStatus::NotPresent);
+            /*auto value = PowerManager::BatteryStatus();
+            VERIFY_ARE_EQUAL(value, BatteryStatus::NotPresent);*/
+
+            PushNotificationManager::Default().Register();
```
danielayala94, line 59: Please revert! These are not our tests! 😅

test/PowerNotifications/PowerNotifications.vcxproj
```diff
@@ -173,6 +173,10 @@
     <None Include="packages.config" />
   </ItemGroup>
   <ItemGroup>
+    <Reference Include="Microsoft.Windows.PushNotifications">
+      <HintPath>..\..\BuildOutput\Debug\x64\WindowsAppRuntime_DLL\Microsoft.Windows.PushNotifications.winmd</HintPath>
+      <IsWinMDFile>true</IsWinMDFile>
+    </Reference>
```
danielayala94, line 179: Revert file please.

test/PushNotificationTests/PackagedSelfContainedTests.cpp
```diff
@@ -0,0 +1,144 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#include "pch.h"
+#include <TestDef.h>
+#include "MockBackgroundTaskInstance.h"
+#include <TlHelp32.h>
```
danielayala94, line 7: What's this file?

test/PushNotificationTests/PackagedSelfContainedTests.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+        TEST_CLASS_CLEANUP(ClassUninit)
+        {
+            ::Test::Bootstrap::CleanupPackages();
+            return true;
+        }
+
+        TEST_METHOD_SETUP(MethodInit)
+        {
+            VERIFY_IS_TRUE(TP::IsPackageRegistered_DynamicDependencyDataStore());
```
danielayala94, line 51: Do we want to verify if the Framework package is registered?

test/PushNotificationTests/PackagedSelfContainedTests.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+        TEST_METHOD_CLEANUP(MethodUninit)
+        {
+            VERIFY_IS_TRUE(TP::IsPackageRegistered_DynamicDependencyDataStore());
+            VERIFY_IS_TRUE(TP::IsPackageRegistered_DynamicDependencyLifetimeManager());
+
+            // Need to keep each TEST_METHOD in clean state.
+            try
+            {
+                PushNotificationManager::Default().UnregisterAll();
```
danielayala94, line 64: I believe we should throw. If there is an issue, we should get notified about it.

test/PushNotificationTests/PackagedSelfContainedTests.cpp
```diff
@@ -0,0 +1,144 @@
...
+        {
+            PushNotificationTestFunctions::ChannelRequestUsingNullRemoteId();
+        }
+
+        TEST_METHOD(ChannelRequestUsingRemoteId)
+        {
+            PushNotificationTestFunctions::ChannelRequestUsingRemoteId();
+        }
+
+        // Currently failing.
```
danielayala94, line 83: Do we have an issue filed for it?

test/PushNotificationTests/PackagedTests.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+        TEST_CLASS_CLEANUP(ClassUninit)
+        {
+            ::Test::Bootstrap::CleanupPackages();
+            return true;
+        }
+
+        TEST_METHOD_SETUP(MethodInit)
+        {
+            ::WindowsAppRuntime::SelfContained::TestInitialize(::Test::Bootstrap::TP::WindowsAppRuntimeFramework::c_PackageFamilyName);
```
danielayala94, line 51: The namespace causes me confusion. This is a non-self-contained packaged app, but we are calling an API for self-contained apps?

test/PushNotificationTests/TestFunctions.cpp
```diff
@@ -0,0 +1,151 @@
...
+}
+
+void PushNotificationTestFunctions::VerifyUnregisterTwice()
+{
+    PushNotificationManager::Default().Register();
+    PushNotificationManager::Default().Unregister();
+    VERIFY_THROWS_HR(PushNotificationManager::Default().Unregister(), E_UNEXPECTED);
+}
+
+void PushNotificationTestFunctions::VerifyUnregisterAll()
```
danielayala94, line 121: What is the difference between VerifyRegisterAndUnregisterAll and this test?

test/PushNotificationTests/TestFunctions.cpp
```diff
@@ -0,0 +1,151 @@
...
+{
+    PushNotificationManager::Default().Register();
+    PushNotificationManager::Default().UnregisterAll();
```
danielayala94, line 124: What about having a test that just calls UnregisterAll? And another which just calls Unregister?

test/PushNotificationTests/TestFunctions.cpp
```diff
@@ -0,0 +1,151 @@
...
+void PushNotificationTestFunctions::VerifyForegroundHandlerSucceeds()
+{
+    PushNotificationManager::Default().PushReceived([](const auto&, PushNotificationReceivedEventArgs const& /* args */) {});
+    PushNotificationManager::Default().Register();
+}
+
+void PushNotificationTestFunctions::VerifyForegroundHandlerFails()
+{
+    PushNotificationManager::Default().Register();
+    VERIFY_THROWS_HR(PushNotificationManager::Default().PushReceived([](const auto&, PushNotificationReceivedEventArgs const& /* args */) {}), HRESULT_FROM_WIN32(ERROR_NOT_FOUND));
```
danielayala94, line 150: What wasn't found in this scenario so that call returns ERROR_NOT_FOUND?

test/PushNotificationTests/TestFunctions.h
```diff
@@ -0,0 +1,18 @@
+﻿class PushNotificationTestFunctions
```
danielayala94, line 1: You don't really need a class to wrap those functions. They can just be under a namespace.

test/PushNotificationTests/UnpackagedSelfContained.cpp
```diff
@@ -0,0 +1,214 @@
...
+                return false;
+            }
+            return true;
+        }
+
+        TEST_CLASS_CLEANUP(ClassUninit)
+        {
+            try
+            {
+
```
danielayala94, line 61: Nit: remove extra line

test/PushNotificationTests/UnpackagedSelfContained.cpp
```diff
@@ -0,0 +1,214 @@
...
+            try
+            {
+                // Cleanup previous installations. Need this to remove any manual installations outside of running this tests.
+                TP::RemovePackage_PushNotificationsLongRunningTask();
+                TP::RemovePackage_DynamicDependencyLifetimeManager();
+                TP::RemovePackage_DynamicDependencyDataStore();
+                TP::RemovePackage_WindowsAppRuntimeFramework();
+
+                TP::AddPackage_WindowsAppRuntimeFramework();       // Installs WARfwk
+                TP::AddPackage_DynamicDependencyDataStore();       // Installs WARmain
```
danielayala94, line 46: Is the comment accurate? (i.e. data store pkg = Main pkg?)

I would remove the other comments.

test/PushNotificationTests/UnpackagedSelfContained.cpp
```diff
@@ -0,0 +1,214 @@
...
+                TP::AddPackage_DynamicDependencyLifetimeManager(); // Installs WARddlm
+                TP::AddPackage_PushNotificationsLongRunningTask(); // Installs the PushNotifications long running task.
+            }
+            catch (...)
+            {
+                return false;
+            }
+            return true;
```
danielayala94, line 54: Please add one line above the return line.

test/PushNotificationTests/UnpackagedSelfContained.cpp
```diff
@@ -0,0 +1,214 @@
...
+                TP::RemovePackage_PushNotificationsLongRunningTask();
+                TP::RemovePackage_DynamicDependencyLifetimeManager();
+                TP::RemovePackage_DynamicDependencyDataStore();
+                TP::RemovePackage_WindowsAppRuntimeFramework();
+            }
+            catch (...)
+            {
+                return false;
+            }
+            return true;
```
danielayala94, line 72: Please add one line above the return line.

test/PushNotificationTests/UnpackagedSelfContained.cpp
```diff
@@ -0,0 +1,214 @@
...
+            VERIFY_IS_TRUE(TP::IsPackageRegistered_DynamicDependencyDataStore());
+            VERIFY_IS_TRUE(TP::IsPackageRegistered_DynamicDependencyLifetimeManager());
+            VERIFY_IS_TRUE(TP::IsPackageRegistered_PushNotificationsLongRunningTask());
+
+            VERIFY_ARE_EQUAL(S_OK, MddBootstrapTestInitialize(Test::Packages::DynamicDependencyLifetimeManager::c_PackageNamePrefix, Test::Packages::DynamicDependencyLifetimeManager::c_PackagePublisherId));
+
+            // Major.Minor version, MinVersion=0 to find any framework package for this major.minor version
+            const UINT32 c_Version_MajorMinor{ Test::Packages::DynamicDependencyLifetimeManager::c_Version_MajorMinor };
+            const PACKAGE_VERSION c_minVersion{};
+            VERIFY_ARE_EQUAL(S_OK, MddBootstrapInitialize(c_Version_MajorMinor, nullptr, c_minVersion));
```
danielayala94, line 87: What is the nullptr param here?

test/PushNotificationTests/UnpackagedSelfContained.cpp
```diff
@@ -0,0 +1,214 @@
...
+            // Need to keep each TEST_METHOD in a clean state in the LRP.
+            try
+            {
+                PushNotificationManager::Default().UnregisterAll();
+            }
+            catch (...)
+            {
+                // We want to unregister regardless of pass/fail to clean the state.
+            }
+            MddBootstrapShutdown();
```
danielayala94, line 103: Does this return HRESULT or is it just void?

test/PushNotificationTests/pch.h
```diff
@@ -6,6 +6,11 @@
 
 #include <unknwn.h>
 
+#ifndef INLINE_TEST_METHOD_MARKUP
+#define INLINE_TEST_METHOD_MARKUP
```
danielayala94, line 10: What's this?

danielayala94 (changes requested):
Great to see this change coming! 😊 Please revert the changes in PowerNotification tests and address the remaining comments in our tests.

test/PushNotificationTests/PackagedSelfContainedTests.cpp
hulumane: I agree. If this throws, we have a much more serious issue. Test shouldn't be swallowing errors.

test/PushNotificationTests/PackagedSelfContainedTests.cpp
```diff
@@ -0,0 +1,144 @@
...
+        TEST_CLASS_CLEANUP(ClassUninit)
+        {
+            ::Test::Bootstrap::CleanupPackages();
+            return true;
+        }
+
+        TEST_METHOD_SETUP(MethodInit)
+        {
+            VERIFY_IS_TRUE(TP::IsPackageRegistered_DynamicDependencyDataStore());
+            VERIFY_IS_TRUE(TP::IsPackageRegistered_DynamicDependencyLifetimeManager());
```
hulumane, line 52: If MethodUnint contains an Unregister API call, it stands to reason that Init should have a corresponding Register call

test/PushNotificationTests/PackagedSelfContainedTests.cpp
hulumane: At thge very least, we need a bool to check if REgister was done by the test and Unregister if the bool is set to true.

test/PushNotificationTests/PackagedSelfContainedTests.cpp
```diff
@@ -0,0 +1,144 @@
...
+        {
+            PushNotificationTestFunctions::VerifyUnregisterAllTwice();
+        }
+
+        TEST_METHOD(VerifyUnregisterAndUnregisterAll)
+        {
+            PushNotificationTestFunctions::VerifyUnregisterAndUnregisterAll();
+        }
+
+        TEST_METHOD(VerifyForegroundHandlerSucceeds)
```
hulumane, line 134: How would we test for activation outside the handler code path? Are there specific tests for that which need to be moved here?

test/PushNotificationTests/PackagedTests.cpp
```diff
@@ -0,0 +1,140 @@
...
+            return true;
+        }
+
+        TEST_METHOD_CLEANUP(MethodUninit)
+        {
+            // Need to keep each TEST_METHOD in clean state.
+            try
```
hulumane, line 58: Just like above, We should not Unregister unless Registration call was successfull.

test/PushNotificationTests/PackagedSelfContainedTests.cpp
pmpurifoy: Tracking now. https://github.com/microsoft/WindowsAppSDK/issues/2392

test/PushNotificationTests/PackagedSelfContainedTests.cpp
```diff
@@ -0,0 +1,144 @@
...
+using namespace winrt::Windows::Foundation;
+using namespace winrt::Windows::Foundation::Collections;
+using namespace winrt::Windows::Management::Deployment;
+using namespace winrt::Windows::Storage;
+using namespace winrt::Windows::System;
+using namespace winrt::Microsoft::Windows::PushNotifications;
+
+namespace Test::PushNotifications
+{
+    class PackagedSelfContainedTests
```
hulumane, line 25: How does this class guarantee that the package is self-contained?

test/PushNotificationTests/PackagedTests.cpp
pmpurifoy: Similarly to MddBootstrapTestInitialize, SelfContained::TestInitialize sets a global string in the SDK used for testing. The implementation looks for a framework that is dynamically generated by the pipeline which will fail in inner loop testing. This API allows inner loop testing using the IsSelfContained API correcty.

test/PushNotificationTests/TestFunctions.cpp
pmpurifoy: There is no difference. Removing.

test/PushNotificationTests/TestFunctions.cpp
pmpurifoy: ERROR_NOT_FOUND is being thrown because there is no registration found. The THROW_HR_MSG will explain this error to the developer.

test/PushNotificationTests/TestFunctions.cpp
hulumane: That should probably fail. But yeah its a good test.

test/PushNotificationTests/TestFunctions.cpp
hulumane: Channel?

test/PushNotificationTests/UnpackagedSelfContained.cpp
pmpurifoy: This is accurate. DynDepDataStore is a part of main pkg

test/PushNotificationTests/UnpackagedSelfContained.cpp
```diff
@@ -0,0 +1,214 @@
...
+        TEST_METHOD(VerifyRegisterAndUnregisterAll)
+        {
+            PushNotificationTestFunctions::VerifyRegisterAndUnregisterAll();
+        }
+
+        TEST_METHOD(MultipleRegister)
+        {
+            auto pushNotificationManager{ PushNotificationManager::Default() };
+            pushNotificationManager.Register();
+            pushNotificationManager.Register();
```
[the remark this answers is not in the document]
hulumane: Atleast it fails in AppNotificationManager. We should be consistent with both of these.

test/PushNotificationTests/UnpackagedSelfContained.cpp
```diff
@@ -0,0 +1,214 @@
...
+        }
+
+        TEST_METHOD(VerifyRegisterAndUnregisterAll)
+        {
+            PushNotificationTestFunctions::VerifyRegisterAndUnregisterAll();
+        }
+
+        TEST_METHOD(MultipleRegister)
+        {
+            auto pushNotificationManager{ PushNotificationManager::Default() };
```
hulumane, line 162: Lets remove the local initialization and directly infer the pushManager via Default

test/PushNotificationTests/TestFunctions.cpp
```diff
@@ -0,0 +1,151 @@
...
+    auto result{ channelOperation.GetResults() };
+    VERIFY_ARE_EQUAL(result.Status(), PushNotificationChannelStatus::CompletedSuccess);
+
+    result.Channel().Close();
+    VERIFY_THROWS_HR(result.Channel().Close(), WPN_E_CHANNEL_CLOSED);
+}
+
+void PushNotificationTestFunctions::VerifyRegisterAndUnregister()
+{
+    auto pushNotificationManager{ PushNotificationManager::Default() };
```
hulumane, line 79: Remove local. Directly infer to Default below

test/PushNotificationTests/TestFunctions.cpp
```diff
@@ -0,0 +1,151 @@
...
+    pushNotificationManager.Register();
+    pushNotificationManager.Unregister();
+}
+
+void PushNotificationTestFunctions::VerifyRegisterAndUnregisterAll()
+{
+    auto pushNotificationManager{ PushNotificationManager::Default() };
```
hulumane, line 86: remove local

test/PushNotificationTests/TestFunctions.cpp
```diff
@@ -0,0 +1,151 @@
...
+    pushNotificationManager.Register();
+    pushNotificationManager.UnregisterAll();
+}
+
+void PushNotificationTestFunctions::MultipleRegister()
+{
+    auto pushNotificationManager{ PushNotificationManager::Default() };
```
hulumane, line 93: remove

test/PushNotificationTests/TestFunctions.cpp
```diff
@@ -0,0 +1,151 @@
...
+}
+
+void PushNotificationTestFunctions::ChannelRequestUsingNullRemoteId()
+{
+    VERIFY_THROWS_HR(PushNotificationManager::Default().CreateChannelAsync(winrt::guid()).get(), E_INVALIDARG);
+}
+
+void PushNotificationTestFunctions::ChannelRequestUsingRemoteId()
+{
+    auto channelOperation{ PushNotificationManager::Default().CreateChannelAsync(c_azureRemoteId) };
```
hulumane, line 56: Remove local. Directly infer Default()

test/PushNotificationTests/TestFunctions.h
hulumane: Makes sense since these are all statics anyway

test/PushNotificationTests/UnpackagedSelfContained.cpp
pmpurifoy: Void

test/PushNotificationTests/pch.h
pmpurifoy: Not needed. Removing.

test/PushNotificationTests/UnpackagedSelfContained.cpp
pmpurifoy: Void

test/PushNotificationTests/UnpackagedSelfContained.cpp
pmpurifoy: The param is versionTag, but I'm changing the bootstrap work to Test::Bootstrap::Setup.

test/PushNotificationTests/TestFunctions.cpp
danielayala94: > ERROR_NOT_FOUND is being thrown because there is no registration found.

App registration? But isn't calling Register() have taken care of it? This is pretty much the reason I initially started the comment thread. A comment in the code would be helpful.

pmpurifoy: Checked it out again. ERROR_NOT_FOUND is being thrown because there is already a registration when trying to register an event handler, which goes against the pattern we have told developers. I don't think ERROR_NOT_FOUND is the correct error code to return to developers.

test/PushNotificationTests/PackagedSelfContainedTests.cpp
pmpurifoy: Not all tests require Register() to be called but UnregisterAll() cleanup must be done. I will send new iteration of what I have for this.

test/PushNotificationTests/PackagedSelfContainedTests.cpp
pmpurifoy: The APIs in the SDK will be treated as Self-contained unless the SelfContained test initialize function is called.

test/PushNotificationTests/PackagedTests.cpp
danielayala94: So, it isn't possible for packaged apps to make use of WinAppSDK as a non-self-contained app without this call?

test/PushNotificationTests/UnpackagedSelfContained.cpp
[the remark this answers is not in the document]
pmpurifoy: Unpackaged self contained scenarios are not supported by PushNotifications. The IsSupported will be used by developers to check if they have access to our APIs. However, our APIs should not throw. This test set verifies self contained behavior when our APIs are gated by IsSupported.

test/PushNotificationTests/TestFunctions.cpp
danielayala94: Thank you @pmpurifoy for diving into it! Please file an issue to address this.

pmpurifoy:
@DrusTheAxe This PR adds the PushNotificationsLongRunningProcess to the Singleton msix package and adds it as a default package to install in TAEF testing. Let me know if you don't want it as part of the default packages installed and we can discuss an alternative.

test/PushNotificationTests/PackagedSelfContainedTests.cpp
pmpurifoy: Cleaned up :)

test/PushNotificationTests/PackagedSelfContainedTests.cpp
pmpurifoy: If the package failed to install on ClassSetup, the test will not run. I think it's an unnecessary check and since none of our tests uninstall the packages, it's better to leave it out IMO.

test/PushNotificationTests/PackagedSelfContainedTests.cpp
pmpurifoy: Using a m_registered bool to track successful registration.

test/PushNotificationTests/PackagedSelfContainedTests.cpp
pmpurifoy: Activation will be handled by a TestApp in another test suite. Looking forward to added improved activation tests soon.

test/PushNotificationTests/PackagedTests.cpp
pmpurifoy: Yup :)

test/PushNotificationTests/TestFunctions.cpp
pmpurifoy: https://github.com/microsoft/WindowsAppSDK/issues/2395

test/PushNotificationTests/UnpackagedTests.cpp
```diff
@@ -0,0 +1,241 @@
...
+using namespace winrt::Windows::Foundation::Collections;
+using namespace winrt::Windows::Management::Deployment;
+using namespace winrt::Windows::Storage;
+using namespace winrt::Windows::System;
+using namespace winrt::Microsoft::Windows::PushNotifications;
+
+
+namespace Test::PushNotifications
+{
+    HRESULT UnpackagedTests::ChannelRequestHelper(IAsyncOperationWithProgress<PushNotificationCreateChannelResult, PushNotificationCreateChannelStatus> const& channelOperation)
```
David-Quero, line 25: Can you create a Helpers file to put common functions that can be used on multiple tests?

test/PushNotificationTests/UnpackagedTests.cpp
```diff
@@ -0,0 +1,241 @@
...
+        else
+        {
+            auto channelOperation{ PushNotificationManager::Default().CreateChannelAsync(winrt::guid()) };
+            VERIFY_ARE_EQUAL(ChannelRequestHelper(channelOperation), E_FAIL);
+        }
+    }
+
+    void UnpackagedTests::ChannelRequestUsingRemoteId()
+    {
+        if (!m_isSelfContained)
```
David-Quero, line 59: What is the purpose of the SelfContained scenario? it seems to me that for self contained almost all tests are going to fail. If so I would remove noise from this file and create tests specifically for self-contained. Also, that way you can unify a lot of tests with the Packaged ones.

pmpurifoy: Packaged, Packaged + SelfContained, Unpackaged, all have the same test behavior. Unpackaged + SelfContained is the only scenario for PushNotifications that needs special attention since none of the APIs will fail.

David-Quero: I didn't know a Packaged app could be SelfContained too, aren't they mutually exclusive?

pmpurifoy: Nope. SelfContained just means that the binaries are shipped with the app instead of having to have a dependency on the WinAppSDK :)

hulumane: This is incorrect. We should use IsSupported(). Not m_isSelfContained

test/PushNotificationTests/UnpackagedTests.h
```diff
@@ -0,0 +1,89 @@
...
+
+namespace Test::PushNotifications
+{
+    class UnpackagedTests
+    {
+    private:
+        HRESULT ChannelRequestHelper(IAsyncOperationWithProgress<PushNotificationCreateChannelResult, PushNotificationCreateChannelStatus> const& channelOperation);
+
+        bool m_registered{ false };
+        bool m_isSelfContained{ false };
```
hulumane, line 28: Please don't use this flag. Instead use IsSupported.

test/PushNotificationTests/UnpackagedTests.cpp
```diff
@@ -95,7 +95,7 @@ namespace Test::PushNotifications
 
     void UnpackagedTests::VerifyUnregisterFails()
     {
-        if (!m_isSelfContained)
+        if (PushNotificationManager::Default().IsSupported())
```
hulumane, line 98: This doesn't look right. If the feature is unsupported, we get an E_UNEXPECTED error?

pmpurifoy: E_UNEXPECTED is what we are throwing when a user alls Unregister without Register.

test/PushNotificationTests/UnpackagedTests.cpp
```diff
@@ -197,7 +197,7 @@ namespace Test::PushNotifications
         PushNotificationManager::Default().UnregisterAll();
         m_registered = false;
 
-        if (!m_isSelfContained)
+        if (PushNotificationManager::Default().IsSupported())
```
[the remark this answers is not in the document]
pmpurifoy: !SelfContained == IsSupported

test/PushNotificationTests/UnpackagedTests.cpp
hulumane: I see.

pmpurifoy: The test still calls Register, it just doesn't expect any API to throw in Unpackaged + SelfContained environment. I can add comments saying "PushNotificationManager APIs called in a SelfContained environment will not throw".

test/PushNotificationTests/UnpackagedTests.cpp
pmpurifoy: Moved to BaseTestSuite

test/DynamicDependency/data/WindowsAppRuntime.Test.Singleton.Msix/WindowsAppRuntime.Test.Singleton.Msix.vcxproj
```diff
@@ -107,6 +124,17 @@
   <ItemGroup>
     <Xml Include="appxmanifest.xml" />
   </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\..\..\..\dev\PushNotifications\PushNotificationsLongRunningTask.ProxyStub\PushNotificationsLongRunningTask.ProxyStub.vcxproj">
```
DrusTheAxe, line 128: BTW you can replace `..\..\..\..\dev\` with `$(RepoRoot)\dev\` then you're not path relative. 

It's a best-practice not to use .. relative path'ing, but we haven't done anything large scale push to clean up the repo (yet). More of an opportunistic change.

NOTE: RepoRoot has no trailing backslash (unlike other paths in VS e.g. $(OutDir) which always end in backslash)

test/PushNotificationTests/BaseTestSuite.cpp
```diff
@@ -0,0 +1,264 @@
...
+using namespace winrt::Windows::Foundation::Collections;
+using namespace winrt::Windows::Management::Deployment;
+using namespace winrt::Windows::Storage;
+using namespace winrt::Windows::System;
+using namespace winrt::Microsoft::Windows::PushNotifications;
+
+void BaseTestSuite::ClassSetup()
+{
+    ::Test::Bootstrap::Setup();
+    bool isSelfContained{ false };
```
DrusTheAxe, line 25: NIT: `bool isSelfContained{};`

No need for the explicit value as it's the same as the default

test/PushNotificationTests/BaseTestSuite.cpp
```diff
@@ -0,0 +1,264 @@
...
+        m_registered = false;
+    }
+}
+
+HRESULT BaseTestSuite::ChannelRequestHelper(IAsyncOperationWithProgress<PushNotificationCreateChannelResult, PushNotificationCreateChannelStatus> const& channelOperation)
+{
+    if (channelOperation.wait_for(c_timeout) != winrt::Windows::Foundation::AsyncStatus::Completed)
+    {
+        channelOperation.Cancel();
+        return HRESULT_FROM_WIN32(ERROR_TIMEOUT); // timed out or failed
```
DrusTheAxe, line 54: No `RETURN_WIN32_ERROR(ERROR_TIMEOUT)` ?

That error info would then show up in test logs

test/PushNotificationTests/BaseTestSuite.cpp
```diff
@@ -0,0 +1,264 @@
...
+void BaseTestSuite::ChannelRequestUsingNullRemoteId()
+{
+    if (PushNotificationManager::Default().IsSupported())
+    {
+        VERIFY_THROWS_HR(PushNotificationManager::Default().CreateChannelAsync(winrt::guid()).get(), E_INVALIDARG);
+    }
+    else
+    {
+        auto channelOperation{ PushNotificationManager::Default().CreateChannelAsync(winrt::guid()) };
+        VERIFY_ARE_EQUAL(ChannelRequestHelper(channelOperation), E_FAIL);
```
DrusTheAxe, line 77: `E_FAIL`? Sad now.

I have a deep dislike for E_FAIL. "I had a problem, and yeah, I know what it is, but not gonna tell you nyah nyah".

That's coming out of the actual PushNotification API? You may want to consider a more specific error code in a future change.

P.S. You can invent new HRESULTs if necessary using FACILITY_ITF. There's caveats to FACILITY_ITF so probably not something to go hog wild on, but when valuable we have the option. See MsixDynamicDependency.h for some examples.

test/PushNotificationTests/BaseTestSuite.cpp
```diff
@@ -0,0 +1,264 @@
...
+    m_registered = true;
+
+    PushNotificationManager::Default().Unregister();
+}
+
+void BaseTestSuite::VerifyUnregisterFails()
+{
+    if (PushNotificationManager::Default().IsSupported())
+    {
+        VERIFY_THROWS_HR(PushNotificationManager::Default().Unregister(), E_UNEXPECTED);
```
DrusTheAxe, line 124: Ooooooh. That's neat. I didn't know TAEF had VERIFY_THROWS*()

test/PushNotificationTests/BaseTestSuite.h
```diff
@@ -0,0 +1,42 @@
...
+        void VerifyUnregisterTwice();
+        void VerifyUnregisterAllTwice();
+        void VerifyUnregisterAndUnregisterAll();
+        void VerifyForegroundHandlerSucceeds();
+        void VerifyForegroundHandlerFails();
+
+    private:
+        HRESULT ChannelRequestHelper(IAsyncOperationWithProgress<PushNotificationCreateChannelResult, PushNotificationCreateChannelStatus> const& channelOperation);
+
+        bool m_registered{ false };
```
DrusTheAxe, line 41: NIT: `bool m_registered{};`

test/inc/TestDef.h
```diff
@@ -23,6 +23,9 @@ static const std::wstring c_testRequestRestartNowRestartedPhasePackagedEventName
 
 static const std::wstring c_testPushPhaseEventName = L"WindowsAppRuntimeTestPushPhaseEventName";
 inline const winrt::hstring c_rawNotificationPayload = L"<toast></toast>";
+static const std::chrono::seconds c_timeout = std::chrono::seconds(300);
```
DrusTheAxe, line 26: Seems like a lot of constants specific to PushNotifications and not across the project. Better served in a PushNotifications directory, or at a minimum a filename less generic e.g.`PushNotifications-Test-Constants.h`

DrusTheAxe (approved):

test/PushNotificationTests/BaseTestSuite.cpp
riverar: `HRESULT_FROM_WIN32(ERROR_NOT_ENOUGH_QUOTA)` 🙃 I kid, I kid.

pmpurifoy: The error is a developer attempting to call a PushNotification APIs in SelfContained environment which is unsupported in this specific scenario. What would be a better existing error code to return? Or should we invent a new HRESULT in this scenario?

DrusTheAxe: `HRESULT_FROM_WIN32(ERROR_NOT_SUPPORTED)` is that what describe

test/PushNotificationTests/BaseTestSuite.cpp
pmpurifoy: It doesn't exist... but I added a define in pch.h that has the same functionality as VERIFY_THROWS_HR.

#define VERIFY_THROWS_HR(expression, hr)        \
            VERIFY_THROWS_SPECIFIC(expression,          \
                winrt::hresult_error,                   \
                [&](winrt::hresult_error e) -> bool     \
                {                                       \
                    return (e.code() == hr);    \
                })

test/PushNotificationTests/BaseTestSuite.cpp
```diff
@@ -0,0 +1,264 @@
...
+using namespace winrt::Windows::Management::Deployment;
+using namespace winrt::Windows::Storage;
+using namespace winrt::Windows::System;
+using namespace winrt::Microsoft::Windows::PushNotifications;
+
+void BaseTestSuite::ClassSetup()
+{
+    ::Test::Bootstrap::Setup();
+    bool isSelfContained{ false };
+    if (SUCCEEDED(TestData::TryGetValue(L"SelfContained", isSelfContained)))
```
hulumane, line 26: Not a huge fan of this if-check. Why not throw if this fails? I mean we don't want our test to run for non-deterministic behavior right?

hulumane: Unless we are saying that we have a valid scenario here where we don't expect SelfContained to be present as a key?

test/PushNotificationTests/BaseTestSuite.cpp
hulumane: Yup. We should always Log and use the macro for error codes IMHO

test/PushNotificationTests/BaseTestSuite.cpp
pmpurifoy: This is basically a switch on the SelfContained scenario. It will run as SelfContained then not SelfContained.

pmpurifoy: But it should never happen where SelfContained isn't set.

test/PushNotificationTests/BaseTestSuite.cpp
```diff
@@ -0,0 +1,264 @@
...
+    {
+        channelOperation.Cancel();
+        return HRESULT_FROM_WIN32(ERROR_TIMEOUT); // timed out or failed
+    }
+
+    auto result{ channelOperation.GetResults() };
+    auto status{ result.Status() };
+    if (status != PushNotificationChannelStatus::CompletedSuccess)
+    {
+        return result.ExtendedError(); // did not produce a channel
```
hulumane, line 61: Use macro please. RETURN_HR(result.ExtendedError());

hulumane: Or better yet. RETURN_HR_IF(result.ExtendedError(), status != PushNotificationChannelStatus::CompletedSuccess);

test/PushNotificationTests/BaseTestSuite.cpp
hulumane: @pmpurifoy Lets open a bug to address this.

pmpurifoy: https://github.com/microsoft/WindowsAppSDK/issues/2416

test/PushNotificationTests/BaseTestSuite.cpp
```diff
@@ -0,0 +1,264 @@
...
+    }
+    else
+    {
+        PushNotificationManager::Default().Register();
+    }
+}
+
+void BaseTestSuite::VerifyMultipleRegisterAndUnregisterAll()
+{
+    PushNotificationManager::Default().Register();
```
hulumane, line 183: The pattern:
 PushNotificationManager::Default().Register();
m_registered = true;
is repeated all over the place. So makes sense to have a helper called:
RegisterWithPushNotificationManager()

test/PushNotificationTests/BaseTestSuite.cpp
```diff
@@ -0,0 +1,264 @@
...
+    m_registered = true;
+
+    PushNotificationManager::Default().UnregisterAll();
```
hulumane, line 186: This pattern is repeated all over the place. So makes sense to have a helper called:
UnregisterAllWithPushNotificationManager()

test/PushNotificationTests/BaseTestSuite.cpp
```diff
@@ -0,0 +1,264 @@
...
+    m_registered = false;
+}
+
+void BaseTestSuite::VerifyUnregisterTwice()
+{
+    PushNotificationManager::Default().Register();
+    m_registered = true;
+
+    PushNotificationManager::Default().Unregister();
```
hulumane, line 201: Shouldn't m_registered be set to false after this? Needs a helper function as pointed to above feedback.

pmpurifoy: Unregister doesn't fully unregister an app from PushNotifications. UnregisterAll wipes the states completely clean.

hulumane: Then we need a tri-state and a bool won't suffice in the tests. Tri state could be mapped by an enum:
enum RegistrationState{
  Registered;
  UnregisteredPartial;
  UnREgisteredFull;
}

Default state at the beginning of every test would be UnregisteredFull. We would then move from UnregisteredFull->Registered
If Unregister is called, we would move from Registered->UnregisteredPartial;
If Unregisterall is called, we would move to UnregisteredFull;

pmpurifoy: Do we need the middle state? At least in terms of test environment, I don't think it's needed.

hulumane: You could get away without it, but the test logic is a bit convoluted without it. Unregister also means the bool is false. Maybe call it m_UnregisteredFully?

hulumane (approved):

test/PushNotificationTests/PushNotification-Test-Constants.h
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+#pragma once
+
+inline const winrt::hstring c_rawNotificationPayload = L"<toast></toast>";
+static const std::chrono::seconds c_timeout = std::chrono::seconds(300);
```
danielayala94, line 6: What's the reason for this var being static instead of inline?

danielayala94 (approved):
:shipit:

test/PushNotificationTests/PushNotification-Test-Constants.h
pmpurifoy: Should be inline. Changing.

PR #2390, opened 2022-04-11
Update local path string

sharath2727:
Description: The current localFolderPath points to an incorrect path. Correct it to use the C:\Users\<user>\AppData\Local\Microsoft\WindowsAppSDK\

Validation: 
1. Use the DemoApp to validate if the iconPath is pointing to the right folder.
2. Added a unittest and validated the contents of the folder.


robertzhou-wpn (approved):

danielayala94 (approved):
Thank you for fixing this!

:shipit:

dev/AppNotifications/ShellLocalization.h
```diff
@@ -9,8 +9,7 @@
 
 namespace Microsoft::Windows::AppNotifications::ShellLocalization
 {
-    const PCWSTR c_localMicrosoftFolder{ LR"(Microsoft)" };
-    const PCWSTR c_localWindowsAppSDKFolder{ LR"(WindowsAppSDK)" };
+    const PCWSTR c_localWindowsAppSDKFolder{ LR"(\Microsoft\WindowsAppSDK\)" };
```
hulumane, line 12: Do we have a unit test to verify that an image file is successfully created in the appropriate path? Strongly recommend that we add one after calling Register API

danielayala94:
There are some improvements we can do on this PR. Working on it.

danielayala94 (approved):

dev/AppNotifications/ShellLocalization.cpp
```diff
@@ -246,7 +247,9 @@ HRESULT Microsoft::Windows::AppNotifications::ShellLocalization::DeleteIconFromC
     std::wstring notificationAppId{ Microsoft::Windows::AppNotifications::Helpers::RetrieveNotificationAppId() };
 
     // path: C:\Users\<currentUser>\AppData\Local\Microsoft\WindowsAppSDK\{AppGUID}.png
-    std::path iconFilePath{ RetrieveLocalFolderPath() / (notificationAppId + c_pngExtension) };
+    std::path iconFilePath{ RetrieveLocalFolderPath() };
```
hulumane, line 250: Why break this up?

dev/AppNotifications/ShellLocalization.cpp
```diff
@@ -219,7 +218,9 @@ HRESULT Microsoft::Windows::AppNotifications::ShellLocalization::RetrieveAssetsF
     std::wstring notificationAppId{ Microsoft::Windows::AppNotifications::Helpers::RetrieveNotificationAppId() };
 
     // path: C:\Users\<currentUser>\AppData\Local\Microsoft\WindowsAppSDK\{AppGUID}.png
-    auto iconFilePath{ RetrieveLocalFolderPath() / (notificationAppId + c_pngExtension) };
+    auto iconFilePath{ RetrieveLocalFolderPath() };
```
hulumane, line 221: Why multiple lines?

dev/AppNotifications/ShellLocalization.cpp
```diff
@@ -188,9 +188,8 @@ inline std::path RetrieveLocalFolderPath()
     THROW_IF_FAILED(SHGetKnownFolderPath(FOLDERID_LocalAppData, 0 /* flags */, nullptr /* access token handle */, &localAppDataPath));
 
     // path: C:\Users\<currentUser>\AppData\Local\Microsoft\WindowsAppSDK
-    std::path localFolderPath{ std::wstring(localAppDataPath.get()) +
-        Microsoft::Windows::AppNotifications::ShellLocalization::c_localMicrosoftFolder +
-        Microsoft::Windows::AppNotifications::ShellLocalization::c_localWindowsAppSDKFolder };
+    std::path localFolderPath{ localAppDataPath.get() };
+    localFolderPath /= Microsoft::Windows::AppNotifications::ShellLocalization::c_localWindowsAppSDKFolder;
```
hulumane, line 192: Why multiple lines?

test/TestApps/ToastNotificationsTestApp/main.cpp
```diff
@@ -1287,6 +1298,35 @@ bool VerifyRemoveAllAsync()
...
+bool VerifyIconPathExists_Unpackaged()
+{
+    try
+    {
+        // Register is already called in main with an explicit appusermodelId
+        wil::unique_cotaskmem_string localAppDataPath;
+        THROW_IF_FAILED(SHGetKnownFolderPath(FOLDERID_LocalAppData, 0 /* flags */, nullptr /* access token handle */, &localAppDataPath));
+
+        std::path iconFilePath{ localAppDataPath.get() };
+        iconFilePath /= c_localWindowsAppSDKFolder;
```
hulumane, line 1310: Why multiple lines?

dev/AppNotifications/ShellLocalization.cpp
danielayala94: I thought to follow some guidance from Howard about using std::filesystem::path, but it was already there. Will revert.

hulumane: Using std::path was the guidance and nothing changes there. We can compress all this to a single line though instead of making it so verbose.

dev/AppNotifications/ShellLocalization.cpp
danielayala94: In this case, we were creating an unnecessary std::wstring. Concatenation can't happen with a raw string to the left.

hulumane: Got it. Yes, unecessary std::wstring conversion is bad. /= is only overloaded for std::path, not raw strings.

hulumane (approved):

dev/AppNotifications/ShellLocalization.cpp
```diff
@@ -246,7 +247,9 @@ HRESULT Microsoft::Windows::AppNotifications::ShellLocalization::DeleteIconFromC
     std::wstring notificationAppId{ Microsoft::Windows::AppNotifications::Helpers::RetrieveNotificationAppId() };
 
     // path: C:\Users\<currentUser>\AppData\Local\Microsoft\WindowsAppSDK\{AppGUID}.png
-    std::path iconFilePath{ RetrieveLocalFolderPath() / (notificationAppId + c_pngExtension) };
+    std::path iconFilePath{ RetrieveLocalFolderPath() };
+    iconFilePath /= notificationAppId;
```
hulumane, line 251: Can we do this iconFilePath = (iconFilePath / notificationAppId) + c_pngExtension?

danielayala94: No, because there is no defined operator+ for std::filesystem::path. The original operator combination works because it does `path / (string + string)` = `path / string`

hulumane: Ok. SO if the original worked in a single line, why are we changing this?

danielayala94: I reverted it to a single line.

PR #2402, opened 2022-04-15
Fix SelfContained always returning true

pmpurifoy:
The IsSelfContained check uses a generated header file from azure pipelines to retrieve the expected framework package family name. After debugging, it was found that the expected package family name was always nullptr. 

This PR fixes the version info header file:
WindowsAppRuntime-VersionInfo.h -> WindowsAppSDK-VersionInfo.h

### DIFF
--- a/dev/Common/WindowsAppRuntime.SelfContained.cpp
+++ b/dev/Common/WindowsAppRuntime.SelfContained.cpp
@@ -3,8 +3,8 @@
 
 #include "pch.h"
 
-#if __has_include(<WindowsAppRuntime-VersionInfo.h>)
-#   include <WindowsAppRuntime-VersionInfo.h>
+#if __has_include(<WindowsAppSDK-VersionInfo.h>)
+#   include <WindowsAppSDK-VersionInfo.h>
 #   define WINDOWSAPPRUNTIME_SELFCONTAINED_EXPECTED_PACKAGEFAMILYNAME   WINDOWSAPPSDK_RUNTIME_PACKAGE_FRAMEWORK_PACKAGEFAMILYNAME_W
 #else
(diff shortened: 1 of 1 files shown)


DrusTheAxe (approved):

jonwis (approved):

riverar:
Is the `WINDOWSAPPRUNTIME_SELFCONTAINED_EXPECTED_PACKAGEFAMILYNAME nullptr` case something you need in production? Would be nice if we could swap that out with an assert or compile-time error to catch issues where the header may be missing on disk for [insert reason here].

DrusTheAxe (changes requested):
This is good but insufficient. The way the pipelines build and flow means this won't fix the issue.

Following up offline

dev/Common/WindowsAppRuntime.SelfContained.cpp
```diff
@@ -3,8 +3,8 @@
 
 #include "pch.h"
 
-#if __has_include(<WindowsAppRuntime-VersionInfo.h>)
-#   include <WindowsAppRuntime-VersionInfo.h>
+#if __has_include(<WindowsAppSDK-VersionInfo.h>)
```
hulumane, line 6: Nice catch. What is that other header file for? Is it plain incorrect?

DrusTheAxe:
This is now obsolete (the code is gone). See https://github.com/microsoft/WindowsAppSDK/pull/2411

DrusTheAxe:
@pmpurifoy this is now obsolete? If so please abandon the PR

pmpurifoy:
> @pmpurifoy this is now obsolete? If so please abandon the PR

Yup.

PR #2424, opened 2022-04-20
Under a .NET SDK version check, bring back our (broken) default file inclusion solution for RESW and image files (applies to non-single project MSIX apps only).

rohanp-msft:
Under a .NET SDK version check, bring back our (broken) default file inclusion solution for RESW and image files (applies to non-single project MSIX apps only).

Reason for change: we don't break apps on upgrade. If they use a sufficiently recent .NET SDK version, they'll get the improvement. Otherwise, they'll get the existing behavior.

**How verified**
In a WinUI with WAP app:
1. If version is less than threshold, do we use the old, WASDK solution? Yes.
2. If version is equal to the threshold, do we use the .NET SDK solution? Yes.
3. If there are duplicate images in Content, do we error? No.
4. If the version is 6.0.2xx, does the comparison still work? Does that evaluate as less than 6.0.300? Yes.

### DIFF
--- a/dev/MRTCore/packaging/MrtCore.PriGen.targets
+++ b/dev/MRTCore/packaging/MrtCore.PriGen.targets
@@ -9,10 +9,9 @@
 
   <!--
     For .NET projects, we use the default file inclusion support that's built into the .NET SDK for the Windows App SDK. For that, we need to set these properties.
+    The support is only available in version 6.0.4 or higher.
   -->
-  <PropertyGroup Condition="'$(UsingMicrosoftNETSdk)' == 'true'">
-    <EnableDefaultContentItems Condition="'$(EnableDefaultContentItems)' == ''">true</EnableDefaultContentItems>
-    <EnableDefaultPRIResourceItems Condition="'$(EnableDefaultPRIResourceItems)' == ''">true</EnableDefaultPRIResourceItems>
+  <PropertyGroup Condition="'$(UsingMicrosoftNETSdk)' == 'true' and $([MSBuild]::VersionGreaterThanOrEquals($(TargetFrameworkVersion), '6.0.4'))">
     <EnableDefaultWindowsAppSdkContentItems Condition="'$(EnableDefaultWindowsAppSdkContentItems)' == ''">true</EnableDefaultWindowsAppSdkContentItems>
     <EnableDefaultWindowsAppSdkPRIResourceItems Condition="'$(EnableDefaultWindowsAppSdkPRIResourceItems)' == ''">true</EnableDefaultWindowsAppSdkPRIResourceItems>
   </PropertyGroup>
--- a/dev/MRTCore/packaging/MrtCore.PriGen.targets
+++ b/dev/MRTCore/packaging/MrtCore.PriGen.targets
@@ -9,9 +9,9 @@
 
   <!--
     For .NET projects, we use the default file inclusion support that's built into the .NET SDK for the Windows App SDK. For that, we need to set these properties.
-    The support is only available in version 6.0.4 or higher.
+    The support is only available in version 6.0.300 or higher.
   -->
-  <PropertyGroup Condition="'$(UsingMicrosoftNETSdk)' == 'true' and $([MSBuild]::VersionGreaterThanOrEquals($(TargetFrameworkVersion), '6.0.4'))">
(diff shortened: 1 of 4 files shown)


rohanp-msft:
Adding @dsplaisted.

dev/MRTCore/packaging/MrtCore.PriGen.targets
```diff
@@ -9,10 +9,9 @@
 
   <!--
     For .NET projects, we use the default file inclusion support that's built into the .NET SDK for the Windows App SDK. For that, we need to set these properties.
+    The support is only available in version 6.0.300 or higher.
   -->
-  <PropertyGroup Condition="'$(UsingMicrosoftNETSdk)' == 'true'">
-    <EnableDefaultContentItems Condition="'$(EnableDefaultContentItems)' == ''">true</EnableDefaultContentItems>
-    <EnableDefaultPRIResourceItems Condition="'$(EnableDefaultPRIResourceItems)' == ''">true</EnableDefaultPRIResourceItems>
+  <PropertyGroup Condition="'$(UsingMicrosoftNETSdk)' == 'true' and $([MSBuild]::VersionGreaterThanOrEquals($(NETCoreSdkVersion), '6.0.300'))">
```
huichen123, line 14: >and $([MSBuild]::VersionGreaterThanOrEquals($(NETCoreSdkVersion), '6.0.300')) [](http://example.com/codeflow?start=64&length=77)

this is not necessary

dev/MRTCore/packaging/MrtCore.props
```diff
@@ -0,0 +1,31 @@
...
+    The app dev can opt out a file by changing the Build Action file property to None in Visual Studio, which
+    will automatically modify the project file accordingly, or by manually modifying the CSPROJ file.
+
+    Yes, the properties used here, including EnableCoreMrtTooling, are defined in MSBuild files imported after
+    this file. This is correct though because ItemGroup definitions are evaluated in a later pass. See
+    https://docs.microsoft.com/en-us/visualstudio/msbuild/build-process-overview?view=vs-2019#evaluation-phase.
+
+    NOTE:
+    This solution is broken. See issue https://github.com/microsoft/WindowsAppSDK/issues/1674. A correct solution
+    now exists in the .NET SDK but is available only in version 6.0.300 or higher. So, continue to use this broken
```
huichen123, line 22: >6.0.300 [](http://example.com/codeflow?start=64&length=7)

just want to double check this version is correct. i thought you mentioned 6.0.4.

dev/MRTCore/packaging/MrtCore.props
```diff
@@ -0,0 +1,31 @@
...
+    solution if we're using an older version. For the .NET SDK feature request with info on why this is
+    insufficient, see https://github.com/dotnet/sdk/issues/24093.
+  -->
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)'=='.NETCoreApp' and $([MSBuild]::VersionLessThan($(NETCoreSdkVersion), '6.0.300')) and '$(EnableDefaultItems)'=='true' and '$(EnableCoreMrtTooling)'=='true'">
+    <Content Include="$(_MrtCoreImageGlobs)" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" Condition="'$(EnableDefaultContentItems)'!='false'" />
+    <PRIResource Include="**/*.resw" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" Condition="'$(EnableDefaultPRIResourceItems)'!='false'"/>
```
huichen123, line 28: do we need to remove from none?

rohanp-msft: We cannot do it here. None is populated later. Doing it in a TARGETS file results in an error. Please see the comment.

dev/MRTCore/packaging/MrtCore.PriGen.targets
rohanp-msft: True. Since it makes the intention explicit, I prefer to keep it. If you think we should remove it, I can do that.

dev/MRTCore/packaging/MrtCore.props
rohanp-msft: Yup, this is right. I was wrong before.

dev/MRTCore/packaging/MrtCore.PriGen.targets
huichen123: I'd remove it if it was me. But I'll leave it up to you.

huichen123 (approved):

PR #2432, opened 2022-04-21
Add new DWrite color APIs.

niklasb-ms:
This change updates the DWriteCore API specification to include new color font APIs, which will be added to both system DWrite and DWriteCore.

The new color font APIs are needed to support new capabilities in version 1 of the OpenType COLR table, including gradients, transforms, and composition modes. The relevant chapter of the OpenType spec is here: https://docs.microsoft.com/en-us/typography/opentype/spec/colr

Note: Much of the new overview material is describing existing APIs.

### DIFF
--- a/specs/dwritecore/DWriteCore.md
+++ b/specs/dwritecore/DWriteCore.md
@@ -1756,9 +1756,9 @@ struct DWRITE_PAINT_COLOR
 
     /// <summary>
     /// Specifies the palette entry index if the color value comes from the font's color palette.
-    /// Otherwise, this member is UINT32_MAX.
+    /// Otherwise, this member is UINT16_MAX.
     /// </summary>
-    UINT32 paletteIndex;
+    UINT16 paletteEntryIndex;
 
     /// <summary>
     /// Specifies an alpha value multiplier in the range 0 to 1 that was used to compute the color
--- a/specs/dwritecore/DWriteCore.md
+++ b/specs/dwritecore/DWriteCore.md
@@ -519,13 +519,16 @@ monochrome glyphs, but a color font also has color representations of some glyph
 in color, an application must use different glyph rendering APIs instead of calling the monochrome
 `DrawGlyphRun` method.
 
-Note that color font support does not affect text layout it at all. The text layout always produces
-monochrome glyph runs. Color is enabled at the glyph rendering level by substituting color
-representations for the monochrome glyphs.
+Because color font support is implemented at the glyph rendering level, it does not affect text
+layout. This is true whether a client uses the `IDWriteTextLayout` interface or implements its own
+text layout algorithm. The mapping of characters to glyphs and the positioning of those glyphs all
+use monochrome glyph IDs and their associated metrics. The output of the text layout process is a
+sequence of monochrome glyph runs. Color font support can be enabled by translating those monochrome
+"base" glyph runs to color glyphs runs at rendering time.
 
 The easiest way to enable color is to call the `DrawGlyphRunWithColor` method instead of calling
-`DrawGyphRun`. Direct2D exposes this method through the `ID2D1DeviceContext7` interface. DirectWrite
-exposes this method through the `IDWriteBitmapRenderTarget3` interface.
+`DrawGlyphRun`. Direct2D exposes this method through the `ID2D1DeviceContext7` interface.
+DirectWrite exposes this method through the `IDWriteBitmapRenderTarget3` interface.
 
 At a lower level, enabling color involves translating the monochrome _base glyph run_ to a sequence
 of _color glyph runs_. Each color glyph run must then be rendered using the appropriate method,
@@ -535,7 +538,7 @@ depending on its _glyph image format_. The translation is performed by the facto
 ### Using TranslateColorGlyphRun
 
 The `TranslateColorGlyphRun` factory method translates a monochrome _base glyph run_ into a sequence
-of _color glyph runs_. There are multiple ways of representiong color glyphs, known as _glyph image
+of _color glyph runs_. There are multiple ways of representing color glyphs, known as _glyph image
 formats_ and identified by the `DWRITE_GLYPH_IMAGE_FORMATS` enumeration. When calling
 `TranslateColorGlyphRun`, the client specifies the combination of glyph image formats it supports.
 The `TranslateColorGlyphRun` returns a _glyph run enumerator_ object
@@ -548,8 +551,15 @@ If the base glyph run does not have color glyphs in any of the requested glyph i
 `TranslateColorGlyphRun` method returns `DWRITE_E_NOCOLOR`, in which case the client should render
 the base glyph run as a monochrome glyph run.
 
+Otherwise, the base glyph run may contain a mix of color glyphs in different glyph image formats as
+well as monochrome glyphs. For each base glyph, the `TranslateColorGlyphRun` method determines the
+available glyph image format (if any) that best matches the requested formats. The base glyph run is
+split into subranges based on the matching glyph image format. Glyphs that do not support any of the
+requested formats are treated as monochrome glyphs.
+
 The glyph image formats fall into four general categories:
 
+-   Monochrome glyphs (`DWRITE_GLYPH_IMAGE_FORMATS_TRUETYPE` or `DWRITE_GLYPH_IMAGE_FORMATS_CFF`)
 -   Layers of solid-color glyphs (`DWRITE_GLYPH_IMAGE_FORMATS_COLR`)
 -   Color bitmaps embedded in the font (`DWRITE_GLYPH_IMAGE_FORMATS_PNG`,
     `DWRITE_GLYPH_IMAGE_FORMATS_JPEG`, `DWRITE_GLYPH_IMAGE_FORMATS_TIFF`, and
@@ -558,6 +568,24 @@ The glyph image formats fall into four general categories:
 -   Vector graphics as a tree of "paint" elements (`DWRITE_GLYPH_IMAGE_FORMATS_COLR_PAINT_TREE`)
 
 The following subsections describe each of these approaches.
+
+### Monochrome glyphs
+
+The `DWRITE_GLYPH_IMAGE_FORMATS_TRUETYPE` and `DWRITE_GLYPH_IMAGE_FORMATS_CFF` glyph image formats
+represent monochrome glyphs, which can be rendered using the `DrawGlyphRun` method. The two formats
+correspond to different ways of representing glyph outlines in the font. An enumerated
+`DWRITE_COLOR_GLYPH_RUN1` structure can specify these formats for one of two reasons:
+
+-   One or more of the base glyphs do not support any of the requested glyph image formats, and so
+    should be rendered as monochrome glyphs.
+-   The best matching glyph image format for a glyph was `DWRITE_GLYPH_IMAGE_FORMATS_COLR`. Glyphs
+    in this image format are rendered by enumerating a monochrome run in a specified color for each
+    layer, as described in the next section.
+
+In the first case, the enumerated glyph run should be rendered using the application-defined
+foreground brush. This is indicated by setting the `paletteIndex` member of the
+`DWRITE_COLOR_GLYPH_RUN` structure to `DWRITE_NO_PALETTE_INDEX` (0xFFFF). If the `paletteIndex`
+member has any other value, the client should render the text in the specified color.
 
 ### Layers of solid color glyphs
 
@@ -574,7 +602,7 @@ their glyph image formats are either `DWRITE_GLYPH_IMAGE_FORMATS_TRUETYPE` or
 
 For example, suppose the base glyph run specifies a smiley face glyph. If we rendered the base glyph
 run using `DrawGlyphRun`, the visual result would be a monochrome representation of a smiley face
-(the base glyph ID). Thd COLR table in the font maps this base glyph ID to an array of _layer
+(the base glyph ID). The COLR table in the font maps this base glyph ID to an array of _layer
 records_, each of which specifies a _replacement glyph ID_ and an index into a font-defined color
 palette. The `IDWriteColorGlyphRunEnumerator` object returned by `TranslateColorGlyphRun` enumerates
 one color glyph run for each of those layer records. Each color glyph run has a position, a single
@@ -1232,34 +1260,82 @@ int __cdecl wmain(int argc, WCHAR** argv)
 
 ### Rendering Color Glyph Runs
 
-In most cases, the simplest way to render glyph runs with color is to use either Direct2D's or
-DirectWrite's `DrawGlyphRunWithColor` method.
+This section shows a sample `TextRenderer::DrawGlyphRun` method that implements the `DrawGlyphRun`
+method of the `IDWriteTextRenderer` interface. The `IDWriteTextRenderer` callback interface is used
+by `IDWriteTextLayout::Draw` to render glyph runs, underlines, and so on. The use of a callback
+interface enables the text layout API to be decoupled from the specific graphics engine. For
+example, it is possible to implement `IDWriteTextRenderer` in terms of Direct2D, in terms of
+`IDWriteBitmapRenderTarget`, or in some other way. This example uses Direct2D.
 
-The following example shows a lower-level approach, in which the client calls
-`TranslateColorGlyphRun` to translate the base glyph run into a sequence of color glyph runs, each
-of which must then be rendered in accordance with its glyph image format. This example assumes the
-client using DirectWrite along with Direct2D.
+The example includes only the `DrawGlyphRun` method. The rest of the `TextRenderer` class is omitted
+for brevity, but the example assumes the class implements the `IDWriteTextRenderer` interface and
+has the following member variables:
 
 ```c++
-HRESULT DrawGlyphRunWithColor(
-    IDWriteFactory8* dwriteFactory,
-    ID2D1DeviceContext7* deviceContext,
-    DWRITE_GLYPH_RUN const& glyphRun,
-    D2D_POINT_2F baselineOrigin,
-    UINT32 colorPaletteIndex,
-    ID2D1Brush* foregroundBrush
-)
+wil::com_ptr<IDWriteFactory4> m_dwriteFactory;
+wil::com_ptr<ID2D1DeviceContext4> m_deviceContext;
+wil::com_ptr<ID2D1Brush> m_foregroundBrush;
+```
+
+The example method renders color glyphs in one of two ways. The easy way is to call
+`ID2D1DeviceContext7::DrawGlyphRunWithColor`. However, the example also includes a fallback code
+path in case the available version of Direct2D does not implement the `ID2D1DeviceContext7`
+interface. The fallback code path uses `TranslateColorGlyphRun`.
+
+```c++
+// Implementation of IDWriteTextRenderer::DrawGlyphRun
+HRESULT STDMETHODCALLTYPE TextRenderer::DrawGlyphRun(
+    _In_opt_ void* clientDrawingContext,
+    FLOAT baselineOriginX,
+    FLOAT baselineOriginY,
+    DWRITE_MEASURING_MODE measuringMode,
+    _In_ DWRITE_GLYPH_RUN const* glyphRun,
+    _In_ DWRITE_GLYPH_RUN_DESCRIPTION const* glyphRunDescription,
+    _In_opt_ IUnknown* clientDrawingEffect
+) noexcept
 {
-    // Get the current world transform for the device context.
-    D2D1_MATRIX_3X2_F transform;
-    deviceContext->GetTransform(&transform);
+    D2D_POINT_2F baselineOrigin{ baselineOriginX, baselineOriginY };
 
-    // Multiply the world transform by the DPI scale to get a world and DPI transform.
+    // Determine the foreground brush to use for this glyph run.
+    // This is the m_foregroundBrush member unless another brush
+    // is specified via the clientDrawingEffect parameter.
+    ID2D1Brush* foregroundBrush = m_foregroundBrush.get();
+    wil::com_ptr<ID2D1Brush> effectBrush;
+    if (clientDrawingEffect != nullptr &&
+        SUCCEEDED(clientDrawingEffect->QueryInterface(&effectBrush)))
+    {
+        foregroundBrush = effectBrush.get();
+    }
+
+    // If available, use the ID2D1DeviceContext7 interface to draw
+    // the glyph run with color the "easy" way.
+    wil::com_ptr<ID2D1DeviceContext7> deviceContext7;
+    if (SUCCEEDED(m_deviceContext->QueryInterface(&deviceContext7)))
+    {
+        deviceContext7->DrawGlyphRunWithColor(
+            baselineOrigin,
+            glyphRun,
+            foregroundBrush,
+            measuringMode,
+            /*colorPaletteIndex*/ 0
+        );
+        return S_OK;
+    }
+
+    // If we fall through to here, we're using an older version of
+    // Direct2D that does not implement the ID2D1DeviceContext7 interface.
+    // We will need to call TranslateColorGlyphRun. First determine the
+    // world to device transform.
+    D2D_MATRIX_3X2_F transform;
+    m_deviceContext->GetTransform(&transform);
     float dpiX, dpiY;
-    deviceContext->GetDpi(&dpiX, &dpiY);
-    transform = transform * D2D1::Matrix3x2F::Scale(dpiX / 96, dpiY / 96);
+    m_deviceContext->GetDpi(&dpiX, &dpiY);
+    transform = transform * D2D1::Matrix3x2F::Scale(dpiX, dpiY);
 
-    // Combination of image formats we support.
+    // Combination of image formats we support. Do *not* include
+    // DWRITE_GLYPH_IMAGE_FORMATS_COLR_PAINT_TREE because this code
+    // path is for an older version of D2D which does not support
+    // the ID2D1DeviceContext7 interface.
     constexpr DWRITE_GLYPH_IMAGE_FORMATS desiredGlyphImageFormats =
         DWRITE_GLYPH_IMAGE_FORMATS_TRUETYPE |
         DWRITE_GLYPH_IMAGE_FORMATS_CFF |
@@ -1268,27 +1344,19 @@ HRESULT DrawGlyphRunWithColor(
         DWRITE_GLYPH_IMAGE_FORMATS_PNG |
         DWRITE_GLYPH_IMAGE_FORMATS_JPEG |
         DWRITE_GLYPH_IMAGE_FORMATS_TIFF |
-        DWRITE_GLYPH_IMAGE_FORMATS_PREMULTIPLIED_B8G8R8A8 |
-        DWRITE_GLYPH_IMAGE_FORMATS_COLR_PAINT_TREE;
+        DWRITE_GLYPH_IMAGE_FORMATS_PREMULTIPLIED_B8G8R8A8;
 
-    // Color glyph runs of type DWRITE_GLYPH_IMAGE_FORMATS_COLR_PAINT_TREE will be
-    // rendered using D2D's DrawPaintGlyphRun, so get the paint feature level from D2D.
-    DWRITE_PAINT_FEATURE_LEVEL paintFeatureLevel = deviceContext->GetColorPaintFeatureLevel();
-
-    // Call TranslateColorGlyphRun to translate the monochrome "base" glyph run
-    // into a sequence of color glyph runs.
+    // Perform color translation.
     wil::com_ptr<IDWriteColorGlyphRunEnumerator1> colorRunEnumerator;
-
-    HRESULT hr = dwriteFactory->TranslateColorGlyphRun(
+    HRESULT hr = m_dwriteFactory->TranslateColorGlyphRun(
         baselineOrigin,
-        &glyphRun,
-        /*glyphRunDescription*/ nullptr,
+        glyphRun,
+        glyphRunDescription,
         desiredGlyphImageFormats,
-        paintFeatureLevel,
-        DWRITE_MEASURING_MODE_NATURAL,
+        measuringMode,
         reinterpret_cast<DWRITE_MATRIX const*>(&transform),
-        colorPaletteIndex,
-        &colorRunEnumerator
+        /*colorPaletteIndex*/ 0,
+        /*out*/ & colorRunEnumerator
     );
 
     // Handle DWRITE_E_NOCOLOR, which is returned if the base glyph run does not
@@ -1296,16 +1364,16 @@ HRESULT DrawGlyphRunWithColor(
     // just render the base glyph run as a monochrome glyph run.
     if (hr == DWRITE_E_NOCOLOR)
     {
-        deviceContext->DrawGlyphRun(
+        m_deviceContext->DrawGlyphRun(
             baselineOrigin,
-            &glyphRun,
+            glyphRun,
             foregroundBrush,
-            DWRITE_MEASURING_MODE_NATURAL
+            measuringMode
         );
         return S_OK;
     }
 
-    // Any other error returned by TranslateColorGyphRun is a failure.
+    // Any other failure HRESULT from TranslateColorGlyphRun is an error.
     RETURN_IF_FAILED(hr);
 
     // Solid color brush to be lazily created if needed.
@@ -1337,7 +1405,10 @@ HRESULT DrawGlyphRunWithColor(
             if (solidBrush == nullptr)
             {
                 // Lazily create the solid color brush with the specified color.
-                RETURN_IF_FAILED(deviceContext->CreateSolidColorBrush(colorGlyphRun->runColor, &solidBrush));
+                RETURN_IF_FAILED(m_deviceContext->CreateSolidColorBrush(
+                    colorGlyphRun->runColor,
+                    &solidBrush
+                    ));
             }
             else
             {
@@ -1360,7 +1431,7 @@ HRESULT DrawGlyphRunWithColor(
         case DWRITE_GLYPH_IMAGE_FORMATS_JPEG:
         case DWRITE_GLYPH_IMAGE_FORMATS_TIFF:
         case DWRITE_GLYPH_IMAGE_FORMATS_PREMULTIPLIED_B8G8R8A8:
-            deviceContext->DrawColorBitmapGlyphRun(
+            m_deviceContext->DrawColorBitmapGlyphRun(
                 colorGlyphRun->glyphImageFormat,
                 baselineOrigin,
                 &colorGlyphRun->glyphRun,
@@ -1370,28 +1441,21 @@ HRESULT DrawGlyphRunWithColor(
             break;
 
         case DWRITE_GLYPH_IMAGE_FORMATS_SVG:
-            deviceContext->DrawSvgGlyphRun(
+            m_deviceContext->DrawSvgGlyphRun(
                 baselineOrigin,
                 &colorGlyphRun->glyphRun,
                 runBrush,
                 /*SVG style*/ nullptr,
-                colorPaletteIndex,
-                colorGlyphRun->measuringMode
-            );
-            break;
-
-        case DWRITE_GLYPH_IMAGE_FORMATS_COLR_PAINT_TREE:
-            deviceContext->DrawPaintGlyphRun(
-                baselineOrigin,
-                &colorGlyphRun->glyphRun,
-                runBrush,
-                colorPaletteIndex,
+                /*colorPaletteIndex*/ 0,
                 colorGlyphRun->measuringMode
             );
             break;
 
         default:
-            deviceContext->DrawGlyphRun(
+            // Treat any other format as a monochrome glyph run.
+            // This is used for monochrome glyphs, or for each layer
+            // of a glyph in DWRITE_GLYPH_IMAGE_FORMATS_COLR.
+            m_deviceContext->DrawGlyphRun(
                 baselineOrigin,
                 &colorGlyphRun->glyphRun,
                 colorGlyphRun->glyphRunDescription,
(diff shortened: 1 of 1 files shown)


specs/dwritecore/DWriteCore.md
```diff
@@ -510,9 +510,189 @@ The following other DirectWrite APIs also use the hybrid font selection algorith
 -   `IDWriteTextLayout` if the typographic font family model is enabled by calling
     `IDWriteTextLayout4::SetFontAxisValues` or `IDWriteTextLayout4::SetAutomaticFontAxes`
 
-# Examples
-
-## Creating an isolated factory
+## Color Font Support
```
niklasb-ms, line 513: Most of this section and its subsections describe _existing_ APIs.
I've added comments to call out new APIs.

SeattlePseudoGenius (commented):
"Translating"

specs/dwritecore/DWriteCore.md
```diff
@@ -1050,9 +1230,185 @@ int __cdecl wmain(int argc, WCHAR** argv)
...
+        if (!haveRun)
+            break;
+
+        // Get a pointer to the color glyph run structure.
+        DWRITE_COLOR_GLYPH_RUN1 const* colorGlyphRun;
+        RETURN_IF_FAILED(colorRunEnumerator->GetCurrentRun(&colorGlyphRun));
+
+        // Determine the brush to use for this color glyph run.
+        ID2D1Brush* runBrush;
+        if (colorGlyphRun->paletteIndex == 0xFFFF)
```
SeattlePseudoGenius, line 1329: Where did this constant come from? Was it somehow defined by the caller in the original glyph run, or is it some systemwide API magic constant?

niklasb-ms: It's a magic palette index meaning the foreground color.

specs/dwritecore/DWriteCore.md
```diff
@@ -510,9 +510,189 @@ The following other DirectWrite APIs also use the hybrid font selection algorith
...
+## Color Font Support
+
+By default, a glyph has a shape but no intrinsic color. Both DirectWrite and Direct2D have
+`DrawGlyphRun` methods that render glyph runs by filling the glyph shapes with a specified text
+color. For convenience, we will refer to this as _monochrome_ glyph rendering. All fonts have
+monochrome glyphs, but a color font also has color representations of some glyphs. To render glyphs
+in color, an application must use different glyph rendering APIs instead of calling the monochrome
+`DrawGlyphRun` method.
+
+Note that color font support does not affect text layout it at all. The text layout always produces
```
SeattlePseudoGenius, line 522: What exactly is 'text layout' in this context? Another separate function of DWrite? Could we be a little more specific here, perhaps referencing the relevant APIs?

specs/dwritecore/DWriteCore.md
```diff
@@ -510,9 +510,189 @@ The following other DirectWrite APIs also use the hybrid font selection algorith
...
+monochrome glyph runs. Color is enabled at the glyph rendering level by substituting color
+representations for the monochrome glyphs.
+
+The easiest way to enable color is to call the `DrawGlyphRunWithColor` method instead of calling
```
SeattlePseudoGenius, line 526: It's unfortunate because DrawGlyphRunWithColor sounds like it should take a color. Essentially, DrawGlyphRun and DrawGlyphRunWithColor are backwards from what you'd expect.

What about DrawColorGlyphRun? The name feels slightly more specific to the purpose and different from the confusion.

specs/dwritecore/DWriteCore.md
```diff
@@ -510,9 +510,189 @@ The following other DirectWrite APIs also use the hybrid font selection algorith
...
+of _color glyph runs_. There are multiple ways of representiong color glyphs, known as _glyph image
+formats_ and identified by the `DWRITE_GLYPH_IMAGE_FORMATS` enumeration. When calling
+`TranslateColorGlyphRun`, the client specifies the combination of glyph image formats it supports.
+The `TranslateColorGlyphRun` returns a _glyph run enumerator_ object
+(`IDWriteColorGlyphRunEnumerator1` interface), which the client uses to iterate over the sequence of
+color glyph runs, represented by the `DWRITE_COLOR_GLYPH_RUN1` structure. Each color glyph run has a
+glyph image format. The client branches on the color run's glyph image format to determine how to
+render the color run.
+
+If the base glyph run does not have color glyphs in any of the requested glyph image formats, the
```
SeattlePseudoGenius, line 547: What happens if the base glyph run contains some color glyphs that the client doesn't specify/support? Are those glyphs "missing" from the final iterator?

specs/dwritecore/DWriteCore.md
```diff
@@ -510,9 +510,189 @@ The following other DirectWrite APIs also use the hybrid font selection algorith
...
+The calling pattern used with this image format is different than with all other glyph image
+formats. That's because what gets rendered at the end of the day is just a series of ordinary,
+monochrome glyph runs -- just with different glyph IDs and different colors than were specified to
+begin with. Because the enumerated color glyph runs are rendered as ordinary monochrome glyph runs,
+their glyph image formats are either `DWRITE_GLYPH_IMAGE_FORMATS_TRUETYPE` or
+`DWRITE_GLYPH_IMAGE_FORMATS_CFF`.
+
+For example, suppose the base glyph run specifies a smiley face glyph. If we rendered the base glyph
+run using `DrawGlyphRun`, the visual result would be a monochrome representation of a smiley face
+(the base glyph ID). Thd COLR table in the font maps this base glyph ID to an array of _layer
```
SeattlePseudoGenius, line 577: "The"

specs/dwritecore/DWriteCore.md
niklasb-ms: _Text layout_ is the process of converting Unicode text + formatting information to a sequence of glyph runs. DWrite has a text layout API (IDWriteTextLayout), but applications can also implement their own text layout. _Text rendering_ takes the glyph runs produced by text layout and converts them to pixels.

specs/dwritecore/DWriteCore.md
```diff
@@ -1050,9 +1230,185 @@ int __cdecl wmain(int argc, WCHAR** argv)
...
+        {
+        case DWRITE_GLYPH_IMAGE_FORMATS_NONE:
+            // do nothing
+            break;
+
+        case DWRITE_GLYPH_IMAGE_FORMATS_PNG:
+        case DWRITE_GLYPH_IMAGE_FORMATS_JPEG:
+        case DWRITE_GLYPH_IMAGE_FORMATS_TIFF:
+        case DWRITE_GLYPH_IMAGE_FORMATS_PREMULTIPLIED_B8G8R8A8:
+            deviceContext->DrawColorBitmapGlyphRun(
```
SeattlePseudoGenius, line 1363: I can't help but notice that all of these calls, more or less, take essentially the same arguments, and are essentially just passed the glyph run to render for the details.

Why have all these different calls then? Why not just have a "DrawSuperNewCoolGlyphRun", pass it the glyph run, and let it worry about the type?

Perhaps just my take knowing virtually nothing about DWrite or glyph rendering at the moment, but I'm struggling to understand the reasoning here, or what splitting these out buys us.

specs/dwritecore/DWriteCore.md
SeattlePseudoGenius: This feels like it warrants a constexpr definition, or at least, a #define.

specs/dwritecore/DWriteCore.md
niklasb-ms: `DrawGlyphRunWithColor` does take a color parameter. The specified color is used to render monochrome glyphs or sometimes even for parts of color glyphs (since a color glyph can reference the foreground color). The "WithColor" in the name is intended to mean "with color fonts enabled."

I thought of `DrawColorGlyphRun`, but I thought that might be confusing because there is an existing `DWRITE_COLOR_GLYPH_RUN` structure, which is returned by `IDWriteColorGlyphRunEnumerator`. Thus, `DrawColorGlyphRun` sounds like the name of a function you would call to draw each of the enumerated color runs after calling `TranslateColorGlyphRun`. That's not what this proposed function is.

Perhaps a better name would be `DrawGlyphRunWithColorFontsEnabled`? It's a bit long, but more explicit. It also nicely parallels an existing flag, `D2D1_DRAW_TEXT_OPTIONS_ENABLE_COLOR_FONT`, which can be passed to the `ID2D1RenderTarget::DrawTextLayout` method.

specs/dwritecore/DWriteCore.md
```diff
@@ -1235,3 +1591,906 @@ DWRITE_BEGIN_INTERFACE(IDWriteFactory8, "202D44F0-84AC-4D6F-B8D0-704B27675CAA")
...
+/// DWRITE_GLYPH_IMAGE_FORMATS_COLR_PAINT_TREE image format represents a color glyph that can be
+/// rendered by drawing paint elements returned by IDWritePaintReader.
+/// </summary>
+#ifndef DWRITE_GLYPH_IMAGE_FORMATS_COLR_PAINT_TREE_DEFINED
+#define DWRITE_GLYPH_IMAGE_FORMATS_COLR_PAINT_TREE_DEFINED
+#define DWRITE_GLYPH_IMAGE_FORMATS_COLR_PAINT_TREE static_cast<DWRITE_GLYPH_IMAGE_FORMATS>(0x00000100)
+#endif
+```
+
+### DWRITE_PAINT_FEATURE_LEVEL Enumeration
```
SeattlePseudoGenius, line 1683: We've got all these paint-specific API additions, but no example to show their usage - the new example simply uses D2D and calls DrawPaintGlyphRun.

I get that the paint API is complex and the original example omitted details for brevity, but if we're adding all these paint APIs as well, we should have an example showing the usage. Finding a good balance of coverage while not going overboard on small details will be key.

specs/dwritecore/DWriteCore.md
niklasb-ms: This is in the existing API, but I agree a #define would make sense.
Maybe DWRITE_PALETTE_INDEX_FOREGROUND?

specs/dwritecore/DWriteCore.md
niklasb-ms: Color glyphs in an unsupported format are treated the same as monochrome glyphs.

If there are _only_ monochrome glyphs (or only color glyphs in unsupported formats) then TranslateColorGlyphRun returns DWRITE_E_NOCOLOR, in which case the client should render the base glyph run.

If there's a mix of color and monochrome glyphs, then TranslateColorGlyphRun chops up the run. Color glyphs are translated to color runs as described in the overview section. Monochrome glyphs are returned as color runs with DWRITE_GLYPH_IMAGE_FORMATS_TRUETYPE or DWRITE_GLYPH_IMAGE_FORMATS_CFF and with paletteIndex = 0xFFFF. Then client then renders these as monochrome glyph runs using the foreground brush.

specs/dwritecore/DWriteCore.md
SeattlePseudoGenius: I see, so when you say "the text layout always produces monochrome glyph runs", do you mean IDWriteTextLayout always produces them?

specs/dwritecore/DWriteCore.md
niklasb-ms: I wrote a console application called DumpPaint that outputs the paint tree for a glyph. I was thinking of turning that into a WindowsAppSDK sample, but I could also adapt part of it to be a sample in the spec.

specs/dwritecore/DWriteCore.md
```diff
@@ -1050,9 +1230,185 @@ int __cdecl wmain(int argc, WCHAR** argv)
 }
 ```
 
-# API Details
+### Rendering Color Glyph Runs
+
+In most cases, the simplest way to render glyph runs with color is to use either Direct2D's or
+DirectWrite's `DrawGlyphRunWithColor` method.
```
SeattlePseudoGenius, line 1236: Though I realize it's almost trivial, it would be good to have a short example of this as well.

specs/dwritecore/DWriteCore.md
```diff
@@ -1050,9 +1230,185 @@ int __cdecl wmain(int argc, WCHAR** argv)
...
 
-## DWRITE_FACTORY_TYPE Enumeration
+The following example shows a lower-level approach, in which the client calls
+`TranslateColorGlyphRun` to translate the base glyph run into a sequence of color glyph runs, each
```
SeattlePseudoGenius, line 1239: Just so I understand what the Translate is actually doing - if you have a color font and create a default glyph run, you can render it in monochrome, but secretly, it stores color information internally so that if the client wishes to render in color, this translation will use that secret color information to create a sequence of glyph runs with color information?

specs/dwritecore/DWriteCore.md
SeattlePseudoGenius: When I read the text, I read that if no glyphs support any of the requested formats, you get back DWRITE_E_NOCOLOR. However I believe the text leaves the case ambiguous where only some of the glyphs are in the supported format. Can we try to clarify the text a bit? Specifically mentioning that the subset of glyphs which are not of the supported types will be "downconverted" to monochrome glyphs?

And on that note, if the downconversion takes place in that case where a subset of glyphs are supported, why do we need DWRITE_E_NOCOLOR at all? Why not just downconvert the entire set of glyph runs to monochrome glyphs, to eliminate the need for the client to handle that one error case?

niklasb-ms: Returning DWRITE_E_NOCOLOR means we don't have to instantiate the enumerator object. Yes, I can better explain this case.

SeattlePseudoGenius: I would argue that avoiding instantiating the enumerator object in this one-off, somewhat rare case, is not worth the additional API complexity of having to check for that special return value and then calling the monochrome method, every time this method is used. Especially true if conversion to monochrome already happens for color glyphs when their format is unsupported but other glyphs are supported. If what we gain from this is the cost of one allocation in a rare case, I think the reduced API complexity is preferable.

niklasb-ms: This is existing API behavior going back to Windows 8.1. The thinking was that the client might call TranslateColorGlyphRun on every glyph run regardless of the font, and for a non-color font this would return DWRITE_E_NOCOLOR. Given this calling pattern, it's not a rare case.

specs/dwritecore/DWriteCore.md
SeattlePseudoGenius: Understood regarding the concerns with the existing type struct. Of all the ideas mentioned, I suppose I do prefer the name you ended up with, DrawGlyphRunWithColor

specs/dwritecore/DWriteCore.md
niklasb-ms: Well, D2D could have a `DrawColorGlyphRun` method that takes a `DWRITE_COLOR_GLYPH_RUN1` and internally has the switch statement. In that case, the client would call `TranslateColorGlyphRun` and then forward each enumerated color glyph run to `DrawColorGlyphRun`. However, if you're going to go that far then it makes more sense to have a higher-level method that calls `TranslateColorGlyphRun` _for_ you, enumerates all the color runs, and then forwards each to the appropriate method. That's what the proposed `DrawGlyphRunWithColor` (or `DrawGlyphRunWithColorFontsEnabled`?) method does.

There are also cases where some clients want to handle different cases differently. E.g., XAML has different handling of monochrome vs. color glyph runs. Monochrome glyph runs are rendered to an alpha texture, while color runs are rendered to a BGRA32 texture.

specs/dwritecore/DWriteCore.md
niklasb-ms: Yes, that's the general idea (except the "secretly" part 😂).

specs/dwritecore/DWriteCore.md
niklasb-ms: I went with `DWRITE_NO_PALETTE_INDEX` since it's more general. It just means a palette index is not specify, which depending on context could mean use the foreground color or just that the color is not relevant.

specs/dwritecore/DWriteCore.md
niklasb-ms: I've added some additional text to hopefully clarify this.

specs/dwritecore/DWriteCore.md
niklasb-ms: I've added the DumpPaintTree sample in the next iteration.

specs/dwritecore/DWriteCore.md
niklasb-ms: I've replaced this with a new sample that has two code paths. If the `ID2D1DeviceContext7` is available, it calls `DrawGlyphRunWithColor`. Otherwise, it calls `TranslateColorGlyphRun`.

specs/dwritecore/DWriteCore.md
niklasb-ms: I'll stick with DrawGlyphRunWithColor, then, unless API reviewers have a better idea.

specs/dwritecore/DWriteCore.md
niklasb-ms: Fixed.

specs/dwritecore/DWriteCore.md
```diff
@@ -1050,9 +1258,591 @@ int __cdecl wmain(int argc, WCHAR** argv)
...
+        if (!haveRun)
+            break;
+
+        // Get a pointer to the color glyph run structure.
+        DWRITE_COLOR_GLYPH_RUN1 const* colorGlyphRun;
+        RETURN_IF_FAILED(colorRunEnumerator->GetCurrentRun(&colorGlyphRun));
+
+        // Determine the brush to use for this color glyph run.
+        ID2D1Brush* runBrush;
+        if (colorGlyphRun->paletteIndex == 0xFFFF)
```
SeattlePseudoGenius, line 1397: Can we change this to use the new #define you added?

specs/dwritecore/DWriteCore.md
```diff
@@ -1050,9 +1258,591 @@ int __cdecl wmain(int argc, WCHAR** argv)
...
+        {
+        case DWRITE_GLYPH_IMAGE_FORMATS_NONE:
+            // do nothing
+            break;
+
+        case DWRITE_GLYPH_IMAGE_FORMATS_PNG:
+        case DWRITE_GLYPH_IMAGE_FORMATS_JPEG:
+        case DWRITE_GLYPH_IMAGE_FORMATS_TIFF:
+        case DWRITE_GLYPH_IMAGE_FORMATS_PREMULTIPLIED_B8G8R8A8:
+            m_deviceContext->DrawColorBitmapGlyphRun(
```
SeattlePseudoGenius, line 1434: These 3 methods share common arguments, such as baselineOrigin, glyphRun, measuringMode, and so on, but the order of these various arguments seems to differ across the methods. Ideally we should make sure the ordering of the common arguments is consistent as possible across the methods. For example, here baselineOrigin is second in the first method, but first in the other methods, and other common params have differing positions/orders in the other methods

Recommend all common parameters come first in these methods, then method-specific params come second. Unclear to me if the ordering differences here are intentional and have rationale or not.

specs/dwritecore/DWriteCore.md
```diff
@@ -1050,9 +1258,591 @@ int __cdecl wmain(int argc, WCHAR** argv)
...
+                          - color: { 0, 0.501961, 0, 1 }
+                        D2D1_GRADIENT_STOP:
+                          - position: 0.5
+                          - color: { 1, 1, 1, 1 }
+                        D2D1_GRADIENT_STOP:
+                          - position: 1
+                          - color: { 1, 0, 0, 1 }
+```
+
+For convenience, the sample defines `Indent` and `PropName` helper types with assocaited stream
```
SeattlePseudoGenius, line 1521: "Associated"

specs/dwritecore/DWriteCore.md
```diff
@@ -1050,9 +1258,591 @@ int __cdecl wmain(int argc, WCHAR** argv)
...
+                << indent << PropName{ "position" } << stop.position << '\n'
+                << indent << PropName{ "color" } << stop.color << '\n';
+        }
+        indent.value--;
+    };
+
+    // Helper to recursively call DumpPaintElement for a child paint element.
+    auto Recurse = [&]()
+    {
+        DumpPaintElement(out, Indent{ indent.value + 1 }, reader, element);
```
SeattlePseudoGenius, line 1581: NIT: Recommend considering some sort of "AutoIndent" class which indents/unindents based on constructor/destructor. Will avoid the need for all the +1/-1 indent logic.

specs/dwritecore/DWriteCore.md
```diff
@@ -1050,9 +1258,591 @@ int __cdecl wmain(int argc, WCHAR** argv)
...
+        break;
+    }
+
+    case DWRITE_PAINT_TYPE_COMPOSITE:
+    {
+        auto const& paint = element.paint.composite;
+        out << indent << "DWRITE_PAINT_TYPE_COMPOSITE:\n"
+            << indent << PropName{ "mode" } << paint.mode << '\n'
+            << indent << PropName{ "children (source, destination)" } << '\n';
+        WriteChildren(2);
```
SeattlePseudoGenius, line 1705: We have all these implicit child counts for various paint types - this sample seems to imply that you need to know how many children are present for each type.

However, it seems that the child count will always be specified in paint.childCount for all paintTypes. Is this correct? If so, I recommend changing all instances of WriteChildren(some constant) with WriteChildren(paint.childCount) in this example. This shows a better API client practice of always using the explicit child count specified by the API rather than assuming the count for any given paint type.

If you're trying to show the expected child count for each paint type, you could perhaps mention it in a comment after each call.

specs/dwritecore/DWriteCore.md
```diff
@@ -1235,3 +2025,906 @@ DWRITE_BEGIN_INTERFACE(IDWriteFactory8, "202D44F0-84AC-4D6F-B8D0-704B27675CAA")
...
+### DWRITE_PAINT_FEATURE_LEVEL Enumeration
+
+```c++
+/// <summary>
+/// Defines known feature level for use with the IDWritePaintReader interface and
+/// related APIs. A feature level represents a level of functionality. For example, it
+/// determines what DWRITE_PAINT_TYPE values might be returned.
+/// </summary>
+/// <remarks>
+/// See the DWRITE_PAINT_TYPE enumration for which paint types are required for each
```
SeattlePseudoGenius, line 2126: "enumeration"

specs/dwritecore/DWriteCore.md
```diff
@@ -1235,3 +2025,906 @@ DWRITE_BEGIN_INTERFACE(IDWriteFactory8, "202D44F0-84AC-4D6F-B8D0-704B27675CAA")
...
+/// </summary>
+struct DWRITE_PAINT_COLOR
+{
+    /// <summary>
+    /// Color value.
+    /// </summary>
+    DWRITE_COLOR_F value;
+
+    /// <summary>
+    /// Specifies the palette entry index if the color value comes from the font's color palette.
```
SeattlePseudoGenius, line 2192: Does the application have any insight into the font's actual color palette? In your examples, we always either use an index of 0, or we detect the forground brush palette index (0xffff) and use the foreground color. Are there ever any cases when an app would choose a different palette index?

specs/dwritecore/DWriteCore.md
```diff
@@ -1235,3 +2025,906 @@ DWRITE_BEGIN_INTERFACE(IDWriteFactory8, "202D44F0-84AC-4D6F-B8D0-704B27675CAA")
...
+
+        /// <summary>
+        /// Valid for paint elements of type DWRITE_PAINT_TYPE_SOLID.
+        /// Specifies a solid color used to fill the current shape or clip.
+        /// This paint element has no child elements.
+        /// </summary>
+        /// <remarks>
+        /// This corresponds to a PaintSolid or PaintVarSolid record in the OpenType COLR table.
+        /// </remarks>
+        typedef DWRITE_PAINT_COLOR PAINT_SOLID;
```
SeattlePseudoGenius, line 2363: Does this typedef really buy us anything? It's only used below. I suggest you just define solid as a DWRITE_PAINT_COLOR.

specs/dwritecore/DWriteCore.md
```diff
@@ -1235,3 +2025,906 @@ DWRITE_BEGIN_INTERFACE(IDWriteFactory8, "202D44F0-84AC-4D6F-B8D0-704B27675CAA")
...
+            float y1;
+
+            /// <summary>
+            /// Radius of the end circle.
+            /// </summary>
+            float radius1;
+        } radialGradient;
+
+        /// <summary>
+        /// Valid for paint elements of type DWRITE_PAINT_TYPE_SWEEPL_GRADIENT.
```
SeattlePseudoGenius, line 2473: SWEEP (not SWEEPL)

specs/dwritecore/DWriteCore.md
```diff
@@ -1235,3 +2025,906 @@ DWRITE_BEGIN_INTERFACE(IDWriteFactory8, "202D44F0-84AC-4D6F-B8D0-704B27675CAA")
...
+
+        /// <summary>
+        /// Valid for paint elements of type DWRITE_PAINT_TYPE_TRANSFORM.
+        /// Specifies an affine transform to be applied to child content.
+        /// This paint element has one child element, which is the transformed content.
+        /// </summary>
+        /// <remarks>
+        /// This corresponds to paint formats 12 through 31 in the OpenType COLR table.
+        /// </remarks>
+        typedef DWRITE_MATRIX PAINT_TRANSFORM;
```
SeattlePseudoGenius, line 2567: Another case where this typedef doesn't really seem like it's worthwhile.

specs/dwritecore/DWriteCore.md
```diff
@@ -1235,3 +2025,906 @@ DWRITE_BEGIN_INTERFACE(IDWriteFactory8, "202D44F0-84AC-4D6F-B8D0-704B27675CAA")
...
+    /// </returns>
+    /// <remarks>
+    /// If the specified glyph index is not a color glyph, the method succeeds, but the paintType
+    /// member of the DWRITE_PAINT_ELEMENT structure is set to DWRITE_PAINT_TYPE_NONE. In this
+    /// case, the application should draw the input glyph as a non-color glyph.
+    /// </remarks>
+    STDMETHOD(SetCurrentGlyph)(
+        UINT32 glyphIndex,
+        _Out_writes_bytes_(structSize) DWRITE_PAINT_ELEMENT* paintElement,
+        UINT32 structSize,
```
SeattlePseudoGenius, line 2623: Why not have this overload define the type of paintElement as void* ? It seems kind of weird for the type to be DWRITE_PAINT_ELEMENT, however the call itself will end up writing an arbitrary number of bytes into the out parameter. This is a logical implication that the type should not be set.

specs/dwritecore/DWriteCore.md
```diff
@@ -1235,3 +2025,906 @@ DWRITE_BEGIN_INTERFACE(IDWriteFactory8, "202D44F0-84AC-4D6F-B8D0-704B27675CAA")
...
+    /// <returns>
+    /// Standard HRESULT error code. The return value is E_INVALIDARG if the current paint element doesn't have
+    /// a next sibling.
+    /// </returns>
+    /// <remarks>
+    /// Whether a paint element has children (and how many) can be determined a priori from its paint type and
+    /// properties. For more information, see DWRITE_PAINT_ELEMENT.
+    /// </remarks>
+    STDMETHOD(MoveToNextSibling)(
+        _Out_writes_bytes_(structSize) DWRITE_PAINT_ELEMENT* paintElement,
```
SeattlePseudoGenius, line 2724: Another variable-sized argument typing concern

specs/dwritecore/DWriteCore.md
niklasb-ms: I disagree. The only paint type with a variable number of children is LAYERS. For all other paint types, the child count is known _a priori_ and is inherent to the semantics of the paint type and cannot change without breaking those semantics. For example, COMPOSITE has two inputs: the source and destination of the compositing operation. The render a COMPOSITE, a client renders the first (source) child to an intermediate, renders the second (destination) child to another intermediate, and then composes the source onto the destination using the specified mode. Given these semantics, what could you possibly do with fewer or greater than 2 children? If there were an explicit _childCount_ member, what could you do with it apart from `assert(childCount == 2)`?

specs/dwritecore/DWriteCore.md
niklasb-ms: Thanks.

specs/dwritecore/DWriteCore.md
niklasb-ms: There are two kinds of indices. A font can have multiple palettes. A _color palette index_ determines which color palette is used and is specified by the client. A _palette entry index_ selects an entry within a palette and is specified by a record in the font (part of a paint record or gradient stop record or COLR v0 layer record).

Zero is a special color palette index, meaning the default palette. Other color palette indices depend on the font and have no standard meaning. The client specifies the color palette index as a parameter to `TranslateColorGlyphRun` or `IDWritePaintReader::SetColorPaletteIndex`.

0xFFFF is a special palette entry index, meaning the foreground brush. In all other cases, the color `value` is determined by using the paletteEntryIndex as an index into the client-specified color palette. The color's alpha value is then multiplied by the `alphaMultiplier`, which also comes from a color record in the font.

It is also possible for a client to define a custom palette instead of using a color palette from the font. CSS allows this. This is one reason `DWRITE_PAINT_COLOR` exposes the information it does, so a client with a custom palette or other custom behavior has the information it needs to compute the color value for a paint element on its own.

specs/dwritecore/DWriteCore.md
```diff
@@ -1235,3 +2025,906 @@ DWRITE_BEGIN_INTERFACE(IDWriteFactory8, "202D44F0-84AC-4D6F-B8D0-704B27675CAA")
...
+    DWRITE_COLOR_F value;
+
+    /// <summary>
+    /// Specifies the palette entry index if the color value comes from the font's color palette.
+    /// Otherwise, this member is DWRITE_NO_PALETTE_INDEX (0xFFFF).
+    /// </summary>
+    UINT16 paletteEntryIndex;
+
+    /// <summary>
+    /// Specifies an alpha value multiplier in the range 0 to 1 that was used to compute the color
```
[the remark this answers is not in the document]
niklasb-ms: No, it's not premultiplied. I've added that the comment for the `value` member. The `alphaMultiplier` is just multiplied with the alpha value of the palette entry to compute the 'a' value of the `value` member.

specs/dwritecore/DWriteCore.md
niklasb-ms: You're right. This doesn't really buy anything. Same with transform.

specs/dwritecore/DWriteCore.md
niklasb-ms: Thanks. Fixed.

specs/dwritecore/DWriteCore.md
niklasb-ms: Yes, I anticipate discussing this with the API review panel.

I agree it's a little weird, but it's not unprecedented. There are quite a few classic Win32 APIs in which the client is expected to set the `cbSize` field of a structure before passing it to a function that be filled in. This is to enable a structure to be versioned while ensuring we never write past the end if the app was compiled against an older version.

In this case, we may want to define new paint types in the future (for COLR v2 and later). I propose to just add those to the existing union instead of having to define a DWRITE_PAINT_ELEMENT1 along with new interfaces and methods to go with it. That _may_ mean the size of the structure increases. However, it should not break ABI compatibility because the DWRITE_PAINT_FEATURE_LEVEL ensures existing clients only see paint types they know about. The size parameter is just an extra measure to ensure we never write past the end of the structure.

specs/dwritecore/DWriteCore.md
```diff
@@ -1235,3 +2025,906 @@ DWRITE_BEGIN_INTERFACE(IDWriteFactory8, "202D44F0-84AC-4D6F-B8D0-704B27675CAA")
...
+    /// Standard HRESULT error code. The return value is E_INVALIDARG if the current paint element doesn't have
+    /// any children.
+    /// </returns>
+    /// <remarks>
+    /// Whether a paint element has children (and how many) can be determined a priori from its paint type and
+    /// properties. For more information, see DWRITE_PAINT_ELEMENT.
+    /// </remarks>
+    STDMETHOD(MoveToFirstChild)(
+        _Out_writes_bytes_(structSize) DWRITE_PAINT_ELEMENT* paintElement,
+        UINT32 structSize = sizeof(DWRITE_PAINT_ELEMENT)
```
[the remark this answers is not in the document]
niklasb-ms: Same as above.

specs/dwritecore/DWriteCore.md
niklasb-ms: Same as above.

specs/dwritecore/DWriteCore.md
niklasb-ms: Agreed.

specs/dwritecore/DWriteCore.md
niklasb-ms: I looked into this, but I think it would require adding more boilerplate code than it's worth, especially as it's all peripheral to the APIs being demonstrated. I've modified the sample to hopefully be a little clearer. It now passes `indent + 1` at a couple of function call sites but never modifies indent within the function.

specs/dwritecore/DWriteCore.md
niklasb-ms: Thanks!

specs/dwritecore/DWriteCore.md
SeattlePseudoGenius: Is it worth adding a paragraph or two basically saying what you describe here? As a reader I was a little unclear when or why you'd ever intentionally specify a palette/index other than 0 or 0xffff, and how you would understand what colors/palette that index would refer to. That is, if this is not already mentioned in some other part of the spec.

niklasb-ms: I've updated the comments to DWRITE_PAINT_COLOR to explain some of the above.

specs/dwritecore/DWriteCore.md
```diff
@@ -1050,9 +1258,604 @@ int __cdecl wmain(int argc, WCHAR** argv)
...
+    _In_ DWRITE_GLYPH_RUN_DESCRIPTION const* glyphRunDescription,
+    _In_opt_ IUnknown* clientDrawingEffect
+) noexcept
+{
+    D2D_POINT_2F baselineOrigin{ baselineOriginX, baselineOriginY };
+
+    // Determine the foreground brush to use for this glyph run.
+    // This is the m_foregroundBrush member unless another brush
+    // is specified via the clientDrawingEffect parameter.
+    ID2D1Brush* foregroundBrush = m_foregroundBrush.get();
```
ChrisGuzak, line 1302: raw COM pointers are dangerous. It seems you can use wil::com_ptr and avoid the danger. There is another raw pointer below `runBrush`.

specs/dwritecore/DWriteCore.md
```diff
@@ -1050,9 +1258,604 @@ int __cdecl wmain(int argc, WCHAR** argv)
...
+    wil::com_ptr<ID2D1Brush> effectBrush;
```
ChrisGuzak, line 1303: this can be `auto effectBrush = wil::com_copy<ID2D1Brush>(clientDrawingEffect);`

specs/dwritecore/DWriteCore.md
```diff
@@ -1235,3 +2038,912 @@ DWRITE_BEGIN_INTERFACE(IDWriteFactory8, "202D44F0-84AC-4D6F-B8D0-704B27675CAA")
...
+    /// The glyph has raw 32-bit premultiplied BGRA data.
+    /// </summary>
+    DWRITE_GLYPH_IMAGE_FORMATS_PREMULTIPLIED_B8G8R8A8 = 0x00000080,
+
+    /// <summary>
+    /// The glyph is represented by a tree of paint elements in the
+    /// font's COLR table.
+    /// </summary>
+    DWRITE_GLYPH_IMAGE_FORMATS_COLR_PAINT_TREE = 0x00000100,
+};
```
ChrisGuzak, line 2108: for flags enums use `DEFINE_ENUM_FLAG_OPERATORS()`

niklasb-ms: It's there in the existing header file, but I neglected to include it in the spec. Will fix.

specs/dwritecore/DWriteCore.md
```diff
@@ -1235,3 +2038,912 @@ DWRITE_BEGIN_INTERFACE(IDWriteFactory8, "202D44F0-84AC-4D6F-B8D0-704B27675CAA")
...
+    DWRITE_PAINT_ATTRIBUTES_USES_PALETTE = 0x01,
+
+    /// <summary>
+    /// Specifies that the color value (or any color value in the glyph) comes from the client-specified
+    /// text color. This means the appearance may depend on the text color, which may be important to
+    /// clients that cache color glyphs.
+    /// </summary>
+    DWRITE_PAINT_ATTRIBUTES_USES_TEXT_COLOR = 0x02
+};
+#ifdef DEFINE_ENUM_FLAG_OPERATORS
```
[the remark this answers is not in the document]
niklasb-ms: All existing uses of DEFINE_ENUM_FLAG_OPERATORS in DWrite header files and in dcommon.h have the #ifdef. I feel like I should either include it here for consistency or remove all the existing #ifdefs.

specs/dwritecore/DWriteCore.md
niklasb-ms: The two raw COM pointers in this function are just aliases to com_ptrs, so not dangerous IMO. However, this being sample code, I can use com_ptr throughout and incur the unnecessary AddRef/Release overhead.

specs/dwritecore/DWriteCore.md
```diff
@@ -1050,9 +1258,604 @@ int __cdecl wmain(int argc, WCHAR** argv)
...
+    wil::com_ptr<ID2D1Brush> effectBrush;
+    if (clientDrawingEffect != nullptr &&
+        SUCCEEDED(clientDrawingEffect->QueryInterface(&effectBrush)))
+    {
+        foregroundBrush = effectBrush.get();
+    }
+
+    // If available, use the ID2D1DeviceContext7 interface to draw
+    // the glyph run with color the "easy" way.
+    wil::com_ptr<ID2D1DeviceContext7> deviceContext7;
```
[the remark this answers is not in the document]
niklasb-ms: Nice. Reminds me of "if let" in Rust.

specs/dwritecore/DWriteCore.md
niklasb-ms: That would throw an exception if the clientDrawingEffect existed by did not implement ID2D1Brush.
OTOH, wil::try_com_query would AV if clientDrawingEffect was NULL.
I've changed it to use try_com_query with an explicit NULL check.

specs/dwritecore/DWriteCore.md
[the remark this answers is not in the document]
niklasb-ms: For now, I'm leaving in the #ifdef for consistency, but if API reviewers think I should remove them all I can do that.

specs/dwritecore/DWriteCore.md
```diff
@@ -510,9 +510,217 @@ The following other DirectWrite APIs also use the hybrid font selection algorith
...
+color glyph runs, represented by the `DWRITE_COLOR_GLYPH_RUN1` structure. Each color glyph run has a
+glyph image format. The client branches on the color run's glyph image format to determine how to
+render the color run.
+
+If the base glyph run does not have color glyphs in any of the requested glyph image formats, the
+`TranslateColorGlyphRun` method returns `DWRITE_E_NOCOLOR`, in which case the client should render
+the base glyph run as a monochrome glyph run.
+
+Otherwise, the base glyph run may contain a mix of color glyphs in different glyph image formats as
+well as monochrome glyphs. For each base glyph, the `TranslateColorGlyphRun` method determines the
```
oldnewthing, line 555: What happens if the font uses a glyph image format that the caller doesn't support? Do we auto-convert it to something the caller *does* support? I think what happens is that we look for fallback formats that the caller supports, and if no fallback can be found, we render as monochrome.

niklasb-ms: You think correctly. We don't convert, but we only return formats the client supports with monochrome being the ultimate fallback.

specs/dwritecore/DWriteCore.md
```diff
@@ -510,9 +510,217 @@ The following other DirectWrite APIs also use the hybrid font selection algorith
...
+`DrawGlyphRun` method.
+
+Because color font support is implemented at the glyph rendering level, it does not affect text
+layout. This is true whether a client uses the `IDWriteTextLayout` interface or implements its own
+text layout algorithm. The mapping of characters to glyphs and the positioning of those glyphs all
+use monochrome glyph IDs and their associated metrics. The output of the text layout process is a
+sequence of monochrome glyph runs. Color font support can be enabled by translating those monochrome
+"base" glyph runs to color glyphs runs at rendering time.
+
+The easiest way to enable color is to call the `DrawGlyphRunWithColor` method instead of calling
```
FrankC-msft, line 529: [already discussed in previously "Resolved" thread, but surfacing here for broader API Review folks consideration.  Consider DrawGlyphRunWithColorFonts to make it more apparent the color (brush) parameter to this function isn't different than DrawGlyphRun

specs/dwritecore/DWriteCore.md
```diff
@@ -1235,3 +2038,916 @@ DWRITE_BEGIN_INTERFACE(IDWriteFactory8, "202D44F0-84AC-4D6F-B8D0-704B27675CAA")
...
+            /// <summary>
+            /// Number of gradient stops. Use the IDWritePaintReader::GetGradientStops method to
+            /// get the gradient stops.
+            /// </summary>
+            UINT32 gradientStopCount;
+
+            /// <summary>
+            /// X coordinate of the start point of the color line.
+            /// </summary>
+            float x0;
```
oldnewthing, line 2415: I assume the coordinate system is explained in the OpenType spec, so I won't ask what they mean. I assume that anybody who is working at this low level will have the OpenType spec and be following along.

specs/dwritecore/DWriteCore.md
```diff
@@ -1235,3 +2038,916 @@ DWRITE_BEGIN_INTERFACE(IDWriteFactory8, "202D44F0-84AC-4D6F-B8D0-704B27675CAA")
...
+    /// any children.
+    /// </returns>
+    /// <remarks>
+    /// Whether a paint element has children (and how many) can be determined a priori from its paint type and
+    /// properties. For more information, see DWRITE_PAINT_ELEMENT.
+    /// </remarks>
+    STDMETHOD(MoveToFirstChild)(
+        _Out_writes_bytes_(structSize) DWRITE_PAINT_ELEMENT* paintElement,
+        UINT32 structSize = sizeof(DWRITE_PAINT_ELEMENT)
+        ) PURE;
```
oldnewthing, line 2730: I assume that if the `DWRITE_PAINT_ELEMENT` structure is extended in the future, the IDWritePaintReader will have to support different sizes corresponding to different versions of the API.

I also assume that all extensions of the `DWRITE_PAINT_ELEMENT` structure will consist of adding new members to the end, so that downlevel callers can just inspect the part they understand.

(One thing I'm not sure of is whether `IDWritePaintReader` is a DWrite-provided interface or an interface that consumers implement. I think it's DWrite-provided, so it will always support all of the things that DWrite does.)

specs/dwritecore/DWriteCore.md
jonwis: RECOMMEND: Make an overload-ish of DrawGlyphRun, with this including the support level parameter at the end. Long term, this becomes the right API to target.

specs/dwritecore/DWriteCore.md
```diff
@@ -510,9 +510,217 @@ The following other DirectWrite APIs also use the hybrid font selection algorith
...
+Because color font support is implemented at the glyph rendering level, it does not affect text
+layout. This is true whether a client uses the `IDWriteTextLayout` interface or implements its own
+text layout algorithm. The mapping of characters to glyphs and the positioning of those glyphs all
+use monochrome glyph IDs and their associated metrics. The output of the text layout process is a
+sequence of monochrome glyph runs. Color font support can be enabled by translating those monochrome
+"base" glyph runs to color glyphs runs at rendering time.
+
+The easiest way to enable color is to call the `DrawGlyphRunWithColor` method instead of calling
+`DrawGlyphRun`. Direct2D exposes this method through the `ID2D1DeviceContext7` interface.
+DirectWrite exposes this method through the `IDWriteBitmapRenderTarget3` interface.
```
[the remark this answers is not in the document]
jonwis: DISCUSSION: These are not all equivalent - you need to provide palette information in some, and awesomeness levels to others. Fonts are backwards compatible; the awesomeness parameter indicates how to operate on a font that is able to be awesome.

The current shape means you need to pass "use awesome" and then "here is the awesome" ... if you don't want awesome, but you still want to call this method, you still need to pass in a bogus/unused awesome value.

The D2D API does take a color palette parameter, but it might not be used for various reasons (B&W font, image font, font without color hinting.)

RECOMMEND: DrawGlyphRunWithColorSupport - the "with" means it's extra, and not really an overload.

specs/dwritecore/DWriteCore.md
```diff
@@ -510,9 +510,217 @@ The following other DirectWrite APIs also use the hybrid font selection algorith
...
+| `DWRITE_PAINT_TYPE_LINEAR_GRADIENT` | Fill the current clip with the specified gradient.                                       |
+| `DWRITE_PAINT_TYPE_RADIAL_GRADIENT` | Fill the current clip with the specified gradient.                                       |
+| `DWRITE_PAINT_TYPE_SWEEP_GRADIENT`  | Fill the current clip with the specified gradient.                                       |
+| `DWRITE_PAINT_TYPE_GLYPH`           | Fill the specified glyph shape with child paint element.                                 |
+| `DWRITE_PAINT_TYPE_COLOR_GLYPH`     | Render the child paint element.                                                          |
+| `DWRITE_PAINT_TYPE_TRANSFORM`       | Render the child paint element with the specified transform.                             |
+| `DWRITE_PAINT_TYPE_COMPOSITE`       | Render the two child paint elements and compose them using the specified composite mode. |
+
+For more information about paint types, see the
+[OpenType COLR Table](https://docs.microsoft.com/en-us/typography/opentype/spec/colr) specification.
```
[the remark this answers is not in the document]
jonwis: RECOMMEND: Fix

specs/dwritecore/DWriteCore.md
```diff
@@ -1050,9 +1258,604 @@ int __cdecl wmain(int argc, WCHAR** argv)
...
+    if (foregroundBrush == nullptr)
+    {
+        foregroundBrush = m_foregroundBrush;
+    }
+
+    // If available, use the ID2D1DeviceContext7 interface to draw
+    // the glyph run with color the "easy" way.
+    if (auto deviceContext7 = m_deviceContext.try_query<ID2D1DeviceContext7>())
+    {
+        deviceContext7->DrawGlyphRunWithColor(
```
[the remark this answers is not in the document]
jonwis: SAMPLE: Yes

specs/dwritecore/DWriteCore.md
jonwis: SAMPLE: Consider annotating the sample to explain some of the assumptions being made.

specs/dwritecore/DWriteCore.md
jonwis: DISCUSSION: Yes, these fields match the names in the OT spec.

specs/dwritecore/DWriteCore.md
jonwis: DISCUSSION: This field is written to by our implementation up to the size, and the caller has definitions of this struct based on the SDK and version level they are using.

RECOMMEND: Zero-fill the struct up to the provided size before use.

specs/dwritecore/DWriteCore.md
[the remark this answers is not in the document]
niklasb-ms: The sample is actually correct because D2D drawing methods return void. If a drawing operation fails, an error is returned from EndDraw. I'm adding a comment to that effect. (I think this design is because drawing operations are batched, so a failure may not be known until the device context is flushed.)

specs/dwritecore/DWriteCore.md
niklasb-ms: Done in the implementation.

specs/dwritecore/DWriteCore.md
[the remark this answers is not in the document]
niklasb-ms: Fixed.

SeattlePseudoGenius (approved):
Approved - all comments addressed.

niklasb-ms:
Add color font APIs to DWriteCore API spec.

PR #2434, opened 2022-04-21
[Ask mode approval pending] [Cherry-pick] Under a .NET SDK version check, bring back our (broken) default file inclusion solution for RESW and image files (applies to non-single project MSIX apps only).

rohanp-msft:
**Ask mode approval pending**

Cherry-pick of https://github.com/microsoft/WindowsAppSDK/pull/2424

Under a .NET SDK version check, bring back our (broken) default file inclusion solution for RESW and image files (applies to non-single project MSIX apps only).

Reason for change: we don't break apps on upgrade. If they use a sufficiently recent .NET SDK version, they'll get the improvement. Otherwise, they'll get the existing behavior.

**How verified**
In a WinUI with WAP app:
1. If version is less than threshold, do we use the old, WASDK solution? Yes.
2. If version is equal to the threshold, do we use the .NET SDK solution? Yes.
3. If there are duplicate images in Content, do we error? No.
4. If the version is 6.0.2xx, does the comparison still work? Does that evaluate as less than 6.0.300? Yes.

### DIFF
--- a/build/NuSpecs/Microsoft.WindowsAppSDK.Foundation.props
+++ b/build/NuSpecs/Microsoft.WindowsAppSDK.Foundation.props
@@ -8,4 +8,10 @@
     <WindowsAppSdkFoundation>true</WindowsAppSdkFoundation>
   </PropertyGroup>
 
+  <!--
+    Ideally, this import would be done only if EnableCoreMrtTooling is true. That's impossible though because that
+    property cannot be defined beforehand.
+  -->
+  <Import Project="$(MSBuildThisFileDirectory)MrtCore.props" />
+
 </Project>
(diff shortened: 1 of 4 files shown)


huichen123 (approved):

rohanp-msft:
Rejected by shiproom. Closing.

riverar:
> Rejected by shiproom. Closing.

Can you share why?

rohanp-msft:
> > Rejected by shiproom. Closing.
> 
> Can you share why?

Yes! Sorry, I didn't intentionally exclude the reason. The bug was considered P2 and it has a workaround (to be documented as a relnote). Per the bug bar, that doesn't make the cut. It's P2 because it doesn’t block apps from giving us feedback on the preview features.

PR #2440, opened 2022-04-26
Fix Unit Testing bug for PushNotifications

pmpurifoy:
Issue Description:
TEST_METHOD setup should be done in the METHOD_SETUP/CLEANUP functions instead of the CLASS_* functions. Previously, the tests were using the incorrect dll to set the VersionInfo::TestInitialize with the package family name of the test package.

Fix:
Move environment setup to METHOD_SETUP.

Validation:
Ran PushNotificationTests unit tests in TAEF.

### DIFF
--- a/test/PushNotificationTests/BaseTestSuite.cpp
+++ b/test/PushNotificationTests/BaseTestSuite.cpp
@@ -21,19 +21,31 @@ using namespace winrt::Microsoft::Windows::PushNotifications;
 
 void BaseTestSuite::ClassSetup()
 {
-    ::Test::Bootstrap::Setup();
+    ::Test::Bootstrap::SetupPackages();
+}
+
+void BaseTestSuite::ClassCleanup()
+{
+    ::Test::Bootstrap::CleanupPackages();
+}
+
+void BaseTestSuite::MethodSetup()
+{
+    ::Test::Bootstrap::SetupBootstrap();
+
     bool isSelfContained{};
     VERIFY_SUCCEEDED(TestData::TryGetValue(L"SelfContained", isSelfContained));
 
     if (!isSelfContained)
     {
         ::WindowsAppRuntime::VersionInfo::TestInitialize(::Test::Bootstrap::TP::WindowsAppRuntimeFramework::c_PackageFamilyName);
+        VERIFY_IS_FALSE(::WindowsAppRuntime::SelfContained::IsSelfContained());
     }
-}
-
-void BaseTestSuite::ClassCleanup()
-{
-    ::Test::Bootstrap::Cleanup();
+    else
+    {
+        ::WindowsAppRuntime::VersionInfo::TestInitialize(L"I_don't_exist_package!");
+        VERIFY_IS_TRUE(::WindowsAppRuntime::SelfContained::IsSelfContained());
+    }
 }
 
 void BaseTestSuite::MethodCleanup()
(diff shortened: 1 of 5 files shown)


loneursid:
The passing rate for push tests before this fix is: 3/56 (3 passing out of 56 total tests) and with the fix it is 45/56. Nice improvement! Do you think there's anything that could / should be addressed in this PR?

Toasts tests have a passing rate of only 43/83. Do you think it's similar issue that should / could be addressed in this PR?

test/PushNotificationTests/BaseTestSuite.cpp
```diff
@@ -21,19 +21,31 @@ using namespace winrt::Microsoft::Windows::PushNotifications;
...
+}
+
+void BaseTestSuite::ClassCleanup()
+{
+    ::Test::Bootstrap::CleanupPackages();
+}
+
+void BaseTestSuite::MethodSetup()
+{
+    ::Test::Bootstrap::SetupBootstrap();
```
hulumane, line 34: The same concern here as with the other PR. If the configuration is per class and not per method, I would expect a single process to be running all the tests. That would mean that this can be moved to classsetup.

pmpurifoy: Right, but since each TEST_METHOD is being run as a separate process, it has to be done in MethodSetup.

test/PushNotificationTests/BaseTestSuite.cpp
```diff
@@ -21,19 +21,31 @@ using namespace winrt::Microsoft::Windows::PushNotifications;
...
     VERIFY_SUCCEEDED(TestData::TryGetValue(L"SelfContained", isSelfContained));
 
     if (!isSelfContained)
     {
         ::WindowsAppRuntime::VersionInfo::TestInitialize(::Test::Bootstrap::TP::WindowsAppRuntimeFramework::c_PackageFamilyName);
+        VERIFY_IS_FALSE(::WindowsAppRuntime::SelfContained::IsSelfContained());
+    }
+    else
+    {
+        ::WindowsAppRuntime::VersionInfo::TestInitialize(L"I_don't_exist_package!");
```
sharath2727, line 46: Do we have to undo this in the method cleanup?

test/PushNotificationTests/BaseTestSuite.cpp
```diff
@@ -21,19 +21,31 @@ using namespace winrt::Microsoft::Windows::PushNotifications;
...
+        VERIFY_IS_TRUE(::WindowsAppRuntime::SelfContained::IsSelfContained());
```
sharath2727, line 47: Also want to understand if any of the packages like the framework, dynamic dependencies are installed manually on the machine before running the tests and now when you run the tests would it fail or would the taef re-install the required packages?

Is there a sanity check before running the tests?

sharath2727 (changes requested):

hulumane (approved):

test/PushNotificationTests/BaseTestSuite.cpp
pmpurifoy: SetupPackages does this. If packages are installed it will uninstall them.

test/PushNotificationTests/BaseTestSuite.cpp
pmpurifoy: VersionInfo::TestShutdown does this.

sharath2727 (approved):

PR #2444, opened 2022-04-27
Refactoring app notification tests

pmpurifoy:
This PR refactors the AppNotification tests into TAEF. Similar to the PushNotificationTests refactor PR, all the unit tests that use to exist in the ToastNotificationTestApp are moving to TAEF. Now error codes will be seen in the TAEF process.

This runs each test scenario under 4 diffrent envs:
- packaged
- unpackaged
- packaged + SelfContained
- unpackaged + SelfContained


test/AppNotificationTests/BaseTestSuite.cpp
```diff
@@ -0,0 +1,640 @@
...
+    VERIFY_SUCCEEDED(TestData::TryGetValue(L"SelfContained", isSelfContained));
+
+    if (!isSelfContained)
+    {
+        ::WindowsAppRuntime::VersionInfo::TestInitialize(::Test::Bootstrap::TP::WindowsAppRuntimeFramework::c_PackageFamilyName);
+        VERIFY_IS_FALSE(::WindowsAppRuntime::SelfContained::IsSelfContained());
+    }
+    else
+    {
+        ::WindowsAppRuntime::VersionInfo::TestInitialize(L"I_don't_exist_package!");
```
hulumane, line 48: Why initialize this at all with a fake package?

test/AppNotificationTests/BaseTestSuite.cpp
```diff
@@ -0,0 +1,640 @@
...
+}
+
+void BaseTestSuite::ClassCleanup()
+{
+    ::Test::Bootstrap::CleanupPackages();
+}
+
+void BaseTestSuite::MethodSetup()
+{
+    ::Test::Bootstrap::SetupBootstrap();
```
hulumane, line 36: Why is this part of MethodSetup and not ClassSetup? Don't we want to do these test operations only once rather than every time a test runs which seems wasteful?

test/AppNotificationTests/BaseTestSuite.cpp
```diff
@@ -0,0 +1,640 @@
...
+
+void BaseTestSuite::VerifyUnregisterTwice()
+{
+    RegisterWithAppNotificationManager();
+    AppNotificationManager::Default().Unregister();
+    VERIFY_THROWS_HR(AppNotificationManager::Default().Unregister(), E_NOT_SET);
+}
+void BaseTestSuite::VerifyToastSettingEnabled()
+{
+    VERIFY_ARE_EQUAL(AppNotificationManager::Default().Setting(), AppNotificationSetting::Enabled);
```
hulumane, line 103: Can we have a negative test?

test/AppNotificationTests/BaseTestSuite.cpp
pmpurifoy: Here is the new guidelines for testing SelfContained: https://github.com/microsoft/WindowsAppSDK/blob/main/docs/Coding-Guidelines/SelfContained.md#unit-and-functional-testing

test/AppNotificationTests/BaseTestSuite.cpp
pmpurifoy: Since TAEF launches a new process to test each method, that process needs to bootstrap instead of the parent process. This pattern is done by Howard as well.

test/AppNotificationTests/BaseTestSuite.cpp
pmpurifoy: Do you know of a way to programmatically set the AppNotificationSetting? I would love to add a negative case :)

test/AppNotificationTests/BaseTestSuite.cpp
pmpurifoy: But this should definitely be a test constant.

test/AppNotificationTests/BaseTestSuite.cpp
hulumane: Can we do a quick experiment by moving this stuff to ClassSetup and ClassCleanup?

test/AppNotificationTests/PackagedTests.h
```diff
@@ -0,0 +1,86 @@
...
+using namespace WEX::Logging;
+using namespace WEX::TestExecution;
+
+class PackagedTests : BaseTestSuite
+{
+    BEGIN_TEST_CLASS(PackagedTests)
+        TEST_CLASS_PROPERTY(L"ThreadingModel", L"MTA")
+        TEST_CLASS_PROPERTY(L"RunFixtureAs:Class", L"RestrictedUser")
+        TEST_CLASS_PROPERTY(L"IsolationLevel", L"Class")
+        TEST_CLASS_PROPERTY(L"Data:SelfContained", L"{true, false}")
```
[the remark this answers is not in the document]
hulumane: I think we need to run te with either true or false flag set correct? Can you add a comment in the PR somewhere as to how you run the test exactly?

test/AppNotificationTests/BaseTestSuite.cpp
hulumane: If you follow the bread crumbs to product code, I believe there is a way to set group policy via registry.

hulumane: I looked at the code. We definitely need a negative test but setting it is not trivial. We may need a test hook of some sort. LEts opene up a seperate work item for a negative test.

test/AppNotificationTests/PackagedTests.h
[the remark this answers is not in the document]
pmpurifoy: It runs the TEST_CLASS as 'true' then as 'false'. No flags are needed.

test/AppNotificationTests/BaseTestSuite.cpp
pmpurifoy: Tested this and unpackaged tests fail.

test/AppNotificationTests/BaseTestSuite.cpp
danielayala94: Loved the package name 😁.

test/AppNotificationTests/BaseTestSuite.cpp
```diff
@@ -0,0 +1,602 @@
...
+    progressData.Value(0.14);
+    progressData.ValueStringOverride(L"14%");
+    toast.Progress(progressData);
+
+    auto progressDataFromToast{ toast.Progress() };
+    VERIFY_ARE_EQUAL(progressDataFromToast.Status(), L"Status");
+    VERIFY_ARE_EQUAL(progressDataFromToast.Title(), L"Title");
+    VERIFY_ARE_EQUAL(progressDataFromToast.Value(), 0.14);
+    VERIFY_ARE_EQUAL(progressDataFromToast.ValueStringOverride(), L"14%");
+    VERIFY_ARE_EQUAL(progressDataFromToast.SequenceNumber(), (uint32_t) 1);
```
danielayala94, line 151: As a general rule, always use modern C++ casts rather than C-styled.

test/AppNotificationTests/AppNotificationHelpers.h
```diff
@@ -0,0 +1,237 @@
...
+    {
+        winrt::AppNotification toast{ CreateToastNotification() };
+
+        toast.Tag(tag.c_str());
+        toast.Group(group.c_str());
+
+        winrt::AppNotificationManager::Default().Show(toast);
+
+        std::vector<winrt::AppNotification> toastVector{ toast };
+        VERIFY_IS_TRUE(VerifyToastsPosted(toastVector));
```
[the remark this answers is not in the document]
hulumane: Agreed. The logic in this helper(VerifyToastsPosted()) should be able to verify multiple payloads.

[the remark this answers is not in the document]
hulumane: Looks like the implementation of the helper does verify multiple toasts. Am I missing something?

hulumane (approved):

test/AppNotificationTests/AppNotificationHelpers.h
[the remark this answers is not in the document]
pmpurifoy: Yup! It takes any number of toasts and verifies that the toasts returned from GetAllAsync are equal.

test/AppNotificationTests/BaseTestSuite.cpp
hulumane: Curious why you cant pass the type directly here instead of casting.

hulumane: You can append the character u or U at the end of integer constant for unsigned int

hulumane: https://www.geeksforgeeks.org/integer-literal-in-c-cpp-prefixes-suffixes/

danielayala94: Ah true that! Adding suffixes is actually better in this case 😊.

test/AppNotificationTests/AppNotificationHelpers.h
```diff
@@ -0,0 +1,237 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
DrusTheAxe, line 1: Helpers? 🤨

test/AppNotificationTests/AppNotificationHelpers.h
```diff
@@ -0,0 +1,237 @@
...
+
+namespace winrt
+{
+    using namespace winrt::Windows::ApplicationModel::Activation;
+    using namespace winrt::Windows::Foundation;
+    using namespace winrt::Microsoft::Windows::AppNotifications;
+    using namespace winrt::Windows::Foundation;
+}
+
+namespace AppNotificationHelpers {
```
DrusTheAxe, line 14: Helpers? 🤨

Reminder: That's one of my trigger words...

This is just code for use by AppNotification tests? Then why not `namespace AppNotification::Test`?

test/AppNotificationTests/AppNotificationHelpers.h
```diff
@@ -0,0 +1,237 @@
...
+        [&] {
+            retrieveNotificationsAsync.Cancel();
+        });
+
+        VERIFY_ARE_EQUAL(retrieveNotificationsAsync.wait_for(std::chrono::seconds(300)), winrt::AsyncStatus::Completed);
+        scope_exit.release();
+
+        auto notifications{ retrieveNotificationsAsync.get() };
+
+        bool found{ false };
```
DrusTheAxe, line 39: NIT: `bool found{};`

No need for explicit value when it's redundant with the default (primitives don't need `{ false/0/0.0/nullptr/etc }` just use `{}`).

test/AppNotificationTests/AppNotificationHelpers.h
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    void VerifyProgressData(const winrt::AppNotificationProgressData& expected, const winrt::AppNotificationProgressData& actual)
+    {
+        if (!expected && !actual)
+        {
+            return;
+        }
+
+        VERIFY_IS_FALSE(!expected || !actual);
```
DrusTheAxe, line 96: You'll find this exceedingly cryptic and difficult to troubleshoot. Failures will show up as VERIFY_IS_FALSE has an expression value of 'false'. Not very useful.

Plus the trip negative is difficult to follow.

How does this differ from
```
VERIFY_IS_TRUE(!!expected);
VERIFY_IS_TRUE(!!actual);
```

(or even better, `VERIFY_IS_TRUE(expected.isValid())` or whatever, if such a named method exists)

P.S. Worst case if you ever do need to use VERIFY_IS_TRUE/FALSE with an expression, add an explicit message so you can tell the values e.g. `VERIFY_IS_FALSE(!expected || !actual, WEX::Common::String().Format(L"Expected:%s Actual:%s", !!expected ? L"true" : L"false", !!actual ? L"true" : L"false"));`

test/AppNotificationTests/AppNotificationHelpers.h
```diff
@@ -0,0 +1,237 @@
...
+            scopeExitGetAll.release();
+
+            auto actualToastVector{ getAllAsync.GetResults() };
+            if (actualToastVector.Size() == (uint32_t) 0)
+            {
+                result = true;
+                break;
+            }
+
+            Sleep(500);
```
DrusTheAxe, line 220: 1. Magic constant
2. Why 500? Why not 100 or 5000?
3. Sleep is generally unwise as it imposes test overhead, and if used to try to ccoordinate cross-thread or -process behavior is an unreliable mechanism to do so.

test/AppNotificationTests/AppNotificationHelpers.h
```diff
@@ -0,0 +1,237 @@
...
+        return result;
+    }
+
+    bool EnsureNoActiveToasts()
+    {
+        auto removeAllAsync = winrt::AppNotificationManager::Default().RemoveAllAsync();
+        if (removeAllAsync.wait_for(c_timeout) != winrt::Windows::Foundation::AsyncStatus::Completed)
+        {
+            removeAllAsync.Cancel();
+            THROW_HR_MSG(E_FAIL, "Failed to remove all active toasts");
```
DrusTheAxe, line 231: `E_FAIL` is a bit blandly misleading. Should be `E_TIMEOUT` as that accurately conveys the error. Or _maybe_ `E_UNEXPECTED` in this case.

If this times out what's `wait_for()` return? If `AsyncStatus::Error` is that the only reason it could fail? If not, how do you know if it failed due to timeout vs other reasons?

P.S. `E_FAIL` is the worst error code, in addition to the obvious useless-tease ("Yeah, something went wrong, but not gonna tell you what, nyah, nyah") it's in various places across Windows, which can complicate troubleshooting if it ever occurs it can be vague what actually happened or why. Strive to never use/return/throw `E_FAIL`.

test/AppNotificationTests/BaseTestSuite.cpp
pmpurifoy: Coming back to this, it is ok to do one-time setup/cleanup in class functions.

test/AppNotificationTests/AppNotificationHelpers.h
```diff
@@ -0,0 +1,237 @@
...
+    {
+        VerifyToastStatus(expectedToastId, true);
+    }
+
+    void VerifyToastIsInactive(UINT32 expectedToastId)
+    {
+        VerifyToastStatus(expectedToastId, false);
+    }
+
+    winrt::AppNotificationProgressData GetToastProgressData(
```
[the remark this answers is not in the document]
pmpurifoy: We use it in many tests, some examples below:

- VerifyUpdateToastProgressDataUsingValidTagAndEmptyGroup
- VerifyUpdateToastProgressDataUsingEmptyTagAndValidGroup
- VerifyUpdateToastProgressDataUsingEmptyTagAndEmptyGroup

test/AppNotificationTests/AppNotificationHelpers.h
pmpurifoy: Renamed to AppNotifications.Test.h

test/AppNotificationTests/BaseTestSuite.cpp
DrusTheAxe: > Coming back to this, it is ok to do one-time setup/cleanup in class functions.

Sure. That's actually the _typical_ use case. Long as your test methods don't _alter_ what's done in setup.

Dynamic Dependencies and DeploymentManager are the exception cases, as they routinely work "before the runtime is (all or partly) installed". As long as AppNotifications isn't doing any 'runtime install/setup' work and just expects "the runtime is properly setup for use" as a prerequisite then you're golden.

One TAEF property to be aware of is if you set IsolationLevel=Method because you want your test methods each run in a new process. DynDep does this in some cases, 'cause, bootstrapping special weirdness. If you need this behavior then when/how your Setup/Cleanup fixtures run may also be...unusual. But this should all be rare birds.

test/AppNotificationTests/BaseTestSuite.cpp
pmpurifoy: https://github.com/microsoft/WindowsAppSDK/issues/2587

test/AppNotificationTests/AppNotificationHelpers.h
pmpurifoy: @hulumane What do you think we should do here? Currently the tests pass without the sleep but I'm unsure how to go about this.

hulumane: This code uses polling to determine that the ToastHistory ultimately is set to 0 within 2.5 seconds. We don't have a better way to do this because the Clear APIs are a fire and forget operation that do thing asynchronously. So if we call ClearAll, the call to VerifyToastsCleared() may not return true because of timing issues. Hence the polling.

As far as the 500ms number is concerned, we should make it 1000. That way this code polls 5 times in 5 seconds which is a little more understandable.

In the future, we could presumably make this an Async API and rely on a ClearAll event to signal completion of the Async operation. Alas, with our current OS setup, this isn't possible.

Please file a bug @pmpurifoy to have an Async equivalent for all our Toast APIs. We can poll our developers to see if this is something they want.

pmpurifoy: https://github.com/microsoft/WindowsAppSDK/issues/2617

test/AppNotificationTests/AppNotifications.Test.h
```diff
@@ -93,7 +95,8 @@ namespace AppNotificationHelpers {
             return;
         }
 
-        VERIFY_IS_FALSE(!expected || !actual);
+        VERIFY_IS_TRUE(!!expected);
+        VERIFY_IS_TRUE(!!actual);
```
danielayala94, line 99: Nit: `VERIFY_IS_NOT_NULL`, so you clearly state the intention of these verification steps, instead of casting.

test/AppNotificationTests/AppNotifications.Test.h
```diff
@@ -156,13 +159,13 @@ namespace AppNotificationHelpers {
...
 
-            VERIFY_ARE_EQUAL(getAllAsync.wait_for(std::chrono::milliseconds(500)), winrt::Windows::Foundation::AsyncStatus::Completed);
+            VERIFY_ARE_EQUAL(getAllAsync.wait_for(c_timeout), winrt::Windows::Foundation::AsyncStatus::Completed);
             scopeExitGetAll.release();
 
             auto actualToastVector{ getAllAsync.GetResults() };
             if (actualToastVector.Size() == expectedToastVector.size())
             {
-                for (int actualToastIndex{ 0 }; (uint32_t)actualToastIndex < actualToastVector.Size(); actualToastIndex++)
+                for (int actualToastIndex{ 0 }; static_cast<uint32_t>(actualToastIndex) < actualToastVector.Size(); actualToastIndex++)
```
danielayala94, line 168: Nit: I wonder if you could have initialized actualToastIndex as an uint32_t and avoid casting. No need to change that 😊.

But I see the `for` below doesn't have any cast, so why casting at all?

test/AppNotificationTests/AppNotifications.Test.h
```diff
@@ -211,24 +215,25 @@ namespace AppNotificationHelpers {
             scopeExitGetAll.release();
 
             auto actualToastVector{ getAllAsync.GetResults() };
-            if (actualToastVector.Size() == (uint32_t) 0)
+            if (actualToastVector.Size() == static_cast<uint32_t>(0))
```
danielayala94, line 218: Nit: Would `size == 0u` be enough?

danielayala94 (approved):

test/AppNotificationTests/AppNotifications.Test.h
pmpurifoy: I missed this one.

test/AppNotificationTests/AppNotifications.Test.h
pmpurifoy: Since vector::size() is type size_t, we have to use uint32_t. Fixed.

test/AppNotificationTests/AppNotificationHelpers.h
DrusTheAxe: > As far as the 500ms number is concerned, we should make it 1000.

Better to do actual timeout than sleep + N loops

```
const auto c_timeout{ 5s };
const auto c_delay{ 25ms };
auto start{ GetTickCount() };
auto elapsed{ start };
for (;;)
{
    IF do something and happy
        break

    elapsed = GetTickCount() - start;
    IF elapsed > c_timeout
        break
    Sleep(c_delay)
}
IF elapsed > c_timeout
    ERROR
```

Cheap timeout loop to ensure you run for X time with small sleeps while awaiting results

Sleeping for 1s is an awefully long time. Small sleeps give up time for plenty of action to happen on other threads (in this process and others) without burning the cpu in a spinloop. 10ms, 25ms and/or 100ms are usually good numbers for these short sleeps. YMMV

DrusTheAxe (approved):

PR #2447, opened 2022-04-28
Introduce Register API to take in assets

sharath2727:
Description: Allow developers to input DisplayName and Icon of their choice to be registered with AppNotificationManager so that these assets can show up when AppNotifications are posted or received.

Validation: Introduced unittests and validated them.

### DIFF
--- a/dev/AppNotifications/AppNotificationManager.cpp
+++ b/dev/AppNotifications/AppNotificationManager.cpp
@@ -177,15 +177,14 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
     {
         THROW_HR_IF_MSG(E_ILLEGAL_METHOD_CALL, AppModel::Identity::IsPackagedProcess(), "Not applicable for packaged applications");
 
-        THROW_HR_IF(E_INVALIDARG, (displayName != winrt::hstring{}) xor (iconUri != nullptr));
+        THROW_HR_IF(E_INVALIDARG, (displayName == winrt::hstring{}) || (iconUri == nullptr));
 
-        std::wstring iconFilePath{};
-        if (iconUri != nullptr)
-        {
-            iconFilePath = iconUri.Path();
-            iconFilePath = iconFilePath.substr(1, iconFilePath.size() - 1);
-            winrt::check_bool(std::filesystem::exists(std::filesystem::path{ iconFilePath }));
-        }
+        std::wstring iconFilePath{ iconUri.Path() };
+        iconFilePath = iconFilePath.substr(1, iconFilePath.size() - 1);
+        // Uri converts backward slashes to forward slashes
+        std::replace(iconFilePath.begin(), iconFilePath.end(), '/', '\\');
+
+        THROW_HR_IF_MSG(E_INVALIDARG, !IsIconFileExtensionSupported(std::filesystem::path{ iconFilePath }), "Icon not supported");
 
         RegisterHelper(displayName, iconFilePath);
     }
--- a/dev/AppNotifications/AppNotificationManager.cpp
+++ b/dev/AppNotifications/AppNotificationManager.cpp
@@ -21,6 +21,7 @@
 #include <winrt/Windows.Foundation.Collections.h>
 #include <WindowsAppRuntime.SelfContained.h>
 #include <ShellLocalization.h>
+#include <filesystem>
 
 using namespace std::literals;
 
@@ -93,7 +94,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
         return m_activatedEventArgs;
     }
 
-    void AppNotificationManager::Register()
+    void AppNotificationManager::RegisterHelper(hstring const& displayName, std::wstring const& iconFilePath)
     {
         HRESULT hr{ S_OK };
 
@@ -122,7 +123,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
                 {
                     THROW_IF_FAILED(PushNotifications_RegisterFullTrustApplication(m_appId.c_str(), GUID_NULL));
 
-                    storedComActivatorGuid = RegisterComActivatorGuidAndAssets();
+                    storedComActivatorGuid = RegisterComActivatorGuidAndAssets(displayName, iconFilePath);
                 }
 
                 if (!WindowsAppRuntime::SelfContained::IsSelfContained())
@@ -165,6 +166,28 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
             hr = wil::ResultFromCaughtException();
             throw;
         }
+    }
+
+    void AppNotificationManager::Register()
+    {
+        RegisterHelper(winrt::hstring{} /* displayName */, std::wstring{} /* iconFilePath */);
+    }
+
+    void AppNotificationManager::Register(hstring const& displayName, winrt::Uri const& iconUri)
+    {
+        THROW_HR_IF_MSG(E_ILLEGAL_METHOD_CALL, AppModel::Identity::IsPackagedProcess(), "Not applicable for packaged applications");
+
+        THROW_HR_IF(E_INVALIDARG, (displayName != winrt::hstring{}) xor (iconUri != nullptr));
+
+        std::wstring iconFilePath{};
+        if (iconUri != nullptr)
+        {
+            iconFilePath = iconUri.Path();
+            iconFilePath = iconFilePath.substr(1, iconFilePath.size() - 1);
+            winrt::check_bool(std::filesystem::exists(std::filesystem::path{ iconFilePath }));
+        }
+
+        RegisterHelper(displayName, iconFilePath);
     }
 
     // This assumes that the caller has taken an exclusive lock
@@ -307,7 +330,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
                     }
                     else
                     { 
-                        registeredClsid = RegisterComActivatorGuidAndAssets();
+                        registeredClsid = RegisterComActivatorGuidAndAssets(winrt::hstring{} /* displayName */, std::wstring{} /* iconUri */);
                     }
 
                     auto notificationCallback{ winrt::create_instance<INotificationActivationCallback>(registeredClsid, CLSCTX_ALL) };
--- a/dev/AppNotifications/AppNotificationUtility.cpp
+++ b/dev/AppNotifications/AppNotificationUtility.cpp
@@ -241,7 +241,7 @@ std::wstring Microsoft::Windows::AppNotifications::Helpers::GetDisplayNameBasedO
     return displayName;
 }
 
-winrt::guid Microsoft::Windows::AppNotifications::Helpers::RegisterComActivatorGuidAndAssets()
+winrt::guid Microsoft::Windows::AppNotifications::Helpers::RegisterComActivatorGuidAndAssets(winrt::hstring const& displayName, std::wstring const& iconFilePath)
 {
     std::wstring registeredGuid;
     auto hr = GetActivatorGuid(registeredGuid);
@@ -258,20 +258,42 @@ winrt::guid Microsoft::Windows::AppNotifications::Helpers::RegisterComActivatorG
         RegisterComServer(comActivatorGuidString);
 
         registeredGuid = comActivatorGuidString.get();
+        std::wstring notificationAppId{ Microsoft::Windows::AppNotifications::Helpers::RetrieveNotificationAppId() };
+        RegisterAssets(notificationAppId, registeredGuid, displayName, iconFilePath);
     }
     else
     {
         THROW_IF_FAILED(hr);
     }
 
-    std::wstring notificationAppId{ Microsoft::Windows::AppNotifications::Helpers::RetrieveNotificationAppId() };
-    RegisterAssets(notificationAppId, registeredGuid);
-
     // Remove braces around the guid string
     return winrt::guid(registeredGuid.substr(1, registeredGuid.size() - 2));
 }
 
-void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring const& appId, std::wstring const& clsid)
+// Try the following techniques to retrieve display name and icon:
+// 1. Assets provided by the user
+// 2. Based on the best app shortcut, using the FrameworkUdk.
+// 3. From the current process.
+// 4. Set a default DisplayName, but leave empty the icon file path so Shell can set a default icon.
+Microsoft::Windows::AppNotifications::ShellLocalization::AppNotificationAssets Microsoft::Windows::AppNotifications::Helpers::RegisterAssetsHelper(winrt::hstring const& displayName, std::wstring const& iconFilePath)
+{
+    Microsoft::Windows::AppNotifications::ShellLocalization::AppNotificationAssets assets{};
+
+    if ((!displayName.empty()) && (!iconFilePath.empty()))
+    {
+        assets.displayName = displayName.c_str();
+        assets.iconFilePath = iconFilePath.c_str();
+    }
+    else if (FAILED(Microsoft::Windows::AppNotifications::ShellLocalization::RetrieveAssetsFromShortcut(assets)) &&
+        FAILED(Microsoft::Windows::AppNotifications::ShellLocalization::RetrieveAssetsFromProcess(assets)))
+    {
+        assets.displayName = GetDisplayNameBasedOnProcessName();
+    }
+
+    return assets;
+}
+
+void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring const& appId, std::wstring const& clsid, winrt::hstring const& displayName, std::wstring const& iconFilePath)
 {
     wil::unique_hkey hKey;
     // subKey: \Software\Classes\AppUserModelId\{AppGUID}
@@ -288,17 +310,7 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring 
         &hKey,
         nullptr /* lpdwDisposition */));
 
-    // Try the following techniques to retrieve display name and icon:
-    // 1. Based on the best app shortcut, using the FrameworkUdk.
-    // 2. From the current process.
-    // 3. Set a default DisplayName, but leave empty the icon file path so Shell can set a default icon.
-    Microsoft::Windows::AppNotifications::ShellLocalization::AppNotificationAssets assets{};
-
-    if (FAILED(Microsoft::Windows::AppNotifications::ShellLocalization::RetrieveAssetsFromShortcut(assets)) &&
-        FAILED(Microsoft::Windows::AppNotifications::ShellLocalization::RetrieveAssetsFromProcess(assets)))
-    {
-        assets.displayName = GetDisplayNameBasedOnProcessName();
-    }
+    Microsoft::Windows::AppNotifications::ShellLocalization::AppNotificationAssets assets{ RegisterAssetsHelper(displayName, iconFilePath) };
 
     RegisterValue(hKey, L"DisplayName", reinterpret_cast<const BYTE*>(assets.displayName.c_str()), REG_EXPAND_SZ, assets.displayName.size() * sizeof(wchar_t));
 
--- a/dev/AppNotifications/AppNotificationUtility.cpp
+++ b/dev/AppNotifications/AppNotificationUtility.cpp
@@ -275,7 +275,7 @@ winrt::guid Microsoft::Windows::AppNotifications::Helpers::RegisterComActivatorG
 // 2. Based on the best app shortcut, using the FrameworkUdk.
 // 3. From the current process.
 // 4. Set a default DisplayName, but leave empty the icon file path so Shell can set a default icon.
-Microsoft::Windows::AppNotifications::ShellLocalization::AppNotificationAssets Microsoft::Windows::AppNotifications::Helpers::RegisterAssetsHelper(winrt::hstring const& displayName, std::wstring const& iconFilePath)
+Microsoft::Windows::AppNotifications::ShellLocalization::AppNotificationAssets Microsoft::Windows::AppNotifications::Helpers::GetAssetsHelper(winrt::hstring const& displayName, std::wstring const& iconFilePath)
 {
     Microsoft::Windows::AppNotifications::ShellLocalization::AppNotificationAssets assets{};
 
@@ -310,7 +310,7 @@ void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring 
         &hKey,
         nullptr /* lpdwDisposition */));
 
-    Microsoft::Windows::AppNotifications::ShellLocalization::AppNotificationAssets assets{ RegisterAssetsHelper(displayName, iconFilePath) };
+    Microsoft::Windows::AppNotifications::ShellLocalization::AppNotificationAssets assets{ GetAssetsHelper(displayName, iconFilePath) };
 
     RegisterValue(hKey, L"DisplayName", reinterpret_cast<const BYTE*>(assets.displayName.c_str()), REG_EXPAND_SZ, assets.displayName.size() * sizeof(wchar_t));
 
--- a/dev/AppNotifications/ShellLocalization.cpp
+++ b/dev/AppNotifications/ShellLocalization.cpp
@@ -167,9 +167,11 @@ void WriteHIconToPngFile(wil::unique_hicon const& hIcon, _In_ PCWSTR pszFileName
     THROW_IF_FAILED(spStreamOut->Commit(STGC_DANGEROUSLYCOMMITMERELYTODISKCACHE));
 }
 
-bool IsIconFileExtensionSupported(std::filesystem::path const& iconFilePath)
+bool Microsoft::Windows::AppNotifications::ShellLocalization::IsIconFileExtensionSupported(std::filesystem::path const& iconFilePath)
 {
     static PCWSTR c_supportedExtensions[]{ L".bmp", L".ico", L".jpg", L".png" };
+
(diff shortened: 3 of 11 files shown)


dev/Common/Common.vcxitems.filters
```diff
@@ -26,15 +26,15 @@
     <ClInclude Include="$(MSBuildThisFileDirectory)WindowsAppRuntime.SelfContained.h">
```
robertzhou-wpn, line 26: Looks to me nothing actual got changed in this file. can we revert and remove this file from this PR?

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -167,6 +168,28 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
+    void AppNotificationManager::Register()
+    {
+        RegisterHelper(winrt::hstring{} /* displayName */, std::wstring{} /* iconFilePath */);
+    }
+
+    void AppNotificationManager::Register(hstring const& displayName, winrt::Uri const& iconUri)
+    {
+        THROW_HR_IF_MSG(E_ILLEGAL_METHOD_CALL, AppModel::Identity::IsPackagedProcess(), "Not applicable for packaged applications");
+
+        THROW_HR_IF(E_INVALIDARG, (displayName != winrt::hstring{}) xor (iconUri != nullptr));
```
hulumane, line 180: xor? Can we use && or || :D?

hulumane: If any of the arguments are empty, we should throw.

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -167,6 +168,28 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
+
+        std::wstring iconFilePath{};
+        if (iconUri != nullptr)
```
hulumane, line 183: Remove this check please.  Its probably better to crash if this is null below.

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -167,6 +168,28 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
+        {
+            iconFilePath = iconUri.Path();
+            iconFilePath = iconFilePath.substr(1, iconFilePath.size() - 1);
+            winrt::check_bool(std::filesystem::exists(std::filesystem::path{ iconFilePath }));
```
hulumane, line 187: You are not checking if the extension is supported or not here.

hulumane: @danielayala94 You said there was a helper for this?

dev/AppNotifications/AppNotificationManager.cpp
sharath2727: I chose xor here because it serves the purpose:

Below are the two combinations of the displayname and icon the xor operation would allow and not throw:
Scenario 1: If (displayName = winrt::hstring{}  &&  iconUri = nullptr) -  In this scenario the windowsappsdk will get the icon from .exe

Scenario 2: If (displayName != winrt::hstring{} && iconUri != nullptr) - In this scenario we pick the developer provided parameters.

Note: winrt::hstring{} means basically setting the parameter to nullptr.

Thanks.

sharath2727: Now once the xor operation allows the right combinations, only when the iconUri != nullptr we check if path exists.

dev/AppNotifications/AppNotificationManager.cpp
danielayala94: I went all the way to `RegisterAssetsHelper` to verify that the file extension isn't being checked.

@sharath2727 There is an inline helper in `ShellLocalization.cpp` -> `IsIconFileExtensionSupported`. Declare it in the header so you can consume it in `AppNotificationManager`.

dev/AppNotifications/AppNotificationManager.cpp
danielayala94: I thought we wanted this API to be all-in or throw (if any of the params is not what we expect, then throw).

dev/AppNotifications/AppNotificationManager.cpp
danielayala94: It isn't probably need after all. I thought we wanted to have both valid DisplayName and IconUri to complete the asset processing thing.

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -258,20 +258,42 @@ winrt::guid Microsoft::Windows::AppNotifications::Helpers::RegisterComActivatorG
...
     return winrt::guid(registeredGuid.substr(1, registeredGuid.size() - 2));
 }
 
-void Microsoft::Windows::AppNotifications::Helpers::RegisterAssets(std::wstring const& appId, std::wstring const& clsid)
+// Try the following techniques to retrieve display name and icon:
+// 1. Assets provided by the user
+// 2. Based on the best app shortcut, using the FrameworkUdk.
+// 3. From the current process.
+// 4. Set a default DisplayName, but leave empty the icon file path so Shell can set a default icon.
+Microsoft::Windows::AppNotifications::ShellLocalization::AppNotificationAssets Microsoft::Windows::AppNotifications::Helpers::RegisterAssetsHelper(winrt::hstring const& displayName, std::wstring const& iconFilePath)
```
danielayala94, line 278: We aren't registering anything yet. This is more like `GetAssetsHelper()`.

specs/AppNotifications/AppNotifications-spec.md
```diff
@@ -139,6 +139,29 @@ int main()
...
+should provide both the assets or not provide them at all. If assets are not provided WinAppSDK
+will determine appropriate DisplayName and Icon. Icon provided by the developer should
+be valid and reside on the local machine only. For Packaged applications, this API is not
+applicable and will throw an exception.
+
+```cpp
+int main()
+{
+    auto manager = winrt::AppNotificationManager::Default();
+    manager.Register(L"AppNotifications", winrt::Windows::Foundation::Uri {L"<Path to Icon>"});
```
danielayala94, line 155: I think it's okay to set a mock path (`C:\Users\<user>\AppData\MyApp\icon.png` or the app folder which may be under ProgramFiles?).

specs/AppNotifications/AppNotifications-spec.md
```diff
@@ -139,6 +139,29 @@ int main()
 }
 ```
 
+## Registering for App Notifications using assets
+
+For Unpackaged applications, the developer calls into Register API with DisplayName and Icon.
```
danielayala94, line 144: `the developer can opt to call into`

The dev isn't forced to call into the new API (it is highly recommended though).

specs/AppNotifications/AppNotifications-spec.md
```diff
@@ -139,6 +139,29 @@ int main()
...
+WinAppSDK will register and display them when an App Notification is received. The developer
```
danielayala94, line 145: nit: Windows App SDK

Also, some rewording will help to express the idea here (which I think is it registering the app and then display a notification?).

test/TestApps/ToastNotificationsTestApp/main.cpp
```diff
@@ -550,10 +550,10 @@ bool VerifyShowToast_Unpackaged()
 	// Registration happens in main()
     auto toastNotificationManager = winrt::AppNotificationManager::Default();
 
-    auto scope_exit = wil::scope_exit(
+    /*auto scope_exit = wil::scope_exit(
```
danielayala94, line 553: What's the reason about commenting?

dev/AppNotifications/AppNotificationManager.cpp
sharath2727: Hm...I thought we should allow the Register(nullptr, nullptr); - I think what you are all suggesting makes sense. They could fall back to using the Register API with no arguments. It simplifies the design further.

Thanks.

specs/AppNotifications/AppNotifications-spec.md
sharath2727: Would like to see what everyone thinks of this before updating. Thanks.

specs/AppNotifications/AppNotifications-spec.md
```diff
@@ -139,6 +139,29 @@ int main()
 }
 ```
 
+## Registering for App Notifications using assets
+
+For Unpackaged applications, the developer can opt to call into Register API with DisplayName and Icon.
```
brendanfly, line 144: "can Register using a custom DisplayName and Icon"?

specs/AppNotifications/AppNotifications-spec.md
```diff
@@ -139,6 +139,29 @@ int main()
...
+WinAppSDK will register the application and display these assets when an App Notification is received.
+The developer should provide both the assets or not provide them at all. If assets are not provided WinAppSDK
+will determine appropriate DisplayName and Icon. Icon provided by the developer should be a valid supported
+format like png, bmp, ico or jpg and should reside on the local machine only. For Packaged applications,
```
brendanfly, line 148: This seems very vague. We should probably support a handful of formats and being very explicit in what those are, throwing an error when not valid.

test/TestApps/ToastNotificationsTestApp/main.cpp
```diff
@@ -1322,6 +1322,171 @@ bool VerifyIconPathExists_Unpackaged()
     return true;
 }
 
+bool VerifyRegisterWithNullDisplayNameFail_Unpackaged()
+{
+    // Register is already called in main with an explicit appusermodelId
+    winrt::AppNotificationManager::Default().UnregisterAll();
+    try
+    {
+        winrt::AppNotificationManager::Default().Register(winrt::hstring{}, winrt::Uri{ LR"(C:\Windows\System32\WindowsSecurityIcon.png)" });
```
brendanfly, line 1331: Is this equivalent to L""? What's that behavior?

test/TestApps/ToastNotificationsTestApp/main.cpp
```diff
@@ -1322,6 +1322,171 @@ bool VerifyIconPathExists_Unpackaged()
...
+    return false;
+}
+
+bool VerifyRegisterWithEmptyDisplayNameAndEmptyIconFail_Unpackaged()
+{
+    // Register is already called in main with an explicit appusermodelId
+    winrt::AppNotificationManager::Default().UnregisterAll();
+    try
+    {
+        winrt::AppNotificationManager::Default().Register(L"", winrt::Uri{ L"" });
```
brendanfly, line 1460: Nevermind, this answers my earlier question :)

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -167,6 +168,27 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
         }
     }
 
+    void AppNotificationManager::Register()
+    {
+        RegisterHelper(winrt::hstring{} /* displayName */, std::wstring{} /* iconFilePath */);
```
brendanfly, line 173: This is OK with empty values?

sharath2727: What this basically means is that when both parameters are empty we fallback to gleaning the icon off process or shortcuts. If everything fails we fallback to default.

Thanks.

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -177,15 +177,14 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
-        std::wstring iconFilePath{};
-        if (iconUri != nullptr)
-        {
-            iconFilePath = iconUri.Path();
-            iconFilePath = iconFilePath.substr(1, iconFilePath.size() - 1);
-            winrt::check_bool(std::filesystem::exists(std::filesystem::path{ iconFilePath }));
-        }
+        std::wstring iconFilePath{ iconUri.Path() };
+        iconFilePath = iconFilePath.substr(1, iconFilePath.size() - 1);
+        // Uri converts backward slashes to forward slashes
```
danielayala94, line 184: Please be more specific on commenting the intention of replacing. Windows understands file paths with backslashes, while URI follows forward slashes.

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -177,15 +177,14 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
+        std::replace(iconFilePath.begin(), iconFilePath.end(), '/', '\\');
+
+        THROW_HR_IF_MSG(E_INVALIDARG, !IsIconFileExtensionSupported(std::filesystem::path{ iconFilePath }), "Icon not supported");
```
danielayala94, line 187: Nit: `Icon file extension not supported.`

dev/AppNotifications/ShellLocalization.cpp
```diff
@@ -167,10 +167,12 @@ void WriteHIconToPngFile(wil::unique_hicon const& hIcon, _In_ PCWSTR pszFileName
     THROW_IF_FAILED(spStreamOut->Commit(STGC_DANGEROUSLYCOMMITMERELYTODISKCACHE));
 }
 
-bool IsIconFileExtensionSupported(std::filesystem::path const& iconFilePath)
+bool Microsoft::Windows::AppNotifications::ShellLocalization::IsIconFileExtensionSupported(std::filesystem::path const& iconFilePath)
 {
     static PCWSTR c_supportedExtensions[]{ L".bmp", L".ico", L".jpg", L".png" };
 
+    winrt::check_bool(std::filesystem::exists(iconFilePath));
```
danielayala94, line 174: There is a slight issue with doing this here.

With this approach, we return false if the file doesn't exist. However, the caller doesn't know if this is because of the icon being not found or if the file extension is unsupported.

danielayala94: I would keep both operations in separate places (maybe do the exists check in the caller instead?).

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -177,15 +177,14 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
     {
         THROW_HR_IF_MSG(E_ILLEGAL_METHOD_CALL, AppModel::Identity::IsPackagedProcess(), "Not applicable for packaged applications");
 
-        THROW_HR_IF(E_INVALIDARG, (displayName != winrt::hstring{}) xor (iconUri != nullptr));
+        THROW_HR_IF(E_INVALIDARG, (displayName == winrt::hstring{}) || (iconUri == nullptr));
```
danielayala94, line 180: Do we want to return a message to the developer if this throws?

danielayala94: Also, can the iconUri be not null but empty?

sharath2727: Windows::Foundation::Uri doesn't allow L"" to be taken in as an input, it throws incorrect parameter error.

specs/AppNotifications/AppNotifications-spec.md
loneursid: I like more specific over too general, as long as there aren't too many options and that we list the most common options. Maybe we use :\Users\<user>\AppData\MyApp\icon.png as proposed by Daniel and we also put a comment with an example path to a folder inside the program file folder. 

Maybe we should show the functions to call in order to retrieve the path(s) to the folders where the app is installed.

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -93,7 +94,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
         return m_activatedEventArgs;
     }
 
-    void AppNotificationManager::Register()
+    void AppNotificationManager::RegisterHelper(hstring const& displayName, std::wstring const& iconFilePath)
```
hulumane, line 97: Why is one argument hstring and the other wstring? We should be consistent

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -167,6 +168,27 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
+    void AppNotificationManager::Register()
+    {
+        RegisterHelper(winrt::hstring{} /* displayName */, std::wstring{} /* iconFilePath */);
+    }
+
+    void AppNotificationManager::Register(hstring const& displayName, winrt::Uri const& iconUri)
+    {
+        THROW_HR_IF_MSG(E_ILLEGAL_METHOD_CALL, AppModel::Identity::IsPackagedProcess(), "Not applicable for packaged applications");
+
+        THROW_HR_IF(E_INVALIDARG, (displayName == winrt::hstring{}) || (iconUri == nullptr));
```
hulumane, line 180: Same issue with iconUri. If it is null, don't check. Let it crash further along. You should instead check for the validity of the uri's content: eg. empty path, invalid path etc.

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -167,6 +168,27 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
+
+        std::wstring iconFilePath{ iconUri.Path() };
+        iconFilePath = iconFilePath.substr(1, iconFilePath.size() - 1);
```
hulumane, line 183: what is this substring for? Can you please explain this?

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -167,6 +168,27 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
+        // Uri converts backward slashes to forward slashes
+        std::replace(iconFilePath.begin(), iconFilePath.end(), '/', '\\');
```
hulumane, line 185: You are actually converting forward slash to back slash. So the comment is incorrect. https://newbedev.com/difference-between-forward-slash-and-backslash-in-file-path

hulumane, line 185: This conversion is extremely fragile. We should not have to do it. Have you checked why we were erroring out with forward slash in the path? We should confirm why we see this behavior before making this change.

dev/AppNotifications/AppNotificationManager.cpp
hulumane: Atleast make sure that the uri is a valid file path and the file exists on disk.

dev/AppNotifications/AppNotificationManager.cpp
hulumane, line 173: Why don't you retrieve the correct displayname and icon path here before calling into the helper? Passing in empty is a strange pattern.

hulumane: Also Register would fail now because you call empty. Please don't use anti patterns and use valid arguments to pass into 
APIs

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -307,7 +329,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
                     }
                     else
                     { 
-                        registeredClsid = RegisterComActivatorGuidAndAssets();
+                        registeredClsid = RegisterComActivatorGuidAndAssets(winrt::hstring{} /* displayName */, std::wstring{} /* iconUri */);
```
hulumane, line 332: This is incorrect too. Don't pass empty arguments please. Instead pass in the overloaded arguments or the one from shortcut/process

hulumane: Actually why do we need to re-register assets at all? Aren't we simply re-routing to another process? @pmpurifoy ? Can you explain this one please and why we need to re-register assets as part of this flow?

dev/AppNotifications/AppNotificationManager.cpp
sharath2727: use RawUri()

dev/AppNotifications/AppNotificationManager.cpp
sharath2727: Call GetActivatorGuid

test/TestApps/ToastNotificationsTestApp/main.cpp
```diff
@@ -1322,6 +1322,171 @@ bool VerifyIconPathExists_Unpackaged()
...
+
+bool VerifyRegisterWithNullIconFail_Unpackaged()
+{
+    // Register is already called in main with an explicit appusermodelId
+    winrt::AppNotificationManager::Default().UnregisterAll();
+    try
+    {
+        winrt::AppNotificationManager::Default().Register(L"AppNotificationApp", nullptr);
+
+        winrt::AppNotificationManager::Default().UnregisterAll();
```
loneursid, line 1352: A better pattern in this situation, would be to call  UnregisterAll() in a scope_exit. It would
1) make it instantly clear that UnregisterAll isn't under test here.
2) though it doesn't apply to this specific case as Unregister can't throw E_INVALIDARG, avoid a situation where the fuction under test does not throw as expected but the cleanup one does. The test would then pass although it should be failing.

test/TestApps/ToastNotificationsTestApp/main.cpp
```diff
@@ -1322,6 +1322,171 @@ bool VerifyIconPathExists_Unpackaged()
...
+    catch (...)
+    {
+        return winrt::to_hresult() == E_INVALIDARG;
+    }
+
+    return true;
+}
+
+bool VerifyRegisterWithDisplayNameAndIcon_Unpackaged()
+{
```
loneursid, line 1382: It's a shame that the only validation we do is that the function didn't throw. I wish we could verify that calling register actually did something useful and valid, i.e.: besides running code that didn't crash / throw...

test/ToastNotificationTests/APITests.cpp
```diff
@@ -583,5 +583,50 @@ namespace Test::ToastNotifications
...
+            RunTestUnpackaged(L"VerifyRegisterWithEmptyIconFail_Unpackaged", testWaitTime());
+        }
+
+        TEST_METHOD(VerifyRegisterWithEmptyDisplayNameAndEmptyIconFail_Unpackaged)
+        {
+            RunTestUnpackaged(L"VerifyRegisterWithEmptyDisplayNameAndEmptyIconFail_Unpackaged", testWaitTime());
+        }
+
+        TEST_METHOD(VerifyRegisterWithAssetsFail)
+        {
```
loneursid, line 628: I assume that we run a single test for packaged just to confirm that register works for packaged, and that we rely on the more complete suite of unpackaged tests to actually verify Register()'s actual behavior.

It would be nice if, we could just run the suite twice: once for unpackaged and one for packaged. It seems the only difference is whether RunTestUnpackaged or RunTest is used to launch the test. It feels like dependency injection could help here. 

Not suggesting you do it now, that's probably beyond the point of this PR but that could be a nice future refactor.

DrusTheAxe: It's possible to detect at compile time if you're built with WindowsAppSDKSelfContained=true or false. At least, I think it is now in Foundation with recent changes I made to .props/.targets

Our test infrastructure does tend to build everything both ways though it was only running SelfContained=false. My PR in the Aggregator fixes that (going in soon).

Sorry limited network access at the moment. If you grip for SelfContained across build\* you should find the C++ and C# symbols code can check for.
Though I suggest not messing with that until my changes go in as that has significant updates to the plumbing

dev/AppNotifications/ShellLocalization.cpp
```diff
@@ -258,10 +260,7 @@ HRESULT Microsoft::Windows::AppNotifications::ShellLocalization::DeleteIconFromC
     std::path iconFilePath{ RetrieveLocalFolderPath() / (notificationAppId + c_pngExtension) };
 
     // If DeleteFile returned FALSE, then deletion failed and we should return the corresponding error code.
-    if (DeleteFile(iconFilePath.c_str()) == FALSE)
-    {
-        THROW_HR(HRESULT_FROM_WIN32(GetLastError()));
-    }
+    winrt::check_bool(DeleteFile(iconFilePath.c_str()) != FALSE);
```
DrusTheAxe, line 263: NIT: `DeleteFileW` is preferable to `DeleteFile` (ensures no ambiguity to the reader)

Is `winrt::check_bool()` equivalent `wil::RETURN_IF_WIN32_BOOL_FALSE()`?

NOTE: It's weird bordering on perverse to explicitly check for == or != for values of `TRUE` or `FALSE`

dev/AppNotifications/AppNotificationManager.cpp
DrusTheAxe, line 97: Helper? 🤨

specs/AppNotifications/AppNotifications-spec.md
jonwis: Something like:

```c++
std::filesystem::path exePath{ GetPathToCurrentExe() };
auto iconPath{ exePath / "displayicon.png" };
```

jonwis: Sample should show use of MRT or another resource loader indirection to fetch that display string, too.

specs/AppNotifications/AppNotifications-spec.md
```diff
@@ -451,6 +474,9 @@ namespace Microsoft.Windows.AppNotifications
         // For Unpackaged apps, the caller process will be registered as the COM server. And assets like displayname and icon will be gleaned from Shell and registered as well.
         void Register();
 
+        // For Unpackaged apps only, the caller process will be registered as the COM server.
+        void Register(String displayName, Windows.Foundation.Uri iconUri);
```
jonwis, line 478: What happens when a Packaged app calls this? Is that an error?

jonwis (commented):
See note about enhancing the samples.

PR #2452, opened 2022-04-30
UndockedRegFreeWinRT auto-initialization

DrusTheAxe:
Enable UndockedRegFreeWinRT automagic initialization, when using WinAppSDK/SelfContained.

NOTE: This is similar to Bootstrapper automagic initialization, when an unpackaged app is using WinAppSDK/MSIX.

You can disable (opt-out) this auto-initialization by defining the project property `<WindowsAppSdkUndockedRegFreeWinRTInitialize>false</WindowsAppSdkUndockedRegFreeWinRTInitialize>`.

If undefined (the default) this is enabled with `<WindowsAppSDKSelfContained>` is `true`

https://task.ms/38121572

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -26,7 +26,7 @@
   </PropertyGroup>
 
   <Target Name="DirectoryBuildPropsInfo">
-    <Message Condition="$(WindowsAppSDKBuildPipeline) == '1'" Importance="High" Text="Directory.Build.props detects WindowsAppSDKBuildPipeline=$(WindowsAppSDKBuildPipeline)"/>
+    <Message Condition="'$(WindowsAppSDKBuildPipeline)' == '1'" Importance="High" Text="Directory.Build.props detects WindowsAppSDKBuildPipeline=$(WindowsAppSDKBuildPipeline)"/>
 
     <Error Condition="!Exists('$(RepoTestCertificatePFX)')" Text="$(RepoTestCertificatePFX) not found. Run '$(RepoRoot)\tools\DevCheck.cmd' to generate the test certificate." />
     <Error Condition="!Exists('$(RepoTestCertificatePWD)')" Text="$(RepoTestCertificatePWD) not found. Run '$(RepoRoot)\tools\DevCheck.cmd' to generate the test certificate." />
@@ -65,4 +65,12 @@
     <RepoTestCertificatePassword>$([System.IO.File]::ReadAllText('$(RepoTestCertificatePWD)').TrimEnd())</RepoTestCertificatePassword>
     <RepoTestCertificatePasswordRedacted>...redacted...</RepoTestCertificatePasswordRedacted>
   </PropertyGroup>
+
+  <!-- 'Clean As We Go' to prevent build agents from running out of disk space -->
+  <PropertyGroup Condition="('$(WindowsAppSDKCleanIntermediateFiles)' == '') and ('$(WindowsAppSDKBuildPipeline)' == '1')">
+    <WindowsAppSDKCleanIntermediateFiles>true</WindowsAppSDKCleanIntermediateFiles>
+  </PropertyGroup>
+  <Target Name="CleanIntermediateFiles" AfterTargets="Build" Condition="'$(WindowsAppSDKCleanIntermediateFiles)' == 'true'">
+    <RemoveDir Directories="$(IntermediateOutputPath)" />
+  </Target>
 </Project>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -66,10 +66,7 @@
     <RepoTestCertificatePasswordRedacted>...redacted...</RepoTestCertificatePasswordRedacted>
   </PropertyGroup>
 
-  <!-- 'Clean As We Go' to prevent build agents from running out of disk space -->
-  <PropertyGroup Condition="('$(WindowsAppSDKCleanIntermediateFiles)' == '') and ('$(WindowsAppSDKBuildPipeline)' == '1')">
-    <WindowsAppSDKCleanIntermediateFiles>true</WindowsAppSDKCleanIntermediateFiles>
-  </PropertyGroup>
+  <!-- 'Clean As We Go' if necessary (specified). Prevents build agents from running out of disk space -->
   <Target Name="CleanIntermediateFiles" AfterTargets="Build" Condition="'$(WindowsAppSDKCleanIntermediateFiles)' == 'true'">
     <RemoveDir Directories="$(IntermediateOutputPath)" />
   </Target>
--- a/dev/UndockedRegFreeWinRT/UndockedRegFreeWinRT-AutoInitializer.cpp
+++ b/dev/UndockedRegFreeWinRT/UndockedRegFreeWinRT-AutoInitializer.cpp
@@ -1,0 +1,25 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#include <Windows.h>
+
+// Do nothing. Just exist.
+//
+// This ensures the including PE file has an import reference
+// to the WindowsAppSDK runtime DLL and thus gets loaded when
+// the including PE file gets loaded.
+
+STDAPI UndockedRegFreeWinRT_EnsureIsLoaded();
+
+// NOTE: Disable optimizations to ensure this unreferenced symbol
+//       doesn't get 'optimized away'. We need the end compiled binary
+//       to import this symbol from the Windows App SDK runtime DLL.
+#pragma optimize("", off)
+namespace Microsoft::Windows::Foundation::UndockedRegFreeWinRT
+{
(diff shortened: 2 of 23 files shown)


dev/UndockedRegFreeWinRT/UndockedRegFreeWinRT-AutoInitializer.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    internal static class NativeMethods
+    {
+        [DllImport("Microsoft.WindowsAppRuntime.dll", CharSet = CharSet.Unicode, ExactSpelling = true)]
+        internal static extern int UndockedRegFreeWinRT_EnsureIsLoaded();
+    }
+
+    class AutoInitialize
+    {
+        [global::System.Runtime.CompilerServices.ModuleInitializer]
```
riverar, line 16: Warning, this module initializer will execute when folks load the assembly in for reflection and possibly other scenarios. We previously discussed this in another PR  https://github.com/microsoft/WindowsAppSDK/pull/1582#issuecomment-942987325; believe one mitigation was to add a `if (Assembly.GetEntryAssembly() != Assembly.GetExecutingAssembly())` guard.

dev/UndockedRegFreeWinRT/UndockedRegFreeWinRT-AutoInitializer.cs
```diff
@@ -0,0 +1,26 @@
...
+        internal static void AccessWindowsAppSDK()
+        {
+            int hr = NativeMethods.UndockedRegFreeWinRT_EnsureIsLoaded();
+            if (hr < 0)
+            {
+                global::System.Environment.Exit(hr);
```
riverar, line 22: I believe in failure cases, this may result in an app appearing to _do nothing_ and confuse users. It may also cause a (temporary) startup loop for packaged apps. (I haven't verified that, but I recall failures during packaged app startup result in retry logic.)

riverar: Perhaps this should throw so WER gets a whack at it? And you get event logging, telemetry, etc.?

dev/UndockedRegFreeWinRT/UndockedRegFreeWinRT-AutoInitializer.cs
DrusTheAxe: That's interesting. I hadn't seen the earlier comment.

@Scottj1s if you concur I'll do this here _and_ add it to `MddBootstrapAutoInitializer.cs`

dev/UndockedRegFreeWinRT/UndockedRegFreeWinRT-AutoInitializer.cs
DrusTheAxe: This is only used when an app has been built using WinAppSDK/SelfContained. The only thing this really does is `LoadLibrary(Microsoft.WindowsAppRuntime.dll)`. The only way that could fail is if the application's missing that file, or something it imports (FrameworkUDK.dll - also SelfContained and placed with the app - and some OS DLLs (kernel32.dll etc)). So while it's somewhat a moral equivalent to MddBootstrapAutoInitializer.cs it's a ton simpler.

I considered baking out a rich palette of options like `MddBootstrapInitialize2(...options)` but that seemed way overkill. There's no lookup, no registered packages, and no real logic (`UndockedRegFreeWinRT_EnsureIsLoaded()` is just `{return S_OK}`). Given this is only in play when an app's using WinAppSDK/SelfContained failures here seem exceedingly rare.

I could change this to just call the function, ignore the return code and use the throwing form of P/Invoke. If the DLL can't load it would throw a .NET exception about a missing native DLL? Been a while since I saw that. Would that give a clearer symptom of the problem?

DrusTheAxe: If the DllImport's target DLL isn't found or otherwise won't load .NET throws a `DllNotFoundException`. This is similar to the C++ experience where an imported DLL can't be resolved when called loads.

`UndockedRegFreeWinRT_EnsureIsLoaded()` does nothing (`{return S_OK}`) so there's no need for any error check. The error check is if .NET can resolve the DLL, which the DllImport handles. Streamlined the code to match.

dev/UndockedRegFreeWinRT/UndockedRegFreeWinRT-AutoInitializer.cs
DrusTheAxe: Added as a proactive change. If problematic I can change or revert.

DrusTheAxe:
Build error due to out of disk space on the build server. Ping'd folks for follow up.

Kicking off another build in case it's a transient problem. Fingers crossed...

dhoehna:
Seems odd to use `true1`<WindowsAppSdkUndockedRegFreeWinRTInitialize>true</WindowsAppSdkUndockedRegFreeWinRTInitialize>` to opt-out of initialization.  This looks like an opt-in to initialization.

dhoehna:
Is there a sample on how to set up Undocked Reg Free Winrt from a selfcontained package?

DrusTheAxe:
> Is there a sample on how to set up Undocked Reg Free Winrt from a selfcontained package?

Yes, sorta. But you won't like it...

Overtly? You can LoadLibraryEx("Microsoft.WindowsAppRuntime.dll", 0, LOAD_WITH_ALTERED_SEARCH_ORDER) and handle errors. Or link importing an export from it (C++) or add a module initializer (C#) like the AutoInit code here does

Or get this PR checked in and then you get auto-init by default if you built with WindowsAppSDKSelfContained=true

DrusTheAxe:
> Seems odd to use `true1`true` to opt-out of initialization. This looks like an opt-in to initialization.

Typo in the description. Fixed. Thansk

dhoehna:
> WindowsAppSDKSelfContained=true

That is what I was searching for.  As long as "set `WindowsAppSDKSelfContained=true`" is somewhere in a sample to help developers get this in their package, I'm all good.

Directory.Build.props
```diff
@@ -65,4 +65,12 @@
     <RepoTestCertificatePassword>$([System.IO.File]::ReadAllText('$(RepoTestCertificatePWD)').TrimEnd())</RepoTestCertificatePassword>
     <RepoTestCertificatePasswordRedacted>...redacted...</RepoTestCertificatePasswordRedacted>
   </PropertyGroup>
+
+  <!-- 'Clean As We Go' to prevent build agents from running out of disk space -->
+  <PropertyGroup Condition="('$(WindowsAppSDKCleanIntermediateFiles)' == '') and ('$(WindowsAppSDKBuildPipeline)' == '1')">
```
Scottj1s, line 70: I would remove the WindowsAppSDKBuildPipeline check (prevents local testing) and have the pipeline set WindowsAppSDKCleanIntermediateFiles

Scottj1s (approved):

DrusTheAxe:
> > WindowsAppSDKSelfContained=true
> 
> That is what I was searching for. As long as "set `WindowsAppSDKSelfContained=true`" is somewhere in a sample to help developers get this in their package, I'm all good.

Nope. More like "DO NOTHING to get this in their build". You have to do something special only if you DON'T want this option.

This mirrors the Bootstrapper's auto-initialization. On by default, only have to do something if you want to opt-out

dev/UndockedRegFreeWinRT/UndockedRegFreeWinRT-AutoInitializer.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+STDAPI UndockedRegFreeWinRT_EnsureIsLoaded();
+
+// NOTE: Disable optimizations to ensure this unreferenced symbol
+//       doesn't get 'optimized away'. We need the end compiled binary
+//       to import this symbol from the Windows App SDK runtime DLL.
+#pragma optimize("", off)
+namespace Microsoft::Windows::Foundation::UndockedRegFreeWinRT
+{
+static void EnsureIsLoaded()
```
sachintaMSFT, line 20: NIT: add indentation to namespace content.

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrapAutoInitializer.cs
```diff
@@ -16,13 +16,22 @@
...
+using System.Runtime.InteropServices;
+
 namespace Microsoft.Windows.ApplicationModel.DynamicDependency.BootstrapCS
 {
     class AutoInitialize
     {
         [global::System.Runtime.CompilerServices.ModuleInitializer]
         internal static void AccessWindowsAppSDK()
         {
+            // Do nothing if we're being loaded for reflection (rather than execcution)
```
sachintaMSFT, line 29: NIT: spelling - execution

sachintaMSFT (approved):

PR #2477, opened 2022-05-05
Support COM activation from elevated processes

pmpurifoy:
Task:
Enable elevated apps to CoCreateInstance the PushNotificationsLongRunningProcess by having single source of Proxy/Stub dll in the framework package MSIX instead of the LRP MSIX.

Status:
Failed to find proxy registration for IID: {60FC21B2-B396-4D49-94F0-7555869FB93C}. This IID is supposed to be associated with the IWpnLrpPlatform but the client cannot find it as of now. After install the Microsoft.WindowsAppRuntime.Framework (test package), I cannot find either of the COM interfaces in the registry.


test/DynamicDependency/data/Microsoft.WindowsAppRuntime.Framework/appxmanifest.xml
```diff
@@ -99,5 +99,17 @@
             <ActivatableClass ActivatableClassId="Microsoft.Windows.Security.AccessControl.SecurityDescriptorHelpers" ThreadingModel="both" />
         </InProcessServer>
     </Extension>
+    <Extension Category="windows.activatableClass.proxyStub">
+        <ProxyStub ClassId="60FC21B2-B396-4D49-94F0-7555869FB93C">
+          <Path>PushNotificationsLongRunningTask.ProxyStub.dll</Path>
```
hulumane, line 104: Who generates this proxystub? Shouldn't it be the singleton that generates this?

pmpurifoy: The ProxyStub dll is generated by the dev/PushNotifications/PushNotificationsLongRunningTask.ProxyStub project. We are moving the dll to the framework package so the client/server can pull the ProxyStub from single dependency.

test/DynamicDependency/data/Microsoft.WindowsAppRuntime.Framework/appxmanifest.xml
```diff
@@ -99,5 +99,17 @@
             <ActivatableClass ActivatableClassId="Microsoft.Windows.Security.AccessControl.SecurityDescriptorHelpers" ThreadingModel="both" />
         </InProcessServer>
     </Extension>
+    <Extension Category="windows.activatableClass.proxyStub">
+        <ProxyStub ClassId="60FC21B2-B396-4D49-94F0-7555869FB93C">
```
brialmsft, line 103: The proxy/stub CLSIDs for this seem incorrect; did you test that this actually works for both interfaces?

The general rule for proxy/stub CLSIDs is that a given set of IDLs built together will share a single proxy/stub CLSID (1 generated dlldata.c file = 1 proxy/stub CLSID). As a _default_, the proxy/stub CLSID for a dlldata.c file is the IID of the first interface in it. That's why you'll often see registrations where the proxy/stub CLSID matches an IID, which tricks you into making the mistake you've made here, and assuming that they _always_ match.
In addition to being a bit confusing, this default is also unstable; an otherwise reasonable change to the code (e.g. adding a new IDL file to the directory, or adding a new interface to the existing file and putting it above the topmost existing interface for style reasons) could shift which interface is listed first in dlldata.c, resulting in the proxy/stub CLSID changing and breaking all these registrations.

My recommendation is to always explicitly set a proxy/stub CLSID when building a proxy/stub DLL. There are 2 pretty typical patterns:

- Generate a completely new proxy/stub CLSID GUID, and use a C define like "PROXY_CLSID_IS={0x1A7F9F39,0xFB98,0x4405,{0xA6,0x6D,0x95,0x4E,0x07,0x71,0x4E,0x2A}}" to set it. The main downside is that you need to use struct syntax for the GUID, which looks a little gross. The upside is that it's completely self-contained.
- Reference an existing GUID symbol using a C define like "PROXY_CLSID=IID_IWpnForegroundSink". This symbol has to be something that can be resolved when linking the proxy/stub DLL, since it's external from dlldata.c's perspective. You could either use an existing IID (as in my example), or generate a new GUID just for your proxy/stub CLSID and use that.

Whichever approach you go with, your DLL should end up with a single predictable proxy/stub CLSID. Then on the registration side, you'll have a single windows.activatableClass.proxyStub extension, with a single ProxyStub element (with a now-predictable ClassId value), and multiple interfaces underneath it.

brialmsft:
It's not touched in this PR, but I'd generally recommend getting rid of the entire "library" block in your IDL file.

libraries are mostly intended for generating typelibs, which you don't make use of. You're following a bad pattern where you use a library block just to generate a CLSID for NotificationsLongRunningPlatform. Much cleaner to replace that whole library block with 1 line:
declare_guid(CLSID_NotificationsLongRunningPlatform, PUSHNOTIFICATIONS_IMPL_CLSID_UUID)

brialmsft (changes requested):

test/DynamicDependency/data/Microsoft.WindowsAppRuntime.Framework/appxmanifest.xml
brialmsft: > Whichever approach you go with, your DLL should end up with a single predictable proxy/stub CLSID.
To expand on this a bit (not relevant to your immediate scenario): you _can_ have a DLL that supports multiple proxy/stub CLSIDs, but only in more complicated scenarios where you're building IDLs in multiple locations. Making this work requires defining different ENTRY_PREFIXes to allow you to combine multiple DllGetClassObject definitions from multiple generated dlldata.c files, and then defining the DLL's _real_ DllGetClassObject manually so that it queries each of the generated prefixed DllGetClassObject functions.

Again, none of this applies here, because this is the simple case: all the IDLs in your proxy/stub DLL are built in a single location. I'm just noting that solutions for the more complicated scenario exist, for future reference.

dev/PushNotifications/PushNotificationUtility.h
```diff
@@ -129,7 +129,9 @@ namespace winrt::Microsoft::Windows::PushNotifications::Helpers
 
     inline wil::com_ptr<INotificationsLongRunningPlatform> GetNotificationPlatform()
     {
-        return wil::CoCreateInstance<NotificationsLongRunningPlatform, INotificationsLongRunningPlatform>(CLSCTX_LOCAL_SERVER);
+        wil::com_ptr<INotificationsLongRunningPlatform> platform;
+        THROW_IF_FAILED(CoCreateInstance(CLSID_NotificationsLongRunningPlatform, nullptr, (CLSCTX_LOCAL_SERVER | CLSCTX_ALLOW_LOWER_TRUST_REGISTRATION), IID_PPV_ARGS(&platform)));
```
brialmsft, line 133: If you prefer WIL, there is a wil::CoCreateInstance that takes CLSID as a normal parameter (rather than the "CLSID carried by the UUID attribute of a template parameter" form you previously used); you don't need to switch to the raw API.

dev/PushNotifications/PushNotificationsLongRunningTask.ProxyStub/NotificationsLongRunningProcess.idl
```diff
@@ -38,13 +38,4 @@ interface INotificationsLongRunningPlatform : IUnknown
     HRESULT RemoveToastRegistrationMapping([in] LPCWSTR processName);
 };
 
-[uuid(PUSHNOTIFICATIONS_LIBID_UUID)]
```
brialmsft, line 41: Should be able to delete definition of PUSHNOTIFICATIONS_LIBID_UUID too?

pmpurifoy:
This does not work for non-elevated process. It was pulling the proxyStub from an already registered package.

dev/PushNotifications/PushNotificationUtility.h
brialmsft, line 133: Warning: CoCreateInstance rejects unknown CLSCTX flags, and CLSCTX_ALLOW_LOWER_TRUST_REGISTRATION will not be available on all machines that support WinAppSdk (the feature is serviced to all RS5+ releases, but you can't always assume that the machine you're running on is fully patched).

Be sure to consider whether the failure mode of passing this flag on an unpatched system is what you want, or if you need to do runtime probing for whether CLSCTX_ALLOW_LOWER_TRUST_REGISTRATION is supported to get correct behavior on those systems.

PR #2493, opened 2022-05-11
UndockedRegFreeWinRT self-contained and auto-initialization support

DrusTheAxe:
Some refinements on the UndockedRegFreeWinRT self-contained and auto-initialization support

Testing here is a matter of modifying the nuget used by IntegrationTests in the Aggregator and then 'porting'/copying those changes from the local nuget to the actual source here. The Aggregator update is blocked until this makes it into the build

A couple of bits are less than elegant but work and can be refined later, when not so blocking/urgent.

### DIFF
--- a/build/NuSpecs/Microsoft.WindowsAppSDK.UndockedRegFreeWinRTCommon.targets
+++ b/build/NuSpecs/Microsoft.WindowsAppSDK.UndockedRegFreeWinRTCommon.targets
@@ -7,4 +7,8 @@
         <WindowsAppSdkUndockedRegFreeWinRTInitialize>true</WindowsAppSdkUndockedRegFreeWinRTInitialize>
     </PropertyGroup>
 
+    <PropertyGroup Condition="'$(WindowsAppSdkUndockedRegFreeWinRTInitializeLoadLibrary)'=='' and '$(WindowsAppSdkUndockedRegFreeWinRTInitialize)'=='true'">
+        <WindowsAppSdkUndockedRegFreeWinRTInitializeLoadLibrary>true</WindowsAppSdkUndockedRegFreeWinRTInitializeLoadLibrary>
+    </PropertyGroup>
+
 </Project>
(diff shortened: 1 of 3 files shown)


build/NuSpecs/Microsoft.WindowsAppSDK.UndockedRegFreeWinRTCommon.targets
```diff
@@ -7,4 +7,8 @@
         <WindowsAppSdkUndockedRegFreeWinRTInitialize>true</WindowsAppSdkUndockedRegFreeWinRTInitialize>
     </PropertyGroup>
 
+    <PropertyGroup Condition="'$(WindowsAppSdkUndockedRegFreeWinRTInitializeLoadLibrary)'=='' and '$(WindowsAppSdkUndockedRegFreeWinRTInitialize)'=='true'">
```
MikeHillberg, line 10: Isn't this always false (can't be both == '' and == 'true'

MikeHillberg (approved):

build/NuSpecs/Microsoft.WindowsAppSDK.UndockedRegFreeWinRTCommon.targets
sachintaMSFT: No. Those two checks are against two different variables - WindowsAppSdkUndockedRegFreeWinRTInitializeLoadLibrary &  WindowsAppSdkUndockedRegFreeWinRTInitialize. The two conditions can be true at the same time.

sachintaMSFT (approved):

build/NuSpecs/Microsoft.WindowsAppSDK.UndockedRegFreeWinRTCommon.targets
DrusTheAxe: I'd like to get rid of `WindowsAppSdkUndockedRegFreeWinRTInitializeLoadLibrary` but trying to statically link in microsoft.windowsappruntime.dll was giving unresolved external symbol for the ..._EnsureIsLoaded export, despite the lib in the link list. So I punted to unblock us and went with a ctor calling LoadLibrary(), and that's used by default. Kept the other logic around so we can experiment later to see what's griefing.

You shouldn't need to do anything. Like the Bootstrapper auto-init, do nothing and if selfcontained you get the URFW auto-init. And unlike the Bootstrapper the only reason that could fail is if your app's self-contained install is horrifically damaged microsoft.windowsappruntime.dll and imports can't be loaded. Pretty hard trick, and if you're in that state WinAppSDK's probably the least of your worries.

PR #2498, opened 2022-05-11
UrfwInitialize() didn't report if ExtRoLoadCatalog() returned error

DrusTheAxe:
UrfwInitialize() didn't report if ExtRoLoadCatalog() errored and returned a failing HRESULT (only trapped thrown exceptions, but the implementation does both)

### DIFF
--- a/dev/UndockedRegFreeWinRT/urfw.cpp
+++ b/dev/UndockedRegFreeWinRT/urfw.cpp
@@ -413,7 +413,7 @@ HRESULT UrfwInitialize() noexcept
     DetourAttach(&(PVOID&)TrueRoResolveNamespace, RoResolveNamespaceDetour);
     try
     {
-        ExtRoLoadCatalog();
+        RETURN_IF_FAILED(ExtRoLoadCatalog());
     }
     catch (...)
     {


kythant (approved):

dev/UndockedRegFreeWinRT/urfw.cpp
```diff
@@ -413,7 +413,7 @@ HRESULT UrfwInitialize() noexcept
     DetourAttach(&(PVOID&)TrueRoResolveNamespace, RoResolveNamespaceDetour);
     try
     {
-        ExtRoLoadCatalog();
+        RETURN_IF_FAILED(ExtRoLoadCatalog());
```
BenJKuhn, line 416: A little odd to mix return_if_failed and try/catch. Any reason for not doing throw_if_failed?

BenJKuhn (approved):

dev/UndockedRegFreeWinRT/urfw.cpp
DrusTheAxe: Notice the catch block. THROW_IF_FAILED to turn around and catch it and return would be machinery to no end. If it were more complex sure, but if wishes were fishes... Probably could just make the function try/CATCH_RETURN() and do away with the try + catch{} in the first place

Most of URFW is error-code based. A few places use std::wstring with all that implies, plus std::make_unique<> i.e. throwing-new, possibly a handful other like exceptions (pardon the pun). But as a rule error-handling rules this codebase.

PR #2503, opened 2022-05-12
Notification Mitigations for 1.1-stable

pmpurifoy:
Issue:
- In elevated scenarios, PushNotifications do not work for both unpackaged + packaged applications and the IsSupported API doesn't include this limitation.
- In elevated scenarios, AppNotification activations do not work for unpackaged applications.

Fix:
- Added IsElevated check to the PushNotification IsSupported API.
- Register COM server and AppID entries in HKLM in the registry

Verified:
Changes were tested by manually running scenarios of unpackaged/packaged + non-elevated/elevated and confirming correct behavior.


pmpurifoy:
Moving toast mitigation to 1.2

PR #2505, opened 2022-05-12
Notification Mitigation - AppNotificationManager::IsSupported

pmpurifoy:
Issue:
In elevated scenarios, PushNotifications do not work for both unpackaged + packaged applications and the IsSupported API doesn't include this limitation. Also for AppNotifications, there is no IsSupported API. Elevated scenarios are currently not supported for AppNotifications so we are added an IsElevated check to this newly added IsSupported API.

Fix:
Added IsElevated check to the PushNotification/AppNotification IsSupported API.

Verified:
Changes were tested by manually running scenarios of unpackaged/packaged + non-elevated/elevated and confirming correct behavior of AppNotification APIs.

### DIFF
--- a/dev/PushNotifications/PushNotificationManager.cpp
+++ b/dev/PushNotifications/PushNotificationManager.cpp
@@ -130,7 +130,7 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
     bool PushNotificationManager::IsSupported()
     {
         // Only scenarios that use the Background Infrastructure component of the OS support Push in the SelfContained case
-        static bool isSupported{ !WindowsAppRuntime::SelfContained::IsSelfContained() || PushNotificationHelpers::IsPackagedAppScenario() };
+        static bool isSupported{ (!WindowsAppRuntime::SelfContained::IsSelfContained() || PushNotificationHelpers::IsPackagedAppScenario()) && !PushNotificationHelpers::IsElevated() };
         return isSupported;
 
     }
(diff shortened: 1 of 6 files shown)


dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -130,7 +130,7 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
     bool PushNotificationManager::IsSupported()
     {
         // Only scenarios that use the Background Infrastructure component of the OS support Push in the SelfContained case
```
hulumane, line 132: Update the comment to:
Packaged App scenario always works
[4:25 PM] Pavan Hullumane
Unpackaged App scenario does not

[4:25 PM] Pavan Hullumane
Unpackaged does not when it is self-contained or it is elevated

dev/PushNotifications/PushNotificationUtility.h
```diff
@@ -143,6 +143,16 @@ namespace winrt::Microsoft::Windows::PushNotifications::Helpers
...
     }
 
+    inline bool IsElevated()
+    {
+        wil::unique_handle processToken;
+        THROW_IF_WIN32_BOOL_FALSE(OpenProcessToken(GetCurrentProcess(), MAXIMUM_ALLOWED, &processToken));
+
+        auto tokenLabel{ wil::get_token_information<TOKEN_MANDATORY_LABEL>(processToken.get()) };
+        DWORD subAuthority = static_cast<DWORD>(*GetSidSubAuthorityCount(tokenLabel->Label.Sid) - 1);
+        return *GetSidSubAuthority(tokenLabel->Label.Sid, subAuthority) >= SECURITY_MANDATORY_HIGH_RID;
```
hulumane, line 153: For elevated, can you add a comment that this equates to HIGH_RID? SYSTEM is not applicable right for elevated?

pmpurifoy: This just blocks all elevated + system callers. I can just use == instead. If that is preferred.

PR #2507, opened 2022-05-12
Notification Mitigation - AppNotificationManager::IsSupported

pmpurifoy:
Branch was based on stable 1.1 branch which led to unnecessary files being merged to main. Previous PR was closed here: https://github.com/microsoft/WindowsAppSDK/pull/2505

Issue:
In elevated scenarios, PushNotifications do not work for both unpackaged + packaged applications and the IsSupported API doesn't include this limitation. Also for AppNotifications, there is no IsSupported API. Elevated scenarios are currently not supported for AppNotifications so we are added an IsElevated check to this newly added IsSupported API.

Fix:
Added IsElevated check to the Push/AppNotification IsSupported API.

Verified:
Changes were tested by manually running scenarios of unpackaged/packaged + non-elevated/elevated and confirming correct behavior of Push/AppNotification APIs.

### DIFF
--- a/dev/AppNotifications/AppNotificationManager.cpp
+++ b/dev/AppNotifications/AppNotificationManager.cpp
@@ -95,7 +95,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
 
     bool AppNotificationManager::IsSupported()
     {
-        static bool isSupported{ !PushNotificationHelpers::IsElevated() };
+        static bool isSupported{ !Security::IntegrityLevel::IsElevated() };
         return isSupported;
     }
 
@@ -447,11 +447,6 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
 
     winrt::Windows::Foundation::IAsyncOperation<winrt::Microsoft::Windows::AppNotifications::AppNotificationProgressResult> AppNotificationManager::UpdateAsync(winrt::Microsoft::Windows::AppNotifications::AppNotificationProgressData const data, hstring const tag)
     {
-        if (!IsSupported())
-        {
-            co_return winrt::AppNotificationProgressResult::Unsupported;
-        }
-
         co_return co_await UpdateAsync(data, tag, L"");
     }
 
@@ -629,7 +624,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
     {
         if (!IsSupported())
         {
-            co_return nullptr;
+            co_return {};
         }
 
         HRESULT hr{ S_OK };
--- a/dev/AppNotifications/AppNotificationManager.cpp
+++ b/dev/AppNotifications/AppNotificationManager.cpp
@@ -93,8 +93,19 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
         return m_activatedEventArgs;
     }
 
+    bool AppNotificationManager::IsSupported()
+    {
+        static bool isSupported{ !PushNotificationHelpers::IsElevated() };
+        return isSupported;
+    }
+
     void AppNotificationManager::Register()
     {
+        if (!IsSupported())
+        {
+            return;
+        }
+
         HRESULT hr{ S_OK };
 
         auto logTelemetry{ wil::scope_exit([&]() {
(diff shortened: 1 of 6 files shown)


dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -93,8 +93,19 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
         return m_activatedEventArgs;
     }
 
+    bool AppNotificationManager::IsSupported()
+    {
+        static bool isSupported{ !PushNotificationHelpers::IsElevated() };
```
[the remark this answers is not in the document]
pmpurifoy: IsSupported in PushNotification vs AppNotifications have similar functions. They both block cloud notifications from their component, but AppNotificationManager allows local toasts which is why it looks a little different/less restrictive than the PushNotificationManager.

dev/PushNotifications/PushNotificationUtility.h
```diff
@@ -138,6 +138,18 @@ namespace winrt::Microsoft::Windows::PushNotifications::Helpers
...
+    inline bool IsElevated()
+    {
+        wil::unique_handle processToken;
+        THROW_IF_WIN32_BOOL_FALSE(OpenProcessToken(GetCurrentProcess(), MAXIMUM_ALLOWED, &processToken));
+
+        auto tokenLabel{ wil::get_token_information<TOKEN_MANDATORY_LABEL>(processToken.get()) };
+        DWORD subAuthority = static_cast<DWORD>(*GetSidSubAuthorityCount(tokenLabel->Label.Sid) - 1);
+
+        // HIGH_RID means the process is elevated
+        return *GetSidSubAuthority(tokenLabel->Label.Sid, subAuthority) == SECURITY_MANDATORY_HIGH_RID;
```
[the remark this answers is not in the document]
pmpurifoy: Nope, that one is much simpler. Replacing.

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -406,11 +447,21 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
 
     winrt::Windows::Foundation::IAsyncOperation<winrt::Microsoft::Windows::AppNotifications::AppNotificationProgressResult> AppNotificationManager::UpdateAsync(winrt::Microsoft::Windows::AppNotifications::AppNotificationProgressData const data, hstring const tag)
     {
+        if (!IsSupported())
```
hulumane, line 450: Is this needed here? I believe the API call below already gates. So ....

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -551,6 +627,11 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
 
     winrt::Windows::Foundation::IAsyncOperation<winrt::Windows::Foundation::Collections::IVector<winrt::Microsoft::Windows::AppNotifications::AppNotification>> AppNotificationManager::GetAllAsync()
     {
+        if (!IsSupported())
+        {
+            co_return nullptr;
```
hulumane, line 632: MAy want to return an empty vecotr here. Not null

dev/AppNotifications/AppNotificationManager.cpp
pmpurifoy: This check was just to be thorough, but it is redundant. Will fix next iteration.

dev/PushNotifications/PushNotificationUtility.h
```diff
@@ -138,6 +138,15 @@ namespace winrt::Microsoft::Windows::PushNotifications::Helpers
         return hasSetTaskEntrypoint;
     }
 
+    inline bool IsElevated()
```
hulumane, line 141: Was this copied over from somewhere else in the SDK? If so, might make sense to move this to the common package.

test/TestApps/ToastNotificationsDemoApp/main.cpp
```diff
@@ -208,9 +209,6 @@ int main()
         std::wcout << std::endl;
     }
 
-    std::wcout << L"Requesting PushNotificationChannel...\n\n";
```
hulumane, line 211: why is this here? Just an orphan we forgot to delete? :(

dev/PushNotifications/PushNotificationUtility.h
pmpurifoy: Did not know Security::IntegrityLevel::IsElevated was provided by Security.IntegrityLevel.h. Switching to that helper API instead.

hulumane (approved):

BenJKuhn (approved):

PR #2545, opened 2022-05-23
Added 4+5 to preview/stable process

DrusTheAxe:


docs/Coding-Guidelines/GitCheckinToMainPolicy.md
```diff
@@ -19,6 +19,8 @@ The following requirements must be met to checkin non-Experimental changes to ma
 1. API Review approval
 2. Tests
 3. Prepared to accept bug reports from the world
+4. Functionality is Feature Complete by a release's fork for Preview1
```
MikeHillberg, line 22: I don't think this is the right place for this note. Code after code complete can go into main for the next release once the  final release branch is set up.

BenJKuhn: Hmm, I think we've been vague about changes coming in after preview 1 / feature complete that are scoped to experimental-only. The wording should probably account for the policy, and the policy needs to be settled: after feature complete do we pause experimental-only changes, or continue to take them? Either way, the spirit of the statement is right, but there's nuance that's not captured.

DrusTheAxe: @BenJKuhn This is for non-Experimental checkins. See just above line 15. Line 27+ are the checkin rules for Experimental.

DrusTheAxe: > Code after code complete can go into main for the next release once the final release branch is set up.

Good point. I added 4b. about checkins for vNext after vCurrent stable is snapped/forked.

DrusTheAxe:
@MikeHillberg reassigning to you as this is more up your alley than mine now

PR #2548, opened 2022-05-24
Bug 39711964: WinAppSDK: DeploymentManager doesn't handle ERROR_INSUFFICIENT_BUFFER

DrusTheAxe:
DeploymentManager.Intiialize() calls InitializeProcThreadAttributeList() to determine the size of the buffer needed but threw an exception if if failed - which it always will because it returns FALSE and `GetLastError() == ERROR_INSUFFICIENT_BUFFER`

### DIFF
--- a/dev/Deployment/DeploymentManager.cpp
+++ b/dev/Deployment/DeploymentManager.cpp
@@ -300,7 +300,7 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
     CATCH_RETURN()
 
 
-    std::wstring DeploymentManager::GenerateDeploymentAgentPath() 
+    std::wstring DeploymentManager::GenerateDeploymentAgentPath()
     {
         // Calculate the path to the restart agent as being in the same directory as the current module.
         wil::unique_hmodule module;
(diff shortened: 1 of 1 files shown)


BenJKuhn (approved):

sachintaMSFT (approved):

jonwis:
How was this update tested?

dev/Deployment/DeploymentManager.cpp
```diff
@@ -327,7 +327,11 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
             attributeCount++;
         }
 
-        THROW_IF_WIN32_BOOL_FALSE(InitializeProcThreadAttributeList(nullptr, attributeCount, 0, &attributeListSize));
+        // attributeCount is always >0 so we need to allocate a buffer. Call InitializeProcThreadAttributeList()
+        // to determine the size needed so we always expect ERROR_INSUFFICIENT_BUFFER.
+        THROW_HR_IF(E_UNEXPECTED, !!InitializeProcThreadAttributeList(nullptr, attributeCount, 0, &attributeListSize));
+        const auto lastError{ GetLastError() };
+        THROW_HR_IF(HRESULT_FROM_WIN32(lastError), lastError != ERROR_INSUFFICIENT_BUFFER);
         PPROC_THREAD_ATTRIBUTE_LIST attributeList = reinterpret_cast<PPROC_THREAD_ATTRIBUTE_LIST>(new BYTE[attributeListSize]);
```
BenJKuhn, line 335: As Alex mentioned earlier, it looks like this is never released. It's a small, fixed, one time leak, so lower priority, but still good to either address or track.

DrusTheAxe:
> How was this update tested?

Partially via local testing with the Initialize_NoCapabilities test under the debugger. The test is wired to run as a Universal app and needs non-trivial changes to run as a DesktopBridge/MediumIL process so debugger introspection plus forcing it to take the breakaway codepath gave partial verification. Next step is cooking a build to use in an E2E test for final verification

PR #2551, opened 2022-05-25
Fixed DeploymentManager.Initialize()

DrusTheAxe:
Bad sprintf (hint: ..."% s", bool == mala fortuna)

CreateProcess ran the DeploymentAgent and waited to retrieve the exit code...but then didn't use the exit code. Fixed to propogate that as expected

### DIFF
--- a/dev/Deployment/DeploymentManager.cpp
+++ b/dev/Deployment/DeploymentManager.cpp
@@ -357,6 +357,7 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
 
         DWORD processExitCode{};
         THROW_IF_WIN32_BOOL_FALSE_MSG(GetExitCodeProcess(processInfo.hProcess, &processExitCode), "CmdLine: %ls, processExitCode: %lu", cmdLine.get(), processExitCode);
+        RETURN_IF_FAILED_MSG(processExitCode, "DeploymentAgent exitcode:0x%X", processExitCode);
         return S_OK;
     }
     CATCH_RETURN()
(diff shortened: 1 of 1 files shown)


BenJKuhn (approved):

cwruss (approved):

dev/Deployment/DeploymentManager.cpp
```diff
@@ -356,8 +356,8 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
         wil::handle_wait(processInfo.hProcess);
 
         DWORD processExitCode{};
-        THROW_IF_WIN32_BOOL_FALSE_MSG(GetExitCodeProcess(processInfo.hProcess, &processExitCode), "CmdLine: %ls, processExitCode: %lu", cmdLine.get(), processExitCode);
-        RETURN_IF_FAILED_MSG(processExitCode, "DeploymentAgent exitcode:0x%X", processExitCode);
+        THROW_IF_WIN32_BOOL_FALSE_MSG(GetExitCodeProcess(processInfo.hProcess, &processExitCode), "CmdLine: %ls, processExitCode: %u", cmdLine.get(), processExitCode);
```
BenJKuhn, line 359: This checks the boolean result of the call but then provides the process exit code that couldn't be retrieved on failure. On failure, it would be more helpful to have a message that provides the result of GetLastError.

DrusTheAxe: Line 359 reports if there's an error retrieving the exit code. If so the ...IF_WIN32_BOOL_FALSE... part tells WIL to report the GetLastError()

If we can _get_ the exit code line 360 notes if the child process' exit code is not-success
Line 361 all is well so we return S_OK

PR #2554, opened 2022-05-26
Deployment agent break away process creation fixes

sachintaMSFT:
Main problem is that CreateProcess is called with PROCESS_CREATION_DESKTOP_APP_BREAKAWAY_OVERRIDE instead of PROCESS_CREATION_DESKTOP_APP_BREAKAWAY_ENABLE_PROCESS_TREE which have opposite affects. The latter is what we need to create a breakaway process outside desktop app runtime environment.

Along with it, updated CreateProcess to start DeploymentAgent.exe in running mode and not in suspend mode (+ xfer foreground rights + resume it) as that's not required.

And telemetry is also reporting failure incorrectly even in case of success due to GetActivity method returning a copy of tracelogger activity object which is short lived instead of a reference to it -this is also fixed.

### DIFF
--- a/dev/Deployment/DeploymentManager.cpp
+++ b/dev/Deployment/DeploymentManager.cpp
@@ -345,10 +345,6 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
         wil::unique_process_information processInfo;
         THROW_IF_WIN32_BOOL_FALSE(CreateProcess(nullptr, cmdLine.get(), nullptr, nullptr, TRUE, EXTENDED_STARTUPINFO_PRESENT, nullptr, nullptr, &info.StartupInfo, &processInfo));
 
-        // Transfer foreground rights to the new process before resuming it.
-        AllowSetForegroundWindow(processInfo.dwProcessId);
-        ResumeThread(processInfo.hThread);
-
         // This API is designed to only return to the caller on failure, otherwise block until process termination.
         // Wait for the agent to exit.  If the agent succeeds, it will terminate this process.  If the agent fails,
         // it can exit or crash.  This API will be able to detect the failure and return.
--- a/dev/Deployment/DeploymentManager.cpp
+++ b/dev/Deployment/DeploymentManager.cpp
@@ -343,7 +343,7 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
         info.lpAttributeList = attributeList;
 
         wil::unique_process_information processInfo;
-        THROW_IF_WIN32_BOOL_FALSE(CreateProcess(nullptr, cmdLine.get(), nullptr, nullptr, TRUE, EXTENDED_STARTUPINFO_PRESENT, nullptr, nullptr, &info.StartupInfo, &processInfo));
+        THROW_IF_WIN32_BOOL_FALSE(CreateProcess(nullptr, cmdLine.get(), nullptr, nullptr, FALSE, EXTENDED_STARTUPINFO_PRESENT, nullptr, nullptr, &info.StartupInfo, &processInfo));
 
         // This API is designed to only return to the caller on failure, otherwise block until process termination.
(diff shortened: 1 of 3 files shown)


dev/Deployment/DeploymentManager.cpp
```diff
@@ -343,8 +343,7 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
         info.lpAttributeList = attributeList;
 
         wil::unique_process_information processInfo;
-        THROW_IF_WIN32_BOOL_FALSE(CreateProcess(nullptr, cmdLine.get(), nullptr, nullptr, TRUE, CREATE_SUSPENDED | EXTENDED_STARTUPINFO_PRESENT, nullptr, nullptr,
-                                                &info.StartupInfo, &processInfo));
+        THROW_IF_WIN32_BOOL_FALSE(CreateProcess(nullptr, cmdLine.get(), nullptr, nullptr, TRUE, EXTENDED_STARTUPINFO_PRESENT, nullptr, nullptr, &info.StartupInfo, &processInfo));
 
         // Transfer foreground rights to the new process before resuming it.
         AllowSetForegroundWindow(processInfo.dwProcessId);
```
jonwis, line 349: Is this `ResumeThread` necessary if the process is not being created suspended anymore?

jonwis (changes requested):

dev/Deployment/DeploymentManager.cpp
```diff
@@ -332,23 +332,17 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
         THROW_IF_WIN32_BOOL_FALSE(InitializeProcThreadAttributeList(attributeList, attributeCount, 0, &attributeListSize));
         auto freeAttributeList{ wil::scope_exit([&] { DeleteProcThreadAttributeList(attributeList); }) };
 
-        // Desktop Bridge applications by default have their child processes break away from the parent process.
-        // In order to recreate the calling process' environment correctly, we must prevent child breakaway semantics
-        // when calling the agent. Additionally the agent must do the same when restarting the caller.
-        DWORD policy{ PROCESS_CREATION_DESKTOP_APP_BREAKAWAY_OVERRIDE };
+        // The process being created will create any child processes outside of the desktop app runtime environment.
+        // This behavior is the default for processes for which no policy has been set.
+        DWORD policy{ PROCESS_CREATION_DESKTOP_APP_BREAKAWAY_ENABLE_PROCESS_TREE };
```
DrusTheAxe, line 337: Please add to the comment the link to the docs https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-updateprocthreadattribute

>PROCESS_CREATION_DESKTOP_APP_BREAKAWAY_OVERRIDE 0x04
>The process being created will run inside the desktop app runtime environment. This policy applies only to the process being created, not its descendants..

>PROCESS_CREATION_DESKTOP_APP_BREAKAWAY_ENABLE_PROCESS_TREE 0x01
>The process being created will create any child processes outside of the desktop app runtime environment. 

DeploymentAgent is derived from RestartAgent so I'm surprised this is a problem but perhaps RestartAgent has subtly different needs. RestartAgent uses the Override option but that's clearly not what we want here. The ProcessTree option gets us out of the packaged context as we want. The naming's confusing (Override doesn't, but ProcessTree does? Gah!)

DeploymentAgent's process token has the WIN://SYSAPPID + WIN://PKG access claim with the Override option but not with the ProcessTree option?

DrusTheAxe (approved):

dev/Deployment/DeploymentManager.cpp
DrusTheAxe: The `ResumeThread` (and `AllowSetForegroundWindow`) is gone as it's unnecessary. See latest commit

dev/Deployment/DeploymentManager.cpp
sachintaMSFT: Added the docs link.

cwruss (approved):

PR #2557, opened 2022-05-27
Refactor register API

sharath2727:
Description: Refactoring the existing implementation to be able to do specialized operation depending on the Application type. This makes it easy to even introduce overloaded Register API that takes in assets to directly call into appropriate helpers to assist in registration process.

Validation: Run TAEF tests

### DIFF
--- a/dev/AppNotifications/AppNotificationManager.cpp
+++ b/dev/AppNotifications/AppNotificationManager.cpp
@@ -140,8 +140,8 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
             // Create event handle before COM Registration otherwise if a notification arrives will lead to race condition
             m_waitHandleForArgs.create();
 
-            // Register AppNotificationManager Object for Shell to activate
-            RegisterActivationCallback(registeredClsid);
+            // Register the AppNotificationManager as a COM server for Shell to Activate and Invoke
+            RegisterComServer(registeredClsid);
         }
         catch (...)
         {
@@ -150,7 +150,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
         }
     }
 
-    void AppNotificationManager::RegisterActivationCallback(winrt::guid const& registeredClsid)
+    void AppNotificationManager::RegisterComServer(winrt::guid const& registeredClsid)
     {
         auto lock{ m_lock.lock_exclusive() };
         THROW_HR_IF_MSG(E_INVALIDARG, m_notificationComActivatorRegistration, "Already Registered for App Notifications!");
@@ -170,7 +170,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
             &m_notificationComActivatorRegistration));
     }
 
-    void AppNotificationManager::RegisterAppNotificationSinkWithNotificationPlatform()
+    void AppNotificationManager::RegisterAppNotificationSinkWithLongRunningPlatform()
     {
         auto notificationPlatform{ PushNotificationHelpers::GetNotificationPlatform() };
         THROW_IF_FAILED(notificationPlatform->AddToastRegistrationMapping(m_processName.c_str(), m_appId.c_str()));
@@ -182,7 +182,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
 
         if (!PushNotificationHelpers::IsPackagedAppScenario() && !WindowsAppRuntime::SelfContained::IsSelfContained())
         {
-            RegisterAppNotificationSinkWithNotificationPlatform();
+            RegisterAppNotificationSinkWithLongRunningPlatform();
         }
 
         return registeredClsid;
@@ -192,14 +192,14 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
     {
         THROW_IF_FAILED(PushNotifications_RegisterFullTrustApplication(m_appId.c_str(), GUID_NULL));
 
-        std::wstring comActivatorGuidString{ GetOrCreateComActivatorGuid() };
         std::wstring notificationAppId{ RetrieveNotificationAppId() };
+        std::wstring comActivatorGuidString{ GetOrCreateComActivatorGuid() };
 
         RegisterAssets(notificationAppId, comActivatorGuidString, assets);
 
         if (!WindowsAppRuntime::SelfContained::IsSelfContained())
         {
-            RegisterAppNotificationSinkWithNotificationPlatform();
+            RegisterAppNotificationSinkWithLongRunningPlatform();
         }
 
         // Remove braces around the guid string
@@ -369,6 +369,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
                         std::wstring registeredClsidString;
                         THROW_IF_FAILED(GetActivatorGuid(registeredClsidString));
 
+                        // Remove braces around the guid string
                         registeredClsid = winrt::guid(registeredClsidString.substr(1, registeredClsidString.size() - 2));
                     }
 
--- a/dev/AppNotifications/AppNotificationManager.cpp
+++ b/dev/AppNotifications/AppNotificationManager.cpp
@@ -24,7 +24,7 @@
 
 using namespace std::literals;
 
-constexpr std::wstring_view expectedAppServerArgs = L"----AppNotificationActivated:"sv;
+constexpr std::wstring_view c_expectedAppServerArgs = L"----AppNotificationActivated:"sv;
 
 namespace winrt
 {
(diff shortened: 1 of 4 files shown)


dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -126,50 +126,22 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
                 m_registering = false;
             }) };
 
-            winrt::guid storedComActivatorGuid{ GUID_NULL };
-            if (!PushNotificationHelpers::IsPackagedAppScenario())
+            winrt::guid registeredClsid{};
+            if (AppModel::Identity::IsPackagedProcess())
```
hulumane, line 130: There is a missing scenario for Packaged Apps here. If it is a Packaged App running as UnpackagedScenario(due to Push), we need to create the correct mapping with LRP. Please sync with @pmpurifoy on details here

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -126,50 +126,22 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
-                }
+                registeredClsid = RegisterPackagedApp();
             }
-
-            winrt::guid registeredClsid{ GUID_NULL };
-            if (AppModel::Identity::IsPackagedProcess())
+            else
             {
-                registeredClsid = PushNotificationHelpers::GetComRegistrationFromRegistry(expectedAppServerArgs.data());
+                AppNotificationAssets assets{ GetAssets() };
```
hulumane, line 136: Can we use std::move semantics to avoid excessive copies here and below?

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -334,11 +362,14 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
-                        registeredClsid = PushNotificationHelpers::GetComRegistrationFromRegistry(expectedAppServerArgs.data());
+                        registeredClsid = PushNotificationHelpers::GetComRegistrationFromRegistry(c_expectedAppServerArgs.data());
                     }
                     else
                     { 
-                        registeredClsid = RegisterComActivatorGuidAndAssets();
+                        std::wstring registeredClsidString;
+                        THROW_IF_FAILED(GetActivatorGuid(registeredClsidString));
+
+                        registeredClsid = winrt::guid(registeredClsidString.substr(1, registeredClsidString.size() - 2));
```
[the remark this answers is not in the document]
hulumane: Bonus points if the helper actually returns a winrt::guid type and we can avoid this extra call.

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -178,6 +150,62 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
+        }
+
+        return registeredClsid;
+    }
+
+    winrt::guid AppNotificationManager::RegisterUnpackagedApp(AppNotificationAssets const& assets)
+    {
+        THROW_IF_FAILED(PushNotifications_RegisterFullTrustApplication(m_appId.c_str(), GUID_NULL));
+
+        std::wstring comActivatorGuidString{ GetOrCreateComActivatorGuid() };
```
hulumane, line 195: Can we have this helper return an actual winrt::GUID minus the braces?

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -178,6 +150,62 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
+
+        RegisterAssets(notificationAppId, comActivatorGuidString, assets);
+
+        if (!WindowsAppRuntime::SelfContained::IsSelfContained())
+        {
+            RegisterAppNotificationSinkWithNotificationPlatform();
+        }
+
+        // Remove braces around the guid string
+        return winrt::guid(comActivatorGuidString.substr(1, comActivatorGuidString.size() - 2));
```
hulumane, line 206: Would be great if the helper above took care of braces removal. Is that possible?

dev/AppNotifications/AppNotificationManager.cpp
hulumane, line 372: General comment. I see a lot of string to guid conversions and vice versa. There are also operations to remove braces from the string. Can we simply store the guid as winrt::guid and use that everywhere instead of the different conversions we have all over the place?

dev/AppNotifications/AppNotificationManager.h
```diff
@@ -46,6 +47,11 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
         winrt::Windows::Foundation::IInspectable Deserialize(winrt::Windows::Foundation::Uri const& uri);
     private:
 
+        winrt::guid RegisterPackagedApp();
+        winrt::guid RegisterUnpackagedApp(::Microsoft::Windows::AppNotifications::ShellLocalization::AppNotificationAssets const& assets);
+        void RegisterAppNotificationSinkWithNotificationPlatform();
```
hulumane, line 52: *WithLongRunningPlatform?

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -241,7 +243,7 @@ std::wstring Microsoft::Windows::AppNotifications::Helpers::GetDisplayNameBasedO
     return displayName;
 }
 
-winrt::guid Microsoft::Windows::AppNotifications::Helpers::RegisterComActivatorGuidAndAssets()
+std::wstring Microsoft::Windows::AppNotifications::Helpers::GetOrCreateComActivatorGuid()
```
hulumane, line 246: This returns an r-value. So we should be using std::move constructs in the caller to avoid unecessary copies.

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -126,50 +126,22 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
-                // Also, STA threads that call CoRegisterClassObject need to use the REGCLS_AGILE flag so that the object is
-                // associated with the neutral apartment. This allows other threads to activate the STA registered thread.
-                THROW_IF_FAILED(::CoRegisterClassObject(
-                    AppModel::Identity::IsPackagedProcess() ? registeredClsid : storedComActivatorGuid,
-                    winrt::make<AppNotificationManagerFactory>().get(),
-                    CLSCTX_LOCAL_SERVER,
-                    activationFlag | REGCLS_AGILE,
-                    &m_notificationComActivatorRegistration));
-            }
+            // Register AppNotificationManager Object for Shell to activate
```
hulumane, line 143: small nit with the comment: // Register the AppNotificationManager as a COM server for Shell to Activate and Invoke

hulumane: Maybe rename the helper below to RegisterComServer

dev/AppNotifications/AppNotificationManager.cpp
pmpurifoy: The reason for some of these conversions is to read/write to the registry. If there is a conversion from winrt::guid<->byte, I believe we can completely remove the std::wstring<->winrt::guid conversions.

pmpurifoy (approved):

dev/AppNotifications/AppNotificationUtility.cpp
sharath2727: Actually, we should probably leave this to the compiler. Based on what I am reading here: https://en.cppreference.com/w/cpp/language/copy_elision - the compiler tries to do a return value optimization first i.e.-"objects are constructed directly into the storage where they would otherwise be copied/moved to." - if it fails then it would switch to move and then finally to copy. Trying to std::move would end up counterintuitive to what is being achieved. I would be avoiding this in the PR. Let me know if you otherwise. Thanks.

dev/AppNotifications/AppNotificationManager.cpp
sharath2727: Actually, we should probably leave this to the compiler. Based on what I am reading here: https://en.cppreference.com/w/cpp/language/copy_elision - the compiler tries to do a return value optimization first i.e.-"objects are constructed directly into the storage where they would otherwise be copied/moved to." - if it fails then it would switch to move and then finally to copy. Trying to std::move would end up counterintuitive to what is being achieved. I would be avoiding this in the PR. Let me know if you otherwise. Thanks.

dev/AppNotifications/AppNotificationManager.cpp
sharath2727: Actually, we should probably leave this to the compiler. Based on what I am reading here: https://en.cppreference.com/w/cpp/language/copy_elision - the compiler tries to do a return value optimization first i.e.-"objects are constructed directly into the storage where they would otherwise be copied/moved to." - if it fails then it would switch to move and then finally to copy. Trying to std::move would end up counterintuitive to what is being achieved. I would be avoiding this in the PR. Let me know if you otherwise. Thanks.

dev/AppNotifications/AppNotificationManager.cpp
sharath2727: I created an internal task for this to track. Thanks.

danielayala94 (approved):
:shipit:

PR #2560, opened 2022-05-27
Remove Timestamp from VSIX version due to WinAppSDK NuGet Version Update

kythant:
Adding a timestamp should not be necessary anymore because WinAppSDK Nuget Versions should be unique.

### DIFF
--- a/dev/VSIX/Extension/Directory.Build.targets
+++ b/dev/VSIX/Extension/Directory.Build.targets
@@ -20,6 +20,13 @@
   </PropertyGroup>
   <PropertyGroup Condition="'$(VSIXVersion)' == ''">
     <!-- Use 3-part Windows App SDK version, less any tagging suffix -->
+    <!-- 
+      WindowsAppSDK release versions now have a timestamp as part of the patch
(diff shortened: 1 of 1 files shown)


dev/VSIX/Extension/Directory.Build.targets
```diff
@@ -22,14 +22,6 @@
     <!-- Use 3-part Windows App SDK version, less any tagging suffix -->
     <VSIXVersion>$(WindowsAppSDKVersion)</VSIXVersion>
     <VSIXVersion Condition="$(VSIXVersion.Contains(&quot;-&quot;))">$(VSIXVersion.Substring(0, $(VSIXVersion.IndexOf("-"))))</VSIXVersion>
-    <!--
```
DefaultRyan, line 25: Probably best to leave a comment about how this version is still required to be unique and increasing, and that this is now satisfied by the new scheme for `WindowsAppSDKVersion`.

DefaultRyan (approved):

DrusTheAxe (approved):

PR #2586, opened 2022-06-07
Update WIL. Update C++/WinRT. Tweak build order

DrusTheAxe:
Updated WIL references to 1.0.220201.1 (latest)

Updated CppWinRT references to 2.0. (latest)

Tweaked a few projects' BuildOrder to avoid build errors due to out-of-order compilation (can't build Foo dependent on Bar before building Bar. At least, not reliably, across a variety of hardware)


sachintaMSFT (approved):

DrusTheAxe:
> Be nice if there was a bot that could do this. Thanks for figuring out the build order, though, that's been annoying. :)

Maestro updates ~half the build dependencies. The other half are currently Maestro-oblivious and static decentralized artifacts.

I'm working on a script that'll report and/or fixup versions based on a central MasterSourceOfAllTruth(TM) file. When we decide to update a dependency the process will be "update version in text file" for Maestro or "update version in text file and run 1-liner command" for non-Maestro-managed dependencies.

I'd like to see the nightly produce a report of any inconsistencies and eventually make it part of the PR build process, error on any inconsistencies so developers adding or updating projects would have to run the script locally to sync their new bits to the standard tooling before checkin.

One step at a time :-)

PR #2596, opened 2022-06-07
Introduce Register Overload for assets

sharath2727:
Description: Allow developers to input DisplayName and Icon of their choice to be registered with AppNotificationManager so that these assets can show up when AppNotifications are posted or received.

Validation: Introduced unit tests and validated them.


dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -286,6 +286,15 @@ AppNotificationAssets Microsoft::Windows::AppNotifications::Helpers::GetAssets()
     return assets;
 }
 
+AppNotificationAssets Microsoft::Windows::AppNotifications::Helpers::GetAssets(std::wstring displayName,std::wstring iconFilePath)
+{
+    winrt::check_bool(std::filesystem::exists(iconFilePath));
```
danielayala94, line 291: What error is thrown if the file doesn't exist?

test/TestApps/ToastNotificationsTestApp/main.cpp
```diff
@@ -1322,6 +1322,171 @@ bool VerifyIconPathExists_Unpackaged()
...
+    }
+    catch (...)
+    {
+        return winrt::to_hresult() == E_POINTER;
+    }
+
+    return false;
+}
+
+bool VerifyRegisterWithAssetsFail()
```
danielayala94, line 1472: What's the reason being of the failure?

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -286,6 +286,15 @@ AppNotificationAssets Microsoft::Windows::AppNotifications::Helpers::GetAssets()
     return assets;
 }
 
+AppNotificationAssets Microsoft::Windows::AppNotifications::Helpers::GetAssets(std::wstring displayName,std::wstring iconFilePath)
```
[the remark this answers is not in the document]
danielayala94: Also, this is more about verifying assets given by the developer, not getting assets.

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -286,6 +286,15 @@ AppNotificationAssets Microsoft::Windows::AppNotifications::Helpers::GetAssets()
...
+{
+    winrt::check_bool(std::filesystem::exists(iconFilePath));
+
+    THROW_HR_IF_MSG(E_INVALIDARG, !IsIconFileExtensionSupported(std::filesystem::path{ iconFilePath }), "Icon format not supported");
```
danielayala94, line 293: You are implicitly creating 3 copies of the iconFilePath: one for the parameter (hstring->wstring), one for the check_bool (wstring->filesystem) and the same for verifying icon file extension. You can at least save one copy.

Also, I'm wondering if you can create a path directly from a hstring?

test/TestApps/ToastNotificationsTestApp/main.cpp
danielayala94: I see this is for packaged. Please add a comment stating it.

danielayala94 (changes requested):

danielayala94:
While there are some changes to do, I think this PR is ok for general review.

dev/AppNotifications/AppNotificationUtility.cpp
sharath2727: ERROR_FILE_NOT_FOUND

test/TestApps/ToastNotificationsTestApp/main.cpp
```diff
@@ -1322,6 +1322,171 @@ bool VerifyIconPathExists_Unpackaged()
     return true;
 }
 
+bool VerifyRegisterWithNullDisplayNameFail_Unpackaged()
```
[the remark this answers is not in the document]
sharath2727: winrt::hstring str = winrt::hstring{}; // this is like assigning nullptr to hstr but if you decode str.c_str() it would return L"". I think I would like to stick with the naming here.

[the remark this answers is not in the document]
sharath2727: Btw there is also another unittest: VerifyRegisterWithEmptyDisplayNameFail_Unpackaged() which specifically checks for L""

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -150,6 +151,54 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
+
+        auto logTelemetry{ wil::scope_exit([&]() {
+            AppNotificationTelemetry::LogRegister(hr, m_appId);
+        }) };
+
+        try
+        {
+            THROW_HR_IF_MSG(E_ILLEGAL_METHOD_CALL, AppModel::Identity::IsPackagedProcess(), "Not applicable for packaged applications");
+
+            THROW_HR_IF(E_INVALIDARG, (displayName == winrt::hstring{}) || (iconUri == nullptr));
```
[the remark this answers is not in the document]
sharath2727: I felt here it is self-explanatory that the provided arguments are wrong. So did not want to. Will keep it this way for now. Thanks.

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -150,6 +151,54 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
     }
 
+    void AppNotificationManager::Register(hstring const& displayName, winrt::Windows::Foundation::Uri const& iconUri)
+    {
+        if (!IsSupported())
+        {
+            return;
+        }
+
+        HRESULT hr{ S_OK };
```
DrusTheAxe, line 161: NIT: `HRESULT hr{};`

No need to specify the defaults if they're 0/0.0/false/nullptr (they're the defaults)

dev/AppNotifications/AppNotificationManager.cpp
DrusTheAxe, line 171: `displayName.empty()`

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -150,6 +151,54 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
+
+            // Create event handle before COM Registration otherwise if a notification arrives will lead to race condition
+            m_waitHandleForArgs.create();
+
+            // Register the AppNotificationManager as a COM server for Shell to Activate and Invoke
+            RegisterComServer(registeredClsid);
+        }
+        catch (...)
+        {
+            hr = wil::ResultFromCaughtException();
```
DrusTheAxe, line 197: Is this catch block needed? Won't C++/WinRT code that called to do do this?

dev/AppNotifications/AppNotificationUtility.h
```diff
@@ -65,4 +65,6 @@ namespace Microsoft::Windows::AppNotifications::Helpers
     std::wstring GetDisplayNameBasedOnProcessName();
 
     Microsoft::Windows::AppNotifications::ShellLocalization::AppNotificationAssets GetAssets();
+
+    Microsoft::Windows::AppNotifications::ShellLocalization::AppNotificationAssets ValidateAssets(std::wstring displayName, std::filesystem::path iconFilePath);
```
DrusTheAxe, line 69: `const std::wstring&` ?

dev/AppNotifications/ShellLocalization.cpp
```diff
@@ -167,7 +167,8 @@ void WriteHIconToPngFile(wil::unique_hicon const& hIcon, _In_ PCWSTR pszFileName
     THROW_IF_FAILED(spStreamOut->Commit(STGC_DANGEROUSLYCOMMITMERELYTODISKCACHE));
 }
 
-bool IsIconFileExtensionSupported(std::filesystem::path const& iconFilePath)
+
+bool Microsoft::Windows::AppNotifications::ShellLocalization::IsIconFileExtensionSupported(std::filesystem::path const& iconFilePath)
```
DrusTheAxe, line 171: `const std::filesystem::path&` ?

riverar: https://hackingcpp.com/cpp/design/east_vs_west_const.html

dev/AppNotifications/AppNotificationManager.cpp
hulumane: Why is there a try catch here at all @sharath2727 ? We are throwing anyway

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -150,6 +151,54 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
     }
 
+    void AppNotificationManager::Register(hstring const& displayName, winrt::Windows::Foundation::Uri const& iconUri)
+    {
+        if (!IsSupported())
+        {
+            return;
+        }
+
+        HRESULT hr{};
```
hulumane, line 161: Nit: Better be explicit here and set it to S_OK

dev/AppNotifications/AppNotificationUtility.cpp
```diff
@@ -286,6 +286,15 @@ AppNotificationAssets Microsoft::Windows::AppNotifications::Helpers::GetAssets()
     return assets;
 }
 
+AppNotificationAssets Microsoft::Windows::AppNotifications::Helpers::ValidateAssets(std::wstring const& displayName, std::filesystem::path iconFilePath)
+{
+    winrt::check_bool(std::filesystem::exists(iconFilePath));
+
+    THROW_HR_IF_MSG(E_INVALIDARG, !IsIconFileExtensionSupported(iconFilePath), "Icon format not supported");
+
+    return AppNotificationAssets{ std::move(displayName), iconFilePath.wstring() };
```
hulumane, line 295: Is the ctor using an R-Value ref? If not, why is std::move needed if it is simply a pass by ref?

hulumane: Ok looks like you don't need r-value semantics to use std::move. So this should technically work.
https://stackoverflow.com/questions/59553364/c-using-move-semantics-to-pass-argument-to-another-function

hulumane: This is the key:
Note on move semantics

If you use move semantics, the arguments of your functions don't necessarily need the double &&. Generally, it is better to omit them (except for perfect forwarding and non-copyable objects like stream objects). The && requires that the passed arguments be rvalues. However, omitting the && does not mean you cannot pass the arguments as rvalues. Let's see this with an example:

dev/AppNotifications/AppNotificationUtility.h
```diff
@@ -65,4 +65,6 @@ namespace Microsoft::Windows::AppNotifications::Helpers
     std::wstring GetDisplayNameBasedOnProcessName();
 
     Microsoft::Windows::AppNotifications::ShellLocalization::AppNotificationAssets GetAssets();
+
+    Microsoft::Windows::AppNotifications::ShellLocalization::AppNotificationAssets ValidateAssets(std::wstring const& displayName, std::filesystem::path iconFilePath);
```
hulumane, line 69: Why is one argument const& and not the other?

dev/AppNotifications/AppNotifications.idl
```diff
@@ -123,6 +123,10 @@ namespace Microsoft.Windows.AppNotifications
         // For Unpackaged apps, the caller process will be registered as the COM server. And assets like displayname and icon will be gleaned from Shell and registered as well.
         void Register();
 
+        // For Unpackaged apps only, the caller process will be registered as the COM server.
```
hulumane, line 126: This statement is untrue.

hulumane: Packaged apps are also registered as a COM server for AppNotifications

hulumane: The only difference is that we define the COM server in a manifest for packaged apps. For unpackaged, we write to registry.

After that we call CoRegister on both.

dev/AppNotifications/ShellLocalization.cpp
hulumane: Yep good to know.

dev/AppNotifications/ShellLocalization.cpp
```diff
@@ -258,10 +259,7 @@ HRESULT Microsoft::Windows::AppNotifications::ShellLocalization::DeleteIconFromC
     std::path iconFilePath{ RetrieveLocalFolderPath() / (notificationAppId + c_pngExtension) };
 
     // If DeleteFile returned FALSE, then deletion failed and we should return the corresponding error code.
-    if (DeleteFile(iconFilePath.c_str()) == FALSE)
-    {
-        THROW_HR(HRESULT_FROM_WIN32(GetLastError()));
-    }
+    winrt::check_bool(DeleteFileW(iconFilePath.c_str()) != FALSE);
```
hulumane, line 262: Always love condensation of code! Good to catch this!

specs/AppNotifications/AppNotifications-spec.md
```diff
@@ -139,6 +139,29 @@ int main()
...
+The developer should provide both the assets or not provide them at all. Icon provided by the developer
+should be a valid supported format. The API supports the formats - png, bmp, jpg. The icon
+should reside on the local machine only otherwise the API throws an exception. For Packaged applications,
+this API is not applicable and will throw an exception.
+
+```cpp
+int main()
+{
+    auto manager = winrt::AppNotificationManager::Default();
+    manager.Register(L"AppNotifications", winrt::Windows::Foundation::Uri {L"<Path to Icon>"});
```
hulumane, line 155: Please use a valid icon path. Ideally, you retrieve the icon path from the process path and then use it to create a Uri object. You need to demonstrate that here.

specs/AppNotifications/AppNotifications-spec.md
```diff
@@ -451,6 +474,9 @@ namespace Microsoft.Windows.AppNotifications
         // For Unpackaged apps, the caller process will be registered as the COM server. And assets like displayname and icon will be gleaned from Shell and registered as well.
         void Register();
 
+         // For Unpackaged apps only, the caller process will be registered as the COM server.
```
hulumane, line 477: Wrong comment. The App is a COM server for both Packaged and Unpackaged apps. We call CoREgister for both. The only difference is that unpackaged apps can override their assets.

hulumane: The comment above this is also confusing

test/TestApps/ToastNotificationsTestApp/main.cpp
```diff
@@ -1322,6 +1322,154 @@ bool VerifyIconPathExists_Unpackaged()
...
+}
+
+bool VerifyRegisterWithEmptyDisplayNameFail_Unpackaged()
+{
+    // Register is already called in main with an explicit appusermodelId
+    winrt::AppNotificationManager::Default().UnregisterAll();
+    try
+    {
+        // hstring treats L"" as assigning nullptr
+        winrt::AppNotificationManager::Default().Register(L"", winrt::Uri{ LR"(C:\Windows\System32\WindowsSecurityIcon.png)" });
```
hulumane, line 1414: What if Windows is installed on d: drive? Hardcoding paths is always bad. The display asset should be in the same path as the process. Then you need not hardcode paths and can retrieve it dynamically.

test/TestApps/ToastNotificationsTestApp/main.cpp
```diff
@@ -1322,6 +1322,154 @@ bool VerifyIconPathExists_Unpackaged()
...
+    return false;
+}
+
+bool VerifyRegisterWithEmptyIconFail_Unpackaged()
+{
+    // Register is already called in main with an explicit appusermodelId
+    winrt::AppNotificationManager::Default().UnregisterAll();
+    try
+    {
+        winrt::AppNotificationManager::Default().Register(L"AppNotificationApp", winrt::Uri{ L"" });
```
hulumane, line 1430: This is a weird test since it is not testing our APIs. winrt::Uri is technically failing here because of invalid arg and throwing

test/TestApps/ToastNotificationsTestApp/main.cpp
```diff
@@ -1322,6 +1322,154 @@ bool VerifyIconPathExists_Unpackaged()
...
+}
+
+bool VerifyRegisterWithAssetsFail()
+{
+    // Register is already called in main with an explicit appusermodelId
+    winrt::AppNotificationManager::Default().UnregisterAll();
+    try
+    {
+        // API fails for Packaged Scenario
+        winrt::AppNotificationManager::Default().Register(L"AppNotificationApp", winrt::Uri{ LR"(C:\InvalidPath\)" });
```
hulumane, line 1463: You don't have any tests to Verify if toasts works after registration. Please think about how your new feature integrates with other feature sets and test those code paths as well.

test/TestApps/ToastNotificationsTestApp/main.cpp
```diff
@@ -1322,6 +1322,154 @@ bool VerifyIconPathExists_Unpackaged()
...
+    return true;
+}
+
+bool VerifyRegisterWithDisplayNameAndIcon_Unpackaged()
+{
+    // Register is already called in main with an explicit appusermodelId
+    winrt::AppNotificationManager::Default().UnregisterAll();
+    try
+    {
+        winrt::AppNotificationManager::Default().Register(L"AppNotificationApp", winrt::Uri{ LR"(C:\Windows\System32\WindowsSecurityIcon.png)" });
```
hulumane, line 1379: We don't check anywhere if these values are actually written correctly to the registry?

dev/AppNotifications/AppNotificationManager.cpp
sharath2727: Yes this is for logging purpose. Once eric introduces activities as part of his work all these will go away. Until then we need to continue to log this way. Thanks.

test/TestApps/ToastNotificationsTestApp/main.cpp
sharath2727: Ah....yeah we could avoid this test.

test/TestApps/ToastNotificationsTestApp/main.cpp
sharath2727: I will create a bug for this internally and address this at a later point in time. Thanks.

specs/AppNotifications/AppNotifications-spec.md
```diff
@@ -139,6 +139,31 @@ int main()
...
+should reside on the local machine only otherwise the API throws an exception. For Packaged applications,
+this API is not applicable and will throw an exception.
+
+```cpp
+int main()
+{
+    auto manager = winrt::AppNotificationManager::Default();
+
+    std::wstring iconFilepath{ std::filesystem::current_path() / "icon.ico" };
+    manager.Register(L"AppNotifications", winrt::Windows::Foundation::Uri {iconFilepath});
```
hulumane, line 157: Rather than hardcode the Displayname, consider using the Processname instead.

dev/AppNotifications/ShellLocalization.cpp
```diff
@@ -259,7 +259,7 @@ HRESULT Microsoft::Windows::AppNotifications::ShellLocalization::DeleteIconFromC
     std::path iconFilePath{ RetrieveLocalFolderPath() / (notificationAppId + c_pngExtension) };
 
     // If DeleteFile returned FALSE, then deletion failed and we should return the corresponding error code.
-    winrt::check_bool(DeleteFileW(iconFilePath.c_str()) != FALSE);
+    RETURN_IF_WIN32_BOOL_FALSE(DeleteFileW(iconFilePath.c_str()));
```
danielayala94, line 262: What was the reason to move away from `winrt::check_bool`?

specs/AppNotifications/AppNotifications-spec.md
```diff
@@ -139,6 +139,31 @@ int main()
...
+should be a valid supported format. The API supports the formats - png, bmp, jpg. The icon
+should reside on the local machine only otherwise the API throws an exception. For Packaged applications,
+this API is not applicable and will throw an exception.
+
+```cpp
+int main()
+{
+    auto manager = winrt::AppNotificationManager::Default();
+
+    std::wstring iconFilepath{ std::filesystem::current_path() / "icon.ico" };
```
danielayala94, line 156: Can the icon be in a different path than the exe, like in a subfolder?

danielayala94 (approved):

dev/AppNotifications/ShellLocalization.cpp
sharath2727: RETURN_IF_WIN32_BOOL_FALSE - I felt this is more appropriate as it directly checks for FALSE flag.

specs/AppNotifications/AppNotifications-spec.md
```diff
@@ -144,17 +144,39 @@ int main()
...
+
+    std::wstring displayName{};
+    wil::GetModuleFileNameExW(GetCurrentProcess(), nullptr, displayName);
```
danielayala94, line 177: The thing that doesn't make me comfortable about this is that we pass a filename as the display name (the dev is free to pass whatever value they want, but the point of this API is to let devs to pass a pretty display name to be shown on the app notifications).

sharath2727: these are just examples to show them they could do something like this but yeah the developers are free to pass whatever pretty value they want and we should adhere to it and we are. :)

hulumane: Why don't you declare a const called c_displayName and assign it to something like "DemoToastApp"? Then reuse that here.

PR #2599, opened 2022-06-08
Fix issue where WinUI C# and C++ item templates are not appearing in WinUI projects

japarson:
## Problem
In a [previous PR](https://github.com/microsoft/WindowsAppSDK/pull/2289), I attempted to prevent the WinUI C# templates from appearing in projects unrelated to WinUI. To do so, I set the [ShowByDefault](https://docs.microsoft.com/en-us/visualstudio/extensibility/showbydefault-visual-studio-templates?view=vs-2022) property to `false`. When this property is `false`, the decision to show/hide the item templates falls to the [AppliesTo](https://docs.microsoft.com/en-us/visualstudio/extensibility/appliesto-element-visual-studio-templates?view=vs-2022) and [TemplateGroupID](https://docs.microsoft.com/en-us/visualstudio/extensibility/templategroupid-element-visual-studio-templates?view=vs-2022) properties.

The C# templates currently hold the value `CSharp + SharedAssetsProject`. However, an examination of the capabilities for a WinUI project reveals that `SharedAssetsProject` is no longer present:
image

The C++ templates currently hold no value for `AppliesTo` and `TemplateGroupID`.

## Solution
For the C# templates, we can adjust the value for `AppliesTo` to be `CSharp + WindowsAppSdk`.

For the C++ templates, we can add the property `AppliesTo` with value `VisualC + WindowsAppSdk`. We can also add the property `TemplateGroupID` with value `WinRT-Native-UAP`, which comes from this table ([link](https://docs.microsoft.com/en-us/visualstudio/extensibility/templategroupid-element-visual-studio-templates?view=vs-2022)):

image

## C#
image

## C++
image

## Work Item
https://github.com/microsoft/microsoft-ui-xaml/issues/7148

### DIFF
--- a/dev/VSIX/ItemTemplates/Desktop/CSharp/BlankWindow/WinUI.Desktop.Cs.BlankWindow.vstemplate
+++ b/dev/VSIX/ItemTemplates/Desktop/CSharp/BlankWindow/WinUI.Desktop.Cs.BlankWindow.vstemplate
@@ -8,7 +8,7 @@
     <ProjectType>CSharp</ProjectType>
     <TemplateID>WinUI.Desktop.Cs.BlankWindow</TemplateID>
     <TemplateGroupID>WinRT-Managed-UAP</TemplateGroupID>
-    <AppliesTo>CSharp + SharedAssetsProject</AppliesTo>
+    <AppliesTo>CSharp + WindowsAppSDK</AppliesTo>
     <NumberOfParentCategoriesToRollUp>2</NumberOfParentCategoriesToRollUp>
     <ShowByDefault>false</ShowByDefault>
     <TargetPlatformName>Windows</TargetPlatformName>
--- a/dev/VSIX/ItemTemplates/Desktop/CppWinRT/BlankWindow/WinUI.Desktop.CppWinRT.BlankWindow.vstemplate
+++ b/dev/VSIX/ItemTemplates/Desktop/CppWinRT/BlankWindow/WinUI.Desktop.CppWinRT.BlankWindow.vstemplate
@@ -7,6 +7,8 @@
     <Icon>WinUI.Desktop.CppWinRT.BlankWindow.ico</Icon>
     <ProjectType>VC</ProjectType>
     <TemplateID>WinUI.Desktop.CppWinRT.BlankWindow</TemplateID>
+    <TemplateGroupID>WinRT-Native-UAP</TemplateGroupID>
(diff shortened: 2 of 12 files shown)


evelynwu-msft (approved):
:shipit:

TRadigk:
I think it broke again in Version 17.3.0 Preview 2.0

evelynwu-msft:
> I think it broke again in Version 17.3.0 Preview 2.0

@TRadigk Could you explain what you mean? This fix hasn't been released yet so it shouldn't be possible for it to have regressed.

SunnieShine:
@evelynwu-msft Hello. Sorry to disturb you.

I use VS 17.3 preview 2 but the same issue encountered.

Pictures:

Template missing:

image

The elementary WinUI VS extensions has been installed:

image

I use VS 17.3p2:

image

Templates has been successfully installed.

image

I don't think that the issue has been fixed in the current version. 😞

TRadigk:
> > I think it broke again in Version 17.3.0 Preview 2.0
> 
> @TRadigk Could you explain what you mean? This fix hasn't been released yet so it shouldn't be possible for it to have regressed.

Thanks for the information. I was quite certain I had no trouble working with the templates recently, thus assuming prior the latest VS preview update. 
Do you know of any workaround, to get it working until the fix is released? Re-installing the extension doesn't do the trick. I am glad I can continue most of my tasks in VS 2022 non-preview, so I am rather curious than limited by VS 2022 preview.

evelynwu-msft:
@TRadigk This will be addressed in the upcoming servicing update for 1.1. Unfortunately, the only workaround until then is to install the WinAppSdk 1.0.4 extension ([C# Visual Studio 2022 extension](https://aka.ms/windowsappsdk/1.0/1.0.4/vsix-2022-cs) and [C++ Visual Studio 2022 extension](https://aka.ms/windowsappsdk/1.0/1.0.4/vsix-2022-cpp)) and manually upgrade the package version when creating a new project (the package upgrade step can be skipped if you're working with an existing project).

evelynwu-msft:
@TRadigk My mistake, the servicing update with the fixed templates has already been released; updating your installed extension to 1.1.2 should resolve this problem for you: [C# Visual Studio 2022 extension](https://aka.ms/windowsappsdk/1.1/1.1.2/vsix-2022-cs) and [C++ Visual Studio 2022 extension](https://aka.ms/windowsappsdk/1.1/1.1.2/vsix-2022-cpp).

PR #2608, opened 2022-06-10
v1.1 SDK throws an access violation from MddBootstrapInitialize's logging code if no matching packages are installed #2592

### LINKED ISSUE #2592: v1.1 SDK throws an access violation from MddBootstrapInitialize's logging code if no matching packages are installed
### Describe the bug

My app currently calls `MddBootstrapInitialize()` when running unpackaged:

https://github.com/fredemmott/OpenKneeboard/blob/794c53568b34f18443027c43135b7a90e5ce47eb/src/app/app-winui3/App.xaml.cpp#L200-L204

After upgrading from 1.0 to 1.1 - changing the c++/winrt package to 2.0.220418.1 (the latest version is compatible with WinUI3 when building with C++20) and windows app SDK 1.1.0, this has the following output:

```
D:\a\_work\1\s\BuildOutput\Release\x64\WindowsAppRuntime_DLL\WindowsAppRuntimeInsights.h(53)\Microsoft.WindowsAppRuntime.Bootstrap.dll!00007FF981FE24DB: (caller: 00007FF981FDE1C6) LogHr(1) tid(5c44) 8007007E The specified module could not be found.
    Msg:[Unable to load resource dll. Microsoft.WindowsAppRuntime.Insights.Resource.dll] 
D:\a\_work\1\s\dev\WindowsAppRuntime_BootstrapDLL\MddBootstrap.cpp(776)\Microsoft.WindowsAppRuntime.Bootstrap.dll!00007FF981FDB0D9: (caller: 00007FF981FD9C2F) LogHr(2) tid(5c44) 80040010 Object is not in any of the inplace active states
    Msg:[Bootstrap.Intitialize: Scanning packages for Major.Minor=1.1, Tag=, MinVersion=1000.516.2156.0] CallContext:[\Initialize] 
Exception thrown at 0x00007FF981FE4403 (Microsoft.WindowsAppRuntime.Bootstrap.dll) in OpenKneeboardApp.exe: 0xC0000005: Access violation writing location 0x00007FF981FF6716.
```

windbg points here:

```C++
void MddBootstrap_StopActivity(
    const std::string failureType,
    const bool isActivityRunning,
    const GUID *activityId,
    const WindowsAppRuntime::MddBootstrap::Activity::Context& activityContext,
    const wil::FailureInfo& failure)
{
    MddBootstrap_WriteEventWithActivity(*failureType.c_str(), activityId);
```

...

called from:

```
    THROW_HR_IF_MSG(HRESULT_FROM_WIN32(ERROR_NO_MATCH), !foundAny, "Enumeration: %ls", criteria.get());
    (void)LOG_HR_MSG(MDD_E_BOOTSTRAP_INITIALIZE_DDLM_FOUND,
                     "Bootstrap.Intitialize: %ls best matches the criteria (%ls) of %d packages scanned",
                     bestFitPackageFullName.c_str(), criteria.get(), packagesScanned);
```

`bestFitPackageFullName` is a null handle.

1. Grab fredemmott/openkneeboard at rev 794c53568b34f18443027c43135b7a90e5ce47eb
2. Apply https://gist.github.com/fredemmott/96216c79d2e0496e128a7923a6ac6777 to update windows app SDK and C++/winrt
3. build with cmake in debug mode
4. run app, e.g. `build\src\app\app-winui3\msbuild-out-Debug\OpenKneeboardApp.exe`

### Expected behavior

App should start, or exit cleanly, depending on state of installed packages.

### Screenshots

_No response_

### NuGet package version

_No response_

### Packaging type

Unpackaged

### Windows version

Windows 10 version 21H2 (19044, November 2021 Update)

### IDE

Other

### Additional context

* using 1.1.0 nuget package, which isn't in list
* using vscode + cmake

DrusTheAxe:
v1.1 SDK throws an access violation from MddBootstrapInitialize's logging code if no matching packages are installed #2592

```
MddBootstrap_WriteEventWithActivity(*failureType.c_str(), activityId);
```

but that first parameter needs to be a **string literal**. Alas, `MddBootstrap_WriteEventWithActivity` is a macro over `TraceLoggingWriteActivity` which relies on its macroness to expand `char foo[sizeof(_eventname_)]` to the size of the string literal to be stored in place. If that parameter's a variable sizeof(char*) == 4 or 8, not the actual length needed. At runtime if the runtime isn't installed the Bootstrapper fails as expected but calling telemetry before returning is bad (usually a crash).

Resolving this unearthed a bug in Dynamic Dependencies returning an error if `LifetimeKind`=`FilePath`|`RegistryKey` because it looked for ApplicationData.LocalFolder for the package family MicrosoftCorporationII.<b>WindowsAppRuntime</b>.Main.<b>1.0</b>_8wekyb3d8bbwe`. This doesn't exist (needs be `WinAppRuntime`) and 1.1+ users shouldn't use or need 1.0 installed. This was corrected. Also, the new error code `MDD_E_WINDOWSAPPRUNTIME_DATASTORE_NOT_FOUND` is returned if we can't find the AppData where we store persisted DynDep definitions (typically because the Main package isn't registered for the usually).

Most of the changes are due to the changed test package names and data trickling through to account for the above fixes.

NOTE: The test runtime packages now include their release number in their name, just like the product names. This wasn't a problem in the past but there are now places in the runtime that dynamically discover the runtime packages and parse out the release info from their package Names. This was causing test failures, so the test runtime packages are now named similarly to the actual runtime (`blahblah-4.1` as the test runtime packages are version 4.1.something.something).

This PR also corrects unrelated test failures in main that were causing TestExplorer's results to be muddied with unrelated noise. So I fixed most of them; following up with feature teams for the few others.

closes #2592

https://task.ms/39995920 for 1.1-servicing consideration

### DIFF
--- a/dev/WindowsAppRuntime_BootstrapDLL/MddBootstrapTracelogging.cpp
+++ b/dev/WindowsAppRuntime_BootstrapDLL/MddBootstrapTracelogging.cpp
@@ -6,13 +6,13 @@
 #include "pch.h"
 
 void MddBootstrap_StopActivity(
-    const std::string failureType,
+    const std::string& failureType,
     const bool isActivityRunning,
     const GUID *activityId,
     const WindowsAppRuntime::MddBootstrap::Activity::Context& activityContext,
     const wil::FailureInfo& failure)
 {
-    MddBootstrap_WriteEventWithActivity(*failureType.c_str(), activityId);
+    MddBootstrap_WriteEventWithActivity(failureType.c_str(), activityId);
 
     if (isActivityRunning)
     {
--- a/dev/WindowsAppRuntime_BootstrapDLL/MddBootstrapTracelogging.cpp
+++ b/dev/WindowsAppRuntime_BootstrapDLL/MddBootstrapTracelogging.cpp
@@ -6,22 +6,18 @@
 #include "pch.h"
 
 void MddBootstrap_StopActivity(
-    const std::string& failureType,
     const bool isActivityRunning,
-    const GUID *activityId,
-    const WindowsAppRuntime::MddBootstrap::Activity::Context& activityContext,
+    WindowsAppRuntime::MddBootstrap::Activity::Context& activityContext,
     const wil::FailureInfo& failure)
 {
-    MddBootstrap_WriteEventWithActivity(failureType.c_str(), activityId);
-
     if (isActivityRunning)
     {
         PWSTR initializationFrameworkPackageFullName{};
-        auto initializationCount{ WindowsAppRuntime::MddBootstrap::Activity::Context::Get().GetInitializeData(initializationFrameworkPackageFullName) };
+        auto initializationCount{ activityContext.GetInitializeData(initializationFrameworkPackageFullName) };
 
         if (activityContext.GetMddBootstrapAPI() == WindowsAppRuntime::MddBootstrap::Activity::MddBootstrapAPI::Initialize)
         {
-            WindowsAppRuntime::MddBootstrap::Activity::Context::Get().GetInitializeActivity().StopWithResult(
+            activityContext.GetInitializeActivity().StopWithResult(
                 failure.hr,
                 static_cast<UINT32>(initializationCount),
                 static_cast<UINT32>(WindowsAppRuntime::MddBootstrap::Activity::Context::GetIntegrityFlags()),
@@ -34,7 +30,7 @@ void MddBootstrap_StopActivity(
         }
         else if (activityContext.GetMddBootstrapAPI() == WindowsAppRuntime::MddBootstrap::Activity::MddBootstrapAPI::Shutdown)
         {
-            WindowsAppRuntime::MddBootstrap::Activity::Context::Get().GetShutdownActivity().StopWithResult(
+            activityContext.GetShutdownActivity().StopWithResult(
                 failure.hr,
                 static_cast<UINT32>(initializationCount),
                 static_cast<UINT32>(activityContext.GetLastFailure().type),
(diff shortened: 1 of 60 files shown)


dev/WindowsAppRuntime_BootstrapDLL/MddBootstrapTracelogging.cpp
```diff
@@ -6,13 +6,13 @@
...
 void MddBootstrap_StopActivity(
-    const std::string failureType,
+    const std::string& failureType,
     const bool isActivityRunning,
     const GUID *activityId,
     const WindowsAppRuntime::MddBootstrap::Activity::Context& activityContext,
     const wil::FailureInfo& failure)
 {
-    MddBootstrap_WriteEventWithActivity(*failureType.c_str(), activityId);
+    MddBootstrap_WriteEventWithActivity(failureType.c_str(), activityId);
```
fredemmott, line 15: AIUI this is still incorrect: the inner ETW macro is documented as requiring a string *literal*, and uses sizeof to get the string length. With this code, it will get the size of a pointer instead:

```C
int main() {
  const char* str = "1234567890123";
  printf("%d %d\n", sizeof("1234567890123"), sizeof(str));
}
```

Output:

```
Execution build compiler returned: 0
Program returned: 0
14 8
```

https://godbolt.org/z/eGG8n794d

fredemmott: I don't think there's a way to fix this without substantially changing this function  (e.g. replacing it with a macro), as ETW does not support variable event names.

DrusTheAxe: Yeah, my local testing is pointing that out (I threw up the PR to get it backed up and optimistically start a pipeline build while I do local build+test). The macro definitions steep several layers deep but string literal's a necessity. Juggling code t comply...

dhoehna:
Looks good to me.

dhoehna (approved):

PR #2629, opened 2022-06-15
Do a compiler options pass on all binaries contributing to WinAppSDK

DrusTheAxe:
Tweak various compiler options for optimization (e.g. /GR-) and correctness/productivity (e.g. /W4).

Project-wide standardization via new \WindowsAppSDK.Build.Cpp.props. See for more details.

Corrected several inconsistencies across projects e.g. given Debug|Release + x86+x64+arm64 = 6 permutations, but some projects had an option set for only 5, or only 1. Most projects were mostly already setting most of these options so their changes were mostly removing the now-redundant project-specific setting and let the top-level project-wide standards handle it.

This now standardizes all C++ projects to...
* Compile (all): `/W4 /sdl /permissive- /std:c++17 /GR-` (except 2 projects that explicitly set `/std:c++20`)
* Compile (debug): `/Od`
* Compile (release): `/GS /GF /GL /Gy /Gw /Ob2 /Oi /Ot /Oy`
* Link (release): `/LTCG:incremental /OPT:ICF /OPT:REF`

https://task.ms/39994837 Foundation

https://task.ms/39994852 MRTCore

### DIFF
--- a/WindowsAppSDK.Build.Cpp.props
+++ b/WindowsAppSDK.Build.Cpp.props
@@ -1,0 +1,78 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <!-- Build with the hybrid CRT (Universal CRT + Static VS CRT (for what little the Universal CRT doesn't cover) -->
+  <Import Project="$(MSBuildThisFileDirectory)HybridCRT.props" />
+
+  <!-- Build with APIscan-friendly compiler+linker options -->
+  <Import Project="$(MSBuildThisFileDirectory)ApiScan.Cpp.props" />
+
+  <!-- Common (all Configurations) options -->
+  <ItemDefinitionGroup>
+    <ClCompile>
+      <!-- /std:c++17 Enable ISO C++17 Standard -->
+      <LanguageStandard>stdcpp17</LanguageStandard>
+
+      <!-- /permissive- Enable Conformance mode (i.e. Disable 'permissive' mode) -->
+      <ConformanceMode>true</ConformanceMode>
+
+      <!-- /W4 Enable max warning level -->
+      <WarningLevel>Level4</WarningLevel>
+
+      <!-- /GR- Disable RTTI (causes binary bloat) -->
+      <RuntimeTypeInfo>false</RuntimeTypeInfo>
+    </ClCompile>
+    <Link>
+    </Link>
+  </ItemDefinitionGroup>
+
+  <!-- Debug-specific options -->
+<!--
+  <ItemDefinitionGroup Condition="'$(Configuration)' == 'Debug'">
+    <ClCompile>
+    </ClCompile>
+    <Link>
+    </Link>
+  </ItemDefinitionGroup>
+-->
+
+  <!-- Release-specific options -->
+  <ItemDefinitionGroup Condition="'$(Configuration)' == 'Release'">
+    <ClCompile>
+      <!-- /GS Enable Control Flow Guard -->
+      <ControlFlowGuard>Guard</ControlFlowGuard>
+
+      <!-- Compile-time optimizations:
+           /GF Enable read-only string pooling (eliminate duplicate strings)
+           /GL Whole Program Optimization
+           /Gy Enable function-level linking
+           /Gw Optimize Global Data (https://docs.microsoft.com/en-us/cpp/build/reference/gw-optimize-global-data)
+        -->
+      <StringPooling>true</StringPooling>
+      <WholeProgramOptimization>true</WholeProgramOptimization>
+      <FunctionLevelLinking>true</FunctionLevelLinking>
+      <AdditionalOptions>%(AdditionalOptions) /Gw</AdditionalOptions>
+
+      <!-- /Ox Enable most speed optimizations
+           NOTE: /Ox is short-hand for multiple options:
+                     /Ob2 Enable inline function expansion (/Ob2 = Any Suitable)
+                     /Oi Enable intrinsic functions
+                     /Ot Favor fast code
+                     /Oy Omit frame pointers
+        -->
+      <InlineFunctionExpansion>AnySuitable</InlineFunctionExpansion>
+      <IntrinsicFunctions>true</IntrinsicFunctions>
+      <FavorSizeOrSpeed>Speed</FavorSizeOrSpeed>
+      <OmitFramePointers>true</OmitFramePointers>
+    </ClCompile>
+    <Link>
+      <!-- Link-time optimizations:
+           /LCTG:incremental Enable Fast Link Time Code Generation
+        -->
+      <EnableCOMDATFolding>true</EnableCOMDATFolding>
+      <OptimizeReferences>true</OptimizeReferences>
+    </Link>
+  </ItemDefinitionGroup>
+
+</Project>
--- a/WindowsAppSDK.Build.Cpp.props
+++ b/WindowsAppSDK.Build.Cpp.props
@@ -20,6 +20,9 @@
       <!-- /W4 Enable max warning level -->
       <WarningLevel>Level4</WarningLevel>
 
+      <!-- /sdk Enable additional security checks-->
+      <SDLCheck>true</SDLCheck>
+
       <!-- /GR- Disable RTTI (causes binary bloat) -->
       <RuntimeTypeInfo>false</RuntimeTypeInfo>
     </ClCompile>
@@ -28,14 +31,14 @@
   </ItemDefinitionGroup>
 
   <!-- Debug-specific options -->
-<!--
   <ItemDefinitionGroup Condition="'$(Configuration)' == 'Debug'">
     <ClCompile>
+      <!-- /Ox- Disable most speed optimizations-->
+      <Optimization>Disabled</Optimization>
     </ClCompile>
     <Link>
     </Link>
   </ItemDefinitionGroup>
--->
 
   <!-- Release-specific options -->
   <ItemDefinitionGroup Condition="'$(Configuration)' == 'Release'">
--- a/WindowsAppSDK.Build.Cpp.props
+++ b/WindowsAppSDK.Build.Cpp.props
@@ -33,7 +33,7 @@
   <!-- Debug-specific options -->
   <ItemDefinitionGroup Condition="'$(Configuration)' == 'Debug'">
     <ClCompile>
-      <!-- /Ox- Disable most speed optimizations-->
+      <!-- /d- Disable optimization -->
       <Optimization>Disabled</Optimization>
     </ClCompile>
     <Link>
@@ -72,7 +72,10 @@
     <Link>
       <!-- Link-time optimizations:
            /LCTG:incremental Enable Fast Link Time Code Generation
+           /OPT:ICF Enable COMDAT folding
+           /OPT:REF Optimize references
         -->
+      <LinkTimeCodeGeneration>UseFastLinkTimeCodeGeneration</LinkTimeCodeGeneration>
       <EnableCOMDATFolding>true</EnableCOMDATFolding>
       <OptimizeReferences>true</OptimizeReferences>
     </Link>
--- a/dev/DeploymentAgent/DeploymentAgent.vcxproj
+++ b/dev/DeploymentAgent/DeploymentAgent.vcxproj
@@ -51,7 +51,6 @@
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)'=='Release'" Label="Configuration">
     <UseDebugLibraries>false</UseDebugLibraries>
-    <WholeProgramOptimization>true</WholeProgramOptimization>
     <LinkIncremental>false</LinkIncremental>
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.props" />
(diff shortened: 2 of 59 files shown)


WindowsAppSDK.Build.Cpp.props
```diff
@@ -71,11 +71,11 @@
     </ClCompile>
     <Link>
       <!-- Link-time optimizations:
-           /LCTG:incremental Enable Fast Link Time Code Generation
+           /LCTG Enable Link Time Code Generation
            /OPT:ICF Enable COMDAT folding
            /OPT:REF Optimize references
         -->
-      <LinkTimeCodeGeneration>UseFastLinkTimeCodeGeneration</LinkTimeCodeGeneration>
+      <LinkTimeCodeGeneration>UseLinkTimeCodeGeneration</LinkTimeCodeGeneration>
```
riverar, line 78: Why?

DrusTheAxe: Guidance I'd gotten was /LTCG. I'm not familiar with the details between them. Does FastLTCG produce the same results in less time? Less optimization?

DrusTheAxe:
FYI size reduces are significant, and if you change to favor Size (`/Os` instead of `/Ot`) you get even more savings.

BEFORE: 2,476 K
AFTER (/Ot): 1,971 K  ==> Shrinkage: ~500K (~20%)
AFTER (/Os): 1,747 K  ==> Shrinkage: ~730K (~30%)

Currently favors speed (/Ot) but favoring size (/Os) is under consideration. If desired it's a trivial 1 line change (in this PR or a later PR)

WindowsAppSDK.Build.Cpp.props
riverar: Is my understanding that UseFastLinkTimeCodeGeneration equates to [`/LTCG:Incremental`](https://docs.microsoft.com/cpp/build/reference/ltcg-link-time-code-generation), and that this change will make inner loop dev miserable. (Untested caveat goes here.)

dmachaj: Incremental linking is bad for performance on release builds.  It seems like a fine thing to do for debug builds.  This whole block is already scoped to release builds so disabling incremental LTCG seems reasonable to me.

WindowsAppSDK.Build.Cpp.props
```diff
@@ -0,0 +1,84 @@
...
+    <Link>
+      <!-- Link-time optimizations:
+           /LCTG Enable Link Time Code Generation
+           /OPT:ICF Enable COMDAT folding
+           /OPT:REF Optimize references
+        -->
+      <LinkTimeCodeGeneration>UseLinkTimeCodeGeneration</LinkTimeCodeGeneration>
+      <EnableCOMDATFolding>true</EnableCOMDATFolding>
+      <OptimizeReferences>true</OptimizeReferences>
+    </Link>
```
dmachaj, line 81: >Link [](http://example.com/codeflow?start=6&length=4)

It would probably be worth explicitly disabling incremental linking too [[https://docs.microsoft.com/en-us/cpp/build/reference/incremental-link-incrementally?view=msvc-170](https://docs.microsoft.com/en-us/cpp/build/reference/incremental-link-incrementally?view=msvc-170)](https://docs.microsoft.com/en-us/cpp/build/reference/incremental-link-incrementally?view=msvc-170).  These other flags should already be disabling it but it would be good to guarantee it.

WindowsAppSDK.Build.Cpp.props
```diff
@@ -0,0 +1,84 @@
...
+  <ItemDefinitionGroup>
+    <ClCompile>
+      <!-- /std:c++17 Enable ISO C++17 Standard -->
+      <LanguageStandard>stdcpp17</LanguageStandard>
+
+      <!-- /permissive- Enable Conformance mode (i.e. Disable 'permissive' mode) -->
+      <ConformanceMode>true</ConformanceMode>
+
+      <!-- /W4 Enable max warning level -->
+      <WarningLevel>Level4</WarningLevel>
```
dmachaj, line 21: >WarningLevel [](http://example.com/codeflow?start=7&length=12)

What about /WX?

WindowsAppSDK.Build.Cpp.props
```diff
@@ -0,0 +1,84 @@
...
+      <!-- /Ox Enable most speed optimizations
+           NOTE: /Ox is short-hand for multiple options:
+                     /Ob2 Enable inline function expansion (/Ob2 = Any Suitable)
+                     /Oi Enable intrinsic functions
+                     /Ot Favor fast code
+                     /Oy Omit frame pointers
+        -->
+      <InlineFunctionExpansion>AnySuitable</InlineFunctionExpansion>
+      <IntrinsicFunctions>true</IntrinsicFunctions>
+      <FavorSizeOrSpeed>Speed</FavorSizeOrSpeed>
```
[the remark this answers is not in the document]
BenJKuhn: Yep, size is sometimes net faster in real world situations (but not always). I agree with defaulting to that, and we can do a full evaluation later & compare in some real world tests to see if we want to keep it that way.

dev/DeploymentAgent/DeploymentAgent.vcxproj
```diff
@@ -51,7 +51,6 @@
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)'=='Release'" Label="Configuration">
     <UseDebugLibraries>false</UseDebugLibraries>
-    <WholeProgramOptimization>true</WholeProgramOptimization>
     <LinkIncremental>false</LinkIncremental>
```
dmachaj, line 54: >LinkIncremental [](http://example.com/codeflow?start=28&length=15)

Consider moving this definition to the shared props file.

dev/Detours/Detours.vcxproj
```diff
@@ -154,26 +121,6 @@
...
-      <SubSystem>
-      </SubSystem>
-      <GenerateDebugInformation>true</GenerateDebugInformation>
-    </Link>
-  </ItemDefinitionGroup>
-  <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|ARM'">
-    <ClCompile>
-      <WarningLevel>Level4</WarningLevel>
       <PreprocessorDefinitions>%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <ConformanceMode>true</ConformanceMode>
```
dmachaj, line 125: >ConformanceMode [](http://example.com/codeflow?start=7&length=15)

/permissive- is now handled by the global props

dmachaj: Same for functionlevellinking.

dev/MRTCore/mrt/Core/src/MRM.cpp
```diff
@@ -19,7 +19,7 @@
 
 using namespace Microsoft::Resources;
 
-typedef struct
+typedef struct _MrmObjects
```
dmachaj, line 22: >_MrmObjects [](http://example.com/codeflow?start=15&length=11)

Needed for `/permissive-`?

test/TestApps/PushNotificationsTestApp/main.cpp
```diff
@@ -252,7 +252,7 @@ bool VerifyForegroundHandlerFails()
     try
     {
         // Register is already called in main
-        PushNotificationManager::Default().PushReceived([](const auto&, PushNotificationReceivedEventArgs const& args) {});
+        PushNotificationManager::Default().PushReceived([](const auto&, PushNotificationReceivedEventArgs const& /*args*/) {});
```
dmachaj, line 255: >* [](http://example.com/codeflow?start=114&length=1)

Mixing patterns on the same line.  The first parameter is nameless instead of commented out.

test/TestApps/ToastNotificationsTestApp/ToastNotificationsTestApp.vcxproj
```diff
@@ -196,36 +183,26 @@
...
       <PreprocessorDefinitions>_DEBUG;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <PrecompiledHeaderFile>pch.h</PrecompiledHeaderFile>
-      <LanguageStandard>stdcpp17</LanguageStandard>
     </ClCompile>
     <Link>
       <SubSystem>Console</SubSystem>
       <AdditionalLibraryDirectories>%(AdditionalLibraryDirectories);$(OutDir)\..\WindowsAppRuntime_DLL</AdditionalLibraryDirectories>
       <AdditionalDependencies>onecore.lib;onecoreuap.lib;Microsoft.WindowsAppRuntime.lib;%(AdditionalDependencies)</AdditionalDependencies>
-      <DelayLoadDLLs>Microsoft.WindowsAppRuntime.Bootstrap.dll</DelayLoadDLLs>
+      <DelayLoadDLLs>Microsoft.WindowsAppRuntime.Bootstrap.dll;%(DelayLoadDLLs)</DelayLoadDLLs>
```
dmachaj, line 194: >DelayLoadDLLs [](http://example.com/codeflow?start=65&length=13)

Good find.  Missing delayloads is bad for perf.

WindowsAppSDK.Build.Cpp.props
DrusTheAxe: That's Foundation's model - folks generally use Debug as it's debuggable (without trying to follow the crazy instruction pointer). Occasionally when someone needs Release (benchmarking etc) they bite the bullet, and the project's not that big that it's painful.

Plus devs can always locally hack settings for local use (just don't checkin).

WindowsAppSDK.Build.Cpp.props
DrusTheAxe: Good idea. It does get disabled but a few projects emit messages about disabling it because of some other setting. Explicit would prevent those messages, so yay. Thanks

WindowsAppSDK.Build.Cpp.props
DrusTheAxe: Not yet. There's a handful of warnings (though you have to Rebuild to see some) but we're into the complex-weird territory (props/targets related, not source code).

/WX is expected in the foreseeable future

WindowsAppSDK.Build.Cpp.props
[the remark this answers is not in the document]
DrusTheAxe: Done! :-)

dev/Detours/Detours.vcxproj
DrusTheAxe: Did a full pass removing all redundancies

DrusTheAxe:
> RemoveUnreferencedCodeData

Detours is intentionally NOT mucked with,. Detours.vcxproj is an MSBuild variant of Detours' makefiles, and Detours has a bunch of low level fiddly presumptions so 2nd best is to not mess with it at all. I've added a comment at the top of Detours.vcxproj to that effect.

1st best is to get the Detours project to provide a nuget. To date they don't provide a nuget and [the Issue requesting one](https://github.com/microsoft/Detours/issues/124) is still open. Until then, Detours.vcxproj is the exception proving the rule.

test/TestApps/ToastNotificationsTestApp/ToastNotificationsTestApp.vcxproj
DrusTheAxe: It's only a test project but yes, missing could be bad.

Worse IMO are the inconsistencies across Configration+Platform. We have 6 permutations. I've found several that only set options for 5, or 1, or somewhere in between. Fixed that up, for any options I messed with. May still have some inconsistencies in other properties.

I've had 'sanitize projects' in my wishlist for ages, and incrementally get closer. There's more to do but only so many hours in a do, so today is better than yesterday, and tomorrow will be better still :-) Just so long as nothing egregious pops up we can get to it incrementally as time permits.

test/TestApps/PushNotificationsTestApp/main.cpp
DrusTheAxe: The latter was a compiler warning at /W4 (was /W3) so I fixed it. I didn't notice the missing parameter. Is that supposed to be "sender"?

dev/MRTCore/mrt/Core/src/MRM.cpp
DrusTheAxe: Yep. The classic Windows style is `typedef struct _Foo { ... } Foo;`.

Better would be simply `struct Foo {...};` since it's just C++ code, not part of a Flat-C API for C consumption. But I was hesitant to make deep changes, and there's other fixes needed in here anyway so I tried to be as minimally invasive as I could and leave a deeper cleanup refactoring to the MRT team follow up

dev/DeploymentAgent/DeploymentAgent.vcxproj
DrusTheAxe: Already did :-)

test/TestApps/PushNotificationsTestApp/main.cpp
dmachaj: Most likely.  To quiet this particular warning you can either comment out the name of the variable (so there is nothing to be unused) or you can delete the name entirely (same effect).  Both are valid options.  I prefer the commented out name in most cases.  My nitpick was just that this line is mixing both patterns instead of sticking to one.

dmachaj (approved):
:shipit:

test/TestApps/PushNotificationsTestApp/main.cpp
DrusTheAxe: I prefer commenting out too. 'Cause today's unused parameter is tomorrow's PITA need-to-use-what-was-the-bloody-thing-called...

My inner OCD rose to the occasion so it's not commented out for consistency :-)

PR #2640, opened 2022-06-17
Fix the .NET SDK installer script used by MRT Core, and don't run the script anymore

jefgen:
This is a follow-up to PR #1916.

The old `versions.props` file used by the script used to have two PropertyGroup XML nodes, and the new file only has one. This means that script ends up *not* getting the version from the new `Versions.props` file, which ultimately means that nothing actually gets installed. (Overall the script is a no-op).

We also need to add a check on `$dotNetSdkVersionLkg` as in the script, as within the `Versions.props` file it's defined to be `$(CsWinRTDependencyDotNetCoreSdkPackageVersion)`. But this doesn't get interpreted by PowerShell, so the string exists verbatim. This later causes the `dotnet-install.ps1` to fail, as it wants a real version number. But if its set to `$(CsWinRTDependencyDotNetCoreSdkPackageVersion)` we should just skip the second install step, since it's the same version.

We also don't actually need to run the `DownloadDotNetCoreSdk.ps1` script in the pipeline anymore, as the .NET SDK is already installed.

### DIFF
--- a/build/build-mrt.yml
+++ b/build/build-mrt.yml
@@ -78,19 +78,19 @@ steps:
 #      PathtoPublish: '${{ parameters.MRTSourcesDirectory }}\msbuild-install-logs'
 #      artifactName: 'installlogs'
 
-- task: powershell@2
-  displayName: 'Installing .NET SDK'
-  inputs:
-    targetType: filePath
-    workingDirectory: ${{ parameters.MRTSourcesDirectory }}\build
-    filePath: ${{ parameters.MRTSourcesDirectory }}\build\DownloadDotNetCoreSdk.ps1
+# - task: powershell@2
+#   displayName: 'Installing .NET SDK'
+#   inputs:
+#     targetType: filePath
+#     workingDirectory: ${{ parameters.MRTSourcesDirectory }}\build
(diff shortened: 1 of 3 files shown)


axelandrejs (approved):

huichen123:
we should not attempt to install if newer version is already installed.

jefgen:
> we should not attempt to install if newer version is already installed.

Sure, but I don't think that will fix the PreFast build though... as it seems to want to use the *exact* same version of the .NET SDK in the global.json file...

huichen123:
> > we should not attempt to install if newer version is already installed.
> 
> Sure, but I don't think that will fix the PreFast build though... as it seems to want to use the _exact_ same version of the .NET SDK in the global.json file...

for the prefast issue, we can remove sdk from global.json, to be inline with everyone else.

jefgen:
> > > we should not attempt to install if newer version is already installed.
> > 
> > Sure, but I don't think that will fix the PreFast build though... as it seems to want to use the _exact_ same version of the .NET SDK in the global.json file...
> 
> for the prefast issue, we can remove sdk from global.json, to be inline with everyone else.

Sure, I can try that in another branch, and see if that works...

huichen123 (approved):

PR #2658, opened 2022-06-22
[AppLifecycle] Properly ignore cleanup event if it can't be opened.

jlaanstra:
If we can't open the cleanup event, ignore it.

Fixes ADO#38606299


jlaanstra:
@dhoehna Can we merge this?

PR #2686, opened 2022-06-28
Fix 'noexcept' C compile errors

AzAgarampur:
Allows bootstrapper APIs to be used from C without a compile-time error complaining about 'noexcept'.


DrusTheAxe:
Cool. Fair ask, though the change needs the source to match the headers or build errors making WinAppSDK. See [Remove noexcept from Bootstrapper API so it can be used from C #2750](https://github.com/microsoft/WindowsAppSDK/pull/2750).

Thanks for the feedback!

Closing this PR as it's superseded by #2750

PR #2723, opened 2022-07-11
WindowsAppRuntimeInstaller.exe fails if newer version present

DrusTheAxe:
Discussion #2708: [WindowsAppRuntime installation error: The package could not be installed because a higher version of this package is already installed](https://github.com/microsoft/WindowsAppSDK/discussions/2708)

>Microsoft.WindowsAppRuntime.1.1_1000.516.2156.0_x64__8wekyb3d8bbwe Package deployment result : 0x80073d06 The package could not be installed because a higher version of this package is already installed. One or more install operations failed. Result: 0x80073d06

That's a bug - it should be OK, not an error.

If v1.1.3 is installed and you run the installer for v1.1.2 we call packageManager.AddPackageAsync(...v1.1.2...) which returns 0x80073d06 == `ERROR_INSTALL_PACKAGE_DOWNGRADE` which is technically an error but functionally it's OK. If you build with 1.1.2 we don't require the runtime packages exactly match 1.1.2, but rather that there needs to be a package equal-or-higher than 1.1.2 (i.e. `>=1.1.2` not `==1.1.2`). If 1.1.3 is installed then all is well.

Problem is we should be trapping `ERROR_INSTALL_PACKAGE_DOWNGRADE` and treating it as success. That's a bug on our part. 

https://task.ms/40411617

FYI: servicing-consider-1.1 = https://task.ms/40411679

### DIFF
--- a/installer/dev/install.cpp
+++ b/installer/dev/install.cpp
@@ -62,6 +62,11 @@ namespace WindowsAppRuntimeInstaller
                 RETURN_IF_FAILED(RegisterPackage(packageProperties->fullName.get()));
                 return S_OK;
             }
+            else if (hrAddPackage == ERROR_INSTALL_PACKAGE_DOWNGRADE)
+            {
+                // Higher version of the package already exists so we're good! Nothing to do!
+                return S_OK;
+            }
             else
             {
                 const auto deploymentResult{ deploymentOperation.GetResults() };


installer/dev/install.cpp
```diff
@@ -62,6 +62,11 @@ namespace WindowsAppRuntimeInstaller
                 RETURN_IF_FAILED(RegisterPackage(packageProperties->fullName.get()));
                 return S_OK;
             }
+            else if (hrAddPackage == ERROR_INSTALL_PACKAGE_DOWNGRADE)
+            {
+                // Higher version of the package already exists so we're good! Nothing to do!
```
sachintaMSFT, line 67: Should we not inform user about it, in console mode ?

sachintaMSFT (changes requested):

installer/dev/install.cpp
DrusTheAxe: > Should we not inform user about it, in console mode ?

Methinks not.

StagePackageAsync() doesn't return `ERROR_INSTALL_PACKAGE_DOWNGRADE`, it just silently succeeds. Thus Stage(1.2.3) doesn't mean 'install 1.2.3' it means 'install 1.2.3 _or higher_ if available' (aka "Version Supercedence").

Frankly I was surprised Add() differs from Stage(). We're discussing treating that as a bug and thus updating PackageManager.Add*() to act the same as Stage. It's a lot more useful this way than complaining about 'attempted downgrade'.

So no, I think we should treat Add like Stage (whether or not the OS implements it that way - yet) and it's merely a documented behavior.

riverar: Does that take into account scenarios where developers bootstrap a specific version and can't use/don't want newer bits?

BenJKuhn: If using the Windows App Runtime MSIX package, you're implicitly opting into using latest servicing update of the package. If you want *exact* bits only, that's what self-contained is for, and is the only way to version-lock. This code path doesn't intersect with the self-contained scenario.

riverar: That does not align with my understanding. Before self-contained came along, WAS published a Bootstrap API that let developers choose what version to anchor to. Are you saying that is no longer the case?

DrusTheAxe: > WAS published a Bootstrap API that let developers choose what version to anchor to

Sorry Rafael but that's not the case. Never has been.

```
STDAPI MddBootstrapInitialize(
    UINT32 majorMinorVersion,
    PCWSTR versionTag,
    PACKAGE_VERSION ***minVersion***)
```

That parameter's called `minVersion`, not `version` or `exactVersion`. You can specify the floor (minimum) version, but higher versions are equally valid, and preferred

dev/DynamicDependency/API/PackageGraph.cpp line 136+

```c++
MddCore::PackageGraph::ResolvePackageDependency(
HRESULT MddCore::PackageGraph::ResolvePackageDependency(
    const MddCore::PackageDependency& packageDependency,
    MddAddPackageDependencyOptions /*options*/,
    wil::unique_process_heap_string& packageFullName)
{
...
    // Find the best fit package in the family (if any)
...
        // Do we already have a higher version under consideration?
        if (bestFit && (bestFit.Version().Version > candidate.Version().Version))
        {
            continue;
        }

        // Package version must meet the minVersion filter
        if (candidate.Version().Version < packageDependency.MinVersion().Version)
```

As spec'd in [3.1.3.1. Resolve Package in DynamicDependencies.md](https://github.com/microsoft/WindowsAppSDK/blob/main/specs/dynamicdependencies/DynamicDependencies.md#3131-resolve-package)

This is nothing new. MSIX has always been biased towards equal-or-higher version matching e.g. `<TargetDeviceFamily MinVersion=x...>` in appxmanifest.xml defines the 'minimum acceptable', there is no 'maximum acceptable' i.e. it's a greater-or-equal match, not equality.

riverar: I'm thinking of the _product_ version (first param), if this PR doesn't affect that then disregard!

Example:
```rust
MddBootstrapInitialize2(
    0x00010001, // 1.1
    WINDOWSAPPSDK_RELEASE_VERSION_TAG_W,
    min_framework_version,
    MddBootstrapInitializeOptions_OnNoMatch_ShowUI,
)
```

DrusTheAxe: Ahhh

This PR doesn't affect that 🙂

The release matching is governed by the release major.minor, which we embed in the package Name (and thus PackageFamilyName) i.e. 1.1.* installers only see, care about and affect release 1.1 packages. So yes you're correct, you 'anchor' to a release. Within that release the specific package version is bounded at the floor only.

riverar: Great, thanks. Glad this turned into a no-op.

cwruss (approved):

DominicMaas:
I think this has regressed.

I get this error when trying to install 1.1.4 on a system with 1.1.5

```
Installing license: MSIX_MAINPACKAGE_LICENSE
Install result : 0x0
Installing license: MSIX_SINGLETONPACKAGE_LICENSE
Install result : 0x0
Deploying package: Microsoft.WindowsAppRuntime.1.1_1004.584.2120.0_x64__8wekyb3d8bbwe
Package deployment result : 0x80073d06 The package could not be installed because a higher version of this package is already installed.
ErrorMessage: Windows cannot install package Microsoft.WindowsAppRuntime.1.1_1004.584.2120.0_x64__8wekyb3d8bbwe because it has version 1004.584.2120.0. A higher version 1005.616.1651.0 of this package is already installed.One or more install operations failed. Result: 0x80073d06
```

I'll open an issue.

PR #2734, opened 2022-07-14
XCOPY command should be <COPY> to ensure execution

DrusTheAxe:
Minor change `<Command>xcopy...` to `<Copy>...` to ensure execution (buttons up a rare problem)


DrusTheAxe:
> Or this should really just use the more appropriate [Copy task](https://docs.microsoft.com/en-us/visualstudio/msbuild/copy-task). 😇

How's it look now? :-)

alexlamtest (approved):

PR #2750, opened 2022-07-19
Remove noexcept from Bootstrapper API so it can be used from C

DrusTheAxe:
As requested by [Fix 'noexcept' C compile errors #2686](https://github.com/microsoft/WindowsAppSDK/pull/2686)

### DIFF
--- a/dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
+++ b/dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
@@ -89,7 +89,7 @@ static std::wstring g_test_mainPackageNamePrefix;
 STDAPI MddBootstrapInitialize(
     UINT32 majorMinorVersion,
     PCWSTR versionTag,
-    PACKAGE_VERSION minVersion)
+    PACKAGE_VERSION minVersion) noexcept
 {
     return MddBootstrapInitialize2(majorMinorVersion, versionTag, minVersion, MddBootstrapInitializeOptions_None);
 }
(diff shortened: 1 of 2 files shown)


DrusTheAxe:
> Consider a model like WIL, which uses a `#define` in the header. So you'd write something like:
> 
> ```c
> #if defined(__cplusplus)
> #define MDD_NOEXCEPT noexcept
> #else
> #define MDD_NOEXCEPT
> 
> STDAPI MddBootstrapInitialize(
>     UINT32 majorMinorVersion,
>     PCWSTR versionTag,
>     PACKAGE_VERSION minVersion) MDD_NOEXCEPT;
> ```
> 
> ... which lets you retain noexcept-ness for C++ and drop it for C (in the header; the implementation should still say noexcept since it should be.)

Is this the standard pattern we're recommending?

Does Windows SDK do this for its Flat-C Win32 APIs?

jonwis:
It's a pattern that appears to work, especially for same-source-multiple-compiler-targets. See also decades of headers that adapt to gcc, G++, EDG, msvc, etc. using macros keyed of the target environment. WIL uses this for flexibility between C++ language levels and compilers.  

The Windows PSDK presumes its headers are consumable from C and assembly, so the headers provide an ABI-stable definition of the platform's exports.  There are certain areas (like nano-COM) that deviate from this and use C++-isms (overloads, templates, noexcept) but many of those also have adaptive markers like DECLARE_INTERFACE and pals.  Or said another way, the Windows PSDK is all "flat C" APIs that cannot declare a noexcept marker on them.

See also https://docs.microsoft.com/en-us/cpp/build/reference/eh-exception-handling-model?view=msvc-170 where you can tell it "all extern C methods are nothrow" ... there's no _guarantee_ that everybody uses it. Awareness of noexcept reduces consuming-binary size.

In this case, I'd use the macro model so a C consumer can use the headers we produce with the DLL exports in the package.

DrusTheAxe:
> In this case, I'd use the macro model so a C consumer can use the headers we produce with the DLL exports in the package.

OK. Thanks Jon.

jonwis (approved):

PR #2751, opened 2022-07-19
Put AccessControl files into transport package

DefaultRyan:
The runtime support for this feature shipped in 1.1, but the changes to put the winmd and headers in the nuget were missed. This fixes it.


jonwis:
Future work - add a sample using the APIs so we'd notice they were missing.

jonwis (approved):

PR #2753, opened 2022-07-20
User/sachinta/fix DeploymentManager's GetStatus

sachintaMSFT:
DeploymentManager GetStatus is searching for Framework package name's prefix (Microsoft.WindowsAppRuntime) to build PackageFamilyNames of Main and Singleton packages before verifying they are installed on the machine. It doesn't work as both Main and Singleton share a different package name prefix (i.e. MicrosoftCorporationII.WinAppRuntime) than that of Framework package prefix


dev/Deployment/DeploymentManager.cpp
```diff
@@ -105,12 +105,12 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
             if (package.versionType == PackageVersionType::Versioned)
             {
                 // Prefix + SubTypeName + VersionIdentifier + Suffix
```
cwruss, line 107: Please consider updating the comments here and above to detail what GetStatus is looking for in terms of verifying the Singleton and Main packages etc.

cwruss (approved):

PR #2760, opened 2022-07-21
AppNotificationContent Builder Core Features - part II

loneursid:
Add TextBox and ComboBox support to the AppNotificationBuilder


dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -280,11 +307,29 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
         return wil::str_printf<std::wstring>(L"%ws%ws%ws", m_inlineImage.c_str(), m_heroImage.c_str(), m_appLogoOverride.c_str());
     }
 
-    std::wstring AppNotificationBuilder::GetButtons()
+    std::wstring AppNotificationBuilder::GetActions()
```
loneursid, line 310: We have an opportunity here to do some extra validation that goes beyond what the xml schema prescribes. For example, we couold ensure that a input-hint in a button elements, refers an existing texbox.

dev/AppNotifications/AppNotificationBuilder/AppNotificationComboBox.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+        return items;
+    }
+
+    winrt::hstring AppNotificationComboBox::ToString()
+    {
+        std::wstring xmlResult{ wil::str_printf<std::wstring>(L"<input id='%ls' type='selection'%ls%ls>%ls</input>",
+            m_id.c_str(),
+            m_title.empty() ? L"" : wil::str_printf<std::wstring>(L" title='%ls'", m_title.c_str()).c_str(),
+            m_selectedItem.empty() ? L"" : wil::str_printf<std::wstring>(L" defaultInput='%ls'", m_selectedItem.c_str()).c_str(),
```
loneursid, line 54: We have an opportunity here to do some extra validation that goes beyond what the xml schema prescribes. For example, we could ensure that the value for SelectedITem refers an existing selection. That's probably something worth exploring in a future iteration.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -194,6 +196,22 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
...
+
+        m_textBoxList.push_back(TextBox{ id });
+        return *this;
+    }
+
+    winrt::Microsoft::Windows::AppNotifications::Builder::AppNotificationBuilder AppNotificationBuilder::AddTextBox(hstring id, hstring placeHolderText, hstring title)
+    {
+        ThrowIfMaxInputItemsExceeded();
+
+        m_textBoxList.push_back(TextBox{ id, true, placeHolderText, title });
```
pmpurifoy, line 211: I recommend converting m_textBoxList to an std::vector<std::wstring> and constructing the input element here. This way, you can remove the TextBox struct completely and just call append in GetActions since you already have the constructed element.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -194,6 +196,22 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
         return *this;
     }
 
+    winrt::Microsoft::Windows::AppNotifications::Builder::AppNotificationBuilder AppNotificationBuilder::AddTextBox(hstring id)
+    {
+        ThrowIfMaxInputItemsExceeded();
+
+        m_textBoxList.push_back(TextBox{ id });
```
pmpurifoy, line 203: Recommend building the input element here. Explanation below this.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -280,11 +307,29 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
...
+        if (m_buttonList.empty() && m_textBoxList.empty() && m_comboBoxList.empty())
+        {
+            return {};
+        }
+        else
         {
             std::wstring result{};
+            for (auto input : m_textBoxList)
+            {
+                auto placeHolderTextAndTitle = input.hasPlaceHolderTextAndTitle
```
pmpurifoy, line 321: Recommend moving this element construction to AddTextBox(...)

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.h
```diff
@@ -59,13 +64,22 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
...
         std::wstring GetDuration();
         std::wstring GetScenario();
         std::wstring GetArguments();
         std::wstring GetButtonStyle();
         std::wstring GetText();
         std::wstring GetImages();
-        std::wstring GetButtons();
+        std::wstring GetActions();
+
+        struct TextBox
```
pmpurifoy, line 76: I think we can remove this if we construct the element inside the AddTextBox(...) functions.

test/AppNotificationBuilderTests/APITests.cpp
```diff
@@ -382,5 +382,118 @@ namespace Test::AppNotification::Builder
...
+        {
+            auto comboBox{ AppNotificationComboBox(L"comboBox1")
+                    .AddItem(L"item1", L"item1 text")
+                    .AddItem(L"item2", L"item2 text")
+                    .AddItem(L"item3", L"item3 text")
+                    .AddItem(L"item4", L"item4 text")
+                    .AddItem(L"item5", L"item5 text") };
+            auto expected{ L"<input id='comboBox1' type='selection'><selection id='item1' content='item1 text'/><selection id='item2' content='item2 text'/><selection id='item3' content='item3 text'/><selection id='item4' content='item4 text'/><selection id='item5' content='item5 text'/></input>" };
+
+            VERIFY_ARE_EQUAL(comboBox.as<winrt::Windows::Foundation::IStringable>().ToString(), expected);
```
pmpurifoy, line 480: Recommend using builder.BuildNotification to test developer flow.

test/AppNotificationBuilderTests/APITests.cpp
```diff
@@ -382,5 +382,118 @@ namespace Test::AppNotification::Builder
...
+        }
+
+        TEST_METHOD(AppNotificationComboBoxAddTwoSelectionItemsWithSameId)
+        {
+            auto comboBox{ AppNotificationComboBox(L"comboBox1")
+                    .AddItem(L"item1", L"item1 text")
+                    .AddItem(L"item1", L"item2 text") };
+            auto expected{ L"<input id='comboBox1' type='selection'><selection id='item1' content='item2 text'/></input>" };
+
+            VERIFY_ARE_EQUAL(comboBox.as<winrt::Windows::Foundation::IStringable>().ToString(), expected);
```
pmpurifoy, line 496: Recommend using builder.BuildNotification to test developer flow.

pmpurifoy (changes requested):

test/AppNotificationBuilderTests/APITests.cpp
loneursid: Notr everything needs to go through the AppNotificationBuilder. These are unit tests, and the smaller they are, the easier it is to maintain and debug them.

Some already go through the AppNotificationBuilder to validate that the content gets inserted correctly. Others are just verifying the class works as per the specs.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -16,6 +16,8 @@ using namespace winrt::Windows::Globalization::DateTimeFormatting;
 
 namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
 {
+    void AppNotificationBuilder::ThrowIfMaxInputItemsExceeded() { THROW_HR_IF_MSG(E_INVALIDARG, m_textBoxList.size() + m_comboBoxList.size() >= c_maxTextInputElements, "Maximum number of input elements added"); };
```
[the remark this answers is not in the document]
loneursid: Yup, I thought I had fixed that in a previous commit but it seems it fell through the crack, somehow...

test/AppNotificationBuilderTests/APITests.cpp
loneursid: Some tests do just that but they don;t all need to do that. It just makes the tests more complex and harder to read / maintain.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.h
loneursid: I think you're right. That would be useful to do extra validation at build time (for example, to ensure that the input hint on a button, actually match a textbox), but that something we can re-add later, if we ever revisit the builder.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
loneursid: done

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
loneursid: done

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
loneursid: done

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.h
loneursid: done

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -195,6 +200,22 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
         return *this;
     }
 
+    winrt::Microsoft::Windows::AppNotifications::Builder::AppNotificationBuilder AppNotificationBuilder::AddTextBox(hstring id)
+    {
+        ThrowIfMaxInputItemsExceeded();
```
hulumane, line 205: Should we check argument to see if it is empty and throw?

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -203,6 +224,15 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
...
+
+        m_comboBoxList.push_back(value);
+
+        return *this;
+    }
+
     winrt::Microsoft::Windows::AppNotifications::Builder::AppNotificationBuilder AppNotificationBuilder::SetTag(hstring const& value)
     {
         m_tag = value;
```
hulumane, line 238: @pmpurifoy I think we need to make sure all these Set*(Hstring) APIs don't take in empty strings. They can be problematic. At the same time, m_tag and m_group should default to empty if they havent been set already

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -281,11 +311,25 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
         return wil::str_printf<std::wstring>(L"%ls%ls%ls", m_inlineImage.c_str(), m_heroImage.c_str(), m_appLogoOverride.c_str());
     }
 
-    std::wstring AppNotificationBuilder::GetButtons()
+    std::wstring AppNotificationBuilder::GetActions()
     {
-        if (m_buttonList.size())
+        if (m_buttonList.empty() && m_textBoxList.empty() && m_comboBoxList.empty())
```
hulumane, line 316: Instead of having this if condition here, we could remove this and do the following below.

return (result.empty()) ? {} : wil::str_printf<std::wstring>(L"<actions>%ls</actions>", result.c_str());

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -316,7 +356,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
         xmlResult.reserve(c_maxAppNotificationPayload);
 
         // Build the button string and fill m_useButtonStyle
-        std::wstring buttons{ GetButtons() };
+        std::wstring actions{ GetActions() };
```
hulumane, line 359: Would be awesome to have a comment on how a final string would look like with the payload.

dev/AppNotifications/AppNotificationBuilder/AppNotificationComboBox.cpp
```diff
@@ -0,0 +1,59 @@
...
+#include "Microsoft.Windows.AppNotifications.Builder.AppNotificationComboBox.g.cpp"
+#include "AppNotificationBuilderUtility.h"
+
+namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
+{
+    AppNotificationComboBox::AppNotificationComboBox(hstring const& id) : m_id(id) { };
+
+    winrt::Microsoft::Windows::AppNotifications::Builder::AppNotificationComboBox AppNotificationComboBox::AddItem(winrt::hstring const& id, winrt::hstring const& content)
+    {
+        THROW_HR_IF_MSG(E_INVALIDARG, m_items.Size() >= c_maxSelectionElements, "Maximum number of items added");
```
hulumane, line 15: We should make sure that id and content are not empty here

dev/AppNotifications/AppNotificationBuilder/AppNotificationComboBox.cpp
```diff
@@ -0,0 +1,59 @@
...
+        THROW_HR_IF_MSG(E_INVALIDARG, id.empty(), "You must provide an id when adding an argument");
+
+        m_items.Insert(id, content);
+
+        return *this;
+    }
+
+    winrt::Microsoft::Windows::AppNotifications::Builder::AppNotificationComboBox AppNotificationComboBox::SetTitle(winrt::hstring const& value)
+    {
+        m_title = value;
```
hulumane, line 25: Please make sure that value isn't empty here

dev/AppNotifications/AppNotificationBuilder/AppNotificationComboBox.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+        return *this;
+    }
+
+    winrt::Microsoft::Windows::AppNotifications::Builder::AppNotificationComboBox AppNotificationComboBox::SetSelectedItem(winrt::hstring const& id)
+    {
+        m_selectedItem = id;
```
hulumane, line 32: id shouldnt be empty here

dev/AppNotifications/AppNotificationBuilder/AppNotificationComboBox.cpp
```diff
@@ -0,0 +1,59 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#include "pch.h"
+#include "AppNotificationComboBox.h"
+#include "Microsoft.Windows.AppNotifications.Builder.AppNotificationComboBox.g.cpp"
+#include "AppNotificationBuilderUtility.h"
+
+namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
+{
+    AppNotificationComboBox::AppNotificationComboBox(hstring const& id) : m_id(id) { };
```
hulumane, line 11: id shouldn't be empty here

dev/AppNotifications/AppNotificationBuilder/AppNotificationComboBox.h
```diff
@@ -0,0 +1,48 @@
...
+
+        winrt::Microsoft::Windows::AppNotifications::Builder::AppNotificationComboBox SetTitle(winrt::hstring const& value);
+
+        winrt::Microsoft::Windows::AppNotifications::Builder::AppNotificationComboBox SetSelectedItem(winrt::hstring const& id);
+
+        // IStringable
+        winrt::hstring ToString();
+
+    private:
+        winrt::hstring m_id;
```
hulumane, line 34: Does winrt::hstring default to empty?

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -195,6 +200,22 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
...
+        ThrowIfMaxInputItemsExceeded();
+
+        m_textBoxList.push_back(wil::str_printf<std::wstring>(L"<input id='%ls' type='text'/>", id.c_str()));
+        return *this;
+    }
+
+    winrt::Microsoft::Windows::AppNotifications::Builder::AppNotificationBuilder AppNotificationBuilder::AddTextBox(hstring id, hstring placeHolderText, hstring title)
+    {
+        ThrowIfMaxInputItemsExceeded();
+
```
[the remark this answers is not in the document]
pmpurifoy: The only required param is id. We should prevent the placeholder/title from being added if empty, but the only time we should throw is if id is empty. 

Main example is developer may only want placeholder OR title instead of both.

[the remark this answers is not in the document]
hulumane: But the string formatting below goes and adds empty title andplaceholdertext as an attribution. Why have the attribute at all if they are empty?

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
hulumane: Me thinks yes :) We havent done any argument validation for string types at all in the builder. We probably should.

dev/AppNotifications/AppNotificationBuilder/AppNotificationComboBox.h
pmpurifoy: Yup. Just tested this.

loneursid: Added empty initializers anyways as this is what the other classes in this module do.

dev/AppNotifications/AppNotificationBuilder/AppNotificationComboBox.cpp
loneursid: The is a check for ID not being empty on the line bellow.

Whether content should be left empty or not should be addressed when we get to this: https://microsoft.visualstudio.com/OS/_workitems/edit/40746502

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
loneursid: done

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
loneursid: Are empty string invalid in the xml schema? The goal for the builder at this point is to produce valid xml and keep the validation to the moinimum required to do so.

there's a deliverable that captures the plans to increase validation: https://microsoft.visualstudio.com/OS/_workitems/edit/40746502

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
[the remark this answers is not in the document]
loneursid: Is adding empty string a problem? Is that something that the xml schema would prohibit?

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
loneursid: done

dev/AppNotifications/AppNotificationBuilder/AppNotificationComboBox.cpp
loneursid: Captured in this deliverable https://microsoft.visualstudio.com/OS/_workitems/edit/40746502.

dev/AppNotifications/AppNotificationBuilder/AppNotificationComboBox.cpp
loneursid: done

dev/AppNotifications/AppNotificationBuilder/AppNotificationComboBox.cpp
loneursid: done

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
loneursid: Captured here: https://microsoft.visualstudio.com/OS/_workitems/edit/40746502

dev/AppNotifications/AppNotificationBuilder/AppNotificationComboBox.cpp
loneursid: Captured here: https://microsoft.visualstudio.com/OS/_workitems/edit/40746502

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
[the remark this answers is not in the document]
loneursid: If adding an empty string isn't illegal, then the work can be addressed when we get to it: https://microsoft.visualstudio.com/OS/_workitems/edit/40746502

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
loneursid: //  <toast launch="action=AppNotificationClick;sequence={}">
    //    <visual>
    //      <binding template="ToastGeneric">
    //        <image placement="appLogoOverride" hint-crop="circle" src="image.png"/>
    //        <text>App Notification with Avatar Image< / text>
    //        <text>This is an example message using XML< / text>
    //      </binding>
    //    </visual>
    //    <actions>
    //      <input id="ComboBox" type="selection" defaultInput="yes">
    //        <selection id="yes" content="Going"/>
    //        <selection id="maybe" content="Maybe"/>
    //        <selection id="no" content="Decline"/>
    //      </input>
    //      <input id="textBox" type="text" placeHolderContent="reply"/>
    //      <action
    //        content="Send"
    //        imageUri="ms-appx://images/Icon.png"
    //        hint-inputId="textBox"
    //        arguments="action=reply&amp;threadId=92187"/>
    //      <action
    //        content="Open App"
    //        arguments="action=OpenApp&amp;sequence={}">
    //    </actions>
    //  </toast>

loneursid: I've tried to cobble something together with an image, some text, a combobox, a textbox and two buttons and I'm not convainced this would be useful.

I think i't better/easier to lookup the schema online when working in the builder.

I'm also concerned that this kind of comment will get out of sync quickly and become a maintenance nightmare.

I could create a unittest that creates an AppNotification simillar to this example comment. At least we know that this would never get out of sync with the code.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -195,6 +200,25 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
         return *this;
     }
 
+    winrt::Microsoft::Windows::AppNotifications::Builder::AppNotificationBuilder AppNotificationBuilder::AddTextBox(hstring id)
+    {
+        ThrowIfMaxInputItemsExceeded();
+        THROW_HR_IF_MSG(E_INVALIDARG, id.empty(), "You must provide an id for the TextBox");
+
+
+        m_textBoxList.push_back(wil::str_printf<std::wstring>(L"<input id='%ls' type='text'/>", id.c_str()));
```
DrusTheAxe, line 209: Is ID restricted to never allow standard entities that need to be escaped?

standard entities are the gang of 5 -- greater-than, less-than, apostraphe, quote, ampersand

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -281,30 +314,34 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
...
-            return {};
+            if (input.ButtonStyle() != AppNotificationButtonStyle::Default)
+            {
+                m_useButtonStyle = true;
+            }
+
+            result.append(input.as<winrt::Windows::Foundation::IStringable>().ToString().c_str());
         }
+
+        return (result.empty()) ? result : wil::str_printf<std::wstring>(L"<actions>%ls</actions>", result.c_str());
```
DrusTheAxe, line 341: NIT: `return result.empty() ? ...`

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
DrusTheAxe: Examples in the spec and documentation are welcome (and encouraged).

Not so useful as a comment.

Maybe useful as a doc-comment, if it can be small and meaningful.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.vcxitems
```diff
@@ -19,12 +19,14 @@
...
     <ClCompile Include="$(MSBuildThisFileDirectory)AppNotificationTextProperties.cpp" />
   </ItemGroup>
   <ItemGroup>
     <ClInclude Include="$(MSBuildThisFileDirectory)AppNotificationBuilder.h" />
     <ClInclude Include="$(MSBuildThisFileDirectory)AppNotificationBuilderUtility.h" />
     <ClInclude Include="$(MSBuildThisFileDirectory)AppNotificationButton.h" />
+    <ClInclude Include="$(MSBuildThisFileDirectory)AppNotificationComboBox.h" />
     <ClInclude Include="$(MSBuildThisFileDirectory)AppNotificationTextProperties.h" />
   </ItemGroup>
 </Project>
```
DrusTheAxe, line 32: NIT: GitHub dislikes files ending without a new line

DrusTheAxe (approved):

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
pmpurifoy: Looks like it doesn't like " & <, but also it doesn't mind empty string. I am introducing an escape character map in my PR :https://github.com/microsoft/WindowsAppSDK/pull/2805 @loneursid. We can add these to this map when my PR merges.

pmpurifoy: We are using percent encoding for our special characters: https://developer.mozilla.org/en-US/docs/Glossary/percent-encoding

Go ahead and merge this PR when ready, I'll raise another PR regarding this issue since it would require more of a sweeping change in AppNotificationBuilder.

loneursid: Yeah, I'll do that. I'll merge this as is, since it's approved and the build is passing and raise a new PR to address nits from this and other PRs in the series

pmpurifoy (approved):

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
pmpurifoy: List of xml escape chars:
"   &quot;
'   &apos;
<   &lt;
>   &gt;
&   &amp;

loneursid: Addressing in: https://github.com/microsoft/WindowsAppSDK/pull/2813

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
loneursid: Fixed in https://github.com/microsoft/WindowsAppSDK/pull/2813

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.vcxitems
loneursid: Done: https://github.com/microsoft/WindowsAppSDK/pull/2813

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
loneursid: Thanks @DrusTheAxe for catching this. It totally slipped past me.

Since we put anything that comes to us through the builder API into single quotes, technically, that the only character we need to escape, right? Or am I missing something?

PR #2765, opened 2022-07-21
Adding copyright headers to required files and enforcing requirement in build pipelines

EricJohnson327:
cs/cpp/h/idl/xaml files all need copyright headers.  This change adds the required copyright information and adds a script VerifyCopyrightHeaders.ps1.  This script is run at the beginning of builds to verify all checked in files are marked correctly.  The script can also be ran locally to update any needed files to make it easier for users adding new files via "VerifyCopyrightHeaders.ps1 -Fix"

### DIFF
--- a/build/ProjectReunion-BuildFoundation.yml
+++ b/build/ProjectReunion-BuildFoundation.yml
@@ -59,7 +59,7 @@ jobs:
     and
     (
       eq(variables['useBuildOutputFromBuildId'],''),
-      in(dependencies.VerifyCopyrightHeaders.result, 'Succeeded', 'SucceededWithIssues', 'Skipped')
+      in(dependencies.VerifyCopyrightHeaders.result, 'Succeeded')
     )
   pool: $(ProjectReunionBuildPool)
   timeoutInMinutes: 120
@@ -141,7 +141,7 @@ jobs:
     and
     (
       eq(variables['useBuildOutputFromBuildId'],''),
-      in(dependencies.VerifyCopyrightHeaders.result, 'Succeeded', 'SucceededWithIssues', 'Skipped')
+      in(dependencies.VerifyCopyrightHeaders.result, 'Succeeded')
     )
   pool: $(ProjectReunionBuildPool)
   timeoutInMinutes: 120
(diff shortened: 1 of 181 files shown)


dev/Detours/disolarm64.cpp
```diff
@@ -1,2 +1,5 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
```
DrusTheAxe, line 2: You're forking Detours source and relicensing it. We shouldn't do this.

[Detours is MIT License](https://github.com/microsoft/Detours/blob/master/LICENSE.md) so it's compatible with us but we should not fork their source. We don't even want their source, just forced to as they don't provide a nuget or equivalent for VS to consume

We can drop this project if/when Detours provides a nuget. Perhaps you'd care to help make that happen instead of forking their source? https://github.com/microsoft/Detours/issues/124

DrusTheAxe:
> cs/cpp/h/idl/xaml files all need copyright headers

What about ps1 and psm1?

What about cmd? "@REM Copyright..."

What about props, targets, csproj, vcxproj, sln? ""

EricJohnson327:
> > cs/cpp/h/idl/xaml files all need copyright headers
> 
> What about ps1 and psm1?
> 
> What about cmd? `@REM Copyright...`
> 
> What about props, targets, csproj, vcxproj*, sln, xml, xsd? `<!-- Copyright... -->`

That's a very good question.  I asked PM about it, but got a "I don't know".  So far I'm just following the task: https://microsoft.visualstudio.com/DefaultCollection/OS/_workitems/edit/36782279  It does specifically say we don't need it for proj files.

dev/Detours/disolarm64.cpp
EricJohnson327: Fixed.  Removed the Detours changes and added an exception to the script.

build/ProjectReunion-BuildFoundation.yml
```diff
@@ -202,6 +223,9 @@ jobs:
 
 - job: BuildMRT
   pool: $(ProjectReunionBuildPool)
+  dependsOn:
+    - VerifyCopyrightHeaders
+  condition: in(dependencies.VerifyCopyrightHeaders.result, 'Succeeded')
```
kythant, line 228: Why are the values in the 'in' conditional different from BuildAnyCPU and Build?

EricJohnson327: I just added the VerifyCopyrightHeaders conditions.  Not sure why BuildMRT already had different conditions.

kythant (approved):

PR #2766, opened 2022-07-22
Rename AppNotification Builder API to match specs

loneursid:
@pmpurifoy - this is a first pass with what was in the specs at the time I looked at it. Ping me when you;re done updating the specs and I'll update this PR accordingly.

All unit tests are passing


pmpurifoy (approved):
This helps a lot. Thanks!

PR #2769, opened 2022-07-22
Making EM fire events.

dhoehna:
For some reason or another EM wasn't firing events.  I changed that.

I believe the reason why was because I wasn't using the correct MACROs(?).

### DIFF
--- a/dev/EnvironmentManager/API/Microsoft.Windows.System.EnvironmentManager.Insights.h
+++ b/dev/EnvironmentManager/API/Microsoft.Windows.System.EnvironmentManager.Insights.h
@@ -2,25 +2,93 @@
 #include "Microsoft.Windows.System.EnvironmentManager.h"
 #include <WindowsAppRuntimeInsights.h>
 
+DECLARE_TRACELOGGING_CLASS(
+EnvironmentManagerTelemetryProvider,
+"Microsoft.Windows.System.EnvironmentManager.Insights",
+(0xfbce41fa, 0xf7bc, 0x538e, 0x11, 0xf2, 0xab, 0xe1, 0xb9, 0xd2, 0xab, 0x23));
+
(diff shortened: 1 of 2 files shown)


DrusTheAxe (approved):

dev/EnvironmentManager/API/Microsoft.Windows.System.EnvironmentManager.Insights.h
```diff
@@ -2,26 +2,94 @@
...
+"Microsoft.Windows.System.EnvironmentManager.Insights",
+(0xfbce41fa, 0xf7bc, 0x538e, 0x11, 0xf2, 0xab, 0xe1, 0xb9, 0xd2, 0xab, 0x23));
 
+class EnvironmentManagerInsights : public wil::TraceLoggingProvider
 {
-    IMPLEMENT_TRACELOGGING_CLASS(
-        EnvironmentManagerInsights,
-        "Microsoft.Windows.System.EnvironmentManager.Insights",
-        (0xfbce41fa, 0xf7bc, 0x538e, 0x11, 0xf2, 0xab, 0xe1, 0xb9, 0xd2, 0xab, 0x23));
+    IMPLEMENT_TELEMETRY_CLASS(EnvironmentManagerInsights, EnvironmentManagerTelemetryProvider);
```
sachintaMSFT, line 12: Just happened to look at this PR.

If you are looking for telemetry, I think you want at least *MEASURES_CLASS. TELEMETRY class doesn't upload any telemetry on retail devices where as MEASURES_CLASS and CRITICAL_CLASS does. The difference between the former two is in what % of retail devices upload telemetry depending on the freshness of the in-market release.

If you are looking for just logging, you shouldn't need TELEMETRY_CLASS and TRACELOGGIN_CLASS should just do.

PR #2780, opened 2022-07-25
AppNotificationBuilder - ProgressBar

loneursid:
Add support for ProgressBar to AppNotificationBuilder


dev/AppNotifications/AppNotificationBuilder/AppNotificationProgressBar.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+        return *this;
+    }
+
+    winrt::hstring AppNotificationProgressBar::ToString()
+    {
+        auto titleValue{ m_titleStatus == BindMode::Value ? m_title : L"{progressTitle}" };
+        auto statusValue{ m_statusStatus == BindMode::Value ? m_status : L"{progressStatus}" };
+        auto valueValue{ m_valueStatus == BindMode::Value ? wil::str_printf<std::wstring>(L"%g", m_value) : L"{progressValue}" };
+        auto valueStringOverrideValue{ m_valueStringOverrideStatus == BindMode::Value ? m_valueStringOverride : L"{progressValueString}" };
```
pmpurifoy, line 103: If we apply the comment from above, we can remove all of these.

dev/AppNotifications/AppNotificationBuilder/AppNotificationProgressBar.cpp
```diff
@@ -0,0 +1,112 @@
...
+#include "AppNotificationProgressBar.h"
+#include "Microsoft.Windows.AppNotifications.Builder.AppNotificationProgressBar.g.cpp"
+
+namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
+{
+    AppNotificationProgressBar::AppNotificationProgressBar()
+        :m_titleStatus{ BindMode::NotSet},
+         m_statusStatus{ BindMode::NotSet },
+         m_valueStatus{ BindMode::NotSet },
+         m_valueStringOverrideStatus{ BindMode::NotSet }
```
pmpurifoy, line 14: We should remove the BindMode::Enum and the member *Status postfix member variables.

For all Bind* functions, we can set the member variable to the constant string. For all Set* functions, we can set the member variable to the passed in param. 

The default ctor should set m_status/m_value to the bind values and leave the rest empty.

dev/AppNotifications/AppNotificationBuilder/AppNotificationProgressBar.cpp
```diff
@@ -0,0 +1,112 @@
...
+    {
+        auto titleValue{ m_titleStatus == BindMode::Value ? m_title : L"{progressTitle}" };
+        auto statusValue{ m_statusStatus == BindMode::Value ? m_status : L"{progressStatus}" };
+        auto valueValue{ m_valueStatus == BindMode::Value ? wil::str_printf<std::wstring>(L"%g", m_value) : L"{progressValue}" };
+        auto valueStringOverrideValue{ m_valueStringOverrideStatus == BindMode::Value ? m_valueStringOverride : L"{progressValueString}" };
+
+        auto title{ m_titleStatus == BindMode::NotSet ? L"" : wil::str_printf<std::wstring>(L" title='%ls'", titleValue.c_str()) };
+        auto status{ wil::str_printf<std::wstring>(L" status='%ls'", statusValue.c_str()) };
+        auto value{ wil::str_printf<std::wstring>(L" value='%ls'", valueValue.c_str()) };
+        auto valueStringOverride{ m_valueStringOverrideStatus == BindMode::NotSet ? L"" : wil::str_printf < std::wstring>(L" valueStringOverride='%ls'", valueStringOverrideValue.c_str()) };
```
pmpurifoy, line 108: If comments from above are addressed, these could turn into:
`auto valueStringOverride{ m_valueOverrideString.empty() ? L"" :  wil::str_printf < std::wstring>(L" valueStringOverride='%ls'", m_valueStringOverride.c_str()) };`

test/AppNotificationBuilderTests/APITests.cpp
```diff
@@ -382,5 +382,84 @@ namespace Test::AppNotification::Builder
...
+
+            VERIFY_ARE_EQUAL(builder.BuildNotification().Payload(), expected);
+        }
+
+        TEST_METHOD(AppNotificationProgressBarDefaults)
+        {
+            auto progressBar{ AppNotificationProgressBar() };
+            auto expected{ L"<progress status='{progressStatus}' value='{progressValue}'/>" };
+
+            VERIFY_ARE_EQUAL(progressBar.as<winrt::Windows::Foundation::IStringable>().ToString(), expected);
```
pmpurifoy, line 417: I'm surprised this works since the object doesn't inherit IStringable in the idl, but maybe . I like the idea of processing everything through BuildNotification instead of casting to IStringable to simulate what devs would actually be doing.

pmpurifoy (changes requested):

dev/AppNotifications/AppNotificationBuilder/AppNotificationProgressBar.cpp
loneursid: What about value? how would you handle that?

pmpurifoy: I would convert m_status from double->winrt::hstring and set the value to {progressValue}.

loneursid: But if the user sets the title property to "title", then calls TitleBind(), then retrive the Title property, he'd get {progreeTitle} but if he does the same for value, what would he get? 0.0, -1.0? we throw?

pmpurifoy: I'm ok with throwing here.

loneursid: Whatever solution we implement, it needs to have the same behavior for users. i.e.: interacting with the title property should work in a similar way as interacting with the value property for example. Users will expect the same behavior.

pmpurifoy: All the other properties are strings which makes them easier, but value is definitely more difficult to handle. If we don't throw, I think 0 or -1 are decent defaults to return.

loneursid: We can't use 0 because it is a valid value.

I'm not a fan of using things like -1. that seems like bad design but if the other reviewers agree that -1 is the correct solution, then I'll change it.

andrewleader (commented):
Looks good to me from a quick 5 minute look-over, I like the BindX() vs SetX(val) options, seems intuitive! I won't mark as approved or denied since I didn't do a deep review :)

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -310,6 +317,18 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
         return m_useButtonStyle ? L" useButtonStyle='true'" : L"";
     }
 
+    std::wstring AppNotificationBuilder::GetProgressBars()
+    {
+        std::wstring result{};
+
```
DrusTheAxe, line 323: NIT: Remove blank line

Looks very C with variables declared in a block before code. C++ style is variables declared with code on 1st use.

Plus nowadays widescreen monitors are the common case so excess blank lines increase vertical height for less readability

dev/AppNotifications/AppNotificationBuilder/AppNotificationProgressBar.cpp
DrusTheAxe: > We can't use 0 because it is a valid value.
> 
> I'm not a fan of using things like -1. that seems like bad design but if the other reviewers agree that -1 is the correct solution, then I'll change it.

I'm not a fan of using valid values as 'no value'. NULL exists for a reason :)

`std::optional<int>` is the Modern C++ way to track a value that may not have a value.

loneursid: Actually a std::variant would be the right solution here, but this isn't the api we have. What we have is a double

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
loneursid: done

dev/AppNotifications/AppNotificationBuilder/AppNotificationProgressBar.cpp
pmpurifoy: What should we use as the binding default double for value? @hulumane

pmpurifoy: The toolkit sets the default to 0: https://github.com/CommunityToolkit/WindowsCommunityToolkit/blob/b21dbe71f64e86ca9bb0ba6cfb4cb4d20e93765a/Microsoft.Toolkit.Uwp.Notifications/Adaptive/AdaptiveProgressBar.cs#L39

If we do it this way, we need to track if the value was manually set to 0 by the dev since it is valid value.

pmpurifoy: @andrewleader  as well for input.

andrewleader: I'm not sure I fully understand the specific problem trying to be addressed here, but I'll try to give some general info.

Note that toast progress bars MUST have a value to be valid (I just tested a toast XML without any value attribute and got a "New notification" notification).

I believe XAML progress bars have a default value of 0, **hence why I decided to just make the default value 0**, even though technically you could argue this is a "nullable" property (but if Value doesn't have a value, you have an invalid toast). I felt just defaulting to 0 is the most intuitive and what developers would "expect", similar to how a string value typically defaults to null, and a new XAML ProgressBar "just works" without setting a value (defaults to 0).

Why would we need to track if the value was manually set to 0 by the dev? Regardless of who sets it to 0, we write it as 0, yeah? Or is this related to returning a value depending on whether BindValue() vs SetValue() was called?

loneursid: @andrewleader : As I understand the specs the value element in the xml can have a value between 0 and 1 or {progressTitle}. That all fine and dandy but we have a property that expose the value as a double. That property can represent a value between 0 and 1 but cannot represent {progressTitle}.

There are 2 options:
1- represent {progressValue} as -1. so the double property can have a value from 0 to 1 or -1.
2- Only use 0 to 1 for the property and maintain an internal flag to indicate whether the value from the property (0 to 1) or {progressValue} should be placed in the xml.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -347,6 +354,17 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
         return m_useButtonStyle ? L" useButtonStyle='true'" : L"";
     }
 
+    std::wstring AppNotificationBuilder::GetProgressBars()
+    {
+        std::wstring result{};
+        for (auto progressBar : m_progressBarList)
+        {
+            result.append(progressBar.as<winrt::Windows::Foundation::IStringable>().ToString());
```
hulumane, line 362: Can you use sprintf? string.append is expensive

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -365,6 +383,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
         xmlResult.append(GetText());
         xmlResult.append(m_attributionText);
         xmlResult.append(GetImages());
+        xmlResult.append(GetProgressBars());
```
hulumane, line 386: Same comment here. string.append is expensive. I thought that we addressed this issue somewhere earlier in Paul's pr and decided not to do this?

dev/AppNotifications/AppNotificationBuilder/AppNotificationProgressBar.cpp
```diff
@@ -0,0 +1,106 @@
...
+    AppNotificationProgressBar::AppNotificationProgressBar()
+        :m_titleStatus{ BindMode::NotSet },
+         m_statusStatus{ BindMode::NotSet },
+         m_valueStatus{ BindMode::NotSet },
+         m_valueStringOverrideStatus{ BindMode::NotSet }
+    {};
+
+    void AppNotificationProgressBar::Title(winrt::hstring const& value)
+    {
+        m_title = value;
```
hulumane, line 19: Do we need locks here and everywhere else?

dev/AppNotifications/AppNotificationBuilder/AppNotificationProgressBar.cpp
```diff
@@ -0,0 +1,106 @@
...
+{
+    AppNotificationProgressBar::AppNotificationProgressBar()
+        :m_titleStatus{ BindMode::NotSet },
+         m_statusStatus{ BindMode::NotSet },
+         m_valueStatus{ BindMode::NotSet },
+         m_valueStringOverrideStatus{ BindMode::NotSet }
+    {};
+
+    void AppNotificationProgressBar::Title(winrt::hstring const& value)
+    {
```
hulumane, line 18: No check for empty args here or anywhere else?

dev/AppNotifications/AppNotificationBuilder/AppNotificationProgressBar.cpp
```diff
@@ -0,0 +1,106 @@
...
+    winrt::Microsoft::Windows::AppNotifications::Builder::AppNotificationProgressBar AppNotificationProgressBar::SetTitle(winrt::hstring const& value)
+    {
+        Title(value);
+
+        return *this;
+    }
+
+    winrt::Microsoft::Windows::AppNotifications::Builder::AppNotificationProgressBar AppNotificationProgressBar::BindTitle()
+    {
+        m_titleStatus = BindMode::Bind;
```
hulumane, line 51: Can we append *BindMode to all this m_ variables?
For example m_tileBindMode, m_statusBindMode

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
loneursid: We use printf everywhere except in this function if I remember correclty. I can turn the whole thing into a printf, no problem

dev/AppNotifications/AppNotificationBuilder/AppNotificationProgressBar.cpp
loneursid: NO locks anywhere. Howard confirmed that we should be locking anything in this class as it isn't intended to be used in a multi-threaded situation

dev/AppNotifications/AppNotificationBuilder/AppNotificationProgressBar.cpp
loneursid: NO, there's a task to handle that in a hollistic manner later. At this point we only check for empty IDs and such but not for normal values.

dev/AppNotifications/AppNotificationBuilder/AppNotificationProgressBar.cpp
```diff
@@ -0,0 +1,106 @@
...
+    {
+        m_valueStringOverrideStatus = BindMode::Bind;
+
+        return *this;
+    }
+
+    winrt::hstring AppNotificationProgressBar::ToString()
+    {
+        return wil::str_printf<std::wstring>(L"<progress%ls%ls%ls%ls/>",
+            m_titleStatus == BindMode::NotSet ? L"" : wil::str_printf<std::wstring>(L" title='%ls'", m_titleStatus == BindMode::Value ? m_title.c_str() : L"{progressTitle}").c_str(),
```
hulumane, line 101: This is a little complicated to read. Can we break this up by assigning each string first and directly passing the arguments to printf rather than doing everything here?

dev/AppNotifications/AppNotificationBuilder/AppNotificationProgressBar.cpp
loneursid: Sure

dev/AppNotifications/AppNotificationBuilder/AppNotificationProgressBar.cpp
loneursid: yup

hulumane: For example, we expect 4 strings here as an argument. So lets initialize the 4 strings and break it up before passing it to printf

dev/AppNotifications/AppNotificationBuilder/AppNotificationProgressBar.cpp
```diff
@@ -0,0 +1,106 @@
...
+    {
+        m_status = value;
+        m_statusStatus = BindMode::Value;
+    }
+
+    void AppNotificationProgressBar::Value(double value)
+    {
+        THROW_HR_IF_MSG(E_INVALIDARG, value < 0.0 || value > 1.0, "You must provide a value between 0.0 and 1.0");
+
+        m_value = value; m_valueStatus = BindMode::Value;
```
hulumane, line 33: 2 lines please

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
pmpurifoy: I believe I only use append in BuildNotification since we reserve the max toast payload size. Append should be 0(1) then since resizing is no longer an issue.

hulumane: This is O(n) for n attributes. Using printf would be O(1) right?

hulumane (approved):

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
loneursid: We agreed to leave as is for now and address later if possible.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
loneursid: Done. Also added a check that we don't exceed the maximin size for the payload.

dev/AppNotifications/AppNotificationBuilder/AppNotificationProgressBar.cpp
loneursid: Done

dev/AppNotifications/AppNotificationBuilder/AppNotificationProgressBar.cpp
loneursid: done

dev/AppNotifications/AppNotificationBuilder/AppNotificationProgressBar.cpp
loneursid: of course...

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
loneursid: I'm pretty sure the difference in performance is negligiible, if any in this case. doing ~15 append isn't realy O(n) since the set of operation is static. It collap[ses to O(1). Doing 15 append() on pre-reserved memory versus calling printf with 15 parameters is going to endup being the same amount of processing. Strings need to be appended in the end.

The sprintf makes this function match the pattern use almost everywhere in the builder and we can conveniently check the size of the string and throw if it exceeds the max payuyload size.

We could have checked and thrown with the previous approach but it would have looked strange to allocate memeory, then check that it hadn't been exceeded.

PR #2739, opened 2022-07-15
AppNotificationContent Builder Core features

pmpurifoy:
This PR contains the implementation of the core features from the AppNotification schema:
- toast
- action
- image
- text
- audio

PR #2786, opened 2022-07-26
AppNotificationBuilder features

pmpurifoy:
This PR is an extension of the other closed PR: https://github.com/microsoft/WindowsAppSDK/pull/2739.

This PR contains the implementation of the core features from the AppNotification schema: https://docs.microsoft.com/en-us/uwp/schemas/tiles/toastschema/root-elements

- toast
- action
- image
- text
- audio


dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -0,0 +1,372 @@
...
+        m_audio = L"<audio silent=\"true\"/>";
+
+        return *this;
+    }
+
+    winrt::Microsoft::Windows::AppNotifications::Builder::AppNotificationBuilder AppNotificationBuilder::AddButton(AppNotificationButton const& value)
+    {
+        THROW_HR_IF(E_INVALIDARG, m_inputList.size() == 5u);
+
+        m_inputList.push_back(value.as<winrt::Windows::Foundation::IStringable>().ToString());
```
loneursid, line 256: I don't think this is the desired behavior. Usually a builder waits until the last minute to lock things up. Capturing the xml for a button at the moment it is added instead of build time, prevents the caller to further modify the button.

loneursid (changes requested):

pmpurifoy:
Need to look into using fmt::format instead of all this string concatenation work.

loneursid:
> Need to look into using fmt::format instead of all this string concatenation work.

Make sure it builds for arm64 platforms. I've seen pipeline failures recently that seems related to the addition of fmt by the builder.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -0,0 +1,372 @@
...
+#include "Microsoft.Windows.AppNotifications.Builder.AppNotificationBuilder.g.cpp"
+#include <fmt/core.h>
+
+using namespace winrt::Windows::Globalization;
+using namespace winrt::Windows::Globalization::DateTimeFormatting;
+namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
+{
+    
+
+    std::wstring AppNotificationBuilder::GetWinSoundEventString(AppNotificationSoundEvent soundEvent)
```
hulumane, line 14: Looks like a utility function. Would recommend that you define a utility helper class for the Builder and put this in there.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -0,0 +1,372 @@
...
+        };
+
+        return c_soundEventMap[soundEvent];
+    }
+
+    winrt::Microsoft::Windows::AppNotifications::Builder::AppNotificationBuilder AppNotificationBuilder::AddArgument(hstring const& key, hstring const& value)
+    {
+        THROW_HR_IF(E_INVALIDARG, key.empty());
+
+        m_arguments.Insert(key, value);
```
hulumane, line 52: Locks? Specifically a write lock

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -0,0 +1,372 @@
...
+        return *this;
+    }
+
+    winrt::Microsoft::Windows::AppNotifications::Builder::AppNotificationBuilder AppNotificationBuilder::SetTimeStamp(winrt::Windows::Foundation::DateTime const& value)
+    {
+        //winrt::hstring currentLanguage { ApplicationLanguages::Languages().First().Current() };
```
hulumane, line 58: Why is this API commented out?

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -0,0 +1,372 @@
...
+        //// DateTimeFormatter dateTimeFormatter{ "longtime" };
+        //winrt::Windows::Foundation::DateTime::
+        //winrt::Windows::Foundation::DateTime { 2022, 1, 13, 16, 25, 30,  }
+        //m_timeStamp = timeStamp;
+        return *this;
+    }
+
+    winrt::Microsoft::Windows::AppNotifications::Builder::AppNotificationBuilder AppNotificationBuilder::SetScenario(AppNotificationScenario const& value)
+    {
+        m_scenario = value;
```
hulumane, line 68: lock

hulumane: Here and everywhere else where we access members.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -0,0 +1,372 @@
...
+        return *this;
+    }
+
+    winrt::Microsoft::Windows::AppNotifications::Builder::AppNotificationBuilder AppNotificationBuilder::AddText(hstring const& text)
+    {
+        THROW_HR_IF(E_INVALIDARG, m_textLines.size() == 3u);
```
hulumane, line 74: Nit: Should this be >= 3u. I understand you would never increment over 3u though

hulumane: Same for the checks below in the other Functions

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -0,0 +1,372 @@
...
+
+        m_textLines.push_back(L"<text>" + text + L"</text>");
+
+        return *this;
+    }
+
+    winrt::Microsoft::Windows::AppNotifications::Builder::AppNotificationBuilder AppNotificationBuilder::AddText(hstring const& text, AppNotificationTextProperties const& properties)
+    {
+        THROW_HR_IF(E_INVALIDARG, m_textLines.size() == 3u);
```
hulumane, line 83: For all of these initial argument checks, would it make sense to also log a message as to why we threw?

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -0,0 +1,372 @@
...
+    {
+        m_scenario = value;
+        return *this;
+    }
+
+    winrt::Microsoft::Windows::AppNotifications::Builder::AppNotificationBuilder AppNotificationBuilder::AddText(hstring const& text)
+    {
+        THROW_HR_IF(E_INVALIDARG, m_textLines.size() == 3u);
+
+        m_textLines.push_back(L"<text>" + text + L"</text>");
```
hulumane, line 76: String concat is generally expensive. The cost likely adds up with the builder pattern. Can we use String format syntax instead?

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -0,0 +1,372 @@
...
+            m_scenario = AppNotificationScenario::IncomingCall;
+        }
+        return *this;
+    }    
+
+    winrt::Microsoft::Windows::AppNotifications::Builder::AppNotificationBuilder AppNotificationBuilder::SetAttributionText(hstring const& text)
+    {
+        THROW_HR_IF(E_INVALIDARG, !m_attributionText.empty());
+
+        m_attributionText = L"<text placement=\"attribution\">" + text + L"</text>";
```
hulumane, line 98: Same concern as above with the cost of string concat.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -0,0 +1,372 @@
...
+
+        return *this;
+    }
+
+    winrt::Microsoft::Windows::AppNotifications::Builder::AppNotificationBuilder AppNotificationBuilder::SetAttributionText(hstring const& text, hstring const& language)
+    {
+        THROW_HR_IF(E_INVALIDARG, !m_attributionText.empty());
+        THROW_HR_IF(E_INVALIDARG, language.empty());
+
+        m_attributionText = L"<text placement=\"attribution\" lang=\"" + language + L"\">" + text + L"</text>";
```
hulumane, line 108: String concat is expensive

hulumane, line 108: General note: Use LR notation and avoid escape characters which are extremely difficult to read.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
pmpurifoy: Yeah I’m looking into other options for string concat. Above I mentioned using fmt::format which builds a string with printf format. Is there an std function I can use for this?

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
pmpurifoy: Should we throw here or the final AppNotification construction? This would apply to all cases @hulumane

loneursid: You should totally throw here. You're adding the button to the builder and we don;t support more than 5. You shouldn't get the string (build the button) until it's needed.

You'll have to keep a reference to the button in your input list, instead of the string already built.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
hulumane: fmtlib is not std c++ right? if I understand correctly and it is simply an open sourced implementation built on top of printf. Other option is to use std::stringstream. 

Would you be able to measure performance for string concat Vs stringstream to see if the benefits warrant the change?

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -0,0 +1,372 @@
+﻿#include "pch.h"
```
DrusTheAxe, line 1: Missing Copyright

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -0,0 +1,372 @@
...
+#include "AppNotificationBuilder.h"
+#include <winrt/Windows.Globalization.h>
+#include <winrt/Windows.Globalization.DateTimeFormatting.h>
+#include "Microsoft.Windows.AppNotifications.Builder.AppNotificationBuilder.g.cpp"
+#include <fmt/core.h>
```
DrusTheAxe, line 6: What's fmt/core.h ?

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -0,0 +1,372 @@
...
+
+using namespace winrt::Windows::Globalization;
+using namespace winrt::Windows::Globalization::DateTimeFormatting;
+namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
```
DrusTheAxe, line 10: NIT: Blank line between usings and namespace

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -0,0 +1,372 @@
...
+{
+    
+
```
DrusTheAxe, line 13: NIT: Delete blank lines 12+13

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -0,0 +1,372 @@
...
+    std::wstring AppNotificationBuilder::GetWinSoundEventString(AppNotificationSoundEvent soundEvent)
+    {
+        static std::map<AppNotificationSoundEvent, std::wstring> c_soundEventMap
```
DrusTheAxe, line 16: If these are all string constants then `PCWSTR` would be better than `std::wstring` (no deep copy)

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -0,0 +1,372 @@
...
+            { AppNotificationSoundEvent::Call4, L"ms-winsoundevent:Notification.Looping.Call4"},
+            { AppNotificationSoundEvent::Call5, L"ms-winsoundevent:Notification.Looping.Call5"},
+            { AppNotificationSoundEvent::Call6, L"ms-winsoundevent:Notification.Looping.Call6"},
+            { AppNotificationSoundEvent::Call7, L"ms-winsoundevent:Notification.Looping.Call7"},
+            { AppNotificationSoundEvent::Call8, L"ms-winsoundevent:Notification.Looping.Call8"},
+            { AppNotificationSoundEvent::Call9, L"ms-winsoundevent:Notification.Looping.Call9"},
+            { AppNotificationSoundEvent::Call10, L"ms-winsoundevent:Notification.Looping.Call10"},
+        };
+
+        return c_soundEventMap[soundEvent];
```
DrusTheAxe, line 45: This is a very Pythonic style but seems heavy for C++. Wouldn't a switch statement be more efficient?

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
DrusTheAxe: Why? `AppNotificationBuilder` is an instance of an inproc object (like std::wstring). Why would you expect 2 threads to concurrently access the same builder instance? One doesn't expect that for std::string, std::vector and zillions other types.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
DrusTheAxe: See other comment re locks

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
DrusTheAxe, line 74: Why 3?

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
DrusTheAxe: [std::string.reserve()](https://cplusplus.com/reference/string/string/reserve/) is your friend. That preallocates a buffer size so if you know a typical worst case you can do this up front then concat e.g.

```
std::wstring s;
s.reserve(4096);
s.append(L"hello");
s.append(123);
s.append(foo);
```

Though `push_back()` may be more efficient than `append()`.

DrusTheAxe: I STRONGLY discourage bringing in fmtlib (or other libraries). Licensing, supply chain tracking, security review and other issues must be addressed for any dependency brought into Windows App SDK. Sufficient justification is needed to add any library.

We have access to Windows, the Standard C++ library, WIL, C++/WinRT and we'll have access to C++20 soon enough which brings `std::format()` and other goodies. I don't see the justification for fmtlib.

If you think there's sufficient justification feel free to make your case. We're currently working on refining the process for dependencies/tool version management so for now I'd suggest starting your proposal to @BenJKuhn, @MikeHillberg and myself and we ran rope in additional folks as appropriate.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
DrusTheAxe: Very likely. Think if you were a developer using WinAppSDK and didn't have the source. Would you understand or could reasonably figure out why an error occurred? If not then a *_MSG() is recommended.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
DrusTheAxe: `wil::str_printf()` is your friend

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -0,0 +1,372 @@
...
+        }
+
+        // Add scenario attribute if set
+        if (m_scenario != AppNotificationScenario::Default)
+        {
+            winrt::hstring scenario{};
+            switch (m_scenario)
+            {
+            case AppNotificationScenario::Alarm:
+                scenario = L"\"alarm\"";
```
DrusTheAxe, line 291: `scenario = LR"("alarm")";`

Raw strings make complex strings easier to read, and R"( before and )" after lets you embed double quotes in a string without escaping them

DrusTheAxe: Are you constructing XML? You can use apostraphe to quote arguments instead of double-quotes which simplifies a lot of your string syntax e.g. line 294

`scenario = L"\"reminder\"";`

produces on line 304 XML of `...scenario="reminder"...`

but `scenario = L"'reminder'";` would produce `...scenario='reminder'...`

DrusTheAxe (changes requested):

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
pmpurifoy: This is the maximum amount of text elements that the schema will support on xml.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
pmpurifoy: Yes, we are building xml. Will move to use ' intsead of of ".

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
pmpurifoy: Adding in next iteration.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
pmpurifoy: This is exactly what I was looking for, thanks!

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
pmpurifoy: Removing as it was a header for the fmt library

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
DrusTheAxe: `3` is a magic constant. Better to use symbolic constants e.g.

```
constexpr size_t c_maxTextElementsSupportedByTheSchema{ 3u };
THROW_HR_IF(E_INVALIDARG, m_textLines.size() >= c_maxTextElementsSupportedByTheSchema);
```

pmpurifoy: Applies for other magic constant maximums.

pmpurifoy:
> Need to look into using fmt::format instead of all this string concatenation work.

Rejecting this idea due to it being a 3rd party library.

loneursid:
@pmpurifoy - Fixed the build so we can be confident the build pipeline are passing and we can merge this PR as soon as it is approved.
  - I fixed a conflict in WindowsAppSDK solution file.
  - Added copyright headers to the idl file and a pch.
  - remove references to fmt/core.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
pmpurifoy: Fixed string concat issues

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
pmpurifoy: Added error messages for THROW_HRs

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
pmpurifoy: Moving to switch statement

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
pmpurifoy: Fixed.

loneursid (approved):
LGTM

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -237,134 +237,150 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
...
 
+    std::wstring AppNotificationBuilder::GetArguments()
+    {
         // Add launch arguments if given arguments
         if (m_arguments.Size())
         {
-            xmlResult = xmlResult + L" launch=\"";
-
-            std::wstring arguments{};
+            std::wstring arguments{ };
```
DrusTheAxe, line 245: NIT: remove space `std::wstring argments{};`

Or even remove the `{}`. For void ctor `std::wstring arguments;` suffices.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -14,7 +14,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
 {
     inline const size_t c_maxAppNotificationPayload{ 5120 };
     inline const uint8_t c_maxTextElements{ 3 };
-    inline const uint8_t c_maxInputElements{ 5 };
+    inline const uint8_t c_maxButtonElements{ 5 };
```
DrusTheAxe: NIT: `constexpr`

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -100,36 +99,26 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
...
         m_attributionText = wil::str_printf<std::wstring>(L"<text placement='attribution'>%ws</text>", text.c_str());
-
         return *this;
     }
 
     winrt::Microsoft::Windows::AppNotifications::Builder::AppNotificationBuilder AppNotificationBuilder::SetAttributionText(hstring const& text, hstring const& language)
     {
-        THROW_HR_IF(E_INVALIDARG, !m_attributionText.empty());
-        THROW_HR_IF(E_INVALIDARG, language.empty());
+        THROW_HR_IF_MSG(E_INVALIDARG, language.empty(), "You must provide a language calling SetAttributionText.");
```
DrusTheAxe, line 87: NIT: Remove the period. Don't need pretty full english sentence for an error message. If it were a paragraph of sentences then yes, but for a single sentence/phase, nah

dev/AppNotifications/AppNotificationBuilder/AppNotificationButton.cpp
```diff
@@ -50,15 +50,15 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
 
     winrt::Microsoft::Windows::AppNotifications::Builder::AppNotificationButton AppNotificationButton::SetInvokeUri(winrt::Windows::Foundation::Uri const& protocolUri)
     {
-        THROW_HR_IF(E_INVALIDARG, m_arguments.Size() > 0u);
+        THROW_HR_IF_MSG(E_INVALIDARG, m_arguments.Size() > 0u, "You cannot add a protocol activation uri after calling AddArgument.");
```
DrusTheAxe, line 64: `!m_arguments.empty()` is better than `.Size() == 0` checks

Ditto `.empty()` is preferred over `.Size() == 0`
Also other AppNotificationBuilder code uses .empty so consistency hath its virtues :-)

dev/AppNotifications/AppNotificationBuilder/AppNotificationButton.cpp
```diff
@@ -11,8 +11,8 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
 
     winrt::Microsoft::Windows::AppNotifications::Builder::AppNotificationButton AppNotificationButton::AddArgument(winrt::hstring const& key, winrt::hstring const& value)
     {
-        THROW_HR_IF(E_INVALIDARG, key.empty());
-        THROW_HR_IF(E_INVALIDARG, m_protocolUri);
+        THROW_HR_IF_MSG(E_INVALIDARG, key.empty(), "You must provide a key when adding an argument.");
+        THROW_HR_IF_MSG(E_INVALIDARG, m_protocolUri, "You cannot add an argument after calling SetInvokeUri.");
```
DrusTheAxe, line 26: NIT: `!m_protocolUri.empty()` ?

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilderUtility.h
```diff
@@ -0,0 +1,72 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#pragma once
+#include "pch.h"
+#include "winrt/Microsoft.Windows.AppNotifications.Builder.h"
+
+inline const size_t c_maxAppNotificationPayload{ 5120 };
```
DrusTheAxe, line 8: `constexpr` is better than `inline const`

same comment applies throught

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilderUtility.h
```diff
@@ -0,0 +1,72 @@
...
+inline const uint8_t c_maxTextElements{ 3 };
+inline const uint8_t c_maxButtonElements{ 5 };
+
+namespace AppNotificationBuilder
+{
+    using namespace winrt::Microsoft::Windows::AppNotifications::Builder;
+}
+
+inline std::wstring GetWinSoundEventString(AppNotificationBuilder::AppNotificationSoundEvent soundEvent)
```
DrusTheAxe, line 17: NIT: Returning these string constants as string objects means malloc+free hits. PCWSTR or maybe `std::wstring_view` is likely more efficient (depends on callers but usually the case)

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
pmpurifoy: I wasn’t sure to add a period or not. Will fix.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -23,12 +26,14 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
...
     winrt::Microsoft::Windows::AppNotifications::Builder::AppNotificationBuilder AppNotificationBuilder::SetTimeStamp(winrt::Windows::Foundation::DateTime const& value)
     {
-        
-        //winrt::std::wstring currentLanguage { ApplicationLanguages::Languages().First().Current() };
-        //// DateTimeFormatter dateTimeFormatter{ "longtime" };
-        //winrt::Windows::Foundation::DateTime::
-        //winrt::Windows::Foundation::DateTime { 2022, 1, 13, 16, 25, 30,  }
-        //m_timeStamp = timeStamp;
+        auto seconds{ winrt::clock::to_time_t(value) };
+        struct tm buf;
```
DrusTheAxe, line 35: NIT: `struct tm buf{};`

same comment applies throught

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -23,12 +26,14 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
...
+        gmtime_s(&buf, &seconds);
+
+        std::wstringstream buffer;
+        buffer << std::put_time(&buf, L"%FT%T");
+
+        m_timeStamp = wil::str_printf<std::wstring>(L" displayTimestamp='%wsZ'", buffer.str().c_str());
```
DrusTheAxe, line 41: Why the trailing 'Z'? Should above be `std::put_time(..., L"%FT%T%Z")` ?

test/AppNotificationBuilderTests/APITests.cpp
```diff
@@ -52,6 +52,22 @@ namespace Test::AppNotification::Builder
...
         }
 
+        TEST_METHOD(AppNotificationBuilderSetTimeStamp)
+        {
+            std::wstring dateTimeFormat{ L"%d/%m/%Y %H:%M:%S" };
+            std::wistringstream ss(L"22/12/2016 01:12:10");
+
+            struct tm dateTime {};
+            ss >> std::get_time(&dateTime, dateTimeFormat.c_str());
+            std::time_t time{ mktime(&dateTime) };
```
DrusTheAxe, line 62: Make a string to convert to struct...seems like a lot of work instead of just `struct tm dateTime{ x, y, z, ... };`

Also, do `std::get_time` etc deal in localtime and locale formatting? Results would depend on the machine and user context when run.

Setting the `tm` struct directly seems simpler and more reliable

DrusTheAxe (approved):

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -16,6 +16,11 @@ using namespace winrt::Windows::Globalization::DateTimeFormatting;
 
 namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
 {
+    bool AppNotificationBuilder::IsUrgentScenarioSupported()
+    {
+        return GetBuildNumber() >= 19041;
```
DrusTheAxe, line 21: Ugh. Version checks. That's a minefield (as in, a reliable answer is a minefield...)

Please use functions in dev\common\IsWindowsVersion.h e.g. `WindowsVersion::IsWindows10_19H10OrGreater()`. That has checks for >=19H1 and >=20H1, and if additional version checks are needed extend as appropos.

That avoids the minefield of GetVersion mechanisms which don't necessarily yield the expected answer

loneursid:
Thaks everyone for all the help. That's a great first step for the builder. I will follow up with a PR to address @DrusTheAxe latest feedback.

DrusTheAxe (approved):
Update the version check as discussed in this PR or a follow up PR. Important for release but merely 'soon' not necessarily 'today'

dev/AppNotifications/AppNotificationBuilder/AppNotificationButton.cpp
pmpurifoy: So in AppNotification xml, you cannot set both Arguments and a ProtocolUri for activation. This is basically saying you can only have one or the other. The reverse check appears in SetInvokeUri as well which verifies AddArguments hasn't been called.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
loneursid: got a PR to address this last round of feedback, here: https://github.com/microsoft/WindowsAppSDK/pull/2801

dev/AppNotifications/AppNotificationBuilder/AppNotificationButton.cpp
loneursid: got a PR to address this last round of feedback, here: https://github.com/microsoft/WindowsAppSDK/pull/2801

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -12,7 +12,9 @@ using namespace winrt::Windows::Globalization::DateTimeFormatting;
 
 namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
 {
-    
+    inline const size_t c_maxAppNotificationPayload{ 5120 };
```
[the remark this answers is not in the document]
loneursid: Addreessed in https://github.com/microsoft/WindowsAppSDK/pull/2801

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -74,18 +77,19 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
 
     winrt::Microsoft::Windows::AppNotifications::Builder::AppNotificationBuilder AppNotificationBuilder::AddText(hstring const& text)
     {
-        THROW_HR_IF(E_INVALIDARG, m_textLines.size() == 3u);
+        THROW_HR_IF(E_INVALIDARG, m_textLines.size() >= c_maxTextElements);
 
-        m_textLines.push_back(L"<text>" + text + L"</text>");
+        m_textLines.push_back(wil::str_printf<std::wstring>(L"<text>%ws</text>", text.c_str()).c_str());
```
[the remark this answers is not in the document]
loneursid: Addressed in https://github.com/microsoft/WindowsAppSDK/pull/2801

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
loneursid: Addressed in https://github.com/microsoft/WindowsAppSDK/pull/2801

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
loneursid: Addressed in https://github.com/microsoft/WindowsAppSDK/pull/2801

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilderUtility.h
loneursid: Addressed in https://github.com/microsoft/WindowsAppSDK/pull/2801

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
loneursid: Addressed in https://github.com/microsoft/WindowsAppSDK/pull/2801

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
pmpurifoy: The reason why it's not in std::put_time is because we want the Z to exist as the raw character in the XML to be processed later in AppNotificationBuilder.Show.

dev/AppNotifications/AppNotificationBuilder/AppNotificationButton.cpp
pmpurifoy: Since this is a winrt::Uri isn't it ok to do it this way? Otherwise, it will have to be m_protocolUri.ToString().empty() or something like it

dev/AppNotifications/AppNotificationBuilder/AppNotificationButton.cpp
pmpurifoy: I guess the issue is winrt::IMap doesn't have an empty function. What do you suggest here?

PR #2801, opened 2022-07-29
AppNotificationBuilder - Code Improvements

loneursid:
Addressing last round of feedback from: https://github.com/microsoft/WindowsAppSDK/pull/2786

Beside nits, one important issue that must addressed in this PR is retrieving the version number. Dev/Common should have a utility function to retrieve the version number. Let's make sure that whatever solution we used, it's the same for all WindowsAppSDK.

From @DrusTheAxe 
> Ugh. Version checks. That's a minefield (as in, a reliable answer is a minefield...)
> 
> Please use functions in dev\common\IsWindowsVersion.h e.g. WindowsVersion::IsWindows10_19H10OrGreater(). That has checks for >=19H1 and >=20H1, and if additional version checks are needed extend as appropos.
> 
> That avoids the minefield of GetVersion mechanisms which don't necessarily yield the expected answer

### DIFF
--- a/dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
+++ b/dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
@@ -32,7 +32,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
     winrt::Microsoft::Windows::AppNotifications::Builder::AppNotificationBuilder AppNotificationBuilder::SetTimeStamp(winrt::Windows::Foundation::DateTime const& value)
     {
         auto seconds{ winrt::clock::to_time_t(value) };
-        struct tm buf;
+        struct tm buf {};
         gmtime_s(&buf, &seconds);
 
         std::wstringstream buffer;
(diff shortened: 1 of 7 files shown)


dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -32,13 +32,13 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
...
         auto seconds{ winrt::clock::to_time_t(value) };
-        struct tm buf;
+        struct tm buf {};
         gmtime_s(&buf, &seconds);
 
         std::wstringstream buffer;
         buffer << std::put_time(&buf, L"%FT%T");
 
-        m_timeStamp = wil::str_printf<std::wstring>(L" displayTimestamp='%wsZ'", buffer.str().c_str());
+        m_timeStamp = wil::str_printf<std::wstring>(L" displayTimestamp='%lsZ'", buffer.str().c_str());
```
loneursid, line 41: From @DrusTheAxe in a different PR: Why the trailing 'Z'? Should above be std::put_time(..., L"%FT%T%Z") ?

pmpurifoy: The reason why it's not in std::put_time is because we want the Z to exist as the raw character in the XML to be processed later in AppNotificationBuilder.Show.

pmpurifoy (approved):

PR #2802, opened 2022-07-30
User/sachinta/fix deployment manager telemetry

sachintaMSFT:
1. Rectified setting up of Deployment Manager Initialize Activity Object within the context object which fixes the crash in packaged app when initialize API is called multiple times within single packaged app session.
- This also helps us in fixing telemetry tracking of Deployment Manager Initialize API because start and stop events of the API are not connected with same activity Id due to the issue being fixed. That impacts telemetry data processing causing us unable to see the filter values that we log in Start event (ex: isElevated, isPackagedProcess, IntegrityLevel etc).
- This also helps us in connecting DeploymentAgent.exe telemetry with DeploymentManager's for a single Initialize API call. Without the fix, DeploymentAgent.exe is always recording the corelated Intialize API call activityID as GUID_NULL.

2. Remove IsFullTrustPackage field when preInitializationStatus is already in OK status (i.e. Main and Singleton packages are already installed).

### DIFF
--- a/dev/Deployment/DeploymentManager.cpp
+++ b/dev/Deployment/DeploymentManager.cpp
@@ -46,10 +46,10 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
 
     winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::DeploymentResult DeploymentManager::Initialize()
     {
-        ::WindowsAppRuntime::Deployment::Activity::Context::Get().GetActivity().Start(false,
-                                                                                      Security::IntegrityLevel::IsElevated(),
-                                                                                      AppModel::Identity::IsPackagedProcess(),
-                                                                                      Security::IntegrityLevel::GetIntegrityLevel());
+        ::WindowsAppRuntime::Deployment::Activity::Context::Get().SetActivity(WindowsAppRuntimeDeployment_TraceLogger::Initialize::Start(false,
+                                                                                        Security::IntegrityLevel::IsElevated(),
+                                                                                        AppModel::Identity::IsPackagedProcess(),
+                                                                                        Security::IntegrityLevel::GetIntegrityLevel()));
 
         FAIL_FAST_HR_IF(HRESULT_FROM_WIN32(APPMODEL_ERROR_NO_PACKAGE), !AppModel::Identity::IsPackagedProcess());
         return Initialize(GetCurrentFrameworkPackageFullName());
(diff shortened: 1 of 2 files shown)


dev/Deployment/DeploymentManager.cpp
```diff
@@ -46,21 +46,22 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
 
     winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::DeploymentResult DeploymentManager::Initialize()
     {
-        ::WindowsAppRuntime::Deployment::Activity::Context::Get().GetActivity().Start(false,
-                                                                                      Security::IntegrityLevel::IsElevated(),
-                                                                                      AppModel::Identity::IsPackagedProcess(),
-                                                                                      Security::IntegrityLevel::GetIntegrityLevel());
+        ::WindowsAppRuntime::Deployment::Activity::Context::Get().SetActivity(WindowsAppRuntimeDeployment_TraceLogger::Initialize::Start(false,
+                                                                                        Security::IntegrityLevel::IsElevated(),
```
DrusTheAxe, line 50: Indentation for 50-52 seems odd

dev/Deployment/DeploymentManager.cpp
```diff
@@ -46,21 +46,22 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
 
     winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::DeploymentResult DeploymentManager::Initialize()
     {
-        ::WindowsAppRuntime::Deployment::Activity::Context::Get().GetActivity().Start(false,
-                                                                                      Security::IntegrityLevel::IsElevated(),
-                                                                                      AppModel::Identity::IsPackagedProcess(),
-                                                                                      Security::IntegrityLevel::GetIntegrityLevel());
+        ::WindowsAppRuntime::Deployment::Activity::Context::Get().SetActivity(WindowsAppRuntimeDeployment_TraceLogger::Initialize::Start(false,
```
DrusTheAxe, line 49: Multiple calls to `::WindowsAppRuntime::Deployment::Activity::Context::Get().blahblah`

SUGGESTION:
```
auto& activityContext{ ::WindowsAppRuntime::Deployment::Activity::Context::Get() };
activityContext.SetActivity(...)
...
activityContext.SetActivity(...)
```

DrusTheAxe (approved):
LGTM modulo some minor comments

PR #2805, opened 2022-08-02
Encoding/Decoding AppNotificationBuilder arguments

pmpurifoy:
This PR contains the implementation of encoding special characters into the activation arguments for AppNotification/Button activation.

The encoded args can be retrieved in the App through AppNotificationActivatedEventArgs.Arguments which returns a decoded mapping of key/values pairs.

Added unit tests to verify correct encoding and manual testing to verify decoding from app activation was returning key/value mapping correctly.

### DIFF
--- a/dev/AppNotifications/AppNotificationActivatedEventArgs.cpp
+++ b/dev/AppNotifications/AppNotificationActivatedEventArgs.cpp
@@ -16,7 +16,8 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
         size_t pos{ 0 };
 
         // Separate the key/value pairs by ';' as the delimiter
-        while ((pos = arguments.find(L';')) != std::wstring::npos) {
+        while ((pos = arguments.find(L';')) != std::wstring::npos)
+        {
             pairs.push_back(arguments.substr(0, pos));
             arguments.erase(0, pos + 1);
         }
--- a/dev/AppNotifications/AppNotificationActivatedEventArgs.cpp
+++ b/dev/AppNotifications/AppNotificationActivatedEventArgs.cpp
@@ -1,0 +1,42 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#include "pch.h"
+#include "AppNotificationActivatedEventArgs.h"
+#include "Microsoft.Windows.AppNotifications.AppNotificationActivatedEventArgs.g.cpp"
+#include "AppNotificationBuilder/AppNotificationBuilderUtility.h"
+
+namespace winrt::Microsoft::Windows::AppNotifications::implementation
+{
+    winrt::Windows::Foundation::Collections::IMap<hstring, hstring> AppNotificationActivatedEventArgs::DecodeArguments(std::wstring arguments)
+    {
+        auto result{ winrt::single_threaded_map<winrt::hstring, winrt::hstring>() };
+
+        std::vector<std::wstring> pairs{};
+        size_t pos{ 0 };
+
+        // Separate the key/value pairs by ';' as the delimiter
+        while ((pos = arguments.find(L';')) != std::wstring::npos) {
+            pairs.push_back(arguments.substr(0, pos));
+            arguments.erase(0, pos + 1);
+        }
+
+        // Need to push back final string
+        pairs.push_back(arguments);
+
+        for (auto pair : pairs)
+        {
+            // Get the key/value individual values separated by '='
+            pos = pair.find(L'=');
+            if (pos == std::wstring::npos)
+            {
+                result.Insert(Decode(pair).c_str(), L"");
+            }
+            else
+            {
+                result.Insert(Decode(pair.substr(0, pos)).c_str(), Decode(pair.substr(pos + 1)).c_str());
+            }
+        }
+        return result;
+    }
+}
--- a/dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
+++ b/dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
@@ -64,7 +64,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
     {
         THROW_HR_IF_MSG(E_INVALIDARG, m_textLines.size() >= c_maxTextElements, "Maximum number of text elements added");
 
-        m_textLines.push_back(wil::str_printf<std::wstring>(L"<text>%ls</text>", EncodeXml(text.c_str()).c_str()).c_str());
+        m_textLines.push_back(wil::str_printf<std::wstring>(L"<text>%ls</text>", EncodeXml(text).c_str()).c_str());
         return *this;
     }
 
@@ -73,7 +73,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
         THROW_HR_IF_MSG(E_INVALIDARG, m_textLines.size() >= c_maxTextElements, "Maximum number of text elements added");
 
         std::wstring props{ properties.as<winrt::Windows::Foundation::IStringable>().ToString() };
-        m_textLines.push_back(wil::str_printf<std::wstring>(L"%ls%ls</text>", props.c_str(), EncodeXml(text.c_str()).c_str()).c_str());
+        m_textLines.push_back(wil::str_printf<std::wstring>(L"%ls%ls</text>", props.c_str(), EncodeXml(text).c_str()).c_str());
 
         if (properties.IncomingCallAlignment())
         {
@@ -84,7 +84,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
 
     winrt::Microsoft::Windows::AppNotifications::Builder::AppNotificationBuilder AppNotificationBuilder::SetAttributionText(hstring const& text)
     {
-        m_attributionText = wil::str_printf<std::wstring>(L"<text placement='attribution'>%ls</text>", EncodeXml(text.c_str()).c_str());
+        m_attributionText = wil::str_printf<std::wstring>(L"<text placement='attribution'>%ls</text>", EncodeXml(text).c_str());
         return *this;
     }
 
@@ -92,7 +92,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
     {
         THROW_HR_IF_MSG(E_INVALIDARG, language.empty(), "You must provide a language calling SetAttributionText");
 
-        m_attributionText = wil::str_printf<std::wstring>(L"<text placement='attribution' lang='%ls'>%ls</text>", language.c_str(), EncodeXml(text.c_str()).c_str());
+        m_attributionText = wil::str_printf<std::wstring>(L"<text placement='attribution' lang='%ls'>%ls</text>", language.c_str(), EncodeXml(text).c_str());
         return *this;
     }
 
@@ -121,7 +121,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
         THROW_HR_IF_MSG(E_INVALIDARG, alternateText.empty(), "You must provide an alternate text string calling SetInlineImage");
 
         std::wstring hintCrop { imageCrop == AppNotificationImageCrop::Circle ? L" hint-crop='circle'" : L"" };
-        m_inlineImage = wil::str_printf<std::wstring>(L"<image src='%ls' alt='%ls'%ls/>", imageUri.ToString().c_str(), alternateText.c_str(), hintCrop.c_str());
+        m_inlineImage = wil::str_printf<std::wstring>(L"<image src='%ls' alt='%ls'%ls/>", imageUri.ToString().c_str(), EncodeXml(alternateText).c_str(), hintCrop.c_str());
 
         return *this;
     }
@@ -151,7 +151,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
         THROW_HR_IF_MSG(E_INVALIDARG, alternateText.empty(), "You must provide an alternate text string calling SetAppLogoOverride");
 
         std::wstring hintCrop{ imageCrop == AppNotificationImageCrop::Circle ? L" hint-crop='circle'" : L"" };
-        m_appLogoOverride = wil::str_printf<std::wstring>(L"<image placement='appLogoOverride' src='%ls' alt='%ls'%ls/>", imageUri.ToString().c_str(), alternateText.c_str(), hintCrop.c_str());
+        m_appLogoOverride = wil::str_printf<std::wstring>(L"<image placement='appLogoOverride' src='%ls' alt='%ls'%ls/>", imageUri.ToString().c_str(), EncodeXml(alternateText).c_str(), hintCrop.c_str());
 
         return *this;
     }
@@ -166,7 +166,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
     {
         THROW_HR_IF_MSG(E_INVALIDARG, alternateText.empty(), "You must provide an alternate text string calling SetHeroImage");
 
-        m_heroImage = wil::str_printf<std::wstring>(L"<image placement='hero' src='%ls' alt='%ls'/>", imageUri.ToString().c_str(), alternateText.c_str());
+        m_heroImage = wil::str_printf<std::wstring>(L"<image placement='hero' src='%ls' alt='%ls'/>", imageUri.ToString().c_str(), EncodeXml(alternateText).c_str());
         return *this;
     }
 
@@ -213,7 +213,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
         THROW_HR_IF_MSG(E_INVALIDARG, id.empty(), "You must provide an id for the TextBox");
 
 
-        m_textBoxList.push_back(wil::str_printf<std::wstring>(L"<input id='%ls' type='text'/>", id.c_str()));
+        m_textBoxList.push_back(wil::str_printf<std::wstring>(L"<input id='%ls' type='text'/>", EncodeXml(id).c_str()));
         return *this;
     }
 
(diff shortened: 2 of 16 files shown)


dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilderUtility.h
```diff
@@ -1,13 +1,16 @@
...
 
-#pragma once
 #include "pch.h"
 #include "winrt/Microsoft.Windows.AppNotifications.Builder.h"
+#include <algorithm>
 
 constexpr size_t c_maxAppNotificationPayload{ 5120 };
 constexpr uint8_t c_maxTextElements{ 3 };
 constexpr uint8_t c_maxButtonElements{ 5 };
+constexpr PCWSTR c_encodedPercent{ L"%25" };
```
hulumane, line 11: You could add the encodings in a map and directly translate them by inferencing the map. That way, you don't have to define new constants every single time for each encoding.

I would also suggest that you add all the % encodings here to the map to be thorough:
https://developer.mozilla.org/en-US/docs/Glossary/percent-encoding

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilderUtility.h
```diff
@@ -70,3 +73,55 @@ inline PCWSTR GetWinSoundEventString(AppNotificationBuilder::AppNotificationSoun
...
     }
 }
+
+inline std::wstring Encode(winrt::hstring const& value)
+{
+    std::wstring encodedValue{};
+    for (auto ch : value)
+    {
+        switch (ch) {
+        case '%':
```
hulumane, line 83: See comment above. Using a map would make this so much more simpler and you could avoid the switch case.

hulumane: You could have something like this:
if (map.contains(ch))
{
    encodedValue.append(map[ch];
}
else
{
 encodedValue.append(ch);
}

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilderUtility.h
```diff
@@ -70,3 +73,55 @@ inline PCWSTR GetWinSoundEventString(AppNotificationBuilder::AppNotificationSoun
...
+            break;
+        default:
+            encodedValue.push_back(ch);
+        }
+    }
+    return encodedValue;
+}
+
+// This function replaces the encoded values with the actual values
+inline std::wstring DecodeString(std::wstring value, std::wstring const& encodedValue, char ch)
```
hulumane, line 100: Can we avoid making a copy of value here?

pmpurifoy: Yeah we can. Need to add another prevPos variable to track where we are in the string. The original idea was to destroy the copied string.

hulumane, line 100: You could define a reverse map to decode. ie. translate 3B to =. By doing that, you would simplify this logic even further and you wouldn't have to pass ch here since the map does it for you.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilderUtility.h
pmpurifoy: I think a map would work best here, but only ; % = need to be encoded. These are the only special characters that have a special behavior inside the xml. The link above is referring to URL encoding which in our case, only use 3.
; separates each key/value pair
= separates the key and value
% since we use % encoding, we encode % to retain the raw value

Will change to map construct.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilderUtility.h
pmpurifoy: Fixed

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilderUtility.h
pmpurifoy: Removing in favor of std::regex_replace

dev/AppNotifications/AppNotificationActivatedEventArgs.cpp
```diff
@@ -0,0 +1,42 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#include "pch.h"
+#include "AppNotificationActivatedEventArgs.h"
+#include "Microsoft.Windows.AppNotifications.AppNotificationActivatedEventArgs.g.cpp"
+#include "AppNotificationBuilder/AppNotificationBuilderUtility.h"
+
+namespace winrt::Microsoft::Windows::AppNotifications::implementation
+{
+    winrt::Windows::Foundation::Collections::IMap<hstring, hstring> AppNotificationActivatedEventArgs::DecodeArguments(std::wstring arguments)
```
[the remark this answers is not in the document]
pmpurifoy: Added decoding tests, but in the future will have to add an automated activation test.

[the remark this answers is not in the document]
pmpurifoy: https://github.com/microsoft/WindowsAppSDK/issues/2812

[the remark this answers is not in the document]
loneursid: Thanks, that will give us some protection / documentation until we get to the automated activation test

loneursid (approved):
LGTM

test/AppNotificationBuilderTests/APITests.cpp
```diff
@@ -347,86 +384,86 @@ namespace Test::AppNotification::Builder
...
+            auto builder{ winrt::AppNotificationBuilder()
                 .AddText(L"Downloading this week's new music...")
-                .AddProgressBar(AppNotificationProgressBar()
+                .AddProgressBar(winrt::AppNotificationProgressBar()
                     .BindTitle()
                     .BindValueStringOverride()) };
-            auto expected{ L"<toast><visual><binding template='ToastGeneric'><text>Downloading this week's new music...</text><progress title='{progressTitle}' status='{progressStatus}' value='{progressValue}' valueStringOverride='{progressValueString}'/></binding></visual></toast>" };
-
+            auto expected{ L"<toast><visual><binding template='ToastGeneric'><text>Downloading this week&apos;s new music...</text><progress title='{progressTitle}' status='{progressStatus}' value='{progressValue}' valueStringOverride='{progressValueString}'/></binding></visual></toast>" };
+            WEX::Logging::Log::Comment(builder.BuildNotification().Payload().c_str());
```
loneursid, line 437: Is this a left over from a debugging session?

pmpurifoy: Yup. I'm still working on unit tests but will remove when finished.

test/AppNotificationBuilderTests/APITests.cpp
```diff
@@ -597,13 +634,34 @@ namespace Test::AppNotification::Builder
...
                 .SetSelectedItem(L""), E_INVALIDARG);
         }
 
         TEST_METHOD(AppNotificationCreateComboBoxWithoutAnId)
         {
-            VERIFY_THROWS_HR(AppNotificationComboBox(L""), E_INVALIDARG);
+            VERIFY_THROWS_HR(winrt::AppNotificationComboBox(L""), E_INVALIDARG);
+        }
+
+        TEST_METHOD(AppNotificationBuilderEscapeXmlCharacters)
```
loneursid, line 646: Thanks for adding these tests

dev/AppNotifications/AppNotificationActivatedEventArgs.cpp
```diff
@@ -0,0 +1,43 @@
...
```
hulumane, line 11: this needs a comment. What is the expected input and output?

A sample would be great

loneursid: The samples will be updated to use the builder next.

hulumane: no i meant a comment

hulumane: This API has a contract. It is important to comment that contract here so that others don't come in and introduce bugs.

loneursid: ok, I'll add a comment in the nits PR today. I don't want to push new changes to this one, unless something is really brocken.

dev/AppNotifications/AppNotificationActivatedEventArgs.cpp
```diff
@@ -0,0 +1,43 @@
...
+        // Need to push back final string
+        pairs.push_back(arguments);
+
+        for (auto pair : pairs)
+        {
+            // Get the key/value individual values separated by '='
+            pos = pair.find(L'=');
+            if (pos == std::wstring::npos)
+            {
+                result.Insert(Decode(pair).c_str(), L"");
```
hulumane, line 34: So is it expected to return empty string as the value for a pair? Can we get away with not putting in anything here?

dev/AppNotifications/AppNotificationActivatedEventArgs.cpp
```diff
@@ -0,0 +1,43 @@
...
+{
+    winrt::Windows::Foundation::Collections::IMap<hstring, hstring> AppNotificationActivatedEventArgs::DecodeArguments(std::wstring arguments)
+    {
+        auto result{ winrt::single_threaded_map<winrt::hstring, winrt::hstring>() };
+
+        std::vector<std::wstring> pairs{};
+        size_t pos{ 0 };
+
+        // Separate the key/value pairs by ';' as the delimiter
+        while ((pos = arguments.find(L';')) != std::wstring::npos)
```
hulumane, line 19: This is a much easier way to tokenize:

int main()
{
    char str[] = "Geeks-for-Geeks";
 
    // Returns first token
    char *token = strtok(str, " ");
 
    // Keep printing tokens while one of the
    // delimiters present in str[].
    while (token != NULL)
    {
        printf("%s\n", token);
        token = strtok(NULL, " ");
    }
 
    return 0;

hulumane: This would avoid that last push_back that you have outside the while loop

hulumane: And you don't have to do that erase :)

dev/AppNotifications/AppNotificationActivatedEventArgs.cpp
```diff
@@ -0,0 +1,43 @@
...
+            arguments.erase(0, pos + 1);
+        }
+
+        // Need to push back final string
+        pairs.push_back(arguments);
+
+        for (auto pair : pairs)
+        {
+            // Get the key/value individual values separated by '='
+            pos = pair.find(L'=');
```
hulumane, line 31: You can use tokenizers here too

dev/AppNotifications/AppNotificationActivatedEventArgs.cpp
```diff
@@ -0,0 +1,43 @@
...
+            if (pos == std::wstring::npos)
+            {
+                result.Insert(Decode(pair).c_str(), L"");
+            }
+            else
+            {
+                result.Insert(Decode(pair.substr(0, pos)).c_str(), Decode(pair.substr(pos + 1)).c_str());
```
hulumane, line 38: Decode(...) already returns a string. So no need of the .c_str() right?

dev/AppNotifications/AppNotificationActivatedEventArgs.cpp
loneursid: Result is a pair of <hstring, hstring. As fas I know there is anything else we can put in this to indicate no value. insert() insists on an hstring won't accept a nullptr.

And since this is part of the IDL, we can really change it to an optional<>

dev/AppNotifications/AppNotificationActivatedEventArgs.cpp
loneursid: Yes, this can be improved. We'll address this in the feature level nits PR that will go in later today.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilderUtility.h
```diff
@@ -16,6 +20,24 @@ namespace AppNotificationBuilder
     using namespace winrt::Microsoft::Windows::AppNotifications::Builder;
 }
 
+inline std::unordered_map<wchar_t, std::wstring> GetXmlEscapeEncodings()
```
hulumane, line 23: Why return the whole array? Instead, let this API take in a char as an argument and return a string.

Call it 
std::wstring GetXmlEscapeEncoding(const wchar_t& charToEncode)

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilderUtility.h
```diff
@@ -16,6 +20,24 @@ namespace AppNotificationBuilder
...
+{
+    static std::unordered_map<wchar_t, std::wstring> encodings = { { L'&', L"&amp;"}, { L'\"', L"&quot;"}, {L'<', L"&lt;"}, {L'>', L"&gt;"}, {L'\'', L"&apos;"}};
+    return encodings;
+}
+
+inline std::unordered_map<wchar_t, std::wstring> GetPercentEncodings()
+{
+    static std::unordered_map<wchar_t, std::wstring> encodings = {{ L'%', L"%25"}, {L';', L"%3B"}, {L'=', L"%3D"} };
```
hulumane, line 31: Just put this in the first map. Why do we need a second API?

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilderUtility.h
```diff
@@ -16,6 +20,24 @@ namespace AppNotificationBuilder
...
+    return encodings;
+}
+
+inline std::unordered_map<std::wstring, wchar_t> GetPercentEncodingsReverse()
+{
+    static std::unordered_map<std::wstring, wchar_t> encodings = { { L"%25", L'%' }, {L"%3B", L';' }, { L"%3D", L'=' } };
```
hulumane, line 37: Same comment as above here. Why return a list when you can just return a char and the API takes in an encodedString

dev/AppNotifications/AppNotificationActivatedEventArgs.cpp
loneursid: Good catch the two extra c_str() are not needed here. We'll add that to the list of things to address today

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -31,7 +31,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
     {
         THROW_HR_IF_MSG(E_INVALIDARG, key.empty(), "You must provide a key when adding an argument");
 
-        m_arguments.Insert(key, value);
+        m_arguments.Insert(EncodeArgument(key.c_str()), EncodeArgument(value.c_str()));
```
hulumane, line 34: Here too. If you can pass this by reference here, you can avoid creating unnecessary copies

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -64,7 +64,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
     {
         THROW_HR_IF_MSG(E_INVALIDARG, m_textLines.size() >= c_maxTextElements, "Maximum number of text elements added");
 
-        m_textLines.push_back(wil::str_printf<std::wstring>(L"<text>%ls</text>", text.c_str()).c_str());
+        m_textLines.push_back(wil::str_printf<std::wstring>(L"<text>%ls</text>", EncodeXml(text).c_str()).c_str());
```
hulumane, line 67: Why c_str()? m_textLines is simply a vector of strings right? No need to convert to raw strings.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -73,7 +73,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
         THROW_HR_IF_MSG(E_INVALIDARG, m_textLines.size() >= c_maxTextElements, "Maximum number of text elements added");
 
         std::wstring props{ properties.as<winrt::Windows::Foundation::IStringable>().ToString() };
-        m_textLines.push_back(wil::str_printf<std::wstring>(L"%ls%ls</text>", props.c_str(), text.c_str()).c_str());
+        m_textLines.push_back(wil::str_printf<std::wstring>(L"%ls%ls</text>", props.c_str(), EncodeXml(text).c_str()).c_str());
```
hulumane, line 76: Same. Why c_str()?

hulumane (approved):

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilderUtility.h
pmpurifoy: We need the second API because the only use for percent encoding is for arguments. %;= don't need any special encoding for anything else.

dev/AppNotifications/AppNotificationActivatedEventArgs.cpp
pmpurifoy: There can only be one '=' per token. Find is simpler/better here since we aren't expecting more than 1 delimiter.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
pmpurifoy: Changing EncodeArgument param from std::wstring -> winrt::hstring to avoid this

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
pmpurifoy: c_str() is needed in str_printf as the param, but the ending c_str() is unnecessary. Fixing for all instances of this.

dev/AppNotifications/AppNotificationActivatedEventArgs.cpp
pmpurifoy: I don't think you can use optional in midl 3, but I think it's fine to have an empty "" value. Developers doing AddArgument(L"Key", L""), is valid. The toolkit allows this behavior as well. In the future, we can add AddArgument(key) for a more explicit "I want an empty value".

PR #2814, opened 2022-08-03
Fix issues with timestamp in AppNotificationBuilder

pmpurifoy:


test/AppNotificationBuilderTests/APITests.cpp
```diff
@@ -54,16 +54,19 @@ namespace Test::AppNotification::Builder
...
+            auto timeAsTimeT{ winrt::clock::to_time_t(time) };
+            localtime_s(&dateTime, &timeAsTimeT);
 
-            auto builder{ AppNotificationBuilder().SetTimeStamp(timeStamp) };
-            auto expected{ L"<toast displayTimestamp='2016-12-22T09:12:10Z'><visual><binding template='ToastGeneric'></binding></visual></toast>" };
+            std::wstringstream buffer;
+            buffer << std::put_time(&dateTime, L"%FT%T%z");
+
+            std::wstring expectedTimestamp{ buffer.str() };
+            expectedTimestamp.insert(expectedTimestamp.size() - 2 /* offset center index */, L":");
```
loneursid, line 68: did you mean to use this c_offsetIndexValue instead of literal 2? Or is it intentional becuse the unit test tries to prevent the value from accidentally changing?

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -38,13 +38,15 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
...
+        struct tm buf {};
+        localtime_s(&buf, &seconds);
 
         std::wstringstream buffer;
-        buffer << std::put_time(&buf, L"%FT%T");
+        buffer << std::put_time(&buf, L"%FT%T%z");
 
-        m_timeStamp = wil::str_printf<std::wstring>(L" displayTimestamp='%lsZ'", buffer.str().c_str());
+        std::wstring expectedTimestamp{ buffer.str() };
+        expectedTimestamp.insert(expectedTimestamp.size() - c_offsetIndexValue, L":");
```
loneursid, line 48: expectedTimestamp ? the "expected" part makes it feel like we're in a unit test

loneursid (changes requested):

test/AppNotificationBuilderTests/APITests.cpp
pmpurifoy: I found out I can include the header in the unit test and use the constant. Will fix.

loneursid: Actually it might not be a bad idea to use the literal in the test, depending on what you are testing for. It may be a good thing that the test breaks whenever someone changes that value....

pmpurifoy: Going to go ahead and use the constant since that was the original intention.

loneursid (approved):
Ship it

PR #2834, opened 2022-08-08
Publish New AppNotificationBuilder API

loneursid:
- Add projection project to build the AppNotification winmd file.
- Copy the new AppNotificationBuilder winmd to the publishing directory
- Publish the new builder API in build/NuSpecs/AppxManifest.xml
- Add contracts to the API in the IDL

I've tested by downloading the nuget, installing it in two of our samples (one C++, one C#) and using the new APIs. I can confirm that the new APIs can be used in the samples and the samples build and run as expected.


dev/Projections/CS/Microsoft.Windows.AppNotifications.Projection/Microsoft.Windows.AppNotifications.Projection.csproj
```diff
@@ -39,6 +39,10 @@
       <HintPath>$(OutDir)..\WindowsAppRuntime_DLL\StrippedWinMD\Microsoft.Windows.AppNotifications.winmd</HintPath>
       <IsWinMDFile>true</IsWinMDFile>
     </Reference>
+	  <Reference Include="Microsoft.Windows.AppNotifications.Builder">
```
loneursid, line 42: Initially, I wanted to follow the same pattern that other APIs have been following and create a new project for the builder but the builder depends on AppNotification and I couldn't figure out how to make that work with two separate projects.

The thing is anyway, there is no point in using the builder without AppNotification as all it can do is build new AppNotification instances.

kythant (approved):

dev/Projections/CS/Microsoft.Windows.AppNotifications.Builder.Projection/Microsoft.Windows.AppNotifications.Builder.Projection.csproj
```diff
@@ -0,0 +1,48 @@
...
+    <DebugType>pdbonly</DebugType>
+    <DebugSymbols>true</DebugSymbols>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <CsWinRTInputs Include="$(OutDir)/**/*.winmd" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Reference Include="Microsoft.Windows.AppNotifications">
```
loneursid, line 38: I was hoping that ia reference to Microsoft.Windows.AppNotifications will fix the build issue. I did fix the initial error, but this is still not good enough.

jonwis:
We should talk about how to test this e2e - is there a test app in the repo that would have consumed this correctly via the nuspec?

There needs to be one in the Samples repo that shows a great use, so we'd have that as a cross-check.

loneursid:
> We should talk about how to test this e2e - is there a test app in the repo that would have consumed this correctly via the nuspec?
> 
> There needs to be one in the Samples repo that shows a great use, so we'd have that as a cross-check.

I've used the samples to test and verify this and I can confirm that the sample builds with the new APIs. The issue is that the pipelines only produce a nuget, no installer or the 4 MSIX needed, so it can't run. The nightly build produces everything, but I'd need to merge this first in order to get the binaries.

loneursid:
> > We should talk about how to test this e2e - is there a test app in the repo that would have consumed this correctly via the nuspec?
> > There needs to be one in the Samples repo that shows a great use, so we'd have that as a cross-check.
> 
> I've used the nuget (downloaded from the output of the build pipeline) and samples to test and verify this and I can confirm that the sample builds with the new APIs. The issue is that the pipelines only produce a nuget, no installer or the 4 MSIX needed, so it can't run. The nightly build produces everything, but I'd need to merge this first in order to get the binaries.

@jonwis - I was wrong about testing. Thaks to @kythant for setting me up straight. I've been able to test the feature e2e using   our samples. The msix were produced by the pipelines. I was just not looking at the right place.

dev/Projections/CS/Microsoft.Windows.AppNotifications.Builder.Projection/Microsoft.Windows.AppNotifications.Builder.Projection.csproj
```diff
@@ -0,0 +1,52 @@
...
+    <DebugType>pdbonly</DebugType>
+    <DebugSymbols>true</DebugSymbols>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <CsWinRTInputs Include="$(OutDir)/**/*.winmd" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Microsoft.Windows.AppNotifications.Projection\Microsoft.Windows.AppNotifications.Projection.csproj" />
```
Scottj1s, line 38: I misunderstood from the email thread. So you have the project reference, but it only picks up the projection assembly and not the metadata?  Odd.  We can debug with a binlog at some point, but the Reference below is fine for now.

Scottj1s (approved):

dev/Projections/CS/Microsoft.Windows.AppNotifications.Builder.Projection/Microsoft.Windows.AppNotifications.Builder.Projection.csproj
loneursid: What I was trying to say is that after adding the dependency on the project, I was expecting that this wouldn't be necessary:
```
<Reference Include="Microsoft.Windows.AppNotifications">
<HintPath>$(OutDir)..\WindowsAppRuntime_DLL\StrippedWinMD\Microsoft.Windows.AppNotifications.winmd</HintPath>
<IsWinMDFile>true</IsWinMDFile>
</Reference>
```

but it turns out it is still needed.

DrusTheAxe:
e2e test -- that's what PRojectReunionInternal's \test<https://dev.azure.com/microsoft/projectreunion/_git/ProjectReunionInternal?path=/test> dirtree is for: IntegratedTests. All features should have coverage.

The issue is that the pipelines only produce a nuget, no installer or the 4 MSIX needed, so it can't run.
Which pipelines<https://dev.azure.com/Microsoft/projectreunion/_build>?

If you have the nuget you don't need the installer. The nuget contains all the MSIX
[cid:2ade13d9-5467-49ee-9244-7da371601b8d]
Unzip that and install them e.g.

powershell -c add-appxpackage -ForceTargetApplicationShutdown %TEMP%\wherever-the-nupkg-lives\tools\MSIX\win10-x64\*.msix

     - Howard
________________________________

PR #2835, opened 2022-08-09
DeploymentManager auto-initializer

DrusTheAxe:
Add an auto-initializer for DeploymentManager (very similar to the Bootstrapper's auto-initializer).

If DeploymentManager Initialize fails...
...Write to the eventlog

...Optionally
* DebugBreak
* DebugBreak if a debugger is attached to the process
* FailFast
* MessageBox (modeled after .NET's) asking if you'd like to install the WindowsAppRuntime and if so, ShellExecute(url) to the download page

The optional behavior is triggered by new `DeploymentInitializeOptions` properties or by environment variables
```
SET MICROSOFT_WINDOWSAPPRUNTIME_DEPLOYMENTMANAGER_INITIALIZE_DEBUGBREAK=1
SET MICROSOFT_WINDOWSAPPRUNTIME_DEPLOYMENTMANAGER_INITIALIZE_SHOWUI=1
SET MICROSOFT_WINDOWSAPPRUNTIME_DEPLOYMENTMANAGER_INITIALIZE_FAILFAST=1
```

Testing partial but incomplete as I need a real build (franken-hacking has been too messy to finish E2E testing). The DeploymentManager auto-initializer is inert/unused unless WindowsAppSdkDeploymentManagerInitialize=true and the default to enable that is commented out for now, making this PR functional but inert so I can get a build turned with real packages for further testing.

*.targets updated to provide build-time control.

* **To disable the auto-initializer** the auto-initialize set the property `WindowsAppSDKDeploymentManagerAutoInitialize`=`false`
* **If the auto-initalizer is enabled** you can define properties to control the options passed via `DeploymentInitializeOptions` to `DeploymentManager.Initialize()`
  * **WindowsAppSDKDeploymentManagerAutoInitializeOptions_Default=true** -- use the default options (OnError_ShowUI)
  * **WindowsAppSDKDeploymentManagerAutoInitializeOptions_None=true** -- use no options (None)
  * **WindowsAppSDKDeploymentManagerAutoInitializeOptions_*=true** -- enable the option where * is one or more of the following
    * OnError_DebugBreak
    * OnError_DebugBreak_IfDebuggerAttached
    * OnError_FailFast
    * OnError_ShowUI
    * OnNoPackageIdentity_NoOp

For example, to disable the auto-initializer add to your *proj file:

```xml
<PropertyGroup>
    <WindowsAppSDKDeploymentManagerAutoInitialize>false</WindowsAppSDKDeploymentManagerAutoInitialize>
</PropertyGroup>
```

or to enable the auto-initializer (default) but use the DebugBreak+FailFast options (instead of the default)

```xml
<PropertyGroup>
    <WindowsAppSDKDeploymentManagerAutoInitializeOptions_OnError_DebugBreak>true</WindowsAppSDKDeploymentManagerAutoInitializeOptions_OnError_DebugBreak>
    <WindowsAppSDKDeploymentManagerAutoInitializeOptions_OnError_ShowUI>true</WindowsAppSDKDeploymentManagerAutoInitializeOptions_OnError_ShowUI>
</PropertyGroup>
```

https://task.ms/40724739


letao-dev (approved):

dev/Deployment/DeploymentManager.cpp
```diff
@@ -15,12 +14,25 @@
...
+        HRESULT hrInitialize,
+        const AppModel::Identity::PackageIdentity& packageIdentity,
+        const std::wstring& release) noexcept;
+
+    HRESULT Initialize_ShowUI_OnError(
+        const AppModel::Identity::PackageIdentity& packageIdentity,
+        const std::wstring& release);
+}
+
 inline void Initialize_StopSuccessActivity(
```
sachintaMSFT, line 29: With these parameter changes, the inline function should be named StopActivity instead of StopSuccessActivity.

dev/Deployment/DeploymentManager.cpp
```diff
@@ -155,7 +155,89 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
...
+                FAIL_FAST_HR_MSG(hr,
+                                 "DeploymentManager initialize for release %ls (MSIX package version %hu.%hu.%hu.%hu)",
+                                 release.c_str(),
+                                 version.Major, version.Minor, version.Build, version.Revision);
+            }
+            Initialize_StopSuccessActivity(initializeActivityContext, deploymentResult.Status(), hr);
+            THROW_HR(hr);
+        }
+
+        // Success!
```
sachintaMSFT, line 232: Looks like you are not logging success activity here for telemetry.

sachintaMSFT (approved):
Minor comments.

PR #2837, opened 2022-08-09
AppNotificationBuilder returning a payload instead of an AppNotification

loneursid:
Alternative to https://github.com/microsoft/WindowsAppSDK/pull/2834 in case no-one can figure out how to set dependencies between AppNotificationBuilder and AppNotification.


loneursid:
The issue this was addressing has been fixed.

PR #2838, opened 2022-08-09
Updating ToastNotificationDemoApp to use the new Builder

loneursid:
This isn't the right / final fix yet, so let's discuus.

One thing is certain though: the code currently checked-in is wrong.


loneursid:
New PR targeting develop instead.

PR #2839, opened 2022-08-09
Update ToastNotificationDemoApp to use the new Builder

loneursid:
This isn't the right / final fix yet, so let's discuus.

One thing is certain though: the code currently checked-in is wrong.


gabbybilka:
@loneursid What is the status of this doc change?

loneursid:
Closing as this is only nice to have work.

Issue #2564, opened 2022-06-02
ARM64 installer won't work on Windows 10


jaimecbernardo:
### Describe the bug

When running the arm64 installer executable on Windows 10, it tries to install a x64 package, which fails, since Windows 10 doesn't have x64 emulation.

### Steps to reproduce the bug

1. Download the arm64 installer. https://aka.ms/windowsappsdk/1.0/1.0.3/windowsappruntimeinstall-1.0.3-arm64.exe
2. Try to run it on a Windows 10 ARM64 machine.

### Screenshots

![GetAttachmentThumbnail](https://user-images.githubusercontent.com/26118718/171608057-7a16cbb3-6ad5-499b-8a0f-a378aafa8057.png)

### Packaging type

Unpackaged

### Windows version

Windows 10 version 21H2 (19044, November 2021 Update)

sachintaMSFT:
The fix for this issue will be serviced in Windows App SDK 1.1.5 that is due in September.

gabbybilka:
:tada:`Microsoft.WindowsAppSDK v1.1.5` has been released which fixes this issue.
		
Handy links:
* [Release Notes](https://docs.microsoft.com/windows/apps/windows-app-sdk/stable-channel#version-115)
* [Official Nuget Release](https://www.nuget.org/packages/Microsoft.WindowsAppSDK/1.1.5)

PR #2841, opened 2022-08-10
Don't install x64 framework package on Windows 10 ARM64 devices as it…

sachintaMSFT:
Don't install x64 framework package on Windows 10 ARM64 devices. Win 10 ARM64 doesn't support x64 apps.
Tested IsWindows10_21H2OrGreater API on Windows 10 and Windows 11 and works as expected.

And fixed an unused variable deploymentOptions in two places (in one place, it is used now. In another, it is removed altogether).

https://github.com/microsoft/WindowsAppSDK/issues/2564

### DIFF
--- a/dev/Common/IsWindowsVersion.h
+++ b/dev/Common/IsWindowsVersion.h
@@ -37,7 +37,7 @@ inline bool IsWindows10_20H1OrGreater()
 }
 inline bool IsWindows11_21H2OrGreater()
 {
-    // GetPackageInfo3() added to kernelbase.dll in NTDDI_WIN10_VB (aka 20H1)
+    // GetMachineTypeAttributes() added to kernelbase.dll in NTDDI_WIN10_CO (aka Windows 11 21H2)
     return IsExportPresent(L"kernelbase.dll", "GetMachineTypeAttributes");
 }
 }
--- a/dev/Common/IsWindowsVersion.h
+++ b/dev/Common/IsWindowsVersion.h
@@ -35,6 +35,11 @@ inline bool IsWindows10_20H1OrGreater()
     // GetPackageInfo3() added to kernelbase.dll in NTDDI_WIN10_VB (aka 20H1)
     return IsExportPresent(L"kernelbase.dll", "GetPackageInfo3");
 }
+inline bool IsWindows11_21H2OrGreater()
+{
+    // GetPackageInfo3() added to kernelbase.dll in NTDDI_WIN10_VB (aka 20H1)
+    return IsExportPresent(L"kernelbase.dll", "GetMachineTypeAttributes");
+}
 }
 
 #endif // __ISWINDOWSVERSION_H
--- a/installer/dev/MachineTypeAttributes.h
+++ b/installer/dev/MachineTypeAttributes.h
@@ -9,7 +9,7 @@ namespace MachineTypeAttributes
     // On systems running Windows 11, this inline function returns if given architecture is supported on the system in user mode.
     // On systems running down level Windows versions (ex: Windows 10), this inline function will return false, always.
     // NOTE: If in case, the GetMachineTypeAttributes API that is currently only available on Windows 11 get serviced to down level Windows versions as well,
-    // then on systems running down level Windows versions (ex: Windows 10), this inline function will function same as on Windows 11.
+    // then on systems running down level Windows versions (ex: Windows 10), this inline function will function the same as on Windows 11.
     inline bool IsWindows11_IsArchitectureSupportedInUserMode(USHORT architecture)
     {
         // GetMachineTypeAttributes() added to kernelbase.dll in NTDDI_WIN10_CO (aka Windows 11 21H2, the first market release version of Windows 11).
@@ -25,9 +25,7 @@ namespace MachineTypeAttributes
             if (!getMachineTypeAttributes)
             {
                 // For now, GetMachineTypeAttributes API is available only on Windows 11 (i.e. builds 22000+).
-                // TODO: If GetMachineTypeAttributes API gets serviced downlevel to Windows versions (ex: Windows 10), then we have to THROW_LAST_ERROR
-                // if GetProcAddressByFunctionDeclaration returned NULL, instead of logging last error and returning false here.
-                LOG_LAST_ERROR();
+                THROW_LAST_ERROR();
                 return false;
             }
             else
--- a/installer/dev/MachineTypeAttributes.h
+++ b/installer/dev/MachineTypeAttributes.h
@@ -6,28 +6,39 @@
 
 namespace MachineTypeAttributes
 {
-    using GetMachineTypeAttributesPtr = HRESULT(WINAPI*)(USHORT Machine, MACHINE_ATTRIBUTES* MachineTypeAttributes);
-
     // On systems running Windows 11, this inline function returns if given architecture is supported on the system in user mode.
     // On systems running down level Windows versions (ex: Windows 10), this inline function will return false, always.
+    // NOTE: If in case, the GetMachineTypeAttributes API that is currently only available on Windows 11 get serviced to down level Windows versions as well,
+    // then on systems running down level Windows versions (ex: Windows 10), this inline function will function same as on Windows 11.
     inline bool IsWindows11_IsArchitectureSupportedInUserMode(USHORT architecture)
     {
         // GetMachineTypeAttributes() added to kernelbase.dll in NTDDI_WIN10_CO (aka Windows 11 21H2, the first market release version of Windows 11).
         // So, if that API is available, it indicates current system is running Windows 11.
         wil::unique_hmodule kernelbaseDll{ LoadLibraryExW(L"kernelbase.dll", nullptr, 0) };
-        if (kernelbaseDll)
+        if (!kernelbaseDll)
         {
-            // GetMachineTypeAttributes API is available only on Windows 11 (i.e. builds 22000+)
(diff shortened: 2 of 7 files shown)


dev/Common/IsWindowsVersion.h
```diff
@@ -35,6 +35,11 @@ inline bool IsWindows10_20H1OrGreater()
     // GetPackageInfo3() added to kernelbase.dll in NTDDI_WIN10_VB (aka 20H1)
     return IsExportPresent(L"kernelbase.dll", "GetPackageInfo3");
 }
+inline bool IsWindows11_21H2OrGreater()
+{
+    // GetPackageInfo3() added to kernelbase.dll in NTDDI_WIN10_VB (aka 20H1)
```
DrusTheAxe, line 40: Please update comment to match code

installer/dev/install.cpp
```diff
@@ -191,9 +189,22 @@ namespace WindowsAppRuntimeInstaller
             return true;
         }
 
-        // On Arm64 systems, all current package architectures are applicable.
+        // On Windows 11 (i.e. builds 22000+) ARM64 systems, all framework package architectures are applicable.
+        // On Windows 10 (i.e. builds 17763-190**) ARM64 systems (which don't support X64 apps), all but x64 framework package architecture is not applicable.
```
DrusTheAxe, line 193: `all but x64 framework package architecture is not applicable`

Odd phrasing, had to re-read a few times to be sure I understood it

SUGGESTION: `..., all framework package architectures are applicable except x64.`

That reads like the Win11 line above "...except x64"

installer/dev/install.cpp
```diff
@@ -191,9 +189,22 @@ namespace WindowsAppRuntimeInstaller
...
+            if (packageProperties->architecture == ProcessorArchitecture::X64)
+            {
+                if (WindowsVersion::IsWindows11_21H2OrGreater())
+                {
+                    return true;
+                }
+                else
+                {
+                    return false;
+                }
```
DrusTheAxe, line 205: Replace lines 198-205 with `return WindowsVersion::IsWindows11_21H2OrGreater();`

No need for the if ...return true else return false.

DrusTheAxe (approved):
LGTM modulo some minor details

riverar:
This seems very fragile. If downlevel gets serviced with these APIs, will all of this behave as expected?

installer/dev/MachineTypeAttributes.h
```diff
@@ -0,0 +1,34 @@
...
+{
+    using GetMachineTypeAttributesPtr = HRESULT(WINAPI*)(USHORT Machine, MACHINE_ATTRIBUTES* MachineTypeAttributes);
+
+    // On systems running Windows 11, this inline function returns if given architecture is supported on the system in user mode.
+    // On systems running down level Windows versions (ex: Windows 10), this inline function will return false, always.
+    inline bool IsWindows11_IsArchitectureSupportedInUserMode(USHORT architecture)
+    {
+        // GetMachineTypeAttributes() added to kernelbase.dll in NTDDI_WIN10_CO (aka Windows 11 21H2, the first market release version of Windows 11).
+        // So, if that API is available, it indicates current system is running Windows 11.
+        wil::unique_hmodule kernelbaseDll{ LoadLibraryExW(L"kernelbase.dll", nullptr, 0) };
```
DrusTheAxe, line 17: where's the thrown exception if LoadLibrary fails?

installer/dev/MachineTypeAttributes.h
```diff
@@ -0,0 +1,34 @@
...
+
+#include "pch.h"
+
+namespace MachineTypeAttributes
+{
+    using GetMachineTypeAttributesPtr = HRESULT(WINAPI*)(USHORT Machine, MACHINE_ATTRIBUTES* MachineTypeAttributes);
+
+    // On systems running Windows 11, this inline function returns if given architecture is supported on the system in user mode.
+    // On systems running down level Windows versions (ex: Windows 10), this inline function will return false, always.
+    inline bool IsWindows11_IsArchitectureSupportedInUserMode(USHORT architecture)
```
DrusTheAxe, line 13: NIT: Logic would be easier to read if it was written as fail/return/end-on-failure e.g.

```
dll = LoadLibrary
if (fail) THROW_BLAH(...)

p = GetProcAddress(...)
if (!p) THROW_LAST_ERROR()

THROW_IF_FAILED(getMachineTypeAttributes(architecture, &machineAttributes));
return WI_IsFlagSet(machineAttributes, MACHINE_ATTRIBUTES::UserEnabled);
```

installer/dev/MachineTypeAttributes.h
```diff
@@ -0,0 +1,34 @@
...
+    {
+        // GetMachineTypeAttributes() added to kernelbase.dll in NTDDI_WIN10_CO (aka Windows 11 21H2, the first market release version of Windows 11).
+        // So, if that API is available, it indicates current system is running Windows 11.
+        wil::unique_hmodule kernelbaseDll{ LoadLibraryExW(L"kernelbase.dll", nullptr, 0) };
+        if (kernelbaseDll)
+        {
+            // GetMachineTypeAttributes API is available only on Windows 11 (i.e. builds 22000+)
+            auto getMachineTypeAttributes{ reinterpret_cast<GetMachineTypeAttributesPtr>(GetProcAddress(kernelbaseDll.get(), "GetMachineTypeAttributes")) };
```
DrusTheAxe, line 21: WIL has a  cleaner way of doing this ==> `GetProcAddressByFunctionDeclaration`

```
const auto getMachineAttributes{ GetProcAddressByFunctionDeclaration(kernelbaseDll.get(), GetMachineTypeAttributes) };
```

SeeWIL's win32_helpers.h (end of the file). Pretty nifty

installer/dev/WindowsAppRuntimeInstall.vcxproj
```diff
@@ -42,7 +42,8 @@
   </ItemGroup>
   <PropertyGroup Label="Configuration">
     <ConfigurationType>Application</ConfigurationType>
-    <PlatformToolset>v142</PlatformToolset>
+    <PlatformToolset>v143</PlatformToolset>
+    <PlatformToolset Condition="'$(VisualStudioVersion)' == '16.0'">v142</PlatformToolset>
```
DrusTheAxe, line 46: Why?

Are you using VS2022? v143 smells like it. Most everything uses v143 (and no conditional logic)

sachintaMSFT: Yes, I am using VS2022. v143 is to accommodate that. Most of WindowsAppRuntime dll projects all use v142 and some are using the conditional logic. I went with conditional logic just to be like that some! But yes, I am fine with just v143.

sachintaMSFT:
> This seems very fragile. If downlevel gets serviced with these APIs, will all of this behave as expected?
Even if suppose the GetMachineTypeAttributes get serviced downlevel, it will still work as expected. Because I am not only checking if the API is available BUT we are also calling the API to check applicability of a given architecture compatibility on the device.

installer/dev/WindowsAppRuntimeInstall.vcxproj
DrusTheAxe: No, most don't

* # referencing v142 = 65
* # referencing v143 = 2 (RestartAgent and Microsoft.Windows.ApplicationModel.Resources) and both conditionally

@kythant is working on the plan to convert us to building with VS2022 but today we (still) require VS2019 to build WinAppSDK

installer/dev/MachineTypeAttributes.h
sachintaMSFT: That's right, I agree with Throwing Load Library failure. But I shouldn't be throwing if the API isn't available because we use that currently to determine then that the system is Windows 10 and shouldn't install x64 packages on arm64. I will instead log the last error and add the error to existing telemetry in a separate field. Thanks for the cue.

By the way, IsExportPresent API in IsWindowsVersion.h follows the same pattern and doesn't throw any errors in loading library or not finding the API.

installer/dev/MachineTypeAttributes.h
sachintaMSFT: This is cool. Didn't know. Thanks for the pointer.

installer/dev/MachineTypeAttributes.h
DrusTheAxe: _But I shouldn't be throwing if the API isn't available_
Right, but it should on errors. GetProcAddress returns null in both cases. We should distinguish Success-vs-Fail-vs-Error e.g.

```
if (!p)
{
     const auto lastError{ GetLastError() };
     THROW_HR_IF(HRESULT_FROM_WIN32(lastError), lastError!=i_forget_the_error_code_for_not_found)
}
```

>By the way, IsExportPresent API in IsWindowsVersion.h follows the same pattern and doesn't throw any errors in loading library or not finding the API.

Cool. That would be a bug :-)

An oversight when that was created. Feel free to improve it with the same sort of error logic you add in here

installer/dev/WindowsAppRuntimeInstall.vcxproj
sachintaMSFT: Ok, I am also saying the same thing that v142 is used most currently. But I learnt now that you and Will are co-ordinating to swtich to using VS 2022 and v143 soon. So I will remove the change to v143 altogether even with conditional usage of v142 and let you both do the switch for all projects at once. Thanks.

PR #2849, opened 2022-08-11
Add Repair option to WinAppSDK Installer

sachintaMSFT:
1. Add repair option to WindowsAppSDK Installer
2. Rectify logic to avoid installing licenses when --msix option is used
3. Add installing Singleton license in local Installer build as well
4. Update Framework TestPackage project for Installer to add/include singleton license file in it's msix package
5. Optimize code at few places


installer/dev/install.cpp
```diff
@@ -277,6 +277,23 @@ namespace WindowsAppRuntimeInstaller
             return;
         }
 
+        PackageManager packageManager;
+        const auto package{ packageManager.FindPackageForUser(L"",packageProperties->fullName.get()) };
+        if (package)
+        {
+            if (WI_IsFlagClear(options, WindowsAppRuntimeInstaller::Options::RepairPackages))
+            {
+                std::wcout << packageProperties->fullName.get() << " package is already installed. Skipping re-installing it." << std::endl;
```
DrusTheAxe, line 286: The message says "re-install" but the option is "repair". The _mechanism_ to repair is re-install but that's an implementation detail

No need for multiple separate print statements

No need to mention both short and long form options. Just mention long, and --help explains the variants (if any)

Narrow strings with wcout is usually requires it do narrow->wide conversion to output. Generally do everything with wide strings.

SUGGESTION:

```
std::wcout << packageProperties->fullName.get() << L" package is already installed." << std::endl
           << L"INFO: If you want to repair a package use the --repair option." << std::endl;
```

Or if you want to go old school

```
wprintf(L"%s package is already installed.\n"
        L"INFO: If you want to repair a package use the --repair option.\n",
        packageProperties->fullName.get());

File: installer/dev/install.h [hunk 2]
```diff
@@ -15,6 +15,7 @@ namespace WindowsAppRuntimeInstaller {
         Quiet               = 0x0004,
         DryRun              = 0x0008,
         ForceDeployment     = 0x0010,
+        RepairPackages      = 0x0020,
```
DrusTheAxe, line 18: Does this only repair packages? I didn't see an option `--repair-packages`

SUGGESTION: Rename `Repair`

DrusTheAxe (approved):
LGTM modulo comments

PR #2853, opened 2022-08-13
New activated log events Push and App Notifications

loneursid:
DCR https://dev.azure.com/microsoft/OS/_workitems/edit/41085624
- **Add telemetry to AppNotificationManager::Activate().** - The new telemetry captures when an App Notification is received and handled to the app in the foreground and in the background whether the app is packaged or unpackaged.

- **Add telemetry to track received Push Notifications path inside the WindowsAppSDK until it is handled to the app.**
    - Adds a telemetry provider to the long running process. The new provider is used to track push notifications destined to unpackaged apps while they make their way through the LRP. It can be extended in the future with other LRP telemetry as needed.
    - Adds correlation vector to the InvokeAll() method so the CV can be passed from the LRP into the SDK and logged.
    - Adds telemetry to track the path of Push Notifications in the SDK, after being received from the LRP (unpackaged apps) or the CoreOS (packaged apps).

**Also in this PR:**
- **Bump toolsets for two of our test projects to 143** - It seems all the projects (in dev and test) have been updated to using the latest toolset. This brings these two test projects in-line with the rest of Reunion.


dev/PushNotifications/PushNotificationsLongRunningTask.ProxyStub/NotificationsLongRunningProcess.idl
```diff
@@ -11,7 +11,7 @@ import "ocidl.idl";
 [pointer_default(unique)]
 interface IWpnForegroundSink : IUnknown
 {
-    HRESULT InvokeAll([in] ULONG length, [in, size_is(length)] byte* data, [out] BOOL* foregroundHandled);
+    HRESULT InvokeAll([in] ULONG length, [in, size_is(length)] byte* data, [in] LPCWSTR correlationVector, [out] BOOL* foregroundHandled);
```
danielayala94, line 14: Is there a chance that the caller of InvokeAll hits the older version of this API?

dev/PushNotifications/PushNotificationsLongRunningTask/PushNotificationLongRunningTaskTelemetry.h
```diff
@@ -0,0 +1,59 @@
...
+    (0x76c12936, 0x0ba7, 0x46ba, 0xbe, 0x2b, 0xce, 0x49, 0x2e, 0x4b, 0xcf, 0x01));
+
+class PushNotificationLongRunningTaskTelemetry : public wil::TraceLoggingProvider
+{
+    IMPLEMENT_TELEMETRY_CLASS(PushNotificationLongRunningTaskTelemetry, PushNotificationLongRunningTaskTelemetryProvider);
+
+public:
+    DEFINE_EVENT_METHOD(LogOnRawNotificationReceived)(
+            winrt::hresult hr,
+            std::wstring const& correlationVector) noexcept try
```
danielayala94, line 20: Nit: fix indentation

dev/PushNotifications/PushNotificationsLongRunningTask/PushNotificationsLongRunningTask.vcxproj
```diff
@@ -102,7 +102,7 @@
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
     <ClCompile>
       <PreprocessorDefinitions>WIN32;_DEBUG;_WINDOWS;%(PreprocessorDefinitions)</PreprocessorDefinitions>
-      <AdditionalIncludeDirectories>%(AdditionalIncludeDirectories);$(MSBuildThisFileDirectory);$(OutDir)..\PushNotificationsLongRunningTask.ProxyStub;$(OutDir)\..\WindowsAppRuntime_DLL;$(OutDir)..\WindowsAppSDK_BootstrapDLL</AdditionalIncludeDirectories>
+      <AdditionalIncludeDirectories>%(AdditionalIncludeDirectories);$(MSBuildThisFileDirectory);$(RepoRoot)\dev\WindowsAppRuntime_Insights;$(OutDir)..\PushNotificationsLongRunningTask.ProxyStub;$(OutDir)\..\WindowsAppRuntime_DLL;$(OutDir)..\WindowsAppSDK_BootstrapDLL</AdditionalIncludeDirectories>
```
danielayala94, line 105: What is WindowsAppRuntime_Insights? Has the DLL name changed?

dev/PushNotifications/PushNotificationsLongRunningTask.ProxyStub/NotificationsLongRunningProcess.idl
loneursid: This is also a concern of mine. I'd like to see if @DrusTheAxe has an opinion on this

dev/PushNotifications/PushNotificationsLongRunningTask/PushNotificationsLongRunningTask.vcxproj
loneursid: It's a new lib that I have to bring in for telemetry support. As far as I understand it's a close code library that is only available when the SDK is built in the pipepline.

I've tested that it works by building the pipleines and running our sample with Tracelog

dev/AppNotifications/AppNotificationTelemetry.h
```diff
@@ -251,6 +251,30 @@ class AppNotificationTelemetry : public wil::TraceLoggingProvider
...
+        if (m_telemetryHelper.ShouldLogEvent())
+        {
+            TraceLoggingClassWriteMeasure(
+                "Activated",
+                TelemetryPrivacyDataTag(PDT_ProductAndServicePerformance),
+                _GENERIC_PARTB_FIELDS_ENABLED,
+                TraceLoggingHexUInt32(hr, "OperationResult"),
+                TraceLoggingWideString(appId.c_str(), "AppId"),
+                TraceLoggingUInt32(arguments.size(), "ArgumentsSize"),
+                TraceLoggingBool(firstNotificationReceived, "firstNotificationReceived"),
```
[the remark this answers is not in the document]
loneursid: The pickier, the merrier :) 

Fixed, thanks

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -810,19 +816,33 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
 
-        return S_OK;
+        auto logTelemetry{ wil::scope_exit([&]() {
+            PushNotificationTelemetry::LogOnRawNotificationReceived(hr, WindowsGetStringRawBuffer(correlationVector, nullptr));
+        }) };
+
+        try
+        {
+            BOOL foregroundHandled = true;
+            THROW_IF_FAILED(InvokeAll(payloadLength, payload, WindowsGetStringRawBuffer(correlationVector, nullptr), &foregroundHandled));
```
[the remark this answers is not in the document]
loneursid: Much nicer, thanks.

dev/PushNotifications/PushNotificationsLongRunningTask/ForegroundSinkManager.cpp
```diff
@@ -17,15 +17,15 @@ void ForegroundSinkManager::Remove(std::wstring const& appId)
...
+bool ForegroundSinkManager::InvokeForegroundHandlers(std::wstring const& appId, winrt::com_array<uint8_t> const& payload, ULONG const& payloadSize, HSTRING correlationVector)
 {
     auto lock = m_lock.lock_exclusive();
 
     auto it = m_foregroundMap.find(appId);
     if (it != m_foregroundMap.end())
     {
         BOOL foregroundHandled = true;
-        if (FAILED(it->second->InvokeAll(payloadSize, payload.data(), &foregroundHandled)))
+        if (FAILED(it->second->InvokeAll(payloadSize, payload.data(), WindowsGetStringRawBuffer(correlationVector, nullptr), &foregroundHandled)))
```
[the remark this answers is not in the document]
loneursid: Much nicer, thanks.

dev/PushNotifications/PushNotificationsLongRunningTask.ProxyStub/NotificationsLongRunningProcess.idl
DrusTheAxe: You're changing `InvokeAll`'s footprint? Why? Seems like a Breaking Change.

When was InvokeAll created? 1.1.0?

Who's directly using the proxy? Who's directly using the stub?

If one side's in the Framework package and the other's in the Singleton then yes, you could have client using 1.1.0 framework and server in Singleton using 1.2.0 -- they expect to communicate but don't agree on the footprint. BadMojo(TM), no?

You could ADD a new `IWpnForegroundSink2` with the new footprint (though it's COM so no method overloading - `InvokeAll2()` or `InvokeAllWithCorrelationVector()` or...), support both (old+new) in the server and change the new client to use the new interface.

dev/PushNotifications/PushNotificationsLongRunningTask/PushNotificationLongRunningTaskTelemetry.h
loneursid: fixed

dev/PushNotifications/PushNotificationsLongRunningTask.ProxyStub/NotificationsLongRunningProcess.idl
loneursid: That's what I was affraid of. I had some hope that because it's not a public API and that the components are both part of / delivered with the SDK that maybe this wouldn't be breacking... but yeah, I'll add a new API and support both in the server.

thanks

DrusTheAxe: > I had some hope that because it's not a public API and that the components are both part of / delivered with the SDK

If they were both in the same package AND not exposed outward then yes. Client+Server would always be matched ('because you can't take half of package-v1 and half of packge-v2).

But PackagedCOM registers in the COM catalog so folks outside the package can use it. That means client+server aren't guaranteed to be in the same atomic distribution unit (package). And if all uses were you'd almost certainly use WinRT instead of PackagedCOM OOP Server -- if no need to share the server to clients outside the package then an OOP WinRT server would be equivalent but simpler.

>I'll add a new API and support both in the server.

Only caveat is if fwk v2 is installed but singleton is v1, your client call will get an error (probably E_NOINTERFACE). You may want to put an explicit check and logging and/or insights into your call if this happens to guide devs/users to recognize the issue and take corrective action.

For instance, if process is running with Singleton v1's identity and someone tries to AddPackage Singleton-v2 _without_ the 'force' option, we see the Singleton's in use so servicing would be disruptive and fail. Then an app built with 1.2 uses 1.2 fwk but tries to talk to 1.1 singleton and error.

For such logging I recommend a combination of WIL `LOG_IF_FAILED[_MSG]`, `TraceLoggingWrite` and write an event to the event log. See `MddBootstrapInitialize_Log()` in dev\WindowsAppRuntime_BootstrapDLL\MddBootstrap.cpp for an example.

DrusTheAxe: P.S. I'd considered making a function in `dev\common` for the common parts but RegisterEventSource+ReportEvent+DeregisterEventSource is so simple but highly variable parameters I didn't see much value. I guess you could create `dev\common\WindowsAppRuntime.EventLog.h`

```c++
namespace wil
{
    typedef unique_any_handle_null<decltype(&::DeregisterEventSource), ::DeregisterEventSource> unique_event_source;
}

namespace Microsoft::WindowsAppRuntime::EventLog
{
inline HRESULT Log(
    const HRESULT eventId,
    _In_ PCWSTR message,
    const WORD severity = EVENTLOG_ERROR_TYPE)
{
    wil::unique_event_source eventlog{ RegisterEventSourceW(nullptr, L"Windows App Runtime");
    RETURN_LAST_ERROR_NUL_NULL(eventlog);

    PCWSTR strings[2]{ L"Windows App Runtime", message };
    return LOG_IF_WIN32_BOOL_FALSE(ReportEventW(eventlog.get(), severity, 0, eventId, nullptr, ARRAYSIZE(strings), 0, strings, nullptr));
}
```

then you can simply call

```
const auto hr{ LOG_IF_FAILED(JustDoIt(x, y)) };
if (FAILED(hr))
{
    auto message{ wil::str_printf<unique_process_heap>(L"Do %s when %y = BadMojo(TM)", x, y) };
    LOG_IF_FAILED(Microsoft::WindowsAppRuntime::EventLog::Log(hr, message.get());
    RETURN_HR(hr);
}
```

loneursid: ok, so that's no fun. Making new version of the IWpnForegroundSink that has an InvokeAllWithCorrelationVector and supporting both implementation is easy enough but it doesn't stop there. The client registers this API has a sync with the server so the server can notify the client when a notification is received.

That means creating a new version of INotificationsLongRunningPlatform with a new version of RegisterForegroundActivator that take a IWpnForegroundSink2 (our new API that supports InvokeAllWithZCorrelationVector).

Which in turn means exposing that new INotificationsLongRunningPlatform to the Library.

loneursid: This is getting we too complicated and I don;t want to jeopardize the entire telemetry feature for InvovkeAll so,I have removed the changes to the IDL from this PR, and I will open a separate PR just for that.

danielayala94 (approved):

PR #2868, opened 2022-08-17
DisplayArea.GetFromDisplayId

deepthi-msft:
Spec for adding a new static method to WinAppSDK 1.2:
DisplayArea.GetFromDisplayId(DisplayId)

### DIFF
--- a/specs/Windowing/DisplayArea.md
+++ b/specs/Windowing/DisplayArea.md
@@ -1,0 +1,74 @@
+DisplayArea API update to facilitate easier use for developers working with DisplayIds
+===
+
+# Background
+Up until WinAppSDK 1.1, the only easy mapping between a
+[Microsoft.UI.DisplayId](https://docs.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.displayid?view=windows-app-sdk-1.1_)
+and a
+[Microsoft.UI.Windowing.DisplayArea](https://docs.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.windowing.displayarea?view=windows-app-sdk-1.1_)
+was if the developer started with a DisplayArea and wanted to find the DisplayId it maps to. If a
+developer wanted to find out what DisplayArea is configured for a given DisplayId they would have
+to enumerate **all** DisplayAreas using the static method
+[DisplayArea.FindAll()](https://docs.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.windowing.displayarea.findall?view=windows-app-sdk-1.1)
+and then compare each DisplayArea returned with the DisplayId for the display they were interested in.
+
+This feature is remedying that complexity by introducing another static method on the DisplayArea
+class that gets the DisplayArea for a given DisplayId if available.
+`DisplayArea.GetFromDisplayId(DisplayId)` returns **null** if the **DisplayId** cannot be matched
+to a valid displayArea.
+
+This matches the pattern we already have in place for the following apis:
+1. DisplayArea given a WindowId -
+ [DisplayArea.GetFromWindowId(WindowId, DisplayAreaFallBack)](https://docs.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.windowing.displayarea.getfromwindowid?view=windows-app-sdk-1.1)
+
+2. AppWindow given a WindowId -
+[AppWindow.GetFromWindowId(WindowId)](https://docs.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.windowing.appwindow.getfromwindowid?view=windows-app-sdk-1.1)
+
+# API Pages
+
+## DisplayArea.GetFromDisplayId static method
+
+This is the new static method added to WinAppSdk 1.2.
+
+The following example shows how an AppWindow could be configured to fully occupy the DisplayArea
+associated with the passed hmonitor/displayId:
+```csharp
+using Microsoft.UI;
+using Microsoft.UI.Windowing;
+using Windows.Graphics;
+
+void ExampleUsage_GetDisplayAreaForDisplay(AppWindow appWindow, IntPtr hmonitor)
+{
+    DisplayId displayId = Win32Interop.GetDisplayIdFromMonitor(hmonitor);
+    DisplayArea displayArea = DisplayArea.GetFromDisplayId(displayId);
+
+    if(displayArea != null)
+    {
+        RectInt32 rect = new RectInt32()
+        {
+            X = 0,
+            Y = 0,
+            Width = displayArea.WorkArea.Width,
+            Height = displayArea.WorkArea.Height
+        };
+        appWindow.MoveAndResize(rect, displayArea);
+    }
+}
+```
+
+# API Details
+
+```csharp
+namespace Microsoft.UI.Windowing
+{
+    [contract(Microsoft.Foundation.WindowsAppSDKContract, 1)]
+    runtimeclass DisplayArea
+    {
+        [contract(Microsoft.Foundation.WindowsAppSDKContract, 1.2)]
+        {
+            // Queries what DisplayArea is configured for the display identified by displayId.
+            static DisplayArea GetFromDisplayId(Microsoft.UI.DisplayId displayId);
+        }
+    }
+}
+```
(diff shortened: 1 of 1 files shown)


jonwis:
We're doing an "email review" of this - please leave your comments in the PR by 5pm Seattle time, 2022-08-19.  @rossnichols  and @deepthi-msft will address feedback in future updates.

(Note that we're still in progress getting the `Microsoft.UI.Windowing` type definitions in the repo - there's links to the exisitng types to review.)

specs/Windowing/DisplayArea.md
```diff
@@ -0,0 +1,74 @@
...
+was if the developer started with a DisplayArea and wanted to find the DisplayId it maps to. If a
+developer wanted to find out what DisplayArea is configured for a given DisplayId they would have
+to enumerate **all** DisplayAreas using the static method
+[DisplayArea.FindAll()](https://docs.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.windowing.displayarea.findall?view=windows-app-sdk-1.1)
+and then compare each DisplayArea returned with the DisplayId for the display they were interested in.
+
+This feature is remedying that complexity by introducing another static method on the DisplayArea
+class that gets the DisplayArea for a given DisplayId if available.
+`DisplayArea.GetFromDisplayId(DisplayId)` returns **null** if the **DisplayId** cannot be matched
+to a valid displayArea.
```
oldnewthing, line 18: Could there be multiple `DisplayArea`s that match the same `DisplayId`? For example, you might have one physical display split into two logical monitors, say, to simulate a Surface Duo.

Or maybe there's some new layout with a taskbar in the middle of the screen, splitting it into top and bottom display areas.

deepthi-msft: For WinAppSDK 1.2 we have a 1:1 mapping between DisplayId and DisplayArea, but this may change in the future:
- We don't want the developers to take a hard dependency on the 1:1 releationship, even though that is what it is today.
- All we promise is that the developer will get *a* DisplayArea for a DisplayId(when DisplayId is valid).

Hence updating the doc with the following statement:
"If more than one DisplayArea matches the DisplayId, the system chooses one deterministically."

FrankC-msft, line 18: nit, capitalization missing on displayArea referencing class name

FrankC-msft: This seems like it creates an idea of a "primary" DisplayArea for a given DisplayId.  (I know you said "deterministically", not "primary", but that's how I would interpret it in practice.)  Is that OK?

SeattlePseudoGenius: Depending on how likely it is that we will break this 1:1 mapping in the future, we may want to future-proof this method, and return back multiple display areas from the call (like an iterable or something). Otherwise, this method becomes incomplete and not particularly useful unless, as Frank mentioned, we introduce some concept of a "primary" display area which we deterministically always return. This feels like an unnecessary complication, and something we should avoid if possible.

deepthi-msft: In future if we have multiple displayAreas for a given displayId, then we would return the “Primary” DisplayArea for a given DisplayId. But there was a discussion, whether we should call it “Primary” or “Default” or “Main”, etc. So, we haven’t landed on the name yet. The main concern regarding naming it “Primary” was there is already a [DisplayArea.Primary](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fdocs.microsoft.com%2Fen-us%2Fwindows%2Fwindows-app-sdk%2Fapi%2Fwinrt%2Fmicrosoft.ui.windowing.displayarea.primary%3Fview%3Dwindows-app-sdk-1.1&data=05%7C01%7Cdemondre%40microsoft.com%7C7c5e1b9007a24433b7b408da855c30ad%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C637968928253414604%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=%2BgoYA4mhdmPpN76PCoR1f6SKin%2Fx9sZ7H19eHPWTs6A%3D&reserved=0) which signifies that it is the Primary monitors displayArea. So, not sure if the name “Primary” would confuse the developers.
 
FUTURE APIs:
In future if we have multiple displayAreas for a displayId in that case we plan to add DisplayArea.FindAll(DisplayId) method.
DisplayArea.GetFromDisplayId(DisplayId) – would return the “Primary” for the displayId.
DisplayArea.FindAll(DisplayId) method – would return the list of all the displayAreas for a given displayId. 

The DisplayArea.FindAll(DisplayId) follows a similar pattern to the following API:
•	[DisplayArea.FindAll()](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fdocs.microsoft.com%2Fen-us%2Fwindows%2Fwindows-app-sdk%2Fapi%2Fwinrt%2Fmicrosoft.ui.windowing.displayarea.findall%3Fview%3Dwindows-app-sdk-1.1&data=05%7C01%7Cdemondre%40microsoft.com%7C7c5e1b9007a24433b7b408da855c30ad%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C637968928253414604%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=U9ykAnylFkzbeVMcxDNbUMHj46Cq58%2BK1Yf4gevfooc%3D&reserved=0) – Retrieves a list of all displayAreas currently present in the system.

DrusTheAxe: Would `DisplayArea.GetFromDisplayId(DisplayId)` == `DispalyArea.FindAll(DisplayId).First()`?

specs/Windowing/DisplayArea.md
```diff
@@ -0,0 +1,74 @@
...
+
+# API Pages
+
+## DisplayArea.GetFromDisplayId static method
+
+This is the new static method added to WinAppSdk 1.2.
+
+The following example shows how an AppWindow could be configured to fully occupy the DisplayArea
+associated with the passed hmonitor/displayId:
+```csharp
```
jonwis, line 35: Nit - blank line between text and formatted blocks

specs/Windowing/DisplayArea.md
deepthi-msft: Right now, since it is 1:1 mapping DisplayArea.GetFromDisplayId(DisplayId) == DispalyArea.FindAll(DisplayId).First() would be true. But in future when we have multiple displayAreas for a displayId we would have to maintain the ordering of the list returned by DisplayArea.FindAll(DisplayId) to make sure the primary is the first in the list.

Not sure if that is a good pattern to follow. At least that is not how DisplayArea.FindAll() works. 

+ @ionvasilian / @tiberiu2 for visibility.

SeattlePseudoGenius: I suppose the point is that when you do move to having multiple display areas per display id, then the semantic of GetFromDisplayId becomes murky - it becomes unclear to developers which would be returned, and the entire API becomes a useless relic, superseded to whatever the FindAll() method returns. This is you should think carefully about how likely it is that this many-to-one relationship is likely to come about in the future. If it's even reasonably likely, I suggest we skip straight to introducing a version of this API that's compatible with that in the future.

oldnewthing: It may be that in the future, there is a monitor with two display areas of equal importance (e.g., Surface Duo left/right), so it's not like there's a "Primary" one. I think we can just sidestep the issue by not giving it a name at all.

"Returns a DisplayArea that corresponds to the specified DisplayId, if one exists. If more than one such DisplayArea exists, the system chooses one to return. This choice will remain consistent until the next display topology change."

I think it's useful to have a "let the system choose one for me" API, because that's all that most developers will care about anyway. "Give me a display area that I can use on this display. If there's more than one, just pick one for me. I don't want to have to deal with the complexity of many-to-one mappings until enough customers demand it." Going straight to FindAll creates complexity that nobody cares about today.

deepthi-msft: I like @oldnewthing's approach. @FrankC-msft and @SeattlePseudoGenius do you have any concerns about this approach?

FrankC-msft: This sounds good to me.  It's a good balance of "simplest API we need right now" and "future-proofed through documentation".

SeattlePseudoGenius: Yes, this sounds good to me as well.

deepthi-msft: Thanks! Updated the docs.

SeattlePseudoGenius (approved):

PR #2869, opened 2022-08-18
DeploymentManager spec update for the new Repair API

sachintaMSFT:
Deployment Manager Spec update to add the new Repair API for 1.3 release. Hence, incrementing the DeploymentContract to version 4.


specs/Deployment/DeploymentAPI.md
```diff
@@ -6,9 +6,9 @@ This is the spec for proposal
 The Windows App SDK (WinAppSDK) for packaged apps involves the framework, main, and singleton
 packages, which are all signed and published by Microsoft. The framework package contains the
 WinAppSDK binaries used at run time, and it is installed with the application. The main package
-contains out-of-process services that are brokered between apps, such as push notifications, and it
-is also needed for the framework to be serviced by the Microsoft Store. The singleton package
-supports a single long-running process that is brokered between apps for features like push
+contains out-of-process services that are brokered between apps, such as Dynamic Dependency data
+store, and it is also needed for the framework to be serviced by the Microsoft Store. The singleton
+package supports a single long-running process that is brokered between apps for features like push
 notifications.
```
DrusTheAxe, line 12: SUGGESTION: Remove the text about what the Main+Singleton do / why they exist, as that ages (badly). Doesn't really matter here what they do, just that they're needed, and thus DeploymentManager is too

specs/Deployment/DeploymentAPI.md
```diff
@@ -42,12 +42,13 @@ are already present on the system, or that the version present is the minimum ve
 the application. Instead, the Deployment API should be used to validate that the WinAppSDK is at the
 correct version and that all required packages are present.
 
-The Deployment API addresses the following two developer scenarios:
+The Deployment API addresses the following three developer scenarios:
```
DrusTheAxe, line 45: SUGGESTION: Remove 'three'. Kinda redundant with the actual list, and if the list changes (again) it's one less edit to have to make

specs/Deployment/DeploymentAPI.md
```diff
@@ -42,12 +42,13 @@ are already present on the system, or that the version present is the minimum ve
...
 
 1. Does this system have the min version of the WinAppSDK main and singleton packages required by my
    app?
 2. How do I install the WinAppSDK main and singleton packages, if they are not already present on
    the system?
+3. How do I repair the already installed WinAppSDK main and singleton pacakges, if needed ?
```
DrusTheAxe, line 51: TYPO: `pacakges`

specs/Deployment/DeploymentAPI.md
```diff
@@ -60,6 +61,14 @@ The API addresses #1 by a simple query of the version needed and a check for OS
     packageManagement restricted capability can use the Deployment API to get these packages
     installed.
 
+#3 has two possible answers.
+
+-   First, the developer can repair these packages directly through the app by reinstalling them, if
```
DrusTheAxe, line 66: NIT: Remove `First, `

specs/Deployment/DeploymentAPI.md
```diff
@@ -143,10 +152,8 @@ by using DeploymentInitializeOptions object and setting ForceDeployment option b
 this API. This option when set will shut down the application(s) associated with WinAppSDK packages,
 update the WinAppSDK packages and restart the application(s).
 
-// OPENISSUE: If this option is set when updating main package, then whether the out of process com
-server from the main package such as push Notifications needs to be explicitly restarted needs to be
-understood. If restart it is needed, then the API will handle explicitly restarting it. This is
-under investigation.
+If this option is set when updating singleton package, then the out of process com server from the
```
DrusTheAxe, line 155: NIT: `Singleton`

It's a proper noun

DrusTheAxe, line 155: NIT: `the Singleton`

DrusTheAxe, line 155: The fact the Singleton has a Packaged COM OOP server is an implementation detail du jour. Better to avoid such details that can age badly.

SUGGESTION: `If this option is set any processes running with the Main and Singleton packages' identities are explicitly restarted (if necessary) after the package updates complete.`

specs/Deployment/DeploymentAPI.md
```diff
@@ -179,12 +186,43 @@ they shut down, to refer to the updated framework package.
     }
 ```
 
+## Repair
+
+This API performs a status check, and if the main and singleton packages are already installed at
+the required version, it will attempt to repair those packages. All information about the version,
```
DrusTheAxe, line 192: So if the installer's got packages v1.2.3.4 and v1.2.3.4+ are already installed this will do a repair, but if v1.2.3.3 (or older) is installed this will...do its install work as usual?

specs/Deployment/DeploymentAPI.md
```diff
@@ -179,12 +186,43 @@ they shut down, to refer to the updated framework package.
...
+channel, and architecture needed are derived from the current WinAppSDK framework package.
+
+Since both the main _and_ the singleton packages may need to be repaired, it is possible that one
+package is repaired successfully but not the other. If so, the returned WindowsAppRuntimeStatus will
+contain the error, if one occurs during package install.
```
DrusTheAxe, line 197: This is unclear. There are 2 things to potentially repair (Main+Singleton). If we do Main and fail what happens? Stop and return error? Also do Singleton but then return the details of the 1st error (Main)? Or...

IOW is this a fail-on-first-error or best-effort-report-first-error?

specs/Deployment/DeploymentAPI.md
```diff
@@ -179,12 +186,43 @@ they shut down, to refer to the updated framework package.
...
+
+```C#
+    if (DeploymentManager.GetStatus().Status != DeploymentStatus.Ok)
+    {
+        // Repair does a status check, and if the status is OK it will attempt to get
+        // the WindowsAppRuntime into a good state by deploying packages. Unlike a simple
+        // status check, Repair can sometimes take several seconds to deploy the packages.
+        // These should be run on a separate thread so as not to hang your app while the
+        // packages deploy.
```
DrusTheAxe, line 206: Is this any different than Initialize()?

specs/Deployment/DeploymentAPI.md
```diff
@@ -179,12 +186,43 @@ they shut down, to refer to the updated framework package.
...
+
+This API performs a status check, and if the main and singleton packages are already installed at
+the required version, it will attempt to repair those packages. All information about the version,
+channel, and architecture needed are derived from the current WinAppSDK framework package.
+
+Since both the main _and_ the singleton packages may need to be repaired, it is possible that one
+package is repaired successfully but not the other. If so, the returned WindowsAppRuntimeStatus will
+contain the error, if one occurs during package install.
+
+```C#
```
DrusTheAxe, line 199: Is this meant to be an example using the API or pseudo-code demonstrating what Repair() does?

specs/Deployment/DeploymentAPI.md
```diff
@@ -179,12 +186,43 @@ they shut down, to refer to the updated framework package.
...
+    if (DeploymentManager.GetStatus().Status != DeploymentStatus.Ok)
+    {
+        // Repair does a status check, and if the status is OK it will attempt to get
+        // the WindowsAppRuntime into a good state by deploying packages. Unlike a simple
+        // status check, Repair can sometimes take several seconds to deploy the packages.
+        // These should be run on a separate thread so as not to hang your app while the
+        // packages deploy.
+        var repairTask = Task.Run(() => DeploymentManager.Repair());
+        repairTask.Wait();
```
DrusTheAxe, line 208: You're kicking the Repair call to a separate task and then immediately .Wait()'ing for it. That makes it kinda pointlerss, no? Is this the same as `var repairResult = DeploymentManager.Repair().get();` but faster / less overhead?

If the intent is to do other work while Repair is running this should be

```c#
var repairTask = ...
...do other work while the repair is running...
repairTask.Wait();
```

specs/Deployment/DeploymentAPI.md
sachintaMSFT: No. It is intended to not install anything new. I take the word "repair" means repair something that is already there. The API works only on already installed packages of the same WinAppSDK version that the framework package (that contains the current DM) belongs to.

In the scenario you asked, it will fail saying no packages installed for this version to be repaired.  Thanks for the cue, it makes sense to explicitly state that here.

Btw, do you disagree with my take on Repair ?

specs/Deployment/DeploymentAPI.md
sachintaMSFT: I copied the language from Initialize. That statement in the spec here seems to indicate best-effort-report-first error. But Initialize code behaves as fail-on-first error. The spec or the code itself needs correction to keep them both in sync. I propose to rectify the spec.

As for Repair, I think either both Initialize and Repair follow same method of fail-on-first error or best-effort-report-first-error. Hence, since Initialize follows the fail-on-first errormethod, I propose Repair should also follow the same fail-on-first error method.

specs/Deployment/DeploymentAPI.md
sachintaMSFT: No difference, here.

Only difference is in functionality that Initialize installs both packages even if either of them is not installed. Repair installs both packages only when both packages are already installed. So, their functionalities are mutually exclusive.

specs/Deployment/DeploymentAPI.md
sachintaMSFT: Example using the API

specs/Deployment/DeploymentAPI.md
sachintaMSFT: I think separate task was used so that a hang in .get() doesn't cause a hang in the main thread of the app. My understanding is .get() is supposed to wait indefinitely for deployment to complete the request. Hence, this extra care for the app.

specs/Deployment/DeploymentAPI.md
sachintaMSFT: Updated the language now to reflect fail-on-first error approach.

specs/Deployment/DeploymentAPI.md
DrusTheAxe: So when should I use --repair? Only when I know WinAppRuntime is installed but maybe broken?

If e.g. Main is registered for the user "windowsappruntimeinstall.exe --repair" won't do anything with it, even if Main's package status != OK?


Sounds kinda backwards. `Install` today is installing what's missing AND re-registering what's already present (which is a form of repair) whereas `Repair` only installs what's not installed. And both consider "installed" to mean "package is registered for user" regardless if it's actually working.

Further, "broken" has 2 degrees of "broken"

1. Known and tagged as 'broken' (PackageStatus != OK)
2. Not explicitly known to Windows as 'broken' but no less busted and needing repair (e.g. register package then delete some regkeys laid down by the install; package status can be OK but the install is busted and in need of repair)

Seems we need more permutations:

1. Install only what's not installed
2. Install what's not installed AND fix packages with package status != ok
3. Install everything (whether or not it's already installed)

Today you're saying 1=`--repair`, 3=`--install` (i.e. default / no --repair) and 2=doesn't exist

Thoughts?

specs/Deployment/DeploymentAPI.md
DrusTheAxe: As an example it makes more sense with the "...do other work..." comment between the task start + wait

Lacking that it's kinda pointless to do the task. This thread is going to kick off the task and then block for .Wait to return. Net result this thread will "hang"/block. No?

specs/Deployment/DeploymentAPI.md
sachintaMSFT: Per our offline discussion on this topic,
repair = Install everything (--repair option for installer)
initialize = Install everything, if even one of the WindowsAppSDK packages is not installed or is not in good state.

DrusTheAxe (approved):

DrusTheAxe (changes requested):
DON'T CHECKIN until API Review

specs/Deployment/DeploymentAPI.md
sachintaMSFT: Added "...do other work..." comment between the task start + wait

specs/Deployment/DeploymentAPI.md
sachintaMSFT: As per offline continuation of this conversation, Initialize and Repair APIs would work on highest servicing revision of framework package installed for given WindowsAppSDK version.

specs/Deployment/DeploymentAPI.md
```diff
@@ -312,6 +502,18 @@ namespace Microsoft.Windows.ApplicationModel.WindowsAppRuntime
...
         [overload("Initialize")]
         static DeploymentResult Initialize(Microsoft.Windows.ApplicationModel.WindowsAppRuntime.DeploymentInitializeOptions deploymentInitializeOptions);
+
+        /// Checks the status of the WindowsAppRuntime of the current package and attempts to
+        /// repair already installed WinAppSDK packages.
+        [contract(DeploymentContract, 3)]
+        static DeploymentResult Repair();
+
+        /// Checks the status of the WindowsAppRuntime of the current package and attempts to
+        /// repair already installed WinAppSDK packages, while applying the DeploymentInitializeOptions
```
sachintaMSFT, line 512: Comment here incorrectly refers to DeploymentInitializeOptions rather than DeploymentRepairOptions

specs/Deployment/DeploymentAPI.md
```diff
@@ -246,12 +264,164 @@ Here's the previous example extended with the `Show UI on error` option:
...
+This API will always attempt to repair WindowsAppRuntime regardless of it's state. For that reason,
+this API should be called only when deployment status is not OK to avoid an unecessary repair even
+when WindowsAppRuntime is in good state. All information about the version, channel, and
+architecture needed are derived from the current WinAppSDK Framework package.
+
+Since both the Main _and_ the Singleton packages will be repaired, it is possible that Main package
+repair has failed and Initialize returns with PackageRepairFailed status (OR) it is possible that
+Main package is repaired successfully but the Singleton package repair has failed and Initialize
+returns with repair has failed and Initialize returns with PackageRepairFailed status (OR) it is
+possible that status. In both cases, the returned WindowsAppRuntimeStatus will contain the error of
```
MikeHillberg, line 278: > possible that status

Looks like there was more sentence intended here

specs/Deployment/DeploymentAPI.md
```diff
@@ -3,101 +3,117 @@
...
 # Description
 
 When a packaged app using the WinAppSDK is deployed from the Store, it will be installed with the
-application's main package and the WinAppSDK framework package. The WinAppSDK main and singleton
+application's Main package and the WinAppSDK Framework package. The WinAppSDK Main and Singleton
 packages may not already be present on the system, or may not be at the version required by the
-application. The Deployment API enables a developer to check when the required WinAppSDK main and
-singleton packages are missing and get them installed.
+application. The Deployment API enables a developer to check when the required WinAppSDK Main and
+Singleton packages are missing and get them installed.
```
MikeHillberg, line 25: Reading through this, it sounds like all of the sample code shown in this spec should be included in all apps?

specs/Deployment/DeploymentAPI.md
```diff
@@ -143,18 +161,16 @@ by using DeploymentInitializeOptions object and setting ForceDeployment option b
 this API. This option when set will shut down the application(s) associated with WinAppSDK packages,
 update the WinAppSDK packages and restart the application(s).
```
MikeHillberg, line 162: If I don't set ForceDeployment true, the deploy will fail? And then the app should interact with the to say "on next reboot" or some such?

specs/Deployment/DeploymentAPI.md
```diff
@@ -246,12 +264,164 @@ Here's the previous example extended with the `Show UI on error` option:
...
+If this option is set when updating Singleton package, then the out of process com server from the
+Singleton package such as push Notifications is explicitly restarted.
+
+When the API is updating Framework package and ForceDeployment option is set, then all dependent
+packages that are NOT currently in use will be immediately re-installed to refer to the updated
+Framework package and all dependent packages that are currently in use will be re-installed, after
+they shut down, to refer to the updated Framework package.
+
+```C#
+    if (DeploymentManager.GetStatus().Status == DeploymentStatus.PackageInstallRequired)
```
FrankC-msft, line 327: This is in the DeploymentRepairOptions section, so this should be checking for == DeploymentStatus.PackageRepairRequired right?

specs/Deployment/DeploymentAPI.md
```diff
@@ -143,18 +161,16 @@ by using DeploymentInitializeOptions object and setting ForceDeployment option b
...
+If this option is set when updating Singleton package, then the out of process com server from the
+Singleton package such as push Notifications is explicitly restarted.
 
-When the API is updating framework package and ForceDeployment option is set, then all dependent
+When the API is updating Framework package and ForceDeployment option is set, then all dependent
 packages that are NOT currently in use will be immediately re-installed to refer to the updated
-framework package and all dependent packages that are currently in use will be re-installed, after
-they shut down, to refer to the updated framework package.
+Framework package and all dependent packages that are currently in use will be re-installed, after
+they shut down, to refer to the updated Framework package.
```
MikeHillberg, line 170: > after they shut down

Does this mean it will force the apps currently using the package to shut down, or does this mean that the Initialize() call could block indefinitely?

specs/Deployment/DeploymentAPI.md
```diff
@@ -246,12 +264,164 @@ Here's the previous example extended with the `Show UI on error` option:
...
+        if options.OnError_ShowUI
+        {
+            MessageBox(...)
+        }
+    }
+    return hr
+}
+```
+
+Here's the previous example extended with the `Show UI on error` option:
```
MikeHillberg, line 389: This example doesn't match the above logic; the logic shows UI, the example doesn't. If the UI built in to the API and explains to the user what the problem is? Or if the logic is correct and the app needs to show UI, it's awkward because they can't use WinAppSDK to show the UI

specs/Deployment/DeploymentAPI.md
```diff
@@ -285,15 +459,31 @@ namespace Microsoft.Windows.ApplicationModel.WindowsAppRuntime
         DeploymentInitializeOptions();
 
         /// Gets or sets a value that indicates whether the processes associated with the
-        /// WindowsAppSDK main and singleton packages will be shut down forcibly if they are
-        /// currently in use, when registering the WinAppSDK packages.
+        /// WinAppSDK Main and Singleton packages will be shut down forcibly if they are
+        /// currently in use, when initializing the WindowsAppRuntime.
```
MikeHillberg, line 463: So when you start my app, you might see some other apps disappear? Is there any kind of interaction with the user first?

specs/Deployment/DeploymentAPI.md
```diff
@@ -3,101 +3,117 @@
...
-The Deployment API addresses the following two developer scenarios:
+The Deployment API addresses the following developer scenarios:
 
-1. Does this system have the min version of the WinAppSDK main and singleton packages required by my
+1. Does this system have the min version of the WinAppSDK Main and Singleton packages required by my
    app?
-2. How do I install the WinAppSDK main and singleton packages, if they are not already present on
+2. How do I install the WinAppSDK Main and Singleton packages, if they are not already present on
    the system?
+3. How do I repair the already installed WinAppSDK Main and Singleton packages, if needed ?
```
FrankC-msft, line 46: Why do we need Repair to be a distinct action than Install/Initialize, instead of having developers [re]install the packages whenever they're "broken".  This is something of a breaking change since we'd previously told developers to Initialize for any status other than .OK, right?

The manual method to repair by "reinstalling from msdn downloads page" is identical for install and repair.  Why is the full-trust-app method different?

specs/Deployment/DeploymentAPI.md
DrusTheAxe: Yes, the UI shown is built into the API (it's currently a MessageBox())

specs/Deployment/DeploymentAPI.md
```diff
@@ -3,101 +3,117 @@
...
-current WinAppSDK framework package.
-
-Since both the main _and_ the singleton packages may need to be installed, it is possible that one
-package is installed successfully but not the other. If so, the returned WindowsAppRuntimeStatus
-will contain the error, if one occurs during package install. There will be no rollback of any
-successfully installed packages. See the
+current WinAppSDK Framework package.
+
+Since both the Main _and_ the Singleton packages may need to be installed, it is possible that Main
+package install has failed and Initialize returns with PackageInstallFailed state (OR) it is
```
FrankC-msft, line 112: What happens if I try to Repair() a package that's in PackageInstallFailed state?  That seems like a natural inclination ("I did Initialize() and it's in some error state now, I guess that means it's broken so I should fix it. Fix -> repair. Call Repair()").  Are we providing DeploymentResult properties that tell the developer they called the wrong method?

specs/Deployment/DeploymentAPI.md
sachintaMSFT: Yes. Need to check for  == DeploymentStatus.PackageRepairRequired

specs/Deployment/DeploymentAPI.md
```diff
@@ -3,101 +3,117 @@
...
 
-If a developer prefers to have the Microsoft Store automatically service the WinAppSDK framework
-package and if features like push notifications are desired, then the WinAppSDK main and singleton
-should also be deployed. Developers must not assume that the WinAppSDK main and singleton packages
-are already present on the system, or that the version present is the minimum version required by
-the application. Instead, the Deployment API should be used to validate that the WinAppSDK is at the
-correct version and that all required packages are present.
+If a developer prefers to have the Microsoft Store automatically service the WinAppSDK Framework
+package and if features like dynamic dependencies, push notifications are desired, then the
+WinAppSDK Main and Singleton should also be deployed. Developers must not assume that the WinAppSDK
```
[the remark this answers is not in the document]
sachintaMSFT: It meant "main and singleton packages also be deployed <apart from the framework package>". Yes, this is saying that Servicing and wanting to use pushNotifications are reasons to call Deployment Manager API to ensure Main and Singleton packages are installed and are in good state for use.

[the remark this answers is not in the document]
MikeHillberg: Should the PushNotification doc pages indicate that it's required that apps run this code?

Is it just PushNotifications, or also AppNotifications?

specs/Deployment/DeploymentAPI.md
sachintaMSFT: Yes, in all Packaged Apps.

specs/Deployment/DeploymentAPI.md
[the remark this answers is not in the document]
sachintaMSFT: Yes, PushNotifications doc should call out that packaged apps should call DeploymentManager Initialize API to enable them use Pushnotifications.

Just PushNotifications. No AppNotifications.

Specifically, Singleton package is necessary for using PushNotifications.

specs/Deployment/DeploymentAPI.md
sachintaMSFT: Intialize installs/re-installs WinAppSDK Main and Singleton packages only when needed (i.e. when one of the packages is not installed or if one of them is in identified bad state). Thus, avoiding the unnecessary perf overhead of the unconditional re-install in every launch of the WinAppSDK dependent Packaged app.

Repair, on the other hand, re-installs those packages unconditionally. This API is important because there are many bad states a package can be in but only a few of them could be identified by Initialize and fix. Hence, we need Developers to just able to call Repair in cases where the WinAppSDK features aren't working due to a bad state that Initialize couldn't identify and fix it.

specs/Deployment/DeploymentAPI.md
sachintaMSFT: Repair() essentially provides an unconditional core functionality of Initialize() - which is to install WindowsAppSDK Main and Singleton licenses and packages. So, if Repair() is tried when Initialize() fails, then it is effectively a retry of Initialize(). If the underlying issue for PackageInstallFailed is transient, then Repair() may succeed. Otherwise, it will continue to be PackageInstallFailed.

specs/Deployment/DeploymentAPI.md
sachintaMSFT: It will only force the Main, Singleton and DDLM packages to shut down in order to update Main, Singleton and DDLM packages (ex: PushNotifications would be shutdown as it's part of Singleton package) and the shutdown packages are restarted automatically by the OS platform after their updates are complete (and in cases such as PushNotifictions which would need explicit reboot as it's a com server, we do reboot it after the update in Initialize()/Repair()).

It doesn't shutdown the WinAppSDK framework package dependent packaged apps and installs the higher version of the framework package side by side without uninstalling the older version. It does update the WinAppSDK framework package dependent packaged apps (in their package graph) to point to higher version of the framework package, once they are shutdown by the user. Once all references to older version of the framework package are updated to refer to the higher version of it, the older version is then completely uninstalled by the OS platform.

specs/Deployment/DeploymentAPI.md
sachintaMSFT: No. Please see one of the above comments for detailed explanation on ForceDeployment. In short, ForceDeployment only shuts down WinAppSDK packages that are dependent on it's framework packages (ex: Main, Singleton and DDLM). It doesn't shut down any non-WinAppSDK packages that declared dependency on WinAppSDK framework package.

specs/Deployment/DeploymentAPI.md
```diff
@@ -246,12 +264,164 @@ Here's the previous example extended with the `Show UI on error` option:
...
 
+## Repair
+
+This API will always attempt to repair WindowsAppRuntime regardless of it's state. For that reason,
+this API should be called only when deployment status is not OK to avoid an unecessary repair even
+when WindowsAppRuntime is in good state. All information about the version, channel, and
+architecture needed are derived from the current WinAppSDK Framework package.
+
+Since both the Main _and_ the Singleton packages will be repaired, it is possible that Main package
+repair has failed and Initialize returns with PackageRepairFailed status (OR) it is possible that
```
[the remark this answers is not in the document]
sachintaMSFT: It is just "or". I should probably just use "or" to avoid such confusion.

sachintaMSFT, line 275: Here, "Initialize returns with PackageRepairFailed status" should actually be "Repair returns with PackageRepairFailed status".

specs/Deployment/DeploymentAPI.md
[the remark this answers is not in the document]
sachintaMSFT: It's actually just 2 cases only.

specs/Deployment/DeploymentAPI.md
MikeHillberg: What if an app is currently using Main/Singleton/DDLM package (and I set Force)? What happens to the app?

specs/Deployment/DeploymentAPI.md
```diff
@@ -3,101 +3,117 @@
...
-        // for the Status.
+        // WindowsAppRuntime is either not installed or not in a good state (can't be both at same time).
+        if (result.Status == DeploymentStatus.PackageInstallRequired)
+        {
+            // Should call Initialize on another thread, or not use features that require the
+            // WindowsAppRuntime. result.ExtendedError has the error code that has the reason
+            // for the Status.
+        }
+        else if (result.Status == DeploymentStatus.PackageRepairRequired ||
+                result.Status == DeploymentStatus.Unknown)
```
FrankC-msft, line 95: what if the status is PackageInstallFailed ?  Shouldn't I want to Repair() it in that case?  (Since repair is unconditional core functionality of Initialize, is PackageInstallFailed a special-case or subset of PackageRepairRequired?)

specs/Deployment/DeploymentAPI.md
```diff
@@ -3,101 +3,117 @@
...
+        // WindowsAppRuntime is installed and is in a good state and ready for use.
     }
     else
     {
-        // WindowsAppRuntime is not in a good state.
-        // Should call Initialize on another thread, or not use features that require the
-        // WindowsAppRuntime. result.ExtendedError has the error code that has the reason
-        // for the Status.
+        // WindowsAppRuntime is either not installed or not in a good state (can't be both at same time).
+        if (result.Status == DeploymentStatus.PackageInstallRequired)
```
oldnewthing, line 88: What happens if an app developer upgrades to WinAppSDK 1.3 but forgets to update this part of their code? In other words, what happens if the app just calls Initialize for all non-OK deployment statuses instead of calling Repair sometimes?

I hope the answer is "Everything still works, though maybe less efficiently than if you had used Repair."

specs/Deployment/DeploymentAPI.md
oldnewthing, line 95: Suppose in WinAppSDK version 1.9 we add a new DeploymentStatus. This code will ignore all new DeploymentStatus error codes. Should it treat all unknown DeploymentStatus values the same as Unknown?

As written, it means that adding a new DeploymentStatus is a breaking change.

specs/Deployment/DeploymentAPI.md
```diff
@@ -106,25 +122,27 @@ concern to some developers. To address these concerns, developers can instead ca
 and verify that the required packages are missing. If so, then call Initialize() on another thread
 to handle the time delay more elegantly.
 
-Once the main and singleton packages have been deployed, they generally do not need to be deployed
-again. If a user were to remove the main or singleton package, the API can be used to reinstall them
-again.
+Once the Main and Singleton packages have been deployed, they generally do not need to be deployed
+again. If a user were to remove the Main or Singleton package, the API can be used to reinstall them
+again. If there is anything wrong with installed Main and Singleton packages, use Repair API to try
```
oldnewthing, line 127: What does "anything wrong" mean? Does it mean "Any DeploymentStatus other than DeploymentStatus.Ok"?

Because that's not what the sample code does. It calls Repair only for selected DeploymentStatus values, not for "anything wrong".

specs/Deployment/DeploymentAPI.md
```diff
@@ -106,25 +122,27 @@ concern to some developers. To address these concerns, developers can instead ca
...
+and repair them.
 
 ```C#
-    if (DeploymentManager.GetStatus().Status != DeploymentStatus.Ok)
+    if (DeploymentManager.GetStatus().Status == DeploymentStatus.PackageInstallRequired)
```
oldnewthing, line 131: This does not do what the previous text recommends. Previous text says, "If there is anything wrong with installed Main and Singleton packages, use Repair API to try and repair them." But in this code, we don't call Repair for "anything wrong". In fact, we're not calling Repair at all!

specs/Deployment/DeploymentAPI.md
```diff
@@ -167,12 +183,13 @@ they shut down, to refer to the updated framework package.
         // These should be run on a separate thread so as not to hang your app while the
         // packages deploy.
         var initializeTask = Task.Run(() => DeploymentManager.Initialize(deploymentInitializeOptions));
+        //...do other work while the repair is running...
```
oldnewthing, line 186: The comment says "repair", but the method we called was "Initialize".

Either the code should be "Repair", to match the comment; or the comment should say "initialize" to match the code.

specs/Deployment/DeploymentAPI.md
```diff
@@ -181,8 +198,8 @@ they shut down, to refer to the updated framework package.
 
 #### OnError
 
-Additional behavior is available if an error occurs in `DeploymentManager.Initialize()`,
-per the following logic:
+Additional behavior is available if an error occurs in `DeploymentManager.Initialize()`, per the
+following logic:
 
 ```
 DeploymentManager::Initialize(...)
```
oldnewthing, line 205: I don't understand what I'm reading. Is this app code? If not, why are we putting non-app code in the documentation?

If app code, please make this real code, not some weird blend of Visual Basic, C++, and C#.

What does "Only needed supported in packaged processes" mean? What is the thing that is needed only in packaged processes?

I can't find any function called `IsPackagedProcess()`. Is it part of WinAppSDK? The only hit I can find is [an internal function inside the BCL](https://referencesource.microsoft.com/#System.Xml/parent/parent/parent/parent/InternalApis/NDP_Common/inc/Win8Helpers.cs,84).

Also, the test appears to be backward. Don't you mean `if !IsPackagedProcess()`?

Also, why can't you call Initialize for an unpackaged process? Earlier, we said that unpackaged full trust apps can use the Deployment API to get the WinAppSDK packages installed.

Please don't use jargon like "pop the debugger" or "pop some UI". 

Is `options.OnError_ShowUI` the same as `deploymentInitializationOptions.OnErrorShowUI`? I thought `deploymentInitializationOptions.OnErrorShowUI` was telling the `DeploymentManager.Initialize` method that it should display UI, not a note to ourselves to remember to show UI later.

specs/Deployment/DeploymentAPI.md
```diff
@@ -234,6 +251,7 @@ Here's the previous example extended with the `Show UI on error` option:
         // These should be run on a separate thread so as not to hang your app while the
         // packages deploy.
         var initializeTask = Task.Run(() => DeploymentManager.Initialize(deploymentInitializeOptions));
+        //...do other work while the repair is running...
```
oldnewthing, line 254: Again, the comment says "repair", but the code says "Initialize". Please get them to agree on which is correct.

specs/Deployment/DeploymentAPI.md
```diff
@@ -246,12 +264,164 @@ Here's the previous example extended with the `Show UI on error` option:
     }
```
oldnewthing, line 264: If `Initialize` did not work, should we attempt a `Repair`?

specs/Deployment/DeploymentAPI.md
```diff
@@ -246,12 +264,164 @@ Here's the previous example extended with the `Show UI on error` option:
...
+
+This API will always attempt to repair WindowsAppRuntime regardless of it's state. For that reason,
+this API should be called only when deployment status is not OK to avoid an unecessary repair even
+when WindowsAppRuntime is in good state. All information about the version, channel, and
+architecture needed are derived from the current WinAppSDK Framework package.
+
+Since both the Main _and_ the Singleton packages will be repaired, it is possible that Main package
+repair has failed and Initialize returns with PackageRepairFailed status (OR) it is possible that
+Main package is repaired successfully but the Singleton package repair has failed and Initialize
+returns with repair has failed and Initialize returns with PackageRepairFailed status (OR) it is
```
oldnewthing, line 277: Seems to be saying the same thing twice. "Initialize returns with repair has failed" and "Initialize returns with PackageRepairFailed".

specs/Deployment/DeploymentAPI.md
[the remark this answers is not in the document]
oldnewthing: I see three cases in the text (separated by "OR"):

1. it is possible that Main package repair has failed and Initialize returns with PackageRepairFailed status
2. it is possible that Main package is repaired successfully but the Singleton package repair has failed and Initialize returns with repair has failed and Initialize returns with PackageRepairFailed status
3.  it is possible that status

Not sure what 3 is trying to say.

specs/Deployment/DeploymentAPI.md
```diff
@@ -246,12 +264,164 @@ Here's the previous example extended with the `Show UI on error` option:
...
+possible that status. In both cases, the returned WindowsAppRuntimeStatus will contain the error of
+first package repair failure. There will be no rollback of any successfully repaired packages. See
+the
+[developer docs](https://docs.microsoft.com/windows/apps/windows-app-sdk/deploy-packaged-apps#address-installation-errors)
+for instructions on how to handle errors.
+
+```C#
+    if (DeploymentManager.GetStatus().Status == DeploymentStatus.PackageRepairRequired ||
+        DeploymentManager.GetStatus().Status == DeploymentStatus.Unknown)
```
oldnewthing, line 286: Shouldn't this really be "if status != DeploymentStatus.PackageRepairFailed"?

We want to attempt a repair for all errors, except the error that says "Don't try to repair; it won't work."

Otherwise, if we add a new status like "PackageRepairRecommended" or "PreviousRepairTimedOut", we will fail to attempt a repair.

specs/Deployment/DeploymentAPI.md
```diff
@@ -246,12 +264,164 @@ Here's the previous example extended with the `Show UI on error` option:
...
+    {
+        // Repair will always attempt to repair WindowsAppRuntime regardless of it's state.
```
oldnewthing, line 288:
```suggestion
        // Repair will always attempt to repair WindowsAppRuntime regardless of its state.
```

specs/Deployment/DeploymentAPI.md
oldnewthing, line 286: Is it right to call `GetStatus()` twice? Could the status change between the two calls?

specs/Deployment/DeploymentAPI.md
```diff
@@ -246,12 +264,164 @@ Here's the previous example extended with the `Show UI on error` option:
...
+API. This option when set will shut down the application(s) associated with WinAppSDK packages,
+update the WinAppSDK packages and restart the application(s).
+
+If this option is set when updating Singleton package, then the out of process com server from the
+Singleton package such as push Notifications is explicitly restarted.
+
+When the API is updating Framework package and ForceDeployment option is set, then all dependent
+packages that are NOT currently in use will be immediately re-installed to refer to the updated
+Framework package and all dependent packages that are currently in use will be re-installed, after
+they shut down, to refer to the updated Framework package.
```
oldnewthing, line 324: Wait, I thought the first paragraph in this section said that if you force deployment, then we don't want for the users to exit the apps; we shut down the apps immediately. But here, we wait for the apps to shut down naturally.

Or is "after they shut down" saying "after we shut them down" (rather than the app shutting down on its own)? Could use some clarity about whether the shutdown is immediate or eventual.

specs/Deployment/DeploymentAPI.md
```diff
@@ -246,12 +264,164 @@ Here's the previous example extended with the `Show UI on error` option:
...
+
+```C#
+    if (DeploymentManager.GetStatus().Status == DeploymentStatus.PackageInstallRequired)
+    {
+        // Create DeploymentRepairOptions object and set ForceDeployment option
+        // to allow for force updating of the WinAppSDK packages even if they are currently in use.
+        var deploymentRepairOptions = new DeploymentRepairOptions();
+        deploymentRepairOptions.ForceDeployment = true;
+
+        // Repair will always attempt to repair WindowsAppRuntime regardless of it's state.
```
oldnewthing, line 334:
```suggestion
        // Repair will always attempt to repair WindowsAppRuntime regardless of its state.
```

specs/Deployment/DeploymentAPI.md
```diff
@@ -246,12 +264,164 @@ Here's the previous example extended with the `Show UI on error` option:
...
+#### OnError
+
+Additional behavior is available if an error occurs in `DeploymentManager.Repair()`, per the
+following logic:
+
+```
+DeploymentManager::Repair(...)
+{
+    // Only needed supported in packaged processes
+    IF IsPackagedProcess()
```
oldnewthing, line 361: Test seems backwards.

mominshaikhdevs:
is this stale or what's the latest update on this?

mominshaikhdevs:
@sachintaMSFT Is this stale or what's the latest update on this?

lauren-ciha:

PR #2897, opened 2022-08-23
Refactor Notification Telemetry to Remove Code Duplication

loneursid:
DCR https://dev.azure.com/microsoft/OS/_workitems/edit/41085624

Refactor the two existing NotificationTelemetryProvider set of files to remove code duplication before introducing additional providers.


pmpurifoy (approved):

pmpurifoy:
Only minor nit is maybe change m_helper->m_telemetryHelper. At this point we have so many helpers it'd be great to clarify which ones we are using.

loneursid:
> Only minor nit is maybe change m_helper->m_telemetryHelper. At this point we have so many helpers it'd be great to clarify which ones we are using.

Addressed, thanks

dev/AppNotifications/AppNotificationTelemetry.h
```diff
@@ -20,16 +21,16 @@ class AppNotificationTelemetry : public wil::TraceLoggingProvider
         winrt::hresult hr,
         std::wstring const& appId) noexcept try
     {
-        if (c_maxEventLimit >= UpdateLogEventCount())
+        if (m_telemetryHelper.ShouldLogEvent())
```
hulumane, line 24: cool. Looks good to me. Whats the advantage of moving to a helper? Reusability for other Telemetry providers?

hulumane (approved):

dev/AppNotifications/AppNotificationTelemetry.h
loneursid: We already had  this code in two separate providers and the upcoming AppNotificationBuilderTelemetryProvider will also need access to this code.

PR #2898, opened 2022-08-23
Add MddGetPackageGraphRevisionId() + M.W.AM.DynamicDependency.PackageDependency.PackageGraphRevisionId API Spec

DrusTheAxe:
Add `MddGetPackageGraphRevisionId()` (Flat-C API) and `PackageGraphRevisionId` property (WinRT API), for alignment with [Windows 11 API / terminology](https://docs.microsoft.com/en-us/windows/apps/desktop/modernize/framework-packages/use-the-dynamic-dependency-api).

GenerationId is soft-deprecated (predecessor equivalent to PackageGraphRevisionId still supported but will go away in a future release).

Also corrected spec missing updates:
* Corrected missing changes per previous API Review and implementation
* Added missing contracts (added in WinAppSDK 1.1)

https://task.ms/41077950

### DIFF
--- a/specs/dynamicdependencies/DynamicDependencies.md
+++ b/specs/dynamicdependencies/DynamicDependencies.md
@@ -1253,7 +1253,13 @@ STDAPI MddGetIdForPackageDependencyContext(
     _In_ MDD_PACKAGEDEPENDENCY_CONTEXT packageDependencyContext,
     _Outptr_result_maybenull_ PWSTR* packageDependencyId) noexcept;
 
-/// Return the package graph's current generation id.
+/// Return the package graph's current revision id.
+STDAPI_(UINT32) MddGetPackageGraphRevisionId() noexcept;
+
+/// Return the package graph's current revision id.
+///
+/// @note This API is deprecated and will be removed in a future release.
+///       Use MddGetPackageGraphRevisionId().
 STDAPI_(UINT32) MddGetGenerationId() noexcept;
 ```
 
@@ -1565,10 +1571,14 @@ namespace Microsoft.Windows.ApplicationModel.DynamicDependency
 ```c# (but really MIDL3)
 namespace Microsoft.Windows.ApplicationModel.DynamicDependency
 {
+[contractversion(2)]
+apicontract DynamicDependencyContract{};
+
 /// CPU architectures to optionally filter available packages against a package dependency.
 /// These generally correspond to processor architecture types supported by MSIX.
 /// @see Windows.System.ProcessorArchitecture
 [flags]
+[contract(DynamicDependencyContract, 1)]
 enum PackageDependencyProcessorArchitectures
 {
     None       = 0,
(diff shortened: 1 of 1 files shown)


DrusTheAxe:
We're doing an "email review" of this - please leave your comments in the PR by 5pm Seattle time, 2022-08-26. @DrusTheAxe will address feedback in future updates.

EHO-makai (commented):
Is it on purpose that we change generation => revision id but make no mention of the "soft deprecation?"

DrusTheAxe:
> Is it on purpose that we change generation => revision id but make no mention of the "soft deprecation?"

Are you referring to a specific line?

EHO-makai:
No, it seems like most of the edits just change generation => revision.  But no mention is made of what you're supposed to do if you had previously used generation.  I'm not asking for a change here necessarily, just wondering if silently doing a "replace" here is what you intended / OK.

DrusTheAxe:
> No, it seems like most of the edits just change generation => revision. But no mention is made of what you're supposed to do if you had previously used generation.

2 words for the same thing.

Was originally called GenerationId in Windows (internal) and then WinAppSDK DynDep API Review. Windows had no public API to get that value so we fixed that with a new public API (`GetPackageGraphRevisionId`) per API Review, which recommended using the term 'revision id' instead of 'generation id'.

This merely updates WinAppSDK to support consist lingo with Windows.

>I'm not asking for a change here necessarily, just wondering if silently doing a "replace" here is what you intended / OK.

Yes. Given it's the same thing, the GetGeneration doc-comments point you to GetRevision, and it's open source so anyone who looks can see it's the same thing<sup>1</sup> I didn't see a need to belabor the point. I expect the release notes and docs will mention this.


<sup>1</sup> The implementation is literally 1-liners `UINT32 MddGetGenerationId() { return MddGetPackageGraphRevisionId(); }`)

EHO-makai (approved):

MikeHillberg:
We have a [template for API specs](https://github.com/microsoft/WindowsAppSDK/blob/main/specs/spec_template.md) (not just for WinRT APIs)

specs/dynamicdependencies/DynamicDependencies.md
```diff
@@ -1841,12 +1850,19 @@ runtimeclass PackageDependency
...
+    /// Successful calls change the package graph's current revision id.
     ///
-    /// @see GenerationId
+    /// @see PackageGraphRevisionId
     PackageDependencyContext Add(AddPackageDependencyOptions options);
 
-    /// Return the package graph's current generation id.
+    /// Return the package graph's current revision id.
+    [contract(DynamicDependencyContract, 2)]
+    static UInt32 PackageGraphRevisionId{ get; };
```
MikeHillberg, line 1860: Standard guideline is to use Int32, even if signed is invalid, unless the API really needs more than 2B values

DrusTheAxe: That's news. This is a wrapper over the Win32 API which returns `uint32`. Still worth going `int32`?

idigdoug: Existing functions already return U32, so probably better to just continue with that.

PR #2900, opened 2022-08-23
Refactor LRP tests for notifications

pmpurifoy:
- Restructured test locations for Notification tests
- Added LRP test dll covering all exposed functions


test/LRPTests/APITests.cpp
```diff
@@ -0,0 +1,166 @@
...
+                nullptr, // thread attributes
+                FALSE,   // inherit handles
+                NORMAL_PRIORITY_CLASS, // creation flags
+                nullptr, // lpEnvironment
+                nullptr, // current directory for the process
+                &startupInfo,
+                &processInfo
+            );
+
+            if (wasProcessCreated == FALSE)
```
danielayala94, line 77: Why do we need this check? Shouldn't we fail the test if process creation failed?

test/LRPTests/APITests.cpp
```diff
@@ -0,0 +1,166 @@
...
+            startupInfo.cb = sizeof(startupInfo);
+
+            // Build the Solution OutDir path where the Startup Task exe is.
+            auto startupTaskExePath = Test::FileSystem::GetSolutionOutDirPath();
+            startupTaskExePath /= Test::Packages::WindowsAppRuntimeSingleton::c_PackageDirName;
+            startupTaskExePath += L".Msix";
+            startupTaskExePath /= L"msix";
+            startupTaskExePath /= L"PushNotificationsLongRunningTask.StartupTask.exe";
+
+            BOOL wasProcessCreated = CreateProcess(
```
danielayala94, line 64: Do you know if there's a wil construct for CreateProcess?

test/LRPTests/APITests.cpp
```diff
@@ -0,0 +1,166 @@
...
+        TEST_METHOD(LaunchLRP_FromCoCreateInstance)
+        {
+            winrt::com_ptr notificationPlatform{ winrt::try_create_instance<INotificationsLongRunningPlatform>(_uuidof(NotificationsLongRunningPlatform), CLSCTX_ALL) };
+            VERIFY_IS_NOT_NULL(notificationPlatform.get());
+        }
+
+        TEST_METHOD(LaunchLRP_FromStartupTask)
+        {
+            STARTUPINFO startupInfo = { 0 };
+            wil::unique_process_information processInfo;
```
danielayala94, line 52: nit: I would move this closer to usage and regroup the startupInfo stuff.

test/LRPTests/APITests.cpp
```diff
@@ -0,0 +1,166 @@
...
+    static const PCWSTR c_appId = L"toastAppId";
+
+    class LRPTests
+    {
+
+    public:
+        BEGIN_TEST_CLASS(LRPTests)
+            TEST_CLASS_PROPERTY(L"Description", L"Windows App SDK Push Notifications Long Running Process tests")
+            TEST_CLASS_PROPERTY(L"ThreadingModel", L"MTA")
+            TEST_CLASS_PROPERTY(L"RunAs:Class", L"RestrictedUser")
```
danielayala94, line 28: What does this mean? Sounds like this is a non-admin user? Then how is it different from not specifying this property at all?

test/LRPTests/APITests.cpp
pmpurifoy: If wasProcessCreated is false, GetLastError will contain the error code and the next line uses VERIFY_SUCCEEDED to log the exact error that was thrown.

test/LRPTests/APITests.cpp
pmpurifoy: I have not found anything related to CreateProcess with wil.

test/LRPTests/APITests.cpp
pmpurifoy: RestrictedUser means that the process is running as a user at medium-IL. This is our only supported scenario, **but in the future** we will need to add ElevatedUser to make sure elevated processes can call into the LRP.

test/LRPTests/APITests.cpp
sharath2727: use winrt for getlasterror

test/LRPTests/APITests.cpp
```diff
@@ -0,0 +1,164 @@
...
+                    return;
+                }
+
+                result = Process32Next(processesSnapshot.get(), &processEntry);
+            }
+
+            VERIFY_IS_FALSE(isRunning);
+        }
+
+        TEST_METHOD(LaunchLRP_FromCoCreateInstance)
```
sharath2727, line 111: do you need this test..I guess other tests are going to be doing this.

test/LRPTests/APITests.cpp
```diff
@@ -0,0 +1,164 @@
...
+            wil::unique_cotaskmem_string lrpAppId;
+            VERIFY_NO_THROW(notificationPlatform->RegisterFullTrustApplication(c_processName, winrt::guid(), &lrpAppId));
+            VERIFY_NO_THROW(notificationPlatform->RegisterLongRunningActivator(c_processName));
+            VERIFY_NO_THROW(notificationPlatform->UnregisterLongRunningActivator(c_processName));
+            VERIFY_NO_THROW(notificationPlatform->UnregisterFullTrustApplication(c_processName));
+        }
+
+        TEST_METHOD(RegisterUnregisterLongRunningActivatorWithClsid)
+        {
+            winrt::com_ptr notificationPlatform{ winrt::try_create_instance<INotificationsLongRunningPlatform>(_uuidof(NotificationsLongRunningPlatform), CLSCTX_ALL) };
```
sharath2727, line 131: this is repeat thing can we move it to a common setup?

test/LRPTests/APITests.cpp
sharath2727: winrt::check_bool will throw lasterror I think.

test/LRPTests/APITests.cpp
danielayala94: What is the default IL we get for the tests if we don't specify a user?

test/LRPTests/APITests.cpp
pmpurifoy: Yup. https://docs.microsoft.com/en-us/uwp/cpp-ref-for-winrt/error-handling/check-bool

test/LRPTests/APITests.cpp
pmpurifoy: If you don't specify a user, TAEF will run in the context of which TAEF is running.

test/LRPTests/APITests.cpp
pmpurifoy: Tbh I was thinking the same thing. Will remove.

test/LRPTests/APITests.cpp
pmpurifoy: Yeah, I'll move this to a common function.

test/LRPTests/APITests.cpp
```diff
@@ -0,0 +1,157 @@
...
+            ZeroMemory(&startupInfo, sizeof(startupInfo));
+            startupInfo.cb = sizeof(startupInfo);
+
+            // Build the Solution OutDir path where the Startup Task exe is.
+            // Path: $OutDir/WindowsAppRuntime.Test.Singleton.Msix/msix/PushNotificationsLongRunningTask.StartupTask.exe
+            auto startupTaskExePath{ Test::FileSystem::GetSolutionOutDirPath() / Test::Packages::WindowsAppRuntimeSingleton::c_PackageDirName };
+            startupTaskExePath += LR"(.Msix/msix/PushNotificationsLongRunningTask.StartupTask.exe)";
+
+            wil::unique_process_information processInfo;
+            winrt::check_bool(CreateProcess(
```
sharath2727, line 65: what do you do with the error here when winrt::check_bool throws the last recorded error.

test/LRPTests/APITests.cpp
```diff
@@ -0,0 +1,157 @@
...
+        BEGIN_TEST_CLASS(LRPTests)
+            TEST_CLASS_PROPERTY(L"Description", L"Windows App SDK Push Notifications Long Running Process tests")
+            TEST_CLASS_PROPERTY(L"ThreadingModel", L"MTA")
+            TEST_CLASS_PROPERTY(L"RunAs:Class", L"RestrictedUser")
+        END_TEST_CLASS()
+
+        winrt::com_ptr<INotificationsLongRunningPlatform> GetNotificationPlatform()
+        {
+            winrt::com_ptr notificationPlatform{ winrt::try_create_instance<INotificationsLongRunningPlatform>(_uuidof(NotificationsLongRunningPlatform), CLSCTX_ALL) };
+            VERIFY_IS_NOT_NULL(notificationPlatform.get());
```
sharath2727, line 34: Optional: wondering if it is succeeded assign it to a class variable? that way you don't have to keep creating and destroying an instance of it. Let me know.

pmpurifoy: As long as a client has a reference to the notificationPlatform, the LRP will stay alive. We don't want this behavior for LaunchLRP_FromStartupTask since we are verifying the LRP is dead then comes alive in the test from the startuptask.exe. If it wasn't for this test, we could use a member variable.

test/LRPTests/APITests.cpp
pmpurifoy: Wrapped the call in VERIFY_NO_THROW. Fixed

test/LRPTests/pch.cpp
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
```
loneursid, line 3: Use new copyright notice:
// Copyright (c) Microsoft Corporation and Contributors.

test/LRPTests/APITests.cpp
```diff
@@ -0,0 +1,157 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
```
loneursid, line 3: Use new copyright notice:
// Copyright (c) Microsoft Corporation and Contributors.

test/LRPTests/APITests.cpp
sharath2727: Thanks for the clarification.

sharath2727 (approved):

PR #2901, opened 2022-08-24
Fix Installer to work on ARM64 on Windows 10

sachintaMSFT:
1. Fix the logic in determining whether a given Procedure is found in the given library or not.
2. Fix console output that wasn't respecting quiet option.
3. Fix Push notifications LRP restart decision logic to be inclusive of ARM64 architecture as well.

Tested and validated the Installer on a real arm64 devices on both Windows 10 and Windows 11 versions.

### DIFF
--- a/dev/Common/IsWindowsVersion.h
+++ b/dev/Common/IsWindowsVersion.h
@@ -20,8 +20,17 @@ inline bool IsExportPresent(
         {
             return true;
         }
+        else
+        {
+            DWORD getProcAddressError = GetLastError();
+            if (getProcAddressError != ERROR_PROC_NOT_FOUND)
+            {
+                THROW_HR(HRESULT_FROM_WIN32(getProcAddressError));
+            }
+            return false;
+        }
     }
-    return false;
+    THROW_LAST_ERROR();
 }
 
 inline bool IsWindows10_19H1OrGreater()
--- a/dev/Common/IsWindowsVersion.h
+++ b/dev/Common/IsWindowsVersion.h
@@ -13,24 +13,20 @@ inline bool IsExportPresent(
     PCSTR functionName)
 {
     wil::unique_hmodule dll{ LoadLibraryExW(filename, nullptr, 0) };
-    if (dll)
+    if (!dll)
     {
-        auto function{ GetProcAddress(dll.get(), functionName) };
-        if (function)
-        {
-            return true;
-        }
-        else
-        {
-            DWORD getProcAddressError = GetLastError();
-            if (getProcAddressError != ERROR_PROC_NOT_FOUND)
-            {
-                THROW_HR(HRESULT_FROM_WIN32(getProcAddressError));
-            }
-            return false;
-        }
+        THROW_LAST_ERROR();
     }
-    THROW_LAST_ERROR();
+
+    auto function{ GetProcAddress(dll.get(), functionName) };
+    if (!function)
+    {
+        DWORD getProcAddressError = GetLastError();
+        THROW_HR_IF(HRESULT_FROM_WIN32(getProcAddressError), getProcAddressError != ERROR_PROC_NOT_FOUND);
+        return false;
+    }
+
+    return true;
 }
 
 inline bool IsWindows10_19H1OrGreater()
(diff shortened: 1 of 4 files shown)


DrusTheAxe:
> …e given library or not.

Ick. GitHub creates PRs with the 1st N chars of the commit comment and splits the rest into the description but...ick. Please fix

dev/Common/IsWindowsVersion.h
```diff
@@ -20,8 +20,17 @@ inline bool IsExportPresent(
         {
             return true;
         }
+        else
+        {
+            DWORD getProcAddressError = GetLastError();
+            if (getProcAddressError != ERROR_PROC_NOT_FOUND)
+            {
+                THROW_HR(HRESULT_FROM_WIN32(getProcAddressError));
```
DrusTheAxe, line 28: THROW_HR_IF

dev/Common/IsWindowsVersion.h
```diff
@@ -20,8 +20,17 @@ inline bool IsExportPresent(
...
+            }
+            return false;
+        }
     }
-    return false;
+    THROW_LAST_ERROR();
```
DrusTheAxe, line 33: Ick. Valid, but the logic's kinda spread.

If the is refactored to early-return this would be clearer.

Also, shouldn't LoadLibrary=ERROR_MOD_NOT_FOUND == false (not error)?

How about this?

```
{
    wil::unique_hmodule dll{ LoadLibraryExW(filename, nullptr, 0) };
    if (!dll)
    {
        const auto lastError{ GetLastError() };
        if (lastError == ERROR_MOD_NOT_FOUND)
        {
            return false;
        }
        THROW_HR(HRESULT_FROM_WIN32(lastError));
    }

    auto function{ GetProcAddress(dll.get(), functionName) };
    if (!function)
    {
        const auto lastError{ GetLastError() };
        if (lastError == ERROR_PROC_NOT_FOUND)
        {
            return false;
        }
        THROW_HR(HRESULT_FROM_WIN32(lastError));
    }

    return true;
}
```

installer/dev/MachineTypeAttributes.h
```diff
@@ -24,20 +24,22 @@ namespace MachineTypeAttributes
...
             if (!getMachineTypeAttributes)
             {
-                // For now, GetMachineTypeAttributes API is available only on Windows 11 (i.e. builds 22000+).
-                THROW_LAST_ERROR();
+                DWORD getProcAddressError = GetLastError();
+                if (getProcAddressError != ERROR_PROC_NOT_FOUND)
+                {
+                    THROW_HR(HRESULT_FROM_WIN32(getProcAddressError));
+                }
+                // If execution comes here, it means the current system is not running Windows 11.
```
DrusTheAxe, line 32: Refactor similar to above suggestion (followed by "now that we have procaddress call it...")

dev/Common/IsWindowsVersion.h
sachintaMSFT: Sure, I like the refactor. But I don't agree with ERROR_MOD_NOT_FOUND. Unless there is an expected case where the module itself could not be found, it should throw an error. Is there an expected case where the kernelbase.dll wouldn't be found ?

DrusTheAxe: It futureproofs against probing for APIs in new DLLs or APIsets.

dev/Common/IsWindowsVersion.h
```diff
@@ -15,7 +15,9 @@ inline bool IsExportPresent(
     wil::unique_hmodule dll{ LoadLibraryExW(filename, nullptr, 0) };
     if (!dll)
     {
-        THROW_LAST_ERROR();
+        DWORD getLoadLibraryError = GetLastError();
```
DrusTheAxe, line 18: NIT: `const`

DrusTheAxe, line 18: NIT: `rcLoadLibrary`

GetLastError = Win32 error code. Common short variable name for that is `rc` (short for `return code`), much like `hr` is the common name for an HRESULT.

If you want multiple `rc` variables then `rcFoo` and `rcBar` is the common pattern, or in this case `rcLoadLibrary` and `rcGetProcAddress`

DrusTheAxe (approved):
LGTM modulo the `getLoadLibraryError` / `rcLoadLibrary` comment

PR #2909, opened 2022-08-26
Add MddGetPackageGraphRevisionId() and PackageGraphRevisionId property

DrusTheAxe:
Add MddGetPackageGraphRevisionId() and PackageGraphRevisionId property.

Detour GetPackageGraphRevisionId() (new in recent Windows previews).

Detach Detour'd functions in reverse order that we Detour'd them.

Moar Tests!

https://task.ms/41077943

### DIFF
--- a/test/DynamicDependency/Test_Win32/Test_GetCurrentPackageInfo.cpp
+++ b/test/DynamicDependency/Test_Win32/Test_GetCurrentPackageInfo.cpp
@@ -252,6 +252,15 @@ namespace Test::DynamicDependency
 
             const auto actualGenerationId{ MddGetGenerationId() };
             VERIFY_ARE_EQUAL(expectedGenerationId, actualGenerationId);
+
+            if (!IsGetPackageGraphRevisionIdSupported())
+            {
+                return;
+            }
+
+            const auto actualPackageGraphRevisionId{ MddGetPackageGraphRevisionId() };
+            const auto expectedPackageGraphRevisionId{ actualGenerationId };
+            VERIFY_ARE_EQUAL(expectedPackageGraphRevisionId, actualPackageGraphRevisionId);
         }
 
         void PrintGetCurrentPackageInfoHeader()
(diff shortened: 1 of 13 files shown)


test/DynamicDependency/Test_Win32/Test_GetCurrentPackageInfo.cpp
```diff
@@ -343,14 +352,27 @@ namespace Test::DynamicDependency
...
-        void EnsureGetCurrentPackageInfo3IsIniitalizedIfExists()
+        typedef UINT32 (WINAPI* GetPackageGraphRevisionIdFunction)();
+        GetPackageGraphRevisionIdFunction m_getPackageGraphRevisionId{};
+
+        void EnsureKernelbaseDllIsLoaded()
         {
             if (!m_kernelbaseDll)
             {
                 wil::unique_hmodule dll{ LoadLibraryW(L"kernelbase.dll") };
                 FAIL_FAST_HR_IF_NULL(E_UNEXPECTED, dll);
```
sachintaMSFT, line 363: Don't you want to log GetLastError() instead for debugging purposes instead of bucketing them to E_UNEXPECTED ?, when Kernel base dll couldn't be found ?

test/DynamicDependency/Test_Win32/Test_GetCurrentPackageInfo.cpp
```diff
@@ -360,17 +382,42 @@ namespace Test::DynamicDependency
...
+                    }
+                }
+
+                m_getPackageGraphRevisionId = std::move(function);
+            }
+        }
+
+        bool IsGetPackageGraphRevisionIdSupported()
+        {
+            EnsureGetPackageGraphRevisionIdIsInitializedIfExists();
```
sachintaMSFT, line 417: Shouldn't this be called only when m_getPackageGraphRevisionId == nullptr to avoid re-running same code in every call  ?

sachintaMSFT (approved):
LGMT except for couple of minor comments in Tests.

test/DynamicDependency/Test_Win32/Test_GetCurrentPackageInfo.cpp
DrusTheAxe: `EnsureGetPackageGraphRevisionIdIsInitializedIfExists()` does the checks to only do work once and cache process-constant info for latest calls. Keeps the is-it-available-or-is-it-not isolated low and not pollute higher level checks, and the work to hit this 2nd+ time is cheap and this is a test not product code so I favored maintainability over ultra squeezing perf

test/DynamicDependency/Test_Win32/Test_GetCurrentPackageInfo.cpp
DrusTheAxe: Hmmm. Technically yes, though it's kernelbase.dll which is loaded into a win32 process before even `DllMain()` is invoked so this should just hit the Loaded Module Table and find the match. Or rather, if it doesn't Windows is in a seriously broken state already and I'd expect maybe and bloodshed well before reaching this point ;-)

PR #2920, opened 2022-08-29
Partially reenable the Integration test to only produce a WinAppSDK package

kythant:
As per demand


loneursid (approved):
Thanks for doing this.

PR #2544, opened 2022-05-23
Making EM RS5+. Limiting the tracking capability to Windows11+ (Spec)

dhoehna:
I realized that the API used to manipulating environment variables can work down level. The restriction to EM was tracking the changes so they could be reverted when a package was removed.

I split up the changes. IsSupported always returns true. If the developer wants to know if the changes will be tracked, they can call the object method AreChangesTracked.

This PR has the spec change and is my formal request for an API review.

PR #2938, opened 2022-09-06
Environment Manager is now RS5+.   Tracking Registry writes is Windows 11+ and with better checks.

dhoehna:
The Spec PR is here: https://github.com/microsoft/WindowsAppSDK/pull/2544

This PR addresses two problems with the EM API.
1. The API will work on RS5+
2. EM did not check for the existence of the Windows app sdk framework package to decide if tracking is enabled.

This PR fixes these issues by
1. IsSupported returns true.
2. A new non-static method AreChangesTrackedis added to query if the registry changes will be stored.

In addition to the above changes the  this PR also did
1. Removed the UWP tests.
2. Add more comments, and fixes to the BuildTransportPackage powershell script.


dev/EnvironmentManager/ChangeTracker/EnvironmentVariableChangeTrackerHelper.h
```diff
@@ -0,0 +1,117 @@
...
+    const auto subKey = std::filesystem::path{ LR"(Software\Microsoft\AppModel\RegistryWriteVirtualization\ExcludedKeys\HKEY_CURRENT_USER/Software/ChangeTracker)" };
+    wil::unique_hkey keyToChangeTracker{};
+    LSTATUS openResult{ RegOpenKeyEx(HKEY_LOCAL_MACHINE, subKey.c_str(), 0, KEY_READ | KEY_WOW64_64KEY, keyToChangeTracker.put()) };
+
+    LOG_IF_WIN32_ERROR_MSG(openResult, "Open key result to check if tracking exists: %i. Errors are treated as \"Do not track changes\"", openResult);
+
+    return openResult == ERROR_SUCCESS;
+}
+
+inline bool IsDllOnSystemAndRegistered()
```
jonwis, line 28: Feels like this doesn't belong in a header. You could put the definition here like this:

```c++
#if defined(__WINDOWSAPPRUNTIME_TEST_METADATA_H)
#define FRAMEWORK_PACKAGE_FAMILY_PATTERN L"Microsoft.WindowsAppRuntime.Framework-%i.%i_8wekyb3d8bbwe"
#else
#define FRAMEWORK_PACKAGE_FAMILY_PATTERN L"Microsoft.WindowsAppRuntime.%i.%i_8wekyb3d8bbwe"
#endif

bool IsDllOnSystemAndRegistered(std::wstring_view frameworkPackageFamily =  FRAMEWORK_PACKAGE_FAMILY_PATTERN);
```

... and then move the body of the method to a .cpp file

dev/EnvironmentManager/ChangeTracker/EnvironmentVariableChangeTrackerHelper.h
```diff
@@ -0,0 +1,117 @@
...
+    // At this point no framework packages are installed for the current user.  However, a framework package might be installed
+    // machine wide.  Check that.  However, the process needs to be elevated to check.
+    if (!Security::IntegrityLevel::IsElevated())
+    {
+        return false;
+    }
+
+    winrt::Windows::Management::Deployment::PackageManager packageManager{};
+    auto currentArch{ AppModel::Identity::GetCurrentArchitecture() };
+    for (uint16_t major = 1; major < maxMajorToSearch; major++)
```
jonwis, line 88: Is it easier to spam a bunch of "find" calls, or just "get all" and do filtering on your own? On my desktop, I've got 257 packages registered for all users (per `get-appxpackage | measure-object Name |% Count`) which is 10x fewer than the 2500 item loop here.

dev/EnvironmentManager/ChangeTracker/EnvironmentVariableChangeTrackerHelper.h
```diff
@@ -0,0 +1,117 @@
...
+                    return true;
+                }
+            }
+        }
+    }
+
+    return false;
+}
+
+inline bool ShouldChangesBeTracked(EnvironmentManager::Scope scope)
```
jonwis, line 107: More things to move to .cpp files

dev/EnvironmentManager/ChangeTracker/EnvironmentVariableChangeTrackerHelper.h
```diff
@@ -0,0 +1,117 @@
...
+
+using namespace winrt::Microsoft::Windows::System::implementation;
+
+static bool DoesChangeTrackingKeyExist()
+{
+    const auto subKey = std::filesystem::path{ LR"(Software\Microsoft\AppModel\RegistryWriteVirtualization\ExcludedKeys\HKEY_CURRENT_USER/Software/ChangeTracker)" };
+    wil::unique_hkey keyToChangeTracker{};
+    LSTATUS openResult{ RegOpenKeyEx(HKEY_LOCAL_MACHINE, subKey.c_str(), 0, KEY_READ | KEY_WOW64_64KEY, keyToChangeTracker.put()) };
+
+    LOG_IF_WIN32_ERROR_MSG(openResult, "Open key result to check if tracking exists: %i. Errors are treated as \"Do not track changes\"", openResult);
```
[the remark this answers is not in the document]
DrusTheAxe: +1

*_MSG are meant for troubleshooting, not english prose.

SUGGESTION: `LOG_IF_WIN32_ERROR_MSG(openResult, "%d", openResult);`

(%i? Is that new?)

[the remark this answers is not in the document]
DrusTheAxe: P.S. `%i`? Is that new?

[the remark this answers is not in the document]
jonwis: On `%i` - https://docs.microsoft.com/en-us/cpp/c-runtime-library/format-specification-syntax-printf-and-wprintf-functions?view=msvc-170 ... same comment as `%d`

dev/EnvironmentManager/ChangeTracker/EnvironmentVariableChangeTrackerHelper.h
```diff
@@ -0,0 +1,117 @@
...
+
+    return openResult == ERROR_SUCCESS;
+}
+
+inline bool IsDllOnSystemAndRegistered()
+{
+
+    // Maybe a better way exists to switch between the test package family name and the production family name.
+    // This does not work because the test code and EM code are in two different processes.
+#if defined(__WINDOWSAPPRUNTIME_TEST_METADATA_H)
```
DrusTheAxe, line 33: >Maybe a better way exists to switch

It does ;-)

In your production code:

```
#include <dev\common\WindowsAppRuntime.VersionInfo.h>
...
const auto frameworkPackageFamily{ ::WindowsAppRuntime::VersionInfo::GetPackageFamilyName() };
```

and in your test call

```
#include <dev\common\WindowsAppRuntime.VersionInfo.h>
...
WindowsAppRuntime::VersionInfo::TestInitialize(::TP::WindowsAppRuntimeFramework::c_PackageFamilyName,
                                               ::TP::WindowsAppRuntimeMain::c_PackageFamilyName);

```

See `Unpackaged_SelfContained` in `test\common\Test_SelfContained.cpp` for an example.

WHY: Compile time hardwired for test vs production data is frowned upon given the difficulty using it appropriately. The better answer is to explicitly set a 'test override' when testing. This is how `MddBootstrapInitialize()` handled it and later more generalized to `WindowsAppRuntime.VersionInfo.h` because SelfContained and others needed the same sort of reindeer games.

dev/EnvironmentManager/ChangeTracker/EnvironmentVariableChangeTrackerHelper.h
```diff
@@ -0,0 +1,117 @@
...
+    constexpr std::wstring_view frameworkPackageFamily{ L"Microsoft.WindowsAppRuntime.%i.%i_8wekyb3d8bbwe" };
+#endif
+
+    std::unique_ptr<WCHAR> frameworkPackageFamilyName{ new WCHAR[PACKAGE_FAMILY_NAME_MAX_LENGTH + 1] };
+
+    // UINT16_MAX is big.  50 should be a reasonable MAJOR and MINOR.
+    constexpr uint16_t maxMajorToSearch{ 50 };
+    constexpr uint16_t maxMinorToSearch{ 50 };
+    bool foundAPackageInstalledForTheUser{};
+    for (uint16_t major = 1; major < maxMajorToSearch; major++)
```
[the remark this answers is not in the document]
DrusTheAxe: What are you trying to accomplish? The intent isn't clear from the code and there's no comment explaining the goal

I don't understand this at all. You're looking for the framework package for a particular release by scanning for releases 1.0 thru 1.50 and 0.* thru 49.*. until you find the lowest version matching?

P.S. This ignores non-stable releases e.g. 1.2-preview3. I'll go out on a limb and guess that's not your intent ;-)

DrusTheAxe (changes requested):

dev/EnvironmentManager/ChangeTracker/EnvironmentVariableChangeTrackerHelper.h
DrusTheAxe: Should go away. See other comment

dev/EnvironmentManager/ChangeTracker/EnvironmentVariableChangeTrackerHelper.h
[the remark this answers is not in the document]
dhoehna: I might have mis-read your message then.  In our email @DrusTheAxe said this.

There’s really 4 possible cases:
A.	Package is registered for current user
B.	Package is registered for another user
C.	Package is registered for machine
D.	Package is staged but not registered
Users can always check A. B+C require admin privilege. D usually requires admin privilege and is rarely useful – except in this case, as Undocked DEHs have a weird lifecycle. Don’t need to be in any ‘registered’ package, just one known to Deployment (i.e. Staged | Registered). See [@Chris Wall (WIN SDE)](mailto:chriwall@microsoft.com) for more details.
 
Would it be better to have a “WillChangesBeTrackedForCurrentUser()”? and “WillChangesBeTrackedForMachine()” instead?
No. Tracking works for everyone or no one if the DEH is present/absent, no difference if it works for user vs machine.
I suggest doing the check mentioned
 
FindPackagesByPackageFamily(Frameworks, &count)
IF count>0
    return true 
IF IsElevated()
    arch = GetCurrentArchitecture()
    FOREACH family in [WAR 1.1, WAR 1.2, etc]
        FOREACH p IN pkgmgr.FindPackages(family, PackageType_Framework)
            IF p.Id.ProcessorArchitecture == arch
                 return true
return false

Is my implementation here not correct?

@jonwis Yeah.  I expected this isn't the best way to find if a framework package is installed.  However, the family name of the framework package has the version in it.  A simple `GetPackagesFromPackageFamily` won't work with the name `Microsoft.WindowsAppRuntime_8wekyb3d8bbw`  Does another API call exist that allows me to get all framework packages on a machine?

dev/EnvironmentManager/ChangeTracker/EnvironmentVariableChangeTrackerHelper.h
dhoehna: Awesome.  Thank you for that.  Code looks a lot less ugly. :)

dev/EnvironmentManager/ChangeTracker/EnvironmentVariableChangeTrackerHelper.h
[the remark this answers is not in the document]
dhoehna: Removed this.  @DrusTheAxe and I discussed this offline.  We agreed to "If Self-contained tracking is not supported."

If we can figure out how to guarantee that the framework dll will be loaded into the system at uninstall time that can be added here as a check.

dev/EnvironmentManager/ChangeTracker/EnvironmentVariableChangeTrackerHelper.h
dhoehna: Removed.

dev/EnvironmentManager/ChangeTracker/EnvironmentVariableChangeTrackerHelper.h
dhoehna: I don't understand what you mean by this comment.  What should be moved to a cpp file?  ShouldChengesBeTracked?  Why?

dhoehna:
Self contained checks work.  

Changes will not be tracked if Windows App SKD is self-contained because windowsAppSDK can not guarantee that the framework bits will be loaded at uninstall time to revert any changes.

dev/EnvironmentManager/ChangeTracker/EnvironmentVariableChangeTrackerHelper.h
dhoehna: Feel free to unresolve if you believe items need to be moved to a .cpp file.

BuildTransportPackage.ps1
```diff
@@ -66,15 +68,11 @@ if(-not (test-path "$fullNugetPath\build\native"))
     new-item -path "$fullNugetPath\build\native" -itemtype "directory"
 }
 
-if(-not (test-path "$fullNugetPath\mrt_raw"))
+if(-not (test-path "$fullNugetPath\lib\net5.0-windows10.0.17763.0"))
```
DrusTheAxe, line 71: Nice but what's it got to do with EM?

dev/EnvironmentManager/ChangeTracker/EnvironmentVariableChangeTrackerHelper.h
```diff
@@ -0,0 +1,48 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+#pragma once
```
DrusTheAxe, line 3: NIT: Blank line after copyright

dev/EnvironmentManager/ChangeTracker/EnvironmentVariableChangeTrackerHelper.h
```diff
@@ -0,0 +1,48 @@
...
+#include <memory>
+#include <wil/result.h>
+#include <appmodel.h>
+#include <winrt/Windows.Management.Deployment.h>
+#include <Microsoft.Windows.System.EnvironmentManager.h>
+#include <AppModel.Identity.h>
+#include <Security.IntegrityLevel.h>
+#include <Microsoft.Windows.System.EnvironmentManager.Insights.h>
+#include <WindowsAppRuntime.SelfContained.h>
+
```
DrusTheAxe, line 15: NIT: Remove extra blank line

dev/EnvironmentManager/ChangeTracker/EnvironmentVariableChangeTrackerHelper.h
```diff
@@ -0,0 +1,48 @@
...
+static bool DoesChangeTrackingKeyExist()
+{
+    const auto subKey = std::filesystem::path{ LR"(Software\Microsoft\AppModel\RegistryWriteVirtualization\ExcludedKeys\HKEY_CURRENT_USER/Software/ChangeTracker)" };
+    wil::unique_hkey keyToChangeTracker{};
+    LSTATUS openResult{ RegOpenKeyEx(HKEY_LOCAL_MACHINE, subKey.c_str(), 0, KEY_READ | KEY_WOW64_64KEY, keyToChangeTracker.put()) };
+
+    if (FAILED_WIN32(openResult))
+    {
+        EnvironmentManagerInsights::TraceLoggingInfo("Could not find the change tracking HKLM key.  Changes will not be tracked");
```
DrusTheAxe, line 27: Where's this log to? If telemetry then the english prose shouldn't be necessary. If the eventlog where, what about localization, yadda yadda. Elsewhere?

dev/EnvironmentManager/ChangeTracker/EnvironmentVariableChangeTrackerHelper.h
```diff
@@ -0,0 +1,48 @@
...
+        LOG_IF_WIN32_ERROR(openResult);
```
DrusTheAxe, line 28: What's the difference between FAILED_WIN32() and LOG_IF_WIN32_ERROR?

If you get here isn't openResult guaranteed to be a WIN32_ERROR? So this should be LOG_WIN32_ERROR(openResult)?

Of course why both? Why not

```
const HRESULT hr{ LOG_IF_WIN32_ERROR(RegOpenKeyEx(...) };
if (FAILED(hr))
{
    EnvironmentManagerInsights::TraceLoggingInfo(...)
    return false;
}

File: dev/EnvironmentManager/ChangeTracker/EnvironmentVariableChangeTrackerHelper.h [hunk 12]
```diff
@@ -0,0 +1,48 @@
...
+
+    return true;
+}
+
+inline bool ShouldChangesBeTracked(EnvironmentManager::Scope scope)
+{
+    // process scope changes are never tracked becuase they go away when the process ends.
+    // Non packaged apps don't have identity.  No way to revert changes, so don't track them.
+    // The framework package needs to be on the system, or, loaded into the system when a package is uninstalled.
+    // CUrrently no way exists to see if the framework dll will be loaded into the system when a self-contained deployment is uninstalled.
```
DrusTheAxe, line 40: You don't need to _use_ the framework package to detect it's _present_
If you're elevated you can check if it's on the machine
If you're not elevated you can check if it's registered to the current user
Or you can say SelfContained simply don't get the tracking feature full stop.

Code seems to be doing the latter. Which is cool, just the comment doesn't quite express that

DrusTheAxe (approved):
LGTM modulo some comments

BuildTransportPackage.ps1
dhoehna: This wasn't supposed to be in here.  I'll remove it.

dev/EnvironmentManager/ChangeTracker/EnvironmentVariableChangeTrackerHelper.h
dhoehna: Added.

dev/EnvironmentManager/ChangeTracker/EnvironmentVariableChangeTrackerHelper.h
dhoehna: Removed.

dev/EnvironmentManager/ChangeTracker/EnvironmentVariableChangeTrackerHelper.h
dhoehna: Yes that.  I wanted to log the WIN32 if it was an error and change the return type depending on the return code.

This was the way I thought to do it.

However, your way is much better.

dev/EnvironmentManager/ChangeTracker/EnvironmentVariableChangeTrackerHelper.h
dhoehna: True.  Since a LOG_IF_WIN32 now exists.  The LOG_ will work.

This would go to telemetry and telemetry would not need a "HKLM key not found"

dev/EnvironmentManager/ChangeTracker/EnvironmentVariableChangeTrackerHelper.h
dhoehna: Changed.

PR #2955, opened 2022-09-09
WinUI 3 new project templates have incorrect comment about OnLaunched

ShashankNay:
Changing comments in the `OnLaunched` method for Desktop applications. These comments held true for UWP but `OnLaunched` always gets called.


Scottj1s (approved):
@ShashankNay, any idea why there's a diff on every line 1?  Do you have nonstandard line terminators set in your editor and/or git config?

bpulliam (approved):

ShashankNay:
> @ShashankNay, any idea why there's a diff on every line 1? Do you have nonstandard line terminators set in your editor and/or git config?

I'm not sure, my git config has autocrlf = true and things look okay in VS.

PR #2958, opened 2022-09-10
Unify payload RawNotification delivery object for packaged/unpackaged apps 

pmpurifoy:

and run the build against your PR branch with the default paramters.

Problem:
Currently, the Push Notifications LRP doesn't not correctly mock IBackgroundTaskInstance::TriggerDetails property when delivering raw notifications to apps using WindowsAppSDK. Instead of storing an IRawNotification object, we cast winrt::hstring->winrt::IInspectable and do some special casing to check if TriggerDetails was IRawNotification or winrt::hstring which causes unnecessary complexity.

Solution:
Only use IRawNotification when getting the payload from TriggerDetails. This means correctly mocking the IRawNotification object from the LRP to send to the target app.

Validation:
Run unit tests and manually deliver payloads to packaged/unpackaged apps.


dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
```diff
@@ -60,54 +49,30 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
     {
         return { m_rawNotificationPayload.data(), m_rawNotificationPayload.data() + (m_rawNotificationPayload.size() * sizeof(uint8_t)) };
     }
 
     winrt::BackgroundTaskDeferral PushNotificationReceivedEventArgs::GetDeferral()
     {
-        if (!m_unpackagedAppScenario)
-        {
-            THROW_HR_IF_NULL_MSG(E_ILLEGAL_METHOD_CALL, m_backgroundTaskInstance, "Foreground activation cannot call this.");
+        THROW_HR_IF_MSG(E_ILLEGAL_METHOD_CALL, m_isForegroundActivation, "Foreground activation cannot call this.");
```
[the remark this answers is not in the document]
pmpurifoy: So what happened in this PR is that instead of switching on unpackaged | packagedAppScenario, we switch on is it foreground | background, which makes more sense because PushActivation contains both foreground/background. 

The reason we no longer have to switch on unpackaged/packaged is because we are correctly mocking the IBackgroundTaskInstance and unpackaged scenarios can call into the background functions and not throw (albeit they will receive dummy values).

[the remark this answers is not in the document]
pmpurifoy: DCRs are only for late features and this is really just a refactoring task, so we can wait for the 1.3 payload to ship this.

dev/PushNotifications/PushRawNotification.h
```diff
@@ -0,0 +1,38 @@
+﻿// Copyright (c) Microsoft Corporation and Contributors.
+// Licensed under the MIT License.
+
+#include <winrt/Windows.Networking.PushNotifications.h>
+#include <winrt/Windows.ApplicationModel.background.h>
+#include <winrt/Windows.Storage.Streams.h>
+#include <winrt/Windows.Security.Cryptography.h>
+
+struct PushRawNotification : winrt::implements<PushRawNotification, winrt::Windows::Networking::PushNotifications::IRawNotification, winrt::Windows::Networking::PushNotifications::IRawNotification2, winrt::Windows::Networking::PushNotifications::IRawNotification3>
```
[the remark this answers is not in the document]
pmpurifoy: The goal of PushRawNotification was to mock the RawNotification object returned by BI. This allows us to have a single constructor that accepts an IBackgroundTaskInstance instead of just wrapping the string as IInspectable and needing two constructors to handle a string or RawNotification.

dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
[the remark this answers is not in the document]
pmpurifoy: I see what you are saying about the behavior change. 

IMO, it feels like it is a product code bug to be able to return background features of PushActivatedEventArgs in a foreground activation context without throwing. This PR correctly throws if the app was activated via the foreground and the app requests a background feature.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -189,8 +189,9 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
                 if (pair.Name() == L"payload")
                 {
                     // Convert escaped components to its normal content from the conversion done in the Long Running Process (see NotificationListener.cpp)
-                    auto payloadAsWstring = winrt::Windows::Foundation::Uri::UnescapeComponent(pair.Value());
-                    eventArgs = winrt::make<winrt::Microsoft::Windows::PushNotifications::implementation::PushNotificationReceivedEventArgs>(payloadAsWstring);
+                    auto payloadAsWstring{ winrt::Windows::Foundation::Uri::UnescapeComponent(pair.Value()) };
```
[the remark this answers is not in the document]
pmpurifoy: Fixed.

dev/PushNotifications/PushRawNotification.h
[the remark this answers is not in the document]
danielayala94: Closing comment after discussing offline, no changes needed 😊.

dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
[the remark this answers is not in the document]
pmpurifoy: What do you think about this change? Should it go through review before moving to product code? @DrusTheAxe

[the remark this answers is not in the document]
pmpurifoy: Reverted change

dev/PushNotifications/PushBackgroundTaskInstance.h
```diff
@@ -1,26 +1,28 @@
...
+﻿// Copyright (c) Microsoft Corporation and Contributors.
 // Licensed under the MIT License.
 
 #include <winrt/Windows.ApplicationModel.background.h>
+#include "PushRawNotification.h"
+#include "PushNotificationDummyDeferral.h"
 
 // Mocks IBackgroundTaskInstance to send raw payloads to packaged apps in
 // PushNotificationBackgroundTask::Run by com activation from PushNotificationsLongRunningProcess
 struct PushBackgroundTaskInstance : winrt::implements<PushBackgroundTaskInstance, winrt::Windows::ApplicationModel::Background::IBackgroundTaskInstance>
```
hulumane, line 10: Isn't this a MockBackgroundTaskInstance technically speaking? If so, shouldn't we rename as such?

dev/PushNotifications/PushBackgroundTaskInstance.h
```diff
@@ -1,26 +1,28 @@
...
 {
     PushBackgroundTaskInstance() {};
-    PushBackgroundTaskInstance(std::wstring const& payload): m_payload(payload) {};
+    PushBackgroundTaskInstance(std::wstring const& payload) : m_rawNotification(winrt::make_self<PushRawNotification>(payload)) {};
 
     winrt::guid InstanceId() { return winrt::guid(); };
     UINT32 SuspendedCount() { return 0; };
     UINT32 Progress() { return 0; };
     UINT32 Progress(UINT32 /* progress */) { return 0; };
     winrt::Windows::ApplicationModel::Background::BackgroundTaskRegistration Task() { return nullptr; };
```
hulumane, line 19: This is returning a nullptr too and could result in a crash just like GetDeferral. Should we track through a seperate bug?

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -870,16 +871,7 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-            {
-                m_backgroundTaskArgs = winrt::make<winrt::Microsoft::Windows::PushNotifications::implementation::PushNotificationReceivedEventArgs>(taskInstance);
-            }
-            else
-            {
-                // Need to mock a RawNotification object instead of winrt boxing: https://github.com/microsoft/WindowsAppSDK/issues/2075
-                winrt::hstring payload{ winrt::unbox_value<winrt::hstring>(taskInstance.TriggerDetails()) };
-                m_backgroundTaskArgs = winrt::make<winrt::Microsoft::Windows::PushNotifications::implementation::PushNotificationReceivedEventArgs>(payload.c_str());
-            }
+            m_backgroundTaskArgs = winrt::make<PushNotificationReceivedEventArgs>(taskInstance);
```
hulumane, line 874: Thanks for fixing. So much easier to look at and read :)

dev/PushNotifications/PushBackgroundTaskInstance.h
pmpurifoy: This is only used for LRP->App. But I can rename to MockBackgroundTaskInstance instead

dev/PushNotifications/PushBackgroundTaskInstance.h
pmpurifoy: We don't expose this attribute in PushNotificationReceivedEventArgs so there is no risk in leaving it as nullptr. Similar to SuspendedCount/InstanceId we don't actually use those properties other than to align with the IBackgroundTaskInstance interface.

dev/PushNotifications/PushNotificationUtility.h
```diff
@@ -8,11 +8,12 @@
 #include "../Common/AppModel.Identity.h"
 #include "wil/stl.h"
 #include "wil/win32_helpers.h"
-#include "PushBackgroundTaskInstance.h"
+#include "MockBackgroundTaskInstance.h"
```
loneursid, line 11: This feels strange. I wouldn't expect mocks to appear outside of test directories.

dev/PushNotifications/MockBackgroundTaskInstance.h
```diff
@@ -1,34 +1,36 @@
-// Copyright (c) Microsoft Corporation and Contributors.
+﻿// Copyright (c) Microsoft Corporation and Contributors.
 // Licensed under the MIT License.
```
sharath2727, line 2: Can you add a description in this format:

Motivation:

What changed:

Validation:

Makes it easy to understand why it is introduced in the first place?

dev/PushNotifications/PushNotificationUtility.h
sharath2727: Yeah that is the first thing I thought when I saw the "mock" naming convention.

dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
```diff
@@ -60,51 +49,45 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
         {
-            THROW_HR_IF_NULL_MSG(E_ILLEGAL_METHOD_CALL, m_backgroundTaskInstance, "Foreground activation cannot call this.");
-
-            return m_backgroundTaskInstance.Canceled(handler);
+            return winrt::event_token{};
         }
         else
         {
-            return { 0 };
+            THROW_HR_IF_MSG(E_ILLEGAL_METHOD_CALL, m_isForegroundActivation, "Foreground activation cannot call this.");
```
sharath2727, line 76: wondering if we need to do this? what is the scenario you think foreground activation can call into this?

pmpurifoy: Foreground cannot call this. Canceled is a property that is only usable in the background activation scenario.

dev/PushNotifications/MockBackgroundTaskInstance.h
pmpurifoy: Yeah I can rephrase it better

dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
sharath2727: then do we need a check paul?

pmpurifoy: Line 76 checks if it was activated via foreground and throws if true.

dev/PushNotifications/PushNotificationUtility.h
pmpurifoy: Rename to LRPBackgroundTaskInstance

loneursid: Works for me. iThe new name doesn't trigger me anymore :)

thanks.

loneursid (approved):
LGTM, but make sure you have addressed feedback from @hulumane and @sharath2727 before merging

dev/PushNotifications/LRPBackgroundTaskInstance.h
```diff
@@ -1,34 +1,36 @@
...
 // Licensed under the MIT License.
 
 #include <winrt/Windows.ApplicationModel.background.h>
+#include "LRPRawNotification.h"
+#include "PushNotificationDummyDeferral.h"
 
 // Mocks IBackgroundTaskInstance to send raw payloads to packaged apps in
 // PushNotificationBackgroundTask::Run by com activation from PushNotificationsLongRunningProcess
-struct PushBackgroundTaskInstance : winrt::implements<PushBackgroundTaskInstance, winrt::Windows::ApplicationModel::Background::IBackgroundTaskInstance>
+struct LRPBackgroundTaskInstance : winrt::implements<LRPBackgroundTaskInstance, winrt::Windows::ApplicationModel::Background::IBackgroundTaskInstance>
```
hulumane, line 10: I still feel that adding acronyms in names should be discouraged. Nobody really outside of our team with specific domain knowledge understands LRP. Maybe we call it LongRunningProcessSourcedTaskInstance which is a clear indication of what this really is?

dev/PushNotifications/LRPRawNotification.h
```diff
@@ -0,0 +1,38 @@
+﻿// Copyright (c) Microsoft Corporation and Contributors.
+// Licensed under the MIT License.
+
+#include <winrt/Windows.Networking.PushNotifications.h>
+#include <winrt/Windows.ApplicationModel.background.h>
+#include <winrt/Windows.Storage.Streams.h>
+#include <winrt/Windows.Security.Cryptography.h>
+
+struct LRPRawNotification : winrt::implements<LRPRawNotification, winrt::Windows::Networking::PushNotifications::IRawNotification, winrt::Windows::Networking::PushNotifications::IRawNotification2, winrt::Windows::Networking::PushNotifications::IRawNotification3>
```
hulumane, line 9: Same comment with LRP here as well.

dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
```diff
@@ -60,51 +49,45 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
+
+            return m_backgroundTaskInstance.Canceled(handler);
         }
     }
 
     void PushNotificationReceivedEventArgs::Canceled(winrt::event_token const& token) noexcept
     {
-        if (!m_unpackagedAppScenario)
+        if (!PushNotificationHelpers::IsPackagedAppScenario())
```
hulumane, line 84: Nit: simplify and avoid an extra line.

if (PushNotificationHelpers::IsPackagedAppScenario())
{
      THROW_HR_IF_MSG(E_ILLEGAL_METHOD_CALL, m_isForegroundActivation, "Foreground activation cannot call this.");

            m_backgroundTaskInstance.Canceled(token);
}

dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
```diff
@@ -60,51 +49,45 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
-
-            return m_backgroundTaskInstance.GetDeferral();
-        }
-        else
-        {
-            auto dummyDeferral = winrt::make<PushNotificationDummyDeferral>();
-            return dummyDeferral.as<winrt::BackgroundTaskDeferral>();
+            return winrt::make<PushNotificationDummyDeferral>().as<BackgroundTaskDeferral>();
         }
+        THROW_HR_IF_MSG(E_ILLEGAL_METHOD_CALL, m_isForegroundActivation, "Foreground activation cannot call this.");
```
hulumane, line 63: So why is this check not at the beginning of the function? Is that on purpose?

dev/PushNotifications/LRPBackgroundTaskInstance.h
hulumane: Yup. In general, never use acronyms because people who inherit code later cannot make head or tail of them. AuthManager for example is a very bad name. Auth could be Authentication. It could also be Authorization. 

Its actually better to be wordy than concise and indicate the meaning of the component being built.

hulumane: And when you add the name, also add a comment that this component should only be instantiated from the Long Running Process.

sharath2727 (approved):

dev/PushNotifications/PushNotificationReceivedEventArgs.cpp
hulumane: Ok so it sounds like this should be a check that is at the beginning of the function and applicable to both packaged/unpackaged scenarios? Or am i wrong?

pmpurifoy: Yup it should be, but since we have decided to noop from beginning, it would change the behavior of the API and that needs further review.

PR #3004, opened 2022-09-28
UrfwInitialize() fails if exe lacks a Fusion manifest

DrusTheAxe:
UrfwInitialize() fails if exe lacks a Fusion manifest

UrfwInitialize() calls ExtRoLoadCatalog() calls CreateActCtxW() to create an activation context with the executable's Fusion manifest. If the exe has no manifest CreateActCtxW() returns ERROR_INVALID_HANDLE with GetLastError=ERROR_RESOURCE_DATA_NOT_FOUND. ExtRoLoadCatalog() detects the failure but (1) `SetLastError(ERROR_OUTOFMEMORY)` due to a developer mistake copy-pasta error handling from a wil::make_unique call (i.e. memory allocation) and (2) treats not-found as an error.

Executables aren't required to have Fusion manfiests. Requiring one is an error.

To repro, create a tiny trivial exe with NO Fusion manifest and 

```
int main()
{
    LoadLibrary("Microsoft.WindowsAppRuntime.dll")
    return 0;
}
```

Run under the debugger you'll see Microsoft.WindowsAppRuntime.dll's DllMain() calls into this code and FAIL_FAST on the error.

https://task.ms/40349008

### DIFF
--- a/dev/UndockedRegFreeWinRT/urfw.cpp
+++ b/dev/UndockedRegFreeWinRT/urfw.cpp
@@ -317,13 +317,13 @@ HRESULT WINAPI RoResolveNamespaceDetour(
 
 HRESULT ExtRoLoadCatalog()
 {
-    WCHAR filePath[MAX_PATH];
+    WCHAR filePath[MAX_PATH]{};
     if (!GetModuleFileNameW(nullptr, filePath, _countof(filePath)))
     {
         return HRESULT_FROM_WIN32(GetLastError());
     }
     std::wstring manifestPath(filePath);
-    HANDLE hActCtx = INVALID_HANDLE_VALUE;
+    HANDLE hActCtx{ INVALID_HANDLE_VALUE };
     auto exit = wil::scope_exit([&]
     {
         if (hActCtx != INVALID_HANDLE_VALUE)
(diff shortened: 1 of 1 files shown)


DrusTheAxe:
Tested via "LoadLibrary.exe Microsoft.WindowsAppRuntime.dll" where the .exe has no Fusion manifest and calls LoadLibrary), plus walking thru the debugger comparing before (DllMain fails thus failing the load) and after (success and happiness and puppy dogs and candy)

DrusTheAxe:
UrfwInitialize() fails if exe lacks a Fusion manifest

`UrfwInitialize()` calls `ExtRoLoadCatalog()` calls `CreateActCtxW()` to create an activation context with the executable's Fusion manifest. If the exe has no manifest `CreateActCtxW()` returns `ERROR_INVALID_HANDLE` with `GetLastError`=`ERROR_RESOURCE_DATA_NOT_FOUND`. `ExtRoLoadCatalog()` detects the failure but (1) `SetLastError(ERROR_OUTOFMEMORY)` due to a developer mistake copy-pasta error handling from a `wil::make_unique` call (i.e. memory allocation) and (2) treats not-found as an error.

Executables aren't required to have Fusion manifests. Requiring one is an error.

To repro, create a tiny trivial exe with NO Fusion manifest and

```
int main()
{
    LoadLibrary("Microsoft.WindowsAppRuntime.dll")
    return 0;
}
```

Run under the debugger you'll see Microsoft.WindowsAppRuntime.dll's `DllMain()` calls into this code and FAIL_FAST on the error.

https://task.ms/40349008

lyahdav:
When attempting to upgrade our unpackaged app that uses WinAppSDK from 1.0.4 to 1.2 preview 2 we started hitting the issue that this PR fixes. @DrusTheAxe will this fix go into WinAppSDK 1.2?

DrusTheAxe:
> When attempting to upgrade our unpackaged app that uses WinAppSDK from 1.0.4 to 1.2 preview 2 we started hitting the issue that this PR fixes. @DrusTheAxe will this fix go into WinAppSDK 1.2?

There's a further refinement to the fix identified today, fixed via https://github.com/microsoft/WindowsAppSDK/pull/3093

In 1.1 if the exe for the process calling `LoadLibrary("microsoft.windowsappruntime.dll") had no embedded SxS manifest we would FAIL_FAST in DllMain.

In 1.2-preview2 you'd succeed if the exe had no resources but still FAIL_FAST if the exe had resources but not a SxS manifest.

PR #3093 fixes that to ignore both cases (not just the one).

As a workaround if your exe has no resources -OR- has resources AND an embedded SxS manifest (even with no meaningful content) it should avoid this problem. I don't know if an empty application.manifest is valid

```
<assembly xmlns="urn:schemas-microsoft-com:asm.v1" manifestVersion="1.0">
</assembly>
```

If not then something harmless would do e.g. `<heapType>` with values other than `SegmentHeap` are ignored per [this](https://learn.microsoft.com/windows/win32/sbscs/application-manifests#heaptype)

```
<assembly xmlns="urn:schemas-microsoft-com:asm.v1" manifestVersion="1.0" xmlns:asmv3="urn:schemas-microsoft-com:asm.v3">
  <asmv3:application>
    <asmv3:windowsSettings xmlns="http://schemas.microsoft.com/SMI/2020/WindowsSettings">
      <heapType>AllOtherValuesAreIgnored</heapType>
    </asmv3:windowsSettings>
  </asmv3:application>
</assembly>
```

We briefly discussed the fix late today. It's expected to be serviced in 1.2.1; it missed the window for 1.2.0, unless it's deemed a recall-class bug. That's the current thinking but please feel free to make your case, feedback is always welcome and after all, squeaky wheel gets the grease :P 

@MikeHillberg for visibility

PR #3027, opened 2022-10-07
Enable elevation for PushNotifications

pmpurifoy:

and run the build against your PR branch with the default paramters.


dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -132,7 +132,7 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
     {
         // Elevated processes are not supported by PushNotifications. UnpackagedAppScenario is not supported when it is self contained
```
DrusTheAxe, line 133: Stale comment?

dev/PushNotifications/PushNotificationUtility.h
```diff
@@ -127,9 +130,18 @@ namespace winrt::Microsoft::Windows::PushNotifications::Helpers
     }
     CATCH_RETURN()
 
+    inline void ClearRoActivateInstanceCache()
+    {
+        // Activating a nonexistent class clears the cache enables the proxyStub to be found for elevated scenarios
+        auto activatableClass{ wil::make_unique_string_nothrow<wil::unique_hstring>(L"FakeClass.ClassName") };
+        winrt::com_ptr<::IInspectable> ptr{};
+
+        RoActivateInstance(activatableClass.get(), ptr.put());
```
DrusTheAxe, line 139: No check of the return code?

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -116,6 +116,8 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
             AppNotificationTelemetry::LogRegister(hr, m_appId);
         }) };
 
+        // Need to clear the RoActivateInstance caching for LRP proxyStub to be found.
```
hulumane, line 119: LRP references being made again here are unclear. Also hy should it be cleared? Is it a bug? If so, what is the bug number?
I would also move this outside of AppNotification APIs if possible. Is it possible to do it during install time?

dev/PushNotifications/PushNotificationUtility.h
```diff
@@ -127,9 +130,18 @@ namespace winrt::Microsoft::Windows::PushNotifications::Helpers
     }
     CATCH_RETURN()
 
+    inline void ClearRoActivateInstanceCache()
+    {
+        // Activating a nonexistent class clears the cache enables the proxyStub to be found for elevated scenarios
+        auto activatableClass{ wil::make_unique_string_nothrow<wil::unique_hstring>(L"FakeClass.ClassName") };
```
hulumane, line 136: What is the perf overhead of doing this? Is there a bug to fix the root cause here that we need to add to comments?

test/LRPTests/APITests.cpp
```diff
@@ -30,7 +30,7 @@ namespace Test::LRP
 
         winrt::com_ptr<INotificationsLongRunningPlatform> GetNotificationPlatform()
         {
-            winrt::com_ptr notificationPlatform{ winrt::try_create_instance<INotificationsLongRunningPlatform>(_uuidof(NotificationsLongRunningPlatform), CLSCTX_ALL) };
+            winrt::com_ptr notificationPlatform{ winrt::try_create_instance<INotificationsLongRunningPlatform>(CLSID_NotificationsLongRunningPlatform, (CLSCTX_LOCAL_SERVER | CLSCTX_ALLOW_LOWER_TRUST_REGISTRATION)) };
```
hulumane, line 33: Why remove CLSCTX_ALL? It should cover all flags including local server and lower trust right?

dev/PushNotifications/PushNotificationsLongRunningTask.ProxyStub/NotificationsLongRunningProcess.idl
```diff
@@ -38,13 +38,4 @@ interface INotificationsLongRunningPlatform : IUnknown
...
-[version(1.0)]
-library NotificationsLongRunningPlatformLibrary
-{
-    [uuid(PUSHNOTIFICATIONS_IMPL_CLSID_UUID)]
-    coclass NotificationsLongRunningPlatform
-    {
-        [default] interface INotificationsLongRunningPlatform;
-    };
-};
+declare_guid(CLSID_NotificationsLongRunningPlatform, PUSHNOTIFICATIONS_IMPL_CLSID_UUID);
```
hulumane, line 41: Interesting. Can you explain why this was changed?

hulumane: For my understanding :)

test/TestApps/PushNotificationsDemoApp/main.cpp
```diff
@@ -90,9 +98,11 @@ winrt::PushNotificationChannel RequestChannel()
 
 int main()
 {
+
```
hulumane, line 101: undo this new line please

test/TestApps/PushNotificationsDemoApp/main.cpp
```diff
@@ -10,7 +10,15 @@
 #include <propkey.h> //PKEY properties
 #include <propsys.h>
 #include <ShObjIdl_core.h>
+#include "WindowsAppRuntime.Test.AppModel.h"
 #include <WindowsAppRuntime.Test.Bootstrap.h>
+#include <WindowsAppRuntime.SelfContained.h>
```
hulumane, line 15: Why are these includes in main.cpp? I don't see any code added to main.cpp that references these newer includes

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -167,6 +169,8 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
...
+        PushNotificationHelpers::ClearRoActivateInstanceCache();
```
hulumane, line 173: Instead of doing this everytime you call CoCreateIntance on LongRunningPlatform, you could do the following.

1) Create a utility method called GetLongRunningPlatform.
2) The first thing this does would be to clear the cache for RoActivate
3) It would then CoCreate a new platform and return it.

By doing this, you have one single method that all components can reuse and a single place for the mitigation.

dev/PushNotifications/PushNotificationUtility.h
brialmsft: OS bug is 39521331. Vestiges of caching logic predating dynamic dependencies cause WinRT proxy/stub creation paths to overlook the fact that the package graph has become non-empty.

As far as perf goes, the bug only requires you to do this fake activation _once_ after dynamic dependencies have taken the package graph size from 0 -> 1. So if this code is only ever run after dynamic dependencies have been added, then you could just make ClearRoActivateInstanceCache do the bogus activation once, and then no-op on all subsequent calls.

hulumane: Thanks. Then wouldn't dynamic dependencies be a better spot to add this rather than notifications? I imagine there may be other feature sets that are also impacted here and could use the goodness.

dev/PushNotifications/PushNotificationUtility.h
brialmsft: What check should be performed? This activation is supposed to fail. You _could_ validate that it fails with the expected REGDB_E_CLASSNOTREG, but that's probably over-validating. If it somehow succeeds (e.g. if someone actually implements a class named "FakeClass.ClassName" within the package graph; or if somebody detours RoActivateInstance to always succeed), I guess you _could_ crash, but barring that you would just want to clean up the activated object and move on, which is done through RAII.

From a style perspective it's probably worth commenting and maybe casting the return to void, just to make it clear that it's being ignored deliberately. But functionally this should be fine.

test/TestApps/PushNotificationsDemoApp/main.cpp
```diff
@@ -10,7 +10,15 @@
...
+#include <WindowsAppRuntime.VersionInfo.h>
+#include <roapi.h>
+#include <roapi.h>
```
hulumane, line 18: roapi.h is included twice

dev/PushNotifications/PushNotificationsLongRunningTask.ProxyStub/NotificationsLongRunningProcess.idl
brialmsft: I'd generally consider using library blocks just to put a CLSID in the header (and uuid lib) to be a minor anti-pattern. We're not actually interested in generating a typelib here, so there's no real functional advantage to the library block.

Arguably it's providing some extra value by describing the interfaces impelemented on the CLSID, but 1) there's nothing enforcing/requiring the interface list to be accurate, so it has no more weight than a comment; and 2) in simple cases like this, convention strongly suggests that CLSID_NotificationsLongRunningPlatform probably implements INotificationsLongRunningPlatform.

test/LRPTests/APITests.cpp
brialmsft: CLSCTX_ALL isn't all of the CLSCTX flags:
`#define CLSCTX_ALL              (CLSCTX_INPROC_SERVER| \
                                 CLSCTX_INPROC_HANDLER| \
                                 CLSCTX_LOCAL_SERVER| \
                                 CLSCTX_REMOTE_SERVER)`

CLSCTX_ALL is for "I have a CLSID and I want to activate it, but I don't know/care whether it's inproc, out-of-proc, on a completely different machine, etc.".
In theory, that's a reasonable default perspective: the client is agnostic about where the server class actually runs, allowing the server location to be treated as an implementation detail. In practice, the fact that CoCreateInstance _allows_ the client to specify this prevents any such agnosticism from being useful. If I want to change my server's activation type from out-of-proc to inproc, I can't unless I'm willing to break any existing clients which pass CLSCTX_LOCAL_SERVER instead of CLSCTX_ALL.

So CLSCTX_ALL doesn't end up adding much real value in most scenarios. In cases like this where your client and server are tightly coupled and the client knows the server is out-of-proc, you might as well just pass CLSCTX_LOCAL_SERVER. It makes the relationship to the server a little clearer, and should also have very minor perf benefits (e.g. avoids some registry queries).

hulumane: Thanks for the detailed explanation. Its good to understand some of the intricacies of COM that we simply take for granted when coding solutions up. I incorrectly assumed ALL mapped to everything. But in any case, it sounds like we shouldn't be using the Big Hammer(Perf, Tight coupling) and it is important to be specific when it comes to COM flags.

dev/PushNotifications/PushNotificationUtility.h
brialmsft: > Then wouldn't dynamic dependencies be a better spot to add this rather than notifications?
That probably would be a little bit cleaner. @DrusTheAxe, thoughts?

dev/PushNotifications/PushNotificationUtility.h
```diff
@@ -127,9 +130,18 @@ namespace winrt::Microsoft::Windows::PushNotifications::Helpers
...
+        winrt::com_ptr<::IInspectable> ptr{};
+
+        RoActivateInstance(activatableClass.get(), ptr.put());
+    }
+
     inline wil::com_ptr<INotificationsLongRunningPlatform> GetNotificationPlatform()
     {
-        return wil::CoCreateInstance<NotificationsLongRunningPlatform, INotificationsLongRunningPlatform>(CLSCTX_LOCAL_SERVER);
+        return wil::CoCreateInstance<INotificationsLongRunningPlatform>(CLSID_NotificationsLongRunningPlatform, (CLSCTX_LOCAL_SERVER | CLSCTX_ALLOW_LOWER_TRUST_REGISTRATION));
```
brialmsft, line 144: It's worth noting that since CLSCTX_ALLOW_LOWER_TRUST_REGISTRATION was serviced downlevel, there's a chance that passing the flag will cause CoCreateInstance to fail (even if the app is non-elevated) due to the flag being unrecognized on an unpatched machine. We did add a mechanism for detecting whether the flag is supported. So if you want to be robust against the possibility of this code running on unpatched downlevel machines, you might need to query.

`
// If supported, allow use of class registrations from less trusted sources.
auto supportMarker = wil::CoCreateInstanceNoThrow<ISupportAllowLowerTrustActivation>(CLSID_ActivationCapabilities);
if (supportMarker)
{
    activatedObject = wil::CoCreateInstance(CLSID_SampleClass, CLSCTX_LOCAL_SERVER |
        CLSCTX_ALLOW_LOWER_TRUST_REGISTRATION);
}
else
{
    // Even if we can't pass CLSCTX_ALLOW_LOWER_TRUST_REGISTRATION, we still want to attempt the activation.
    // This might fail to find the class when run elevated on a downlevel SKU, but it will work otherwise.
    activatedObject = wil::CoCreateInstance(CLSID_SampleClass, CLSCTX_LOCAL_SERVER);
}
`

test/TestApps/PushNotificationsDemoApp/main.cpp
pmpurifoy: AppModel can be removed, but self contained is supposed to check if the env is self-contained. I am going to add a check here.

hulumane: Right. Lets not add to the header bloat if at all possible.

dev/PushNotifications/PushNotificationManager.cpp
pmpurifoy: Fixed

dev/PushNotifications/PushNotificationUtility.h
```diff
@@ -141,7 +141,19 @@ namespace winrt::Microsoft::Windows::PushNotifications::Helpers
 
     inline wil::com_ptr<INotificationsLongRunningPlatform> GetNotificationPlatform()
     {
-        return wil::CoCreateInstance<INotificationsLongRunningPlatform>(CLSID_NotificationsLongRunningPlatform, (CLSCTX_LOCAL_SERVER | CLSCTX_ALLOW_LOWER_TRUST_REGISTRATION));
+        // If supported, allow use of class registrations from less trusted sources.
+        auto supportMarker{ wil::CoCreateInstanceNoThrow<ISupportAllowLowerTrustActivation>(CLSID_ActivationCapabilities) };
```
brialmsft, line 145: Probably best to cache supportMarker; it shouldn't change within the app's lifetime.

pmpurifoy: This will be wrapped up std::call_once

dev/PushNotifications/PushNotificationUtility.h
pmpurifoy: I'm not sure if this is correct, but I thought packaged apps don't need DynamicDependencies. And this fix is for both packaged/unpackaged apps so I don't think it can go in DynDep.

DrusTheAxe: What permutations of app+WinAppSDK need this?

image

pmpurifoy: Packaged+Unpackaged | MSIX. We don't support SelfContained since we can't guarantee the PushNotificationsLRP is installed.

DrusTheAxe: >the bug only requires you to do this fake activation once after dynamic dependencies have taken the package graph size from 0 -> 1

Is this only true of WinAppSDK's DynDep or also true for the [Win11 OS APIs](https://learn.microsoft.com/windows/win32/api/appmodel/nf-appmodel-addpackagedependency)?

pmpurifoy: The reason packaged wasn't working in the first place because it couldn't get the proxyStub from a trusted source. This PR adds the proxyStub to the WASfwk which is the fix for packaged. I just mixed up the two issues. Oops.

This means only unpackaged | MSIX (WAS) is broken due to the bug.

brialmsft: > Is this only true of WinAppSDK's DynDep or also true for the [Win11 OS APIs](https://learn.microsoft.com/windows/win32/api/appmodel/nf-appmodel-addpackagedependency)?

The bug is in the OS code, so it definitely affects the docked APIs. I'm pretty sure that WinAppSDK's DynDep doesn't manage to avoid the problem either, since IIRC the undocked DynDeps are focused primarily on activation, and don't do anything special for proxy/stub creation (which is reasonable; proxy/stub creation is way less straightforward to detour than inproc activation).

> If the bug is only when package graph goes from empty to not-empty (0->1) then why does Packaged need this?

Yeah, I don't follow that either. There are other parts of this change that are definitely relevant to both packaged and unpackaged apps, but the issue that ClearRoActivateInstanceCache is mitigating should only apply to processes that initially have an empty package graph.

dev/AppNotifications/AppNotificationManager.cpp
```diff
@@ -60,6 +60,12 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
     winrt::Microsoft::Windows::AppNotifications::AppNotificationManager AppNotificationManager::Default()
     {
         auto appProperties{ winrt::CoreApplication::Properties() };
+        static std::once_flag flag;
+        std::call_once(flag, []()
```
hulumane, line 64: Looking at the code below, you don't even need this call_once lambda. You can do this in the else clause below which is under the lock.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -130,15 +130,21 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
...
 
     }
 
     winrt::Microsoft::Windows::PushNotifications::PushNotificationManager PushNotificationManager::Default()
     {
+        static std::once_flag flag;
+        std::call_once(flag, []()
+        {
+            // Need to clear the RoActivateInstance caching for LRP proxyStub to be found.
+            PushNotificationHelpers::ClearRoActivateInstanceCache();
```
hulumane, line 145: SAme comment here. No need of the call_once lambda. Just move this to the else clause.

dev/PushNotifications/PushNotificationUtility.h
hulumane: @pmpurifoy What about RS5 scenario for MSIX. Don't they also rely on the Long Running Process to activate the app in question?

dev/PushNotifications/PushNotificationUtility.h
```diff
@@ -127,9 +130,30 @@ namespace winrt::Microsoft::Windows::PushNotifications::Helpers
...
+    }
+
     inline wil::com_ptr<INotificationsLongRunningPlatform> GetNotificationPlatform()
     {
-        return wil::CoCreateInstance<NotificationsLongRunningPlatform, INotificationsLongRunningPlatform>(CLSCTX_LOCAL_SERVER);
+        // If supported, allow use of class registrations from less trusted sources.
+        auto supportMarker{ wil::CoCreateInstanceNoThrow<ISupportAllowLowerTrustActivation>(CLSID_ActivationCapabilities) };
+        if (supportMarker)
+        {
+            return wil::CoCreateInstance<INotificationsLongRunningPlatform>(CLSID_NotificationsLongRunningPlatform, CLSCTX_LOCAL_SERVER |
```
hulumane, line 148: So Enabling the lower trust flag allows for what exactly here? LRP relaunching the process or callbacks registered with LRP being invoked correctly?

dev/PushNotifications/PushNotificationUtility.h
pmpurifoy: As long as the packaged app has a declared dependency on the WASfwk package that contains the proxyStub, it will work since now it can trust the source of the proxyStub.

dev/PushNotifications/PushNotificationUtility.h
pmpurifoy: It allows elevated clients to get a reference to the LRP platform object.

The lower trust flag allows an elevated process to CoCreateInstance a lower trust COM server, which is in our case the elevated client -> LRP.

dev/AppNotifications/AppNotificationManager.cpp
pmpurifoy: Fixed.

dev/PushNotifications/PushNotificationUtility.h
pmpurifoy: Fixed.

dev/AppNotifications/AppNotificationManager.cpp
pmpurifoy: This is under discussion in other thread.

dev/AppNotifications/AppNotificationManager.cpp
pmpurifoy: Fixed LRP reference. Location discussion in other thread.

dev/PushNotifications/PushNotificationUtility.h
pmpurifoy: Adding comment and cast to explain impl.

pmpurifoy: Actually, I'm going to check for the REGDB_E_CLASSNOTREG. It feels weird to not verify the return value.

brialmsft: It's ultimately your choice. Error codes are a bit of a gray area in terms of "is this an implementation detail, or a stable part of the API's contract?" My general view is that if an API doesn't specifically document that it returns a specific error code for a given failure mode, then taking a dependency on that detail risks getting broken if the API's behavior changes in the future.

But with that said, there's enough precedent for checking for REGDB_E_CLASSNOTREG from COM/WinRT activations that it's definitely de facto part of the contract of those APIs. I'd say it's extremely unlikely we'd ever be able to change it, due to compat risk. So in practice, validating that specific error code is unlikely to be harmful.

dev/PushNotifications/PushNotificationUtility.h
DrusTheAxe: >As far as perf goes, the workaround for the bug only requires you to do this fake activation once 

Yes, but you didn't answer the question. MddBootstrapInitialize is likely the DynDep caller causing the pkggraph 0->1 transition. This is one of the 1st things an unpackaged app does in main and that's always a hotpath. Do you have a sense of the perf impact?

Doing this ClearRoCache in DynDep would be a tax on all unpackaged apps' critical startup path is one reason to do it in PushNotifications - defer it until needed so it's not in app startup hotpath, and only those who need it take the hit (Pay-to-Play)

If/when others besides PushNotifications need this we could move `ClearRoActivateInstanceCache()` to `dev\Common\Microsoft.RoApi.h` for reuse.

brialmsft: > As far as perf goes, the workaround for the bug only requires you to do this fake activation once

Perf impact would primarily be loading the private hive (activationStore.dat). The hive remains loaded, so it's a one-time cost. That's on top of any cost payed by the undocked detoured RoActivateInstance before reaching the docked code. That's definitely a cost that will be paid eventually, but I can't say for certain whether that happens within the startup critical path right now.

In terms of total perf cost (i.e. ignoring the fact that the critical path is critical), it would arguably be best to do this right after the 0 -> 1 transition, rather than waiting to some later point where the package graph size may be > 1. That way we're loading/querying as few packages as possible before failing. Not sure that matters in practice, if nobody other than MddBootstrapInitialize is setting up DynDeps for these apps.

Pay-to-play is almost certainly better for perf. The risk is that it's creating a potential pitfall for any other WASDK components that try to use WinRT-style proxy registrations on systems without the OS bugfix. So that's basically a maintainability/development risk weighed against a perf risk. I don't think either risk is prohibitive.

pmpurifoy: I just moved the function to dev/common/Microsoft.RoApi.h with a use case description for all WAS components to pay-to-play. Let me know if anything else is needed for this PR to be approved 😄

brialmsft (approved):

dev/Common/Microsoft.RoApi.h
```diff
@@ -0,0 +1,27 @@
...
+    // in manifest should call this function before due to a RoActivateInstance caching bug.
+    inline void ClearRoActivateInstanceCache()
+    {
+        // Activating a nonexistent class clears the cache enables the proxyStub to be found for elevated scenarios
+        auto activatableClass{ wil::make_unique_string_nothrow<wil::unique_hstring>(L"FakeClass.ClassName") };
+        winrt::com_ptr<::IInspectable> ptr{};
+
+        // Function is expected fail and if anything else happens we should throw.
+        winrt::hresult hr{ RoActivateInstance(activatableClass.get(), ptr.put()) };
+        THROW_HR_IF(hr, hr != REGDB_E_CLASSNOTREG);
```
hulumane, line 23: You probably want to assert that hr didn't succeed first. Unlikely, but someone could actually define a class with the name you put in above and the activation passes.

hulumane: Also if I understand correct, you cannot THROW S_OK. The macro would complain.

dev/PushNotifications/PushNotificationUtility.h
```diff
@@ -129,7 +130,19 @@ namespace winrt::Microsoft::Windows::PushNotifications::Helpers
...
+                CLSCTX_ALLOW_LOWER_TRUST_REGISTRATION);
+        }
+        else
+        {
+            // Even if we can't pass CLSCTX_ALLOW_LOWER_TRUST_REGISTRATION, we still want to attempt the activation.
+            // This might fail to find the class when run elevated on a downlevel SKU, but it will work otherwise.
+            return wil::CoCreateInstance<INotificationsLongRunningPlatform>(CLSID_NotificationsLongRunningPlatform, CLSCTX_LOCAL_SERVER);
```
hulumane, line 144: Should IsSupported() return false for this case?

dev/PushNotifications/PushNotificationsLongRunningTask.StartupTask/PushNotificationsLongRunningPlatform-Startup.h
```diff
@@ -24,9 +25,21 @@ inline HRESULT StartupNotificationsLongRunningPlatform()
     {
         try
         {
-            wil::com_ptr<INotificationsLongRunningPlatform> longRunningProcessPlatform
-            { wil::CoCreateInstance<NotificationsLongRunningPlatform, INotificationsLongRunningPlatform>(CLSCTX_LOCAL_SERVER) };
+            ::Microsoft::RoApi::ClearRoActivateInstanceCache();
 
+            // If supported, allow use of class registrations from less trusted sources.
+            auto supportMarker{ wil::CoCreateInstanceNoThrow<ISupportAllowLowerTrustActivation>(CLSID_ActivationCapabilities) };
```
hulumane, line 31: Common code. Likely requires a Utility function

dev/PushNotifications/PushNotificationUtility.h
hulumane: Would it make sense for IsSupported to also make this call?

pmpurifoy: Need to discuss how to address this in another PR

dev/PushNotifications/PushNotificationsLongRunningTask.StartupTask/PushNotificationsLongRunningPlatform-Startup.h
pmpurifoy: Done.

dev/Common/Microsoft.RoApi.h
pmpurifoy: Yup, just added another error line.

dev/PushNotifications/PushNotificationUtility.h
pmpurifoy: Tracking issue: https://github.com/microsoft/WindowsAppSDK/issues/3063

dev/PushNotifications/PushNotificationsLongRunningTask.StartupTask/PushNotificationsLongRunningPlatform-Startup.h
```diff
@@ -5,6 +5,8 @@
 
 #include <wil/com.h>
 #include <wil/resource.h>
+#include "../../Common/Microsoft.RoApi.h"
+#include "../../PushNotifications/NotificationPlatformActivation.h"
```
hulumane, line 9: In general, we should avoid relative paths for header files. Instead the project file should include the dirs that we care about and then, this wouldn't be needed.

Same comment for other files where relative headers are used.

hulumane (approved):

PR #3029, opened 2022-10-07
MRTCore take ms-resource URI

huichen123:
WinUI3 needs to pass in full URI in some cases. Make MRTCore handle that. When full URI is provided, the parent resource map is ignored.


axelandrejs (approved):

dev/MRTCore/mrt/Core/src/MRM.cpp
```diff
@@ -28,12 +28,55 @@ typedef struct _MrmObjects
...
+    return ((wcslen(resourceId) > static_cast<size_t>(ResourceUriPrefixLength)) && (CompareStringOrdinal(ResourceUriPrefix, ResourceUriPrefixLength, resourceId, ResourceUriPrefixLength, TRUE) == CSTR_EQUAL));
+}
+
+HRESULT GetRootAndRelativeIdFromUri(
+    _In_ PCWSTR resourceUri,
+    _Out_writes_z_(rootSize) PWSTR root,
+    unsigned int rootSize,
+    _Out_ unsigned int* indexOfRelatvieId)
+{
+    // We should have checked resourceUri length is longer than ResourceUriPrefixLength before calling this
```
robertzhou-wpn, line 48: Wondering where we do this check?

huichen123: The function is always called after IsResourceUri. The check is done in IsResourceUri.

PR #3076, opened 2022-10-24
Remove VersionInfo in 1.2 branch

MikeHillberg:
We want to make the VersionInfo APIs experimental in main then cherry pick that to 1.2-stable, where the APIs will be stripped out since it's a stable branch. But we're having trouble with the experimental mechanism, so while we work that out in main, for now just delete from the 1.2-stable branch.


DrusTheAxe (approved):
#sadness

Issue #2634, opened 2022-06-16
Unpackaged framework-dependent app crashes if fusion manifest not present/embedded


riverar:
### Describe the bug

In scenarios where the fusion manifest is not embedded into the output executable, an app bootstrapping with Windows App SDK will fail to start.

Repro project: [repro.zip](https://github.com/microsoft/WindowsAppSDK/files/8916829/repro.zip)

### Steps to reproduce the bug

1. Open repro project provided above
2. Compile/execute output
3. Observe app does not start correctly
4. Edit the .vcxproj and replace all instances of `<EmbedManifest>false</EmbedManifest>` with `<EmbedManifest>true</EmbedManifest>`.
5. Compile/execute output
6. Observe app starts correctly

### Expected behavior

App to start correctly regardless of fusion manifest presence.

### NuGet package version

1.1.0-preview3

### Packaging type

Unpackaged

### Windows version

Insider Build (xxxxx)

### IDE

Visual Studio 2022

### Additional context

Impacted Rust bindings, which does not use fusion manifests.

It appears the bootstrapping component is attempting to locate reg-free WinRT information. In *non-framework-dependent* apps, that is expected. In *framework-dependent* apps it is not. The fusion manifest in these scenarios will have very little aside from optional common controls/DPI flags.

<details>
<summary>Output window</summary>

```
'repro.exe' (Win32): Loaded 'C:\Sources\was-repros\redacted\x64\Debug\repro\repro.exe'. Symbols loaded.
'repro.exe' (Win32): Loaded 'C:\Windows\System32\ntdll.dll'. Symbols loaded.
'repro.exe' (Win32): Loaded 'C:\Windows\System32\kernel32.dll'. Symbols loaded.
'repro.exe' (Win32): Loaded 'C:\Windows\System32\KernelBase.dll'. Symbols loaded.
'repro.exe' (Win32): Loaded 'C:\Windows\System32\combase.dll'. Symbol loading disabled by Include/Exclude setting.
'repro.exe' (Win32): Loaded 'C:\Windows\System32\ucrtbase.dll'. Symbols loaded.
'repro.exe' (Win32): Loaded 'C:\Windows\System32\vcruntime140d.dll'. Symbols loaded.
'repro.exe' (Win32): Loaded 'C:\Windows\System32\rpcrt4.dll'. Symbols loaded.
'repro.exe' (Win32): Loaded 'C:\Windows\System32\msvcp140d.dll'. Symbols loaded.
[...]

PR #3093, opened 2022-11-01
Microsoft.WindowsAppRuntime.dll!DllMain fails if exe has resources but no SxS manifest

DrusTheAxe:
Microsoft.WindowsAppRuntime.dll fails in DllMain if loaded by a process whose exe has resources but not an embedded SxS fusion manifest. Those are distinct error codes from `CreateActCtxW()` but we only trapped for one. Updated to trap for both conditions.

https://task.ms/41853965


dmachaj (approved):
I built your branch locally and confirmed that Microsoft.WindowsAppRuntime.dll now passes the check (the crash is gone)

👍

riverar:
Might want to check if this affects #2634

PR #3118, opened 2022-11-07
Introducing WIL activities in the App Notification Builder

danielayala94:
This is the first round of telemetry changes to introduce WIL activities in place of the single telemetry events. This PR covers all changes required for the App Notification builder.

In subsequent PRs I will do the cleaning work needed on NotificationTelemetryHelper.h.

**How verified:** Recorded AppNotificationBuilderTelemetry traces on an etl file and I was able to see the new activities there.

### DIFF
--- a/dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
+++ b/dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
@@ -395,6 +395,8 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
         appNotification.Tag(m_tag);
         appNotification.Group(m_group);
 
+        logTelemetry.Stop();
+
         return appNotification;
     }
 }
--- a/dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
+++ b/dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 #include "pch.h"
+#include "NotificationTelemetryHelper.h"
 #include "AppNotificationBuilder.h"
 #include "AppNotificationBuilderTelemetry.h"
 #include <winrt/Windows.Globalization.h>
@@ -84,7 +85,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
             m_scenario = AppNotificationScenario::IncomingCall;
         }
         return *this;
-    }    
+    }
 
     winrt::Microsoft::Windows::AppNotifications::Builder::AppNotificationBuilder AppNotificationBuilder::SetAttributionText(hstring const& text)
     {
@@ -124,7 +125,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
     {
         THROW_HR_IF_MSG(E_INVALIDARG, alternateText.empty(), "You must provide an alternate text string calling SetInlineImage");
 
-        std::wstring hintCrop { imageCrop == AppNotificationImageCrop::Circle ? L" hint-crop='circle'" : L"" };
+        std::wstring hintCrop{ imageCrop == AppNotificationImageCrop::Circle ? L" hint-crop='circle'" : L"" };
         m_inlineImage = wil::str_printf<std::wstring>(L"<image src='%ls' alt='%ls'%ls/>", imageUri.ToString().c_str(), EncodeXml(alternateText).c_str(), hintCrop.c_str());
 
         return *this;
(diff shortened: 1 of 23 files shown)


dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.cpp
```diff
@@ -370,42 +371,30 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
 
     winrt::Microsoft::Windows::AppNotifications::AppNotification AppNotificationBuilder::BuildNotification()
     {
-        HRESULT hr{ S_OK };
+        auto logTelemetry = AppNotificationBuilderTelemetry::BuildNotification::Start(m_telemetryHelper);
```
pmpurifoy, line 374: {}-init applies throughout :)

dev/AppNotifications/AppNotificationBuilder/AppNotificationButton.h
```diff
@@ -73,6 +73,8 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
         winrt::hstring m_inputId{};
         bool m_useContextMenuPlacement{};
         AppNotificationButtonStyle m_buttonStyle { AppNotificationButtonStyle::Default };
+
+        NotificationTelemetryHelper m_telemetryHelper{};
```
hulumane, line 77: Can we not use the telemetry helper class from NotificationBuilder rather than declare and reinit again?

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilderTelemetry.h
```diff
@@ -16,86 +16,59 @@ class AppNotificationBuilderTelemetry : public wil::TraceLoggingProvider
...
-                TelemetryPrivacyDataTag(PDT_ProductAndServicePerformance),
-                _GENERIC_PARTB_FIELDS_ENABLED,
-                TraceLoggingHexUInt32(hr, "OperationResult"),
-                TraceLoggingBool(m_telemetryHelper.IsPackagedApp(), "IsAppPackaged"),
-                TraceLoggingWideString(m_telemetryHelper.GetAppName().c_str(), "AppName"));
-        }
+        TraceLoggingClassWriteStart(
+            TextPropertiesToString,
+            _GENERIC_PARTB_FIELDS_ENABLED,
+            TraceLoggingBool(notificationTelemetryHelper.IsPackagedApp(), "IsAppPackaged"),
```
hulumane, line 70: I hope these values are cached and not calling into state repo every single time. It would significantly impact perf.

dev/AppNotifications/AppNotificationBuilder/AppNotificationComboBox.h
```diff
@@ -37,6 +37,8 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
         winrt::hstring m_selectedItem{};
 
         std::wstring GetSelectionItems();
+
+        NotificationTelemetryHelper m_telemetryHelper{};
```
hulumane, line 41: SAme concern here. Can we not use the instance from NotificationBuilder here?

dev/AppNotifications/AppNotificationBuilder/AppNotificationProgressBar.h
```diff
@@ -51,6 +51,8 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
         double m_value;
         BindMode m_valueStringOverrideBindMode;
         winrt::hstring m_valueStringOverride;
+
+        NotificationTelemetryHelper m_telemetryHelper{};
```
hulumane, line 55: Same concern as above.

dev/AppNotifications/AppNotificationBuilder/AppNotificationTextProperties.h
```diff
@@ -34,6 +34,8 @@ namespace winrt::Microsoft::Windows::AppNotifications::Builder::implementation
         int m_maxLines{ 0 };
         winrt::hstring m_language{};
         bool m_useCallScenarioAlign{};
+
+        NotificationTelemetryHelper m_telemetryHelper{};
```
hulumane, line 38: Same concern. 

In essence, we should figure out a way to use a global rather than re-initializing this over and over again.

Also the implementation of helper should cache the values of AppId and IsPackaged to avoid expensive RPC calls to state repo.

dev/Common/NotificationTelemetryHelper.h
```diff
@@ -6,23 +6,25 @@
...
+        m_appName = m_isPackagedApp ? GetAppNamePackaged() : GetAppNameUnpackaged();
+    }
+
     inline bool ShouldLogEvent()
     {
         return c_maxEventLimit >= UpdateLogEventCount();
     }
 
-    inline bool IsPackagedApp() const
+    inline const bool IsPackagedApp() const
```
loneursid, line 20: Why return a const bool and not just a bool> It's not like your returning a ref or a pointer here.

danielayala94: Thank you for the advice! I realized that setting const to a thing returned by value isn't really needed.

hulumane (approved):

loneursid (approved):

PR #3127, opened 2022-11-10
Introducing WIL activities for AppNotifications

danielayala94:
Now it's time to bring WIL activities goodness to App Notifications.

**How verified:** While running the AppNotificationTests suite, I recorded AppNotificationTelemetry traces in an etl file and I was able to see the new activities there.

**Bonus:** I fixed a couple of minor test bugs on AppNotificationTests.

### DIFF
--- a/dev/AppNotifications/AppNotificationTelemetry.h
+++ b/dev/AppNotifications/AppNotificationTelemetry.h
@@ -177,7 +177,7 @@ public:
         std::wstring const& appId)
     {
         TraceLoggingClassWriteStart(
-            RemoveAllAsyncActivity,
+            RemoveAllAsync,
             _GENERIC_PARTB_FIELDS_ENABLED,
             TraceLoggingWideString(appId.c_str(), "AppId"),
             TraceLoggingBool(notificationTelemetryHelper.IsPackagedApp(), "IsAppPackaged"),
--- a/dev/AppNotifications/AppNotificationTelemetry.h
+++ b/dev/AppNotifications/AppNotificationTelemetry.h
@@ -11,270 +11,211 @@ DECLARE_TRACELOGGING_CLASS(AppNotificationTelemetryProvider,
     // {1825c850-a487-537d-b768-f0ab298d2565}
     (0x1825c850, 0xa487, 0x537d, 0xb7, 0x68, 0xf0, 0xab, 0x29, 0x8d, 0x25, 0x65));
 
-
 class AppNotificationTelemetry : public wil::TraceLoggingProvider
 {
     IMPLEMENT_TELEMETRY_CLASS(AppNotificationTelemetry, AppNotificationTelemetryProvider);
 
 public:
-    DEFINE_EVENT_METHOD(LogRegister)(
-        winrt::hresult hr,
-        std::wstring const& appId) noexcept try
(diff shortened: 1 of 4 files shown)


test/AppNotificationTests/UnpackagedTests.cpp
```diff
@@ -110,7 +110,7 @@ void UnpackagedTests::VerifyGetAllAsyncWithZeroActiveToast()
 
 void UnpackagedTests::VerifyGetAllAsyncWithOneActiveToast()
 {
-    BaseTestSuite::VerifyGetAllAsyncWithZeroActiveToast();
```
loneursid, line 113: Good catch. Would be great if there was a way to automate all that... maybe there's something taht can be done with templates in the future...

dev/AppNotifications/AppNotificationTelemetry.h
```diff
@@ -11,270 +11,211 @@ DECLARE_TRACELOGGING_CLASS(AppNotificationTelemetryProvider,
...
-                TraceLoggingWideString(m_telemetryHelper.GetAppName().c_str(), "AppName"));
-        }
+        TraceLoggingClassWriteStart(
+            Register,
+            _GENERIC_PARTB_FIELDS_ENABLED,
+            TraceLoggingWideString(appId.c_str(), "AppId"),
+            TraceLoggingBool(notificationTelemetryHelper.IsPackagedApp(), "IsAppPackaged"),
+            TraceLoggingWideString(notificationTelemetryHelper.GetAppName().c_str(), "AppName"));
     }
-    CATCH_LOG()
```
loneursid, line 36: We used to swallow and log any exception thrown by the provider, what happens to them now?

loneursid (commented):
Just a question

test/AppNotificationTests/AppNotifications.Test.h
```diff
@@ -129,8 +129,9 @@ namespace AppNotifications::Test
             break;
 
         case ExpectedTransientProperties::DEFAULT:
+            // The default priority value is Default, and the default for SuppressDisplay is false.
             VERIFY_ARE_EQUAL(winrt::Microsoft::Windows::AppNotifications::AppNotificationPriority::Default, actual.Priority());
-            VERIFY_IS_TRUE(actual.SuppressDisplay());
+            VERIFY_IS_FALSE(actual.SuppressDisplay());
```
hulumane, line 134: Was an existing test failing because of this?

hulumane (approved):

test/AppNotificationTests/AppNotifications.Test.h
danielayala94: No, because of the other test invoking the wrong BaseTestSuite function 😆. Once I fixed the naming thing, this verification step bug showed up.

dev/AppNotifications/AppNotificationTelemetry.h
danielayala94: Then it will throw 😅. I'll fix this in the upcoming PR.

It's worth noticing that we can't really control handling exceptions thrown by Stop() anyway unless wrapping it on a try catch on the caller, which seems overkill to me.

PR #3131, opened 2022-11-11
Introducing WIL activities for Push Notifications

danielayala94:
The third and last part of the _WIL activities into WNP_ trilogy.

This PR replaces Push Notifications telemetry events with activities. It also cleans throttling logic on NotificationTelemetryHelper and adds exception handling on all Start events.

**How verified:** Ran all WNP WASDK unit tests and analyzed an ETL file containing all our activities.


dev/PushNotifications/PushNotificationTelemetry.h
```diff
@@ -17,122 +17,90 @@ class PushNotificationTelemetry : public wil::TraceLoggingProvider
...
+    BEGIN_COMPLIANT_MEASURES_ACTIVITY_CLASS(UnregisterAll, PDT_ProductAndServicePerformance);
+    DEFINE_ACTIVITY_START(NotificationTelemetryHelper& notificationTelemetryHelper) noexcept try
     {
-        if (m_telemetryHelper.ShouldLogEvent())
-        {
-            TraceLoggingClassWriteMeasure(
-                "UnregisterAll",
-                TelemetryPrivacyDataTag(PDT_ProductAndServicePerformance),
-                _GENERIC_PARTB_FIELDS_ENABLED,
-                TraceLoggingHexUInt32(hr, "OperationResult"),
```
loneursid, line 95: Is the activity somehow capturing and logging the hr, or did we just drop that information from the logs?

loneursid (approved):
Thanks for re-adding "noexcept" to the providers.

Have left a question but otherwise looks good.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilderTelemetry.h
```diff
@@ -18,57 +18,62 @@ class AppNotificationBuilderTelemetry : public wil::TraceLoggingProvider
...
     BEGIN_COMPLIANT_MEASURES_ACTIVITY_CLASS(BuildNotification, PDT_ProductAndServicePerformance);
-    DEFINE_ACTIVITY_START(NotificationTelemetryHelper& notificationTelemetryHelper)
-    {
-        TraceLoggingClassWriteStart(
-            BuildNotification,
-            _GENERIC_PARTB_FIELDS_ENABLED,
-            TraceLoggingBool(notificationTelemetryHelper.IsPackagedApp(), "IsAppPackaged"),
-            TraceLoggingWideString(notificationTelemetryHelper.GetAppName().c_str(), "AppName"));
-    }
+        DEFINE_ACTIVITY_START(NotificationTelemetryHelper& notificationTelemetryHelper) noexcept try
```
hulumane, line 21: Did we use try CATCH_LOG pattern in other helpers too?

dev/PushNotifications/PushNotificationChannel.cpp
```diff
@@ -56,6 +55,8 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
 
             hr = channelCloseException.code();
 
+            logTelemetry.StopWithResult(hr);
```
hulumane, line 58: Antipattern. No need of stop. The throw below in the if condition should log the hresult automatically.

hulumane: ERROR_NOT_FOUND is by design if the registration doesn't exist. So its ok to ignore. Please remove the stop event.

danielayala94: Actually ERROR_NOT_FOUND is still being logged because of the THROW macro above. I'm going to remove the StopWithResult().

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilderTelemetry.h
danielayala94: Do you mean if we use try and CATCH_LOG on all providers? If so, then yes, we do.

dev/PushNotifications/PushNotificationTelemetry.h
danielayala94: Yes! Any failure coming from a THROW macro during the lifetime of the activity gets logged.

Calling Stop() means that the operation associated to the activity finished with S_OK. If the function threw before reaching Stop(), then we log the last recorded failure.

hulumane (approved):

PR #3135, opened 2022-11-11
Plumbing CV from the LRP to the SDK

danielayala94:
Making the CV to be logged on the SDK activity, by introducing a new interface - IWpnNotificationSink2.


dev/PushNotifications/PushNotificationsLongRunningTask.ProxyStub/NotificationsLongRunningProcess.idl
```diff
@@ -14,6 +14,18 @@ interface IWpnForegroundSink : IUnknown
...
 
+[object]
+[uuid(PUSHNOTIFICATIONS_FOREGROUNDSINK2_CLSID_UUID)]
+[pointer_default(unique)]
+interface IWpnForegroundSink2 : IUnknown
+{
+    HRESULT InvokeAllWithCorrelationVector(
+        [in] ULONG length,
+        [in, size_is(length)] byte* data,
+        [in] LPCWSTR correlationVector,
```
jonwis, line 25: Is this copied from the OS somewhere? Should this be a WinRT type instead of "flat IDL"?

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
```diff
@@ -36,12 +36,15 @@ HRESULT NotificationListener::RuntimeClassInitialize(
...
+STDMETHODIMP_(HRESULT __stdcall) NotificationListener::OnRawNotificationReceived(
+    unsigned int payloadLength,
+    _In_ byte* payload,
+    _In_ HSTRING correlationVector) noexcept try
 {
     HRESULT hr{ S_OK };
 
     auto logTelemetry{ wil::scope_exit([&]() {
-        PushNotificationLongRunningTaskTelemetry::LogOnRawNotificationReceived(hr, WindowsGetStringRawBuffer(correlationVector, nullptr));
+        PushNotificationLongRunningTaskTelemetry::LogOnRawNotificationReceived(hr, wil::str_raw_ptr(correlationVector));
```
jonwis, line 47: Consider changing the signature here so that the telemetry method takes HSTRING on its own, and it does the conversion inside the method.

dev/PushNotifications/PushNotificationTelemetry.h
```diff
@@ -117,8 +117,9 @@ class PushNotificationTelemetry : public wil::TraceLoggingProvider
     }
     CATCH_LOG()
 
-        DEFINE_EVENT_METHOD(LogInvokeAll)(
-            winrt::hresult hr) noexcept try
+    DEFINE_EVENT_METHOD(LogInvokeAll)(
```
loneursid, line 120: Will this become an activity once the PR to move push notifications from events to activities is merged?

danielayala94: Yes. I think that's one of the reasons this PR is still a draft 😅.

danielayala94: I merged against the latest develop commit, which has the activities I introduced.

dev/PushNotifications/PushNotificationsLongRunningTask.ProxyStub/NotificationsLongRunningProcess.idl
danielayala94: @pmpurifoy do you have context on why the sink interfaces were added here?

loneursid (approved):
Ship it!

dev/PushNotifications/PushNotificationsLongRunningTask/ForegroundSinkManager.cpp
```diff
@@ -17,19 +19,46 @@ void ForegroundSinkManager::Remove(std::wstring const& appId)
...
     auto it = m_foregroundMap.find(appId);
     if (it != m_foregroundMap.end())
     {
+        // IWpnForegroundSink2 declares InvokeAll expecting a correlation vector.
+        // The SDK may not know about this interface, so we make sure
+        // if the foreground sink implements it, then route the payload appropriately.
         BOOL foregroundHandled = true;
-        if (FAILED(it->second->InvokeAll(payloadSize, payload.data(), &foregroundHandled)))
+        auto foregroundSink = it->second;
+        ComPtr<IWpnForegroundSink2> foregroundSink2;
```
pmpurifoy, line 38: It probably is low impact, but can we use winrt::com_ptr here?

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
```diff
@@ -36,39 +36,30 @@ HRESULT NotificationListener::RuntimeClassInitialize(
...
-
-        return hr;
-    }
-    catch (...)
-    {
-        hr = wil::ResultFromCaughtException();
-        throw;
     }
+
+    logTelemetry.Stop();
```
pmpurifoy, line 61: Can we use scope_exit here instead for stop telemetry?

tools/GeneratePushNotificationsOverrides.ps1
```diff
@@ -57,6 +57,11 @@ $content_json=@"
 			"CLSID": {
 				"UUID": "25604D55-9B17-426F-9D67-2B11B3A65598"
 			}
+		},
+        "ForegroundSink2": {
```
pmpurifoy, line 61: Nit: Align with other ComInterfaces

dev/PushNotifications/PushNotificationsLongRunningTask/ForegroundSinkManager.cpp
danielayala94: No, because the foreground sinks are stored as ComPtrs in the list.

danielayala94: Unless we change the list to use winrt, but I think it's out of scope.

dev/PushNotifications/PushNotificationsLongRunningTask/NotificationListener.cpp
danielayala94: Good catch!

PR #3148, opened 2022-11-14
Remove UWP templates

bpulliam:

This change removes the UWP templates that still existed under experimental.
- The first commit removes the templates themselves and the entries in the project and solution files.
- The second removes the strings that were only used in the UWP templates.
- The third contains a set of other fixes to the projects and strings exposed by this change.


dev/VSIX/Extension/Cpp/Common/fr-FR/VSPackage.fr-FR.resx
```diff
@@ -204,21 +204,12 @@
     <value>Modèle de projet pour la création d’une application de bureau basée sur la Bibliothèque d’interface utilisateur Windows (WinUI 3) ainsi qu’un package MSIX pour le chargement indépendant ou la distribution via le Microsoft Store.</value>
   </data>
   <data name="1033" xml:space="preserve">
-    <value>Composant Windows Runtime (WinUI 3 dans UWP)</value>
```
evelynwu-msft, line 207: >dans UWP) [](http://example.com/codeflow?start=46&length=9)

The changes to this file (and all the other non-`en-us` ones) won't be preserved since it comes from the localization service. On the other hand this is a translation error... I'll need to look into how to file a bug against localization so that it gets updated in the next run of the localization pipeline.

evelynwu-msft (approved):
:shipit:

PR #3170, opened 2022-11-22
Fix Various Issues in the BuildTransportPackage.ps1

kythant:
1) Fix Channel variable not following down to Generate-TerminalVelocityFeatures.ps1 correctly

2) Remove unused Build-DevProject.yml that somehow got brought back. 

3) Added clean code

4) Fix Generate Overrides is not being called in the inner dev loop.


alexlamtest (commented):
Non-blocking: Old parameter "LocalPackagesPath" is obsolete? While new parameter "Clean" might be worth a comment, because this script would then perform cleaning and then exit immediately?

alexlamtest (approved):

PR #3171, opened 2022-11-22
Update to C++20

DrusTheAxe:
Update C++ standard in use to C++20 (was C++17)

NOTE: This has uncovered several bugs. Fixing them is in progress.


loneursid:
@DrusTheAxe - Looks like you are targeting main while you should have targeted develop, no? I see a bunch of changes in this PR that do not seem to have anything ot do with updating to C++ 20. For example, some notification specs files are being renamed (moved to different folder, in fact)

DrusTheAxe:
Good catch. Yes, the changes target `develop` it's just the PR was mis-targeted. Fixing

bpulliam (approved):

loneursid:
@DrusTheAxe - re: Switch /await:strict to /await for now until Notifications updates their code to be /await:strict compatible, is there a bug tracking this?

PR #3176, opened 2022-11-23
Adding product version info to more binaries

JJBrychell:
Create a common resource include file for binary versioning information and added rc files (that use that include) for binaries that ship with the SDK.


DrusTheAxe:
Nice but you missed some projects

* dev\DynamicDependencyDataStore\DynamicDependency.DataStore.ProxyStub
* dev\DynamicDependencyLifetimeManager\DynamicDependencyLifetimeManager.ProxyStub

DrusTheAxe:
The PR's description has the generic template. Please update

JJBrychell:
@DrusTheAxe: My task was to add versions to the files that are shipped in the sdk.  Those two projects  don't appear in the installed directory.  Am I missing something?

DrusTheAxe (approved):

fredemmott:
A similar change has been made to WinUIEdit.dll , however as it already included a 10.x version number (from the windows DDK), this breaks MSI installers of self-contained apps  - https://github.com/microsoft/microsoft-ui-xaml/issues/8518

Commenting here as I'm unable to find a commit/PR/issue here for that DLL and it seems related - perhaps a closed-source change within MSFT?

PR #3189, opened 2022-11-28
Update WindowsAppRuntimeInstall.vcxproj

alexlamtest:
Adding the "/Qspectre" flag for building of the WindowsAppRuntimeInstall.exe in the Aggregation pipeline.

### DIFF
--- a/installer/dev/WindowsAppRuntimeInstall.vcxproj
+++ b/installer/dev/WindowsAppRuntimeInstall.vcxproj
@@ -76,7 +76,7 @@
       <PrecompiledHeaderFile>pch.h</PrecompiledHeaderFile>
       <PrecompiledHeaderOutputFile>$(IntDir)pch.pch</PrecompiledHeaderOutputFile>
       <PreprocessorDefinitions>_CONSOLE;WIN32_LEAN_AND_MEAN;WINRT_LEAN_AND_MEAN;%(PreprocessorDefinitions)</PreprocessorDefinitions>
-      <AdditionalOptions>%(AdditionalOptions) /bigobj</AdditionalOptions>
+      <AdditionalOptions>%(AdditionalOptions) /bigobj /Qspectre</AdditionalOptions>
     </ClCompile>
   </ItemDefinitionGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)'=='Debug'">


codendone (approved):

kythant (approved):

DrusTheAxe:
The Installer's built by running msbuild on \installer\WindowsAppRuntimeInstall.sln. That's not picking up \Directory.Build.props because VS stops looking up the directory tree when it hits the .sln's directory? Installer should be picking up the root's Directory.Build.props for multiple reasons (current, plus future too). This can be solved by adding \installer\Directory.Build.props

```
<Project InitialTargets="DirectoryBuildPropsInfo" ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">

  <Import Project="$(RepoRoot)\Directory.Build.props" />

</Project>
```

?

P.S. MRTCore is built from `\dev\MRTcore\mrt\MrtCore.sln`. Same issues and solution -- add `\dev\MRTcore\mrt\Directory.Build.props` import'ing the root copy?

installer/dev/WindowsAppRuntimeInstall.vcxproj
```diff
@@ -76,7 +76,7 @@
       <PrecompiledHeaderFile>pch.h</PrecompiledHeaderFile>
       <PrecompiledHeaderOutputFile>$(IntDir)pch.pch</PrecompiledHeaderOutputFile>
       <PreprocessorDefinitions>_CONSOLE;WIN32_LEAN_AND_MEAN;WINRT_LEAN_AND_MEAN;%(PreprocessorDefinitions)</PreprocessorDefinitions>
-      <AdditionalOptions>%(AdditionalOptions) /bigobj</AdditionalOptions>
+      <AdditionalOptions>%(AdditionalOptions) /bigobj /Qspectre</AdditionalOptions>
```
DrusTheAxe, line 79: Not using the root's Directory.Build.props is bad (for this and other reasons), no? See other comment

P.S. MRTCore has the same issue?

DrusTheAxe (changes requested):

installer/dev/WindowsAppRuntimeInstall.vcxproj
bpulliam: Separate bugs filed for both installer and MRT to use the root's props files.

bpulliam (approved):

installer/dev/WindowsAppRuntimeInstall.vcxproj
DrusTheAxe: Also nuget.config. There's 4 copies in the tree. We should only be using the 1 in the root.

bpulliam:
Given #3357, this should no longer be necessary.

PR #3265, opened 2022-12-20
Remove ToastNotificationTests folder from .sln

pmpurifoy:

Removing this from the main branch since it is causing issues and is unused.


alexlamtest (approved):

loneursid (changes requested):
Looks like this PR is accidentally targeting main instead of develop. Are you sure main is the right target here?

PR #3327, opened 2023-01-12
Update WindowsAppSDK-RunHelixTests-Job.yml

alexlamtest:
Replacing the Helix queues for 21h1 with corresponding new Windows 10 Helix queues, because 21h1 is becoming out of service, so we can stop running tests on it. The new Windows 10 Helix queues will be kept up to date, always pointing to the latest public Windows 10 release.


kythant (approved):

codendone (approved):
:shipit:

PR #3335, opened 2023-01-13
Fix PushNotificationChannel::ExpirationTime incorrectly returning DateTime

pmpurifoy:

This PR fixes the ExpirationTime bug in PushNotificationChannel. The conversion from the ABI to the winrt type did not match and this PR manually does the conversion. This also adds unit a test with the change for verification.


danielayala94 (approved):

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -121,7 +121,14 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
 
         THROW_IF_FAILED(operationalCode);
 
-        winrt::copy_from_abi(channelInfo.channelExpiryTime, &channelExpiryTime);
+        // Need to manually convert from the ABI::DateTime type to the winrt::DateTime type since they have different contracts.
+        FILETIME filetime;
+        LARGE_INTEGER largeTime;
+        largeTime.QuadPart = channelExpiryTime.UniversalTime;
```
hulumane, line 127: Is this in UTC? I believe that expiration time is always UTC, but it would be best to check

pmpurifoy: From [docs](https://learn.microsoft.com/en-us/windows/win32/api/minwinbase/ns-minwinbase-filetime): Contains a 64-bit value representing the number of 100-nanosecond intervals since January 1, 1601 (UTC).

So we should be good.

test/PushNotificationTests/BaseTestSuite.cpp
```diff
@@ -113,6 +114,27 @@ void BaseTestSuite::ChannelRequestUsingRemoteId()
...
+        auto channel{ channelOperation.GetResults().Channel() };
+        auto expirationTime{ channel.ExpirationTime() };
+        auto expiryBound{ winrt::clock::now() + std::chrono::days(31) };
+
+        // Need to add 30 days to match expiration time.
+        VERIFY_IS_LESS_THAN(expirationTime, expiryBound);
+    }
+    else
+    {
+        auto channelOperation{ PushNotificationManager::Default().CreateChannelAsync(c_azureRemoteId) };
```
hulumane, line 133: How was the negative test verified?

test/PushNotificationTests/BaseTestSuite.cpp
```diff
@@ -113,6 +114,27 @@ void BaseTestSuite::ChannelRequestUsingRemoteId()
...
+void BaseTestSuite::ChannelRequestCheckExpirationTime()
+{
+    if (PushNotificationManager::Default().IsSupported())
+    {
+        auto channelOperation{ PushNotificationManager::Default().CreateChannelAsync(c_azureRemoteId) };
+        VERIFY_SUCCEEDED(ChannelRequestHelper(channelOperation));
+
+        auto channel{ channelOperation.GetResults().Channel() };
+        auto expirationTime{ channel.ExpirationTime() };
+        auto expiryBound{ winrt::clock::now() + std::chrono::days(31) };
```
hulumane, line 126: Add a comment here that we want granularity within 31 days because we expect the expiry to be a maximum of 30 days from now

hulumane: If you want better granularity, you can use std::chrono:hours and have the expiration be std::chrono::hours(30*24 + 1). Then the granularity changes to a max of 1 hr delta.

test/PushNotificationTests/BaseTestSuite.cpp
pmpurifoy: In our negative tests, we return E_FAIL. This is not ExpirationTime specific but moreso to cover unsupported WAS scenarios.
https://github.com/microsoft/WindowsAppSDK/blob/77761e244289fda6b3d5f14c7bded189fed4fb89/dev/PushNotifications/PushNotificationManager.cpp#L208

test/PushNotificationTests/BaseTestSuite.cpp
```diff
@@ -113,6 +114,27 @@ void BaseTestSuite::ChannelRequestUsingRemoteId()
...
+
+        // Need to add 30 days to match expiration time.
+        VERIFY_IS_LESS_THAN(expirationTime, expiryBound);
```
hulumane, line 129: Are both of these in UTC? We need to make an apples to apples comparison.

test/PushNotificationTests/BaseTestSuite.cpp
hulumane: I get that but did we run the negative test case too or is this just eye-balling :)

test/PushNotificationTests/BaseTestSuite.cpp
pmpurifoy: Yes. These are both UTC.

hulumane (approved):

test/PushNotificationTests/BaseTestSuite.cpp
pmpurifoy: Just eye-balling the negative test. Since this is a common negative test that we have trusted in the past (downlevel testing), we should be ok.

dev/PushNotifications/PushNotificationManager.cpp
```diff
@@ -121,7 +121,14 @@ namespace winrt::Microsoft::Windows::PushNotifications::implementation
 
         THROW_IF_FAILED(operationalCode);
 
-        winrt::copy_from_abi(channelInfo.channelExpiryTime, &channelExpiryTime);
+        // Need to manually convert from the ABI::DateTime type to the winrt::DateTime type since they have different contracts.
```
robertzhou-wpn, line 124: minor: Maybe we put this time conversion into a util function.

PR #3358, opened 2023-01-24
dev.cpp.*.com-oopserver[-main] + Kozani[Remote]Manager lacking C++/WinRT support

DrusTheAxe:
dev.cpp.*.com-oopserver[-main] templates (and thus Kozani[Remote]Manager) were lacking C++/WinRT support - you could build but not consume WinRT


rhuang-msft (approved):
Thanks! Can't wait to try this out!

sachintaMSFT (approved):

PR #3366, opened 2023-01-26
Eliminate relative package locations for nuget

bpulliam:

A new variable is defined in Directory.Build.props for NugetPackageDirectory that can be referenced in csproj and vcxproj files.  This eliminates the need for "[..\]*" when defining packa directories.  This is much easier to maintain and eliminates the need to change project files if they are moved around the repo.

There were also instances where wil versions were not using the variables, so those were fixed as well.


kythant:
There is already a RepoRoot defined in
https://github.com/microsoft/WindowsAppSDK/blob/226235a1013bc917f22654ace73e4e289c92e6bf/build/EnsureOutputLayout.props

I suggest we consolidate it so we don't have two properties doing the same thing.

DrusTheAxe (approved):
:-)

Issue #3137, opened 2022-11-12
DeploymentAgent.exe crashing when calling DeploymentManager::Initialize() on Windows 10


triplef:
### Describe the bug

Launching a packaged app using the Windows App SDK v1.1 or v1.2 on Windows 10, which calls `DeploymentManager::Initialize()` on launch, results in DeploymentAgent.exe crashing and the following deployment result in the app if the machine does not already have the Main and Singleton packages installed:

- using Windows App SDK v.1.1: `0xc0000409` ([Report-v1.1.wer.txt](https://github.com/microsoft/WindowsAppSDK/files/9994474/Report-v1.1.wer.txt))
- using Windows App SDK v.1.2: `0x80070005` ([Report-v1.2.wer.txt](https://github.com/microsoft/WindowsAppSDK/files/9994475/Report-v1.2.wer.txt))

The issue is not reproducible when launching the app on Windows 11.

The app launches correctly if the Windows App SDK from [here](https://learn.microsoft.com/en-us/windows/apps/windows-app-sdk/downloads) is installed before.

Additional information in discussion https://github.com/microsoft/WindowsAppSDK/discussions/3125.

### Steps to reproduce the bug

1. Create an app with the Windows App SDK calling the DeploymentManager like this from its main() function on launch:
```c++
DeploymentInitializeOptions deploymentOptions;
auto deploymentResult{ DeploymentManager::Initialize(deploymentOptions) };
if (deploymentResult.Status() != DeploymentStatus::Ok) {
	const HRESULT hr{ deploymentResult.ExtendedError().value };
	wprintf(L"Error 0x%X initializing Windows App SDK\n", hr);
	return hr;
}
```
2. Add the following dependency in its AppxManifest:
```xml
<PackageDependency Name="Microsoft.WindowsAppRuntime.1.2" MinVersion="2000.677.1750.0" Publisher="CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US" />
```
3. Create an MSIX from the app.
4. Install the MSIX on a machine running Windows 10 which has no Windows App SDK packages installed (run `get-appxpackage micro*win*appruntime* | remove-appxpackage` before).

### Expected behavior

1. The `Microsoft.WindowsAppRuntime.1.2` package is installed during installation of the MSIX ✅
2. The additional Main and Singleton packages are installed when the app is launched ❌
[...]

PR #3367, opened 2023-01-27
Support TAEF testing in foundation pipeline

pmpurifoy:

Components that have test dependencies on the test packages in the WinAppSDK are now able to automate their unit tests in the Foundation pipeline! By adding a .testdef that follows the following schema:
```
{
    "Tests": [
        {
            "Description": "This module tests the push notifications component in WinAppSDK.",
            "Filename" (required): "PushNotificationTests.dll",
            "Parameters": "",
            "Architectures" (required): ["x86", "x64", "arm64"],
            "Status": "Enabled" 
        }
    ]
}
```

The pipeline with search for these testmds and run tests based on your configurations such as architecture, TAEF parameters, and status. Add this testmd to your project and publish it to your buildoutput. Sample is below.
`    
    <Copy SourceFiles="PushNotificationTests.testdef" DestinationFolder="$(OutDir)" />
  </Target>
`


build/AzurePipelinesTemplates/WindowsAppSDK-RunTests-Steps.yml
```diff
@@ -0,0 +1,115 @@
...
+    inputs:
+      artifactName: 'TestCert'
+      downloadPath: $(Build.SourcesDirectory)\BuildOutput
+
+  - task: powerShell@2
+    displayName: 'Enable developer mode'
+    inputs:
+      targetType: 'inline'
+      script: |
+            $RegistryKeyPath = "HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\AppModelUnlock"
```
DrusTheAxe, line 36: What's this needed for? Just curious

build/AzurePipelinesTemplates/WindowsAppSDK-RunTestsInPipeline-Job.yml
```diff
@@ -0,0 +1,120 @@
...
+  
+jobs:
+- job: ${{ parameters.jobName }}
+  dependsOn: ${{ parameters.dependsOn }}
+  strategy:
+    # Debug/Release buildConfiguration below refers to the build of the tests/apps. The OS image bits are always Release build.
+    matrix:
+      # The test matrix for x64 is currently larger; this is where we exercise most variations.
+      ${{ if not( parameters.isArm64Platfrom ) }}:
+        RS5_x86chk:
```
DrusTheAxe, line 26: We use chk builds? I thought those went the way of the dodo.

build/scripts/DownloadDotNetRuntimeInstaller.ps1
```diff
@@ -0,0 +1,43 @@
+[CmdLetBinding()]
+Param(
+    [string]$Platform = "x64",
+    [string]$OutputDirectory = ""
```
DrusTheAxe, line 4: Why default it? Why not

```
[string]$OutputDirectory
```

Won't `Join-Path $null "foo"` work equally well as `Join-Path "" "foo"` ?

build/scripts/DownloadDotNetRuntimeInstaller.ps1
```diff
@@ -0,0 +1,43 @@
...
+)
+
+$ProgressPreference = "SilentlyContinue"
+$ErrorActionPreference = "Stop"
+
+$outputPath = Join-Path $OutputDirectory "dotnet-windowsdesktop-runtime-installer.exe"
+if(!(Test-Path $OutputDirectory))
+{
+    New-Item -ItemType Directory -Path $OutputDirectory | Out-Null
```
DrusTheAxe, line 13: Which is recommended practice?

```
New-Item BlahBlah | Out-Null
```

or

```
$null = New-Item BlahBlah
```

build/scripts/DownloadDotNetRuntimeInstaller.ps1
```diff
@@ -0,0 +1,43 @@
+[CmdLetBinding()]
+Param(
+    [string]$Platform = "x64",
```
DrusTheAxe, line 3: Why default `$Platform`? Why not always require it (no default)?

This comment applies throughout

DrusTheAxe, line 3: Is this an open ended string or a limited allowed list? Seems like the latter, in which case shouldn't it be an enum (well, Powershell's expression of an enum)?

build/scripts/DownloadVCLibsDesktop.ps1
```diff
@@ -0,0 +1,27 @@
+[CmdLetBinding()]
+Param(
+    [string]$Platform = "x64",
+    [string]$OutputDirectory = "D:\WindowsAppSDK\build\scripts"
+)
+
+$ProgressPreference = "SilentlyContinue"
+$ErrorActionPreference = "Stop"
```
DrusTheAxe, line 8: Very nice, but you should also specify

```
Set-StrictMode -Version 3.0
```

I added that to DevCheck.ps1 recently and it spotted a problem, and stopped me from creating a couple new ones. Plus it's good practice so there's that too :P

This comment applies throughout

build/scripts/DownloadVCLibsDesktop.ps1
```diff
@@ -0,0 +1,27 @@
...
+
+$outputPath = Join-Path $OutputDirectory "Microsoft.VCLibs.$Platform.14.00.Desktop.appx"
+if(!(Test-Path $OutputDirectory))
```
DrusTheAxe, line 11: Should this be

```
if (!(TestPath foo))
```

or

```
if (-not(TestPath foo))
```

Which is recommended practice? I ask as I've usually seen `-not` but I don't know which is best practice

build/scripts/DownloadVCLibsDesktop.ps1
```diff
@@ -0,0 +1,27 @@
...
+
+if(!(Test-Path $outputPath))
+{
+    Write-Host "Downloading $downloadurl to $outputPath"
+    Invoke-WebRequest $downloadurl -OutFile $outputPath
+}
+else
+{
+    Write-Host "The file '$outputPath' already exists. Skipping download."
+}
```
DrusTheAxe, line 27: NIT: Some tools (like GitHub diff) are cranky if the file doesn't end with the newline character

build/AzurePipelinesTemplates/WindowsAppSDK-RunTests-Steps.yml
pmpurifoy: This is used for putting the machine in Developer mode.

TestAll.ps1
```diff
@@ -0,0 +1,110 @@
+#Requires -RunAsAdministrator
```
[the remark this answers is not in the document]
pmpurifoy: Originally this script was running DevCheck which has to be in Admin mode, but since this is only triggering the tests now, we can remove.

BUT it is a directive that only allows the script to run as admin.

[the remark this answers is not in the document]
pmpurifoy: Dev Inner Loop / Test pipeline

TestAll.ps1
```diff
@@ -0,0 +1,110 @@
...
+      Builds the WindowsAppSDK aggregate NuGet package.
+
+Switches:
+
+  -Platform <platform>
+      Only test the selected platform.
+      Example: -Platform x64
+
+  -Configuration <configuration>
+      Only build the selected configuration(s).
```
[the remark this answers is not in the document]
pmpurifoy: Since this script is being used by both the build pipeline and local, I think both parameters should support lists. 

Platforms: "x64,x86"
Configuration:"release,debug"

But the pipeline will only submit the pair, not a list: x64 | release.

TestAll.ps1
```diff
@@ -0,0 +1,110 @@
...
+      Example: -Configuration debug
+      Example: -Configuration "debug,release"
+
+  -AzureBuildStep <step>
+      Used by Azure to selectively run parts of the script in order to account for yaml-specific actions like signing.
+      Example: -AzureBuildStep "CopyFilesFromTransportPackages"
+
+  -ImageName <name>
+      Used by Azure to specify which SKU is being tested.
+      Example: -ImageName "Windows.10.Enterprise.RS5"
```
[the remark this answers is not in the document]
pmpurifoy: Also I just realized that we don't need the SKU parameter.... Removing.

TestAll.ps1
```diff
@@ -0,0 +1,110 @@
...
+  -Help
+      Display this usage message.
+"@
+  Exit
+}
+$lastexitcode = 0
+
+if (($AzureBuildStep -eq "all") -Or ($AzureBuildStep -eq "DisplayInfo")) {
+    # Display OS build/version info
+    Get-Item -Path 'HKLM:\Software\Microsoft\Windows NT\CurrentVersion'
```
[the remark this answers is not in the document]
pmpurifoy: We don't really need this at all. Removing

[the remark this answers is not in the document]
pmpurifoy: We already know the OS build/version as local dev / test machine. Removing.

TestAll.ps1
```diff
@@ -0,0 +1,110 @@
...
+    }
+}
+
+$TotalTime = (Get-Date)-$StartTime
+$TotalMinutes = $TotalTime.Minutes
+$TotalSeconds = $TotalTime.Seconds
+Write-Host @"
+
+Total Running Time:
+$TotalMinutes minutes and $TotalSeconds seconds
```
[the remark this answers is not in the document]
pmpurifoy: Moving to one line.

build/AzurePipelinesTemplates/WindowsAppSDK-RunTestsInPipeline-Job.yml
pmpurifoy: This is the same list as machines from Aggregator.

Are chk builds valid? @alexlamtest / @kythant

build/scripts/DownloadDotNetRuntimeInstaller.ps1
pmpurifoy: Just found out that Join-Path doesn't work with null OR empty string. Making these params required.

build/AzurePipelinesTemplates/WindowsAppSDK-RunTests-Steps.yml
```diff
@@ -0,0 +1,115 @@
+parameters:
+  buildPlatform: ''
+  buildConfiguration: ''
+  ImageName: ''
+  TaefSelect: '*'
+  BinaryCompatSwitch: ''
+  testLocale: ''
+steps:
+  # Download the BuildOutput from the Build stage(s).
```
kythant, line 9: Do we want to run TestAll.ps1 -AzureBuildStep DisplayInfo here?

build/ProjectReunion-BuildFoundation.yml
```diff
@@ -148,6 +148,39 @@ jobs:
       PathtoPublish: '$(build.SourcesDirectory)\BuildOutput'
       artifactName: 'BuildOutput'
 
+  - task: PowerShell@2
```
kythant, line 151: I think this should maybe go into
AzurePipelinesTemplates\WindowsAppSDK-SetupBuildEnvironment-Steps.yml

build/ProjectReunion-BuildFoundation.yml
```diff
@@ -263,6 +296,22 @@ jobs:
...
     condition: eq(variables['Build.SourceBranch'], 'refs/heads/develop')
 
+# Arm64 uses a different VM type than x64, thus requiring them to be in different
+# pools, hence, we spilt the two scenarios into separate jobs.
+- template: AzurePipelinesTemplates\WindowsAppSDK-RunTestsInPipeline-Job.yml
+  parameters:
+    jobName: PipelineTestsX64
+    isArm64Platfrom: false
+    dependsOn:
+      - Build
```
kythant, line 306: Should depend on all of these job below
      - Build
      - BuildAnyCPU
      - BuildAndTestMRT

build/ProjectReunion-BuildFoundation.yml
```diff
@@ -263,6 +296,22 @@ jobs:
...
+
+- template: AzurePipelinesTemplates\WindowsAppSDK-RunTestsInPipeline-Job.yml
+  parameters:
+    jobName: PipelineTestsArm64
+    isArm64Platfrom: true
+    dependsOn:
+      - Build
```
kythant, line 313: Should depend on all of these job below
      - Build
      - BuildAnyCPU
      - BuildAndTestMRT

build/scripts/DownloadDotNetRuntimeInstaller.ps1
pmpurifoy: Yeah, I'm updating this to mandatory

pmpurifoy: Aggregator takes a string separated by commas. "x64,x86"

pmpurifoy: But for this installer, it is only invoked for one platform at a time.

build/scripts/DownloadVCLibsDesktop.ps1
pmpurifoy: I'm pretty -not is the recommended practice. Fixing.

build/ProjectReunion-BuildFoundation.yml
pmpurifoy: I was considering each of the jobs and found that the only job testing has a dependency on is the Build job.

kythant: But if in the future, we want to test MRT core in these pipeline tests, it would need to depend on BuildAndTestMRT.

build/ProjectReunion-BuildFoundation.yml
pmpurifoy: I was considering each of the jobs and found that the only job testing has a dependency on is the Build job.

build/ProjectReunion-BuildFoundation.yml
pmpurifoy: That's a good point. This PR should prep for MRT testing as well.

kythant: Ideally all those jobs will be in a staged called "Build" and a "Test" stage would depend on the "Build"stage.
It would require some refactoring to move all these jobs into stages.

Stretch goal for you :)

TestAll.ps1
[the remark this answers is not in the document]
pmpurifoy: Looks like dism needs elevation.

build/AzurePipelinesTemplates/WindowsAppSDK-RunTests-Steps.yml
pmpurifoy: Do we need DisplayInfo?

pmpurifoy: Or can I pull this out of TestAll and put it in RunTests-Steps.yml instead?

kythant: The DisplayInfo code? 
I'll leave it up to you.

But I am noticing that we don't really have a need for "AzureBuildStep" in TestAll.ps1.
We introduced this parameter in the aggregator's code because we needed to run Azure tasks in between the code in  BuildAll or TestAll. 

Also one difference between the testall code in the aggregator and testall code in Foundation is that the test code gets built in Foundation's buildall code whereas the testcode in the aggregator get's built in the aggregator's testall code.

So maybe we should move building the foundation's test code from buildall to testall.

pmpurifoy: "But I am noticing that we don't really have a need for "AzureBuildStep" in TestAll.ps1." +1

This is because there as two dev/test slns correct? I think we can move it out and let TestAll be pure testing for now.

TestAll.ps1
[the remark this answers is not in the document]
pmpurifoy: Deleting this and moving DisplayInfo to RunTestSteps.yml

build/AzurePipelinesTemplates/WindowsAppSDK-RunTests-Steps.yml
kythant: Ah yes. Don't need to be in this PR for sure. It will be for future refactors.

TestAll.ps1
```diff
@@ -0,0 +1,110 @@
...
+    [string]$Platform = "x64",
+    [string]$Configuration = "release",
+    [string]$ImageName = "",
+    [switch]$AzurePipelineBuild = $false,
+    [switch]$Help = $false
+)
+
+$StartTime = Get-Date
+
+if ($help) {
```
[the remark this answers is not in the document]
pmpurifoy: Changing vars to all camel-case in script params.

TestAll.ps1
```diff
@@ -0,0 +1,110 @@
+#Requires -RunAsAdministrator
+
+Param(
+    [string]$OutputFolder = "",
+    [string]$AzureBuildStep = "all",
+    [string]$Platform = "x64",
+    [string]$Configuration = "release",
+    [string]$ImageName = "",
```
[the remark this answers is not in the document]
pmpurifoy: All params are now required.

build/AzurePipelinesTemplates/WindowsAppSDK-RunTests-Steps.yml
DrusTheAxe: Talk to @EricJohnson327

I'm fond of "build=build and test=test". If we need a "do everything" command then it should be something else

```
magic.ps1 -build ... ==> & build.cmd ...
magic.ps1 -test ... ==> & test.cmd ...
etc
```

DrusTheAxe: >move building the foundation's test code from buildall to testall

I'm fond of "build=build and test=test". If we need a "do everything" command then it should be something else

```
magic.ps1 -build ... ==> & build.cmd ...
magic.ps1 -test ... ==> & test.cmd ...
etc
```

build/scripts/DownloadDotNetRuntimeInstaller.ps1
DrusTheAxe: > Just found out that Join-Path doesn't work with null OR empty string

Sure does

`$meh = Join-Path "$foo" "$bar"`

Then it doesn't matter if $foo or $bar are $null or "" :-)

build/AzurePipelinesTemplates/WindowsAppSDK-RunTestsInPipeline-Job.yml
DrusTheAxe: > This is the same list as machines from Aggregator.

Can we use a shared list across repos?

Should we?

If we can/should we add eng\TestEnvironments.xml with the various definitions, let Maestro flow that across the repos and make ADO use that list?

build/scripts/DownloadDotNetRuntimeInstaller.ps1
pmpurifoy: Didn't know this but looks like $null is the better practice. Replacing.

TestAll.ps1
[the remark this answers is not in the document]
DrusTheAxe: DevCheck doesn't require admin
_Some operations_ in DevCheck require admin

Adding+Removing certs and registering/starting/stopping services

Search for "Get-IsAdmin" in DevCheck.ps1 for what needs admin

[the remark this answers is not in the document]
pmpurifoy: Since TestAll isn't doing any of those operations, they're all preinstalled for us, we can remove admin requirement.

build/scripts/DownloadDotNetRuntimeInstaller.ps1
pmpurifoy: Looks like it's better to do $null

TestAll.ps1
```diff
@@ -0,0 +1,76 @@
+<#  
+.SYNOPSIS
+    Run the TAEF dlls generated by WinAppSDK
+
+.DESCRIPTION
+    The TestAll script will take the folder input and look for subfolders containing a .testmd file. WinAppSDK
```
DrusTheAxe, line 6: RECOMMENDATION: Pick a unique extension

This is similar-but-different from `*.testmd` Windows has (for some time). Please pick a different (unique) extension to avoid confusion.

SUGGESTIONS: `.testdef` (TEST DEFinition) but yeah, my namingFu feels underpowered atm so suggestions welcome

alexlamtest: +1 for using a different extension to avoid confusion with the existing testmd, ".testConfig" and ".testProp" are also options.

build/AzurePipelinesTemplates/WindowsAppSDK-RunTests-Steps.yml
```diff
@@ -0,0 +1,141 @@
...
+
+            Write-Host "Get-WinSystemLocale"
+            Get-WinSystemLocale
+  
+  - task: PowerShell@2
+    displayName: 'Run TAEF Tests'
+    inputs:
+      filePath: 'TestAll.ps1'
+      arguments: >
+        -OutputFolder "$(Build.SourcesDirectory)\BuildOutput\BuildOutput"
```
alexlamtest, line 138: Are we stuck with "...\BuildOutput\BuildOutput", or can we use a more meaning name like "...\BuildOutput\TestResults" or something.

build/scripts/DownloadDotNetRuntimeInstaller.ps1
```diff
@@ -0,0 +1,45 @@
...
+{
+    $downloadurl = $downloadurlx64
+}
+else
+{
+    Write-Host "dotnet Desktop runtime not available for arch $Platform"
+    return
+}
+
+
```
alexlamtest, line 36: Nit: double blank lines?

TestAll.ps1
```diff
@@ -0,0 +1,76 @@
...
+    components define a testmd with the following schema and runs the TAEF dll in the subfolder.
+    
+    Example:
+    {
+        "Tests": [
+            {
+                "Description": "Enabled: This module tests the push notifications component in WinAppSDK.",
+                "Filename": "PushNotificationTests.dll",
```
alexlamtest, line 14: In the PR description you said filename and architecture are mandatory, consider mentioning that here too.

build/AzurePipelinesTemplates/WindowsAppSDK-RunTestsInPipeline-Job.yml
DrusTheAxe: The 'chk' refers to the **test binaries** and not the OS? That's a very different matter - and confusingAH (Windows builds come in fre + chk flavors, WinAppSDK binaries come in release + debug flavors)

I like the idea of running some debug test binaries, but not at the cost of not running release bits on that SKU. We ship release bits. We should be testing release bits on all supported platforms. If we ALSO want to run debug binaries on some SKUs I could agree with that. But if I'm reading this right we're running debug bits on RS5 and _not_ testing our release bits on RS5. Not cool.

TestAll.ps1
pmpurifoy: Using testdef

TestAll.ps1
pmpurifoy: I don't think I'm allowed to add a comment to JSON. Maybe I can add a sample testdef file to the repo for components to copy/paste?

build/AzurePipelinesTemplates/WindowsAppSDK-RunTests-Steps.yml
pmpurifoy: We are currently stuck with BuildOutput/BuildOutput until we can split the BuildOutput into dev/test folders.

TestAll.ps1
DrusTheAxe: > I don't think I'm allowed to add a comment to JSON. Maybe I can add a sample testdef file to the repo for components to copy/paste?

Correct, [JSON](https://www.json.org/json-en.html) has no 'comment' concept. That's what Description would do

Also nothing stops you from adding other data if the parser is forgiving (not strict) e.g.

```
{
    "Tests": [
        {
            "Description": "Enabled: This module tests the push notifications component in WinAppSDK.",
            "Filename": "PushNotificationTests.dll",
            "Parameters": "",
            "Architectures": ["x86", "x64", "arm64"],
            "Status": "Enabled",
            "Warning": "Danger Will Robinson! Danger!",
            "Room Service": "Waffles, Omelette and Pancakes available. Call 1-800-HUNG-RY1"
        }
    ]
}
```

>Maybe I can add a sample testdef file to the repo for components to copy/paste?

No need because you're going to add a _working real-world_ `Test.testdef`(?) to `tools\ProjectTemplates\test.cpp.dll.taef\` right? This PR or follow on, same diff

```
{
    "Tests": [
        {
            "Description": "PurojekutoTenpuret tests for feature Tokucho",
            "Filename": "PurojekutoTenpuret.dll",
            "Parameters": "",
            "Architectures": ["x86", "x64", "arm64"],
            "Status": "Enabled",
        }
    ]
}
```

test/PushNotificationTests/PushNotificationTests.testdef
```diff
@@ -0,0 +1,11 @@
+{
+    "Tests": [
+        {
+            "Description": "Enabled: This module tests the push notifications component in WinAppSDK.",
```
DrusTheAxe, line 4: Why is 'Enabled' in the description?

Does `TestAll` have a `-List` command to list the tests and their status?

If not when are you adding it? :P

See `List-ProjectTemplates` in `tools\AddProject.ps1`

DrusTheAxe (changes requested):

test/PushNotificationTests/PushNotificationTests.testdef
pmpurifoy: Oh, the enabled was an old comment. Will remove.

build/AzurePipelinesTemplates/WindowsAppSDK-RunTestsInPipeline-Job.yml
pmpurifoy: Renamed to FRE

TestAll.ps1
pmpurifoy: Done.

test/PushNotificationTests/PushNotificationTests.testdef
pmpurifoy: Added -List feature :)

TestAll.ps1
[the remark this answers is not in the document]
pmpurifoy: For v1 I don't think we have to support lists. Will follow up with this as a feature.

build/AzurePipelinesTemplates/WindowsAppSDK-RunTestsInPipeline-Job.yml
```diff
@@ -0,0 +1,120 @@
...
+  dependsOn: ${{ parameters.dependsOn }}
+  strategy:
+    # Debug/Release buildConfiguration below refers to the build of the tests/apps. The OS image bits are always Release build.
+    matrix:
+      # The test matrix for x64 is currently larger; this is where we exercise most variations.
+      ${{ if not( parameters.isArm64Platfrom ) }}:
+        RS5_x86fre:
+          imageName: Windows.10.Enterprise.RS5
+          buildPlatform: x86
+          buildConfiguration: release
```
alexlamtest, line 29: Is there any reason why only "release" config is being tested?
Would it be feasible at this point to exercise the "debug" config for a few OS versions? And similar for the smaller arm64 group?

build/AzurePipelinesTemplates/WindowsAppSDK-RunTests-Steps.yml
```diff
@@ -0,0 +1,143 @@
...
+            Write-Host "Get-WinUILanguageOverride"
+            Get-WinUILanguageOverride
+
+            Write-Host "Get-WinSystemLocale"
+            Get-WinSystemLocale
+  
+  - task: PowerShell@2
+    displayName: 'Run TAEF Tests'
+    inputs:
+      filePath: 'TestAll.ps1'
```
alexlamtest, line 136: This script today only detects/runs PushNotification's tests, right?
Are you aware of the status of any other existing tests, in terms whether they will pass/fail when being run in pipeline?

test/PushNotificationTests/PushNotificationTests.testdef
```diff
@@ -0,0 +1,18 @@
+{
+    "Tests": [
+        {
+            "Description": "Test unpackaged scenarios for PushNotifications",
+            "Filename": "PushNotificationTests.dll",
```
alexlamtest, line 5: 1) I'm curious whether your new tests need to be run elevated or not?
2) It might help the reader if the owner of a test can call out anything special about a particular test right here in this .testdef file, say, the test needs to run elevated, or in a low-IL container, something like that, in an optional Notes property, or the alike. Consider starting this practice from the beginning.

TestAll.ps1
```diff
@@ -0,0 +1,132 @@
+<#  
+.SYNOPSIS
+    Run the TAEF dlls generated by WinAppSDK
+
+.DESCRIPTION
+    The TestAll script will take the folder input and look for subfolders containing a .testdef file. WinAppSDK
```
alexlamtest, line 6: It may not be obvious for new developers that they can get their teaf test to run in the pipeline by onboarding them via the .testdef file. I presume the developer can also manually run TestAll.cmd on a local dev machine to get onboarded taef tests to run?
I recommend either update readme.md at the root of this repo, or add a readme.md under the \test folder, to make this test onboarding process more discoverable.

DrusTheAxe: Readme? Too low a bar. This information belongs @ https://github.com/microsoft/WindowsAppSDK/blob/main/docs/Coding-Guidelines.md

1. Add new page https://github.com/microsoft/WindowsAppSDK/blob/main/docs/Coding-Guidelines/Tests.md (sp?)
2. Add new link to that @ https://github.com/microsoft/WindowsAppSDK/blob/main/docs/Coding-Guidelines.md

The docs should explain how to author and use/run *.testdef files

Mention tests can be run via VS' Test Explorer or command line then provide another chapter (or link to another doc) about running tests in the pipeline.

test/PushNotificationTests/PushNotificationTests.testdef
DrusTheAxe: I strongly recommend against #2. Tests can specify RunAs: _and other properties_, and they can change. Code is the final authority; mentioning redundant information in docs is just a matter of time before docs are stale and misleading.

One thing that would be notable is if a test is 'destructive'. As a general rule unit/functional tests in the repo can be run on the local dev desktop. If a test would damage a dev's desktop it could be worth noting in a .testdef is it's not something easily identified by quickly eyeballing code, that `TestAll` doesn't run such tests by default and one could do it manually via `TestAll -Destructive` (hopefully in a VM and not the local desktop).

Are there other subtle test prereqs e.g. doesndidn't Power tests require another machine at some point? Still? If a test required a 2nd machine that could be wroth defining (if it's clear enough to articulate and worth as an option like 'Destructive'). Likewise if a test requires a reboot (before? after? during?) that could be noteworthy.

There's a lot of downstream opportunity here but we should rush to far toofast. First get the basic behavior working and checked in. Followed by docs. Then additional properties get interesting. My top 2 requests/desires after this PR...

1. **"TestPhase": [ "PR", "CI", "Nightly", "Release", "*" ]** -- Which phases(sp?) should run this? If not specified default = * (all). Update test infra so e.g. I can define a test that runs in the release and nightly builds but not in a PR build (I'd like to define a _subset_ of tests for that :-))
2. **"Platform": [ "+*", "-RS5*", "-19H*" ]** -- Which platform(s) should this run on? If not specified default = +* but one can include/exclude e.g. "[ "+*", "-RS5*", "-*de-DE*" ]" == include(+) all, exclude(-)RS5 (all arch, all lang, all release/debug) and exclude(-) all German images. Details TBD, goal is how do I specify a test doesn't run on certain images? e.g. How do I specify my tests only run on Win11 (do not run on Win10) because my feature only runs on Win11

build/AzurePipelinesTemplates/WindowsAppSDK-RunTests-Steps.yml
DrusTheAxe: I have tests I've been itching to light up but the test infrastructure had (significant) gaps needing love before I could (and I've been too busy to plug them myself). Paul's done an awesome job here not only plugging those gaps but simplifying the process to be practical for developers to use productively.

Lighting up my tests will still be work but with all this will be nearly trivial for me to do. _That_ I can find time for sooner than later :-)

In theory _most_ existing UT/FT can be lit up in the pipeline. A few have errors last I looked and/or may have errors when run in the pipeline and/or in all the test environments. More significantly, not all tests _should_ be run in all test passes -- e.g. we may not want all UT/FT to light up in PR/CI test phases. That's soluble with a little work (see other comment). Nothing that should hold up this PR, just lots of exciting opportunities for future PRs :D

Only real Q is if we want each team to make those calls or a more centralized effort by some/1 dev. I recommend the former as much as possible. The judgement if not-100% should run in all phases requires subject matter experts to make the right call. If the effort to make a .testdef is low (as it looks like) pinging teams to make a small investment to automate using existing tests should be an easy sell.

build/AzurePipelinesTemplates/WindowsAppSDK-RunTestsInPipeline-Job.yml
pmpurifoy: We are currently only building for release on the Build job. [https://github.com/microsoft/WindowsAppSDK/blob/73b7848a2acac05d310d6c5730b744a356f44480/build/ProjectReunion-BuildFoundation.yml#L53](url)

TestAll.ps1
pmpurifoy: After this PR, I'm planning on writing a doc to go into Coding Guidelines

build/AzurePipelinesTemplates/WindowsAppSDK-RunTests-Steps.yml
pmpurifoy: This script detects any test folder with a testdef and looks for the component filename.dll described from the testdef. But I'm not sure of the status of all component testing whether they are stable tests or not when ran in the pipeline.

pmpurifoy:
Logging these comments for future reference/improvements:
**Image selection**: https://github.com/microsoft/WindowsAppSDK/pull/3367#discussion_r1097094753
**Doc to Coding Guidelines**: https://github.com/microsoft/WindowsAppSDK/pull/3367#discussion_r1097074269

build/AzurePipelinesTemplates/WindowsAppSDK-RunTestsInPipeline-Job.yml
alexlamtest: I'm ok with not introducing the debug config in this PR.

build/AzurePipelinesTemplates/WindowsAppSDK-RunTests-Steps.yml
alexlamtest: Yeah, this is one of the items I'd like to follow up once this PR is in. Agreed that this is nonblocking.

DrusTheAxe (approved):
Awesome work

test/PushNotificationTests/PushNotificationTests.testdef
alexlamtest: I hope there is no "destructive" tests today. It's against guidance. Best for these to not exist, but if they do, they should be called out clearly. Agreed that details of a test that might evolve over time shouldn't be in the ,testdef, but being 'destructive' feels like something good to be clear. Others may disagree, but I think of "requires elevation" as one property of a test that does not change evolve quickly. When a developer uses the -List feature on a local machine, seeing what tests require elevation might save the developer some headache when some tests fail to run locally, only because they need to run elevated.

TestAll.ps1
alexlamtest: Sounds good. But how discoverable is Coding-Guidelines.md itself? I suspect most new developers would first look for readme.md. So, it might still be worthy to put a link to Coding-Guidelines.md in readme.md. Again, this is a nonblocking suggestion.

alexlamtest (approved):

TestAll.ps1
DrusTheAxe: > how discoverable is Coding-Guidelines.md itself?

Good question. We have a gap but it's easy to fix. I'll post a PR

DrusTheAxe: Added the missing link(s). Plus some assorted fixing and polish while I was in the neighborhood

https://github.com/microsoft/WindowsAppSDK/pull/3409

test/PushNotificationTests/PushNotificationTests.testdef
```diff
@@ -0,0 +1,18 @@
...
+            "Parameters": "/name:UnpackagedTests#metadataSet1::VerifyForegroundHandlerFails",
+            "Architectures": ["x64"],
```
hulumane, line 7: Should this be enabled for arm64 at some point too?

test/PushNotificationTests/PushNotificationTests.testdef
```diff
@@ -0,0 +1,18 @@
+{
+    "Tests": [
+        {
+            "Description": "Test unpackaged scenarios for PushNotifications",
+            "Filename": "PushNotificationTests.dll",
+            "Parameters": "/name:UnpackagedTests#metadataSet1::VerifyForegroundHandlerFails",
```
hulumane, line 6: Whats the part proceeding the #? Is this specifically running only a single test in the suite?

test/PushNotificationTests/PushNotificationTests.testdef
```diff
@@ -0,0 +1,18 @@
...
+            "Architectures": ["x64"],
+            "Status": "Disabled" 
+        },
+        {
+            "Description": "Test packaged scenarios for PushNotifications",
+            "Filename": "PushNotificationTests.dll",
+            "Parameters": "/name:PackagedTests#metadataSet1::VerifyForegroundHandlerFails",
+            "Architectures": ["x64"],
+            "Status": "Disabled"
```
hulumane, line 15: Will this be Enabled at some later point?

test/PushNotificationTests/PushNotificationTests.testdef
pmpurifoy: We can metadataSet1 is the value assigned by TAEF with conditional properties. This property is SelfContained='true'. MetadataSet0 is SelfContained='false'. Our tests are currently failing some activations due to an existing WinAppSDK issue: https://github.com/microsoft/WindowsAppSDK/issues/3137#issuecomment-1332851404, so for now I disabled our tests.

test/PushNotificationTests/PushNotificationTests.testdef
pmpurifoy: We can enable this for all archs but this is just the template for us. They are disabled due to WinAppSDK issue: https://github.com/microsoft/WindowsAppSDK/issues/3137#issuecomment-1332851404. I'll come back here with a better testdef in the future, but for now I am just laying down the basic foundation for WinAppSDK testing as a whole. I'll deal with notifications soon after.

alexlamtest (approved):

PR #3383, opened 2023-01-30
KozaniManager in-proc WinRT APIs and OOP classic COM APIs to activate remote apps

rhuang-msft:
Created KozaniManager in-proc WinRT APIs and OOP classic COM APIs to activate remote apps. Added tests and test package to validate the changes.


rhuang-msft:
@DrusTheAxe , @wcheng-msft , @cwruss , @sachintaMSFT - I have the PR open for review. Please take a look. Thanks!

rhuang-msft:
@rhuang-msft - test PR comment communication.

dev/Kozani/KozaniManagerRuntime/M.K.MR.ManagerRuntimeManager.h
```diff
@@ -3,16 +3,25 @@
...
+        void ActivateRemoteApplication(
+            Windows::ApplicationModel::Activation::ActivationKind activationKind,
+            winrt::hstring appUserModelId,
+            winrt::hstring connectionRdpFilePath,
+            winrt::hstring additionalSettingsFilePath,
+            Windows::ApplicationModel::Activation::IActivatedEventArgs args,
+            IInspectable statusCallback);
+
+    private:
+        wil::com_ptr<IKozaniManager> m_kozaniManager;
```
DrusTheAxe, line 24: You'll want to put in a static call and not implicitly get the manager

As expressed here the caller pattern is

```
ManagerRuntime::ManagerRuntimeManager manager;
manager.ActivateRemoteApplication(...);
```

Better to have the calling pattern

```
auto manager{ ManagerRuntime::ManagerRuntimeManager::Open(...) }
manager.ActivateRemoteApplication(...);
```

Leaves open the door for us to support more granular pre-activation work (e.g. establish connection w/o running any app, configure behavior, other odd things). Also easier to provide correct lifetiming - the former snippet implicitly creates the manager on 1st use, whereas the latter it's explicit as part of the act of acquiring a manager.

P.S. And no, that doesn't mean we need ActivateRemoteApplication() and other checks doing any sort of `!m_kozaniManager` null check before use. Since there's no way to instantiate a Manager w/o calling Open (or whatever we call it) m_kozaniManager is never null when someone gets their hands on it, but with no potential concurrency issue. And yes, no one _should_ call the same instance from 2 threads, but people do all kinds of crazy things like that - sometimes by accident, sometimes intentionally probing for errors, crashes, MSRC vulnerabilities, etc. The static Open pattern helps prevent that (mis)use inherent to the use.

test/Kozani/KozaniManagerRuntimeTests/KozaniManagerRuntimeTests.cpp
```diff
@@ -2,12 +2,70 @@
...
+{
+    constexpr PCWSTR c_PackageDirName = L"KozaniManager";
+    constexpr PCWSTR c_PackageFullName = L"KozaniManager_1.0.0.0_neutral__8wekyb3d8bbwe";
+}
+
 namespace Test::KozaniManagerRuntimeTests
 {
+    struct MyKozaniStatusCallback : winrt::implements<MyKozaniStatusCallback, IKozaniStausCallback, winrt::Windows::Foundation::IInspectable>
+    {
+#pragma region IKozaniStausCallback_methods
```
DrusTheAxe, line 21: Cute. Is it useful?

test/Kozani/KozaniManagerRuntimeTests/KozaniManagerRuntimeTests.cpp
```diff
@@ -2,12 +2,70 @@
...
+            return m_isActivationFailed;
+        }
+
+        bool IsClosed()
+        {
+            return m_isClosed;
+        }
+
+    private:
+        bool m_isActivated;
```
DrusTheAxe, line 64: NIT: `bool m_isActivated{};`

Ensure locals and attributes are initialized and ES.23: [Prefer the {}-initializer syntax](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#es23-prefer-the--initializer-syntax)

test/Kozani/KozaniManagerRuntimeTests/KozaniManagerRuntimeTests.cpp
```diff
@@ -18,18 +76,37 @@ namespace Test::KozaniManagerRuntimeTests
...
         TEST_CLASS_CLEANUP(ClassCleanup)
         {
+            ::TP::RemovePackageIfNecessary(KozaniManagerTestPackage::c_PackageFullName);
             ::TB::Cleanup();
             return true;
         }
 
-        // TODO Replace this with your own tests
-        TEST_METHOD(JustDoIt)
+        TEST_METHOD(ActivateRemoteApplicationTests)
```
DrusTheAxe, line 90: Tests? Plural?

1 test should do 1 thing. If there's multiple tests please make multiple methods. Easier to debug, repro, yadda yadda.

Plus a 'Test*' suffix isn't needed since the method's in a BlahBlahTests class.

This has 1 test so SUGGESTION: Rename to `ActivateRemoteApplication_Launch`

test/Kozani/KozaniManagerRuntimeTests/KozaniManagerRuntimeTests.cpp
```diff
@@ -18,18 +76,37 @@ namespace Test::KozaniManagerRuntimeTests
...
         {
+            try
+            {
+                winrt::Microsoft::Kozani::ManagerRuntime::ManagerRuntimeManager runtimeManager;
+                auto statusCallback{ winrt::make_self<MyKozaniStatusCallback>() };
+
+                runtimeManager.ActivateRemoteApplication(winrt::Windows::ApplicationModel::Activation::ActivationKind::Launch,
+                    L"Microsoft.WindowsCalculator_8wekyb3d8bbwe!App", L"c:\\data\\connection.rdp", L"c:\\data\\additionalSettings.txt",
```
DrusTheAxe, line 98: I assume the hardcoded pathnames are temporary hacks soon to be replaced...

SUGGESTION: Create `connection.rdp` and `additionalSettings.txt` in the project directory, tell VS to copy them to OutDir on build. See https://stackoverflow.com/questions/18743907/visual-studio-how-to-copy-to-output-directory-without-copying-the-folder-stru for more details, but with Content instead of ContentWithTargetPath if the data files are in the project's root (next to the .vcxproj)

P.S. `PreserveNewest` will update the OutDir copy if the source is newer, `Always` copies it every time you build, `Never` obviously never does. `PreserveNewest` is usually what you'd want.

test/Kozani/KozaniManagerRuntimeTests/KozaniManagerRuntimeTests.cpp
```diff
@@ -18,18 +76,37 @@ namespace Test::KozaniManagerRuntimeTests
...
+            {
+                winrt::Microsoft::Kozani::ManagerRuntime::ManagerRuntimeManager runtimeManager;
+                auto statusCallback{ winrt::make_self<MyKozaniStatusCallback>() };
+
+                runtimeManager.ActivateRemoteApplication(winrt::Windows::ApplicationModel::Activation::ActivationKind::Launch,
+                    L"Microsoft.WindowsCalculator_8wekyb3d8bbwe!App", L"c:\\data\\connection.rdp", L"c:\\data\\additionalSettings.txt", 
+                    nullptr,    // IActivatedEventArgs
+                    statusCallback.as<winrt::Windows::Foundation::IInspectable>());
+
+                VERIFY_IS_TRUE(statusCallback->IsActivated(), L"IKozaniStausCallback::OnActivated() should have been called.");
```
DrusTheAxe, line 102: Is the message needed or redundant?

test/Kozani/KozaniManagerRuntimeTests/KozaniManagerRuntimeTests.cpp
```diff
@@ -18,18 +76,37 @@ namespace Test::KozaniManagerRuntimeTests
...
+            }
+            catch (const winrt::hresult_error& e)
+            {
+                VERIFY_FAIL(WEX::Common::String().Format(
+                    L"ActivateRemoteApplication() failed with 0x%x, error message: %s",
+                    e.code(), e.message().c_str()));
```
DrusTheAxe, line 108: If you don't catch it what does TAEF output? Usually it's message is meaningful enough, doing your own catch/printf is redundant

test/Kozani/data/KozaniManager.Msix/KozaniManager.Msix.vcxproj
```diff
@@ -0,0 +1,111 @@
...
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Debug|x64">
+      <Configuration>Debug</Configuration>
+      <Platform>x64</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Release|x64">
+      <Configuration>Release</Configuration>
+      <Platform>x64</Platform>
+    </ProjectConfiguration>
+  </ItemGroup>
```
DrusTheAxe, line 20: Did you create the project via VS File/New...?

Just a hunch. Since there's no ARM64. And lines 28+ are highly redundant and not the optimized form we use. And...

Please use the standard project templates per `tools\AddProject -List`. Notably, `test.cpp.dll.taef` or `test.cpp.dll.msixpackage`. Do we need to add C++/WinRT support to `test.cpp.dll.msixpackage` or is there more to it you need?

DrusTheAxe (changes requested):

test/Kozani/data/KozaniManager.Msix/KozaniManager.Msix.vcxproj
rhuang-msft: This is copied and modified from existing msxi project (D:\Projects\WindowsAppSDK\test\DynamicDependency\data\DynamicDependency.DataStore.Msix\DynamicDependencies.DataStore.Msix.vcxproj). The current template does not meet the need as here we only need to package existing dll and exe from other projects. No need to add new dll code.

test/Kozani/KozaniManagerRuntimeTests/KozaniManagerRuntimeTests.cpp
rhuang-msft: VERIFY_FAIL will fail the test. TAEF will properly handle it and mark the test as failed.

test/Kozani/KozaniManagerRuntimeTests/KozaniManagerRuntimeTests.cpp
rhuang-msft: The message will explain what is expected and when the condition is not met and the test fails, it will tell the dev what is the expected behavior.

test/Kozani/KozaniManagerRuntimeTests/KozaniManagerRuntimeTests.cpp
rhuang-msft: yes, it is temporary to get through the API testing. No real functionality implemented yet so just a string to pass through will be enough. When it is implemented we will change the test to do meaningful work.

dev/Kozani/KozaniManager/main.cpp
```diff
@@ -22,13 +25,21 @@ static constexpr GUID KozaniManager_guid { PR_KOZANIMANAGER_CLSID_GUID };
...
+        LPCWSTR appUserModelId,
+        LPCWSTR connectionRdpFilePath,
+        LPCWSTR additionalSettingsFilePath,
+        ::IInspectable* activatedEventArgs,
+        IKozaniStausCallback* statusCallback,
+        DWORD associatedLocalProcessId)
     {
-        return S_OK;
-    }
+        // ToDo: temporary code to enable initial testing of the in-proc WinRT API and OOP COM API. Will be replaced with real impl later.
```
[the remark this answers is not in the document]
rhuang-msft: Should we put ADO link to the open source project? Isn't ADO link internal?

test/Kozani/KozaniManagerRuntimeTests/KozaniManagerRuntimeTests.cpp
rhuang-msft: renamed.

dev/Kozani/KozaniManager/main.cpp
```diff
@@ -22,13 +25,21 @@ static constexpr GUID KozaniManager_guid { PR_KOZANIMANAGER_CLSID_GUID };
 
 struct __declspec(uuid(PR_KOZANIMANAGER_CLSID_STRING)) KozaniManagerImpl WrlFinal : RuntimeClass<RuntimeClassFlags<ClassicCom>, IKozaniManager>
 {
-    STDMETHODIMP Initialize()
+    STDMETHODIMP ActivateRemoteApplication(
+        INT32 activationKind,
+        LPCWSTR appUserModelId,
+        LPCWSTR connectionRdpFilePath,
+        LPCWSTR additionalSettingsFilePath,
+        ::IInspectable* activatedEventArgs,
```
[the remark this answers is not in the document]
rhuang-msft: To make sure it is the root namespace, not to confused with winrt::XXX::IInspectable.

dev/Kozani/KozaniManager/main.cpp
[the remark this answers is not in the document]
DrusTheAxe: `https://task.ms/12345` is a public-friendly URL redirector. Follow the link and if you're authenticated and have access it takes you to the ADO item (if not you get access denied).

ADO item numbers aren't secrets. Just opaque cookies for reference. No problem sharing them here via task.ms

test/Kozani/KozaniManagerRuntimeTests/KozaniManagerRuntimeTests.cpp
DrusTheAxe: A thrown exception boiling up out of your test method is caught by TAEF and fails the test. All properly handled, no need for you to add your own catch block redundant with TAEF's behavior we get for free

test/Kozani/KozaniManagerRuntimeTests/KozaniManagerRuntimeTests.cpp
DrusTheAxe: Just wondering how useful that is here. Happily deferring to your judgement

test/Kozani/data/KozaniManager.Msix/KozaniManager.Msix.vcxproj
DrusTheAxe: >No need to add new dll code.

Then you'll 

Compare with the template `tools\ProjectTemplates\test.cpp.dll.msixpackage\PurojekutoTenpuret.vcxproj` and you'll quickly see how the config differs If you're just packaging and don't need to compile anything into a dll or exe you'll want a new project template (more of a 'utility' than `<ConfigurationType>DynamicLibrary</>`).

Or even simpler, you don't need a new project. Your project producing your test dll can ALSO package it. Go into your `.vcxproj` that makes your test dll and add `makemsix.targets` support e.g. the core of it's

```
  <PropertyGroup>
    <MakeMsixOutputFilename>$(MSBuildProjectName).msix</MakeMsixOutputFilename>
  </PropertyGroup>
  <ItemGroup>
    <MakeMsixInputs Include="$(MSBuildThisFileDirectory)appxmanifest.xml" />
    <MakeMsixInputs Include="$(MSBuildThisFileDirectory)logo.png" />
    <MakeMsixInputs Include="$(OutDir)$(TargetFileName)" />
  </ItemGroup>
  <PropertyGroup>
    <PackageCertificateKeyFile>$(RepoTestCertificatePFX)</PackageCertificateKeyFile>
    <PackageCertificatePassword>$(RepoTestCertificatePassword)</PackageCertificatePassword>
  </PropertyGroup>
  <Import Project="$(RepoRoot)\MakeMsix.targets" />
```

That adds a post-build step so after your project compiles its DLL it then runs commands in MakeMSIX.targets (notably, makeappexe+signtool.exe) to create your .msix.

No need for an "msix-dedicated"(ish) project.

dev/Kozani/KozaniManager/KozaniManager.vcxproj.filters
```diff
@@ -45,4 +42,12 @@
   <ItemGroup>
     <None Include="packages.config" />
   </ItemGroup>
-</Project>
+  <ItemGroup>
+    <Natvis Include="$(MSBuildThisFileDirectory)..\..\natvis\wil.natvis" />
```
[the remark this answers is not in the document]
rhuang-msft: Not sure how it got there. VS put it there? Removed.

dev/Kozani/KozaniManager/main.cpp
```diff
@@ -13,6 +13,9 @@
 
 #include <wrl\module.h>
 
+//#include <winrt/Microsoft.Kozani.ManagerRuntime.h>
```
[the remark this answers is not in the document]
rhuang-msft: removed

dev/Kozani/KozaniManager/main.cpp
```diff
@@ -22,13 +25,21 @@ static constexpr GUID KozaniManager_guid { PR_KOZANIMANAGER_CLSID_GUID };
 
 struct __declspec(uuid(PR_KOZANIMANAGER_CLSID_STRING)) KozaniManagerImpl WrlFinal : RuntimeClass<RuntimeClassFlags<ClassicCom>, IKozaniManager>
 {
-    STDMETHODIMP Initialize()
+    STDMETHODIMP ActivateRemoteApplication(
+        INT32 activationKind,
+        LPCWSTR appUserModelId,
```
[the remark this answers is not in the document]
rhuang-msft: replaced LPCWSTR with PCWSTR. In IDL we have to use LPCWSTR as PCWSTR is not recognized by MIDL. Was to match the IDL in cpp but I agree it is not needed.

dev/Kozani/KozaniManager/KozaniManager.vcxproj.filters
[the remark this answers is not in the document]
rhuang-msft: Not sure how it got there. Maybe VS automatically putting it there when I messed around to get something build. Removed.

dev/Kozani/KozaniManager/main.cpp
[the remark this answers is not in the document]
rhuang-msft: Added ADO link.

dev/Kozani/KozaniManagerInterface/KozaniManagerInterface.idl
```diff
@@ -3,13 +3,33 @@
 
 import "oaidl.idl";
 import "ocidl.idl";
+import "inspectable.idl";
+
+[object]
+[uuid(5239c3b0-9e90-11ed-87cd-0800200c9a66)]
+[pointer_default(unique)]
+interface IKozaniStausCallback : IInspectable
```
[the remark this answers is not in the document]
rhuang-msft: Yes, the callback will be implemented in cpp/winrt, which will not interface with IUnknown. The WinRT IDL will not take IUnknown as parameter.

dev/Kozani/KozaniManagerInterface/KozaniManagerInterface.idl
```diff
@@ -3,13 +3,33 @@
...
+{
+    HRESULT OnActivated(DWORD pid);
+
+    HRESULT OnActivationFailed(
+        DWORD errorCode,
+        [in, string] LPCWSTR errorMessage);
+
+    HRESULT OnClosed();
```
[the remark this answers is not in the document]
rhuang-msft: This is a classic COM interface, cannot reference ICLosable.

dev/Kozani/KozaniManagerInterface/KozaniManagerInterface.idl
```diff
@@ -3,13 +3,33 @@
...
 import "oaidl.idl";
 import "ocidl.idl";
+import "inspectable.idl";
+
+[object]
+[uuid(5239c3b0-9e90-11ed-87cd-0800200c9a66)]
+[pointer_default(unique)]
+interface IKozaniStausCallback : IInspectable
+{
+    HRESULT OnActivated(DWORD pid);
```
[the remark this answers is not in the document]
rhuang-msft: No, this is classic COM interface, derived from IInspectable so it can be passed as a parameter as IInspectable through WinRT APIs.

dev/Kozani/KozaniManagerInterface/KozaniManagerInterface.idl
```diff
@@ -3,13 +3,33 @@
...
+
+    HRESULT OnActivationFailed(
+        DWORD errorCode,
+        [in, string] LPCWSTR errorMessage);
```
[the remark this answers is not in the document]
rhuang-msft: This is classic COM with classic COM marshalling.

dev/Kozani/KozaniManagerRuntime/KozaniManagerRuntime.vcxproj.filters
```diff
@@ -53,6 +53,9 @@
     <ClCompile Include="M.K.MR.ManagerRuntimeManager.cpp">
       <Filter>Source Files</Filter>
     </ClCompile>
+    <ClCompile Include="$(GeneratedFilesDir)module.g.cpp">
```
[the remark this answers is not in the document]
rhuang-msft: I did exactly the same steps you mentioned. This is not manually added. Must be done by VS.

test/Kozani/KozaniManagerRuntimeTests/KozaniManagerRuntimeTests.cpp
rhuang-msft: Yes, this is how I validated the API E2E across processes actually works. It uncovered code that compiled but didn't work OOP due to missing proxy stub, etc.

test/Kozani/data/KozaniManager.Msix/KozaniManager.Msix.vcxproj
rhuang-msft: I am following existing pattern in WinAppSDK project for "msix-dedicated"(ish) projects, which I found intuitive and convenient for creating test MSIX packages. Those packages can refer to multiple code projects that build various exe and dlls. A dedicated packaging project to organize relevant build outputs makes sense and it is the model followed by product code like MSRDC. They has dedicated MSIX packaging project like this as well. Note that the test MSIX package can be used in different test projects so for a test code project to build it seems less intuitive.

test/Kozani/KozaniManagerRuntimeTests/KozaniManagerRuntimeTests.cpp
rhuang-msft: Yes, it is very useful for people to understand what a test failure means.

wcheng-msft, line 102: >aus [](http://example.com/codeflow?start=73&length=3)

missing t

dev/Kozani/KozaniManagerInterface/KozaniManagerInterface.idl
wcheng-msft, line 11: >aus [](http://example.com/codeflow?start=19&length=3)

missing t

rhuang-msft: Fixed the typo.

test/Kozani/KozaniManagerRuntimeTests/KozaniManagerRuntimeTests.cpp
rhuang-msft: Fixed.

test/Kozani/KozaniManagerRuntimeTests/KozaniManagerRuntimeTests.cpp
rhuang-msft: Member variables should be by default set to default value. Anyway, added {}.

test/Kozani/KozaniManagerRuntimeTests/KozaniManagerRuntimeTests.cpp
rhuang-msft: Ok, removed try catch.

dev/Kozani/KozaniManagerRuntime/M.K.MR.ManagerRuntimeManager.h
rhuang-msft: Updated to ManagerRuntime::ManagerRuntimeManager::Create(...)

dev/Kozani/KozaniManagerRuntime/M.K.MR.ManagerRuntimeManager.cpp
```diff
@@ -9,6 +9,16 @@
 
 namespace winrt::Microsoft::Kozani::ManagerRuntime::implementation
 {
+    ManagerRuntimeManager::ManagerRuntimeManager()
+    {
+        m_kozaniManager = wil::CoCreateInstance<KozaniManager, IKozaniManager>(CLSCTX_LOCAL_SERVER);
```
DrusTheAxe, line 14: Seems a bit odd to do in the ctor but the ctor's private, right? Only way to get here is via the static Create()?

test/Kozani/KozaniManagerRuntimeTests/KozaniManagerRuntimeTests.cpp
```diff
@@ -16,9 +16,9 @@ namespace KozaniManagerTestPackage
 
 namespace Test::KozaniManagerRuntimeTests
 {
-    struct MyKozaniStatusCallback : winrt::implements<MyKozaniStatusCallback, IKozaniStausCallback, winrt::Windows::Foundation::IInspectable>
+    struct MyKozaniStatusCallback : winrt::implements<MyKozaniStatusCallback, IKozaniStatusCallback, winrt::Windows::Foundation::IInspectable>
     {
-#pragma region IKozaniStausCallback_methods
+#pragma region IKozaniStatusCallback_methods
```
DrusTheAxe, line 21: Does the region name need to be 1 word?

(been a while since I used this pragma)

DrusTheAxe (approved):

dev/Kozani/KozaniManagerRuntime/M.K.MR.ManagerRuntimeManager.cpp
rhuang-msft: we cannot put ctor in private as it will cause cpp/winrt build error. I tried and it failed. It is the same pattern as all the other APIs in this solution. None of them have private ctor.

test/Kozani/KozaniManagerRuntimeTests/KozaniManagerRuntimeTests.cpp
rhuang-msft: it can have space.

PR #3388, opened 2023-02-01
[CherryPick] Fix PushNotificationChannel::ExpirationTime

pmpurifoy:
…eTime (#3335)

* Initial fix
The ExpirationTime being returned by PushNotificationChannel has been returning incorrectly. Instead of returning value 30 days from the request, it is returning some value in year 1602. This is because originally, the code was using winrt::copy_from_abi to copy the DateTime. The fix is to take the ExpirationTime from the ABI and convert it to a winrt::DateTime.

* Add unit test
Added a unit test verifying the returned channel is within the expected range.

* Change constraint

* Update upper bound

* Change upper bound to 1 min

WARNING:

### DIFF
--- a/test/PushNotificationTests/BaseTestSuite.cpp
+++ b/test/PushNotificationTests/BaseTestSuite.cpp
@@ -18,6 +18,8 @@ using namespace winrt::Windows::Management::Deployment;
 using namespace winrt::Windows::Storage;
 using namespace winrt::Windows::System;
 using namespace winrt::Microsoft::Windows::PushNotifications;
+using namespace std::literals;
+using namespace std::chrono;
 
 void BaseTestSuite::ClassSetup()
 {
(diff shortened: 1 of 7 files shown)


test/PushNotificationTests/BaseTestSuite.cpp
```diff
@@ -113,6 +115,27 @@ void BaseTestSuite::ChannelRequestUsingRemoteId()
...
+{
+    if (PushNotificationManager::Default().IsSupported())
+    {
+        auto channelOperation{ PushNotificationManager::Default().CreateChannelAsync(c_azureRemoteId) };
+        VERIFY_SUCCEEDED(ChannelRequestHelper(channelOperation));
+
+        auto channel{ channelOperation.GetResults().Channel() };
+        auto expirationTime{ channel.ExpirationTime() };
+        auto expiryBound{ winrt::clock::now() + (hours(24) * 30) + minutes(1) };
+
```
hulumane, line 128: You checked the upper bound but forgot to check the lower bound

hulumane: This means that if Expiry time returns 1800/01/01, the test would pass incorrectly

hulumane: The lower bound would be the current time.

hulumane (approved):

PR #2648, opened 2022-06-20
AppNotificationContent Builder 1.2 spec

pmpurifoy:
This PR introduces the AppNotificationContent builder spec to the Windows App SDK. Developers will use this construct to build rich AppNotification XML payloads in a much easier way.

UI Components that will be supported for 1.2:
- Text
- Image
- TextBox
- ComboBox
- Audio
- Button
- ProgressBar

Developers will be able to retrieve the local AppNotification payload using AppNotificationContent::GetXml.

PR #3393, opened 2023-02-02
AppNotificationComboBox Builder Bug Fixes

loneursid:
This PR addresses two bugs:
1. AppNotificationComboBox should respect the developer's wish when ordering its items in xml.
When generating its xml output, AppNotificationComboBox sorts the items according to their keys (items are key value pairs). The issue with that strategy is that the key names are completely unrelated to the sorting order. AppNotificationComboBox should simply list the items in the order they were added by the developer.

2. AppNotificationComboBox should immediately report an error when a duplicate item is added
AppNotificationComboBox items are represented by key  value properties. If a developer adds an item with the same key twice, the old item is replaced with the new one and no error is reported. This behavior is almost never what a developer wants. Adding two items with the same key in a combobox is almost always an error on the part of the developer. Instead of silently replacing old item with the new one, AppNotificationComboBox should detect the situation and immediately report the error to the developer.

There are other issues with how properties are managed / validated (or not validated for that matter) but these will be addressed in a separate PR as they also involve other parts of the builder: https://github.com/microsoft/WindowsAppSDK/pull/3405

@MikeHillberg was invited to the review as we are making a few changes to the API, including deprecating a property.


dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.idl
```diff
@@ -118,6 +118,24 @@ namespace Microsoft.Windows.AppNotifications.Builder
         AppNotificationComboBox SetSelectedItem(String id);
     };
 
+    [contract(AppNotificationBuilderContract, 2)]
+    runtimeclass AppNotificationComboBox2
+    {
```
loneursid, line 123: There might be code outthere that relies on the flawed behavior of the initial version of the AppNotificationComboBox builder, hence the need for a new version.

There is also a slight change in the API: while the original AppNotificationComboBox builder was exposing the key value map used to keep the items, this version does not. For one, the new implementation uses two internal data structures: a map to keep the key value pairs and a vector to track the ordering of the items. Both structures must be kept in sync.

There is a point to be made that all the items could be kept in the vector and the map is overkill but doing so would still require some extra code.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.idl
```diff
@@ -118,6 +118,24 @@ namespace Microsoft.Windows.AppNotifications.Builder
...
+        AppNotificationComboBox2(String id);
+
+        String Title;
+        String SelectedItem;
+
```
loneursid, line 128: Not a fan of exposing the class innards this way. Having a getter seems resonable but a setter seems more trouble than it is worh. These two setters, don't offer any validation contrairy to the fluent setter which do some validation.

Even worse, the fluent setter for the title encodes the value into xml, while the basic setter for the title does not. If the caller doesn't properly encode the value, then the builder, at best will produce a notification that can't be used, at worse, it will crash! This pattern is in use throught all the builder API and addressing it is beyond the scope of this PR.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.idl
```diff
@@ -118,6 +118,24 @@ namespace Microsoft.Windows.AppNotifications.Builder
         AppNotificationComboBox SetSelectedItem(String id);
     };
 
+    [contract(AppNotificationBuilderContract, 2)]
+    runtimeclass AppNotificationComboBox2
```
loneursid, line 122: AppNotificationComboBox2 isn't the greatest name. I'm open to suggestion, if someone has a better idea. The main difference is that this version keeps the items in the order they were added by the caller, rather than re-ordering them arbitrarily.

dev/AppNotifications/AppNotificationBuilder/AppNotificationComboBox2.h
```diff
@@ -0,0 +1,46 @@
...
+        winrt::Microsoft::Windows::AppNotifications::Builder::AppNotificationComboBox2 SetTitle(winrt::hstring const& value);
+
+        winrt::Microsoft::Windows::AppNotifications::Builder::AppNotificationComboBox2 SetSelectedItem(winrt::hstring const& id);
+
+        // IStringable
+        winrt::hstring ToString();
+
+    private:
+        winrt::hstring m_id{};
+        winrt::Windows::Foundation::Collections::IVector<winrt::hstring> m_ids{ winrt::single_threaded_vector<winrt::hstring>() };
```
loneursid, line 32: Since the vector and the map aren't exposed as properties in this version of the class, maybe we should ditch the winrt stuff and use STL instead.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.idl
pmpurifoy: Exposing these setters is moreso for C# developers. https://github.com/microsoft/WindowsAppSDK/pull/2648#discussion_r924785223

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.idl
pmpurifoy: My thoughts were that this is a bug to be fixed and didn't need an entire other object to be designed. Anyways, if we want to follow std library style we can do AppNotificationUnorderedComboBox. That would imply that the AppNotificationComboBox has some ordering (which it does) and make sense as a separate class.

But I hope that we don't have to do this, I'd rather just fix the impl inside the original ComboBox.

loneursid: AppNotificationUnorderedComboBox, unordered implies that the developers doesn't have control on the order. Usually, it means ordered by hashtags for efficiency. In this case we very much want the developers to control the order.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.idl
loneursid: Then shouldn't we apply the same level of validation we apply to fluent setters? We shouldn't let external developers just set these values without doing any validation on them.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.idl
loneursid: You can't keep the original runtimeclass because we can't expose the internal map anymore as the innards of the class have changed.

If we want to keep exposing a properties for our C# friends, then it would have to be a vector (of KeyValuePairs) instead of a map.

Either way, that's a foundamental change to the API and can only be achieved with a new runtimeclass

hulumane, line 122: Can you follow up with our API reps on this? It will be a point of contention in the API review. Traditionally, we add more contracts and update the versioning in the existing runtime class. See wpnapps idl on how this is accomplished. 

Older properties that are not applicable anymore should go through a deprecation ritual. ie. You could say E_NOTIMPL  and throw an exception.

loneursid: Done. We're going to go the deprecation road. For now, the caller should see a warning that the old property is being deprecated. if the developer use the old property they will incur a copy as we have to convert data from the old structure to the new one and vice-versa. The good news is that we can't have more than 5 items, so the copy won't have any performance impact.

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.idl
```diff
@@ -99,12 +99,22 @@ namespace Microsoft.Windows.AppNotifications.Builder
...
+        String Content;
+    };
+
     [contract(AppNotificationBuilderContract, 1)]
     runtimeclass AppNotificationComboBox
     {
         AppNotificationComboBox(String id);
 
+        [deprecated("Use ItemList instead of Items. For more info, see MSDN.", deprecate, AppNotificationBuilderContract, 2)]
         Windows.Foundation.Collections.IMap<String, String> Items;
```
loneursid, line 115: The Items property has been deprecated and replaced by the itemList property (couldn't think of a better name, but open to suggestions).

In this implementation, AppNotificationComboBox does the conversion between this old property and its internal representation. The alternative would be to throw a "not implemented" error. Throwing is easier but I'm not sure this would be appropriate behavior. @MikeHillberg do you have an opinion on this?

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.idl
loneursid: Fair enough. Innards a spilling out once more...

dev/AppNotifications/AppNotificationBuilder/AppNotificationBuilder.idl
loneursid: Resolving, there's noneed for a new runtimeclass.

dev/AppNotifications/AppNotificationBuilder/AppNotificationComboBox2.h
loneursid: Resolving as this has become irrelevant. The vector is exposed through a property.

PR #3398, opened 2023-02-03
Patch version of TAEF referenced by DevCheck

DrusTheAxe:
Quick patch to unblock folks.

More dynamic update using Version.Dependencies.xml info coming in a future PR

### DIFF
--- a/DevCheck.ps1
+++ b/DevCheck.ps1
@@ -605,7 +605,7 @@ function Repair-TAEFService
 
     $root = Get-ProjectRoot
     $cpu = Get-CpuArchitecture
-    $taef = 'Microsoft.Taef.10.58.210222006-develop'
+    $taef = 'Microsoft.Taef.10.75.221207001'
     $path = "$root\packages\$taef\build\Binaries\$cpu\Wex.Services.exe"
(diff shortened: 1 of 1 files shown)


pmpurifoy (approved):

DrusTheAxe:
Quick patch to unblock folks.

More dynamic update using Version.Dependencies.xml info coming in a future PR

DevCheck.ps1
```diff
@@ -605,7 +605,7 @@ function Repair-TAEFService
 
     $root = Get-ProjectRoot
     $cpu = Get-CpuArchitecture
-    $taef = 'Microsoft.Taef.10.58.210222006-develop'
+    $taef = 'Microsoft.Taef.10.75.221207001'
```
bpulliam, line 608: Why not keep this value in Version.Details.xml?

DrusTheAxe: Time.

This is just a quick'n'dirty patch to unblock folks. Another (more complex) PR will use Version.Dependencies.xml, as you'd expect

Patience please. Rome wasn't built in a fortnight :P

bpulliam (approved):

PR #3399, opened 2023-02-04
Refactoring MRTCore and Consolidate Nuget.configs

kythant:
Refactor MRTCore to eliminate relative paths with the definition of $(MRTCoreRoot)

Also consolidate all nuget.configs.


alexlamtest (approved):
Nit: nonblocking - the PR description appears to be unpopulated. The PR title might give the impression that only MRTCore is impacted but in fact changes also affect \build\..., hence, some short elaborations might be helpful to the reader.

kythant:
> Nit: nonblocking - the PR description appears to be unpopulated. The PR title might give the impression that only MRTCore is impacted but in fact changes also affect \build..., hence, some short elaborations might be helpful to the reader.

Fixed!

PR #3403, opened 2023-02-06
VSWhere

DrusTheAxe:
Change Visual Studio detection to use the locally installed vswhere.exe (%ProgramFiles(X86)%\Microsoft Visual Studio\Installer\vswhere.exe) unless -OnlineVSWhere is specified. The locally installed version is sufficient now that we require VS2022 to create WinAppSDK.

TAEF service handling is much smarter now. -CheckTAEFService (or -All / no-parameters) verifies the TAEF service's version matches the version in `eng\Version.Dependencies.xml`. If an older version is detected the current TAEF service is stopped and uninstalled and the (newer) one is installed and started.

NOTE: Detecting a newer version is OK, though a WARNING message is output.

New commands

* -OnlineVSWhere => Download and use the latest vswhere.exe on the network
* -StartTAEFService => Start the TAEF service
* -StopTAEFService => Stop the TAEF service

Error and Warning messages are now colorized (most weren't).

Fixed syntax error if -NoInteractive was specified and a prompt was needed for a certificate password.


DevCheck.ps1
```diff
@@ -186,6 +207,105 @@ function Get-CpuArchitecture
...
+        return $cmp
+    }
+
+    $cmp = [int]$lefty[1] - [int]$righty[1]
+    if ($cmp -ne 0)
+    {
+        return $cmp
+    }
+
+    $cmp = [int]$lefty[1] - [int]$righty[1]
```
alexlamtest, line 232: Why [1] again?

DevCheck.ps1
```diff
@@ -186,6 +207,105 @@ function Get-CpuArchitecture
...
+{
+    # Absolute Path = "%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe"
+
+    $programfilesx86 = [Environment]::GetFolderPath([Environment+SpecialFolder]::ProgramFilesX86)
+    $path = Join-Path $programfilesx86 "Microsoft Visual Studio\Installer\vswhere.exe"
+    if (-not(Test-Path -Path $path -PathType Leaf))
+    {
+        return $null
+    }
+    $path
```
alexlamtest, line 279: Get-TAEFPackageVersion above uses "return $taef.Version", so do "return $path" here also for consistency?

alexlamtest (approved):

PR #3429, opened 2023-02-13
API spec document for msix packaging in winappsdk

sreading-MSFT:
API spec document for msix packaging in winappsdk

WARNING:

### DIFF
--- a/specs/WinRT/MakeMSIX-spec.md
+++ b/specs/WinRT/MakeMSIX-spec.md
@@ -1,398 +1,87 @@
 # MakeMSIX API in Windows App SDK
 - [1. Background](#1-background)
-- [2. Existing interfaces and options](#2-existing-interfaces-and-options)
-  - [2.1. Makeappx](#21-makeappx)
-    - [2.1.1. pack](#211-pack)
-    - [2.1.2. bundle](#212-bundle)
-    - [2.1.3. unpack](#213-unpack)
-    - [2.1.4. unbundle](#214-unbundle)
-  - [2.2. Makemsix](#21-makemsix)
-    - [2.2.1. pack](#221-pack)
-    - [2.2.2. bundle](#222-bundle)
-    - [2.2.3. unpack](#223-unpack)
-    - [2.2.4. unbundle](#224-unbundle)
-  - [2.3. MSIXMgr](#23-msixmgr)
-    - [2.3.1. unpack](#231-unpack)
-- [3. Summary of existing options in relation to proposed api](#3-summary-of-existing-options-in-relation-to-proposed-api)
-- [4. Examples](#4-examples)
-    - [4.1. Packing a package](#41-packing-a-package)
-    - [4.2. Unbundling and unpacking a package to change information](#42-unbundling-and-unpacking-a-package-to-change-information)
-    - [4.3. Helper methods for examples](#43-helper-methods-for-examples) 
-- [5. API Details](#5-api-details)
+- [2. Summary of existing options in relation to proposed API](#2-summary-of-existing-options-in-relation-to-proposed-API)
+- [3. Examples](#4-examples)
+    - [3.1. Packing a package](#41-packing-a-package)
+    - [3.2. Unbundling and unpacking a package to change information](#32-unbundling-and-unpacking-a-package-to-change-information)
+    - [3.3. Helper methods for examples](#33-helper-methods-for-examples) 
+- [4. API Details](#4-API-details)
+- [5. Appendix](#5-Appendix)
+- [6. Existing interfaces and options](#6-existing-interfaces-and-options)
+  - [6.1. Makeappx](#61-makeappx)
+    - [6.1.1. pack](#611-pack)
+    - [6.1.2. bundle](#612-bundle)
+    - [6.1.3. unpack](#613-unpack)
+    - [6.1.4. unbundle](#614-unbundle)
+  - [6.2. Makemsix](#62-makemsix)
+    - [6.2.1. pack](#621-pack)
+    - [6.2.2. bundle](#622-bundle)
+    - [6.2.3. unpack](#623-unpack)
+    - [6.2.4. unbundle](#624-unbundle)
+  - [6.3. MSIXMgr](#63-msixmgr)
+    - [6.3.1. unpack](#631-unpack)
 
 # 1. Background
 
-The MakeMSIX api allows developers to create app packages for distribution. The api
-provides a WinRT interface that joins the functionality from multiple existing command line tools in
-one developer friendly location to allow creation of msix\appx packages and msixbundle\appxbundle packages,
-unpacking of those packages, and creation of package images for Azure App Attach.
-In addition to functionality from those existing tools, this api adds support for creation of Kozani
-packages.
+The MakeMSIX API allows developers to create app packages for distribution. The API provides a WinRT
+interface that joins the functionality from multiple existing command line tools in one developer
+friendly location to allow creation of msix\appx packages and msixbundle\appxbundle packages,
+unpacking of those packages, and creation of package images for Azure App Attach. In addition to
+functionality from those existing tools, this API adds support for creation of Kozani packages.
 
-# 2. Existing interfaces and options
+# 2. Summary of existing options in relation to proposed API
 
-The existing tools are:
--   [The makeappx command line tool](https://learn.microsoft.com/en-us/windows/win32/appxpkg/make-appx-package--makeappx-exe-)
-    Shipped as part of the Windows Development Kit.
-    Exposes pack, unpack, bundle, and unbundle commands, as well as encryption support
-    via a custom key-file list file, and creation of packages via a custom mapping file
-    to allow creation of packages with different payloads from the same folder.
-
--   [The makemsix command line tool and dll](https://github.com/Microsoft/msix-packaging)
-    Available through the github project.
-    Exposes pack, unpack, unbundle, and bundle* commands (*Bundle command only supports creation of
-    flat bundles. Flat bundles are bundles where package locations are stored as external path references
-    rather than being stored inside the msixbundle file itself. Flat bundles referred to as
-    sparse bundles in the makeappx documentation).
-    
--   [The msixmgr command line tool](https://github.com/Microsoft/msix-packaging and 
-    https://learn.microsoft.com/en-us/azure/virtual-desktop/app-attach-msixmgr)
-    Available through github and as a direct download from the learn.microsoft.com site.
-    Exposes "unpack" command to convert msix packages to vhd, vhdx, cim image files.
-
-None of the existing tools provide a good programmatic interface for all aspects of packaging.
-
-The makeappx.exe built into windows provides the only complete implementation of packaging
-and bundling but requires CreateProcess and does not return any specific errors from process exit.
-It also does not support any of the azure app attach scenarios.
-
-The makemsix tool is designed as a cross-platform implementation of packaging, but does not support
-a complete set of bundling features. The functionality is exposed as a command line as well as flat dll exports.
-
-The msixmgr tool is mostly for specific azure app attach packaging scenarios that are not covered by
-the other tools. To use it programmatically would require either using CreateProcess or building
-the msix-packaging github project and linking the static lib it produces into a project.
-
-## 2.1. Makeappx
-
-### 2.1.1. Pack
-Usage:
-------
-    MakeAppx pack [options] /d <content directory> /p <output package name>
-    MakeAppx pack [options] /f <mapping file> /p <output package name>
-    MakeAppx pack [options] /m <app package manifest> /f <mapping file> /p <output package name>
-    MakeAppx pack [options] /r /m <app package manifest> /f <mapping file> /p <output package name>
-    MakeAppx pack [options] /d <content directory> /ep <encrypted output package name> /kf <key file>
-    MakeAppx pack [options] /d <content directory> /ep <encrypted output package name> /kt
-
-Options:
---------
-    /h, /hashAlgorithm: Specifies a hash algorithm to use for creating the
-        block map. Valid algorithm IDs are SHA256, SHA384, and SHA512. The
-        default is SHA256.
-    /nv, /noValidation: Skips validation that ensures the package will be
-        installable on Windows. The validation include: existence of files
-        referenced in manifest, ContentGroupMap correctness, and additional
-        manifest validation on Protocols and FileTypeAssociation. By default,
-        all semantic validation is performed.
-    /nfv, /noFileValidation: Skips validation that ensures that the files used
-        to create the package (from a folder through /d or layout file through
-        /l) exist and are accessible. By default, all files must exist and not
-        be read locked. If a file is inaccessible, it will not be added to the
-        package.
-    /nc, /noCompress: Prevents MakeAppx from compressing files in the package.
-        By default, files in the package are compressed based on detected file
-        type.
-    /m: Specifies the path to an input app manifest which will be used as the
-        basis for generating the output app package or resource package's
-        manifest. When you use this option, you must also use /f and include a
-        [ResourceMetadata] section in the mapping file to specify the resource
-        dimensions to be included in the generated manifest.
-    /r: Builds a resource package. You must use the /m option with this.
-    /kf: Use this option to encrypt or decrypt the package or bundle using a
-        key file. This option cannot be combined with /kt.
-    /kt: Use this option to encrypt or decrypt the package or bundle using the
-        global test key. This option cannot be combined with /kf.
-    /cgm: The input content group map (CGM) file path used to create a
-        stream-able package. Providing a content group map through this option
-        will disable the check that all files in the content group map exist in
-        the package.
-    /pri, /makepriExeFullPath: You can use /pri to override the default
-        MakePri.exe path with the custom fullpath from which makeappx.exe will
-        launch the tool from when needed.
-    /mp: Specifies the path to a main package in the same package family as the
-        resource package being built.  You should provide this when encrypting
-        a resource package or when the resource package contains a content
-        group map.
-    /pb, /publisherBridging: Use this option to add publisher bridging entries
-        to the package or bundle.  Publisher bridging is useful when the new
-        issued cert subject name changed but the old publisher name is still
-        desired to maintain package identity continuity.
-    /np, /noParallel: Use this option to disable parallel execution of this
-        command.
-    /ml, /memoryLimit: Specify maximum memory in bytes that MakeAppx should
-        consume while executing in parallel. By default, this value is set to
-        half of total physical memory.
-    /o, /overwrite: Forces the output to overwrite any existing files with the
-        same name. By default, the user is asked whether to overwrite existing
-        files with the same name. You can't use this option with /no.
-    /no, /noOverwrite: Prevents the output from overwriting any existing files
-        with the same name. By default, the user is asked whether to overwrite
-        existing files with the same name. You can't use this option with /o.
-    /v, /verbose: Enables verbose output of messages to the console.
-    /?, /help: Displays this help text.
-
-### 2.1.2. Bundle
-
-Usage:
-------
-    MakeAppx bundle [options] /d <content directory> /p <output bundle name>
-    MakeAppx bundle [options] /f <mapping file> /p <output bundle name>
-    MakeAppx bundle [options] /d <content directory> /ep <encrypted output bundle name> /kf MyKeyFile.txt
-    MakeAppx bundle [options] /f <mapping file> /ep <encrypted output bundle name> /kt
-
-Options:
---------
-    /bv: Specifies the version number of the bundle being created. The version
-        must be in dotted-quad notation of four integers
-        <Major>.<Minor>.<Build>.<Revision> ranging from 0 to 65535 each. If the
-        /bv option is not specified or is set to 0.0.0.0, the bundle is created
-        using the current date-time formatted as the version:
-        <Year>.<Month-Day>.<Hour-Minute>.<Second-Millisecond>.
-    /mo: Generates a bundle manifest only, instead of a full bundle. Input
-        files must all be package manifests in XML format if this option is
-        specified.
-    /fb: Generates a fully sparse bundle where all packages are references to
-        packages that exist outside of the bundle file
-    /pri, /makepriExeFullPath: You can use /pri to override the default
-        MakePri.exe path with the custom fullpath from which makeappx.exe will
-        launch the tool from when needed.
-    /kf: Use this option to encrypt or decrypt the package or bundle using a
-        key file. This option cannot be combined with /kt.
-    /kt: Use this option to encrypt or decrypt the package or bundle using the
-        global test key. This option cannot be combined with /kf.
-    /o, /overwrite: Forces the output to overwrite any existing files with the
-        same name. By default, the user is asked whether to overwrite existing
-        files with the same name. You can't use this option with /no.
-    /no, /noOverwrite: Prevents the output from overwriting any existing files
-        with the same name. By default, the user is asked whether to overwrite
-        existing files with the same name. You can't use this option with /o.
-    /v, /verbose: Enables verbose output of messages to the console.
-    /?, /help: Displays this help text.
-
-### 2.1.3. Unpack
-Usage:
-------
-    MakeAppx unpack [options] /p <input package name> /d <output directory>
-    MakeAppx unpack [options] /ep <encrypted input package name> /d <output directory> /kf <key file>
-    MakeAppx unpack [options] /ep <encrypted input package name> /d <output directory> /kt
-
-Options:
---------
-    /pfn: Unpacks all files to a subdirectory under the specified output path,
-        named after the package full name.
-    /nv, /noValidation: Skips validation that ensures the package will be
-        installable on Windows. The validation include: existence of files
-        referenced in manifest, ContentGroupMap correctness, and additional
-        manifest validation on Protocols and FileTypeAssociation. By default,
-        all semantic validation is performed.
-    /kf: Use this option to encrypt or decrypt the package or bundle using a
-        key file. This option cannot be combined with /kt.
-    /kt: Use this option to encrypt or decrypt the package or bundle using the
-        global test key. This option cannot be combined with /kf.
-    /nd: Skips decryption when unpacking an encrypted package or bundle.
-    /o, /overwrite: Forces the output to overwrite any existing files with the
-        same name. By default, the user is asked whether to overwrite existing
-        files with the same name. You can't use this option with /no.
-    /no, /noOverwrite: Prevents the output from overwriting any existing files
-        with the same name. By default, the user is asked whether to overwrite
-        existing files with the same name. You can't use this option with /o.
-    /v, /verbose: Enables verbose output of messages to the console.
-    /?, /help: Displays this help text.
-
-### 2.1.4. Unbundle
-
-Usage:
-------
-    MakeAppx unbundle [options] /p <input bundle name> /d <output directory>
-    MakeAppx unbundle [options] /ep <encrypted input bundle name> /d <output directory> /kf <key file>
-    MakeAppx unbundle [options] /ep <encrypted input bundle name> /d <output directory> /kt
-
-Options:
---------
-    /pfn: Unpacks all files to a subdirectory under the specified output path,
-        named after the package full name.
-    /kf: Use this option to encrypt or decrypt the package or bundle using a
-        key file. This option cannot be combined with /kt.
-    /kt: Use this option to encrypt or decrypt the package or bundle using the
-        global test key. This option cannot be combined with /kf.
-    /nd: Skips decryption when unpacking an encrypted package or bundle.
-    /o, /overwrite: Forces the output to overwrite any existing files with the
-        same name. By default, the user is asked whether to overwrite existing
-        files with the same name. You can't use this option with /no.
-    /no, /noOverwrite: Prevents the output from overwriting any existing files
-        with the same name. By default, the user is asked whether to overwrite
-        existing files with the same name. You can't use this option with /o.
-    /v, /verbose: Enables verbose output of messages to the console.
-    /?, /help: Displays this help text.
-
-## 2.2. MakeMsix
-
-MakeMSIX is generally a subset of MakeAppx's functionality. It lacks support for
-encryption, content groups, mapping files, and bundle support is limited to
-only allowing creation of flat bundles (bundles where the appxbundle file itself
-only references but does not contain the main and resource appxpackages)
-
-It does add the ability to optionally unpack all packages inside the bundle
-in one command using the -pfn-flat option, as in:
-makemsix.exe unbundle -p <bundle> -d <directory> -pfn-flat
-
-There is, however, no good option for reversing this command when recreating the
-packages and bundle. Callers need to call makemsix pack on each individual
-package folder and then follow it with a makemsix bundle call to generate a
-flat bundle.
-
-### 2.2.1. Pack
-Usage:
----------------
-    makemsix.exe pack -d <directory> -p <package> [options]
-
-Options:
----------------
-    -d <directory> {Required}
-        Input directory path.
-    -p <package> {Required}
-        Output package file path.
-    -? [Flag]
-        Displays this help text.
-
-### 2.2.2. Bundle
-Usage:
----------------
-    makemsix.exe bundle -p <outputBundle> [options]
-
-Options:
----------------
-    -d <inputDirectory>
-        Input directory path.
-    -p <outputBundle> {Required}
-        Output bundle file path.
-    -f <mappingFile>
-        Mapping file path.
-    -bv <version>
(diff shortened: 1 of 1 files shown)


specs/WinRT/MakeMSIX-spec.md
```diff
@@ -0,0 +1,640 @@
...
+For more details see:
+
+-   [The makeappx command line tool](https://learn.microsoft.com/en-us/windows/win32/appxpkg/make-appx-package--makeappx-exe-)
+    Shipped as part of the Windows Development Kit.
+    Exposes pack, unpack, bundle, and unbundle commands, as well as encryption support
+    via a custom key-file list file, and mapping .
+
+-   [The makemsix command line tool and dll](https://github.com/Microsoft/msix-packaging)
+    Available through the github project.
+    Exposes pack, unpack, unbundle, and bundle* commands (*Bundle command is not implemented).
```
DrusTheAxe, line 17: msix-packaging's makemsix has a bundle command but it does nothing?

sreading-MSFT: Well, I spell it out more clearly later, maybe this is overstating the issue. Only flat bundles are supported. Maybe that's what everyone uses now? The failure when trying to create a non-flat bundle is definitely not well documented by the tool, it's just an E_NOTIMPL failure in code that gets ignored and then the user sees a failure message "The bundle must contain at least one app package targeting a known processor architecture." because it never actually processed any of the payload packages:
https://github.com/microsoft/msix-packaging/blob/master/src/msix/pack/AppxBundleWriter.cpp#L117

wjk:
Why, exactly, are full MSIX bundles not supported here? This makes it impossible to offer a single file to download containing every architecture supported by the application, or to deploy the App SDK runtime MSIX package alongside an application that uses it, which is rather important.

As an aside, as far as I am concerned, MSIX is a useless technology outside of the Windows Store. The code signing requirement makes it impossible for solo developers like me to distribute MSIX packages outside of the Windows Store (because, no, I cannot “just” purchase a certificate because they are so expensive, not to mention the fact that I would have to form an LLC before the CAs would sell me one because non-EV code-signing certificates are for some reason not available). Furthermore, it is impossible for an MSIX packaged application to be installed for all users on a computer, like all other Windows software (and, no, provisioning the package for all users is not an option because [a] there is no way to do it outside of manual PowerShell commands and [b] if the user were to delete the MSIX file after installing it, which would be a logical thing for them to do, the provisioning operation would then fail). And even then it is far too easy to accidentally uninstall the program, after which it will not be re-provisioned.

As much as I support the Windows Store — and I support it a lot — it is not a useful distribution method for serious applications. Since Microsoft has killed off the Store for Business, it is now impossible to download an app except by tying it to an individual’s consumer Microsoft account. Worse, the `Windows.Services.Store` API is a complete joke. There is no way to debug the use of these APIs without completing and publishing your application, having it go through App Review — and then you have to buy a copy of your own app with your own credit card (and this is somehow a deliberate design decision). I always liked the older `Windows.ApplicationModel.Store` API because it is very easy to test through the use of the `CurrentAppSimulator` class. However, that API requires CoreWindow, which is now very deprecated. As a result, developers who distribute their apps on the Windows Store have to handle their own billing and licensing, which negates the main advantage of a centralized app store. Payments should be handled by the platform.

DrusTheAxe:
> Why, exactly, are full MSIX bundles not supported here?

We're lighting up some new functionality and there's no change to bundles needed for it, hence the reason it's out of scope for this tool at this time. While I'd like to see this tool expand to encompass more functionality in time (including bundles) there's only so many hours in a day and you've gotta crawl before you walk before you run. So our focus here is on gaps current tools don't address.

Existing tools will continue to work just fine for bundles e.g. `makeappx.exe`. If there's issues or enhancements you'd like to see addressed please file an Issue so we can track the ask and get it onto our radar. Thank you.

DrusTheAxe:
>The code signing requirement

Are you aware of [AddPackageOptions.AllowUnsigned](https://learn.microsoft.com/uwp/api/windows.management.deployment.addpackageoptions.allowunsigned?view=winrt-22621#windows-management-deployment-addpackageoptions-allowunsigned)?

If you make an unsigned package (i.e. just run `makeappx.exe`, no `signtool.exe`) you can install it via [PackageManager.AddPackageByUriAsync](https://learn.microsoft.com/uwp/api/windows.management.deployment.packagemanager.addpackagebyuriasync?view=winrt-22621) introduced in 10.0.19041.0 (aka 20H1) e.g.

```
packageManager.AddPackageByUriAsync("file://c:/foo/bar.msix", new AddPackageOptions(){ AllowUnsigned=true })
```

You can also accomplish this via the Powershell cmdlet `Add-AppxPackage -AllowUnsigned ...`

>no, provisioning the package for all users is not an option because [a] there is no way to do it outside of manual PowerShell commands

There are multiple ways I'm aware of you can provision a package

1. PackageManager.[ProvisionPackageForAllUsersAsync()](https://learn.microsoft.com/uwp/api/windows.management.deployment.packagemanager.provisionpackageforallusersasync?view=winrt-22621)
2. Powershell cmdlet `Add-AppxProvisionedPackage`
3. DISM.exe [/Add-ProvisionedAppxPackage](https://learn.microsoft.com/windows-hardware/manufacture/desktop/dism-app-package--appx-or-appxbundle--servicing-command-line-options?view=windows-11#add-provisionedappxpackage)
4. Intune
5. [MSIX Hero](https://github.com/marcinotorowski/MSIX-Hero)

To name a few. There's probably other ways out there than I know :-)

Provisioning a package affects more than just the current user so provisioning (regardless of how) requires the caller have admin privilege. The `Remarks` section on the [ProvisionPackageForAllUsersAsync()](https://learn.microsoft.com/uwp/api/windows.management.deployment.packagemanager.provisionpackageforallusersasync?view=winrt-22621) page has more details.

>[b] if the user were to delete the MSIX file after installing it...the provisioning operation would then fail)

This is not how provisioning works. When you provision a package it's staged on the system e.g. `C:\Program Files\WindowsApps\LolzCatzVidz_1.2.3.4_x64__1234567890abc\...`. That's locked down and managed by the Deployment stack; a normal user can't delete that (and if you're elevated you've already pwned the system).

Once you've successfully provisioned a package the package source is no longer needed. If you provisioned a package with a `file:` URL using a physically local .msix file then you can delete the .msix file and it doesn't matter. The package directory (`C:\Program Files\WindowsApps\...`) is unaffected. Likewise if you provisioned a package with an `http:` or other network URL once provisioned you don't need network connectivity to that source to use the provisioned package.

>And even then it is far too easy to accidentally uninstall the program, after which it will not be re-provisioned.

To be pedantic, when you login after packages are provisioned packages (or a user's 1st login) all provisioned packages are registered for they user. The user can "uninstall" the package but that's just deregistering it for the user. The package is still present on the system (you have to deregister a package for all users AND deprovision the package to actually remove it from the system). If the user e.g. right-clicks on a provisioned package like Calculator and says Uninstall the package is deregistered for the user (StartMenu items, file type associations, etc all gone). After all the user is in control of their experience. Do you find this user experience is "too easy"? In what ways?

This is well off the beaten path for this PR so if you have additional feedback or questions please create an Issue or Discussion for follow up. Thank you.

specs/WinRT/MakeMSIX-spec.md
```diff
@@ -0,0 +1,959 @@
...
+    /// <summary>
+    /// Static methods for creating and unpacking packages.
+    /// </summary>
+    [contract(MakeMSIXContract, 1)]
+    runtimeclass MakeMSIXManager
+    {
+        /// <summary>
+        /// Creates packages from the directoryPathToPack as specified by the packOptions.
+        /// </summary>
+        static Windows.Foundation.IAsyncAction Pack(String directoryPathToPack, PackOptions packOptions);
```
DrusTheAxe, line 922: SUGGESTION: rename `directoryPathToPack` to `inputPath`

specs/WinRT/MakeMSIX-spec.md
```diff
@@ -0,0 +1,959 @@
...
+        /// <summary>
+        /// Creates bundles from the directoryPathToBundle as specified by the bundleOptions.
+        /// </summary>
+        static Windows.Foundation.IAsyncAction Bundle(String directoryPathToBundle, BundleOptions bundleOptions);
```
DrusTheAxe, line 926: SUGGESTION: rename `directoryPathToBundle` to `inputPath`

specs/WinRT/MakeMSIX-spec.md
```diff
@@ -0,0 +1,959 @@
...
+        /// <summary>
+        /// Unpacks a package at packageFilePathToUnpack as specified by the unpackOptions.
+        /// </summary>
+        static Windows.Foundation.IAsyncAction Unpack(String packageFilePathToUnpack, UnpackOptions unpackOptions);
```
DrusTheAxe, line 930: SUGGESTION: rename `packageFilePathToUnpack` to `inputFileName`

These functions' 1st parameter tries to convey 2 pieces of information

1. This is the input or source file
2. This is a package/bundle

Which leads to somewhat complex and confusing naming.

SUGGESTION: Drop #2 and just name the input parameter something simple and unambiguous, e.g. `inputFileName`

DrusTheAxe, line 930: All these take 1 required input parameter and 1 required Options object but you don't have to set any of these properties in the Options objects. Is that distinction useful?

What's the minimum required parameters for these APIs? It's hard to imagine just 1 input filename is required, but no output filename? What's required to use these e.g. Pack = inputPath + outputFileName(OrPath if LooseFile)

Implies these APIs should take 3 parameters: input, output, options

specs/WinRT/MakeMSIX-spec.md
```diff
@@ -0,0 +1,959 @@
...
+```c#
+// Copyright (c) Microsoft Corporation and Contributors.
+// Licensed under the MIT License.
+
+namespace Microsoft.Kozani.MakeMSIX
+{
+    [contractversion(1)]
+    apicontract MakeMSIXContract{};
+
+    /// Options for creating a package
```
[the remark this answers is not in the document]
DrusTheAxe: Ohhhhhh. PackOptions are options for the Pack() command to create a package vs CreateKozaniPackageOptions are options to the ??? command to create a LocalVariant package. This was confusing to me (think about that statement..) but OK.

Q: What options differ? Is creating a LocalVariant package the same as creating a package plus additional options?

What options differ?

If there's `Pack(input, packOptions)` would it make more sense to have `CreateKozani(input, packOptions, createKozaniOptions)`? If CreateKozani == Pack + additional behavior and options...

specs/WinRT/MakeMSIX-spec.md
```diff
@@ -0,0 +1,959 @@
...
+        static Windows.Foundation.IAsyncAction Bundle(String directoryPathToBundle, BundleOptions bundleOptions);
+        /// <summary>
+        /// Unpacks a package at packageFilePathToUnpack as specified by the unpackOptions.
+        /// </summary>
+        static Windows.Foundation.IAsyncAction Unpack(String packageFilePathToUnpack, UnpackOptions unpackOptions);
+        /// <summary>
+        /// Unbundles a package at bundleFilePathToUnbundle as specified by the unbundleOptions.
+        /// </summary>
+        static Windows.Foundation.IAsyncAction Unbundle(
+            String bundleFilePathToUnbundle,
```
DrusTheAxe, line 935: SUGGESTION: rename `bundleFilePathToUnbundle` to `inputFileName`

specs/WinRT/MakeMSIX-spec.md
```diff
@@ -0,0 +1,959 @@
...
+        static Windows.Foundation.IAsyncAction CreateKozaniPackage(
+            String packageFilePathToConvert,
+            CreateKozaniPackageOptions createKozaniPackageOptions);
+
+        /// <summary>
+        /// Creates a mountable image that contains the packages at packageFilePathsToAdd.
+        /// Can be used for Azure App Attach.
+        /// </summary>
+        static Windows.Foundation.IAsyncAction CreateMountableImage(
+            Windows.Foundation.Collections.IVector<String> packageFilePathsToAdd,
```
DrusTheAxe, line 951: SUGGESTION: rename `packageFilePathsToAdd` to `inputFileNames`

specs/WinRT/MakeMSIX-spec.md
```diff
@@ -0,0 +1,959 @@
...
+        /// If false, the operation will fail if a file already exists.
+        /// </summary>
+        Boolean OverwriteFiles{ get; set; };
+    };
+
+    /// <summary>
+    /// 
+    /// </summary>
+    [contract(MakeMSIXContract, 1)]
+    runtimeclass CreateMountableImageOptions
```
DrusTheAxe, line 886: Missing property?

```
enum MountableImageFormat
{
    Auto,
    CIM,
    VHD,
    VHDX,
}
MoutableImageFormat Format;
```

see earlier comment

specs/WinRT/MakeMSIX-spec.md
DrusTheAxe, line 922: Naming

`CreateKozaniPackage()` to create a Kozani package from an existing package? But `Pack()`, not `CreatePackage()`?

The usual WinRT API naming for APIs works better, even if terms differ from tool command line syntax e.g.

* pack = CreatePackage()
* bundle = CreateBundle()
* unpack = ExtractPackage()
* unbundle = ExtractBundle()
* ??? = CreateKozaniLocalVariant() or maybe just CreateKozani()
* ??? = CreateMountableImage()

P.S. .NET calls it ZipFile.CreateFromDirectory() and ZipFile.ExtractToDirectory() hence the Create+Extract verbs here.

DrusTheAxe (changes requested):

specs/WinRT/MakeMSIX-spec.md
```diff
@@ -0,0 +1,959 @@
...
+        /// Output path for the image file.
+        /// Valid file extensions are vhd, vhdx, and CIM.
+        /// </summary>
+        String ImageFilePath{ get; set; };
+
+        /// <summary>
+        /// Root directory path for packages inside the image.
+        /// Defaults to "WindowsApps"
+        /// </summary>
+        String PackageRootDirectoryInImage{ get; set; };
```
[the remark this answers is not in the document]
sreading-MSFT: This is a path that gets created within the OuputFile mountable image. So, if you mount the OutputFile you'd see like M:\WindowsApps or M:\<PackageRootDirectoryInImage>. Let me know what was confusing about the comments. Was just taken from msixmgr docs, but can do better in these docs.

specs/WinRT/MakeMSIX-spec.md
```diff
@@ -0,0 +1,959 @@
...
+        /// <summary>
+        /// Output path for the kozani packaged file.
+        /// Format must match input file. If packageFilePathToConvert is an appx or msix, then PackageFilePath must be an appx or msix.
+        /// If packageFilePathToConvert is an appxbundle or msixbundle, then PackageFilePath must be an appxbundle or msixbundle.
+        /// </summary>
+        String PackageFilePath{ get; set; };
+
+        /// <summary>
+        /// Optional replacement package publisher.
+        /// If not set, publisher from appxmanifest.xml is unchaged.
```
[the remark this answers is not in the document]
sreading-MSFT: Yes.

specs/WinRT/MakeMSIX-spec.md
```diff
@@ -0,0 +1,959 @@
...
+        /// <summary>
+        /// Output path for the full packaged file.
+        /// </summary>
+        String PackageFilePath{ get; set; };
+
+        /// <summary>
+        /// Overwrite PackageFilePath if it already exists.
+        /// Defaults to true.
+        /// </summary>
+        Boolean OverwriteFiles{ get; set; };
```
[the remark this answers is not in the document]
sreading-MSFT: I understand how Prompt would work in the context of makeappx or any command line tool, but not sure how it would work for a winrt API. It seems reasonable to let the caller check the filesystem for existence of the file first if they care. Barring that, the api prototype currently checks for existence of the output file up front before doing any work and returns HRESULT(ERROR_FILE_EXISTS) when the file exists and Overwrite is false, which the caller could use to show a prompt.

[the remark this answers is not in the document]
sreading-MSFT: Are there other apis in WindowsAppSDK that pop UI up? I haven't looked at what it would take to localize ui and make that work in the WindowsAppSDK context.

specs/WinRT/MakeMSIX-spec.md
```diff
@@ -0,0 +1,959 @@
...
+        /// Format must match input file. If packageFilePathToConvert is an appx or msix, then PackageFilePath must be an appx or msix.
+        /// If packageFilePathToConvert is an appxbundle or msixbundle, then PackageFilePath must be an appxbundle or msixbundle.
+        /// </summary>
+        String PackageFilePath{ get; set; };
+
+        /// <summary>
+        /// Optional replacement package publisher.
+        /// If not set, publisher from appxmanifest.xml is unchaged.
+        /// </summary>
+        String PackagePublisher{ get; set; };
```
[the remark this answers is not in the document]
sreading-MSFT: Agreed. Upon implementing the prototype, I realized I do need this packageidentity class because it would be incredibly annoying to have an api that lets you change the publisher name but no api that reads the publisher name from a package. Have added GetPackageIdentity as well.

specs/WinRT/MakeMSIX-spec.md
```diff
@@ -0,0 +1,959 @@
...
+
+        /// <summary>
+        /// Optional replacement package name.
+        /// If not set, name from appxmanifest.xml is unchaged.
+        /// </summary>
+        String PackageName{ get; set; };
```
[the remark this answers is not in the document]
sreading-MSFT: Yeah, will add.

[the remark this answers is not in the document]
DrusTheAxe: Yes. In fact, most everything in package identity should be overrideable. When I create a LocalVariant package I might decide there's nothing in the LocalVariant that's architecture-dependent so the servers need x86/x64/arm64 but the LocalVariant can be neutral.

The LocalVariant is a variant of another package. If you can provide alternative parts of identity I don't see why not all of it. Overrides for all 5 parts of the identity tuple - name, publisher, version, architecture, resourceid

specs/WinRT/MakeMSIX-spec.md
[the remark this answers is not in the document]
DrusTheAxe: Ahhhh. I didn't realize that.

That makes more sense now. Ignore my comment

specs/WinRT/MakeMSIX-spec.md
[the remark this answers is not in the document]
DrusTheAxe: Ahhh. My bad. I mixed up the API and the wrapper tool.

Prompting is entirely a tool thing. The API is a boolean. Any prompting should be handled by the tool e.g.

MAKEMSIX PACK --Overwrite=Prompt... == IF file exists Prompt for Y/N then call API with options.OverwriteFileIfExists=Y/N

specs/WinRT/MakeMSIX-spec.md
sreading-MSFT: Good point, I like that a lot better and will change. I think this is probably a remnant of having earlier versions of this api where things were grouped together more and the minimal set of options was less easy to discover. Agree that making that clear is a primary concern.

specs/WinRT/MakeMSIX-spec.md
```diff
@@ -0,0 +1,959 @@
...
+order to create an image of the correct size. This behavior is more consistent with other types of package creation and eliminates the api differences
+between CIM files and VHD\VHDX files. 
+
+Since the api surface supports creation of the CIM, VHD, VHDX file as well as creation of msix and msixbundle files, callers may
+find it strange to need to first pack their directory to an msix, and then create the vhdx from the msix, rather than just
+creating the vhdx directly. However, signing is not integrated into any of these tools, and is a separate step that
+is done after packaging is complete. The Appx SIP knows how to sign packages, but not vhdx and other mounted directories. The result
+is that creating a vhdx with an appx signature file (appxsignature.p7x) requires first creating the package, then signing it, then
+creating the vhdx. Allowing creation of a CIM, VHD, or VHDX file directly from a directory seems likely to cause confusion for 
+developers who may not realize that they won't be able to sign the package in the image. This api also proposes always applying acls
```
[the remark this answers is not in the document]
sreading-MSFT: The first one - no direct cim\vhd\vhdx from a package layout directory. My claim is that since other apis on the system do not recognize cim\vhd\vhdx as a package in and of itself, rather than just a container for packages, that it will be confusing to have this api treat those formats as if they're themselves a type of package just because it's possible to create a cim\vhd\vhdx with only one package. A layout directory is packed, then signed, then added to a cim\vhd\vhdx in separate steps.

specs/WinRT/MakeMSIX-spec.md
```diff
@@ -0,0 +1,959 @@
...
+        /// <summary>
+        /// Overwrite PackageFilePath if it already exists.
+        /// Defaults to true.
+        /// </summary>
+        Boolean OverwriteFiles{ get; set; };
+        /// <summary>
+        /// Validates elements of the package during creation.
+        /// Defaults to false.
+        /// </summary>
+        Boolean ValidateFiles{ get; set; };
```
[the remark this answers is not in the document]
sreading-MSFT: Version I intend to implement is "all files referenced by the manifest do actually exist". The windows api for package creation always do manifest validation at the level of verifying that the manifest reader can parse the xml and be initialized, e.g. find a package object. That can't be turned off.

specs/WinRT/MakeMSIX-spec.md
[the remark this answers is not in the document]
sreading-MSFT: Multiple comments dealing with roughly similar questions - choosing this one for response. The top level question I think is whether to merge kozani package creation into the other apis as an option, or to have it as a separate thing. As you noticed, the other apis all take an input of a specific format (i.e. CreatePackage requires a layout directory), an output that they always write in a specific format, an options object, and then perform one operation on the input file to produce the output.

Creation of Kozani packages could follow that model. A caller would call ExtractBundle, iterate through msix files and call ExtractPackage,  iterate directories and call CreatePackage with something like MakeKozaniPackage(true) and other options in CreatePackageOptions like RemoveExtensions(true), call CreateBundle with any Kozani specific options.

For Kozani though I think we know that by far the most common operation that people want to do is to convert from an existing package to Kozani. They may have a layout directory, or a package, or a bundle - and they just want that same thing in Kozani form. In some cases they may have large groups of packages, some of which are bundles and others of which are packages, that they want to convert. In this way, creation of a kozani package is more like an operation like AddPackage, where any type of package format (layout, package, bundle) can be handled internally, than it is like the other apis in this spec (like CreatePackage) where the input and output are always one format. 
Forcing people to unpack the entire package just to recreate it is a waste of processor time and disk space, and forcing them to special case bundles vs packages adds unnecessary code complexity for this one most common scenario. Of course those operations are all available if there's some custom manifest changes the caller wants to write.

In other words, since all the other apis transform an existing package or set of packages to some other format, whereas Kozani package creation is mostly intended to transform an existing package into a different version of the same package format, the only way to make Kozani package creation fit into the existing api space is to require callers to first transform the package out of the format they want, so that they can call another api with a kozani option to transform it back into the format they want.

Since I'm proposing having CreateKozaniPackage handle all package types internally, the equivalent option set is more like CreateKozaniOptions == CreatePackageOptions + CreateBundleOptions + additional kozani specific options (+ (potentially) CreateMountableImageOptions). Since these are still just packages I would mostly assume anything in the "additional kozani specific options" could be moved into either\both of CreatePackageOptions or CreateBundleOptions, just that some options are very unlikely to be useful for those operations and so seem likely to do nothing more than add confusion for people. As an example, in the update to the spec I just pushed I added "boolean RemoveExtensions" as field on the CreateKozaniPackageOptions. There is no inherent reason this couldn't be in CreatePackageOptions as well.

But I think the intention of having CreateKozaniPackage handle all types of packages internally does mean it makes sense to have its own options object. There may be cause to add the OutputFormat option you suggested in another comment, which wouldn't make sense for the other apis. Or there may be differences in behavior between operations using the same options. For example, since main packages could all have different versions and a bundle manifest isn't required for bundling, CreateBundle defaults to creating a bundle version based on timestamp if Version is not explicitly specified. But CreateKozaniPackage does require an existing package, so it makes more sense to maintain the version from the existing manifest if Version is not set.

specs/WinRT/MakeMSIX-spec.md
```diff
@@ -0,0 +1,959 @@
...
+        /// Overwrite PackageFilePath if it already exists.
+        /// Defaults to true.
+        /// </summary>
+        Boolean OverwriteFiles{ get; set; };
+
+        /// <summary>
+        /// Validates elements of the package during creation.
+        /// Defaults to false.
+        /// </summary>
+        Boolean ValidateFiles{ get; set; };
```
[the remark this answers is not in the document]
sreading-MSFT: The proposal right now is to just use extension\path to determine the input type, and it seems like input and output arg formats should be mirrored here if foo.bar being an MSIX file is a common scenario. Is there high-demand for creating and using files that don't match their typical extensions? I assume callers could easily rename the output afterwards, so this seems to affect input files more than output files.

specs/WinRT/MakeMSIX-spec.md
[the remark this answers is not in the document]
DrusTheAxe: > the intention of having CreateKozaniPackage handle all types of packages internally does mean it makes sense to have its own options object

+1

specs/WinRT/MakeMSIX-spec.md
```diff
@@ -0,0 +1,640 @@
...
+    /// Result of the packaging operation
+    /// </summary>
+    [contract(MakeMSIXContract, 1)]
+    enum PackagingResultStatus
+    {
+        Ok,
+        InvalidOptions,
+        FileAlreadyExistsError,
+        InternalError,
+    };
```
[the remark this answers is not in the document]
sreading-MSFT: Moved to IAsyncAction which uses hresults in its ErrorCode property

sreading-MSFT:

---
In reply to: [1472758116](https://github.com/microsoft/WindowsAppSDK/pull/3429#issuecomment-1472758116) [](http://example.com/codeflow?ancestors=1472758116)

specs/WinRT/MakeMSIX-spec.md
sreading-MSFT: Still not sure of benefit of this vs using file extensions. Not opposed, just not sure who wouldnt use Auto.

specs/WinRT/MakeMSIX-spec.md
```diff
@@ -0,0 +1,1002 @@
...
+
+The MakeMSIX API allows developers to create app packages for distribution. 
+The API provides a WinRT interface that supports:
+1. Creation of msix packages and msixbundle bundle packages,
+2. Unpacking and unbundling of packages
+3. Creation of vhd\vdx\cim images for Azure App Attach. 
+4. Creation of Kozani packages.
+
+Creation of Kozani packages is a new feature that has not existed in other msix packaging tools. The
+other features have previously been available in command line tools discussed in the appendix.
```
DrusTheAxe, line 34: SUGGESTION: Other features ***in this spec*** have previously...

specs/WinRT/MakeMSIX-spec.md
```diff
@@ -0,0 +1,1002 @@
...
+
+Kozani packages are msix packages that have been optimized for size by removing unnecessary
+resources. All existing tools that work with msix packages are compatible with Kozani packages.
+Kozani packages can be created manually without this API by modifying the files, manifest, and
+resource index of an existing package using existing win32 apis.
```
DrusTheAxe, line 39: SUGGESTION: ...using existing APIs.

specs/WinRT/MakeMSIX-spec.md
```diff
@@ -0,0 +1,1002 @@
...
+
+# 2. Examples
+
+## 2.1. Packing a package
+
+```cpp
```
DrusTheAxe, line 45: Samples should be written in C# unless there's reasons another language is needed (C++ specific details, not callable by C#, etc)

RECOMMEND: Change the samples to C# e.g.

```
void CreatePackagesFromFolderExample()
{
     // Scenario: ...
     ....
     string packageLayoutRootPath = "D:\\test\\ContosoApp1\\BuildOutput\\PackageLayouts";
    ...
}
```

specs/WinRT/MakeMSIX-spec.md
```diff
@@ -0,0 +1,1002 @@
...
+    HRESULT CreateProcessAndWaitForExitCode(std::wstring commandLine, DWORD& exitCode)
+    {
+        STARTUPINFO startupInfo;
+        ZeroMemory(&startupInfo, sizeof(startupInfo));
+        startupInfo.cb = sizeof(startupInfo);
+
+        PROCESS_INFORMATION processInformation;
+        ZeroMemory(&processInformation, sizeof(processInformation));
+
+        if (!CreateProcess(
```
DrusTheAxe, line 233: C#ize :-)

FYI C# `p = new Process(); ... p.Start(); p.WaitForExit();` is the C# equivalent to CreateProcess, WaitForSingleHandle, etc

specs/WinRT/MakeMSIX-spec.md
```diff
@@ -0,0 +1,1002 @@
...
+    apicontract MakeMSIXContract{};
+
+    /// Options for creating a package
+    [contract(MakeMSIXContract, 1)]
+    runtimeclass CreatePackageOptions
+    {
+        /// Created by callers to pass options into CreatePackage()
+        CreatePackageOptions();
+
+        /// Optional replacement package name.
```
DrusTheAxe, line 383: SUGGESTION: `/// Optional override for package name`

same comment applies to other replacement/override properties

specs/WinRT/MakeMSIX-spec.md
```diff
@@ -0,0 +1,1002 @@
...
+        /// If version is 0, value from AppxManifest.xml is unchanged.
+        Windows.ApplicationModel.PackageVersion Version;
+
+        /// Languages supported by the output package
+        /// If field is empty, all languages from the original package are maintained.
+        Windows.Foundation.Collections.IVector<String> Languages{ get; };
+
+        /// ScaleFactors supported by the output package
+        /// If field is empty, all scale factors from the original package are maintained.
+        Windows.Foundation.Collections.IVector<UInt32> ScaleFactors{ get; };
```
DrusTheAxe, line 433: This refers to e.g. `<Package><Resources><Resource Scale="150">`?

Are there other `<Resource>`s deserving the same support?

[Docs](https://learn.microsoft.com/uwp/schemas/appxpackage/uapmanifestschema/element-resource) note `<Resource>` also supports Language and DXFeatureLevel

specs/WinRT/MakeMSIX-spec.md
```diff
@@ -0,0 +1,1002 @@
...
+    };
+
+	/// Identity of a package
+    runtimeclass PackageIdentity
+    {
+        /// Name of the package
+        String Name{ get; };
+        
+        /// Publisher of the package
+        String Publisher{ get; };
```
DrusTheAxe, line 514: Add PublisherId ?

specs/WinRT/MakeMSIX-spec.md
```diff
@@ -0,0 +1,1002 @@
...
+    runtimeclass PackageInformation
+    {
+		/// Identity of the package
+        PackageIdentity Identity{ get; };
+
+        /// Languages supported by the package
+        Windows.Foundation.Collections.IVectorView<String> Languages{ get; };
+
+        /// ScaleFactors supported by the package
+        Windows.Foundation.Collections.IVectorView<UInt32> ScaleFactors{ get; };
```
DrusTheAxe, line 542: AppxManifest has a 3rd `<Resource>` property, `DXFeatureLevel`. Should be added?

https://learn.microsoft.com/uwp/schemas/appxpackage/uapmanifestschema/element-resource

specs/WinRT/MakeMSIX-spec.md
```diff
@@ -0,0 +1,1002 @@
...
+}
+```
+
+# 4. Appendix
+
+# 5. Existing interfaces and options
+
+The existing tools are:
+-   [Makeappx.exe](https://learn.microsoft.com/en-us/windows/win32/appxpkg/make-appx-package--makeappx-exe-)
+    Shipped in Windows' Platform SDK.
```
DrusTheAxe, line 607: SUGGESTION: Suggestion make these sub-bullets

See Preview -- indenting here just means "one paragraph, word wrap it for me"

e.g.

```
The existing tools are:
- [MakeAppx.exe] blahblah
  - Foo blah
  - Bar blah
  - Ack ack ack

DrusTheAxe (changes requested):

specs/WinRT/MakeMSIX-spec.md
sreading-MSFT: Thought this was unlikely to ever be relevant for CreateKozaniPackage, but yeah, for completeness better to just have it.

mominshaikhdevs:
@sreading-MSFT whats the latest update on this?

lauren-ciha:

PR #3459, opened 2023-02-23
Fix issue where scope_exit instances weren't being held in local variables.

loneursid:
scope_exit is not needed with activity based telemetry, besides the scope_exit constructs were not working as expected as they weren't held by a local variable.


pmpurifoy (approved):

danielayala94 (approved):
Thanks for fixing this!

:shipit:

Issue #3460, opened 2023-02-23
Visual Studio Templates 1.2.230217.4 manifest supported OS Windows 8.1


Balkoth:
### Describe the bug

I don't know if it always was like this but this entry

```xml
<compatibility xmlns="urn:schemas-microsoft-com:compatibility.v1">
  <application>
    <!--The ID below informs the system that this application is compatible with OS features first introduced in Windows 8. 
    For more info see https://docs.microsoft.com/windows/win32/sysinfo/targeting-your-application-at-windows-8-1 
    
    It is also necessary to support features in unpackaged applications, for example the custom titlebar implementation.-->
    <supportedOS Id="{4a2f28e3-53b9-4441-ba9c-d69d4a4a6e38}" />
  </application>
</compatibility>
```

seems wrong to me.

Should be:

```xml
<compatibility xmlns="urn:schemas-microsoft-com:compatibility.v1">
  <application>
    <supportedOS Id="{8e0f7a12-bfb3-4fe8-b9a5-48fd50a15a9a}" />
  </application>
</compatibility>
```

### Steps to reproduce the bug

Create a new packaged application throuth the template.

### NuGet package version

Windows App SDK 1.2.4: 1.2.230217.4

### Packaging type

Packaged (MSIX), Unpackaged

### Windows version

Windows 10 version 22H2 (19045, 2022 Update)

### IDE

Visual Studio 2022

castorix:
It is OK,  {4a2f28e3-53b9-4441-ba9c-d69d4a4a6e38}" is Windows 8 :

https://learn.microsoft.com/en-us/windows/win32/sysinfo/targeting-your-application-at-windows-8-1

Balkoth:
Isn't Windows 10 version 1809 the minimum requirement for Windows App SDK Apps?

JaiganeshKumaran:
Good that Microsoft still supports the best version of Windows ever made.

castorix:
> Isn't Windows 10 version 1809 the minimum requirement for Windows App SDK Apps?

You set Windows 8 for example to support layered child windows (Note at [Layered Windows](https://learn.microsoft.com/en-us/windows/win32/winmsg/window-features#layered-windows)), like the Caption window (_DRAG_BAR_WINDOW_CLASS_ class)

DarranRowe:
@Balkoth 
I understand what you mean, you would expect the Windows 10 compatibility declaration.

Balkoth:
[...]

PR #3493, opened 2023-03-09
Change supportedOS guid to Win10 and improve comment

bpulliam:
Fixes https://github.com/microsoft/WindowsAppSDK/issues/3460

@Balkoth makes a fair point. The supportedOS logic checks the maximum version, so the Windows 10 GUID will always include Windows 8 and for WinAppSDK will be more accurate.


codendone (approved):
:shipit:

PR #3522, opened 2023-03-20
Do not add xbf files to UpToDateCheckInput as they are output files.

jlaanstra:
Fixes ADO#43580700. We've tested this fix for the last two weeks inside Phone Link and have seen significant improvements in inner dev loop for xaml files and observed no other issues with this. /cc @bpulliam 

WARNING:


jlaanstra:
@bpulliam buildfailure doesn't look related to this change?

Scottj1s:
Hmm, build log says "StageAndPack was canceled"

/azp run

Scottj1s:
@rohanp-msft any concerns here?

rohanp-msft:
I think the change itself is fine. I'm thinking about the duplicate logic in `_GenerateProjectPriConfigurationFiles`. I'll reply soon for that. Also, FYI @loneursid (ownership of MRT has been transferred).

---
In reply to: [1480287836](https://github.com/microsoft/WindowsAppSDK/pull/3522#issuecomment-1480287836) [](http://example.com/codeflow?ancestors=1480287836,1480287836)

jlaanstra:
> I think the change itself is fine. I'm thinking about the duplicate logic in `_GenerateProjectPriConfigurationFiles`. I'll reply soon for that. Also, since ownership of MRT has been transferred, FYI @loneursid.
> 
> In reply to: [1480287836](https://github.com/microsoft/WindowsAppSDK/pull/3522#issuecomment-1480287836)

You still want to reference the xbf for the pri file I think?

rohanp-msft:
> > I think the change itself is fine. I'm thinking about the duplicate logic in `_GenerateProjectPriConfigurationFiles`. I'll reply soon for that. Also, since ownership of MRT has been transferred, FYI @loneursid.
> > In reply to: [1480287836](https://github.com/microsoft/WindowsAppSDK/pull/3522#issuecomment-1480287836)
> 
> You still want to reference the xbf for the pri file I think?

Yeah. I think we should keep _GenerateProjectPriConfigurationFiles as is.

rohanp-msft:
@jlaanstra Any thoughts on why there is perf improvement with this change? If there is a change in an XBF, I would think there would be a corresponding change in the corresponding XAML file. In that case, the incremental build behavior will be the same if we used XAML files instead of XBF files in `UpToDateCheckInput`.

jlaanstra:
> @jlaanstra Any thoughts on why there is perf improvement with this change? If there is a change in an XBF, I would think there would be a corresponding change in the corresponding XAML file. In that case, the incremental build behavior will be the same if we used XAML files instead of XBF files in `UpToDateCheckInput`.

There is no perf improvements to a single build. However, without this change VS will always consider every subsequent build out of date as the xbf file was marked as an input file, but it was modified after the build started.

rohanp-msft (approved):
:shipit:

jlaanstra:
@Scottj1s let's merge this?

Scottj1s (approved):

rohanp-msft:
FYI @loneursid

PR #3062, opened 2022-10-17
Update pipelines to use BuildTransportPackage.ps1 

kythant:
Refactor pipelines to use BuildTransportPackage.ps1. Currently the yml and the ps1 does exactly the same thing so it is important that we need to consolidate them. Much of the BuildTransportPackage.ps1 needed refactoring so that the pipeline can use it in steps. Running the BuildTransportPackage.ps1 in the innerloop will run all the steps by default. 

- Updated CI and BuildFoundation to use Powershell Script
- Enable Helix Tests in BuildFoundation.yml if branch is develop
- Move DevCheck to the root directory
- Deleted much of the unused AzurePipelines yml templates or ones that are outdated by the script

PR #3529, opened 2023-03-23
Re-enable PREFast scanning in build pipeline

alexlamtest:
**The Problem:**
The PR [Update pipelines to use BuildTransportPackage.ps1 (#3062) · microsoft/WindowsAppSDK@8224ec5 (github.com)](https://github.com/microsoft/WindowsAppSDK/commit/8224ec5c7d73ea2c90ddac4350c594c0f60c993b) in late 2022 changed the way the build pipeline builds projects from using the VSBuild/MSBuild tasks, to invoking MSBuild.exe directly via BuildAll.ps1. This move is beneficial in the sense that it allows the projects to be built in the pipeline and on developers' local machines with the same code, BuildAll.ps1.
Unfortunately, that PR blocked PREFast scans, because PREFast's tasks only supports building the projects with the VSBuild/MSBuild tasks, and directly calling MSBuild.exe is unsupported.

**Proposed Solution:**
To maximally preserve a converged code path for building the projects both in the pipeline and on local dev machines, while allowing PREFast to run, we continue to pivot on using BuildAll.ps1 for setting up the build environment. When a pipeline run does _not_ intend to trigger a PREFast scan, we keep the existing flow unchanged. OTOH, when a pipeline run intends to trigger a PREFast scan, do not use the call to MSBuild.exe from BuildAll.ps1 to build the projects, but invoke the VSBuild/MSBuild tasks to perform the build operation instead, the same way as before the linked PR above.
As part of this effort, a couple of scripts are updated/fixed.

Although the above approach did unblock PREFast to run again, the PREFast scan itself is currently hitting an error in some scenarios. This problem is being tracked separately by ADO Bug 43856157, because the Release/1.2-Stable, which is unrelated to this PR, is currently hitting the same error. Therefore, in this PR, PREFast is not fully enabled to avoid hitting those errors.
But even the partially enabled PREFast scans have merits, we already received meaningful bugs auto filed by TSA to track the PREFast issues found.

**How built:**
- Changes will go through PR validation.

**How tested:**
- The sample run below succeeded, with PREFast and other Binary Analysis scans enabled. You can see PREFast results (for arm64) under the Scans tab:
https://dev.azure.com/microsoft/ProjectReunion/_build/results?buildId=66952864&view=results
- When PREFast is disabled, but other Binary Analysis scans are enabled, the sample pipeline run succeeded:
https://dev.azure.com/microsoft/ProjectReunion/_build/results?buildId=66828516&view=results
- This is a successful sample pipeline run with all Binary Analysis scans are disabled (baseline scenario):
https://dev.azure.com/microsoft/ProjectReunion/_build/results?buildId=66828409&view=results
----

WARNING:


build/AzurePipelinesTemplates/WindowsAppSDK-BuildFoundation-job.yml
```diff
@@ -19,10 +19,14 @@ parameters:
     type: boolean
     default: true
   - name: "enableLicenseInstall"
+    type: boolean
     default: true
   - name: "isWeeklyRun"
     type: boolean
     default: False
+  - name: "solutionPath"
```
kythant, line 27: Do we need this to be a parameter? Will ever be anything other than WindowsAppRuntime.sln?

kythant (approved):

build/AzurePipelinesTemplates/WindowsAppSDK-BuildFoundation-job.yml
alexlamtest: This is a hold-over from release/1.2-stable/build/ProjectReunion-BuildFoundation.yml and WindowsAppSDK-BuildProject-Steps.yml. As you can see below, it was used for other stuff back then. Hence, I'm deferring the cleanup of this parameter in case we need to lean further back to the old 1.2-stable pattern to unblock PREFast.

  - template: AzurePipelinesTemplates\WindowsAppSDK-BuildProject-Steps.yml
    parameters:
      solutionPath: 'dev\Bootstrap\CS\Microsoft.WindowsAppRuntime.Bootstrap.Net\Microsoft.WindowsAppRuntime.Bootstrap.Net.csproj'

Issue #3201, opened 2022-12-01
WinUI templates should not show up in Blend


mgoertz-msft:
### Describe the bug

VS and Blend do not have a designer for WinUI, so the WinUI templates should not be available in Blend, which is primarily for design experiences.

### Steps to reproduce the bug

1. Install Visual Studio and add the .NET Desktop and UWP workloads (make sure the Windows App SDK C# templates are selected in the optional components for the .NET Desktop workload)
2. Launch Blend from the Start menu
3. Click Create new project

WinUI templates show up with unfriendly titles and descriptions.

### Expected behavior

WinUI templates should not even be available in Blend, since there is no designer for WinUI.

### Screenshots

![image](https://user-images.githubusercontent.com/14282894/205119144-794d664f-fa28-446f-b90c-7401ea6e5de5.png)

### NuGet package version

None

### Windows version

Windows 11 version 21H2 (22000)

### IDE

Visual Studio 2022-preview

MSLukeWest:
The likely fix is to just remove this line from the .vstemplate files:

`<AppIdFilter>blend</AppIdFilter>`

MSLukeWest:
@evelynwu-msft This change seems to have made it into the latest 1.4 templates, as I no longer see these templates in Blend.  So, this issue can probably be closed.

PR #3557, opened 2023-04-03
Updating Item and Project templates to make them not appear in VS Blend

MSLukeWest:
Fixing #3201 

Updating WinUI Visual Studio project and item templates to make them no longer appear in Blend.  VS and Blend do not have a designer for WinUI, so the WinUI templates should not be available in Blend, which is primarily for design experiences.

The fix is to remove this line from each .vstemplate file:

`<AppIdFilter>blend</AppIdFilter>`


MSLukeWest:
FYI @evelynwu-msft.  Can you take a look and help me get this submitted?

evelynwu-msft (approved):

evelynwu-msft:
Thanks for making this change!

PR #3571, opened 2023-04-13
Disable BuildMRT in public CI builds

kythant:
Disable BuildMRT in public CI builds

Microsoft.Taef's version update won't be publicly available for the time being


codendone (approved):
:shipit:

PR #3599, opened 2023-04-28
Fix: FAIL_FAST_FATAL_APP_EXIT_c0000409_Microsoft.Windows.ApplicationModel.Resources.dll!winrt::com_array_winrt::hstring_::clear

loneursid:
Fix for [Watson Failure] caused by FAIL_FAST_FATAL_APP_EXIT_c0000409_Microsoft.Windows.ApplicationModel.Resources.dll!winrt::com_array_winrt::hstring_::clear

Consensus is that this error is likely due to concurrent access of class ResourceContext. I've added locks to all public methods to ensure there's no race.

WARNING:


dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/src/Microsoft.Windows.ApplicationModel.Resources.vcxproj
```diff
@@ -171,6 +171,7 @@
...
     <Import Project="$(NugetPackageDirectory)\Microsoft.Build.Tasks.Git.$(MicrosoftBuildTasksGitVersion)\build\Microsoft.Build.Tasks.Git.targets" Condition="Exists('$(NugetPackageDirectory)\Microsoft.Build.Tasks.Git.$(MicrosoftBuildTasksGitVersion)\build\Microsoft.Build.Tasks.Git.targets')" />
     <Import Project="$(NugetPackageDirectory)\Microsoft.SourceLink.GitHub.$(MicrosoftSourceLinkGitHubVersion)\build\Microsoft.SourceLink.GitHub.targets" Condition="Exists('$(NugetPackageDirectory)\Microsoft.SourceLink.GitHub.$(MicrosoftSourceLinkGitHubVersion)\build\Microsoft.SourceLink.GitHub.targets')" />
+    <Import Project="..\..\..\..\..\Packages\Microsoft.Windows.ImplementationLibrary.1.0.220914.1\build\native\Microsoft.Windows.ImplementationLibrary.targets" Condition="Exists('..\..\..\..\..\Packages\Microsoft.Windows.ImplementationLibrary.1.0.220914.1\build\native\Microsoft.Windows.ImplementationLibrary.targets')" />
```
huichen123, line 174: >..\..\..\..\..\Packages [](http://example.com/codeflow?start=21&length=23)

NugetPackageDirectory

dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/src/ResourceContext.cpp
```diff
@@ -1,15 +1,76 @@
...
+ResourceContext::~ResourceContext()
+{
+    winrt::slim_lock_guard guard {m_lock};
+    MrmDestroyResourceContext(m_resourceContext);
+}
+
+
+winrt::Windows::Foundation::Collections::IMap<hstring, hstring> ResourceContext::QualifierValues()
+{
+    m_lock.lock_shared();
```
huichen123, line 24: >m_lock.lock_shared(); [](http://example.com/codeflow?start=4&length=21)

is this needed?

dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/src/ResourceContext.cpp
```diff
@@ -1,15 +1,76 @@
...
+        if (!eachValue.Value().empty())
+        {
+            winrt::check_hresult(MrmSetQualifier(m_resourceContext, eachValue.Key().c_str(), eachValue.Value().c_str()));
+        }
+    }
+}
+
+MrmContextHandle ResourceContext::GetContextHandle()
+{
+    winrt::slim_shared_lock_guard guard {m_lock};
```
huichen123, line 65: >winrt::slim_shared_lock_guard guard {m_lock}; [](http://example.com/codeflow?start=4&length=45)

is this needed?

dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/src/Microsoft.Windows.ApplicationModel.Resources.vcxproj
loneursid: It's strange that we have to manually edit these files. I would have expected the Nuget Manager to use the project constansts automatically.

dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/src/ResourceContext.cpp
loneursid: Because there's a possibility the dtor get called on a separate thread.

dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/src/ResourceContext.cpp
loneursid: Unless we move initialization to the ctor, in which case we loose the benifits of lazy initialization without this guard we're at risk of having a race.

huichen123: would this be sufficient?
if (!IsInitialized())
    {
        m_lock.lock();

        if (!IsInitialized())
        InitializeQualifierValueMap();

        m_lock.unlock();

    }

    return m_qualifierValueMap;

loneursid: no, becasue you can't garantee that !IsInitialized() is atomic (I mean, even if we could make the check with a single operation, instead of calling a function, we would also have to make it atomic becasue of caches, etc.). Since the operation isn't atomic, the only way to avoid potential races is to take a lock on it.

dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/src/ResourceContext.cpp
huichen123: can dtor be called when you still have a reference?

loneursid: Can the app obtains a reference and then call the dtor? Sure, and there's nothing that can be done about that, at least not in the PR.

The caller gets a MrmContextHandle which is defined be some macros. I'm not familiar with how the type is defined but it doesn't seem to be ref counted, so whenever the dtor is called, the handle will be released and all references handed out will become dangling.

Introducing some sort of reference counting or other mechanism to cope with this is beyond the scope of this fix.

At least now we garantee that the dtor won't race with GetContextHandle.

huichen123: i think winrt object is already refcounted, and you don't have to worry about someone else calling dtor when you still hold a ref.

loneursid: Cool, thanks for confirming.

dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/src/ResourceContext.cpp
huichen123: then can you do the lock inside IsInitialized()? also create lock object, and there is no need to introduce wil.

loneursid: what is the issue with wil?

dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/src/ResourceContext.cpp
```diff
@@ -1,15 +1,76 @@
...
+    winrt::slim_lock_guard guard {m_lock};
+    MrmDestroyResourceContext(m_resourceContext);
+}
+
+
+winrt::Windows::Foundation::Collections::IMap<hstring, hstring> ResourceContext::QualifierValues()
+{
+    m_lock.lock_shared();
+
+    auto cleanupOnFailure = wil::scope_exit([&] { m_lock.unlock_shared(); });
```
jonwis, line 26: FWIW, STL has https://en.cppreference.com/w/cpp/thread/call_once that makes this pattern easier.

There's no issue with WIL, of course. Other components in this repo use it just fine.

dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/src/ResourceContext.cpp
```diff
@@ -1,15 +1,76 @@
...
+#include <wil\resource.h>
 
 const wchar_t c_languageQualifierName[] = L"Language";
 
 namespace winrt::Microsoft::Windows::ApplicationModel::Resources::implementation
 {
+
+ResourceContext::~ResourceContext()
+{
+    winrt::slim_lock_guard guard {m_lock};
```
jonwis, line 17: Per comments below, this lock shouldn't be necessary - there should be no more references to the object when it's having its final release handled.

dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/src/ResourceContext.cpp
loneursid: Done, now using call_once, thanks

dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/src/ResourceContext.cpp
loneursid: removed

dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/src/ResourceContext.cpp
loneursid: All this is gone, now using call_once.

huichen123 (approved):
:shipit:

Issue #1780, opened 2021-11-12
Win32 Desktop functions not available in WinUI 3 Runtime Component.


timmarriott:
### Describe the bug

Attempting to bring up a FilePicker in from a Windows Runtime Component which was started using the WinUI 3 template.   When using #include <shobjidl_core.h> I get a large number of compile issues.  It seems that the Component is not a Desktop component so large parts of the header are #ifdef'ed out.

### Steps to reproduce the bug

Create a Windows Runtime Component with a Page and a button in which the handler brings up a FileOpenPicker, and reference it in a Blank Page App.  The component has the following code:

#include <shobjidl_core.h>
#include <winrt/Windows.Storage>

Windows::Foundation::IAsyncAction PopupPickerAsync()
{
    auto picker = Windows::Storage::Pickers::FileOpenPicker();
  
    picker.as<IInitializeWithWindow>()->Initialize(hwnd);  // This was hard to find for C++  hwnd is in a static singleton in my app
 
    .....
    Windows::Storage::StorageFile infile = co_await picker.PickSingleFileAsync();j
}

### NuGet package version

1.0.0-preview3

### Packaging type

Packaged (MSIX)

### Windows version

Windows 11 (22000)

### IDE

Visual Studio 2022

andrewleader:
@evelynwu-msft is this a similar "Desktop APIs not available from C++" bug in 1.0 Preview 3? In this case it seems it's from a WinUI 3 Runtime Component (a library) rather than the app itself, but is it the same bug?

evelynwu-msft:
@andrewleader Similar, but it's a different bug. I suspect that the change we made ages ago to allow C++ Windows Runtime Components to be useable by Desktop WinUI apps (by setting `<DesktopCompatible>true</DesktopCompatible>`) doesn't actually lift the UWP-only API restriction.

Jaiganeshkumaran:
The description for the Windows Runtime Component (WinUI 3) C++ project template says that it can be used in both UWP and desktop apps which means it shouldn't access Desktop only APIs.

timmarriott:
Ok,  then there is no way to bring up file/folder pickers from a windows runtime component since they now need access to the HWND and a Win32 only api to associate it with the picker.

On Nov 12, 2021, at 9:28 PM, Jaiganesh Kumaran ***@***.***> wrote:

﻿
This email is not from Hexagon’s Office 365 instance. Please be careful while clicking links, opening attachments, or replying to this email.


[...]

PR #3619, opened 2023-05-09
Improved C++ item & templates

JaiganeshKumaran:

WARNING:

Unfortunately, while C# item & project templates were cleaned up in 1.3.0, C++ item & project templates weren't.

Changes made in this pull request:
- Removed unnecessary MyProperty property and myButton_Click handler (except for MainWindow in the project template).
- Removed unnecessary include of g.cpp file. Almost every app-level XAML type must have a public constructor, which means the g.cpp file will get generated.
-  Replaced boxing a hardcoded type name string with the winrt::xaml_typename function. The previous one will actually crash at runtime because it isn't a TypeName.
- Included Windows.UI.Xaml.Interop.h in pch.h because its TypeName struct is still used in WinUI 3 and is required to get the winrt::xaml_typename function.
- Made the WinUI 3 runtime component "desktop-only" as WinUI 3 cannot be used by UWP apps. This is a fix for #1780.


dev/VSIX/ProjectTemplates/Desktop/CppWinRT/SingleProjectPackagedApp/App.cpp
```diff
@@ -1,48 +1,42 @@
...
+namespace $safeprojectname$::implementation
 {
-    InitializeComponent();
+    /// <summary>
+    /// Initializes the singleton application object.  This is the first line of authored code
+    /// executed, and as such is the logical equivalent of main() or WinMain().
+    /// </summary>
+    App::App()
+    {
+        InitializeComponent();
```
sylveon, line 19: Remove this one too

dev/VSIX/ProjectTemplates/Neutral/CppWinRT/RuntimeComponent/Class.cpp
```diff
@@ -1,24 +1,10 @@
-// Copyright (c) Microsoft Corporation and Contributors.
-// Licensed under the MIT License.
-
 #include "pch.h"
 #include "Class.h"
-#if __has_include("Class.g.cpp")
```
sylveon, line 6: Keep this, as simple classes might not have a public constructor

dev/VSIX/ProjectTemplates/Desktop/CppWinRT/SingleProjectPackagedApp/App.cpp
sylveon: I'm not sure what you're currently doing works. You should probably make an InitializeComponent function which calls the base one and then sets up the UnhandledException handler. Also, avoid passing `this` in the lambda?

JaiganeshKumaran: A custom InitializeComponent member function is only required to access XAML elements part of markup. Events can be set up in the constructor. Not sure why `this` was captured by the lambda before.

dev/VSIX/ProjectTemplates/Neutral/CppWinRT/RuntimeComponent/Class.cpp
JaiganeshKumaran: I also realise that users can opt out of component optimisations, which gets rid of the g.cpp file.

DrusTheAxe:
@Scottj1s

JaiganeshKumaran:
There are a few other things that could be done. Currently, the default templates enable dot prefix for C++/WinRT generated files in a sub-namespace. For example, if you have a runtime class under the View namespace, the generated file you'll need to include will be Views.RuntimeClass.g.h - however due to a long-standing XAML Compiler bug (https://github.com/microsoft/microsoft-ui-xaml/issues/7652), it will not compile since the file generated by the XAML Compiler always seems to be under a subfolder.

Another thing is if you add one of these to a project whose root namespace itself is nested, you'll see the templates emit dots in the C++ header and source instead of double colons. I would also really appreciate being able to create an item in a subfolder and it's automatically put into a sub-namespace.

Lastly, I see that the WinUI 3 item templates show up even when using WinUI 3. Should we move them all from neutral to desktop?

sotanakamura:
Could you add filters to organize files shown in Solution Explorer? The C++ project template has many more files than C#.

dev/VSIX/ProjectTemplates/Desktop/CppWinRT/RuntimeComponent/WinUI.Desktop.CppWinRT.RuntimeComponent.vstemplate
```diff
@@ -1,10 +1,10 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <VSTemplate Version="3.0.0" xmlns="http://schemas.microsoft.com/developer/vstemplate/2005" Type="Project">
   <TemplateData>
     <Name ID="1033" Package="FIXME-PACKAGEGUID" />
     <Description ID="1035" Package="FIXME-PACKAGEGUID" />
-    <Icon>WinUI.Neutral.CppWinRT.RuntimeComponent.ico</Icon>
-    <TemplateID>Microsoft.WinUI.Neutral.CppWinRT.RuntimeComponent</TemplateID>
+    <Icon>WinUI.Desktop.CppWinRT.RuntimeComponent.ico</Icon>
+    <TemplateID>Microsoft.WinUI.Desktopy.CppWinRT.RuntimeComponent</TemplateID>
```
evelynwu-msft, line 7: >Desktopy [](http://example.com/codeflow?start=32&length=8)

Nitpick: typo

evelynwu-msft:
@JaiganeshKumaran thank you so much for taking the time to make these improvements to our project templates. We reviewed this PR internally and we like it a lot although there are some things that we believe merit further discussion. Unfortunately, because of internal scheduling constraints we don't have sufficient time to have that additional discussion *and* include your work here in the WinAppSDK 1.4 release *and* resolve merge conflicts with other in-flight work that we know will occur. So here's the plan we devised to fast-track the uncontroversial parts of your PR:

1. I will create a new PR into the `develop` branch containing the portions of this PR that we know for certain we want to accept. This will allow for expedited code flow into the release branch.
2. Once things have settled down and you have had a chance to merge `main` into your topic branch we will pick up this PR where we left off.
3. The portions of this PR that weren't taken into the separate PR will target WinAppSDK 1.5. 

Please rest assured that we're not trying to take credit for your work. I would much prefer it if you could drive the process from beginning to end but unfortunately our schedule won't permit that and this is what we came up with as a compromise to allow most of your changes in this PR to ship in WinAppSDK 1.4.

sylveon:
You can make the commits in your PR as authored by Jaiganesh but committed by you - GitHub will keep the credits to him.

dev/VSIX/ItemTemplates/Neutral/CppWinRT/TemplatedControl/TemplatedControl.cpp
```diff
@@ -17,16 +12,6 @@ namespace winrt::$rootnamespace$::implementation
...
-
-    int32_t $safeitemname$::MyProperty()
-    {
-        throw hresult_not_implemented();
-    }
-
-    void $safeitemname$::MyProperty(int32_t /* value */)
-    {
-        throw hresult_not_implemented();
+        DefaultStyleKey(winrt::box_value(xaml_typename<class_type>()));
```
evelynwu-msft, line 15: >xaml_typename<class_type>() [](http://example.com/codeflow?start=41&length=27)

XAML accepts boxed strings so this change isn't necessary. Can you provide a min repro project for the crash you were encountering with the previous incarnation?

sylveon: It's not a crash, it's a compile error. C++/WinRT `hstring` does not implicitly convert to `IInspectable`. It needs to be explicitly done (via `winrt::box_value`). The template won't compile by default.

evelynwu-msft: That's correct. However, the change being proposed here is simply from `hstring` to `xaml_typename`; the enclosing `winrt::box_value()` call isn't being touched.

evelynwu-msft: The primary benefit of using `xaml_typename<class_type>()` over 'hstring' that I'm aware of is that it simplifies renaming the class. But the argument being made here instead is that there's an *error* associated with the `hstring` variant so that's why I want to dig deeper.

JaiganeshKumaran: You're right DefaultStyleKey can be set to a boxed hstring as well, it doesn't have to be a TypeName. As an alternative, winrt::name_of could be used instead of winrt::xaml_typename. There are a few differences between the name_of and xaml_typename, for example, name_of<Windows::Foundation::Point> returns "Windows.Foundation.Point" while xaml_typename<Windows::Foundation::Point> returns "Point". This doesn't matter in this case though.

PR #3627, opened 2023-05-12
Update MRTCore ManagedTests to use TAEF and onboard it to Pipeline Tests

kythant:
Update MRTCore ManagedTests to use TAEF and onboard it to Pipeline Tests.

Also updated the MRT Core yml to not use the VSTest


alexlamtest (commented):
How about removing the reference to this test from the .yml file that builds MRTCore?

kythant:
> How about removing the reference to this test from the .yml file that builds MRTCore?

Oh yes I forgot about this.

alexlamtest (approved):

PR #3640, opened 2023-05-17
API Spec for Microsoft.UI.ThemeSettings class

JesseCol:
This PR adds the proposed spec for the Microsoft.UI.ThemeSettings class.


specs/themes/ThemeSettings.md
```diff
@@ -0,0 +1,157 @@
...
+[AccessibilitySettings](https://learn.microsoft.com/en-us/uwp/api/windows.ui.viewmanagement.accessibilitysettings?view=winrt-22621)
+API to query the system's high-contrast setting, and detect when it changes.  But in a Win32/non-UWP context, the
+**AccessibilitySettings.HighContrastChanged** event never fires.  This is because the event relies on the **CoreWindow**
+object to handle the `WM_THEMECHANGED` message and fire an internal event to the **AccessibilitySettings** object -- but
+in a non-UWP, Win32/WinAppSDK context, no **CoreWindow** exists.
+
+The API described in this doc, **ThemeSettings**, is a WinAppSDK WinRT API that provides the functionality of
+**AccessibilitySettings**, but for a Win32 environment.  We expect to add other APIs to this type over time.
+
+The **ThemeSettings.HighContrast** and **ThemeSettings.HighContrastScheme** properties shown below are basically identical to the
```
FrankC-msft, line 17: The HighContrast Boolean seems redundant with HighContrastScheme having a non-empty value. (especially given the below comment that we'll return null/empty rather than the last active Scheme when HC is off.)  Is consistency with the existing APIs valuable in this case?

specs/themes/ThemeSettings.md
```diff
@@ -0,0 +1,157 @@
...
+## Open Issues and Implementation Issues
+In the experimental API today, HighContrastScheme returns a stale value if HighContrast is off.  We should change this to return
+the empty string.
+
+It seems easy to misuse this API by accidentally allowing the ThemeSettings object to go out of scope, which will mean the event
+never fires.
+
+In the original plan we weren't going to add HighContrast or HighContrastScheme.  Do we need them?  Should we remove?
+
+Note that today in the code the **Changed** object is called **HighContrastChanged** because it currently only applies to HighContrast.
```
FrankC-msft, line 145: Does this mean that changing HighContrastScheme from Aquatic to Dusk wouldn't proc a ThemeSettings::Changed event?  That sounds like a bug whether we have a peer Boolean or not.  UX trying to respect the current scheme will be left out of date.

JaiganeshKumaran:
This should have a few other things:
- App theme currently used
- App theme to override (if this is the default, app theme should return the default theme)
- Default theme selected in Settings to be used in apps
- Theme selected for the Windows shell in Settings (aka system theme, not very essential).
- Whether the option to apply accent colour on the title bar is on or off.
- Whether the option to apply accent on the taskbar is on or off (not very essential).

The app theme override property should probably persist automatically and be applied by the framework transparently. Maybe the class should implement INotifyPropertyChanged or be a DependencyObject to easily bind that to a XAML settings page.

specs/themes/ThemeSettings.md
```diff
@@ -0,0 +1,157 @@
...
+```
+
+# Appendix
+
+## Open Issues and Implementation Issues
+In the experimental API today, HighContrastScheme returns a stale value if HighContrast is off.  We should change this to return
+the empty string.
+
+It seems easy to misuse this API by accidentally allowing the ThemeSettings object to go out of scope, which will mean the event
+never fires.
```
oldnewthing, line 141: I think this is okay. This is a fairly well-established pattern: If you want something to raise events, you have to keep it alive (either directly or indirectly).

specs/themes/ThemeSettings.md
oldnewthing: I'm hoping this just means "only applies to HighContrast *(state or theme)*." I.e., "Since all of the properties on this object are HC-related, the event name is currently HCChanged. But this new API is going to use the Changed event to notify for any changes to the object, even future ones that may not be HC-related."

JesseCol: Sorry for the confusion.  Raymond is correct about what I meant.
(Changing HighContrastScheme from Aquatic to Dusk will raise the ThemeSettings.Changed event).

specs/themes/ThemeSettings.md
```diff
@@ -0,0 +1,157 @@
...
+
+In UWP, apps can use the
+[AccessibilitySettings](https://learn.microsoft.com/en-us/uwp/api/windows.ui.viewmanagement.accessibilitysettings?view=winrt-22621)
+API to query the system's high-contrast setting, and detect when it changes.  But in a Win32/non-UWP context, the
+**AccessibilitySettings.HighContrastChanged** event never fires.  This is because the event relies on the **CoreWindow**
+object to handle the `WM_THEMECHANGED` message and fire an internal event to the **AccessibilitySettings** object -- but
+in a non-UWP, Win32/WinAppSDK context, no **CoreWindow** exists.
+
+The API described in this doc, **ThemeSettings**, is a WinAppSDK WinRT API that provides the functionality of
+**AccessibilitySettings**, but for a Win32 environment.  We expect to add other APIs to this type over time.
```
david-risney, line 15: Other uses I see for 'theme' include ApplicationTheme, ElementTheme, ShareUITheme all of which are about light mode / dark mode. To my eye it looks odd to have a ThemeSettings and only accessibility / high contrast settings on it. Is the expected APIs to add to this over time include the notion of theme used elsewhere?

MikeHillberg: The expectation is to add more theme APIs here that we should have full APIs for:
* We need an API for light/dark mode, which currently you have to read from the registry
* System colors currently exposed by [GetSysColor](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getsyscolor) (Win32)
* Accent colors currently exposed by [UISettings.GetColorValue](https://docs.microsoft.com/en-us/uwp/api/windows.ui.viewmanagement.uisettings.getcolorvalu)

specs/themes/ThemeSettings.md
MikeHillberg: Discussion: What the API enables is to turn HC on and off without having to also pick which HC theme. The Win10 Settings UI had two controls, a toggle and a combo box, this was probably originally designed to match that.  (The API does not guarantee that when false that the string will be null/empty.)

specs/themes/ThemeSettings.md
```diff
@@ -0,0 +1,157 @@
...
+    String HighContrastScheme { get; } 
+  }
+}
+```
+
+# Appendix
+
+## Open Issues and Implementation Issues
+In the experimental API today, HighContrastScheme returns a stale value if HighContrast is off.  We should change this to return
+the empty string.
```
MikeHillberg, line 138: Discussion: This is the behavior of the underlying system API and so we shouldn't try to policy override it here. Also, this API will be available on Win10 where the Settings UI gives this meaning (see earlier comment)

specs/themes/ThemeSettings.md
MikeHillberg: Discussion: Yes, this is pattern is common and OK

specs/themes/ThemeSettings.md
MikeHillberg: Discussion: The `Changed` event is following the aggregate change event pattern here, meaning that when any of it changes the one event raises. This wraps WM_THEMECHANGED, so being more granular such as HighContrastChanged would actually be difficult to do, and the aggregate pattern is a not-uncommon one today.

specs/themes/ThemeSettings.md
JesseCol: Added discussion to background section.

specs/themes/ThemeSettings.md
JesseCol: I've removed this from the appendix.

specs/themes/ThemeSettings.md
JesseCol: I've removed this from the appendix.

specs/themes/ThemeSettings.md
JesseCol: Added notes to background section.

PR #3672, opened 2023-06-13
Change to using the centrally shared copy of DevCheck (from eng\common)

DrusTheAxe:


kythant:
Need to update yml to use the .cmd instead

DrusTheAxe:
> Need to update yml to use the .cmd instead

Good catch! But nah, anyone using the .ps1 can just change the path. Only reason for the cmd is end user interactive convenience, the yml specify the .ps1 they can (and should) continue doing so

jonwis (approved):

PR #3679, opened 2023-06-15
Remove experimental-ness from DeploymentContract

DrusTheAxe:
Remove experimental-ness from DeploymentContract. The API has parts that are experimental, but the contract isn't


dev/Deployment/Deployment.idl
```diff
@@ -5,7 +5,6 @@
 
 namespace Microsoft.Windows.ApplicationModel.WindowsAppRuntime
 {
-    [feature(Feature_DeploymentRepair)]
     [contractversion(4)]
```
jonwis, line 8: Does the [feature] tag need to be applied to other types that were added in v4?

DrusTheAxe: The 2 types new in v4 were already tagged (line 20+76). There's a separate thread discussing if that should change for 1.4 or (more likely) 1.5, but it's not a problem. [feature] on apicontract was the surprise build breaker.

jonwis (approved):

PR #3683, opened 2023-06-18
Fix PostBuild event for copying Bootstrap DLL

JaiganeshKumaran:
Fixes #3682.

A microsoft employee must use /azp run to validate using the pipelines below.

WARNING:
Comments made by azure-pipelines bot maybe inaccurate. 
Please see pipeline link to verify that the build is being ran.

For status checks on the develop branch, please use TransportPackage-Foundation-PR
(https://microsoft.visualstudio.com/ProjectReunion/_build?definitionId=81063&_a=summary)
and run the build against your PR branch with the default parameters.

For status checks on the main branch, please use microsoft.ProjectReunion
(https://dev.azure.com/ms/ProjectReunion/_build?definitionId=391&_a=summary)
and run the build against your PR branch with the default parameters.

PR #3710, opened 2023-07-07
Revert back to old xcopy command for now to unblock build pipelines.

codendone:
The attempted fix for #3708 caused problems in build pipelines due to /-I parameter not being known in that environment. Reverting back to previous state for now before #3683.


codendone:
wrong target branch, and the dropdown isn't working to let me change it.

PR #3714, opened 2023-07-09
KozaniHostRuntime implementation and FTA and protocol launch support

rhuang-msft:
### KozaniHostRuntime 

**Added logics to get remote app AUMID from app extension of the Local Variant Package (LVP). 

**Added logics to get the configuration file paths (connection.rdp file and AdditionalSettings.txt) from LVP's app data folder.**

**Implemented IKozaniStatusCallback interface to handle remote app status changes.** It will exit the KozaniHostRuntime process if the remote app is terminated, or the remote app activation fails, or the activation does not start a new instance (like UWPs that do not start a new app process for multiple activations). 

### KozaniManager
Fixed various bugs in process lifetime tracking, race condition, etc.

Moved Remote Desktop Client (RDC) launcher code to KozaniManagerRuntime.dll, which is loaded in-proc by the KozaniHostRuntime.exe to ensure RDC is launched in foreground. KozaniManager.exe is a background process with no UIs. Launching RDC by it will not guarantee the RDC login window to show up in foreground. Instead, it will be launched minimized with flashing icon at the taskbar, unless RDC itself pops up a warning dialog in the first connection to the new host. KozaniHostRuntime.exe on the other hand, has foreground privileges as it is launched by the user by activating the LVP app. Launching RDC from KozaniHostRuntime.exe will guarantee the RDC login dialog shows up in foreground. 

### KozaniRemoteManager
Added support for FTA and protocol launch in remote desktop server side. 

Moved activation code to KozaniRemoteManagerLauncher.exe to ensure the activated apps show up in foreground. KozaniRemoteManager.exe is a background process with no UIs. Launching apps from it will not guarantee the app will have foreground. If the app activation causes a new process to be created, then the windows from the new process will show up in foreground. However, if the activation does not create a new process, then the app window will not show up in foreground. UWP apps are single instancing and once launched, subsequent activation will not create a new process. It will be problematic if the activation is done from KozaniRemoteManager.exe as if the UWP app is already running and not in foreground (e.g., minimized), the activation will not bring the app to foreground, like nothing has happened. Win32 apps have the same problem if they implement single instancing. KozaniRemoteManagerLauncher.exe on the other hand, has foreground privileges as it is launched by remote desktop shell by RAIL. It is also a new process each time it is launched by the client. Activating apps from it will guarantee the apps will show up in foreground, whether a new process is created or not.

Used IApplicationActivationManager::ActivateForFile for UWP FTA activation, IApplicationActivationManager::ActivateForProtocol for UWP protocol activation because these 2 APIs only works for UWPs. Used ShellExecuteEx for Packaged desktop apps FTA and protocol activation. Implemented registry lookup logics to find the ProgId of the FTA and protocol association for the app to be used in ShellExecuteEx.


DrusTheAxe:
>However, if it is activated by file type association (FTA) or protocol, the command line will not contain the parameter. That's why we need additional logics to parse the manifest xml to get that info for FTA and protocol activation

Can't you query the ActivatedEventArgs to determine why you're KozaniHostRuntime is being activatied (e.g. `ActivationKind.FileTypeAssociation`)?

https://learn.microsoft.com/en-us/windows/apps/desktop/modernize/get-activation-info-for-packaged-apps

or event better https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.windows.applifecycle.appinstance.getactivatedeventargs?view=windows-app-sdk-1.3

(Any games involving "Find, Load and Parse appxmanifest.xml" is strongly discouraged for the yuuuge perf implications)

DrusTheAxe (changes requested):
>However, if it is activated by file type association (FTA) or protocol, the command line will not contain the parameter. That's why we need additional logics to parse the manifest xml to get that info for FTA and protocol activation

Can't you query the ActivatedEventArgs to determine why you're KozaniHostRuntime is being activatied (e.g. `ActivationKind.FileTypeAssociation`)?

https://learn.microsoft.com/en-us/windows/apps/desktop/modernize/get-activation-info-for-packaged-apps

or event better https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.windows.applifecycle.appinstance.getactivatedeventargs?view=windows-app-sdk-1.3

(Any games involving "Find, Load and Parse appxmanifest.xml" is strongly discouraged for the yuuuge perf implications)

rhuang-msft:
@DrusTheAxe , the problem is not because I can't get why KozaniHostRuntime is being activated. Of course we can get that info easily from ActivatedEventArgs. The problem is what the AUMID is for the remote app, associated with the LVP app being activated. It is encoded in the manifest of the LVP, like this:

<Application Id="App" uap10:HostId="KozaniHostRuntime" uap10:Parameters="-aumid Microsoft.WindowsCalculator_8wekyb3d8bbwe!App">

The uap10:Parameters will be passed to the KozaniHostRuntime.exe as command line parameter IF the LVP app is activated by launch contract (i.e., clicking the tile to launch it). However, if it is activated by FTA or Protocol, then the command line parameter will NOT contain the info. Instead, it will be the file path for FTA, or uri for protocol launch. That's why we need to load up the manifest to get the remote app AUMID. LVP's manifest is very small and perf hit is minimal.

DrusTheAxe:
> The problem is what the AUMID is for the remote app, associated with the LVP app being activated. It is encoded in the manifest of the LVP, like this:

Ahhh. So when the LVP app runs you need to know the associated app on the server to run?

RECOMMENDATION: Use appextension.

Add to the LVP's manifest

```
<uap3:Extension Category="windows.appExtension">
    <uap3:AppExtension Name="com.microsoft.kozani.localvariant.remoteapp.AUMID" Id="kozani.remoteapp.PRAID" DisplayName="Kozani RemoteApp REMOTEAUMID">
        <uap3:Properties>
            <RemoteApp AppUserModelId="REMOTEAUMID"/>
        </>
    </>
</>
```

where PRAID=the parent `<Application>`'s PRAID, AUMID=the parent `<Application>`'s AUMID and REMOTEAUMID=the remoteapp's AUMID. Then KozaniHostRuntime.exe can use the AppExtension API to retrieve this e.g.

```
aumid = GetCurrentApplicationUserModelId()
pkgfamilyname,praid = ParseApplicationUserModelId(aumid)
catalog = AppExtensionCatalog.Open("com.microsoft.kozani.localvariant.remoteapp." + aumid)
list<AppExtension> aes = co_await catalog.FindAllAsync()
IPropertySet properties = co_await aes[0].GetExtensionPropertiesAsync()
remoteaumid = properties.Lookup("RemoteApp").Lookup("@AppUserModelId")
```

The manifest is parsed when the app's installed and AppExtension is the generic metadata mechanism with APIs working off that already parsed and optimized data. Leverage that.

"Find, Load and Parse appxmanifest.xml" is strongly discouraged.

rhuang-msft:
@DrusTheAxe , I have updated the PR with the use of app extension. Please review. Thx.

rhuang-msft:
@Howard ***@***.***>, can you take a look at my updated PR? Used app extension for remote app AUMID.

Thanks
Ruke

dev/Kozani/KozaniHostRuntime/main.cpp
```diff
@@ -112,60 +115,68 @@ void LogArgs(int argc, PWSTR* argv)
...
-
-    for (int i = 0; i < argc; i++)
+    WCHAR lvpAumid[APPLICATION_USER_MODEL_ID_MAX_LENGTH + 1]{};
+    UINT32 lvpAumidLength{ ARRAYSIZE(lvpAumid) };
+    THROW_IF_WIN32_ERROR(GetCurrentApplicationUserModelId(&lvpAumidLength, lvpAumid));
+
+    WCHAR lvpFamilyName[PACKAGE_FAMILY_NAME_MAX_LENGTH + 1]{};
+    WCHAR lvpRelativeId[PACKAGE_RELATIVE_APPLICATION_ID_MAX_LENGTH + 1]{};
+    UINT32 lvpFamilyNameLength{ ARRAYSIZE(lvpFamilyName) };
+    UINT32 lvpRelativeIdLength{ ARRAYSIZE(lvpRelativeId) };
```
DrusTheAxe, line 143: SUGGESTION: `lvpPraid`

You use `Aumid` below but not `Praid` here

dev/Kozani/KozaniHostRuntime/main.cpp
```diff
@@ -112,60 +115,68 @@ void LogArgs(int argc, PWSTR* argv)
...
+    IVectorView<AppExtensions::AppExtension> appExtensions{ catalog.FindAllAsync().get() };
+    for (const auto appExtension : appExtensions)
+    {
+        if (appExtension.Id() == lvpRelativeId)
+        {
+            const auto properties{ appExtension.GetExtensionPropertiesAsync().get() };
+            const auto property{ properties.Lookup(L"RemoteAppAumid").as<IMap<winrt::hstring, winrt::Windows::Foundation::IInspectable>>() };
+            const auto remoteAumid{ winrt::unbox_value<winrt::hstring>(property.Lookup(L"#text")) };
+
+            auto otherData{ properties.Lookup(L"OtherData").as<IMap<winrt::hstring, winrt::Windows::Foundation::IInspectable>>() };
```
DrusTheAxe, line 169: What's OtherData? I don't see it mentioned in the XML snippet comment above

dev/Kozani/KozaniHostRuntime/main.cpp
```diff
@@ -112,60 +115,68 @@ void LogArgs(int argc, PWSTR* argv)
...
+            const auto otherDataText{ winrt::unbox_value<winrt::hstring>(otherData.Lookup(L"#text")) };
+
+            auto moreotherData{ properties.Lookup(L"MoreOtherData").as<IMap<winrt::hstring, winrt::Windows::Foundation::IInspectable>>() };
```
DrusTheAxe, line 172: What's MoreOtherData? I don't see it mentioned in the XML snippet comment above

dev/Kozani/KozaniHostRuntime/main.cpp
```diff
@@ -112,60 +115,68 @@ void LogArgs(int argc, PWSTR* argv)
...
+            const auto moreotherDataText{ winrt::unbox_value<winrt::hstring>(moreotherData.Lookup(L"#text")) };
+
+            return remoteAumid;
+        }
+    }
+
+    THROW_WIN32_MSG(ERROR_NOT_FOUND, "Cannot find remote app aumid from app extension with name: %ls, Id: %ls", appExtensionName.c_str(), lvpRelativeId);
```
DrusTheAxe, line 179: Change `Id:` to `PRAID:`

dev/Kozani/KozaniHostRuntime/main.cpp
```diff
@@ -194,33 +205,38 @@ void GetConfigurationFiles(std::wstring& rdpFilePath, std::wstring& additionalSe
...
-        // For launch activation, we can simply get the AUMID from command line arguments.
-        remoteAumid = GetRemoteAumidFromCommandLineArgs(pCmdLine);
-    }
-    else
-    {
-        remoteAumid = GetRemoteAumidFromManifest();
-    }
+#ifdef WAIT_FOR_DEBUGGER
+    WaitForDebugger();
+#endif
```
DrusTheAxe, line 229: Debugging code?

Obsolete and should be deleted?

DrusTheAxe (approved):
Nicely done. Only 2 Qs come to mind:

#1 Manifest extension XML
Is that doc'd? I see the comment in dev/Kozani/KozaniHostRuntime/main.cpp line 118+ which looks close (see comments there). Is that meant to doc the expected info in the manifest?

#2 E2E testing
Is there an automated E2E test for this here? Or in the Aggregator as an Integration test? Is that not possible without a remote environment?

dev/Kozani/KozaniHostRuntime/main.cpp
rhuang-msft: Changed.

dev/Kozani/KozaniHostRuntime/main.cpp
rhuang-msft: Removed. Used for testing, forgot to remove.

dev/Kozani/KozaniHostRuntime/main.cpp
rhuang-msft: Removed. Used for testing, forgot to remove.

dev/Kozani/KozaniHostRuntime/main.cpp
rhuang-msft: The AppExtension element only has the "Id" attribute. The log is specific to describe the AppExtension element.

rhuang-msft:
> Nicely done. Only 2 Qs come to mind:
> 
> #1 Manifest extension XML Is that doc'd? I see the comment in dev/Kozani/KozaniHostRuntime/main.cpp line 118+ which looks close (see comments there). Is that meant to doc the expected info in the manifest?
> 
The doc is not created yet. Right now the comments in line 118 serves the documentation.

> #2 E2E testing Is there an automated E2E test for this here? Or in the Aggregator as an Integration test? Is that not possible without a remote environment?
No automatic E2E testing as of now. Yes, it needs a remove environment for E2E.

dev/Kozani/KozaniHostRuntime/main.cpp
rhuang-msft: Leave it here so it can be enabled to facilitate debugging easily.

PR #3729, opened 2023-07-13
Use latest C++/WinRT in C++ project templates

evelynwu-msft:
Updating VS project templates to use the latest C++/WinRT Nuget package. The version we were previously using in some cases did not emit the fully-qualified name for types in the function signatures of projected types (e.g. the projection for `Microsoft.UI.Xaml.Markup.ComponentConnector`; the projected `Connect()` method had the signature `void Connect(int32_t connectionId, Windows::Foundation::IInspectable const& target)` rather than `void Connect(int32_t connectionId, winrt::Windows::Foundation::IInspectable const& target)`. Because the `Microsoft.UI.Xaml` namespace now has an API that consumes a type from the `Microsoft.Windows.ApplicationModel.Resources` namespace, the non-fully qualified name resulted in a namespace conflict during compilation. The updated version of C++/WinRT introduced by this change addresses that oversight.


JJBrychell (approved):

Scottj1s:
Thanks @evelynwu-msft !

PR #3745, opened 2023-07-20
Various fixes for C++ project and item templates

evelynwu-msft:
This is a partial adoption of #3619 by @JaiganeshKumaran in order to expedite the merging of the changes from that PR which do not need additional time for discussion. The individual changes that were extracted from the aforementioned PR are as follows:

- Made the WinUI 3 runtime component "desktop-only" as WinUI 3 cannot be used by UWP apps. This is a fix for #1780. (Note that unlike the original PR the source directory structure is not being changed in order to minimize potential merge conflicts.)

Additionally, this PR replaces `<ProjectTypeTag>uwp</ProjectTypeTag>` with `<ProjectTypeTag>WinUI</ProjectTypeTag>` (or just adds the latter if the former was not present at all) in the item templates. I don't believe this tag is surfaced anywhere in the Visual Studio UI but the consistency makes me happier.

### DIFF
--- a/dev/VSIX/ItemTemplates/Desktop/CSharp/BlankWindow/WinUI.Desktop.Cs.BlankWindow.vstemplate
+++ b/dev/VSIX/ItemTemplates/Desktop/CSharp/BlankWindow/WinUI.Desktop.Cs.BlankWindow.vstemplate
@@ -15,7 +15,7 @@
     <LanguageTag>csharp</LanguageTag>
     <PlatformTag>windows</PlatformTag>
     <ProjectTypeTag>desktop</ProjectTypeTag>
-    <ProjectTypeTag>WinUI</ProjectTypeTag>
+    <ProjectTypeTag>winui</ProjectTypeTag>
   </TemplateData>
   <TemplateContent>
     <ProjectItem OpenInEditor="true" ReplaceParameters="true" ItemType="Page" CustomTool="MSBuild:Compile" TargetFileName="$fileinputname$.xaml">BlankWindow.xaml</ProjectItem>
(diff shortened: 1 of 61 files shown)


dev/VSIX/ItemTemplates/Desktop/CSharp/BlankWindow/WinUI.Desktop.Cs.BlankWindow.vstemplate
```diff
@@ -15,6 +15,7 @@
     <LanguageTag>csharp</LanguageTag>
     <PlatformTag>windows</PlatformTag>
     <ProjectTypeTag>desktop</ProjectTypeTag>
+    <ProjectTypeTag>WinUI</ProjectTypeTag>
```
bpulliam, line 18: From this page, the value should be lowercase ("winui"): https://learn.microsoft.com/en-us/visualstudio/ide/template-tags?view=vs-2022

dev/VSIX/ProjectTemplates/Desktop/CppWinRT/PackagedApp/BlankApp/App.h
```diff
@@ -14,6 +11,6 @@ namespace winrt::$safeprojectname$::implementation
         void OnLaunched(Microsoft::UI::Xaml::LaunchActivatedEventArgs const&);
 
     private:
-        winrt::Microsoft::UI::Xaml::Window window{ nullptr };
+        winrt::Microsoft::UI::Xaml::Window m_window{ nullptr };
```
bpulliam, line 14: This is potentially a breaking change. If we keep it, it should be explicitly called out and justified.

bpulliam (approved):

dev/VSIX/ProjectTemplates/Desktop/CppWinRT/PackagedApp/BlankApp/App.h
evelynwu-msft: That's a good catch. I'll pull it out.

dotMorten:
@evelynwu-msft You made all the categories lower-case:

image

Looks really weird: 
image

cc @riverar

PR #3754, opened 2023-07-21
[release/1.4-stable] Various fixes for C++ project and item templates (#3745)

evelynwu-msft:
This is a partial adoption of #3619 by @JaiganeshKumaran in order to expedite the merging of the changes from that PR which do not need additional time for discussion. The individual changes that were extracted from the aforementioned PR are as follows:

- Made the WinUI 3 runtime component "desktop-only" as WinUI 3 cannot be used by UWP apps. This is a fix for #1780. (Note that unlike the original PR the source directory structure is not being changed in order to minimize potential merge conflicts.)

Additionally, this PR replaces `<ProjectTypeTag>uwp</ProjectTypeTag>` with `<ProjectTypeTag>WinUI</ProjectTypeTag>` (or just adds the latter if the former was not present at all) in the item templates. I don't believe this tag is surfaced anywhere in the Visual Studio UI but the consistency makes me happier.

---------

Co-authored-by: Jaiganésh Kumaran <jaiganesh.kumaran@outlook.com>
(cherry picked from commit 18ada2ffa013e488916125ca96a7007259250e05)

### DIFF
--- a/dev/VSIX/ProjectTemplates/Neutral/CppWinRT/RuntimeComponent/pch.h
+++ b/dev/VSIX/ProjectTemplates/Neutral/CppWinRT/RuntimeComponent/pch.h
@@ -1,16 +1,26 @@
-// Copyright (c) Microsoft Corporation and Contributors.
-// Licensed under the MIT License.
-
 #pragma once
+#include <windows.h>
 #include <unknwn.h>
+#include <restrictederrorinfo.h>
+#include <hstring.h>
+
+// Undefine GetCurrentTime macro to prevent
+// conflict with Storyboard::GetCurrentTime
+#undef GetCurrentTime
+
 #include <winrt/Windows.Foundation.h>
 #include <winrt/Windows.Foundation.Collections.h>
+#include <winrt/Windows.ApplicationModel.Activation.h>
+#include <winrt/Windows.UI.Xaml.Interop.h>
 #include <winrt/Microsoft.UI.Composition.h>
 #include <winrt/Microsoft.UI.Xaml.h>
 #include <winrt/Microsoft.UI.Xaml.Controls.h>
 #include <winrt/Microsoft.UI.Xaml.Controls.Primitives.h>
 #include <winrt/Microsoft.UI.Xaml.Data.h>
+#include <winrt/Microsoft.UI.Xaml.Interop.h>
 #include <winrt/Microsoft.UI.Xaml.Markup.h>
+#include <winrt/Microsoft.UI.Xaml.Media.h>
 #include <winrt/Microsoft.UI.Xaml.Navigation.h>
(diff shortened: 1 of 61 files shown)


codendone (approved):
:shipit:

dev/VSIX/ProjectTemplates/Neutral/CppWinRT/RuntimeComponent/pch.h
```diff
@@ -1,16 +1,26 @@
-// Copyright (c) Microsoft Corporation and Contributors.
-// Licensed under the MIT License.
-
 #pragma once
+#include <windows.h>
```
dmachaj, line 2: >windows [](http://example.com/codeflow?start=10&length=7)

`#define WIN32_LEAN_AND_MEAN`?

dev/VSIX/ProjectTemplates/Neutral/CppWinRT/RuntimeComponent/pch.h
```diff
@@ -1,16 +1,26 @@
...
 #include <winrt/Microsoft.UI.Xaml.Controls.h>
 #include <winrt/Microsoft.UI.Xaml.Controls.Primitives.h>
 #include <winrt/Microsoft.UI.Xaml.Data.h>
+#include <winrt/Microsoft.UI.Xaml.Interop.h>
 #include <winrt/Microsoft.UI.Xaml.Markup.h>
+#include <winrt/Microsoft.UI.Xaml.Media.h>
 #include <winrt/Microsoft.UI.Xaml.Navigation.h>
+#include <winrt/Microsoft.UI.Xaml.Shapes.h>
 #include <winrt/Microsoft.UI.Dispatching.h>
 #include <wil/cppwinrt_helpers.h>
```
dmachaj, line 26: >cppwinrt_helpers [](http://example.com/codeflow?start=14&length=16)

Does this include `<wil/cppwinrt.h>` to enable WIL/cppwinrt interop?

dmachaj, line 26: >include [](http://example.com/codeflow?start=1&length=7)

_Consider_ including `<wil/result_originate.h>` so that error logging automatically calls RoOriginateError.

evelynwu-msft: No. That header file's sole purpose is to support async calls (WIL is where `resume_foreground` ended up landing). More generally, I recall there was resistance in the past to the project templates automatically opting in developers to using WIL which is why we limit our usage to that one header file. If the functionality hadn't ended up in WIL then we wouldn't even have a Nuget package reference to it in the first place.

Issue #3760, opened 2023-07-25
The 1.4.230518007-experimental1 installer is flaky


tristanlabelle:
### Describe the bug

We install WindowsAppSDK `1.4.230518007-experimental1` in CI and it has started failing randomly after working for several weeks:

```
Installing license: MSIX_MAINPACKAGE_LICENSE
Install License result : 0x0 
Installing license: MSIX_SINGLETONPACKAGE_LICENSE
Install License result : 0x0 
Deploying package: Microsoft.WindowsAppRuntime.1.4-experimental1_4000.868.150.0_x64__8wekyb3d8bbwe
Package deployment result : 0x0 
Deploying package: Microsoft.WindowsAppRuntime.1.4-experimental1_4000.868.150.0_x86__8wekyb3d8bbwe
Package deployment result : 0x0 
Deploying package: MicrosoftCorporationII.WinAppRuntime.Main.1.4-e1_4000.868.150.0_x64__8wekyb3d8bbwe
Package deployment result : 0x0 
Error: Process completed with exit code 1.
```

### Steps to reproduce the bug

Our CI uses a `windows-2022` GitHub runner with this step:

```yaml
    - name: Install WindowsAppRuntime 1.4
      shell: pwsh
      run: |
        $installer = "$env:TEMP\installer.exe"
        invoke-WebRequest -Uri https://aka.ms/windowsappsdk/1.4/1.4.230518007-experimental1/windowsappruntimeinstall-x64.exe -Outfile $Installer
        & $installer
```

### Expected behavior

The installer should reliably succeed, or at least provide error information when it fails.

### Screenshots

History of successes vs failures:
![image](https://github.com/microsoft/WindowsAppSDK/assets/2314287/03fd4b4e-15f2-4b89-a30e-aff77e54a312)

### NuGet package version

Windows App SDK 1.4 Experimental 1: 1.4.230518007-experimental1

### Packaging type

Unpackaged

### Additional context

Windows version: `windows-2022` github runner, see: https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md
We tried capturing an ETL trace but it seems like the process is ending unexpectedly and the trace doesn't contain any useful information.

PR #3796, opened 2023-08-17
Fix crashes in the installer

codendone:
Fix crashes in the installer, which appears to account for 95%+ of crashes in WindowsAppRuntimeInstall.exe.

Errors fixed:
* InstallActivityContext.cpp was passing an hstring to StringCchPrintf for the activity Id. **Fix**: call ".data()" on the hstring to properly pass the WCHAR*.
* InstallActivityContext.cpp was passing std::wstring to StringCchPrintf for the resource (package) id and error text. **Fix**: call ".c_str()" on the std::wstring to properly pass the WCHAR*.
* InstallActivityContext.cpp for ProvisionPackage failures wasn't passing the resource/package id for one of the "%s" arguments. **Fix**: Pass m_currentResourceId.c_str() for that argument.
* InstallActivityContext.cpp for StagePackage failures was passing a GUID for one of the "%s" arguments. **Fix**: convert the GUID to a string, like is done for active Ids elsewhere.
* GetAndLogDeploymentOperationResult() didn't always return a value. **Fix**: put back the "return S_OK" which was removed in a previous change.
* DeregisterInstallerEventSourceW() was returning a "const BOOL&", which was a reference to a stack variable. **Fix**: Fixed that   compiler warning by making it not a reference.

This change is intended to address issues reported in #3760.

### DIFF
--- a/installer/dev/InstallActivityContext.cpp
+++ b/installer/dev/InstallActivityContext.cpp
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation and Contributors.
+// Copyright (c) Microsoft Corporation and Contributors.
 // Licensed under the MIT License.
 
 #include "pch.h"
@@ -65,7 +65,7 @@ BOOL WindowsAppRuntimeInstaller::InstallActivity::Context::LogInstallerCommandLi
         WCHAR message[1024]{};
         PCWSTR messageFormat{ L"Windows App Runtime Installer (ActivityId: %s) is called with Command Line Args: %s" };
         auto installerActivityId{ winrt::to_hstring(*m_activity.Id()) };
-        FAIL_FAST_IF_FAILED(StringCchPrintfW(message, ARRAYSIZE(message), messageFormat, installerActivityId, cmdlineArgs));
+        FAIL_FAST_IF_FAILED(StringCchPrintfW(message, ARRAYSIZE(message), messageFormat, installerActivityId.data(), cmdlineArgs));
         PCWSTR messageStrings[1]{ message };
         return ReportEventW(m_hEventLog, EVENTLOG_INFORMATION_TYPE, 0, eventId, nullptr, ARRAYSIZE(messageStrings), 0, messageStrings, nullptr);
     }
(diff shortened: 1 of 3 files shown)


installer/dev/InstallActivityContext.cpp
```diff
@@ -65,7 +65,7 @@ BOOL WindowsAppRuntimeInstaller::InstallActivity::Context::LogInstallerCommandLi
         WCHAR message[1024]{};
         PCWSTR messageFormat{ L"Windows App Runtime Installer (ActivityId: %s) is called with Command Line Args: %s" };
         auto installerActivityId{ winrt::to_hstring(*m_activity.Id()) };
-        FAIL_FAST_IF_FAILED(StringCchPrintfW(message, ARRAYSIZE(message), messageFormat, installerActivityId, cmdlineArgs));
+        FAIL_FAST_IF_FAILED(StringCchPrintfW(message, ARRAYSIZE(message), messageFormat, installerActivityId.data(), cmdlineArgs));
```
Scottj1s, line 68: @bpulliam had concern about whether winrt::data() was safe for empty strings.  appears so:
https://github.com/microsoft/cppwinrt/blob/9b453cfc518bdaa7e2ff590526c4883457fd6065/strings/base_string.h#L272C28-L272C28

Scottj1s (approved):

tristanlabelle:
Thank you so much! Would there be a new version of that installer published afterwards or would this only apply to future new releases?

codendone:
@tristanlabelle The first step will be to get this into the 1.4 stable release, which is coming soon. If there is a need for this on 1.3.x, we can also look at that.

tristanlabelle:
@codendone Thanks! Will that apply to all 1.4 releases? We are dependent on features in `1.4-experimental1` and would like to see fixes there too.

Also we are running into the issue I reported even when running the individual 4 MSIXes without the exe installer, so I wonder if there's something else going on.

codendone:
@tristanlabelle The current plan is to only have the fix in future releases. Many of the experimental APIs became stable in 1.4-preview2 (and will be in the upcoming 1.4-stable release) -- any chance that release has all of the functionality you need?

PR #3820, opened 2023-08-26
PackageDeploymentManager/PackageRuntimeManager v0 (experimental)

DrusTheAxe:
PackageDeploymentManager is an evolved and enhanced form of core Windows deployment APIs (e.g. [Windows.Management.Deployment.PackageManager](https://learn.microsoft.com/uwp/api/windows.management.deployment.packagemanager?view=winrt-22621)). These new APIs will provide much of Windows' functionality with enhancements and improved ease-of-use.

PackageRuntimeManager provides a complementary set of APIs for runtime access. This is particularly important given how a PackageSet can be an abstraction of multiple packages. PackageRuntimeManager makes it easy to use sets of packages via [Dynamic Dependencies](https://learn.microsoft.com/en-us/windows/apps/desktop/modernize/framework-packages/use-the-dynamic-dependency-api).

Much of this functionality is possible today with existing APIs, but with more effort or complexity on the developers part.

**NOTE:** This is an early 'v0' implementation for experimentation purposes to unblock other early development. More functionality and refinements are coming RSN. A full API spec detailing the API surface and behavior will be forthcoming in a future PR (and of course subsequent code changes per outcome of the review. This code is experimental until then).

The API lives under `dev\PackageManager\API`, and the associated tests under `test\PackageManager\API`. All tests pass.

This PR also updates the repository to TAEF 10.82.230714001 and C++/WinRT 2.0.230706.1.

TODO:

* Implement PackageRuntimeManager

Future PRs will...

* Change EnsurePackageSetIsReadyAsync() to use PackageDeploymentManager.AddPackageByUriAsync() instead of its copy of driving PackageManager.AddPackageAsync()
* Implement PackageDeploymentManager.AddPackageAsync()
* Implement PackageDeploymentManager.AddPackageSetAsync()
* Implement Remove*()
* Implement PackageSetManager for packageset definition persistence/lookup
* Implement *PackageSetById*()
* Improve progress reporting of *Async()
* Implement DeploymentProcessingModel.Parallel support
* Implement DeploymentPriority support
* Implement PackageVolume.Status()
* Implement PackageVolume.FixMe() (and rename to more official name)
* Implement PackageVolumeManager.FindPackageVolumeByName() via FrameworkUdk
* Replace FindPackageOptions placeholder with actual query (Find*()/Get*()/Exists*()) design
* Telemetry
* API spec for formalization of the APIs


dev/Common/AppModel.Identity.h
```diff
@@ -109,6 +112,120 @@ inline winrt::Windows::System::ProcessorArchitecture ParseShortArchitecture(_In_
...
+#endif // defined(____x_ABI_CWindows_CApplicationModel_CIPackageId_INTERFACE_DEFINED__)
+
+#if defined(WINRT_Windows_ApplicationModel_2_H)
+    winrt::Windows::ApplicationModel::PackageVersion ToWinrtPackageVersion() const
+    {
+        return winrt::Windows::ApplicationModel::PackageVersion{ Major, Minor, Build, Revision };
+    }
+#endif // defined(WINRT_Windows_ApplicationModel_2_H)
+
+#if defined(_XSTRING_) && defined(_STRSAFE_H_INCLUDED_) && defined(WI_VERIFY)
```
jonwis, line 188: Again, isn't this a core piece of the WASDK _implementation_ ? Why use the if guard here? If this is going into the SDK to be used by a C++ consumer, it's also OK to just require "to use PackageVersion you have to include string."  Can you use https://en.cppreference.com/w/cpp/utility/format/format instead so it's `return std::format(...)` ? Be idiomatic when possible.

dev/PackageManager/API/M.W.M.D.AddPackageOptions.cpp
```diff
@@ -0,0 +1,171 @@
...
+#include <IsWindowsVersion.h>
+
+#include "M.W.M.D.AddPackageOptions.h"
+#include "Microsoft.Windows.Management.Deployment.AddPackageOptions.g.cpp"
+
+namespace winrt::Microsoft::Windows::Management::Deployment::implementation
+{
+    winrt::Microsoft::Windows::Management::Deployment::PackageVolume AddPackageOptions::TargetVolume()
+    {
+        return m_targetVolume;
```
jonwis, line 15: There's a version of WIL that lets you put all these simple properties right into the header file. See https://github.com/microsoft/wil/blob/master/include/wil/cppwinrt_authoring.h for how to use it.

dev/PackageManager/API/PackageManager.idl
```diff
@@ -0,0 +1,383 @@
...
+//import "Windows.Management.Deployment.idl";
+//import "Microsoft.Windows.ApplicationModel.winmd;"
+
+//import "Windows.ApplicationModel.idl";
+import "M.AM.DynamicDependency.idl";
+//import "Microsoft.Windows.ApplicationModel.DynamicDependency.winmd";
+
+//import "Windows.Management.Deployment.idl";
+
+namespace Microsoft.Windows.Management.Deployment
```
jonwis, line 14: Presumably we'll have an API review of these coming soon?

dev/PackageManager/API/PackageManager.idl
```diff
@@ -0,0 +1,383 @@
...
+        InProgress = 1,         // The request is in progress
+        CompletedSuccess = 2,   // The request completed successfully
+        CompletedFailure = 3,   // The request failed with some critical internal error. //TODO:Check ExtendedError for HRESULT
+    };
+
+    [contract(PackageDeploymentContract, 1)]
+    runtimeclass PackageDeploymentProgress
+    {
+        PackageDeploymentProgressStatus status;
+        UInt32 percentage;
```
jonwis, line 156: This was an error on our part. We should have made this a `Single` since there are multi-gb packages in the world.

dev/PackageManager/API/PackageManager.idl
```diff
@@ -0,0 +1,383 @@
...
+        // Return true if the package set's content are present and available for use
+        Boolean IsPackageSetReady(PackageSet packageSet);
+
+        Boolean IsPackageSetReadyById(String packageSetId);
+
+        // Check if the necessary package(s) needed by the scenario are present
+        // and available for use and if not then Make It So.
+        // If the necessary packages(s) are not present on the system
+        // then make them available (download, install, etc).
+        // If the necessary AI Models are present and available this is equivalent to IsReady(id).
```
jonwis, line 322: Is this limited to AI, or would any package set be available here?

test/VersionInfo/packages.config
```diff
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="Microsoft.Taef" version="10.75.221207001" targetFramework="native" />
-  <package id="Microsoft.Windows.CppWinRT" version="2.0.221121.5" targetFramework="native" />
+  <package id="Microsoft.Taef" version="10.82.230714001" targetFramework="native" />
```
jonwis, line 3: Is there a "repo wide packages config" that we should be using here? Like a top-level "directory.props" or something that can be included?

dev/PackageManager/API/PackageManager.idl
cwruss: None of this should be limited to a specific type of packageset.

dev/PackageManager/API/M.W.M.D.AddPackageOptions.cpp
DrusTheAxe: I'm not following. I see in cppwinrt_authoring.h support for observer/change notifications and some things related to `single_threaded_property` but just a lot of implementation. No clear code (or docs or examples) doing what you describe. What line#s are you referring to? Or link to WIL docs on the subject?

dev/PackageManager/API/PackageManager.idl
DrusTheAxe: Yep. This is early experimental work to unblock other early folks early work. There'll be an API Spec coming soon, in a later PR based on this work plus feedback from that early use.

dev/PackageManager/API/PackageManager.idl
DrusTheAxe: Good idea, though Double seems more right than Single here. Using Tempo I see many progress properties are uint, some are double (e.g. MediaElement's.DownloadProgress) and I only see one using float (AnimationController), and notably PackageCatalog uses Double e.g. [`PackageInstallingEventArgs.Progress` property](https://learn.microsoft.com/uwp/api/windows.applicationmodel.packageinstallingeventargs.progress?view=winrt-22621#windows-applicationmodel-packageinstallingeventargs-progress)

dev/PackageManager/API/PackageManager.idl
DrusTheAxe: That's a typo from an early doc I started from. Carl's right, this API is about MSIX packages regardless of their content or use. Fixed the comment.

test/VersionInfo/packages.config
DrusTheAxe: Unfortunately no. VS supports that for *proj files (and thus C# since they only reference packages in *.csproj), but for C++ VS requires Nuget's packages.config and that has no macro or #include mechanism. That's why WinAppSDK built our own.

Dependencies MUST be listed in [eng\Version.Details.xml](https://github.com/microsoft/WindowsAppSDK/blob/main/eng/Version.Details.xml) if Maestro should automagically update when newer versions are available or [eng\Version.Dependencies.xml](https://github.com/microsoft/WindowsAppSDK/blob/main/eng/Version.Dependencies.xml) for manual updates only.

After editing those, run `DevCheck.cmd -CheckDependencies` to verify everything in the repo uses only sanctioned dependencies and only the sanctioned versions, or `DevCheck.cmd -SyncDependencies` to update everything in the repo as necessary.

The build pipelines include a `-CheckDependencies` step so you can't commit anything not matching eng\Version.*.xml.

This -Check/SyncDependencies was introduced late last year after repeated

dev/Common/AppModel.Identity.h
```diff
@@ -109,6 +112,120 @@ inline winrt::Windows::System::ProcessorArchitecture ParseShortArchitecture(_In_
...
+    PackageVersion(std::uint64_t version)
+    {
+        Version = version;
+    }
+
+// NOTE: There's no preprocessor symbol for Windows.ApplicationModel.PackageVersion
+//       or Windows.ApplicationModel.h so test for Windows.ApplicationModel.IPackageId
+//       as a close proxy for the ABI struct PackageVersion is defined.
+#if defined(____x_ABI_CWindows_CApplicationModel_CIPackageId_INTERFACE_DEFINED__)
+    PackageVersion(ABI::Windows::ApplicationModel::PackageVersion packageVersion) :
```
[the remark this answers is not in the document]
DrusTheAxe: This is forked from [MddBootstrap.h](https://github.com/microsoft/WindowsAppSDK/blob/main/dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.h) with some minor needed additions. I'm planning to move this to WIL hence the conditional behavior.

https://task.ms/46211363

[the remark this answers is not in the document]
DrusTheAxe: That's interesting. Thanks!

dev/Common/AppModel.Identity.h
DrusTheAxe: I forgot WinAppSDK switched to C++20. Updating to std::format()

Speaking of...would this work

```
const auto v{ Microsoft::Windows::ApplicationModel::PackageVersion(1,2,3,4) };
const auto message{ std::format(L"This is {}", v) };  //=>"This is 1.2.3.4"
```

if this was defined

```
template <>
struct std::formatter<PACKAGE_VERSION>
{
    constexpr auto parse(std::format_parse_context& ctx)
    {
        return ctx.begin();
    }
    auto format(const PACKAGE_VERSION& version, std::format_context& ctx)
    {
        std::format(std::format_to(ctx.out(), "{}.{}.{}.{}", version.major, version.minor, version.build, version.revision);
    }
};
```

?

DrusTheAxe: FYI I just learned about [/Zc:__cplusplus](https://learn.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-170#remarks)

I also learned [here](https://learn.microsoft.com/cpp/preprocessor/predefined-macros?view=msvc-170)  `_MSVC_LANG` "is set to 202002L if the [/std:c++20](https://learn.microsoft.com/cpp/build/reference/std-specify-language-standard-version?view=msvc-170) compiler option is specified".

Moral of the story, if you need to test the C++ language version use both

```
#if (__cplusplus >= 202002L) || ((defined(_MSVC_LANG) && (_MSVC_LANG >= 202002L)))
```

DrusTheAxe:
Build failed due to transient infrastructure error. Trying again

dev/PackageManager/API/PackageDeploymentResolver.cpp
```diff
@@ -0,0 +1,272 @@
...
+}
+
+static PCWSTR GetSystemSupportedArchitecturesAsString(const std::uint16_t nativeMachine)
+{
+    switch (nativeMachine)
+    {
+    case IMAGE_FILE_MACHINE_I386: return L"neutral,x86";
+    case IMAGE_FILE_MACHINE_AMD64:return L"neutral,x86,x64";
+    case IMAGE_FILE_MACHINE_ARM:  return L"neutral,x86,arm";
+    case IMAGE_FILE_MACHINE_ARM64:return L"neutral,x86,arm,arm64,x86onArm64";
```
sreading-MSFT, line 105: >, [](http://example.com/codeflow?start=54&length=1)

Should be X64 as well.

dev/PackageManager/API/PackageManager.idl
```diff
@@ -0,0 +1,407 @@
...
+    [contract(PackageDeploymentContract, 1)]
+    runtimeclass PackageDeploymentProgress
+    {
+        PackageDeploymentProgressStatus status;
+        Double percentage;
+    }
+
+    [feature(Feature_PackageManager)]
+    [contract(PackageDeploymentContract, 1)]
+    runtimeclass PackageDeploymentResult
```
sreading-MSFT, line 176: >PackageDeploymentResult [](http://example.com/codeflow?start=17&length=23)

This all seems to refer to a singular package whereas all the APIs that return it are for Sets. Would generally worry about trying to build a batched API on top of a windows API that only supports one package per operation. Seems like ideally the c++\winrt concurrency apis would just be straightforward enough to use that making an installing a vector of packages would be straightforward for callers to do themselves, but maybe they aren't.

dev/PackageManager/API/PackageDeploymentResolver.cpp
DrusTheAxe: Good catch! Thanks

dev/PackageManager/API/PackageManager.idl
DrusTheAxe: ByDesign. Yes, perhaps, in the future there'll be a good reason to add more-granular-results info.

Maybe. It's not very meaningful today.

The complex data about any problems with a deployment operation are logged (extensively) in the eventlog.

And there's precedent NOT to do that ==> `pkgmgr.AddPackageAsync("file://c:/foo/bar.msixbundle")` can install multiple packages (e.g. Main/x64 + Resource/en-us + Resource/scale-150 + ...), and possibly even fail during the batch operation (e.g. Main/x64 succeeds but Resource/en-us fails) and yet there's only a singular result about the operation, not granular failure details (Main/x64=success, Resource/en-us=fail, Resource/scale-150=not-even-tried, ...).

It's an interesting idea for a future enhancement, perhaps. It hasn't been a requirement to date, nor an impediment or pain point. Hence the reason it's not in the design.

>installing a vector of packages would be straightforward for callers to do themselves, but maybe they aren't.

That's one of those "The only difference between theory and practice is, in theory, there's no difference between theory and practice, whereas in practice..."

Simple concept, but the execution quickly gets complicated. One of the benefits of `PackageDeploymentManager` and `PackageSet`s is _we_, deeply experienced in the space and immersed in the complexities and nuances, can do the heavy lifting to implement such things many would find challenging to do well, for everyone to reap the benefit.

The v0 goal is to unblock early development. This is more of a marathon than a sprint. This is the first PR, not the last :-)

sreading-MSFT (approved):
:shipit:

dev/PackageManager/API/M.W.M.D.PackageVolumeManager.cpp
```diff
@@ -0,0 +1,54 @@
...
+    winrt::Windows::Foundation::Collections::IVector<winrt::Microsoft::Windows::Management::Deployment::PackageVolume> PackageVolumeManager::FindPackageVolumes()
+    {
+        winrt::Windows::Foundation::Collections::IVector<winrt::Microsoft::Windows::Management::Deployment::PackageVolume> microsoftPackageVolumes{ winrt::single_threaded_vector<winrt::Microsoft::Windows::Management::Deployment::PackageVolume>() };
+
+        winrt::Windows::Management::Deployment::PackageManager packageManager;
+        winrt::Windows::Foundation::Collections::IIterable<winrt::Windows::Management::Deployment::PackageVolume> windowsPackageVolumes{ packageManager.FindPackageVolumes() };
+        for (winrt::Windows::Management::Deployment::PackageVolume windowsPackageVolume : windowsPackageVolumes)
+        {
+            //winrt::Microsoft::Windows::Management::Deployment::PackageVolume microsoftPackageVolume{ windowsPackageVolume };
+            auto microsoftPackageVolume{ winrt::make<winrt::Microsoft::Windows::Management::Deployment::implementation::PackageVolume>(windowsPackageVolume) };
```
sreading-MSFT, line 21: Can be deleted.

cwruss (approved):

PR #3833, opened 2023-09-01
Fixed a redundant error check in Test-PackageDependency.Tests.ps1

DrusTheAxe:


alexlamtest (commented):
Thanks for fixing this instance.
Is the similar instance in Test-MsixDynamicDependency.Tests.ps1 being intentionally skipped?

DrusTheAxe:
> Thanks for fixing this instance. Is the similar instance in Test-MsixDynamicDependency.Tests.ps1 being intentionally skipped?

Same redundancy. No surprise as the cmdlet tests were derived from the function tests.

Fixed to match

alexlamtest (approved):

PR #3996, opened 2023-11-28
Updated WindowsAppSDK-RunTestsInPipeline-Job.yml

kythant:
Pipeline runs seem to be failing consistently due to failure to acquire a co_refresh VM for testing.
OSGWiki says cobalt Home/Pro SKUs hit End of Support on Oct 10, 2023. That appears to explain it.
Skipping testing on co_refresh to unblock pipeline runs. Cobalt is 21h2. We are still testing the Windows.11.Enterprise.MultiSession.21h2 OS image which has another year before end of support, so we don't lose much test coverage with this change.


DrusTheAxe:
"21H2" is both Win10 and Win11. You're referring to the Win11 21H2 (aka SV1)?

https://en.wikipedia.org/wiki/Windows_10_version_history
image

https://en.wikipedia.org/wiki/Windows_11_version_history
image


>We are still testing the Windows.11.Enterprise.MultiSession.21h2...so we don't lose much test coverage with this change.

Zero loss if you count supported releases. Win11 21H2 Home/Pro supported ended last month (Oct'2023). Win11 21H2 Enterprise plus Win11 22H2 are still supported and their test pipelines are still humming so all good

DrusTheAxe (approved):
LGTM

PR #4012, opened 2023-12-02
Update MSBuild PropertyGroup for Extended .NET Runtime Identifier Support

ShashankNay:
**Current Behavior:**
The current MSBuild logic checks if the Target Framework Moniker (TFM) is NET8 and conditionally sets the runtime identifiers to win-x86, win-x64, win-arm64 (from win10-x86, win10-x64, win10-arm64).

**Proposed Changes:**
Instead of checking for a specific TFM (like net8.0), the logic should be based on whether the target framework version is greater than or equal to 8. 

**Reason for Change:**
This change ensures our build process stays forward-compatible with upcoming .NET versions, starting from .NET 9. It also simplifies our project templates by removing the need for future updates when new .NET versions are released.


Scottj1s (approved):
:shipit:

PR #4020, opened 2023-12-06
[1.4] Correct Binary Deployment for .NET8 Breaking Change Support

bpulliam:
.NET8 introduced a [conditional breaking change](https://github.com/dotnet/docs/issues/35398) that influences how the runtime host looks for RID-specific assets. In NET8, support for non-portable RIDs (runtime identifiers specific to versions and distributions, such as win10-x86) is discontinued. The folder structure of the Microsoft.WindowsAppSDK NuGet is structured with assets marked using these non-portable RIDs:

The changes made to the `CopyFilesToStagingDir.ps1` script along with the project template changes should help support the .NET8 scenario. The selection of the project template's RIDs is conditional and based on the TFM chosen by the user. If NET8 is chosen, the RIDs become portable, enabling the runtime host to locate the RID-specific assets that were previously unattainable.

### DIFF
--- a/build/CopyFilesToStagingDir.ps1
+++ b/build/CopyFilesToStagingDir.ps1
@@ -137,7 +137,9 @@ PublishFile $FullBuildOutput\WindowsAppRuntime_DLL\Microsoft.WindowsAppRuntime.l
 #
 # MSIX Framework package - DLLs/EXEs
 PublishFile $FullBuildOutput\WindowsAppRuntime_DLL\Microsoft.WindowsAppRuntime.dll $NugetDir\runtimes\win10-$Platform\native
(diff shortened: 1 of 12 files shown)


Scottj1s:
We should cherry pick this change instead (future proof):
https://github.com/microsoft/WindowsAppSDK/pull/4012

Scottj1s (approved):
Are all .NET 8 changes accounted for here?

ShashankNay (approved):

Scottj1s:
> Can you please deploy a package version with this fix ? Neither the last 1.4.231115000 nor 1.5.231202003-experimental1 contains it.
> 
> ty in advance

It's on the way - thanks for your patience

PR #4035, opened 2023-12-14
Add custom titlebar to default Cs and Cpp WinUI template

pratikone:
This change adds custom titlebnar to Cs and Cpp template. As a result, any hello world WinUI app created via the template has custom titlebar already applied to it. This provides a better experience for developers as default option takes care of styling of the window.  It works better with overall theming.
For example : 

current hello world app in dark theme (notice the mismatch between app content and OS provided titlebar)
image (1)

this is with custom titlebar applied (theme consistency extends till the titlebar area too)
image

This is a better experience right out of the box. Similar default is followed by MAUI and Template Studio projects.

### DIFF
--- a/dev/VSIX/ItemTemplates/Desktop/CppWinRT/BlankWindow/BlankWindow.h
+++ b/dev/VSIX/ItemTemplates/Desktop/CppWinRT/BlankWindow/BlankWindow.h
@@ -10,6 +10,7 @@ namespace winrt::$rootnamespace$::implementation
         {
             // Xaml objects should not call InitializeComponent during construction.
             // See https://github.com/microsoft/cppwinrt/tree/master/nuget#initializecomponent
+            ExtendsContentIntoTitleBar(true); // provides default WinUI custom title bar experience
         }
 
         int32_t MyProperty();
--- a/dev/VSIX/ItemTemplates/Desktop/CSharp/BlankWindow/BlankWindow.xaml.cs
+++ b/dev/VSIX/ItemTemplates/Desktop/CSharp/BlankWindow/BlankWindow.xaml.cs
@@ -26,6 +26,7 @@ namespace $rootnamespace$
         public $safeitemname$()
         {
             this.InitializeComponent();
+            ExtendsContentIntoTitleBar = true; // provides default WinUI custom title bar experience
         }
     }
 }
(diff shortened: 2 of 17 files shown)


Scottj1s:
just item templates or project templates too?

dev/VSIX/ItemTemplates/Desktop/CppWinRT/BlankWindow/BlankWindow.h
```diff
@@ -10,6 +10,7 @@ namespace winrt::$rootnamespace$::implementation
         {
             // Xaml objects should not call InitializeComponent during construction.
             // See https://github.com/microsoft/cppwinrt/tree/master/nuget#initializecomponent
+            ExtendsContentIntoTitleBar(true); // provides default WinUI custom title bar experience
```
evelynwu-msft, line 13: >ExtendsContentIntoTitleBar(true); // provides default WinUI custom title bar experience [](http://example.com/codeflow?start=12&length=87)

This feels incomplete. Should the template include a full example (similar to what's shown in [https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.window.settitlebar?view=windows-app-sdk-1.4#microsoft-ui-xaml-window-settitlebar(microsoft-ui-xaml-uielement](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.window.settitlebar?view=windows-app-sdk-1.4#microsoft-ui-xaml-window-settitlebar(microsoft-ui-xaml-uielement))) to demonstrate the complete usage?

pratikone: this is the complete use.  Calling this api creates a default custom titlebar which extends the entire width of the window i.e. same area as system titlebar would have taken.  the docs are not updated strangely. I have had updated them while back.  See my merged PR : Merged docs PR : https://github.com/MicrosoftDocs/winapps-winrt-api/pull/89
 I will dig that up. Thanks for bringing this to my attention.

pratikone:
> just item templates or project templates too?

oops. fixed it. now project templates and item templates both have it.

bpulliam (approved):

ramjke:
The custom title bar produces a serious bug (https://github.com/microsoft/microsoft-ui-xaml/issues/10357). It should be eliminated before making the custom title bar the default option.

PR #4040, opened 2023-12-15
Project Template updates to improve .NET 8 Compatibility and Unit Test Support

ShashankNay:
- Default TFM set to NET8.
- Publish profiles updated to use correct paths.
- Unit test projects now support .NET8+.
- Added new project templates for C# and C++/WinRT WinUI Unit Test Apps using MSTest and CppUnitTestFramework, respectively.


Scottj1s:
Needs the rest of the unit test project template work as well

NuGet.config
```diff
@@ -6,8 +6,10 @@
   </config>
   <packageSources>
     <clear />
+    
     <add key="ProjectReunion internal" value="https://microsoft.pkgs.visualstudio.com/ProjectReunion/_packaging/Project.Reunion.nuget.internal/nuget/v3/index.json" />
-    <add key="WindowsAppSDK Protobuf" value="tools/nuget" />
+    <add key="WindowsAppSDK Local" value="tools/nuget" />
```
Scottj1s, line 11: was this for local tests/builds? need to be permanent?

Scottj1s (approved):

PR #4044, opened 2023-12-16
Start running tests on Win11 23h2 Profession arm64 OS image

alexlamtest:
Start running tests on Win11 23h2 Professional arm64 OS image

--------

WARNING:

### DIFF
--- a/build/AzurePipelinesTemplates/WindowsAppSDK-RunTestsInPipeline-Job.yml
+++ b/build/AzurePipelinesTemplates/WindowsAppSDK-RunTestsInPipeline-Job.yml
@@ -85,6 +85,12 @@ jobs:
           buildPlatform: arm64
           buildConfiguration: release
           testLocale: en-US
+        Win11_Pro_23h2_arm64fre:
+          imageName: Windows.11.Professional.23h2.Arm64
+          buildPlatform: arm64
+          buildConfiguration: release
+          testLocale: en-US
+
   pool:
     type: windows
     isCustom: true


bpulliam (approved):

build/AzurePipelinesTemplates/WindowsAppSDK-RunTestsInPipeline-Job.yml
```diff
@@ -85,6 +85,12 @@ jobs:
           buildPlatform: arm64
           buildConfiguration: release
           testLocale: en-US
+        Win11_Pro_23h2_arm64fre:
+          imageName: Windows.11.Professional.23h2.Arm64
```
DrusTheAxe, line 89: Is this equivalent to x64's `rs_prerelease_x64fre`?

Why 23H2 Professional for arm64 but no Professional for x64?

Names are very inconsistent between x64 and arm64 making it hard to understand if there's parity or even what some images are (`Windows.11.Profession.Arm64` is 22H2? 21H2? rs_prerelease? Gah!)

DrusTheAxe (approved):

build/AzurePipelinesTemplates/WindowsAppSDK-RunTestsInPipeline-Job.yml
alexlamtest: Win11_Pro_23h2_arm64fre does not correspond to rs_prerelease_x64fre. My understanding is that the latter is routinely updated with new feature code, while the former is latched onto a servicing branch updated with bug fixes only.
Regarding "but no Professional for x64?", we are not pursuing a complete test matrix here, just sampling some configurations + platform combos, to save resources.
The Names we use follow the names of the corresponding base OS images.  My understanding is that Windows.11.Profession.Arm64 attempts to stay on a "reasonably new" released of Win11 Pro, it is not tied to a specific version such as 22h2, 21h2, etc., while Win11_Pro_23h2_arm64fre, as its name suggests, is tied to 23h2.

Issue #2639, opened 2022-06-17
Microsoft.Storage.ApplicationData


DrusTheAxe:
### Describe the bug

# Proposal: Microsoft.Storage.ApplicationData

## Summary
Provide a modern enhanced version of [Windows.Storage.ApplicationData](https://docs.microsoft.com/uwp/api/windows.storage.applicationdata?view=winrt-22621) and [Windows.Management.Core.ApplicationDataManager](https://docs.microsoft.com/uwp/api/Windows.Management.Core.ApplicationDataManager?view=winrt-22621)

Microsoft-internal task [NNN](https://task.ms/NNN)

## Rationale
* `Windows.Storage.ApplicationData.Current` only works fir you have package identity AND running in AppContainer
* `Windows.Management.Core.ApplicationDataManager.CreateForPackageFamily` is equivalent but requires you have package identity AND running MediumIL or higher
* Unpackaged apps have long been advised to use `%LOCALAPPDATA%\publisher\product`, `%LOCALAPPDATA%\Temp` and `HKCU\SOFTWARE\publisher\product` but ApplicationData doesn't support that
* `Windows.Storage.ApplicationData` has several deprecated and obsolete features (e.g. [.RoamingFolder](https://docs.microsoft.com/uwp/api/windows.storage.applicationdata.roamingfolder?view=winrt-22621))
* `ApplicationData.*Folder` only provide access to the filesystem via `StorageFolder`. There's no way to get their `.Path` without going thrug a `StorageFolder` incurring significant performance overhead
* `ApplicationData` has async methods which should be available synchronously e.g. `ClearAsync()` should offer `Clear() as a synchronous variant (or replacement?)

Windows App SDK can and should provide a modern revision of the `ApplicationData` API, addressing the issues above.

## Scope
| Capability | Priority |
| :---------- | :------- |
| APIs available to packaged and unpackaged applications | Must |
| APIs offer filesystem access without requiring use of StorageFolder | Must |
| WinRT API | Must |

## Important Notes

Here's a sketch of the general shape of the API to express the intent of the ask. The actual API when designed/spec'd/reviewed/approved may differ.

### WinRT API
Microsoft.Storage.ApplicationData.idl
```c# (closest to midl3)
namespace Microsoft.Storage.ApplicationData
{
[contract(ApplicationData, 1)]
enum ApplicationDataLocality
{
    Local = 0,
    LocalCache = 3,
    SharedLocal = 4,
    Temporary = 2,
}

[...]

PR #4056, opened 2024-01-02
Microsoft.Storage.ApplicationData API Spec

DrusTheAxe:
ApplicationData APIs comparable to those in namespaces [Windows.Storage](https://learn.microsoft.com/uwp/api/windows.storage) and [Windows.Management.Core](https://learn.microsoft.com/uwp/api/windows.management.core) but with additional functionality, improved developer experience and performance optimizations.

Issue #2639

Microsoft-internal tasks [22308619](https://task.ms/22308619), [30002314](https://task.ms/30002314), [39401257](https://task.ms/39401257), [39414843](https://task.ms/39414843), [40050113](https://task.ms/40050113) and [48332357](https://task.ms/48332357).

### DIFF
--- a/specs/applicationdata/ApplicationData.md
+++ b/specs/applicationdata/ApplicationData.md
@@ -1,0 +1,389 @@
+# 1. Application Data
+
+- [1. Application Data](#1-application-data)
+- [2. Background](#2-background)
+- [3. Description](#3-description)
+  - [3.1. ApplicationData and ApplicationDataManager functionality](#31-applicationdata-and-applicationdatamanager-functionality)
+  - [3.2. \*Path properties](#32-path-properties)
+  - [3.3. Reliable access](#33-reliable-access)
+  - [3.4. Per-Machine data store](#34-per-machine-data-store)
+  - [3.5. Unpackaged app data stores](#35-unpackaged-app-data-stores)
+- [4. Examples](#4-examples)
+  - [4.1. Packaged app using LocalPath](#41-packaged-app-using-localpath)
+  - [4.2. Unpackaged app using package'd app's LocalPath](#42-unpackaged-app-using-packaged-apps-localpath)
+  - [4.3. Packaged app using MachinePath](#43-packaged-app-using-machinepath)
+  - [4.4. Unpackaged app using LocalPath](#44-unpackaged-app-using-localpath)
+- [5. Remarks](#5-remarks)
+  - [5.1. Platform Support](#51-platform-support)
+- [6. API Details](#6-api-details)
+
+This feature provides application data APIs comparable to those in namespaces
+[Windows.Storage](https://learn.microsoft.com/uwp/api/windows.storage) and
+[Windows.Management.Core](https://learn.microsoft.com/uwp/api/windows.management.core) but with
+additional functionality, improved developer experience and performance optimizations.
+
+This feature provides APIs to access and manipulate app data stores.
+
+- [1. Application Data](#1-application-data)
+
+# 2. Background
+
+Windows provides data stores for packaged applications to store data by user and package family via
+WinRT APIs in the Windows.Storage and Windows.Management.Core namespaces. These API provide the
+means to create, use and manage data in these data stores.
+
+The Application Data APIs were originally introduced in Windows 8. However, as one of the earliest
+WinRT APIs they include some patterns out of step with current API practices and recommendations.
+Windows App SDK offers the opportunity to provide a new generation of APIs in line with the latest
+patterns and recommendations for an improved developer experience as well as functional enhancements
+and improved runtime efficiencies.
+
+Microsoft-internal tasks [22308619](https://task.ms/22308619),
+[30002314](https://task.ms/30002314), [39401257](https://task.ms/39401257),
+[39414843](https://task.ms/39414843), [40050113](https://task.ms/40050113) and
+[48332357](https://task.ms/48332357).
+
+# 3. Description
+
+This API supports app data stores focusing on the following scenarios:
+
+* ApplicationData and ApplicationDataManager functionality
+* *Path properties
+* Reliable access
+* Per-Machine data store
+* Unpackaged app data stores
+
+## 3.1. ApplicationData and ApplicationDataManager functionality
+
+This API provides an equivalent superset of
+[Windows.Storage.ApplicationData](https://learn.microsoft.com/uwp/api/windows.storage.applicationdata),
+with the following notable differences:
+
+* **No roaming functionality** -- We don't provide equivalents to Windows.Storage.ApplicationData's
+      [RoamingSettings](https://learn.microsoft.com/uwp/api/windows.storage.applicationdata.roamingsettings),
+      [RoamingStorageQuota](https://learn.microsoft.com/uwp/api/windows.storage.applicationdata.roamingstoragequota),
+      [Version](https://learn.microsoft.com/uwp/api/windows.storage.applicationdata.version),
+      [SetVersionAsync()](https://learn.microsoft.com/uwp/api/windows.storage.applicationdata.setversionasync),
+      [SignalDataChanged()](https://learn.microsoft.com/uwp/api/windows.storage.applicationdata.signaldatachanged)
+      and
+      [DataChanged](https://learn.microsoft.com/uwp/api/windows.storage.applicationdata.datachanged). See [5.1. Platform Support](#51-platform-support) for more details.
+* **GetDefault() works for all packaged apps** -- Equivalent to .Current but works for all packaged
+  apps (not just callers running in an AppContainer). See
+  [3.3. Reliable access](#33-reliable-access).
+* **`GetForPackageFamily()`** -- Equivalent to Windows.Storage.ApplicationDataManager.CreateForPackageFamily().
+* **`GetForUser()`** -- Synchronous equivalent to GetForUserAsync().
+* **`GetForUnpackaged()`** -- Equivalent functionality for unpackaged apps. See
+  [3.5. Unpackaged app data stores](#35-unpackaged-app-data-stores).
+
+## 3.2. *Path properties
+
+New *Path equivalents to Windows.Storage.ApplicationData's *Folder but returning a `String` rather than `StorageFolder`.
+
+## 3.3. Reliable access
+
+`Windows.Storage.ApplicationData.Current` only works for packaged apps **running in an AppContainer**. Processes not running in an AppContainer (e.g. MediumIL) must use the equivalent `Windows.Management.Core.ApplicationDataManager.CreateForPackageFamily(Windows.ApplicationModel.Package.Current.Id.FamilyName)`.
+
+Microsoft.Windows.Storage.ApplicationData provides consistent behavior as developers expect e.g. `ApplicationData.Current` works regardless if the caller's running in an AppContainer or not.
+
+## 3.4. Per-Machine data store
+
+New properties and methods to access per-machine data:
+
+* ApplicationDataLocality.Machine
+* ClearMachineFolderAsync()
+* IsMachinePathSupported
+* MachineFolder
+* MachinePath
+
+## 3.5. Unpackaged app data stores
+
+Unpackaged applications can use the ApplicationData API to access app data stores in the classic locations:
+
+|Packaged                        |Unpackaged                                                         |
+|--------------------------------|-------------------------------------------------------------------|
+|ApplicationData.LocalCachePath  |%LOCALAPPDATA%\<publisher>\<product>                               |
+|ApplicationData.LocalPath       |%LOCALAPPDATA%\<publisher>\<product>                               |
+|ApplicationData.MachinePath     |%ProgramData%\<publisher>\<product>                                |
+|ApplicationData.RoamingPath     |%APPDATA%                                                          |
+|ApplicationData.TemporaryPath   |[GetTempPath2W()](https://learn.microsoft.com/windows/win32/api/fileapi/nf-fileapi-gettemppath2w)<sup>1</sup>|
+|ApplicationData.LocalSettings   |HKCU\SOFTWARE\Classes\Local Settings\Software\<publisher>\<product>|
+|ApplicationData.LocalSettings   |HKCU\SOFTWARE\<publisher>\<product>                                |
+
+<sup>1</sup>
+[GetTempPath2W()](https://learn.microsoft.com/windows/win32/api/fileapi/nf-fileapi-gettemppath2w)
+returns `C:\Windows\SystemTemp` for SYSTEM processes else for non-SYSTEM processes checks for the
+existence of environment variables in the following order and uses the first path found:
+
+1. The path specified by the TMP environment variable.
+2. The path specified by the TEMP environment variable.
+3. The path specified by the USERPROFILE environment variable.
+4. The Windows directory.
+
+See https://learn.microsoft.com/windows/win32/api/fileapi/nf-fileapi-gettemppath2w for more details.
+
+GetTempPath2W() is available on Windows version &gt;= 10.0.20348.0. `ApplicationData.TemporaryPath`
+performs the equivalent logic on older systems (aka Polyfill).
+
+# 4. Examples
+
+## 4.1. Packaged app using LocalPath
+
+Fabrikam packaged app creating a file in the LocalPath data store.
+
+```c#
+void SaveData(string data)
+{
+    var applicationData = ApplicationData.GetDefault();
+    var filename = Path.Combine(applicationData.LocalPath, "stuff.dat");
+    File.WriteAllText(filename, data, System.Text.UTF8Encoding);
+}
+```
+
+## 4.2. Unpackaged app using package'd app's LocalPath
+
+An unpackaged app app creating a file in the Fabrikam app's LocalPath data store.
+
+```c#
+void SaveData(string data)
+{
+    string packageFamilyName = "Fabrikam.Contoso_1234567890abc";
+    var applicationData = ApplicationData.GetForPackageFamily(packageFamilyName);
+    var filename = Path.Combine(applicationData.LocalPath, "stuff.dat");
+    File.WriteAllText(filename, data, System.Text.UTF8Encoding);
+}
+```
+
+## 4.3. Packaged app using MachinePath
+
+Fabrikam packaged app running elevated and creating a file in the LocalPath data store.
+
+```c#
+void SaveData(string data)
+{
+    var applicationData = ApplicationData.GetDefault();
+    var filename = Path.Combine(applicationData.MachinePath, "importantstuff.dat");
+    File.WriteAllText(filename, data, System.Text.UTF8Encoding);
+}
+```
+
+## 4.4. Unpackaged app using LocalPath
+
+An unpackaged app app creating a file its LocalPath data store.
+
+```c#
+void SaveData(string data)
+{
+    string publisher = "Fabrikam";
+    string product = "Contoso";
+    var applicationData = ApplicationData.GetForUnpackaged(publisher, product);
+    var filename = Path.Combine(applicationData.LocalPath, "stuff.dat");
+    File.WriteAllText(filename, data, System.Text.UTF8Encoding);
+}
+```
+
+# 5. Remarks
+
+## 5.1. Platform Support
+
+Per https://learn.microsoft.com/uwp/api/windows.storage.applicationdata.roamingfolder#remarks:
+  > Roaming data and settings is no longer supported as of Windows 11. The recommended replacement
+      is Azure App Service. Azure App Service is widely supported, well documented, reliable, and
+      supports cross-platform/cross-ecosystem scenarios such as iOS, Android and web. Settings
+      stored here no longer roam (as of Windows 11), but the settings store is still available. 
+
+We provide `RoamingFolder` and `RoamingSettings` equivalents but they're only as functional as
+Windows provides (i.e. no data roaming on Windows after 1909 aka 19H2 aka 10.0.18363.0).
+
+We don't provide equivalents to Windows.Storage.ApplicationData's
+[RoamingSettings](https://learn.microsoft.com/uwp/api/windows.storage.applicationdata.roamingsettings),
+[RoamingStorageQuota](https://learn.microsoft.com/uwp/api/windows.storage.applicationdata.roamingstoragequota),
+[Version](https://learn.microsoft.com/uwp/api/windows.storage.applicationdata.version),
+[SetVersionAsync()](https://learn.microsoft.com/uwp/api/windows.storage.applicationdata.setversionasync),
+[SignalDataChanged()](https://learn.microsoft.com/uwp/api/windows.storage.applicationdata.signaldatachanged)
+and [DataChanged](https://learn.microsoft.com/uwp/api/windows.storage.applicationdata.datachanged).
+If you need this functionality use the OS APIs in `Windows.Storage.ApplicationData`.
+
+# 6. API Details
+
+```c# (but really MIDL3)
+namespace Microsoft.Windows.Storage
+{
+    [contractversion(1)]
+    apicontract ApplicationDataContract{};
+
+    /// Specifies the type of data store.
+    /// @see https://learn.microsoft.com/uwp/api/windows.storage.applicationdatalocality
+    [contract(ApplicationDataContract, 1)]
+    enum ApplicationDataLocality
+    {
+        // NOTE: Values 0-99 match Windows.Storage.ApplicationDataLocality.
+        //       Values 1000+ are unique to us here.
+        Local = 0,
+        LocalCache = 3,
+        Roaming = 1,
+        SharedLocal = 4,
+        Temporary = 2,
+        Machine = 1000,
+    };
+
+    /// Specifies options for creating application data containers or returning existing containers.
+    [contract(ApplicationDataContract, 1)]
+    enum ApplicationDataCreateDisposition
+    {
+        /// Always returns the specified container. Creates the container if it does not exist.
+        Always = 0,
+
+        // Returns the specified container only if it already exists. Raises an exception of type System.Exception if the specified container does not exist.
+        Existing,
+    };
+
+    /// Represents a container for settings. The methods and properties of this class support creating, deleting, enumerating, and traversing the container hierarchy.
+    /// @see https://learn.microsoft.com/uwp/api/windows.storage.applicationdatacontainer
+    [contract(ApplicationDataContract, 1)]
+    runtimeclass ApplicationDataContainer : Windows.Foundation.IClosable
+    {
+        /// Return the child settings containers of the settings container.
+        /// https://learn.microsoft.com/uwp/api/windows.storage.applicationdatacontainer.containers
+        IMap<String, ApplicationDataContainer> Containers { get; };
+
+        /// Return the name of the current settings container.
+        /// https://learn.microsoft.com/uwp/api/windows.storage.applicationdatacontainer.name
+        String Name { get; };
+
+        /// Return the type of the app data store that is associated with the settings container.
+        /// @see https://learn.microsoft.com/uwp/api/windows.storage.applicationdatacontainer.locality
+        ApplicationDataLocality Locality { get; };
+
+        /// Return an object that represents the settings in the settings container.
+        /// @see https://learn.microsoft.com/uwp/api/windows.storage.applicationdatacontainer.values
+        Windows.Foundation.Collections.IPropertySet Values { get; };
+
+        /// Create or open the specified settings container in the settings container.
+        /// @see https://learn.microsoft.com/uwp/api/windows.storage.applicationdatacontainer.createcontainer
+        ApplicationDataContainer CreateContainer(String name, ApplicationDataCreateDisposition disposition);
+
+        /// Delete the specified settings container, its subcontainers, and all settings in the hierarchy.
+        /// @see https://learn.microsoft.com/uwp/api/windows.storage.applicationdatacontainer.deletecontainer
+        void DeleteContainer(String name);
+    };
+
+    /// Provides access to the application data store.
+    /// @see https://learn.microsoft.com/uwp/api/windows.storage.applicationdata
+    [contract(ApplicationDataContract, 1)]
+    runtimeclass ApplicationData : Windows.Foundation.IClosable
+    {
+        /// Get an instance of ApplicationData for the current user.
+        /// @note Requires the current process has package identity.
+        /// @note This is equivalent to Windows.Storage.ApplicationData.Current if running in an AppContainer.
+        /// @note This is equivalent to Windows.Storage.ApplicationDataManager.CreateForPackageFamily(GetCurrentPackageFamilyName()) if not running in an AppContainer.
+        static ApplicationData GetDefault();
+
+        /// Get an instance of ApplicationData for the specified user.
+        /// @note This is equivalent to Windows.Storage.ApplicationDataManager.GetForUserAsync()
+        /// @see https://learn.microsoft.com/uwp/api/windows.storage.applicationdata.getforuserasync
+        static ApplicationData GetForUser(Windows.System.User user);
+
+        /// Get an instance of ApplicationData for the specified package family for the current user.
+        /// @note This is equivalent to Windows.Storage.ApplicationDataManager.CreateForPackageFamily()
+        /// @see https://learn.microsoft.com/uwp/api/windows.management.core.applicationdatamanager.createforpackagefamily
+        static ApplicationData GetForPackageFamily(String packageFamilyName);
+
+        /// Get an instance of ApplicationData for the specified unpackaged app for the current user.
+        static ApplicationData GetForUnpackaged(String publisher, String name);
+
+        /// Return true if the package family supports the machine data store.
+        Boolean IsMachinePathSupported();
+
+        /// Return the path for the local cache data store not included in backup and restore operations.
+        /// @see https://learn.microsoft.com/uwp/api/windows.storage.applicationdata.localcachefolder
+        String LocalCachePath { get; };
+
+        /// Return the path for the local data store. This location is backed up to the cloud.
+        /// @see https://learn.microsoft.com/uwp/api/windows.storage.applicationdata.localfolder
+        String LocalPath { get; };
+
+        /// Return the path for the machine data store.
+        /// @note This location is per-machine (not per-user like other *Path properties).
+        String MachinePath { get; };
+
+        /// Return the path for the roaming data store.
+        /// @warning Roaming data and settings are no longer supported after Windows 10 version 1901 (aka 19H1 aka 10.0.18363.0).
+        /// @see https://learn.microsoft.com/uwp/api/windows.storage.applicationdata.roamingfolder
+        String RoamingPath { get; };
+
+        /// Return the path for the shared data store.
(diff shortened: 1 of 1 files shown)


specs/applicationdata/ApplicationData.md
```diff
@@ -0,0 +1,428 @@
...
+      [SetVersionAsync()](https://learn.microsoft.com/uwp/api/windows.storage.applicationdata.setversionasync),
+      [SignalDataChanged()](https://learn.microsoft.com/uwp/api/windows.storage.applicationdata.signaldatachanged)
+      and
+      [DataChanged](https://learn.microsoft.com/uwp/api/windows.storage.applicationdata.datachanged). See [5.1. Platform Support](#51-platform-support) for more details.
+* **GetDefault() works for all packaged apps** -- Equivalent to .Current but works for all packaged
+  apps (not just callers running in an AppContainer). See
+  [3.3. Reliable access](#33-reliable-access).
+* **`GetForPackageFamily()`** -- Equivalent to Windows.Storage.ApplicationDataManager.CreateForPackageFamily().
+* **`GetForUser()`** -- Synchronous equivalent to GetForUserAsync().
+* **`GetForUnpackaged()`** -- Equivalent functionality for unpackaged apps. See
```
jonwis, line 77: Can `GetDefault()` just work for unpackaged, using some kind of hint in the resource section of the executable as the key to produce the path?

specs/applicationdata/ApplicationData.md
```diff
@@ -0,0 +1,428 @@
...
+  [3.5. Unpackaged app data stores](#35-unpackaged-app-data-stores).
+
+## 3.2. *Path properties
+
+New *Path equivalents to Windows.Storage.ApplicationData's *Folder but returning a `String` rather than `StorageFolder`.
+
+## 3.3. Reliable access
+
+`Windows.Storage.ApplicationData.Current` only works for packaged apps **running in an AppContainer**. Processes not running in an AppContainer (e.g. MediumIL) must use the equivalent `Windows.Management.Core.ApplicationDataManager.CreateForPackageFamily(Windows.ApplicationModel.Package.Current.Id.FamilyName)`.
```
jonwis, line 86: Please use a tool like `prettier` or your favorite markdown editing package to format the document and word-wrap at about 100 characters.

specs/applicationdata/ApplicationData.md
```diff
@@ -0,0 +1,428 @@
...
+BUILTIN\Users:(OI)(CI)(Rc,S,RD,REA,X,RA)
+S-1-15-3-2977037414-864741429-1129033548-1928484290-1803339615-1058153653-556172075:(OI)(CI)(RX)
+BUILTIN\Users:(OI)(CI)(R)
+NT SERVICE\TrustedInstaller:(I)(F)
+NT SERVICE\TrustedInstaller:(I)(CI)(IO)(F)
+NT AUTHORITY\SYSTEM:(I)(F)
+NT AUTHORITY\SYSTEM:(I)(OI)(CI)(IO)(F)
+BUILTIN\Administrators:(I)(RX)
+BUILTIN\Administrators:(I)(OI)(CI)(IO)(GR,GE)
+S-1-15-3-1024-3635283841-2530182609-996808640-1887759898-3848208603-3313616867-983405619-2501854204:(I)(RX)
```
jonwis, line 110: What are these S-1-15 SIDs? Maybe add a `//` about them?

specs/applicationdata/ApplicationData.md
```diff
@@ -0,0 +1,428 @@
...
+S-1-15-3-1024-3635283841-2530182609-996808640-1887759898-3848208603-3313616867-983405619-2501854204:(I)(OI)(CI)(IO)(GR,GE)
+```
```
jonwis, line 112: What's a use case for this? My naive read is that only TrustedInstaller and System get write access. How is an app supposed to use this?

specs/applicationdata/ApplicationData.md
```diff
@@ -0,0 +1,428 @@
...
+
+New properties and methods to access per-machine data:
+
+* ApplicationDataLocality.Machine
+* ClearMachineFolderAsync()
+* IsMachinePathSupported
+* MachineFolder
+* MachinePath
+
+**NOTE:** The per-machine APIs require the caller has the package family registered for the user -or- is running as SYSTEM.
```
jonwis, line 144: When would this result in access denied? Like, "An app fooing the bar with zot does not have access."

specs/applicationdata/ApplicationData.md
```diff
@@ -0,0 +1,428 @@
...
+Package families can optionally have a per-machine data store.
+
+### 3.4.1. Machine Path/Folder
+
+The path for a package family’s "Machine Folder" is located at `%ProgramData%\Microsoft\Windows\AppRepository\ApplicationData\...packagefamilyname...\Machine`.
+
+This directory is ACL’d similarly to a package’s System Metadata directory (`%ProgramData%\Microsoft\Windows\AppRepository\Packages\...pkgfullname...`) e.g.
+
+```
+BUILTIN\Users:(OI)(CI)(Rc,S,RD,REA,X,RA)
```
jonwis, line 101: FrancisH has come up with an interesting conditional ACE in an SDDL that grants access only when the caller is a certain package identity and running as a user. Handy way to cloak a directory from casual probing by other apps.

specs/applicationdata/ApplicationData.md
```diff
@@ -0,0 +1,428 @@
...
+* ClearMachineFolderAsync()
+* IsMachinePathSupported
+* MachineFolder
+* MachinePath
+
+**NOTE:** The per-machine APIs require the caller has the package family registered for the user -or- is running as SYSTEM.
+
+## 3.5. Unpackaged app data stores
+
+Unpackaged applications can use the ApplicationData API to access app data stores in the classic locations:
```
jonwis, line 148: How does the app specify the publisher & product information?

specs/applicationdata/ApplicationData.md
```diff
@@ -0,0 +1,428 @@
...
+|--------------------------------|-------------------------------------------------------------------|
+|ApplicationData.LocalCachePath  |%LOCALAPPDATA%\<publisher>\<product>                               |
+|ApplicationData.LocalPath       |%LOCALAPPDATA%\<publisher>\<product>                               |
+|ApplicationData.MachinePath     |%ProgramData%\<publisher>\<product>                                |
+|ApplicationData.RoamingPath     |%APPDATA%                                                          |
+|ApplicationData.TemporaryPath   |[GetTempPath2W()](https://learn.microsoft.com/windows/win32/api/fileapi/nf-fileapi-gettemppath2w)<sup>1</sup>|
+|ApplicationData.LocalSettings   |HKCU\SOFTWARE\Classes\Local Settings\Software\<publisher>\<product>|
+|ApplicationData.LocalSettings   |HKCU\SOFTWARE\<publisher>\<product>                                |
+
+<sup>1</sup>
```
jonwis, line 160: Consider not replicating the documentation here, just say "Is the return value of GetTempPath2W()[1]" / "1: See MSDN for more information on how this method works."

specs/applicationdata/ApplicationData.md
```diff
@@ -0,0 +1,428 @@
...
+
+## 5.1. Platform Support
+
+Per https://learn.microsoft.com/uwp/api/windows.storage.applicationdata.roamingfolder#remarks:
+  > Roaming data and settings is no longer supported as of Windows 11. The recommended replacement
+      is Azure App Service. Azure App Service is widely supported, well documented, reliable, and
+      supports cross-platform/cross-ecosystem scenarios such as iOS, Android and web. Settings
+      stored here no longer roam (as of Windows 11), but the settings store is still available.
+
+We provide `RoamingFolder` and `RoamingSettings` equivalents but they're only as functional as
```
jonwis, line 243: Should we take this opportunity to just not have these on the new type? If you need those, go use the in-box type.

specs/applicationdata/ApplicationData.md
```diff
@@ -0,0 +1,428 @@
...
+        void DeleteContainer(String name);
+    };
+
+    /// Provides access to the application data store.
+    /// @see https://learn.microsoft.com/uwp/api/windows.storage.applicationdata
+    [contract(ApplicationDataContract, 1)]
+    runtimeclass ApplicationData : Windows.Foundation.IClosable
+    {
+        /// Get an instance of ApplicationData for the current user.
+        /// @note Requires the current process has package identity.
```
jonwis, line 325: What happens when called from unpackaged? An exception?

specs/applicationdata/ApplicationData.md
```diff
@@ -0,0 +1,428 @@
...
+        /// @see https://learn.microsoft.com/uwp/api/windows.storage.applicationdata.getforuserasync
+        static ApplicationData GetForUser(Windows.System.User user);
+
+        /// Get an instance of ApplicationData for the specified package family for the current user.
+        /// @note This is equivalent to Windows.Storage.ApplicationDataManager.CreateForPackageFamily()
+        /// @see https://learn.microsoft.com/uwp/api/windows.management.core.applicationdatamanager.createforpackagefamily
+        static ApplicationData GetForPackageFamily(String packageFamilyName);
+
+        /// Get an instance of ApplicationData for the specified unpackaged app for the current user.
+        static ApplicationData GetForUnpackaged(String publisher, String name);
```
jonwis, line 341: See earlier note - can we use the resource APIs to pull the publisher and product names out of the exe? Or out of the sxs manifest if one is present on the app?

specs/applicationdata/ApplicationData.md
```diff
@@ -0,0 +1,428 @@
...
+
+        /// Return true if the package family supports the machine data store.
+        Boolean IsMachinePathSupported();
+
+        /// Return the path for the local cache data store not included in backup and restore operations.
+        /// @note This is equivalent to Windows.Storage.ApplicationDataManager.LocalCacheFolder().Path()
+        /// @see https://learn.microsoft.com/uwp/api/windows.storage.applicationdata.localcachefolder
+        String LocalCachePath { get; };
+
+        /// Return the path for the local data store. This location is backed up to the cloud.
```
jonwis, line 351: Are local paths really backed up to the cloud?

specs/applicationdata/ApplicationData.md
DrusTheAxe: Possibly. Interesting idea. Any prior art? In particular, any current practices doing this we could leverage? Or would it be new for new folks who wanted to use this new behavior?

Regardless, it's interesting but probably not urgent (v2?).

specs/applicationdata/ApplicationData.md
DrusTheAxe: Interesting. Color me curious; I'll contact him to learn more. Thanks for the tidbit.

specs/applicationdata/ApplicationData.md
DrusTheAxe: S-1-15-3 is the 'package capability sid'. @oldnewthing has a post about it (of course :P) https://devblogs.microsoft.com/oldnewthing/20220503-00/?p=106557

Though he's slightly mistaken: S-1-15-3 is a package (not app) capability SID. S-1-15-2-... is a PackageSID derived from a package family name, S-1-15-3-1024-... is a 'group' equivalent. The pkgdir is ACL'd with an ACE for the package's S-1-15-3-... ACE.

specs/applicationdata/ApplicationData.md
DrusTheAxe: Yes, but Admins get read/execute access and Users get read access.

An app can write content if it runs elevated, or is a Service running as LocalSystem, etc. The latter is the driving use case -- a packaged Service writing content in its package family's MachinePath for read+execute access by other processes (not necessarily in its package or even packaged at all).

And in the cases where some of the content needs access other than the default rights we create, the Service will create a subdir with the needed permissions. That's no different than some packaged apps do today -- create ApplicationData.LocalFolder\somedir ACL'd with different (non-default) permissions for their scenario.

specs/applicationdata/ApplicationData.md
DrusTheAxe: You're right. I misread the ACL not realizing Administrator and User has access (regardless packaged or not).

Access should be governed by the path's ACL. The API shouldn't make presumptions (especially not false ones). After the path's just a string, not until _access_ managed per ACLs is Access/Denied relevant.

RECOMMEND: Remove line 148.

specs/applicationdata/ApplicationData.md
DrusTheAxe: By parameter e.g. `GetForUnpackaged(publisher, product)`. The moral equivalent to `ApplicationDataManager.CreateForPackageFamily(pkgfamilyname).

I'm not aware of any existing practice for `GetDefault()` to automagically discover the caller's info. Your earlier suggestion could address that. See other comment thread.

specs/applicationdata/ApplicationData.md
DrusTheAxe: RECOMMEND: Make the change.

specs/applicationdata/ApplicationData.md
DrusTheAxe: Seemed reasonable to keep these 2 (we omitted all other roaming related properties/methods).

Roaming only stopped roaming after 19H2 and WinAppSDK runs down to RS5. Plus some apps use RoamingFolder and/or RoamingSettings even today (despite not roaming) 'cause compat (e.g. Tempo by @MikeHillberg).

Plus for unpackaged apps RoamingFolder === `%USERPROFILE%\AppData\Roaming` and RoamingSettings=`HKCU\SOFTWARE\...` and those are still supported.

specs/applicationdata/ApplicationData.md
DrusTheAxe: Yes. Same as an unpackaged process calling `Windows.Storage.ApplicationData.Current`, you get HRESULT_FROM_WIN32(APPMODEL_ERROR_NO_PACKAGE).

Unless we come up with a way for an unpackaged app to implicitly provide the information (e.g. via embedded resource, as discussed in other comment).

specs/applicationdata/ApplicationData.md
DrusTheAxe: Maybe. See other comment.

specs/applicationdata/ApplicationData.md
DrusTheAxe: Yes.

That's an exact quote from the docs for ApplicationData.LocalFolder.

LocalCacheFolder is similar to LocalFolder but it's _not_ backed up, as the docs mention:

>If the device that runs this app supports app data backup, any content placed on LocalFolder will be eligible to be backed up to the cloud. Therefore if you want to persist data and keep it local to the device you should consider using [LocalCacheFolder](https://learn.microsoft.com/en-us/uwp/api/windows.storage.applicationdata.localcachefolder?view=winrt-22621#windows-storage-applicationdata-localcachefolder).

If you have machine-specific data then you should use LocalCacheFolder, e.g. a license key for the Fabrikam MuffinMaker application which is tied to the machine it's installed on. No point having backup restore such data to another machine as such files could be restored but their content would be wrong for the other machine.

specs/applicationdata/ApplicationData.md
DrusTheAxe: Sorry about that. Thanks for the reminder. Fixed.

specs/applicationdata/ApplicationData.md
jonwis: Consider updating the text so the "why should you care" is more clear. Otherwise apps will think "oh hey I can easily break the user security model and share stuff with other users on the same machine!"

specs/applicationdata/ApplicationData.md
jonwis: Personally, I'd just remove them here as "go forward" apps won't care and we don't want to give people false hope.

specs/applicationdata/ApplicationData.md
jonwis: Do you know of devices that support app data back up? I'm surprised by this.

specs/applicationdata/ApplicationData.md
jonwis: See the implementation of Limited Access Features where you have to put a special field in the resource section for the executable.  https://learn.microsoft.com/en-us/windows/win32/menurc/versioninfo-resource is related but the ProductName and CompanyName are human-readable, not identifiers (and localized, often.)

DrusTheAxe: So we can define some named resources and fetch via

```
std::wstring GetPublisher()
{
    HRSRC resource{ ::FindResourceExW(nullptr, L"ApplicationData", "Publisher", 0)
    THROW_LAST_ERROR_IF_NULL(resource);
    const DWORD dataSize{ ::SizeofResource(nullptr, resource) };
    THROW_LAST_ERROR_IF(size == 0);
    auto data{ reinterpret_cast<wchar_t*>(::LoadResource(nullptr, resource));
    THROW_LAST_ERROR_IF_NULL(data);
    return std::wstring(data, dataSize / sizeofof(*data));
}
```

and likewise for "Product". Nice! Thanks Jon

specs/applicationdata/ApplicationData.md
```diff
@@ -0,0 +1,432 @@
...
+
+* **No roaming functionality** -- We don't provide equivalents to Windows.Storage.ApplicationData's
+  [RoamingSettings](https://learn.microsoft.com/uwp/api/windows.storage.applicationdata.roamingsettings),
+  [RoamingStorageQuota](https://learn.microsoft.com/uwp/api/windows.storage.applicationdata.roamingstoragequota),
+  [Version](https://learn.microsoft.com/uwp/api/windows.storage.applicationdata.version),
+  [SetVersionAsync()](https://learn.microsoft.com/uwp/api/windows.storage.applicationdata.setversionasync),
+  [SignalDataChanged()](https://learn.microsoft.com/uwp/api/windows.storage.applicationdata.signaldatachanged)
+  and
+  [DataChanged](https://learn.microsoft.com/uwp/api/windows.storage.applicationdata.datachanged).
+  See [5.1. Platform Support](#51-platform-support) for more details.
```
oldnewthing, line 72: Also, no SharedLocal functionality (shared per user by all apps from the same publisher). Is that basically the same as the new Machine folder?

specs/applicationdata/ApplicationData.md
```diff
@@ -0,0 +1,432 @@
...
+        /// @see https://learn.microsoft.com/uwp/api/windows.storage.applicationdata.getforuserasync
+        static ApplicationData GetForUser(Windows.System.User user);
+
+        /// Get an instance of ApplicationData for the specified package family for the current user.
+        /// @note This is equivalent to Windows.Storage.ApplicationDataManager.CreateForPackageFamily()
+        /// @see https://learn.microsoft.com/uwp/api/windows.management.core.applicationdatamanager.createforpackagefamily
+        static ApplicationData GetForPackageFamily(String packageFamilyName);
+
+        /// Get an instance of ApplicationData for the specified unpackaged app for the current user.
+        static ApplicationData GetForUnpackaged(String publisher, String name);
```
oldnewthing, line 345: Is there some correspondence between PackageFamilyName and Publisher/Name?

specs/applicationdata/ApplicationData.md
oldnewthing: I dimly recall that we stopped backing them up, but I may have gotten that confused with roaming.

specs/applicationdata/ApplicationData.md
```diff
@@ -0,0 +1,432 @@
...
+        String LocalCachePath { get; };
+
+        /// Return the path for the local data store. This location is backed up to the cloud.
+        /// @note This is equivalent to Windows.Storage.ApplicationDataManager.LocalFolder().Path()
+        /// @see https://learn.microsoft.com/uwp/api/windows.storage.applicationdata.localfolder
+        String LocalPath { get; };
+
+        /// Return the path for the machine data store.
+        /// @note This location is per-machine (not per-user like other *Path properties).
+        String MachinePath { get; };
```
oldnewthing, line 362: What happens if you ask for MachinePath and MachineFolder when IsMachinePathSupported is false?

specs/applicationdata/ApplicationData.md
```diff
@@ -0,0 +1,432 @@
...
+        /// Get an instance of ApplicationData for the specified package family for the current user.
+        /// @note This is equivalent to Windows.Storage.ApplicationDataManager.CreateForPackageFamily()
+        /// @see https://learn.microsoft.com/uwp/api/windows.management.core.applicationdatamanager.createforpackagefamily
+        static ApplicationData GetForPackageFamily(String packageFamilyName);
+
+        /// Get an instance of ApplicationData for the specified unpackaged app for the current user.
+        static ApplicationData GetForUnpackaged(String publisher, String name);
+
+        /// Return true if the package family supports the machine data store.
+        Boolean IsMachinePathSupported();
```
oldnewthing, line 348: Should this be a property instead of a method?

specs/applicationdata/ApplicationData.md
```diff
@@ -0,0 +1,432 @@
...
+
+|Packaged                        |Unpackaged                                                         |
+|--------------------------------|-------------------------------------------------------------------|
+|ApplicationData.LocalCachePath  |%LOCALAPPDATA%\<publisher>\<product>                               |
+|ApplicationData.LocalPath       |%LOCALAPPDATA%\<publisher>\<product>                               |
+|ApplicationData.MachinePath     |%ProgramData%\<publisher>\<product>                                |
+|ApplicationData.RoamingPath     |%APPDATA%                                                          |
+|ApplicationData.TemporaryPath   |[GetTempPath2W()](https://learn.microsoft.com/windows/win32/api/fileapi/nf-fileapi-gettemppath2w)<sup>1</sup>|
+|ApplicationData.LocalSettings   |HKCU\SOFTWARE\Classes\Local Settings\Software\<publisher>\<product>|
+|ApplicationData.LocalSettings   |HKCU\SOFTWARE\<publisher>\<product>                                |
```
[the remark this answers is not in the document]
DrusTheAxe: ApplicationData is per-user per-packagefamily. %USERPROFILE%\AppData\Local\Packages\...pkgfamilyname... is the root for a user's applicationdata for a package family. Under that is where .LocalFolder, .LocalCacheFolder and other resources are stored.

specs/applicationdata/ApplicationData.md
DrusTheAxe: No. Key difference is Machine = per-machine, not per-user. SharedLocal is ACL'd for access by apps across packages by the publisher, but still managed per-user. Machine isn't per-user.

SharedLocal = per-user per-publisher
Machine = per-machine per-packagefamily

I guess you could say Machine a higher level escalation of Local
* Local = per-user per-packagefamily
* Machine = per-machine per-packagefamily

SharedLocal is a bit different axis from that, as SharedLocal is still per-user but being per-publisher it's cross-packagefamily. Given e.g. FantastyFootball, FantasyHockey and FantasyBaseball package families all by publisher=FabrikamSports &lt;g&gt; and you and I both have all 3 packages registered the machine will have...
* 6 Local paths (3 in your %USERPROFILE%, 3 in mine)
* 3 Machine paths (assuming all 3 packages opt-in via their appxmanifest)
* 2 SharedLocal paths (1 in your %USERPROFILE%, 1 in mine)

So yes, there's some vague similiarities of SharedLocal and Machine in that they offer wider access than just the 1 package family for user, but SharedLocal's still a per-user thing. Machine isn't.

The scenario for Machine is I need to store some files on the machine every user may need to Read but could be sizable, so I don't want every user having their own copy. I can make 1 copy on the machine, ACL'd so only my PackagedService will modify and manage, and everyone else can read and share the 1 copy. In the simple extreme case, a Citrix, VMware or AVD system could have 100 users concurrently logged in using the system. If I use SharedLocal to share these large files I'd have 100 copies. Using Machine I can have 1.

specs/applicationdata/ApplicationData.md
DrusTheAxe: PackageFamilyName = package Name + PublisherId (which is a hashed derivative of Publisher)

The unpackaged moral equivalent is Publisher + Product

For instance, if one made an app packaged and unpackaged...
* Packaged PackageFamilyName = iTunes_1234567890abc (or whatever the PublisherId from Publisher=CN=Apple, ...)
* Unpackaged Publisher + Product = Apple + iTunes

Are there better terms for unpackaged?

specs/applicationdata/ApplicationData.md
DrusTheAxe: If IsMachinePathSupported()=false
* MachinePath returns empty string (std::wstring void ctor)
* MachineFolder returns null

In the spirit of "GetPrinters()" on a system that doesn't support printers, you get back a valid empty result. Likewise, MachinePath/Folder give you a valid empty result you can't do more with.

specs/applicationdata/ApplicationData.md
DrusTheAxe: Good point!

RECOMMEND: Make IsMachinePathSupported a get property

specs/applicationdata/ApplicationData.md
```diff
@@ -0,0 +1,428 @@
...
```
oldnewthing, line 162: Um, `LocalSettings` is in this table twice, with two different values. And we're missing SharedLocal.

specs/applicationdata/ApplicationData.md
```diff
@@ -0,0 +1,428 @@
...
+
+        /// Remove all data from the specified data store.
+        /// @see https://learn.microsoft.com/uwp/api/windows.storage.applicationdata.clearasync
+        Windows.Foundation.IAsyncAction ClearAsync(ApplicationDataLocality locality);
+
+        /// Remove all data from the shared data store in the specified subfolder.
+        /// @see SharedLocalPath
+        /// @see SharedLocalFolder
+        /// @see https://learn.microsoft.com/uwp/api/windows.storage.applicationdata.clearpublishercachefolderasync
+        Windows.Foundation.IAsyncAction ClearPublisherCacheFolderAsync(String folderName);
```
oldnewthing, line 421: This is inconsistent with the Windows.Storage API (a hidden broken egg).

In Windows.Storage, the shared publisher directory is structured like this:

* `%LOCALAPPDATA%\Publishers\<publisher>\SharedLocal`: Returned by SharedLocalFolder
* `%LOCALAPPDATA%\Publishers\<publisher>\<blah>`: Returned by GetPublisherCacheFolder("\<blah\>") and cleared by ClearPublisherCacheFolderAsync("\<blah\>")

But given the description "remove all data from the shared data store in the specified subfolder", it looks like the WinAppSDK version is

* `%MYSTERYROOTDIRECTORY%`: Returned by SharedLocalFolder
* `%MYSTERYROOTDIRECTORY%\<blah>`: Returned by GetPublisherCacheFolder("\<blah\>") and cleared by ClearPublisherCacheFolderAsync("\<blah\>")

In Windows.Storage, the GetPublisherCacheFolder is not visible to SharedLocal, but in WinAppSDK, they are. This can lead to mysterious bugs like "The old code deleted everything in the SharedLocal directory. That used to leave the publisher cache folders alone, but now it wipes them out!"

specs/applicationdata/ApplicationData.md
```diff
@@ -0,0 +1,428 @@
...
+
+        /// Return the path for the shared data store for the publisher of the app.
```
oldnewthing, line 423: GetPublisherCachePath and GetPublisherCacheFolder comments seem to be wrong, since they say that they ignore the `folderName` parameter and just give you the SharedLocalPath / SharedLocalFolder.

specs/applicationdata/ApplicationData.md
DrusTheAxe: RECOMMEND: Add comment per Raymond's feedback

specs/applicationdata/ApplicationData.md
DrusTheAxe: Sounds good to me

RECOMMEND: Remove `RoamingFolder`, `RoamingPath` and `RoamingSettings`

specs/applicationdata/ApplicationData.md
DrusTheAxe: Document - of course.

And yes, consistent with SharedLocalFolder and other "null"-esque experiences.

specs/applicationdata/ApplicationData.md
```diff
@@ -0,0 +1,428 @@
...
+
+Unpackaged applications can use the ApplicationData API to access app data stores in the classic locations:
+
+|Packaged                        |Unpackaged                                                         |
+|--------------------------------|-------------------------------------------------------------------|
+|ApplicationData.LocalCachePath  |%LOCALAPPDATA%\<publisher>\<product>                               |
+|ApplicationData.LocalPath       |%LOCALAPPDATA%\<publisher>\<product>                               |
+|ApplicationData.MachinePath     |%ProgramData%\<publisher>\<product>                                |
+|ApplicationData.RoamingPath     |%APPDATA%                                                          |
+|ApplicationData.TemporaryPath   |[GetTempPath2W()](https://learn.microsoft.com/windows/win32/api/fileapi/nf-fileapi-gettemppath2w)<sup>1</sup>|
```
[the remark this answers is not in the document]
DrusTheAxe: Any idea where we've provided _written_ guidance to unpackaged apps?

"Get Temp dir" (and its various API variations over the years) is a common practice for decades and I recall the "store your data/config under HKCU\SOFTWARE\company\product (or %LOCALAPPDATA%)" guidance when Win95 came out, and cultural norms and common guidance and practice since. But I don't know where it's formally documented these days. Wondering if that says anything about TEMP, and if not if there's any de facto common practice re temp (I've seen both styles - subdir and general mosh pit).

I'm a little hesitant to dictate ApplicationData.GetForUnpackaged("x", "y").TemporaryFolder == GetTempPath2W() + "\x\y" as someone will have to create that directory and, presumably destroy it. We could create it on 1st use but who destroys it? Natural expectation is whoever creates it, but won't have any idea when it's "done" and can be removed, and if the app doesn't create it would they routinely think to remove it? Or do we just crate large swathes of temp files hanging around 'forever' (or if/until someone runs DiskCleanupWizard etc)?

specs/applicationdata/ApplicationData.md
```diff
@@ -0,0 +1,428 @@
...
+## 3.5. Unpackaged app data stores
+
+Unpackaged applications can use the ApplicationData API to access app data stores in the classic locations:
+
+|Packaged                        |Unpackaged                                                         |
+|--------------------------------|-------------------------------------------------------------------|
+|ApplicationData.LocalCachePath  |%LOCALAPPDATA%\<publisher>\<product>                               |
+|ApplicationData.LocalPath       |%LOCALAPPDATA%\<publisher>\<product>                               |
+|ApplicationData.MachinePath     |%ProgramData%\<publisher>\<product>                                |
+|ApplicationData.RoamingPath     |%APPDATA%                                                          |
```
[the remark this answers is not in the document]
DrusTheAxe: Correct. Typo in the spec

RECOMMEND: Fix (if RoamingPath doesn't go away)

specs/applicationdata/ApplicationData.md
DrusTheAxe: 2nd one is a typo, should be `.RoamingSettings`

In the classic Win32 world going years back, `HKCU\SOFTWARE` roams except for the `SOFTWARE\Classes\Local Settings` subtree.

specs/applicationdata/ApplicationData.md
DrusTheAxe: Good point. Intent is PublisherCacheFolder and SharedLocalFolder are handled per Windows.Storage.ApplicationData (since we do nothing special here, just passthru to the OS to handle them).

RECOMMEND: Fix spec

specs/applicationdata/ApplicationData.md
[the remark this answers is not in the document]
oldnewthing: I think Storage Sense runs around cleaning up stale files in TEMP directories, as well as cleaning up empty directories. I mean, *somebody* has to. Otherwise, those files just accumulate indefinitely!

Having a practical difference between packaged and unpackaged apps seems wrong. "I upgraded my app to packaged, and stuff doesn't work any more." Or worse "I downgraded my app from packaged to unpackaged, and now my app crashes (because it now encounters sharing violations on files in its TemporaryFolder)."

specs/applicationdata/ApplicationData.md
DrusTheAxe: "Identity for unpackaged" is an interesting ongoing topic.

RECOMMEND: Defer this to a future date.

specs/applicationdata/ApplicationData.md
DrusTheAxe: CONSIDER: Feature enhancement based on FrancisH example

specs/applicationdata/ApplicationData.md
DrusTheAxe: CONSIDER: Add more text about WHY and consider-security-implications

specs/applicationdata/ApplicationData.md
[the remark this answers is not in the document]
DrusTheAxe: CONSIDER: Unpackaged `TemporaryPath` = `GetTempPath2W()\publisher\product`

specs/applicationdata/ApplicationData.md
DrusTheAxe: RECOMMEND: User 'publisher' and 'product'

specs/applicationdata/ApplicationData.md
DrusTheAxe: RECOMMEND: Follow up with Windows.Storage.ApplicationData owner to possibly update the docs if the 'backup' comment is stale. If that changes also update here to match.

DrusTheAxe (commented):
Updates per API Review feedback

specs/applicationdata/ApplicationData.md
```diff
@@ -0,0 +1,430 @@
...
+* **`GetForUnpackaged()`** -- Equivalent functionality for unpackaged apps. See
+  [3.5. Unpackaged app data stores](#35-unpackaged-app-data-stores).
+
+## 3.2. *Path properties
+
+New *Path equivalents to Windows.Storage.ApplicationData's *Folder but returning a `String` rather than `StorageFolder`.
+
+## 3.3. Reliable access
+
+`Windows.Storage.ApplicationData.Current` only works for packaged apps **running in an
```
ahmed605, line 87: `Windows.Storage.ApplicationData.Current` works for any packaged app, it doesn't require AC, `Windows.Management.Core.ApplicationDataManager.CreateForPackageFamily` is for accessing app data for *other* packages

riverar: Stumbled on this old comment while helping someone and want to clarify:
* Packaged applications running in AppContainer should use `Windows.Storage.ApplicationData`
* Packaged applications not running in AppContainer (e.g., runFullTrust) can use `Windows.Storage.ApplicationDataManager` or classical methods
* Unpackaged applications should stick to classical methods

ahmed605: @riverar but `Windows.Storage.ApplicationData` should work for packaged apps not using AC too, I remember trying it when I made that comment and it was working

riverar: Yeah, it'll only work in some scenarios on a best-effort (not guaranteed) basis; Microsoft has confirmed it's not designed for use outside AC. Just wanted to leave a note to prevent folks from wondering why this is a high-crash API when they roll out to production. 😅

PR #4057, opened 2024-01-02
Microsoft.Windows.Storage.ApplicationData

DrusTheAxe:
Implementation of Microsoft.Storage.ApplicationData API Spec #4056

Issue https://github.com/microsoft/WindowsAppSDK/issues/2639

Microsoft-internal tasks [22308619](https://task.ms/22308619), [30002314](https://task.ms/30002314), [39401257](https://task.ms/39401257), [39414843](https://task.ms/39414843), [40050113](https://task.ms/40050113) and [48332357](https://task.ms/48332357).

vFuture (ApplicationDataContract v2, vFuture)
* GetForUnpackaged() backed by an unpackaged variant of M.W.S.ApplicationData.cpp + M.W.S.ApplicationDataContainer.cpp


dev/ApplicationData/ApplicationData.idl
```diff
@@ -0,0 +1,190 @@
...
+        /// @see LocalFolder
+        /// @see RoamingPath
+        /// @see RoamingFolder
+        /// @see TemporaryPath
+        /// @see TemporaryFolder
+        [method_name("ClearAsync")]
+        Windows.Foundation.IAsyncAction ClearAllAsync();
+
+        /// Remove all data from the specified data store.
+        Windows.Foundation.IAsyncAction ClearAsync(ApplicationDataLocality locality);
```
riverar, line 170: Is this `method_name` attribute needed?

I can't remember how does this work out in terms of ABI. We have two methods here that are going to want to be `ClearAsync`--is that expected? (Not all languages support overloading.)

DrusTheAxe: Yep. ByDesign.

Overload by arity so overloading works but ABI has different names (since WinRT is COM at the ABI level and COM doesn't support method overloading). How that's expressed by language projections depends on the projection, e.g. C++/WinRT supports method overloading so callers see

```
ClearAsync()
ClearAsync(Locality)
```

as they'd expect. For languages that don't support method overloading I'd expect they'd expose the ABI method names but that's really up to the projection tech.

P.S. I never remember if the method is the ABI and method_name() is the overloaded name or vice verse but MIDL takes this fine and errors the other way so it's hard to get it wrong :P

riverar: Right, but in terms of ABI, I think:
* ClearAllAsync requests emission as `ClearAsync`
* ClearAsync default requests emission as `ClearAsync`

So it seems like this creates a conflict that might be handled differently per projection. (I haven't tried it yet.)

riverar, line 170: In terms of ABI
* `ClearAllAsync` requests emission as `ClearAsync`
* `ClearAsync` default requests emission as `ClearAsync`

This creates a conflict that is resolved (via midl) as:
* `ClearAllAsync()` becomes `ClearAsync()`
* `ClearAsync(locality)` becomes `ClearAsync2(locality)`

This is not ideal for languages that don't support overloading (e.g., Rust) and seems entirely avoidable with a better naming scheme that more closely aligns with the behavior of the function.

Suggestions:
* Remove `method_name` attribute to keep the `ClearAllAsync` name. The intent--to clear all localities--is better communicated with the All qualifier. Without it, it's unclear which localities are chosen and requires the developer to consult the docs.
* Leave `ClearAsync` as-is

DrusTheAxe: Hmmm. Good point. Well, at least it's consistent with the OS API

Thanks for the feedback

DrusTheAxe: FYI per API Review we decided to drop ClearAllAsync and ClearMachineAsync and just have ClearAsync(locality).

The Clear-vs-ClearAll was a Win8-era convenience but as ApplicationData's evolved over the years 'All' lost some meaning (All...as-of-10+years-ago...). If we were doing it from scratch today we wouldn't have the overload and gasp! We are doing it from scratch (sorta) :-)

riverar: Thanks for the heads up, sounds reasonable.

rhuang-msft (approved):

dev/ApplicationData/ApplicationData.idl
```diff
@@ -0,0 +1,167 @@
...
+        /// @see https://learn.microsoft.com/uwp/api/windows.storage.applicationdatacontainer.locality
+        ApplicationDataLocality Locality { get; };
+
+        /// Return an object that represents the settings in the settings container.
+        /// @see https://learn.microsoft.com/uwp/api/windows.storage.applicationdatacontainer.values
+        Windows.Foundation.Collections.IPropertySet Values { get; };
+
+        /// Create or open the specified settings container in the settings container.
+        /// @see https://learn.microsoft.com/uwp/api/windows.storage.applicationdatacontainer.createcontainer
+        ApplicationDataContainer CreateContainer(String name, ApplicationDataCreateDisposition disposition);
```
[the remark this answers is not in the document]
DrusTheAxe: This is Create not Open for parity with the Windows API. [Windows.Storage.ApplicationDataContainer](https://learn.microsoft.com/uwp/api/windows.storage.applicationdatacontainer?view=winrt-26100) has CreateContainer and no OpenContainer.

P.S. Even if no precedent Create is the right verb e.g.

* RegOpenKeyEx() opens an existing regkey or error
* RegCreateKeyEx() creates a regkey if new or opens it exists

Common pattern across Windows APIs -- create = create + open, with variations over open (open, open but return disposition=created|opened, etc)

dev/ApplicationData/ApplicationData.idl
```diff
@@ -0,0 +1,167 @@
+// Copyright (c) Microsoft Corporation and Contributors.
+// Licensed under the MIT License.
+
+#include <TerminalVelocityFeatures-ApplicationData.h>
+
+namespace Microsoft.Windows.Storage
+{
+    [contractversion(2)]
```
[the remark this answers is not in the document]
DrusTheAxe: Because `GetForUnpackaged()` isn't ready at this time so rather than hold up the rest (Perfect is Enemy of the Good, etc) this is marked v2 and experimental to be addressed later

ghost1372:
Hi @DrusTheAxe 
Can we use this new api in. Next release? (v1.6-exp2 ?)

PR #4065, opened 2024-01-05
Fix Variables Field Being Defined Twice in Nightly Yml Template

kythant:
Fix Variables Field Being Defined Twice in Nightly Yml Template


alexlamtest (approved):
LGTM

Issue #2597, opened 2022-06-08
PublishSingleFile with WindowsAppSDKSelfContained produces non runnable application


Balkoth:
### Describe the bug

If `PublishSingleFile` and `WindowsAppSDKSelfContained` are set to true while publishing, an exe is produced which crashes on startup.

### Steps to reproduce the bug

1. Add <WindowsAppSDKSelfContained>true</WindowsAppSDKSelfContained> to your project file
2. Add <PublishSingleFile>true</PublishSingleFile> to your publish file
3. Publish the application
4. Try to run and see it crashing

### Packaging type

Unpackaged

### Windows version

Windows 10 version 21H2 (19044, November 2021 Update)

### IDE

Visual Studio 2022

### Additional context

NuGet package version is 1.1.0

DefaultRyan:
I've created an internal task to track this: https://task.ms/39946393 and assigning to @DrusTheAxe

DrusTheAxe:
Do you have Microsoft.WindowsAppRuntime.dll as a file along with your "single file"?

Does your project define `<IncludeNativeLibrariesForSelfExtract>true</IncludeNativeLibrariesForSelfExtract>`?

Do you set the property `IncludeNativeLibrariesForSelfExtract`?

------

PublishSingleFile has variation of behavior. Depending how you use it you may or may not run into problems

https://docs.microsoft.com/en-us/dotnet/core/deploying/single-file/overview#output-differences-from-net-3x
>Bundling all application-dependent files into a single binary provides an application developer with the attractive option to deploy and distribute the application as a single file. This deployment model has been available since .NET Core 3.0 and has been enhanced in .NET 5. Previously in .NET Core 3.0, when a user runs your single-file app, .NET Core host first extracts all files to a directory before running the application. **.NET 5 improves this experience by directly running the code without the need to extract the files from the app.**

Which would be problematic as [DllImport(Microsoft.WindowsAppRuntime.dll,...)] == LoadLibrary(Microsoft.WindowsAppRuntime.dll) but .NET 5's not quite that simple as noted further down

>Output differences from .NET 3.x
[...]

PR #4067, opened 2024-01-06
PublishSingleFile dll and resources.pri redirection support

Scottj1s:
PublishSingleFile creates a self-extracting exe that deposits embedded files into a temporary folder.  As noted in #2597, this causes dll SxS redirection issues, as well as resources.pri lookup issues.  The support here (which requires coordinated changes in aggregator) sets a base directory env var, which is used to configure the generated self-contained app.manifest entries to enable custom SxS redirection via the undocumented loadFrom attribute, e.g.:
`<asmv3:file name="microsoft.ui.xaml.dll" loadFrom="%MICROSOFT_WINDOWSAPPRUNTIME_BASE_DIRECTORY%microsoft.ui.xaml.dll">`

### DIFF
--- a/dev/UndockedRegFreeWinRT/catalog.cpp
+++ b/dev/UndockedRegFreeWinRT/catalog.cpp
@@ -1,9 +1,10 @@
-﻿// Copyright (c) Microsoft Corporation and Contributors.
+// Copyright (c) Microsoft Corporation and Contributors.
 // Licensed under the MIT License.
 
 #include <pch.h>
 
 #include "catalog.h"
+#include "Microsoft.Utf8.h"
 #include "TypeResolution.h"
 
 #include <activation.h>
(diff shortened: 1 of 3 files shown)


dev/UndockedRegFreeWinRT/catalog.cpp
JesseCol, line 1: These changes to UndockedRegFreeWinRT -- do we know that the Windows implementation behaves the same way?  (on all supported platforms)

dev/UndockedRegFreeWinRT/catalog.cpp
```diff
@@ -185,16 +208,48 @@ HRESULT ParseRootManifestFromXmlReaderInput(IUnknown* input)
...
     PCWSTR localName = nullptr;
     PCWSTR value = nullptr;
-    hr = xmlReader->MoveToAttributeByName(L"name", nullptr);
-    RETURN_HR_IF(HRESULT_FROM_WIN32(ERROR_SXS_MANIFEST_PARSE_ERROR), hr != S_OK);
-    RETURN_IF_FAILED(xmlReader->GetValue(&value, nullptr));
-    std::wstring fileName{ !value ? L"" : value };
-    RETURN_HR_IF(HRESULT_FROM_WIN32(ERROR_SXS_MANIFEST_PARSE_ERROR), fileName.empty());
+    std::wstring fileName;
     auto locale = _create_locale(LC_ALL, "C");
+    // Using this pattern intead of calling multiple MoveToAttributeByName improves performance
```
JesseCol, line 215: Typo: "instead"

dev/UndockedRegFreeWinRT/catalog.cpp
```diff
@@ -185,16 +208,48 @@ HRESULT ParseRootManifestFromXmlReaderInput(IUnknown* input)
...
+    HRESULT hr = xmlReader->MoveToFirstAttribute();
```
JesseCol, line 216: Presumably, since we change the way we parse things here, we could _possibly_ make a 1.4 app that was working fine fail at runtime when built with 1.5, correct?  I'm not saying I see such a problem, I'm just trying to make sure I understand the stakes.

JesseCol (approved):

dev/UndockedRegFreeWinRT/catalog.cpp
Scottj1s: Yes, in theory

dev/UndockedRegFreeWinRT/catalog.cpp
Scottj1s: Yes, I've tested with SystemRegFreeWinRT as well (disabling UndockedRegFreeWinRT)

PR #4069, opened 2024-01-08
Fix XML Syntax Error in RuntimeIdentifiers Conditions

ShashankNay:
**Issue**: The current XML configuration in our project file uses a less-than sign (<) in the Condition attribute for RuntimeIdentifiers. This syntax is causing XML parsing errors because the less-than sign is interpreted as the start of a new tag.

**Solution**: This pull request resolves the XML parsing issue by replacing the less-than sign (<) with its XML entity equivalent &lt;. This change ensures that the XML parser correctly interprets the character as part of the attribute value, rather than as XML syntax.

The reference to the `NuGetPackageInstaller` WizardExtension is also removed as the Dynamic Retrieval logic has not been cherry-picked into a release branch from `main`

### DIFF
--- a/dev/VSIX/ProjectTemplates/Desktop/CSharp/UnitTestApp/WinUI.Desktop.Cs.UnitTestApp.vstemplate
+++ b/dev/VSIX/ProjectTemplates/Desktop/CSharp/UnitTestApp/WinUI.Desktop.Cs.UnitTestApp.vstemplate
@@ -58,6 +58,10 @@
     <Assembly>Microsoft.VisualStudio.WinRT.TemplateWizards, Version=16.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a</Assembly>
     <FullClassName>Microsoft.VisualStudio.WinRT.TemplateWizards.UpdatePublisherInManifestWizard</FullClassName>
   </WizardExtension>
+  <WizardExtension>
+    <Assembly>WindowsAppSDK.Cs.Extension.Dev17, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null</Assembly>
+    <FullClassName>WindowsAppSDK.TemplateUtilities.NuGetPackageInstaller</FullClassName>
+  </WizardExtension>
   <WizardData>
     <packages>
       <package id="Microsoft.WindowsAppSDK" />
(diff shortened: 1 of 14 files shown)


dev/VSIX/ProjectTemplates/Desktop/CSharp/UnitTestApp/WinUI.Desktop.Cs.UnitTestApp.vstemplate
```diff
@@ -58,10 +58,6 @@
     <Assembly>Microsoft.VisualStudio.WinRT.TemplateWizards, Version=16.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a</Assembly>
     <FullClassName>Microsoft.VisualStudio.WinRT.TemplateWizards.UpdatePublisherInManifestWizard</FullClassName>
   </WizardExtension>
-  <WizardExtension>
```
kmahone, line 61: Why is this not needed for this project?

ShashankNay: The XML Syntax error exists both in `main` and `release/1.4-stable`. The [Wizard](https://github.com/microsoft/WindowsAppSDK/commit/514f12a81ea0944e52c97b46552c8f969663c25d), however, only exists in `main`. To fix the error in `release/1.4-stable` , I thought it would make sense to remove this here in `main` so that with one cherry-pick we could take this into `release/1.4-stable`. We would need to include the Dynamic Retrieval logic to `release/1.4-stable` from `main` soon, and then we could reintroduce this reference back into the project -  though I'm not sure if both can be done in time.  If this line remains in `release/1.4-stable` without the Wizard itself, it leads to a file not found exception.

kmahone: Ok, so this NuGetPackageInstaller should be there in main, but not in release/1.4. It currently is in both places. You plan to fix this by removing it in main with this PR and then cherry-picking that change to release/1.4. And then creating a separate follow up PR to add it back to main? Is that accurate?

I think you are over-complicating things. Revert this change from this PR since it is not needed in main. Then create a separate PR directly into release/1.4 that removes this NuGetPackageInstaller.

ShashankNay: You're right, that sounds like a better course of action. I didn't know I could PR directly into release/1.4

Scottj1s: agreed

Scottj1s (approved):

PR #4083, opened 2024-01-16
48525090 - Skipping a test case to temporarily avoid hitting this bug

alexlamtest:
Temporarily skip test case FullLifecycle_FilePathLifetime_Frameworks_WindowsAppRuntime_MathAdd_Elevated() to unblock pipeline runs.
---------

WARNING:

### DIFF
--- a/test/DynamicDependency/Test_Win32/Test_Win32.h
+++ b/test/DynamicDependency/Test_Win32/Test_Win32.h
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation and Contributors.
+// Copyright (c) Microsoft Corporation and Contributors.
 // Licensed under the MIT License.
 
 #pragma once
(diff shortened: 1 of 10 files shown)


Scottj1s (approved):
thanks!

DrusTheAxe (approved):
Enhanced PR built locally and tested. A bit broader brush than I'd like but necessary for now

test/DynamicDependency/Test_Win32/Test_Win32.h
```diff
@@ -246,6 +246,11 @@ namespace Test::DynamicDependency
 
         TEST_CLASS_SETUP(Setup_Elevated)
         {
+            if (/*TODO ::WindowsVersion::IsWindows11_23H1OrGreater() &&*/ !::Test::TAEF::IsEnabled(L"DynamicDependency.23H1"))
```
Scottj1s, line 249: intend to comment out the IsWindows11_23H1OrGreater() check? (same below)

DrusTheAxe: Yes because that checks for a new export in 23H2

right now the parameter isn't set (it's an override for manual investigation) so this is always if (!false) == if (true) and thus we always Skip here

That commented out block should make that if (OS>=23H2 && !false) == (OS>=23H2) but the new symbol it's probing for is _really_ new - newer than the test images we're currently using. So for now we always Skip here. In a few weeks when we have a new test image with the new code we can uncomment this comment and change the test from

if (!override) == always Skip

to

if (OS>=23H2 and !override) == always skip **23H2**

At least until we figure out why that 1 test image is flaky here

PR #4099, opened 2024-01-20
Restore PublishSingleFile with fix for failing Aggregator tests

Scottj1s:


kythant (approved):

dev/MRTCore/mrt/Core/src/MRM.cpp
```diff
@@ -989,6 +989,12 @@ STDAPI MrmGetFilePathFromName(_In_opt_ PCWSTR filename, _Outptr_ PWSTR* filePath
     PCWSTR filenameToUse = filename;
     if (filename == nullptr || *filename == L'\0')
     {
+        wil::unique_cotaskmem_string baseDir;
+        if (SUCCEEDED(wil::TryGetEnvironmentVariableW(L"MICROSOFT_WINDOWSAPPRUNTIME_BASE_DIRECTORY", baseDir)) && baseDir)
```
Scottj1s, line 993: added check for empty baseDir to prevent StringCchLengthW failure on line 996 - TryGetEnvironmentVariableW succeeds when the env var isn't found, returning empty string

PR #4132, opened 2024-01-30
Make Nightly Build Non-Official

kythant:
Make Nightly Build Non-Official. Nightly builds do not need to be on official.


alexlamtest (approved):
If there is further motivation to make this change than the generic reason "if we don't need to be official then don't", can we document that in the PR Overview?

kythant:
> If there is further motivation to make this change than the generic reason "if we don't need to be official then don't", can we document that in the PR Overview?

Done!

PR #4144, opened 2024-01-31
PreFast compliance fixes - Installer

pratikone:
These set of changes are flagged by PreFast static analysis tool. Fixing for that.

### DIFF
--- a/installer/dev/tracelogging.h
+++ b/installer/dev/tracelogging.h
@@ -30,6 +30,7 @@ void __stdcall wilResultLoggingCallback(const wil::FailureInfo& failure) noexcep
 
 class WindowsAppRuntimeInstaller_TraceLogger final : public wil::TraceLoggingProvider
 {
+    #pragma prefast(disable:6387, "PreFast gets confused over tracelogging parameters")
     IMPLEMENT_TRACELOGGING_CLASS(
         WindowsAppRuntimeInstaller_TraceLogger,
         "Microsoft.WindowsAppRuntimeInstaller",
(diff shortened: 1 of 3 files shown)


alexlamtest (approved):

installer/dev/tracelogging.h
```diff
@@ -30,6 +30,7 @@ void __stdcall wilResultLoggingCallback(const wil::FailureInfo& failure) noexcep
 
 class WindowsAppRuntimeInstaller_TraceLogger final : public wil::TraceLoggingProvider
 {
+    #pragma prefast(disable:6387, "PreFast gets confused over tracelogging parameters")
```
JesseCol, line 33: I think "disable" disables the warning for the rest of the translation unit.  Does prefast support "suppress" here instead to just disable the warning for the next line?  Alternatively, can you do prefast(push) and (pop) like you can for warnings?  https://learn.microsoft.com/en-us/cpp/preprocessor/warning?view=msvc-170

evelynwu-msft: Yes, that's possible: [https://learn.microsoft.com/en-us/previous-versions/windows/embedded/gg155764(v=winembedded.70)#parameters](https://learn.microsoft.com/en-us/previous-versions/windows/embedded/gg155764(v=winembedded.70)#parameters)

evelynwu-msft: Although for the purpose this usage I think changing `disable` to `suppress` is sufficient since we only need it for the following statement.

pratikone: good idea. i didn't know one could do that.

bpulliam (approved):

PR #4135, opened 2024-01-30
Revert "DynamicDependencies: Use Win11 APIs if available"

kythant:
…(#4108)" (#4118)"

This reverts commit daf549809c22bfc82543acad9381c955d125096d.

A microsoft employee must use /azp run to validate using the pipelines below.

WARNING:
Comments made by azure-pipelines bot maybe inaccurate.
Please see pipeline link to verify that the build is being ran.

For status checks on the main branch, please use TransportPackage-Foundation-PR
(https://microsoft.visualstudio.com/ProjectReunion/_build?definitionId=81063&_a=summary)
and run the build against your PR branch with the default parameters.

PR #4171, opened 2024-02-07
[1.5] CP of Revert "Revert "DynamicDependencies

bpulliam:
…4 #4083 (#4108)" (#4118)" (#4135)" (#4140)

* Revert "Revert "DynamicDependencies: Use Win11 APIs if available #3704 #4083 (#4108)" (#4118)" (#4135)"

This reverts commit 117ed20d739a5e6b53a93f00bf4c29ece06a335b.

* Disable DynamicDependency-delegate-to-OS by changing IsSupported() (not reverting the world)

* PackageRuntimeManager tests rely on OS DynDep functionality but Setup was a parallel version check when it should have been IsSupported. Fixed so they're all using the same answer

* Doh! Had the check reversed -- if (foo) should have been if (!foo). Fixed

* Tweak TryCreatePackageDependency when calling the OS with an invaliud package family name we get back STATEREPOSITORY_E_DEPENDENCY_NOT_RESOLVED but should return to the caller E_INVALIDARG

* Remove false skip-tests check. The surgical IsSupported disables all delegation to OS APIs. When that changes we WANT the tests to run and canary any problem

* Bad test passing too long a VersionTag. Tweaked Bootstrapper to spew WIL message and deal instead of FAIL_FAST

* Fixed PackageDependency.m_packageFullName was never set and .PackageId string fields (e.g. .name) weren't properly set, though they weren't used so not noticed or harm (before the resolution fix - next sentence...). Fixed TryCreatePackageDependency() wasn't resolving the dependency unless the MddCreatePackageDependencyOptions_DoNotVerifyDependencyResolution option was specified. Fixed up some test issues as well asupdates in sync with the fixes (above)

* Fixed tests to correctly match (fixed) implementation

* Improved some test logging. Fided up a few tests out of date with the latest behavior

WARNING:


bpulliam:
There are as many removals here as there are additions. My primary desire is to not ship things that are incomplete or not being used.

DrusTheAxe:
> There are as many removals here as there are additions. My primary desire is to not ship things that are incomplete or not being used.

DynDep delegating to the OS made a bunch of changes. Problems cropped up so a chainsaw approach was taken to unblock the repo - revert everything. The more recent work identified root causes which are under development but the 3way merge hell makes any progress there exceedingly difficult so the chainsaw was replaced with a scalpel - change IsSupported() to return false and avoid those new problematic codepaths for now. This centralizes the 'disable' to a single line rather than scattered across dozens of places (including added hacks which are probably obsolete with the chainsaw but given the increasingly twisted path the commits have taken it's hard to say).

Given the current code it's easy to say with confidence with it does and how it behaves in product.

Removing MSIXInstallFromPath is just a side bonus.

DrusTheAxe (approved):

PR #4179, opened 2024-02-09
Fixed Microsoft.Windows.Management.Deployment.Projection.csproj

DrusTheAxe:
Microsoft.Windows.Management.Deployment.Projection.csproj was using 2nd `<CSRWinRTIncludes>` to resolve the DynDep type dependency when that's a singular 'namespace of the project's output' so the projection had DynDep types and not PkgMgmt. Changed it to a `<ProjectReference Include=...\DynDep.csproj>` and happiness.

Verified projection assembly content via ILSpy.

KUDOS to Steve Otteson for the magic incantation!


DrusTheAxe:
FYI https://task.ms/48917378

PR #4231, opened 2024-02-26
PackageDeploymentManager.IsReadyOrNewerAvailable() API spec

DrusTheAxe:
Add Is*ReadyOrNewerAvailable() and EnsureReadyOptions.RegisterNewerIfAvailable=boolean

https://task.ms/48914090

### DIFF
--- a/specs/packagemanager/PackageManagement.md
+++ b/specs/packagemanager/PackageManagement.md
@@ -12,6 +12,7 @@ but with additional functionality, improved developer experience and performance
   - [3.3. Is\*ReadyOrNewerAvailable()](#33-isreadyorneweravailable)
   - [3.4. Ensure\*Ready()](#34-ensureready)
     - [3.4.1. Why Is\*Ready() Given Ensure\*Ready()?](#341-why-isready-given-ensureready)
+    - [3.4.2. EnsureReadyOptions.RegisterNewerIfAvailable](#342-ensurereadyoptionsregisternewerifavailable)
   - [3.5. Repair](#35-repair)
   - [3.6. Reset](#36-reset)
   - [3.7. IsPackageRegistrationPending](#37-ispackageregistrationpending)
--- a/specs/packagemanager/PackageManagement.md
+++ b/specs/packagemanager/PackageManagement.md
@@ -137,8 +137,8 @@ Legend:
 
 ## 3.2. Is\*Ready()
 
-Is\*Ready() methods determine if the target is installed and ready for use. Reasons why a package is
-not ready can include:
+Is\*Ready() methods determine if the target is installed (registered) and ready for use. Reasons why
+a package is not ready can include:
 
 * The package is not present on the machine
 * The package is present on the machine but not registered for the user
@@ -150,19 +150,26 @@ can be used.
 
 ## 3.3. Is\*ReadyOrNewerAvailable()
 
-Is\*Ready() methods determine if the target is installed and ready for use or a newer version is
-locally available. Reasons why a package is not ready can include:
+Is\*ReadyOrNewerAvailable() methods determine if the target is installed (registered) and ready for
+use AND if a newer version is locally available. Reasons why a package is ready but a newer version
+is available can include:
 
-* The package is not present on the machine
-* The package is present on the machine but not registered for the user
-* The package is registered for the user but is not in a healthy status e.g. its
-  Package.Status=Tampered
 * The package is registered for the user but a newer version is available locally on the machine
   e.g. a user has package v1 registered and Windows Update downloaded and staged v2 in the
   background while app(s) are running using v1.
 
 Is\*ReadyOrNewerAvailable() methods are a quick test to determine if more (costly) work is needed
 before the target can be used.
+
+**NOTE:** Is\*ReadyOrNewerAvailable() can only determine a newer package is available than a package
+installed (registered) and ready for use. If Is\*Ready() returns `false` then
+Is\*ReadyOrNewerAvailable() will never return `NewerAvailable`.
+
+**NOTE:** Is\*ReadyOrNewerAvailable() does not require admin privilege.
+[PackageManager.FindPackages](https://learn.microsoft.com/en-us/uwp/api/windows.management.deployment.packagemanager.findpackages?view=winrt-22621)()
+and
+[PackageManager.FindPackagesForUser](https://learn.microsoft.com/en-us/uwp/api/windows.management.deployment.packagemanager.findpackagesforuser?view=winrt-22621)(user!=currentuser...)
+can enumerate packages regardless if staged or registered for the user but requires admin privilege.
 
 ## 3.4. Ensure\*Ready()
 
@@ -651,7 +658,7 @@ if (options.IsLimitToExistingPackagesSupported)
 ```c# (but really MIDL3)
 namespace Microsoft.Windows.Management.Deployment
 {
-    [contractversion(1)]
+    [contractversion(2)]
     apicontract PackageDeploymentContract{};
 
     /// Represents a package storage volume.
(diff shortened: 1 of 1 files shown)


specs/packagemanager/PackageManagement.md
```diff
@@ -145,7 +148,23 @@ not ready can include:
 Is\*Ready() methods are a quick test to determine if more (costly) work is needed before the target
 can be used.
 
-## 3.3. Ensure\*Ready()
+## 3.3. Is\*ReadyOrNewerAvailable()
```
jonwis, line 151: This renumbering is causing a large ripple in the change. Future, just add a new number to the end of the list.

specs/packagemanager/PackageManagement.md
```diff
@@ -667,6 +715,15 @@ namespace Microsoft.Windows.Management.Deployment
         UsePreference,
     };
 
+    /// Defines the stub behavior for an app package that is being added or staged.
+    [contract(PackageDeploymentContract, 1)]
```
jonwis, line 719: Have we already shipped the v1 of this contract?

specs/packagemanager/PackageManagement.md
```diff
@@ -854,6 +911,8 @@ namespace Microsoft.Windows.Management.Deployment
         EnsureReadyOptions();
 
         AddPackageOptions AddPackageOptions { get; };
+
+        Boolean RegisterNewerIfAvailable { get; };
```
jonwis, line 915: Get-only, or should be get/set?

specs/packagemanager/PackageManagement.md
```diff
@@ -145,7 +148,23 @@ not ready can include:
 Is\*Ready() methods are a quick test to determine if more (costly) work is needed before the target
 can be used.
 
-## 3.3. Ensure\*Ready()
+## 3.3. Is\*ReadyOrNewerAvailable()
+
+Is\*Ready() methods determine if the target is installed and ready for use or a newer version is
```
oldnewthing, line 153:
```suggestion
Is\*ReadyOrNewerAvailable() methods determine if the target is installed and ready for use or a newer version is
```
?

specs/packagemanager/PackageManagement.md
```diff
@@ -145,7 +148,23 @@ not ready can include:
...
+locally available. Reasons why a package is not ready can include:
+
+* The package is not present on the machine
+* The package is present on the machine but not registered for the user
+* The package is registered for the user but is not in a healthy status e.g. its
+  Package.Status=Tampered
+* The package is registered for the user but a newer version is available locally on the machine
+  e.g. a user has package v1 registered and Windows Update downloaded and staged v2 in the
+  background while app(s) are running using v1.
```
oldnewthing, line 162: But in this last case, IsReady is false, but IsReadyOrNewerAvailable is true?

specs/packagemanager/PackageManagement.md
```diff
@@ -145,7 +148,23 @@ not ready can include:
...
+
```
oldnewthing, line 163: If the package is (say) not healthy, but a newer version is downloaded and staged, then IsReadyOrNewerAvailable returns true?

This section is confusing because the API under discussion is Is\*ReadyOrNewerAvailable, but all the text talks about is the Ready part and not the NewerAvailable part.

specs/packagemanager/PackageManagement.md
oldnewthing: Yeah, copy/pasta. This should be read-write. The AddPackageOptions is read-only because it is a runtime class. The property is read-only, but you can mutate the returned AddPackageObjects object.

specs/packagemanager/PackageManagement.md
```diff
@@ -873,6 +932,17 @@ namespace Microsoft.Windows.Management.Deployment
...
+        // IsReadyOrNewerAvailable
+
+        // Return true if the package(s) are present and available for use
+
+        PackageReadyOrNewerAvailableStatus IsPackageReadyOrNewerAvailable(String package);
+
+        PackageReadyOrNewerAvailableStatus IsPackageReadyOrNewerAvailableByUri(Windows.Foundation.Uri packageUri);
+
+        PackageReadyOrNewerAvailableStatus IsPackageSetReadyOrNewerAvailable(PackageSet packageSet);
+
```
chrilaMSFT, line 945: why not just have a IsNewerAvailable* methods we already have IsPackageReady*

specs/packagemanager/PackageManagement.md
DrusTheAxe: If package not health but newer version staged then IsReadyOrNewerAvailable=true.

Per previous comment the phrasing is wrong and being updated.

specs/packagemanager/PackageManagement.md
DrusTheAxe: Yes - 1.5.0 just shipped yesterday (was a surprise).

RECOMMEND: Change contract to v2 for the new bits.

specs/packagemanager/PackageManagement.md
DrusTheAxe: Yes, thanks.

RECOMMEND: Support get+set

specs/packagemanager/PackageManagement.md
DrusTheAxe: That would require admin privilege but the ask's for MediumIL processes.

The requirement to only detect newer for registered packages is to prevent an Elevation-of-Privilege. Enumerating packages not registered for the user requires admin privilege, regardless if the user has anything related already registered. PackageManager.FindPackages*() only return packages not registered for the user if called with admin privilege e.g. see 'Remarks' in PackageManager.[FindPackages](https://learn.microsoft.com/uwp/api/windows.management.deployment.packagemanager.findpackagesforuser)() and PackageManager.[FindPackagesForUser](https://learn.microsoft.com/uwp/api/windows.management.deployment.packagemanager.findpackagesforuser)()).

If not for this admin privilege caveat this would be "IsNewerAvailable(criteria)" and non-elevated processes could probe for anything regardless what the user has registered.

DrusTheAxe: Another reason to make IsReadyOrNewerAvailable a superset of IsReady rather than add a parallel IsNewerAvailable is performance.

We can't reduce the work done by IsNewerAvailable w/o the EoP issue, but we could return boolean rather than a superset enum. But doing so requires apps also call IsReady to determine what to do e.g.

```
if (!IsReady(x))
    AddPackage(x)
else if IsNewerAvailable(x)
    RegisterPackageFamily(x)
```

NOTE: Ensure*ReadyAsync() does this internally. Dev convenience but doesn't change the perf costs.

We've already been asked to make IsReady faster. As IsReady and IsNewerAvailable implementations heavily overlap, providing the functionality in 1 API provides efficiencies we can't get with separate functions.

DrusTheAxe: DISCUSSION: Explanation why IsNewerAvailable can't skip its is-registered logic to compute its answer plus the perf impacts of separate functions vs 1 superset function.

DrusTheAxe (commented):
Feedback review, commented and addressed.

specs/packagemanager/PackageManagement.md
```diff
@@ -873,6 +940,20 @@ namespace Microsoft.Windows.Management.Deployment
...
+        [contractversion(2)]
+        PackageReadyOrNewerAvailableStatus IsPackageReadyOrNewerAvailable(String package);
+
+        [contractversion(2)]
+        PackageReadyOrNewerAvailableStatus IsPackageReadyOrNewerAvailableByUri(Windows.Foundation.Uri packageUri);
+
+        [contractversion(2)]
+        PackageReadyOrNewerAvailableStatus IsPackageSetReadyOrNewerAvailable(PackageSet packageSet);
+
```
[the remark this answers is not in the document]
DrusTheAxe: It was...and then 1.5 shipped (sooner than expected) so contract version is now bumped

cwruss (approved):

PR #4267, opened 2024-03-14
 Implement Dynamic NuGet Retrieval in WindowsAppSDK C++ Wizard

ShashankNay:
Introducing a new `NuGetPackageInstaller` wizard to the WindowsAppSDK, enabling dynamic retrieval and installation of the latest NuGet packages during C++ project setup. The implementation bypasses static versioning, utilizing `IVsPackageInstaller` to ensure the use of current packages without manual updates. The removal of direct NuGet package inclusion from the VSIX has successfully reduced the file size from 105 MB to 5 MB, enhancing the efficiency of the installation experience.

Changes contain:
- Asynchronous NuGet package installations within the project creation lifecycle.
- A maintained list of essential packages in `NuGetPackageList.cs`, centralizing dependency management.
- Enhanced error handling mechanisms reporting directly to Visual Studio's Activity Log for developer clarity.

### DIFF
--- a/dev/VSIX/Extension/Cpp/Dev17/NuGetPackageList.cs
+++ b/dev/VSIX/Extension/Cpp/Dev17/NuGetPackageList.cs
@@ -1,0 +1,16 @@
+using System;
+using System.Collections.Generic;
+
+namespace WindowsAppSDK.Cpp.Extension
+{
+    public static class NuGetPackageList
+    {
+        public static readonly List<string> Packages = new List<string>
+        {
+            "Microsoft.WindowsAppSDK",
+            "Microsoft.Windows.CppWinRT",
+            "Microsoft.Windows.SDK.BuildTools",
+            "Microsoft.Windows.ImplementationLibrary",
+        };
+    }
+}
--- a/dev/VSIX/Extension/Cpp/Dev17/WizardImplementationCpp.cs
+++ b/dev/VSIX/Extension/Cpp/Dev17/WizardImplementationCpp.cs
@@ -80,6 +80,7 @@ namespace WindowsAppSDK.TemplateUtilities.Cpp
         }
         private void LogError(string message)
         {
+            ThreadHelper.ThrowIfNotOnUIThread();
             ThreadHelper.JoinableTaskFactory.Run(async delegate
             {
                 await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
--- a/dev/VSIX/Extension/Cpp/Dev17/WizardImplementationCpp.cs
+++ b/dev/VSIX/Extension/Cpp/Dev17/WizardImplementationCpp.cs
@@ -1,0 +1,98 @@
+// Copyright (c) Microsoft Corporation and Contributors.
+// Licensed under the MIT License
+
+using EnvDTE;
+using Microsoft.VisualStudio.ComponentModelHost;
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Shell.Interop;
+using Microsoft.VisualStudio.TemplateWizard;
+using NuGet.VisualStudio;
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using WindowsAppSDK.Cpp.Extension;
+
+namespace WindowsAppSDK.TemplateUtilities.Cpp
+{
+    public class NuGetPackageInstaller : IWizard
+    {
+        private Project _project;
+        private IComponentModel _componentModel;
+        public void RunStarted(object automationObject, Dictionary<string, string> replacementsDictionary, WizardRunKind runKind, object[] customParams)
+        {
+            ThreadHelper.ThrowIfNotOnUIThread();
+            _componentModel = (IComponentModel)ServiceProvider.GlobalProvider.GetService(typeof(SComponentModel));
+        }
+        public void ProjectFinishedGenerating(Project project)
+        {
+            _project = project;
+
+            // Ensure we're on the main thread, as required by the ProjectFinishedGenerating method and DTE operations
+            ThreadHelper.JoinableTaskFactory.Run(async () =>
+            {
+                await InstallNuGetPackagesAsync(_project);
+            });
+        }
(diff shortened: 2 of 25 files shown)


dev/VSIX/Extension/Cpp/Dev17/WizardImplementationCpp.cs
```diff
@@ -0,0 +1,99 @@
...
+            // Ensure we're on the main thread, as required by the ProjectFinishedGenerating method and DTE operations
+            ThreadHelper.JoinableTaskFactory.Run(async () =>
+            {
+                await InstallNuGetPackagesAsync(_project);
+            });
+        }
+        private async Task InstallNuGetPackagesAsync(Project project)
+        {
+            await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+            foreach (var packageId in NuGetPackageList.Packages)
```
Scottj1s, line 39: with multiple support now, can we make Microsoft.Windows.SDK.BuildTools dynamic in the C# templates?

also, a bit of cleanup: 
```
      <CustomParameter Name="$WindowsAppSDKNupkgVersion$" Value="FIXME-Verify-Directory.Build.Targets-XmlPoke-Queries" />
      <CustomParameter Name="$WindowsSDKBuildToolsNupkgVersion$" Value="FIXME-Verify-Directory.Build.Targets-XmlPoke-Queries" />
```
etc.

dev/VSIX/Extension/Cpp/Dev17/NuGetPackageList.cs
```diff
@@ -0,0 +1,16 @@
+using System;
+using System.Collections.Generic;
+
+namespace WindowsAppSDK.Cpp.Extension
+{
+    public static class NuGetPackageList
+    {
+        public static readonly List<string> Packages = new List<string>
```
Scottj1s, line 8: I'd prefer to see this in the templates:
```
  <WizardData>
    <packages>
      <package id="Microsoft.WindowsAppSDK" />
      <package id="Microsoft.Windows.CppWinRT" />
      <package id="Microsoft.Windows.SDK.BuildTools" />
      <package id="Microsoft.Windows.ImplementationLibrary" />
    </packages>      
  </WizardData>
```

dev/VSIX/Extension/Cpp/Dev17/WizardImplementationCpp.cs
Scottj1s: I guess this is a separate implementation from C#?  But would still like to see multiple support added there.

dev/VSIX/Extension/Cpp/Dev17/WizardImplementationCpp.cs
Scottj1s, line 1: there's a fair amount of commonality between this and WizardImplementation.cs.  I think it would be better to combine them and fork the logic on language (project extension)

dev/VSIX/Extension/Cpp/Dev17/NuGetPackageList.cs
ShashankNay: I tried that initially and get this error message during project creation:
image
The Visual Studio template processing still expects each <package> element to have both an "id" and a "version" attribute with valid values. At that point I thought about using another name instead of packages to accomplish this, but it also required additional xml parsing logic within the Wizard so I abandoned that idea.

dev/VSIX/Extension/Cpp/Dev17/WizardImplementationCpp.cs
ShashankNay: Yes, it's a separate implementation that allows us to install multiple packages compared to the existing C# method that was better suited for only one package.

dev/VSIX/Extension/Cpp/Dev17/WizardImplementationCpp.cs
ShashankNay: Since each VSIX is packaged separately and would require its own Wizard.dll, what would be the best way to combine them? You're right - there is significant overlap between both files and the only real differences are the underlying mechanism (packages.config vs packagereference) and the NuGet packages themselves.

evelynwu-msft: What about factoring out the common code into a class library project that each VSIX project references? Or alternatively factor it out into a shared project if you don't want each extension to gain an additional DLL.

dev/VSIX/Extension/Cpp/Dev17/WizardImplementationCpp.cs
Scottj1s: C# would also benefit from multi-package support (BuildTools)

dev/VSIX/Extension/Cpp/Dev17/NuGetPackageList.cs
Scottj1s: Here's the brittle implementation you're tripping on:
https://github.com/NuGet/NuGet.Client/blob/6917e6c883d45f45672e20d4b1c45758dad2fa84/src/NuGet.Clients/NuGet.VisualStudio.Implementation/VsTemplateWizard.cs#L97

I'd recommend trying again with unique element names.

dev/VSIX/Extension/Cpp/Dev17/WizardImplementationCpp.cs
```diff
@@ -0,0 +1,99 @@
...
+        {
+        }
+        public void RunFinished()
+        {
+
+        }
+        private async Task InstallNuGetPackageAsync(string packageId)
+        {
+            await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+            IVsPackageInstaller installer = _componentModel.GetService<IVsPackageInstaller>();
```
ghost1372, line 65: there is a `IVsPackageInstaller2` class which contain a new method called `InstallLatestPackage` so this line:
```
installer.InstallPackage(null, _project, packageId, "", false);
``` 
can be changed to this:
```
installer.InstallLatestPackage(null, _project, packageId, false, false);
```

PR #4271, opened 2024-03-18
Better logging support, replaces LOG_HR_MSG with DebugLog

pratikone:
This PR replaces and refactors ways to do logging in WinAppSDK codebase. 
Currently, LOG_HR_MSG from wil library is used for logging purposes too even though its usage is for error-reporting.

In current scenario, we want to print logging information to Debug Console (using OutputDebugString). This PR introduces a new logging function to do the same : `DebugLog` in `Logging.h`.
With this change, any place which is using LOG_HR_MSG for just diagnostic logging, I will replace it with this function call and also remove any macros associated with it.

Another change is refactoring of` Mddbootstrap.cpp`'s `FindDDLMViaAppExtension()` to be cleaner.

At 2-3 places, I have left custom winappruntime provided error code intact because it is throwing them. A user code could be setup to catch them and hence, to save on backward compatibility, I am keeping them intact. (thanks @Scottj1s for the suggestion)


dev/Common/Logging.h
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) Microsoft Corporation and Contributors.
+// Licensed under the MIT License.
+
+#if !defined(__LOGGING_H)
```
Scottj1s, line 4: #pragma once is more efficient than include guards, which are now obsolete

dev/Common/Logging.h
```diff
@@ -0,0 +1,31 @@
...
+#define __LOGGING_H
+
+#include <Windows.h>
+#include <source_location>
+#include <sstream>
+
+namespace Common::Logging
+{
+
```
Scottj1s, line 13: nit: blank line

dev/Common/Logging.h
```diff
@@ -0,0 +1,31 @@
...
+    // use std::format in msg parameter to have a printf like varargs syntax for the debug message
+    inline void DebugLog(const wchar_t* msg,  const std::source_location location = std::source_location::current())
+    {
+        
+        std::wstringstream filename ;
+        filename << location.file_name(); // converting char to wstring (and wchar_t implicitly)
```
Scottj1s, line 19: probably don't need the stream for this

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -322,23 +322,23 @@ void VerifyInitializationIsCompatible(
     GetFrameworkPackageFamilyName(majorMinorVersion, versionTag);
 
     // Is the initialization request compatible with the current initialization state?
-    THROW_HR_IF_MSG(MDD_E_BOOTSTRAP_INITIALIZE_INCOMPATIBLE,
+    THROW_HR_IF_MSG(E_FAIL,
```
Scottj1s, line 325: Does this HRESULT find its way to user code?  If so, we can't change it.  (See Liskov Substitution Principle)

dev/PackageManager/API/PackageDeploymentResolver.cpp
```diff
@@ -226,21 +220,13 @@ winrt::hstring Microsoft::Windows::ApplicationModel::PackageDeploymentResolver::
...
-                                     "PackageDeploymentResolver: %ls not applicable. Architecture (%ls) doesn't match system supported architectures (0x%X %ls)",
-                                     candidateFullName.c_str(), ::AppModel::Identity::GetArchitectureAsString(candidateArchitecture),
-                                     static_cast<std::uint32_t>(supportedArchitectures), GetSystemSupportedArchitecturesAsString(nativeMachine));
                     continue;
                 }
             }
             else
             {
                 if (!IsArchitectureInArchitectures(candidateArchitecture, processorArchitectureFilter))
                 {
```
bpulliam, line 229: a comment here to explain would be helpful.

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -982,9 +980,6 @@ void FindDDLMViaEnumeration(
         version.Revision = packageVersion.Revision;
         if (version.Version < minVersion.Version)
         {
-            (void)LOG_HR_MSG(MDD_E_BOOTSTRAP_INITIALIZE_DDLM_SCAN_NO_MATCH,
-                             "Bootstrap.Intitialize: %ls not applicable. Version doesn't match MinVersion criteria (%ls)",
-                             packageFullName.c_str(), criteria.get());
             continue;
```
Scottj1s, line 983: why did we remove logging altogether here (and below)?

dev/PackageManager/API/PackageDeploymentResolver.cpp
```diff
@@ -226,21 +220,13 @@ winrt::hstring Microsoft::Windows::ApplicationModel::PackageDeploymentResolver::
                 const auto supportedArchitectures{ GetSystemSupportedArchitectures(nativeMachine) };
                 if (!IsArchitectureInArchitectures(candidateArchitecture, supportedArchitectures))
                 {
```
bpulliam, line 222: a comment here to explain would be helpful.

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -993,41 +988,28 @@ void FindDDLMViaEnumeration(
...
             bestFitPackageFullName = packageId.FullName();
             continue;
         }
     }
-    THROW_HR_IF_MSG(STATEREPOSITORY_E_DEPENDENCY_NOT_RESOLVED, !foundAny, "Enumeration: %ls", criteria.get());
-    (void)LOG_HR_MSG(MDD_E_BOOTSTRAP_INITIALIZE_DDLM_FOUND,
-                     "Bootstrap.Intitialize: %ls best matches the criteria (%ls) of %d packages scanned",
-                     bestFitPackageFullName.c_str(), criteria.get(), packagesScanned);
+    // if bestFitVersion is 0, the value we initialized it with,  then we didn't find a suitable candidate
+    THROW_HR_IF_MSG(E_FAIL, bestFitVersion.Version == 0, "No suitable version matching criteria is found. Criteria: %ls", criteria.get());
```
Scottj1s, line 1007: here too - if the HRESULT is user-visible, we have to maintain it

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -993,41 +988,28 @@ void FindDDLMViaEnumeration(
...
+    {
+        wchar_t printmsg[128];
+        Common::Logging::DebugLog(std::format(L"Bootstrap.Intitialize: {} best matches the criteria ({}) of {} packages scanned",
```
Scottj1s, line 1010: is this a generalized roll-up log message to replace all the removed ones above?

test/DynamicDependency/Test_Win32/TestMddBootstrap.cpp
```diff
@@ -190,7 +190,7 @@ namespace Test::DynamicDependency
             // Major.Minor = 0.0 == No such framework package
             const UINT32 doesNotExist{};
             const PACKAGE_VERSION minVersionMatchAny{};
-            VERIFY_ARE_EQUAL(STATEREPOSITORY_E_DEPENDENCY_NOT_RESOLVED, MddBootstrapInitialize(doesNotExist, nullptr, minVersionMatchAny));
+            VERIFY_ARE_EQUAL(E_FAIL, MddBootstrapInitialize(doesNotExist, nullptr, minVersionMatchAny));
```
Scottj1s, line 193: definitely cannot change HRESULTs returned by MddBootstrapInitialize

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
pratikone: it is inside a void functrion so I assume, no.
 still user can have a try catch block with this error. i will replace with original error message from Win32

Scottj1s: right - return type of the function is irrelevant to exceptions thrown

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
pratikone: I can add the comments back. Those messages were diagnostic logging and would add 5-10 lines on each build if you have multiple Winappsdk packages -  main, ddlm, etc. It is just a search result and I didn't want to increase noise in user visible area. Hence, decided to remove unnecessary logging. We are not losing any helpful debugging info from this. This code tells which packages it considered to be invalid. That list can be long.

Scottj1s: Thanks for explanation - I'm fine with the change, just wanted to clarify.

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
pratikone: I decided against it. the log messages were like 
Testing foo package but it didn't match . arch : x86, looking : x64.

those messages add up a lot if you have more than 1 winappsdk package and provide little diagnostic value. There are 5-10 lines of these in VS console output which would confuse developers. Hence, removed them

dev/Common/Logging.h
pratikone: all other solutions I checked are error-prone and complicated. like `mbstowcs_s`. Any other way you can suggest ?

Scottj1s: can't you just pass const char* file_name() to std::format directly?https://en.cppreference.com/w/cpp/utility/source_location/file_name

dev/Common/Logging.h
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Common::Logging
+{
+    // use std::format in msg parameter to have a printf like varargs syntax for the debug message
+    inline void DebugLog(const wchar_t* msg,  const std::source_location location = std::source_location::current())
+    {
+        std::wstringstream filename ;
+        filename << location.file_name(); // converting char to wstring (and wchar_t implicitly)
+        // format is "foo.cpp (66) : debug message"
+        const std::wstring printmsg = std::format(L"{} ({}) : {}\n", filename.str(), location.line(), msg);
```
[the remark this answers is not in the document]
pratikone: it doesn't work because location_filen_name() returns a wstring. std::format doesn't support mixing string and wstring together and results in a compilation error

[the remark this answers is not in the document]
pratikone: thanks Scott

PR #4282, opened 2024-03-22
Fix incremental build issue where Microsoft.WindowsAppRuntime.Bootstrap.dll is always written to output

jevansaks:
PostBuildEvent is unconditionally run and xcopy always does the copy, making downstream projects think that they need to rebuild too because their references rebuild.

Switch to using a Target that leverages the built-in Copy task that already does the timestamp checks for us.

### DIFF
--- a/build/NuSpecs/WindowsAppSDK-Nuget-Native.C.props
+++ b/build/NuSpecs/WindowsAppSDK-Nuget-Native.C.props
@@ -28,9 +28,12 @@
         %(AdditionalDependencies);
       </AdditionalDependencies>
     </Link>
-    <PostBuildEvent>
-      <Command>xcopy.exe /y "$(MSBuildThisFileDirectory)..\..\runtimes\win10-$(_WindowsAppSDKFoundationPlatform)\native\Microsoft.WindowsAppRuntime.Bootstrap.dll" "$(OutDir)"</Command>
-    </PostBuildEvent>
   </ItemDefinitionGroup>
+
+  <Target Name="CopyMicrosoftWindowsAppRuntimeBootstrapdllToOutDir" Condition="'$(AppxPackage)' != 'true'" AfterTargets="Build">
+    <Copy
+      SourceFiles="$(MSBuildThisFileDirectory)..\..\runtimes\win10-$(_WindowsAppSDKFoundationPlatform)\native\Microsoft.WindowsAppRuntime.Bootstrap.dll"
+      DestinationFolder="$(OutDir)"/>
+  </Target>
 
 </Project>


build/NuSpecs/WindowsAppSDK-Nuget-Native.C.props
```diff
@@ -28,9 +28,12 @@
         %(AdditionalDependencies);
       </AdditionalDependencies>
     </Link>
-    <PostBuildEvent>
-      <Command>xcopy.exe /y "$(MSBuildThisFileDirectory)..\..\runtimes\win10-$(_WindowsAppSDKFoundationPlatform)\native\Microsoft.WindowsAppRuntime.Bootstrap.dll" "$(OutDir)"</Command>
-    </PostBuildEvent>
   </ItemDefinitionGroup>
 
+  <Target Name="CopyMicrosoftWindowsAppRuntimeBootstrapdllToOutDir" Condition="'$(AppxPackage)' != 'true'" AfterTargets="Build">
```
jonwis, line 33: Why is this conditioned on "not an appx package"? I guess because generally appx uses the FWP model of binding and not bootstrap?

jevansaks: I don't know, I just maintained the condition from where it was previously. Probably a question for whoever added this -- @DrusTheAxe added in https://github.com/microsoft/WindowsAppSDK/pull/827.

I also see that Howard tried to make a change to Copy in #2734 and put it back on #2758. Hopefully someone more knowledgeable about these can help verify this change doesn't re-break any of those scenarios. I suspect the change in #2734 didn't work because it wasn't invoking the task in a Target.

jonwis: OK. I'm generally supportive of the change. If you wanted a slam-dunk no-questions-asked change, just add the `/d` parameter to xcopy, which only copies if the source is newer than the target.

jonwis (approved):

build/NuSpecs/WindowsAppSDK-Nuget-Native.C.props
DrusTheAxe: >Why is this conditioned on "not an appx package"?
Because Bootstrapper is only needed when WinAppSDK MSIX packages via WinAppSDK DynamicDependencies, AND WinAppSDK DynDep only supports unpackaged apps.

But isn't this necessary-but-insufficient -- the condition should be "I an not a packaged app AND i am not using WinAppSDKSelfContained"?

```
Condition="('$(AppxPackage)' != 'true') and ('$(WindowsAppSDKSelfContained)' != 'true')"
```

?

DrusTheAxe, line 33: >'$(AppxPackage)' != 'true'

Is that the right test or should it be `'$(WindowsPackageType)'=='None'` ?

@Scottj1s @bpulliam @evelynwu-msft what's the latest practices for "My VS project is not making an MSIX package"?

jevansaks: This condition came from the ItemGroup above.

jonwis:
@kythant or @bpulliam or @evelynwu-msft - what's going on with the build pipelines here?

kythant:
> @kythant or @bpulliam or @evelynwu-msft - what's going on with the build pipelines here?

The triggers won't work because the PR is coming from a forked copy of the repository. The branch cannot be seen from this pipeline https://microsoft.visualstudio.com/DefaultCollection/ProjectReunion/_build?definitionId=81063

jonwis:
> > @kythant or @bpulliam or @evelynwu-msft - what's going on with the build pipelines here?
> 
> The triggers won't work because the PR is coming from a forked copy of the repository. The branch cannot be seen from this pipeline https://microsoft.visualstudio.com/DefaultCollection/ProjectReunion/_build?definitionId=81063

So what should @jevansaks do? How do we get this built and validated?

bpulliam:
I made the change earlier this morning, and I now see that it has been changed again (I don't see the history).

It should work for you now.

jevansaks:
Builds from forks are enabled but without secrets they can't do anything. Abandoning for now until the team spends time to make forks work. Re-opened as #4296

PR #4328, opened 2024-04-11
49601674: [Compliance] address pre-existing PREfast errors and enable break-on-error setting for PREfast

alexlamtest:
Addressed the pre-existing errors reported by PREfast, thus allowing us to enable the "break the build on errors" setting for PREfast. 
Testing was done in the corresponding internal pipeline, the pipeline build was successful with the "break on error" setting. 
PR validation has succeeded.

----

WARNING:

### DIFF
--- a/dev/Deployment/DeploymentManager.cpp
+++ b/dev/Deployment/DeploymentManager.cpp
@@ -414,8 +414,8 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
                 deploymentResult.ActivityId());
         }
 
-        return !deploymentOperationHResult ? deploymentOperationHResult :
-            (deploymentOperationExtendedHResult ? deploymentOperationExtendedHResult : deploymentOperationHResult);
+        return FAILED(deploymentOperationHResult) ? deploymentOperationHResult :
+            (FAILED(deploymentOperationExtendedHResult) ? deploymentOperationExtendedHResult : deploymentOperationHResult);
     }
     CATCH_RETURN()
 
--- a/dev/Deployment/DeploymentManager.cpp
+++ b/dev/Deployment/DeploymentManager.cpp
@@ -414,7 +414,9 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
                 deploymentResult.ActivityId());
         }
 
-        return FAILED(deploymentOperationHResult) ? deploymentOperationHResult :
+        // If deploymentOperationHResult indicates success, take that, ignore deploymentOperationExtendedHResult.
+        // Otherwise, return deploymentOperationExtendedHResult if there is an error in it, if not, return deploymentOperationHResult.
+        return SUCCEEDED(deploymentOperationHResult) ? deploymentOperationHResult :
             (FAILED(deploymentOperationExtendedHResult) ? deploymentOperationExtendedHResult : deploymentOperationHResult);
     }
     CATCH_RETURN()
--- a/dev/MRTCore/mrt/mrm/UnitTests/testMap.cpp
+++ b/dev/MRTCore/mrt/mrm/UnitTests/testMap.cpp
@@ -230,7 +230,7 @@ bool TestResourceMap::TryAddCandidatesFromTestVars(
         for (unsigned i = 0; i < candidates.GetSize(); i++)
         {
             Log::Comment(tmp.Format(L"[ candidate %d: %s ]", i, (PCWSTR)candidates[i]));
-            if (FAILED(candidate.InitFromList(candidates[i])))
+            if (!candidate.InitFromList(candidates[i]))
             {
                 Log::Warning(L" [ Couldn't parse candidate string ] ");
                 continue;
@@ -511,7 +511,7 @@ TestResourceMap::VerifyCandidates(_In_ TestMapPrivate* pMap, _In_ TestDecisionIn
         for (unsigned iSpec = 0; iSpec < specs.GetSize(); iSpec++)
         {
             Log::Comment(tmp.Format(L"[ Candidate %d: %s ]", iSpec, (PCWSTR)specs[iSpec]));
-            if (FAILED(spec.InitFromList(specs[iSpec])))
+            if (!spec.InitFromList(specs[iSpec]))
             {
                 Log::Warning(tmp.Format(L"[ Error parsing spec %d in %sExpectedCandidateInfo ]", iSpec, pVarPrefix));
                 continue;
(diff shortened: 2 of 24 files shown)


dev/MRTCore/mrt/mrm/UnitTests/testMap.cpp
```diff
@@ -715,7 +715,7 @@ TestResourceMap::VerifyDescendentScopes(_In_ TestMapPrivate* pMap, _In_ PCWSTR p
     if (pMap->pSubtree == nullptr)
     {
         Log::Warning(tmp.Format(L"[ VerifyDescendentScopess(\"%s\") requires subtree ]", pVarPrefix));
```
bpulliam, line 717: nit:mispelled function name

installer/dev/console.cpp
```diff
@@ -48,10 +48,10 @@ void WindowsAppRuntimeInstaller::Console::DisplayError(const HRESULT hr)
 
     HRESULT hResult = hr;
 
-    if (installActivityContext.GetDeploymentErrorHresult() &&
-        installActivityContext.GetInstallStage() == InstallStage::StagePackage ||
+    if ((installActivityContext.GetDeploymentErrorHresult() != S_OK) &&
```
bpulliam, line 51: should be FAILED()

installer/dev/console.cpp
```diff
@@ -70,8 +70,8 @@ void WindowsAppRuntimeInstaller::Console::DisplayError(const HRESULT hr)
     }
 
     // Don't log redundant Hr information
-    if (installActivityContext.GetDeploymentErrorExtendedHResult() != S_OK &&
-        installActivityContext.GetDeploymentErrorExtendedHResult() != hResult &&
+    if ((installActivityContext.GetDeploymentErrorExtendedHResult() != S_OK) &&
```
bpulliam, line 73: should be FAILED()

dev/Deployment/DeploymentManager.cpp
```diff
@@ -414,8 +414,8 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
                 deploymentResult.ActivityId());
         }
 
-        return !deploymentOperationHResult ? deploymentOperationHResult :
-            (deploymentOperationExtendedHResult ? deploymentOperationExtendedHResult : deploymentOperationHResult);
+        return FAILED(deploymentOperationHResult) ? deploymentOperationHResult :
```
bpulliam, line 417: just double-checking - this is a functional change (and I believe a bug fix).

dev/MRTCore/mrt/mrm/include/mrm/platform/CoreQualifierTypes.h
```diff
@@ -60,7 +60,7 @@ class DXFeatureLevelQualifierType : public EnumerationQualifierType
 
     virtual ~DXFeatureLevelQualifierType() {}
 
-    HRESULT Evaluate(_In_ const IQualifier* pQualifier, _In_ PCWSTR pValue, _Outptr_ double* score) const;
+    HRESULT Evaluate(_In_ const IQualifier* pQualifier, _In_ PCWSTR pValue, _Out_ double* score) const;
```
bpulliam, line 63: this fix was not clear. More detail please.

dev/MRTCore/mrt/mrm/UnitTests/testMap.cpp
alexlamtest: Updated.

installer/dev/console.cpp
alexlamtest: Updated, thanks!

installer/dev/console.cpp
alexlamtest: Updated, thanks!

dev/MRTCore/mrt/mrm/include/mrm/platform/CoreQualifierTypes.h
alexlamtest: Sure, for context, here's the original error:
```
Tool: PREfast: Rule: C6504 ((Unknown Rule Name)). 
Invalid annotation: 'Null' property may only be used on values of pointer, pointer-to-member, array, or reference type: Function 'Evaluate' _Param_(3). Actual type 'double'.
```
Fixed by matching with another 2 instances of the Evaluate() method in "class ContrastQualifierType" (line 16) and "class ScaleQualifierType" (line 44) that did _not_ trigger this PREfast issue.

I think the right use of _Outptr_ in this context should look like "_Outptr_ double** score". PREfast spotted the missing "*", but the code author likely meant to use _Out_ like in the other 2 instances in the same source, instead of really meaning to say "_Outptr_ double** score".

dev/Deployment/DeploymentManager.cpp
alexlamtest: For context, here's the original error:
```
##[error]18. PREfast Error C6230 - File: file:///C:/__w/1/s/dev/Deployment/DeploymentManager.cpp. Line: 418. Column 14. 
Tool: PREfast: Rule: C6230 ((Unknown Rule Name)). 
Implicit cast between semantically different integer types.
```
My understanding of the intention of the code author is:
- if deploymentOperationHResult contains an error then just return it and be done
- otherwise, if deploymentOperationExtendedHResult contains an error then just return it be done
- otherwise, return deploymentOperationHResult which should have contained "success"

I think PREfast does not like the use of HRESULT types deploymentOperationHResult and deploymentOperationExtendedHResult where booleans are expected in the logical expression, so I added FAILED() to do the conversion. Looks like the original code can handle S_OK (i.e., 0) being returned in deploymentOperationHResult and deploymentOperationExtendedHResult fine, but S_FALSE (i.e., 1) would not be handled correctly, but in reality this probably moot, because I don't recall the Package Manager ever using S_FALSE.

alexlamtest: Assuming the "!" near the front is intended probably minimizes the risk. Updated the fix and added comment for a possible interpretation. Thanks!

installer/dev/console.cpp
```diff
@@ -70,8 +70,8 @@ void WindowsAppRuntimeInstaller::Console::DisplayError(const HRESULT hr)
     }
 
     // Don't log redundant Hr information
-    if (installActivityContext.GetDeploymentErrorExtendedHResult() != S_OK &&
```
DrusTheAxe, line 73: This doesn't return S_FALSE or other SUCCEEDED cases we want to ignore here?

DrusTheAxe (approved):
Nicely done. Minor non-blocking comments

bpulliam (approved):

installer/dev/console.cpp
alexlamtest: We are now not logging S_FALSE, which indicates success, as this method is named DisplayError(), so I'm _guessing_ that S_FALSE is out of scope. If that's a wrong guess, my apologies, and will happily fix it next time.

Issue #4312, opened 2024-04-06
ResourceLoader can't find resources


YexuanXiao:
### Describe the bug

I added Strings/en-US/Settings.resw in my packaged app, but ResourceLoader("Settings") throws an exception.

Everything works fine when using a parameterless constructor, or when referencing the resource in XAML.

### Steps to reproduce the bug

Using [ResourceLoader(String)](https://learn.microsoft.com/zh-cn/windows/windows-app-sdk/api/winrt/microsoft.windows.applicationmodel.resources.resourceloader.-ctor?view=windows-app-sdk-1.5#microsoft-windows-applicationmodel-resources-resourceloader-ctor(system-string)).

### Expected behavior

Constructs a ResourceLoader that references the specified resource.

### NuGet package version

Windows App SDK 1.5.1: 1.5.240311000

### Packaging type

Packaged (MSIX)

### Windows version

Insider Build (26090)

### IDE

Visual Studio 2022-preview

### Additional context

Documents:

https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.windows.applicationmodel.resources.resourceloader.-ctor?view=windows-app-sdk-1.5#microsoft-windows-applicationmodel-resources-resourceloader-ctor(system-string)

https://learn.microsoft.com/en-us/windows/apps/windows-app-sdk/mrtcore/localize-strings?source=recommendations

YexuanXiao:
The old and obsolete API Windows.ApplicationModel.Resources.ResourceLoader has the correct behavior, but it's not a real solution to the problem.

roxk:
I'm not sure if this is intended but you need to say which pri you want to use now. So, instead of just `ResourceLoader("Settings")`, you should write `ResourceLoader(ResourceLoader.GetDefaultResourceFilePath(), "Settings")`.

If this is intended, doc should be updated.

ghost1372:
duplicate #2406

bpulliam:
duplicate

Issue #2406, opened 2022-04-16
ResourceLoader can not load the specified resource


ghost1372:
### Describe the bug

i have multiple resources (resw) files 
![Untitled](https://user-images.githubusercontent.com/9213496/163683630-29591007-2bf3-4ec5-aec3-a7ea38e684b7.png)

i can load resources from **Resources.resw** file with this code:
`myButton.Content = new ResourceLoader().GetString("String1");`

but i can not load resources from **ErrorMessage.resw** file, i used this code:
`myButton.Content = new ResourceLoader("ErrorMessage").GetString("msg");`

but i get exception:
Exception thrown at 0x76C8ED42 (KernelBase.dll) in App15.exe: WinRT originate error - 0x80070002 : 'The system cannot find the file specified.'.

In this case, using the following namespace works without any problems
`myButton.Content = Windows.ApplicationModel.Resources.ResourceLoader.GetForViewIndependentUse("ErrorMessage").GetString("msg");
`

### Steps to reproduce the bug

1.Create a new WinUI 3 Project
2.Create multiple Resource file
3. try to load resource from resw files.
4. you should use **Microsoft.Windows.ApplicationModel.Resources** 

### NuGet package version

1.0.0

### Packaging type

Packaged (MSIX)

### Windows version

Windows 11 version 21H2 (22000)

### IDE

Visual Studio 2022

huichen123:
can you do:
myButton.Content = new ResourceLoader("resources.pri", "ErrorMessage").GetString("msg");

riverar:
@axelandrejs Not sure this is ready to close. It seems like there's still a bug here?

axelandrejs:
There are two constructors. One takes the file path, one the file path and the scope. You can't instantiate for a scope without also specifying the file. As such I don't think there is a bug here.

riverar:
@axelandrejs Not sure what that means.

I think there are a few bugs here interacting with each other.

1. User has a file named `ErrorMessage.resw`. They are constructing a `ResourceLoader` with a filename as input using the constructor `ResourceLoader(string fileName)`.

The documentation indicates that an input of `ErrorMessage` will result in attempting to load `ErrorMessage.resw`, but that's not the case. It will attempt to load `ErrorMessage` (no extension) and fail.

[...]

PR #4329, opened 2024-04-11
Adjust ResourceLoader(String) to match documents

YexuanXiao:
In documents [ResourceLoader](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.windows.applicationmodel.resources.resourceloader.-ctor?view=windows-app-sdk-1.5#microsoft-windows-applicationmodel-resources-resourceloader-ctor(system-string)) and [WindowsAppSDK Localize strings in your UI and the app package manifest](https://learn.microsoft.com/en-us/windows/apps/windows-app-sdk/mrtcore/localize-strings?source=recommendations), str in Microsoft.Windows.ApplicationModel.Resources.ResourceLoader.ResourceLoader(String str) represents resourceMap, usually the file name of resw. In the current implementation, str represents the file name of pri the str of the old obsolete API [Windows.ApplicationModel.Resources.ResourceLoader(String str)](https://learn.microsoft.com/en-us/uwp/api/windows.applicationmodel.resources.resourceloader.-ctor?view=winrt-22621) represents the resourceMap.

So making the two more consistent seems to be a better solution, rather than using Microsoft.Windows.ApplicationModel.Resources.ResourceLoader.GetDefaultResourceFilePath().

documents:

```csharp
ResourceLoader("ErrorMessages") // implicit "resources.pri" or "modulename.pri"
```

implement:
```csharp
ResourceLoader(ResourceLoader.GetDefaultResourceFilePath(), "ErrorMessages")
```

this pr:
same as documents

### breaking change

before:
```csharp
ResourceLoader("filename.pri") // implicit "Resources" subtree
```
after:
```csharp
ResourceLoader("filename.pri", "Resources")
```

Original issue: https://github.com/microsoft/WindowsAppSDK/issues/4312 https://github.com/microsoft/WindowsAppSDK/issues/2406

### DIFF
--- a/dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/src/Microsoft.Windows.ApplicationModel.Resources.idl
+++ b/dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/src/Microsoft.Windows.ApplicationModel.Resources.idl
@@ -11,7 +11,7 @@ namespace Microsoft.Windows.ApplicationModel.Resources
     runtimeclass ResourceLoader
     {
         ResourceLoader();
-        ResourceLoader(String fileName);
+        ResourceLoader(String resourceMap);
         ResourceLoader(String fileName, String resourceMap);
 
         static String GetDefaultResourceFilePath();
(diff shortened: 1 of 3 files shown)


lauren-ciha:
Thank you for your contribution! In this case, the docs need an update. [Here is the PR for that change](https://github.com/MicrosoftDocs/winapps-winrt-api/pull/122).

PR #4370, opened 2024-04-26
Completing addition of Nuget Installation Wizard to all .vstemplate files

ShashankNay:
This change removes package references to C++ item template classes and adds the NuGetInstaller to the C# class library.


kythant (approved):

Scottj1s:
I'm a bit confused - why do item templates need package references?

Scottj1s (approved):

PR #4372, opened 2024-04-26
Improved Logging v3

DrusTheAxe:
Revise [Better logging support, replaces LOG_HR_MSG with DebugLog #4271](https://github.com/microsoft/WindowsAppSDK/pull/4271) as discussed with @pratikone 

Goal: Preserve non-failure information for troubleshooting production environments while not routing 'false failure' information to the debug console (e.g. VS' Output window when debugging an app). The evolutionary story:

v1: LOG_HR_MSG(E_NOTFAILURE_INFORMATIONAL_EVENT, "...info...") provided information for troubleshooting non-dev scenarios but also showed up in the debugger console as (false) errors. Meets non-dev needs but fails to meet desired dev experience

v2: OutputDebugString("...info...") routed non-failure troubleshooting into to the debugger console as non-failure information but not captured by non-developer scenarios (FeedbackHub, wpr, more). Meets desired dev experience but regressed non-dev scenarios.

v3 (this PR): TraceLoggingWrite() routing non-failure troubleshooting info to common non-dev support mechanisms (FeedbackHub, wpr, more). Meets desired dev and non-dev scenarios. Also provides a small performance improvement.

NOTE: v3 doesn't route non-failure information to the debug console. This isn't a problem as those developers interested in the information can use the debugger to inspect the processing flows for equivalent result (e.g. "Package Foo v1.2.3.4 is installed but why isn't it being used?"), or enable ETW capture, or use non-debugger solutions.

Also updated Coding-Guidelines to spell out standard practices for error handling and logging (previously cultural and oral tradition but not written down)

Or one could say, the error handling and logging policies are now written down and the egregious offenders are brought into compliance :-)

Also deleted some dead code (e.g. #if 1...#else...dead code...#endif) and obsolete custom error codes (noted as 'reserved' to prevent future use of those values, to avoid ambiguity)


dev/DynamicDependency/API/MsixDynamicDependency.h
```diff
@@ -8,10 +8,18 @@
 
 #include <stdint.h>
 
-// MSIX Dynamic Dependency HRESULT: Data Store not found (Windows App Runtime's Main package not registered?)
+/// MSIX Dynamic Dependency HRESULT: Windows App Runtime is not in the package graph.
+#define MDD_E_WINDOWSAPPRUNTIME_NOT_IN_PACKAGE_GRAPH    _HRESULT_TYPEDEF_(0x80040001L)
```
bpulliam, line 12: I understand we should maintain back compat, but reusing/creating HRESULTs is not a good pattern.  Running err in razzle gives this output: # for hex 0x80040001 / decimal -2147221503
  E_DISCVOL_VOLUME_TOO_SMALL                                     discvol.h
  DU_E_OUTOFKERNELRESOURCES                                      dusererror.h
# There are not enough Kernel resources to perform the
# operation
  E_COLORCPL_NO_SUCH_DEVICE                                      icmpriv.h
  E_NOTVALIDFORANIMATEDIMAGE                                     shimgdata.h
  TAPI_E_NOTENOUGHMEMORY                                         tapi3err.h
# The buffer passed in to this method was not big enough.
  IAP_E_BAD_PRODUCTID_LIST                                       windowsstoresdkerrors.h
# The supplied product list is bad. Check the Store event log
# for more details.
  OLE_E_ADVF                                                     winerror.h
# Invalid advise flags
  XAUDIOPACKETSTATUS_XMA_BUFFER_ERROR                            xauddefs.h
# as an HRESULT: Severity: FAILURE (1), FACILITY_IO_ERROR_CODE (0x4), Code 0x1
# for hex 0x1 / decimal 1
  STATUS_WAIT_1                                                  ntstatus.h
# as an HRESULT: Severity: FAILURE (1), FACILITY_ITF (0x4), Code 0x1
  ERROR_INVALID_FUNCTION                                         winerror.h
# Incorrect function.
# 10 matches found for "80040001"

DrusTheAxe: > I understand we should maintain back compat, but reusing/creating HRESULTs is not a good pattern.  

Yes and no. 

Yes, it's good. This HRESULT is a FACILITY_ITF (4) which is an open namespace. ALL 0x8004xxxx are, by definition, not guaranteed unique across space and time and dependent on current context. This is how you define new HRESULTs if you're not inside the Windows build and can stake out new unique values in winerror.h

That said, No it's bad as guidelines recommend only using values 0x0200 - 0xFFFF

https://learn.microsoft.com/en-us/windows/win32/com/codes-in-facility-itf

> All the COM-defined FACILITY_ITF codes have a code value in the range of 0x0000-0x01FF. While it is legal to use any codes in FACILITY_ITF, it is recommended that only code values in the range of 0x0200-0xFFFF be used. This recommendation is made as a means of reducing confusion with any COM-defined errors.

I'd like to change the value but was waiting for 2.0 when breaking changes are acceptable.

Unless you feel strongly about it and approve breaking the usual compat policy in minor updates? If so I'll make the change in main for 1.6

pratikone (changes requested):

pratikone (approved):

dev/PackageManager/API/M.W.M.D.PackageDeploymentManager.cpp
```diff
@@ -54,14 +54,15 @@ namespace winrt::Microsoft::Windows::Management::Deployment::implementation
...
             {
-                Common::Logging::DebugLog(std::format(L"Id={} PackageFamilyName={} MinVersion={}.{}.{}.{} ArchitectureFilter:0x{}",
-                            packageSetItem.Id().c_str(),
-                            packageSetItem.PackageFamilyName().c_str(),
-                            packageSetItem.MinVersion().Major,
-                            packageSetItem.MinVersion().Minor,
-                            packageSetItem.MinVersion().Build,
-                            packageSetItem.MinVersion().Revision,
-                            static_cast<std::uint32_t>(packageSetItem.ProcessorArchitectureFilter())));
+                (void)LOG_HR_MSG(MSIXPACKAGEMANAGER_E_PACKAGE_SCAN_FAILED,
```
Scottj1s, line 57: LOG_HR_MSG is what Pratik was trying to avoid, as it clutters OutputDebugString logging with meaningless HRESULT descriptions.

dev/PackageManager/API/PackageDeploymentResolver.cpp
```diff
@@ -179,12 +181,14 @@ winrt::hstring Microsoft::Windows::ApplicationModel::PackageDeploymentResolver::
     auto packages{ packageManager.FindPackagesForUserWithPackageTypes(winrt::hstring(), packageFamilyName, packageTypes) };
 
     // Find the/any match
-    auto criteria{ wil::str_printf<wil::unique_cotaskmem_string>(L"PackageFamilyName=%ls MinVersion=%hu.%hu.%hu.%hu ArchitectureFilter:0x%X",
-                                                                 packageFamilyName.c_str(),
-                                                                 minVersion.Major, minVersion.Minor,
-                                                                 minVersion.Build, minVersion.Revision,
-                                                                 static_cast<std::uint32_t>(processorArchitectureFilter)) };
-    Common::Logging::DebugLog(std::format(L"PackageDeploymentResolver: Scanning packages ({})", criteria.get()));
+    TraceLoggingWrite(
```
Scottj1s, line 184: Does TraceLoggingWrite write to both ETL and OutputDebugString?

dev/DynamicDependency/API/MsixDynamicDependency.h
Scottj1s: I like some of the contrary responses here: https://stackoverflow.com/questions/53199912/encoding-hresult-values-use-facility-itf-or-set-bit-29

I.e., mostly "no" 

I agree if we're going to change HRESULTs, it would have to wait until 2.0.  I'd suggest we shouldn't even do it then.  Ideally, we would reuse well-known HRESULTs that have some semblance of the WAR operation failure.  But we are stuck with these.

dev/PackageManager/API/PackageDeploymentResolver.cpp
```diff
@@ -260,13 +307,22 @@ winrt::hstring Microsoft::Windows::ApplicationModel::PackageDeploymentResolver::
     // Did we find what we're looking for?
     if (bestFitPackageFullName.empty())
     {
-        Common::Logging::DebugLog(std::format(L"PackageDeploymentResolver: No match ({})",
-                         criteria.get()));
+        (void)LOG_HR_MSG(MSIXPACKAGEMANAGER_E_PACKAGE_SCAN_FAILED,
```
Scottj1s, line 310: another return to LOG_HR_MSG - unless these are absolutely necessary, we should avoid them and just go with direct TraceLoggingWrite/OutputDebugString calls to avoid the FACILITY_ITF confusion that Pratik was originally trying to address.  I would think all calls to LOG_HR_MSG are not really error scenarios (they are logging after all), so could be avoided.

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -886,14 +879,18 @@ void FindDDLMViaEnumeration(
...
     const auto c_packageTypes{ winrt::Windows::Management::Deployment::PackageTypes::Main };
     auto packages{ packageManager.FindPackagesForUserWithPackageTypes(currentUser, c_packageTypes) };
+    TraceLoggingWrite(
+        WindowsAppRuntimeBootstrap_TraceLogger::Provider(),
+        "Bootstrap.Initialize.DDLM.Scan",
+        TraceLoggingHexUInt32(majorMinorVersion, "Criteria.MajorMinorVersion"),
+        TraceLoggingWideString(!versionTag ? L"" : versionTag, "Criteria.VersionTag"),
+        TraceLoggingHexUInt64(minVersion.Version, "Criteria.MinVersion"),
+        TraceLoggingLevel(WINEVENT_LEVEL_VERBOSE),
+        TelemetryPrivacyDataTag(PDT_ProductAndServicePerformance));
```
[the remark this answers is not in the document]
Scottj1s: I was thinking the same thing initially, but it wouldn't be trivial - a variadic template function in order to capture variable nested calls for tracing strings and ints.  probably not worth it.

docs/Coding-Guidelines/ErrorHandlingAndLogging.md
```diff
@@ -0,0 +1,17 @@
+# Error Handling and Logging
```
Scottj1s, line 1: like this - thx

docs/Coding-Guidelines/ErrorHandlingAndLogging.md
```diff
@@ -0,0 +1,17 @@
...
+<span style="color:green">**DO**</span> Throw exceptions to report exceptional failures.
+
+<span style="color:green">**DO**</span> Use [WIL error handling helpers](https://github.com/Microsoft/wil/wiki/Error-handling-helpers) when handling error code failures (`THROW_*()`, `RETURN_*()`, `FAIL_FAST_*()`, `LOG_*()`).
+
+# Logging
+
+<span style="color:green">**DO**</span> Rely on [WIL error handling helpers](https://github.com/Microsoft/wil/wiki/Error-handling-helpers) to log failures.
+
+ <span style="color:green">**DO**</span> Use `TraceLoggingWrite()` to log non-failure information.
+
```
Scottj1s, line 14:
```suggestion

<span style="color:green">**DON'T**</span> Define error 'HRESULTS' using FACILITY_ITF.  This is a general-purpose category that results in confusing logging messages.  Instead, use an existing HRESULT that approximates the error described. ```

Scottj1s:
> NOTE: v3 doesn't route non-failure information to the debug console. This isn't a problem as those developers interested in the information can use the debugger to inspect the processing flows for equivalent result (e.g. "Package Foo v1.2.3.4 is installed but why isn't it being used?"), or enable ETW capture, or use non-debugger solutions.

"inspect the processing flows" 
what does this mean? sounds hand-wavy and cumbersome, vs simply observing debug spew

"or use non-debugger solutions"
what does this mean?  if logging messages are now only routed to ETL, what other options are there?

I think we should strive to preserve the ODS behavior of LOG_HR_MSG, without using it.  Maybe we need a wrapper for TraceLoggingWrite to achieve that.

dev/PackageManager/API/PackageDeploymentResolver.cpp
DrusTheAxe: ByDesign. The others were changed as they weren't Errors, they're Information that used WIL to log and thus had to be a FAILED(hr).

In this case it is an error. This reports the error details but the function returns empty string for not found. If it returned HRESULT this would be RETURN_HR(). We return empty string rather than throwing an exception as this isn't an exceptional case and exception flow up the stack would have be inappropriate.

Maybe in theory this could be rearchitected to throw but the ripple impact up the stack can get invasive so it's not worth the effort.

docs/Coding-Guidelines/ErrorHandlingAndLogging.md
DrusTheAxe: Made it simple - document the rules, then Uhoh! We have a violator! Fix that! :P

dev/PackageManager/API/PackageDeploymentResolver.cpp
DrusTheAxe: TraceLoggingWrite() writes to ETW

OutputDebugString() writes to ODS

WIL writes to both (technically WIL writes to ODS and an optional callback, but we register a callback in the usual way to wire up ETW support)

But for these Informational events we don't want ODS.

ETW is important for non-dev support scenarios but if you're a dev in VS you either don't care about this information, or if you really do you can debug into the source to see the actual code flow (equally or more meaningful than the previous LOG_HR to VS' Output window)

docs/Coding-Guidelines/ErrorHandlingAndLogging.md
DrusTheAxe: FACILITY_ITF is how you define new HRESULTs (if you're not in the Windows source tree).

You should use existing HRESULTs if possible, but if not then yes you should use FACILITY_ITF to define a new HRESULT.

How's this?

<span style="color:green">**DO**</span> Use existing 'HRESULTS' if you can. Be careful to pick an existing HRESULT whose symbolic name and message convey the intent and don't mislead the developer, administrator or user.

Scottj1s: that works

dev/PackageManager/API/PackageDeploymentResolver.cpp
Scottj1s: Are we confident that nobody will miss this ODS output - that this is not a breaking change that people will mind?

DrusTheAxe:
> "inspect the processing flows" what does this mean? sounds hand-wavy and cumbersome, vs simply observing debug spew

If you're in the debugger walk thru the debugger. Or just capture ETW events for the meaningful data.

Question is, if you're in the debugger, do you care about this data? In this case methinks no, so there's little value in ODS here. If you do you can capture ETW events, or if you really care set a breakpoint in this code and see the details. The latter will be exceedingly rare.

> "or use non-debugger solutions" what does this mean?

Use WPR.EXE or other mechanisms that capture ETW events.

>what other options are there?

Root question, there's only...

* ETW
* OutputDebugString
* EventLog
* roll-it-yourself (e.g. sprintf-append-to-text-file)

ETW and ODS are fundamentally different approaches

* ETW records raw data and formatting is deferred to view time
* ODS records a string. Any data must be formatted into it at runtime

ODS has other issues (e.g. it's a session? machine? global resource). WIL partially mitigates this by scoping its calls to ODS only when debugging via a check of `IsDebuggerPresent()`. WIL also has lots advanced behavior and little details already baked out e.g. IsDebuggerPresent() is called once and cached to minimize the overhead of later checks. It's a shame WIL has no `LOGINFO_*()` as its functionality is pretty deep.

I started with taking a stab at adding a `LOGINFO_HR[_MSG]()` but WIL's implementation is deeply layered. To get LOG_HR style functionality but w/o the `FAIL_FAST_HR_IF(..., type=Logging and SUCCEEDED(HR))` required changing callstacks a dozen levels down to pass a additional parameter to not fault on LOGINFO, or cloning out a dozen+ functions. At which point it occurred to me TraceLoggingWrite is the norm how we handle such things and a better fit than trying to double-down on forking WIL.

> I think we should strive to preserve the ODS behavior of LOG_HR_MSG (without using LOG_HR_MSG). Maybe we need a wrapper for TraceLoggingWrite to achieve that.

> I had hoped (maybe naively) that WIL would provide some kind of layered API or hooks to allow something like LOG_HR_MSG but without an HR - LOG_MSG.

WIL is hard-wired to fault if called by one of the LOG*() functions.

If you're interested I think the most fruitful option is to revisit the `LOGINFO_HR[_MSG]()` approach, for a net-result like `LOG_HR[_MSG]()` except (1) any HRESULT (no failfast if SUCCEEDED(hr)), and (2) the string handed to ODS is prefixed with "Information:..." or some such (nothing implying an 'error')

We can discuss that further offline if you're interested

Scottj1s:
> We can discuss that further offline if you're interested

Not necessary - just want to be clear that we're not okay with a breaking change that removes ETL, but we are okay with a breaking change that removes ODS.

dev/PackageManager/API/M.W.M.D.PackageDeploymentManager.cpp
Scottj1s: how was this resolved?

PR #4378, opened 2024-04-30
Fix Prefast C6101 on Foundation Solution 

kythant:

WARNING:

### DIFF
--- a/dev/UndockedRegFreeWinRT/typeresolution.cpp
+++ b/dev/UndockedRegFreeWinRT/typeresolution.cpp
@@ -231,9 +231,8 @@ namespace UndockedRegFreeWinRT
         }
 
         wchar_t szCandidateFileName[MAX_PATH + 1]{};
+        RETURN_IF_FAILED(WindowsCreateString(L"", 0, phstrMetaDataFilePath));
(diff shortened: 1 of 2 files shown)


dev/UndockedRegFreeWinRT/typeresolution.cpp
```diff
@@ -213,8 +221,19 @@ namespace UndockedRegFreeWinRT
...
+            *ppMetaDataImport = nullptr;
+        }
+        if (pmdTypeDef != nullptr)
+        {
+            *pmdTypeDef = 0;
+        }
+
         wchar_t szCandidateFileName[MAX_PATH + 1]{};
         HRESULT hr{ StringCchCopy(szCandidateFileName, ARRAYSIZE(szCandidateFileName), pszFullName) };
+        hr = WindowsCreateString(L"", 0, phstrMetaDataFilePath);
```
alexlamtest, line 235: Is it intentional that we now overwrite the hr result from the stringCchCopy() w/o first checking for error?

kythant: oh good point

PR #4397, opened 2024-05-08
Fix initialization of _InOut_ variable

bpulliam:
A recent prefast bug fix initialized an _Out_ variable athat was actually an _InOut_, so that the initialization clobbered the existing value. The parameter was not being used as an out param, so the correct fix is to change it to a const.


Scottj1s (approved):

Scottj1s:
actually a pure in param, so a const & could be used

PR #4414, opened 2024-05-14
Changing Default Platform to x64

ShashankNay:
Changed the preferred solution configuration from `Debug|x86` to `Debug|x64` in the project template to align with the target platform requirements.


Scottj1s (approved):
thx!

PR #4451, opened 2024-05-27
Fix missing trimming annotation and set couple cswinrt properties

manodasanW:
We have `IsTrimmable` set in our projection projects.  But this doesn't show up in the resulting assembly as an assembly attribute.  This turns out it is due to at some point we started using a custom `AssemblyInfo.cs` which meant the attribute that was generated in the default one is no longer there.  While checking the assembly I also noticed the `IIDOptimizer` is inserting a type that is unused given nothing needs to be optimized and that the vtable lookup generation table was generated because CsWinRT failed to properly detect this as a projection project.  So, disabling those with the respective properties to control them.

WARNING:

### DIFF
--- a/dev/Projections/Directory.Build.props
+++ b/dev/Projections/Directory.Build.props
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE-CODE in the project root for license information. -->
+<!-- This is a special MSBuild file that is parsed before everything else when MSBuild detects it in our directory structure.
+     Adding Sdk="Microsoft.NET.Sdk" to a project causes NuGet files to be generated very early on, so we need these defines to be here
+     to ensure that all of our build output files end up in the same location. -->
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="..\..\Directory.Build.props" />
+
+  <PropertyGroup>
+    <!-- Both the IID optimizer and the CsWinRT lookup table aren't needed with the new CsWinRT version,
+         so disabling them here until they are disabled automatically by CsWinRT. -->
+    <CsWinRTIIDOptimizerOptOut>true</CsWinRTIIDOptimizerOptOut>
+    <CsWinRTCcwLookupTableGeneratorEnabled>false</CsWinRTCcwLookupTableGeneratorEnabled>
+  </PropertyGroup>
+</Project>
--- a/eng/common/VersionInfo/AssemblyInfo.cs
+++ b/eng/common/VersionInfo/AssemblyInfo.cs
@@ -15,3 +15,5 @@ using System.Runtime.InteropServices;
 
 [assembly: AssemblyVersion("0.0")]
 [assembly: AssemblyFileVersion("0.0")]
+
+[assembly: AssemblyMetadata("IsTrimmable", "True")]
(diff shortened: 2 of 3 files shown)


dev/Projections/Directory.Build.props
```diff
@@ -0,0 +1,15 @@
...
+<!-- This is a special MSBuild file that is parsed before everything else when MSBuild detects it in our directory structure.
+     Adding Sdk="Microsoft.NET.Sdk" to a project causes NuGet files to be generated very early on, so we need these defines to be here
+     to ensure that all of our build output files end up in the same location. -->
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="..\..\Directory.Build.props" />
+
+  <PropertyGroup>
+    <!-- Both the IID optimizer and the CsWinRT lookup table aren't needed with the new CsWinRT version,
+         so disabling them here until they are disabled automatically by CsWinRT. -->
+    <CsWinRTIIDOptimizerOptOut>true</CsWinRTIIDOptimizerOptOut>
```
Scottj1s, line 12: Is the IIDOptimizer totally obsoleted then, or only when PublishAOT=true?  Is there any downside to leaving it on (e.g., in feeder repos), other than the unused type you mention?

manodasanW: yea it is going to be no longer needed in general for newly generated projections at least.  We generate all the IIDs within the projection as we need to expose them as a property in the static methods class.  If someone still had a call to like GuidGenerator in their own code, it can still optimize those if they want to keep it.

manodasanW: There is no downside other than not using it anymore is allowing us to fix some of the bugs that came with it such as breaking sourcelink support from rewriting pdbs / dlls.

Scottj1s: should all feeder projections also be updated?

manodasanW: I have similar changes for WinUI.  IXP is not impacted by this as they didn't run IIDOptimizer.  Otherwise, others mainly just Widgets would get these by default in the next CsWinRT update I do, I just didn't want to do that right now.  I decided to do it here manually for now due to the # of projections we produce in this repo.

dev/Projections/Directory.Build.props
```diff
@@ -0,0 +1,15 @@
...
+    <CsWinRTCcwLookupTableGeneratorEnabled>false</CsWinRTCcwLookupTableGeneratorEnabled>
```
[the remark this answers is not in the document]
manodasanW: Yes, once I get the next CsWinRT update (after current preview one), this would not be needed anymore.   I mainly put this just here because this repo produces multiple projections and the same thing is generated in each one when it is not needed.

eng/common/VersionInfo/AssemblyInfo.cs
```diff
@@ -15,3 +15,5 @@
 
 [assembly: AssemblyVersion("0.0")]
 [assembly: AssemblyFileVersion("0.0")]
+
+[assembly: AssemblyMetadata("IsTrimmable", "True")]
```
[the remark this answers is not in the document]
manodasanW: This is just an issue here. The projection projects all are already marked trimmable here similar to feeder projections too.  But we are overriding the AssemblyInfo.cs which means the effect of setting the property doesn't take effect as we aren't using the C# generated default AssemblyInfo.cs which seems unique here in this repo.

Scottj1s (approved):

PR #4453, opened 2024-05-28
PackageDeploymentManager: IsReadyOrNewerAvailable(), FrameworkUdk update, Implement not-implemented functions, Tests!

DrusTheAxe:
This PR does multiple things
* Add contract version 2
  * Is\*ReadyOrNewerAvailable\*()
  * EnsureReadyOptions.RegisterNewerIfAvailable=bool
  * Is\*Provisioned\*()
* Fix progress updates when PackageSet.Items.Count>=2 (was (re)set to 0-1.0 for each item, not collectively for the whole operation)
* Update to recent FrameworkUdk
* Implement previously not-implemented functions (Remove, Repair, Reset)
* Expanded test coverage
* Updated TAEF to 10.92.240405002
* Fixed DevCheck error accessing non-existent registry value if DeveloperMode was never set. Changed to use more resilient mechanism; this was the last of the older less-resilient syntax in the code

KNOWN ISSUES:
1. Need updated Platform SDK to resolve known issues
2. Register*Async() errors on 20H2/Vb3.
3. Reset*Async() succeeds but ends w/o the package being registered
4. RemovePackageByUriAsync() throws E_NOTIMPL
5. RemovePackageSetAsync() throws E_NOTIMPL if PackageUri=ms-uup
6. Provision\*Async() not supported yet
7. Deprovision\*Async() not supported yet

https://task.ms/49555034
https://task.ms/48914090
https://task.ms/48964562


dev/PackageManager/API/M.W.M.D.PackageDeploymentManager.cpp
rhuang-msft, line 1: General question: why do we reimplement/create wrappers on public PackageManager APIs?

DrusTheAxe: If all we did was wrap OS APIs it would be pointless. Not something we do. The WinAppSDK package management APIs bring value to the table in multiple ways including:
* New functionality
* Platform availability / polyfill
* Developer experience
* Error handling
* Modern design practices

WinAppSDK's [overarching goal](https://github.com/microsoft/WindowsAppSDK/?tab=readme-ov-file#version-support) is to "provide libraries, frameworks, components, and tools that you can use in your apps to access powerful Windows platform functionality from all kinds of apps ***on many versions of Windows***" (emphasis mine). That's difficult to do by enhancing ' PackageManager APIs.

Some of what you see as 'wrapping' may be functionally correct but inaccurate. PackageManager is one of the oldest WinRT APIs, with many designs dating back to earliest Win8 days. Practices and patterns have evolved since then in many ways (big and small) leaving PackageManager increasing awkward and difficult to use. 'Modernizing' PackageManager would pose severe (unacceptable) compatibility issues.

For one example, PackageManager's error handling is problematic with projections and common practices. To be fair, PackageManager was designed, implemented and continued to evolve for many years as an ABI (or using WRL and like thin ABI-oriented libraries). This well predates C++/WinRT, C#/WinRT, Rust/WinRT and other modern projections. This led to the necessity of the `PackageDeploymentResult` object having both `Error` and `ExtendedError` properties (see [lines 114-119](https://github.com/microsoft/WindowsAppSDK/blob/main/dev/PackageManager/API/PackageManager.idl)) plus some creative implementation in `PackageDeploymentManager`. This is possible for developers to handle themselves but at extreme cost, and requiring an advanced understanding of the APIs and behavior. And even then it's very challenging, unless you have access to PackageManager's full source implementation (i.e. unless you're a Windows developer). This is an example where only we (Microsoft) can do it well.

For another example, PackageManager.RemovePackageAsync(pkgfullname) errors with `HRESULT_FROM_WIN32(ERROR_INSTALL_PACKAGE_NOT_FOUND)` if the pkg isn't found, but handling that is cumbersome and difficult (see error handling comment above). In a common case the goal is the package isn't present when done, but whether the API removed it or it was already gone isn't interesting or meaningful ('Tis the destination the matters, not the journey). The `RemovePackageOptions.FailIfNotFound` option solves this issue. We could have added that to Windows' PackageManager APIs but doing so across many Windows releases is challenging, and leaves apps with the problem of does a particular machine they're running on have an OS update with the enhancement? WinAppSDK gives developers the ability to _rely_ on the feature as they can **guarantee** the API's present on the machine when they run. YourApp can't upgrade or patch the OS, but YourApp can install the WinAppSDK if necessary.

For all these reasons and more the WinAppSDK package management APIs were designed to be fully featured and not a a partial supplement to PackageManager.

If you have additional questions we can discuss it offline.

rhuang-msft: Sounds good. Thanks!

rhuang-msft (approved):

PR #4479, opened 2024-06-07
[mrtcore] Microsoft.Windows.Globalization.ApplicationLanguages API specs

stefansjfw:

Changing the application language using [PrimaryLanguageOverride](https://learn.microsoft.com/uwp/api/windows.globalization.applicationlanguages.primarylanguageoverride) property from `Windows.Globalization.ApplicationLanguages` namespace is not supported for WindowsAppSDK un-packaged apps, only for packaged. To support language change for unpackaged applications `Microsoft.Windows.Globalization.ApplicationLanguages` type is introduced in
MRTCore. `Microsoft.Windows.Globalization.ApplicationLanguages.PrimaryLanguageOverride` property is supported both for packaged and unpackaged WindowsAppSDK apps.

WARNING:

### DIFF
--- a/specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
+++ b/specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
@@ -44,9 +44,10 @@ Note that these extensions can affect the numeral system or calendar used by glo
 ## ApplicationLanguages.PrimaryLanguageOverride property
 
 Gets or sets an override for the app's preferred language, expressed as a
-[BCP-47 language tag](https://tools.ietf.org/html/bcp47). Unlike
+[BCP-47 language tag](https://tools.ietf.org/html/bcp47). This setting is global for the running
+process. Unlike
 [Windows.Globalization.ApplicationLanguages.PrimaryLanguageOverride](https://learn.microsoft.com/uwp/api/windows.globalization.applicationlanguages.primarylanguageoverride),
-this setting is global for the running process and is not persisted.
+this setting is not persisted between app sessions.
 
 ```c#
 public static string PrimaryLanguageOverride { get; set; }
--- a/specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
+++ b/specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
@@ -1,24 +1,23 @@
 # ApplicationLanguages
 
-This new type (`Microsoft.Windows.ApplicationModel.Resources.ApplicationLanguages`) is a wrapper for
-the OS `Windows.Globalization.ApplicationLanguages` type with the updated behavior for the
-`PrimaryLanguageOverride` property to be supported for un-packaged apps.
+This new type (`Microsoft.Windows.Globalization.ApplicationLanguages`) is a wrapper for the OS
+`Windows.Globalization.ApplicationLanguages` type with the updated behavior for the
+`PrimaryLanguageOverride` property to be supported for unpackaged apps.
 
 # Background
 
 Changing the application language using
 [PrimaryLanguageOverride](https://learn.microsoft.com/uwp/api/windows.globalization.applicationlanguages.primarylanguageoverride)
 property from `Windows.Globalization.ApplicationLanguages` namespace is not supported for
-WindowsAppSDK un-packaged apps, only for packaged. To support language change for un-packaged
-applications `ApplicationLanguages` type is introduced in
-`Microsoft.Windows.ApplicationModel.Resources` MRTCore namespace.
-`Microsoft.Windows.ApplicationModel.Resources.ApplicationLanguages.PrimaryLanguageOverride` property
-is supported both for packaged and un-packaged WindowsAppSDK apps.
+WindowsAppSDK unpackaged apps, only for packaged. To support language change for unpackaged
+applications, `Microsoft.Windows.Globalization.ApplicationLanguages` type is introduced in
+MRTCore.`Microsoft.Windows.Globalization.ApplicationLanguages.PrimaryLanguageOverride` property is
+supported both for packaged and unpackaged WindowsAppSDK apps.
 
-This new type (`Microsoft.Windows.ApplicationModel.Resources.ApplicationLanguages`) is a wrapper for
-the OS type with the updated behavior for the `PrimaryLanguageOverride` to support un-packaged apps.
-`Microsoft.Windows.ApplicationModel.Resources.ApplicationLanguages` should serve as a replacement
-for the existing OS type (`Windows.Globalization.ApplicationLanguages.PrimaryLanguageOverride`).
+This new type (`Microsoft.Windows.Globalization.ApplicationLanguages`) is a wrapper for the OS type
+with the updated behavior for the `PrimaryLanguageOverride` to support unpackaged apps.
+`Microsoft.Windows.Globalization.ApplicationLanguages` should serve as a replacement for the
+existing OS type (`Windows.Globalization.ApplicationLanguages.PrimaryLanguageOverride`).
 
 `GetLanguagesForUser(User)` method from `Windows.Globalization.ApplicationLanguages` is being
 intentionally left off due to not being relevant for WinAppSDK.
@@ -44,10 +43,10 @@ claims about conventions that are used in the US such as the measurement system 
 Example:
 
 ```c#
-Microsoft.Windows.ApplicationModel.Resources.ApplicationLanguages.PrimaryLanguageOverride = "en-US";
+Microsoft.Windows.Globalization.ApplicationLanguages.PrimaryLanguageOverride = "en-US";
 ```
 
-Language tags support the Unicode extensions "ca-" and "nu-". (See Unicode Key/Type Definitions.)
+Language tags support the Unicode extensions "ca-" and "nu-" (See Unicode Key/Type Definitions.).
 Note that these extensions can affect the numeral system or calendar used by globalization objects.
 
 ## ApplicationLanguages.Languages property
@@ -78,18 +77,17 @@ section in
 
 ## ApplicationLanguages.ManifestLanguages property
 
-Gets the app's declared list of supported languages.
+Gets the declared list of supported languages in the app's manifest, or empty list for unpackaged
+applications.
 
 # Property Value
 
 [IReadOnlyList](https://learn.microsoft.com/dotnet/api/system.collections.generic.ireadonlylist-1)<[String](https://learn.microsoft.com/dotnet/api/system.string)>
 
-The list of supported languages declared in the app's manifest. For un-packaged applications,
-exception is thrown.
-
-Note: This property returns the same values as the language list exposed by
+For packaged applications, the list of supported languages declared in the app's manifest, same as
+the language list exposed by
 [Windows.Globalization.ApplicationLanguages.ManifestLanguages](https://learn.microsoft.com/uwp/api/windows.globalization.applicationlanguages.manifestlanguages)
-property.
+property. For unpackaged applications, empty list.
 
 # Remarks
 
@@ -108,7 +106,7 @@ Gets or sets an override for the app's preferred language, expressed as a
 wrapper for
 [Windows.Globalization.ApplicationLanguages.PrimaryLanguageOverride](https://learn.microsoft.com/uwp/api/windows.globalization.applicationlanguages.primarylanguageoverride)
 property and behaves the same. This setting is global for the running process. Unlike for packaged
-apps, for un-packaged apps, this setting is not persisted between app sessions.
+apps, for unpackaged apps, this setting is not persisted between app sessions.
 
 ```c#
 public static string PrimaryLanguageOverride { get; set; }
@@ -119,10 +117,16 @@ public static string PrimaryLanguageOverride { get; set; }
 [String](https://learn.microsoft.com/dotnet/api/system.string)
 
 A [BCP-47 language tag](https://tools.ietf.org/html/bcp47). The app can set it to override the
-language. It must be a single language tag; a delimited list of language tags will result in default
-language being used.
+language. It must be a single language tag; if delimited list of language tags or any value not
+representing a language tag is used, exception is thrown.
 
-The PrimaryLanguageOverride property should only be set to languages available for the app.
+The PrimaryLanguageOverride property should only be set to languages available for the app. For
+packaged apps, languages available for the app to use are limited to those languages included in the
+main app package manifest. The ApplicationLanguages.ManifestLanguages property reflects language
+resource packages that are available for the user, and returns an appropriate set of languages that
+can be used for setting the PrimaryLanguageOverride property. For unpackaged apps, languages
+available for the app to use are limited to languages for which your app has explicitely declared
+support.
 
 When your app gets the value, PrimaryLanguageOverride returns either a single language tag (if your
 app has previously set the property) or an empty string.
@@ -135,14 +139,15 @@ is used to override that behavior by setting a single specific language.
 
 For packaged apps, `PrimaryLanguageOverride` setting is a wrapper for
 [Windows.Globalization.ApplicationLanguages.PrimaryLanguageOverride](https://learn.microsoft.com/uwp/api/windows.globalization.applicationlanguages.primarylanguageoverride)
-property and behaves the same. It is persisted between app sessions.
+property and behaves the same. It is persisted between app sessions. The property should only be set
+based on user input presented in settings UI.
 
-For un-packaged apps, `PrimaryLanguageOverride` setting is not persisted between sessions. It should
+For unpackaged apps, `PrimaryLanguageOverride` setting is not persisted between sessions. It should
 be set each time the app is loaded. You should set PrimaryLanguageOverride setting in early stage of
 app loading, before any resource is loaded.
 
-The property should only be set based on user input presented in settings UI. It can be read at any
-time. If the property has never been set, it returns an empty string.
+The property can be read at any time. If the property has never been set, it returns an empty
+string.
 
 When you set the PrimaryLanguageOverride, this will be immediately reflected in resources loaded
 afterwards, both resources loaded from the code and the resources loaded by XAML. However, this
@@ -162,9 +167,9 @@ public partial class App : Application
         // This will cause the app to use the German (Germany) resources
         // Both resources loaded manually with ResourceLoader and XAML resources fetched with x:Uid
         // will be German (Germany)
-        Microsoft.Windows.ApplicationModel.Resources.ApplicationLanguages.PrimaryLanguageOverride = "de-DE";
+        Microsoft.Windows.Globalization.ApplicationLanguages.PrimaryLanguageOverride = "de-DE";
 
-        Microsoft.Windows.ApplicationModel.Resources.ResourceLoader resourceLoader = new Microsoft.Windows.ApplicationModel.Resources.ResourceLoader();
+        var resourceLoader = new Microsoft.Windows.ApplicationModel.Resources.ResourceLoader();
         var resourceString = resourceLoader.GetString("SampleString");
 
         this.InitializeComponent();
@@ -176,7 +181,7 @@ public partial class App : Application
 # API Details
 
 ```c# (but really MIDL3)
-namespace Microsoft.Windows.ApplicationModel.Resources
+namespace Microsoft.Windows.Globalization
 {
   // ...
   [contract(MrtCoreContract, 2)]
--- a/specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
+++ b/specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
@@ -1,27 +1,36 @@
-# ApplicationLanguages.PrimaryLanguageOverride
+# ApplicationLanguages
 
-This feature provides the ability to change the application language for WindowsAppSDK applications,
-comparable to the one in
-[Windows.Globalization.ApplicationLanguages.PrimaryLanguageOverride](https://learn.microsoft.com/uwp/api/windows.globalization.applicationlanguages).
-The benefit of this addition is that it applies to both packaged and un-packaged WindowsAppSDK
-applications, whereas the existing one is not supported for un-packaged.
+This new type (`Microsoft.Windows.ApplicationModel.Resources.ApplicationLanguages`) is a wrapper for
+the OS `Windows.Globalization.ApplicationLanguages` type with the updated behavior for the
+`PrimaryLanguageOverride` property to be supported for un-packaged apps.
 
 # Background
 
 Changing the application language using
 [PrimaryLanguageOverride](https://learn.microsoft.com/uwp/api/windows.globalization.applicationlanguages.primarylanguageoverride)
 property from `Windows.Globalization.ApplicationLanguages` namespace is not supported for
-WindowsAppSDK un-packaged apps, only for packaged. To support the same behavior for un-packaged
-applications `PrimaryLanguageOverride` property is introduced in
-`Microsoft.Windows.ApplicationModel.Resources` MRTCore namespace. This property
-(`Microsoft.Windows.ApplicationModel.Resources.PrimaryLanguageOverride`) overrides the behavior of
-the existing one (`Windows.Globalization.ApplicationLanguages.PrimaryLanguageOverride`).
+WindowsAppSDK un-packaged apps, only for packaged. To support language change for un-packaged
+applications `ApplicationLanguages` type is introduced in
+`Microsoft.Windows.ApplicationModel.Resources` MRTCore namespace.
+`Microsoft.Windows.ApplicationModel.Resources.ApplicationLanguages.PrimaryLanguageOverride` property
+is supported both for packaged and un-packaged WindowsAppSDK apps.
+
+This new type (`Microsoft.Windows.ApplicationModel.Resources.ApplicationLanguages`) is a wrapper for
+the OS type with the updated behavior for the `PrimaryLanguageOverride` to support un-packaged apps.
+`Microsoft.Windows.ApplicationModel.Resources.ApplicationLanguages` should serve as a replacement
+for the existing OS type (`Windows.Globalization.ApplicationLanguages.PrimaryLanguageOverride`).
+
+`GetLanguagesForUser(User)` method from `Windows.Globalization.ApplicationLanguages` is being
+intentionally left off due to not being relevant for WinAppSDK.
 
 # API Pages
 
 ## ApplicationLanguages class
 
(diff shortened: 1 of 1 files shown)


specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
```diff
@@ -0,0 +1,121 @@
...
+    // ...
+    public App()
+    {
+        // Set the primary language override to German (Germany)
+        // This will cause the app to use the German (Germany) resources
+        // Both resources loaded manually with ResourceLoader and XAML resources fetched with x:Uid
+        // will be German (Germany)
+        Microsoft.Windows.ApplicationModel.Resources.ApplicationLanguages.PrimaryLanguageOverride = "de-DE";
+
+        this.InitializeComponent();
```
kmahone, line 103: This sample is improved over what was there before. I think it would be even better if it closed the loop and included a line of code that retrieves a localized resource programmatically, to show PrimaryLanguageOverride having an effect.

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
```diff
@@ -0,0 +1,121 @@
...
+# Background
+
+Changing the application language using
+[PrimaryLanguageOverride](https://learn.microsoft.com/uwp/api/windows.globalization.applicationlanguages.primarylanguageoverride)
+property from `Windows.Globalization.ApplicationLanguages` namespace is not supported for
+WindowsAppSDK un-packaged apps, only for packaged. To support the same behavior for un-packaged
+applications `PrimaryLanguageOverride` property is introduced in
+`Microsoft.Windows.ApplicationModel.Resources` MRTCore namespace. This property
+(`Microsoft.Windows.ApplicationModel.Resources.PrimaryLanguageOverride`) overrides the behavior of
+the existing one (`Windows.Globalization.ApplicationLanguages.PrimaryLanguageOverride`).
```
kmahone, line 18: "FooOverride overrides the behavior of BarOverride" is too many instances of the word 'override'! 

Maybe:
> This property
(`Microsoft.Windows.ApplicationModel.Resources.PrimaryLanguageOverride`) **takes precedence over**
the existing one (`Windows.Globalization.ApplicationLanguages.PrimaryLanguageOverride`).

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
```diff
@@ -0,0 +1,121 @@
...
+```
+
+Language tags support the Unicode extensions "ca-" and "nu-". (See Unicode Key/Type Definitions.)
+Note that these extensions can affect the numeral system or calendar used by globalization objects.
+
+## ApplicationLanguages.PrimaryLanguageOverride property
+
+Gets or sets an override for the app's preferred language, expressed as a
+[BCP-47 language tag](https://tools.ietf.org/html/bcp47). This setting is global for the running
+process and is not persisted.
```
kmahone, line 48: I would call out explicitly here that this is a difference in behavior from the `Windows.Globalization` api.

kmahone (changes requested):

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
stefansjfw: updated

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
stefansjfw: sounds good. Updated

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
stefansjfw: done.Thank you

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
```diff
@@ -0,0 +1,125 @@
...
+
+# Remarks
+
+Apps normally run with language settings determined by the system by comparing the languages
+supported by the app with the language preferences of the user. The PrimaryLanguageOverride property
+is used to override that behavior by setting a single specific language.
+
+The PrimaryLanguageOverride setting is not persisted between sessions. It should be set each time
+the app is loaded. You should set PrimaryLanguageOverride setting in early stage of app loading,
+before any resource is loaded. It should only be set based on user input presented in settings UI.
```
huichen123, line 76: is it possible to set early enough so that WinUI will be able to respect it?

stefansjfw: Setting it before InitializeComponent in App() constructor like presented here https://github.com/microsoft/WindowsAppSDK-Samples/pull/358/files#diff-af77dcd97ddae80357680b47edbe8940ebd597b31bf81dd152af69067e46f701R40
worked for me in every scenario I tested

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
stefansjfw: fixed

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
```diff
@@ -0,0 +1,206 @@
+# ApplicationLanguages.PrimaryLanguageOverride
```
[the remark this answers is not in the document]
kmahone: Also update the blurb below now that we are lifting the full class. Keep the detail that the motivation is to update the behavior of PrimaryLanguageOverride.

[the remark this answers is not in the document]
kmahone: We can also explicitly state that this new type is primarily a wrapper for the OS type, just with the updated behavior for PrimaryLanguageOverride.

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
```diff
@@ -0,0 +1,206 @@
...
+language being used.
+
+The PrimaryLanguageOverride property should only be set to languages available for the app.
+
+When your app gets the value, PrimaryLanguageOverride returns either a single language tag (if your
+app has previously set the property) or an empty string.
+
+# Remarks
+
+Apps normally run with language settings determined by the system by comparing the languages
```
kmahone, line 122: Is the new design that setting this property will also set the underlying OS property? If so, state as much. Also, if this is the case, then the property will be persisted for packaged apps. So we should state that also.

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
```diff
@@ -0,0 +1,206 @@
...
+        Microsoft.Windows.ApplicationModel.Resources.ResourceLoader resourceLoader = new Microsoft.Windows.ApplicationModel.Resources.ResourceLoader();
+        var resourceString = resourceLoader.GetString("SampleString");
+
+        this.InitializeComponent();
+    }
+    // ...
+}
+```
+
+## ApplicationLanguages.GetLanguagesForUser(User) Method
```
kmahone, line 165: I think MUA is not applicable to WinAppSDK apps, so I think we can actually remove this from this type.

Remove the api here and then update the Background to state that it is being intentionally left off due to not being relevant for WinAppSDK.

kmahone (changes requested):

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
```diff
@@ -0,0 +1,190 @@
...
+
+# Property Value
+
+[String](https://learn.microsoft.com/dotnet/api/system.string)
+
+A [BCP-47 language tag](https://tools.ietf.org/html/bcp47). The app can set it to override the
+language. It must be a single language tag; a delimited list of language tags will result in default
+language being used.
+
+The PrimaryLanguageOverride property should only be set to languages available for the app.
```
FrankC-msft, line 125: "should be", what happens when a developer sets an override that isn't one of the supported languages?  Is the "set" ignored, and getting the property again still returns the previous value?

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
```diff
@@ -0,0 +1,190 @@
...
+```c#
+public static string PrimaryLanguageOverride { get; set; }
+```
+
+# Property Value
+
+[String](https://learn.microsoft.com/dotnet/api/system.string)
+
+A [BCP-47 language tag](https://tools.ietf.org/html/bcp47). The app can set it to override the
+language. It must be a single language tag; a delimited list of language tags will result in default
```
FrankC-msft, line 122: Does "default language" mean "it takes the first language in your list", "the property keeps the value it had before you tried to set it to a list", or "the property gets reset to empty by this attempt"?

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
```diff
@@ -0,0 +1,190 @@
...
+
+```c#
+public static string PrimaryLanguageOverride { get; set; }
+```
+
+# Property Value
+
+[String](https://learn.microsoft.com/dotnet/api/system.string)
+
+A [BCP-47 language tag](https://tools.ietf.org/html/bcp47). The app can set it to override the
```
FrankC-msft, line 121: If I set a string that can't be a well-formed language tag, does that throw a usage error or is it ignored?

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
```diff
@@ -0,0 +1,190 @@
...
+```
+
+# API Details
+
+```c# (but really MIDL3)
+namespace Microsoft.Windows.ApplicationModel.Resources
+{
+  // ...
+  [contract(MrtCoreContract, 2)]
+  runtimeclass ApplicationLanguages
```
DrusTheAxe, line 183: Do we routinely use plural names in WinRT? Should be this `ApplicationLanguage`?

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
```diff
@@ -0,0 +1,190 @@
...
+  {
+      static IVectorView<String> Languages { get; };
```
DrusTheAxe, line 185: Why not `IVector<String>`?

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
```diff
@@ -0,0 +1,190 @@
...
+      static IVectorView<String> ManifestLanguages { get; };
```
DrusTheAxe, line 186: `IVector<String>`?

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
```diff
@@ -0,0 +1,190 @@
...
+    // ...
+    public App()
+    {
+        // Set the primary language override to German (Germany)
+        // This will cause the app to use the German (Germany) resources
+        // Both resources loaded manually with ResourceLoader and XAML resources fetched with x:Uid
+        // will be German (Germany)
+        Microsoft.Windows.ApplicationModel.Resources.ApplicationLanguages.PrimaryLanguageOverride = "de-DE";
+
+        Microsoft.Windows.ApplicationModel.Resources.ResourceLoader resourceLoader = new Microsoft.Windows.ApplicationModel.Resources.ResourceLoader();
```
DrusTheAxe, line 167: NIT: `var resourceLoader = ...`

No need to repeat the type given it's the ctor on the right side of the `=`

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
DrusTheAxe: Or is it an error when setting to an unsupported language?

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
```diff
@@ -0,0 +1,190 @@
...
+
+# Remarks
+
+Apps normally run with language settings determined by the system by comparing the languages
+supported by the app with the language preferences of the user. The PrimaryLanguageOverride property
+is used to override that behavior by setting a single specific language.
+
+For packaged apps, `PrimaryLanguageOverride` setting is a wrapper for
+[Windows.Globalization.ApplicationLanguages.PrimaryLanguageOverride](https://learn.microsoft.com/uwp/api/windows.globalization.applicationlanguages.primarylanguageoverride)
+property and behaves the same. It is persisted between app sessions.
```
DrusTheAxe, line 138: What is an 'app session'?

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
```diff
@@ -0,0 +1,190 @@
...
+
+For un-packaged apps, `PrimaryLanguageOverride` setting is not persisted between sessions. It should
```
DrusTheAxe, line 140: NIT: unpackaged (no hyphen)

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
```diff
@@ -0,0 +1,190 @@
...
+be set each time the app is loaded. You should set PrimaryLanguageOverride setting in early stage of
+app loading, before any resource is loaded.
+
+The property should only be set based on user input presented in settings UI. It can be read at any
```
DrusTheAxe, line 144: 'should'. And what if it isn't?

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
stefansjfw: We align with how Windows.Globalization property behaves:
 - setting an unsupported language is ignored - results in the default language being used. No errors/exceptions.
 - setting a non-language tag would throw an exception saying "The parameter is incorrect."

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
stefansjfw: setting a string that is not  a language tag would throw an exception saying "The parameter is incorrect."

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
stefansjfw: Actually, this is wrong. Setting delimited list of language tags will result in "The parameter is incorrect." as well. This is copy-pasted from W.G. property docs, so the docs are wrong there. This line needs to be fixed.

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
stefansjfw: > Is the "set" ignored, and getting the property again still returns the previous value?

Yes. This is correct.

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
stefansjfw: Single run of an application - i.e. session starts when app is ran, and ends when application is closed. This is copy-pasted from W.G. property docs. It is used there without additional explanation

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
stefansjfw: This line should be moved to "packaged apps" part, after "It is persisted between app sessions" part. It means that it should not be set on every app launch, but only on user input from app settings.

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
stefansjfw: the idea was to mimic existing Windows.Globalization namespace type https://learn.microsoft.com/en-us/uwp/api/windows.globalization.applicationlanguages?view=winrt-22621

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
stefansjfw: To align with Windows.globalization property used in the background . and get IReadOnlyList in C#

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
stefansjfw: To align with Windows.globalization property used in the background . and get IReadOnlyList in C#

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
```diff
@@ -0,0 +1,190 @@
...
+`Microsoft.Windows.ApplicationModel.Resources` MRTCore namespace.
+`Microsoft.Windows.ApplicationModel.Resources.ApplicationLanguages.PrimaryLanguageOverride` property
+is supported both for packaged and un-packaged WindowsAppSDK apps.
+
+This new type (`Microsoft.Windows.ApplicationModel.Resources.ApplicationLanguages`) is a wrapper for
+the OS type with the updated behavior for the `PrimaryLanguageOverride` to support un-packaged apps.
+`Microsoft.Windows.ApplicationModel.Resources.ApplicationLanguages` should serve as a replacement
+for the existing OS type (`Windows.Globalization.ApplicationLanguages.PrimaryLanguageOverride`).
+
+`GetLanguagesForUser(User)` method from `Windows.Globalization.ApplicationLanguages` is being
```
david-risney, line 23: The (User) part is confusing to me. Is this supposed to be (ForUser) like there are two methods one with ForUser at the end?

```suggestion
`GetLanguages(ForUser)` method from `Windows.Globalization.ApplicationLanguages` is being
```

stefansjfw: It's the argument of the method. The method name is GetLanguagesForUser and it takes one argument. User is the type of the argument

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
```diff
@@ -0,0 +1,190 @@
+# ApplicationLanguages
+
+This new type (`Microsoft.Windows.ApplicationModel.Resources.ApplicationLanguages`) is a wrapper for
+the OS `Windows.Globalization.ApplicationLanguages` type with the updated behavior for the
+`PrimaryLanguageOverride` property to be supported for un-packaged apps.
```
MikeHillberg, line 5: Where WinAppSDK has a replacement for a Platform SDK API, we should match the namespace other than prefixing with `Microsoft.`. Moving this from Globalization to Resources sounds like an org alignment

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
```diff
@@ -0,0 +1,190 @@
...
+
+# Background
+
+Changing the application language using
+[PrimaryLanguageOverride](https://learn.microsoft.com/uwp/api/windows.globalization.applicationlanguages.primarylanguageoverride)
+property from `Windows.Globalization.ApplicationLanguages` namespace is not supported for
+WindowsAppSDK un-packaged apps, only for packaged. To support language change for un-packaged
```
MikeHillberg, line 12: Nit, the docs don't hyphenate

```suggestion
WindowsAppSDK un-packaged apps, only for packaged. To support language change for unpackaged
```

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
MikeHillberg: I'm still a bit confused on when we want to use the singleton pattern vs the static pattern.
@jonwis FYI

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
```diff
@@ -0,0 +1,190 @@
...
+
+This new type (`Microsoft.Windows.ApplicationModel.Resources.ApplicationLanguages`) is a wrapper for
+the OS type with the updated behavior for the `PrimaryLanguageOverride` to support un-packaged apps.
+`Microsoft.Windows.ApplicationModel.Resources.ApplicationLanguages` should serve as a replacement
+for the existing OS type (`Windows.Globalization.ApplicationLanguages.PrimaryLanguageOverride`).
+
+`GetLanguagesForUser(User)` method from `Windows.Globalization.ApplicationLanguages` is being
+intentionally left off due to not being relevant for WinAppSDK.
+
+# API Pages
```
MikeHillberg, line 26: This is all copy/paste from existing docs? Or are there differences in here beyond a couple of mentions of unpackaged?

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
```diff
@@ -0,0 +1,190 @@
...
+profile language list). The app runtime language list is set to this intersection (if the
+intersection is not empty), or to just the app's default language (if the intersection is empty).
+For more detail, see the
+[App runtime language list](https://learn.microsoft.com/windows/apps/design/globalizing/manage-language-and-region#app-runtime-language-list)
+section in
+[Understand user profile languages and app manifest languages](https://learn.microsoft.com/windows/apps/design/globalizing/manage-language-and-region).
+
+## ApplicationLanguages.ManifestLanguages property
+
+Gets the app's declared list of supported languages.
```
MikeHillberg, line 81: What if you have no manifest (unpackaged app)?

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
kmahone: RECOMMEND: Update to Microsoft.Windows.Globalization.ApplicationLanguages.

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
kmahone: Fix.

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
kmahone: RECOMENDED: Leave as is. This new type does not have the ...ForUser variant anyway.

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
kmahone: Correct. Other than the comments around packed vs unpackaged or UWP vs Win32.

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
kmahone: RECOMMEND: 
1. Make this return an empty list instead of throwing an exception.
2. Update the doc here to state that it returns an empty list.

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
kmahone: RECOMMEND: Update the doc to state that this api throws an exception if given an invalid string.

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
kmahone: RECOMMEND: Update this doc to state that it throws exception. 

CONSIDER: Update existing public doc.

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
kmahone: RECOMENDED: 

Update docs to be clearer about what "unsupported" language is. 
Update docs to say what happens when an invalid value is set.

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
kmahone: RECOMMEND: Leave as-is.

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
kmahone: Fix.

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
kmahone: Move to "packaged apps" section.

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
kmahone: Fix.

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
kmahone: Keep as-is since it is a copy of the OS api.

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
kmahone: Keep as is to match existing api.

specs/mrtcore/PrimaryLanguageOverride/PrimaryLanguageOverride.md
kmahone: Keep as-is to match existing api.

kmahone (approved):
:shipit:

PR #4484, opened 2024-06-10
Package Management: Is Provisioned API spec

DrusTheAxe:
API spec for IsProvisioned overloads

- IsPackageProvisioned
- IsPackageProvisionedByUri
- IsPackageSetProvisioned

### DIFF
--- a/specs/packagemanager/PackageManagement.md
+++ b/specs/packagemanager/PackageManagement.md
@@ -127,7 +127,7 @@ The following table shows the supported permutations of verbs and targets:
 |Remove                  |   X    |    X     |        WAS        |      OS/WAS     |   X    |    X     | OS/WAS  |    WAS     |
 |Repair                  |   X    |    X     |        WAS        |       WAS       |   X    |    X     |  WAS    |    WAS     |
 |Reset                   |   X    |    X     |        WAS        |       WAS       |   X    |    X     |  WAS    |    WAS     |
-|IsProvisioned           |   X    |    X     |       OS/WAS      |      OS/WAS     |   X    |    X     |  WAS    |    WAS     |
+|IsProvisioned           |   X    |    X     |       OS/WAS      |        X        |   X    |    X     |  WAS    |    WAS     |
 |Provision               |   X    |    X     |       OS/WAS      |        X        |   X    |    X     |  WAS    |    WAS     |
 |Deprovision             |   X    |    X     |       OS/WAS      |        X        |   X    |    X     |  WAS    |    WAS     |
 
@@ -651,13 +651,18 @@ Fabrikam app installing Contoso's Muffin and Waffle packages for all users if ne
 ```c#
 void Install()
 {
+    // We want to check what's provisioned by PackageFamilyName. If something's
+    // not provisioned we'll also need MinVersion and PackageUri to Stage and
+    // Provision. But checking if a family is provisioned only supports some
+    // URI schemes (notably, not the one we need for our staging work). So we'll
+    // define the PackageSet with the families we want checked and if something's
+    // not provisioned we'll add the additional properties needed.
+
     var packageSet = new PackageSet() {
-        Items = { new PackageSetItem() { PackageFamilyName = "contoso.muffin_1234567890abc",
-                                         MinVersion = ToVersion(1, 2, 3, 4),
-                                         PackageUri = new Uri("c:\\contoso\\muffin-1.2.3.4.msix") },
-                { new PackageSetItem() { PackageFamilyName = "contoso.waffle_1234567890abc",
-                                         MinVersion = ToVersion(2, 4, 6, 8),
-                                         PackageUri = new Uri("https://contoso.com/waffle-2.4.6.8.msix") } };
+        Items = { new PackageSetItem() { PackageFamilyName = "contoso.muffin_1234567890abc" },
+                { new PackageSetItem() { PackageFamilyName = "contoso.waffle_1234567890abc" }
+        }
+    };
 
     var packageDeploymentManager = PackageDeploymentManager.GetDefault();
     if (packageDeploymentManager.IsPackageSetProvisioned(packageSet))
@@ -670,6 +675,11 @@ void Install()
     {
         return;
     }
+
+    packageSet.Items()[0].MinVersion(ToVersion(1, 2, 3, 4));
+    packageSet.Items()[0].PackageUri(new Uri("c:\\contoso\\muffin-1.2.3.4.msix"));
+    packageSet.Items()[1].MinVersion(ToVersion(2, 4, 6, 8));
+    packageSet.Items()[1].PackageUri(new Uri("https://contoso.com/waffle-2.4.6.8.msix"));
 
     var stageOptions = new StagePackageOptions();
     var deploymentResult = await packageDeploymentManager.StagePackageSetReadyAsync(packageSet, options);
--- a/specs/packagemanager/PackageManagement.md
+++ b/specs/packagemanager/PackageManagement.md
@@ -336,7 +336,7 @@ package management APIs in Windows (e.g. Windows.Management.Deployment.PackageMa
   requested package is installed.
 * Many `PackageManager` operations accept a target package as a file but require it expressed as a
   `Uri`. `PackageDeploymentManager` provides overrides also accepting it as a `String`.
-* `PackageManager.RemovePackageByFullNameAsyn(p)` fails if the specified package isn't found.
+* `PackageManager.RemovePackageByFullNameAsync(p)` fails if the specified package isn't found.
   `PackageDeploymentManager` succeeds as the requested package is not present at the end of the
   operation.
   * This follows the core deployment principle "'Tis not the journey that matters but the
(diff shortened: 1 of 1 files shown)


specs/packagemanager/PackageManagement.md
```diff
@@ -344,6 +347,12 @@ package management APIs in Windows (e.g. Windows.Management.Deployment.PackageMa
   not PackageFamilyName. `PackageDeploymentManager` provides a richer API accepting additional
```
oldnewthing, line 347: Typo on line 339: `RemovePackageByFullNameAsyn` should be `RemovePackageByFullNameAsyn`.

specs/packagemanager/PackageManagement.md
```diff
@@ -344,6 +347,12 @@ package management APIs in Windows (e.g. Windows.Management.Deployment.PackageMa
...
   identifiers.
 
+## 3.12 Is*Provisioned()
+
+Is\*Provisioned\*() methods determine if the target is provisioned.
+
+These methods require administrative privileges.
```
oldnewthing, line 354: If I pass a package that is not even downloaded (much less provisioned), does it fail with "Invalid package name" or does it return false?

specs/packagemanager/PackageManagement.md
```diff
@@ -635,6 +644,66 @@ void CheckAndFixPackageVolume(string packageStorePath)
...
+    var packageSet = new PackageSet() {
+        Items = { new PackageSetItem() { PackageFamilyName = "contoso.muffin_1234567890abc",
+                                         MinVersion = ToVersion(1, 2, 3, 4),
+                                         PackageUri = new Uri("c:\\contoso\\muffin-1.2.3.4.msix") },
+                { new PackageSetItem() { PackageFamilyName = "contoso.waffle_1234567890abc",
+                                         MinVersion = ToVersion(2, 4, 6, 8),
+                                         PackageUri = new Uri("https://contoso.com/waffle-2.4.6.8.msix") } };
+
+    var packageDeploymentManager = PackageDeploymentManager.GetDefault();
+    if (!packageDeploymentManager.IsPackageSetProvisioned(packageSet))
```
oldnewthing, line 663: Does this check only that the specified package families have been provisioned (at any version), or does it also check the version numbers of each package in the set?

If the currently provisioned version is *newer* than the one requested, does Is\*Provisioned return "true" (it's even better than you dreamed!) or "false" (the exact version you are asking about is not provisioned)?

chrilaMSFT: Related is there an exact version check available sometimes packages break compatibility between versions

chrilaMSFT, line 663: >IsPackageSetProvisioned [](http://example.com/codeflow?start=34&length=23)

why use a set?  Would it be better for caller to call in a loop then they can figure out which package is provisioned and which is not.

specs/packagemanager/PackageManagement.md
```diff
@@ -635,6 +644,66 @@ void CheckAndFixPackageVolume(string packageStorePath)
...
+    {
+        bool ok = PromptUserForConfirmation();
```
chrilaMSFT, line 665: >PromptUserForConfirmation() [](http://example.com/codeflow?start=18&length=27)

how does user know which packages will install?

specs/packagemanager/PackageManagement.md
```diff
@@ -635,6 +644,66 @@ void CheckAndFixPackageVolume(string packageStorePath)
...
+        if (!ok)
+        {
+            return;
+        }
+    }
```
chrilaMSFT, line 670: >} [](http://example.com/codeflow?start=4&length=1)

Do we need to do anything if packageDeploymentManager.IsPackageSetProvisioned(packageSet) returns true I thought that means that there was no need to provision the package.

specs/packagemanager/PackageManagement.md
DrusTheAxe: False

Whether absent or present but not provisioned it's still "not provisioned". If you asked is Raymond in the kitchen it doesn't matter if Raymon is in another room in the house or not in the house at all, either way he's not in the kitchen :-)

If you pass in invalid data (e.g. something that's not a syntactically valid pkgfamilyname, pkgfullname or URI) you get E_INVALIDARG.

specs/packagemanager/PackageManagement.md
DrusTheAxe: Depends on your scenario. One I've seen is needing N packages provisioned and if 1+ aren't a simple user confirmation "I have stuff to do OK?" and if OK then do it, where details like package names aren't needed, desired or meaningful to the user.

Also, `PackagDeploymentManager` strives to offer consistent overloads across the various verbs accepting a package identity, a URI and a packageset. One of the reasons for PackageDeploymentManager's creation is PackageManager's consistently inconsistent API surface with corresponding usability pain points.

I don't see this is one of the more popular overloads across the API surface, but still non-zero.

specs/packagemanager/PackageManagement.md
DrusTheAxe: You assume the user needs to know, or would understand.

I know at least one prominent and recent similar user experience where the user is asked to confirm some package work at the functional level without sharing the implementation details like package names. And to be honest, even I would have a hard time understanding what the specific packages mean, other than the half dozen packages collectively are needed to power a feature.

For non-interactive post-mortem reporting or analysis deployment generates logs of all its activity so telling a user "build car?" rather than "assemble engine, doors, spark plugs, timing belt, ..." can be a valid (and sometimes better) user experience.

specs/packagemanager/PackageManagement.md
DrusTheAxe: Doh! Copy/paste error. Yes, we should provision

RECOMMEND: Fix the sample

specs/packagemanager/PackageManagement.md
DrusTheAxe: Lolz! The typo fix has a typo :P

RECOMMEND: Fix the typo - Async, not Asyn

specs/packagemanager/PackageManagement.md
DrusTheAxe: >chrilaMSFT is there an exact version check available sometimes packages break compatibility between versions

Provisioning is based on Package Family so 'exact version' is meaningless.

If you're interested in what's staged on the machine you can do an exact match via `pkgmgr.FindPackage(pkgfullname)` and if you get Package object it's found. I think you get a specific error condition if not found (`HRESULT_FROM_WIN32(ERROR_NOT_FOUND)`?) rather than the usual `null` as PackageManager is one of the oldest WinRT APIs (early Win8 days) and doesn't follow a lot of common practices by today's standards. That's one of the reasons for `PackageDeploymentManager`, as a "If you created PackageManager today...".

Similarly you can explicitly check if a package is registered for a user via `pkgmgr.FindPackageForUser("", pkgfullname)`, with similar caveats about detecting not-found.

You can also do these via non-WinRT APIs e.g. [GetStagedPackageOrigin()](https://learn.microsoft.com/windows/win32/api/appmodel/nf-appmodel-getstagedpackageorigin), [FindPackagesByPackageFamily()](https://learn.microsoft.com/windows/win32/api/appmodel/nf-appmodel-findpackagesbypackagefamily) and other APIs (usually in appmodel.h).

`PackageDeploymentManager` provides a superset of most things in `PackageManager` but notably no `FindPackage*()` methods. Not yet. Doing that right is more involved than it sounds and improved 'query' APIs weren't as urgent a need as 'do' methods e.g. `EnsurePackageSetIsReady()`. I expect we'll get there one day, but more urgent matters first.

DrusTheAxe: > Does this check only that the specified package families have been provisioned (at any version), or does it also check the version numbers of each package in the set?

The former. Provisioning is based on Package Family. You don't provision Foo-v1, you provision Foo. If you provision Foo-v1 via `DISM` or the powershell `Add-AppxProvisionedPackage` then you're really staging Foo-v1 and provisioning Foo.

The sample implies it but shouldn't.

The table in 3.1. API Structure erroneously states IsProvisioned supports PackageFullName. 

RECOMMEND: Improve the sample. Also fix the table (as noted above).

specs/packagemanager/PackageManagement.md
```diff
@@ -125,6 +127,7 @@ The following table shows the supported permutations of verbs and targets:
 |Remove                  |   X    |    X     |        WAS        |      OS/WAS     |   X    |    X     | OS/WAS  |    WAS     |
 |Repair                  |   X    |    X     |        WAS        |       WAS       |   X    |    X     |  WAS    |    WAS     |
 |Reset                   |   X    |    X     |        WAS        |       WAS       |   X    |    X     |  WAS    |    WAS     |
+|IsProvisioned           |   X    |    X     |       OS/WAS      |      OS/WAS     |   X    |    X     |  WAS    |    WAS     |
```
DrusTheAxe, line 130: Provisioning is based on package family.

RECOMMEND: Change table marking IsProvisioned/PackageFullName=X (not supported)

specs/packagemanager/PackageManagement.md
oldnewthing: Particularly misleading since there is a MinVersion that is apparently ignored? Or is that what you are referring to by "the sample implies it but shouldn't"?

But even so, the fact that there's a PackageSet variant, and PackageSetItems have a MinVersion, the existence of a MinVersion is unavoidable.

chrilaMSFT: I am confused there is a package version and url for where to install the package but we don't check the version so url is not used if any package in the family is installed.  We have apis that can detect the version individually and can do the same logic.  Why do we need the new API.  Why not just write a helper function that is open source.

specs/packagemanager/PackageManagement.md
chrilaMSFT: If user is already installing the app why do we need another prompt to install dependencies?

specs/packagemanager/PackageManagement.md
chrilaMSFT: still missing?

chrilaMSFT (commented):
Looks like the else statement is still missing

chrilaMSFT (approved):

rhuang-msft (approved):

PR #4516, opened 2024-06-27
Setup Validation Stage for Installer in the Foundation PR build

kythant:
This new stage will download the WindowsAppSDK package from the latest Aggregator Nightly build off main.
It will then proceed to build the installer using downloaded WindowsAppSDK package. 

WindowsAppSDK\build\AzurePipelinesTemplates\WindowsAppSDK-BuildInstaller-Steps.yml 
is a stripped-down version of 
WindowsAppSDKAggregator/build/AzurePipelinesTemplates/WindowsAppSDK-CreateInstaller-Steps.yml

I've also copied over 
https://microsoft.visualstudio.com/ProjectReunion/_git/WindowsAppSDKAggregator?path=/build/Scripts/CreateInstallerOverrideHeader.ps1
and 
https://microsoft.visualstudio.com/ProjectReunion/_git/WindowsAppSDKAggregator?path=/build/Scripts/ExtractMSIXFromNuget.ps1
into this PR

The plan for the future is to have both builds use the same template which would work for both OneBranch and non-OneBranch builds. The WindowsAppSDKAggregator will have its version deleted and would use the template from this repository. 

For now, in the interest of time, we will keep these two templates at the risk of them diverging and the template only works for non-OneBranch builds. 

There is a concern that I have with the circular nature of this. What if the installer changes require changes on the WindowsAppSDK NuGet as well? This means that a previous build of the WindowsAppSDK NuGet may not work and as a result, the validation will fail and we will need a way to override this. 

Test run with broken change to show that it will failed when a broken change is introduced
https://microsoft.visualstudio.com/DefaultCollection/ProjectReunion/_build/results?buildId=102273256&view=results


build/WindowsAppSDK-BuildInstaller-PR.yml
```diff
@@ -0,0 +1,11 @@
+# This file has been adapted to use as a non-OneBranch PR validation yml in the WindowsAppSDK Foundation repository
```
kythant, line 1: I will remove this file

build/AzurePipelinesTemplates/WindowsAppSDK-BuildInstaller-Stage.yml
```diff
@@ -0,0 +1,29 @@
...
+      maxParallel: 10
+      matrix:
+        Release_x86:
+          buildPlatform: 'x86'
+          buildConfiguration: 'Release'
+        Release_x64:
+          buildPlatform: 'x64'
+          buildConfiguration: 'Release'
+        Release_arm64:
+          buildPlatform: 'ARM64'
```
DrusTheAxe, line 22: NIT: Why ARM64 and not arm64 like the other architectures?

DrusTheAxe (approved):

build/AzurePipelinesTemplates/WindowsAppSDK-BuildInstaller-Stage.yml
kythant: Good point. I will change and see if everything still builds

PR #4524, opened 2024-06-28
Update TouchdownBuildTask to Version 4

kythant:
Update TouchdownBuildTask to Version 4 which eliminates the use of secrets by moving to using federated credentails

Switched our TDBuild registration to use an existing ProjectReunion App registration that already has a federated credentials setup


kythant:
Test run
https://microsoft.visualstudio.com/ProjectReunion/_build/results?buildId=102371224&view=results

Scottj1s (approved):

PR #4539, opened 2024-07-05
Enable nullable reference types on new projects (.NET guidance) and fix warning

Scottj1s:
[Nullable reference types](https://learn.microsoft.com/en-us/dotnet/csharp/nullable-references) states:
"Beginning with .NET 6, new projects include the <Nullable>enable</Nullable> element in all project templates."

Following suit with WinAppSDK C# templates, and addressing warning on m_window.


jevansaks (approved):
Thanks!

PR #4545, opened 2024-07-08
Allow Installer validation stage to download from a build that has failed

kythant:
A nightly could fail in the test stages but succeeds in the Aggregator stage. 

This change allows it so the Foundation build can still download the WindowsAppSDK Nuget package even if the build partially fail.

### DIFF
--- a/build/AzurePipelinesTemplates/WindowsAppSDK-BuildInstaller-Steps.yml
+++ b/build/AzurePipelinesTemplates/WindowsAppSDK-BuildInstaller-Steps.yml
@@ -18,7 +18,7 @@ parameters:
 steps:
 - ${{ if eq(parameters.UseCurrentBuild, 'True') }}:
   - task: DownloadPipelineArtifact@2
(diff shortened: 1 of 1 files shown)


build/AzurePipelinesTemplates/WindowsAppSDK-BuildInstaller-Steps.yml
```diff
@@ -27,12 +27,14 @@ steps:
   # This step downloads the WindowsAppSDK NuGet package from last successful build of the Aggregator's Nightly build
```
alexlamtest, line 27: Nit: comment needs update to include failed builds?

alexlamtest (approved):

build/AzurePipelinesTemplates/WindowsAppSDK-BuildInstaller-Steps.yml
kythant: oh good call out

Scottj1s (approved):

Issue #1034, opened 2021-07-07
WinUi 3 desktop app doesn't open camera. and is there any way that can open default system camera and capture image


sangitamalviya:
CameraCaptureUI captureUI = new CameraCaptureUI();

            StorageFile photo = await captureUI.CaptureFileAsync(CameraCaptureUIMode.Photo);

            if (photo != null)
            {
                CameraCapture.Source = new BitmapImage(new Uri(photo.Path));
            }

> CameraCaptureUI  it doesn't open camera. even  CaptureElement  does not find on UI 
![Capture](https://user-images.githubusercontent.com/53295795/124348234-65c11800-dc06-11eb-81fc-c203ca965788.PNG)

Jaiganeshkumaran:
I think CameraCaptureUI requires you to invoke IInitializeWithWindow and pass the window handle.

sangitamalviya:
hi @Jaiganeshkumaran ,[]( https://github.com/microsoft/microsoft-ui-xaml/issues/5354 )here FileOpenPicker  IInitializeWithWindow . I was same tried  for CameraCaptureUI IInitializeWithWindow,but it's giving Exception "Element Not Found".I do not know CameraCaptureUI have different Guid or not.

jonwis:
Looks like the `CameraCaptureUI` does not support `IInitializeWithWindow`, and expects the caller to be on a `CoreWindow` thread.  The implementation uses multiple pickers and launchers that need the "with window" treatment. It's an interesting candidate for open-sourcing the implementation for improvement.

craigomatic:
Is there any workaround in the meantime?

applefanbois:
Just do exactly like Apple and everything will be fine. I know this new Windows SDK is 15 years behind the SDK of Android and Apple, so I won't be too mean. Our app cannot release until we have access to the camera, we need to be able to scan documents with a camera.

Jaiganeshkumaran:
> Looks like the `CameraCaptureUI` does not support `IInitializeWithWindow`, and expects the caller to be on a `CoreWindow` thread. The implementation uses multiple pickers and launchers that need the "with window" treatment. It's an interesting candidate for open-sourcing the implementation for improvement.

CameraCaptureUI is probably implemented as a wrapper on Launcher::LaunchUriForResultsAsync as it opens the Windows Camera app in a modal dialog, but Launcher::LaunchUriForResultsAsync doesn't work in desktop apps.

[...]

Issue #219, opened 2020-10-01
AppContainer for Win32 apps


sylveon:
# Proposal: AppContainer for Win32 apps

## Summary
This proposal attempts to bring the advantages of app containers (limited permissions and file system access) to Win32, by documenting how a packaged Win32 app can be hosted in one, and allowing Win32-only APIs which are not currently allowed to be called from an AppContainer.

Currently, this can be achieved by using `EntryPoint="Windows.PartialTrustApplication"` in the appx manifest. By setting this, your Win32 app is launched in the AppContainer:

![image](https://user-images.githubusercontent.com/6440374/94751632-24d66300-0357-11eb-946b-b3e361cac99b.png)

But, this entry point is largely undocumented (its only reference on the entire internet is an [electron PR](https://github.com/electron/electron/pull/15340)), and some things are broken.

For example, tray icons (with `ShellNotifyIcon`) are simply not working, and return an access denied error.

`OpenProcess`, even with only the `PROCESS_QUERY_LIMITED_INFORMATION` permission, also returns an access denied. This permission is mostly harmless and only allows a very limited subset of APIs, none of which can do anything harmful to the process. There is little to zero security implication in allowing this.

As the screenshot above shows, I've tried doing this with my own app, and these are the two big limitations I've encountered (as well as the general lack of documentation around this topic).

## Rationale
* This makes Win32 apps safer for everyone. Long gone are programming mistakes or bugs that would accidentally wipe your whole drive.
* Allows to host Win32 apps in Windows 10X natively, without the need for virtualization. Since Win32 apps in AppContainers are under the same restriction than UWP apps, there is no need to use an expensive virtualization stack to run them, but instead a thin compatibility layer (if any). By doing this, it makes Win32 apps faster and more energy efficient under 10X.
* Allows developers to ship Win32 apps to the Microsoft Store without requesting access to the `runFullTrust` restricted capability
   * This would allow to make sandboxing more fine-grained rather than all or nothing, for example a specific restricted capability to enable more process access permissions under the sandbox.
   * It can also bring the possibility to lockdown `runFullTrust` to apps that actually do need to run completely outside the container for legitimate reasons, rather than every Win32 app. 

## Scope

| Capability | Priority |
| :---------- | :------- |
| This proposal will allow developers to host Win32 apps in the AppContainer with ease | Must |
| This proposal will allow developers to use tray icons from within an AppContainer | Must |
[...]

PR #4576, opened 2024-07-17
CameraCaptureUI API Spec

Saharsh979:
This is API Spec for Issue: #1034 

The spec details the implementation of CameraCaptureUI tailored for desktop applications, enabling seamless integration of native camera capture functionalities across Windows platforms supported by WinAppSDK.

The current WinRT CameraCaptureUI, designed primarily for UWP, presents challenges in desktop environments due to dependencies on CoreWindow and lack of IInitializeWithWindow support. Developers resort to complex interop solutions for integration. In response, the CameraCaptureUI API for WinAppSDK provides a streamlined solution with feature parity, now supporting windowID in the constructor for enhanced desktop compatibility.


mominshaikhdevs:
I hope this PR doesn't get stuck and gets merged asap.

specs/CameraCaptureUI/CameraCaptureUI.md
```diff
@@ -0,0 +1,315 @@
...
+| Name | Description | Parameters | Returns |
+|-|-|-|-|
+| CaptureFileAsync(CameraCaptureUIMode) | Launches the CameraCaptureUI user interface.| __mode__ CameraCaptureUIMode | __IAsyncOperation< StorageFile >__ |
+
+### 3.1.4 Remarks
+CameraCaptureUI provides a full window UI experience for capturing audio, video, and images. It provides controls for setting a time delay on photo captures, trimming video, and for adjusting the camera's settings such as video resolution, the audio device, brightness, and contrast.
+
+Call CaptureFileAsync to launch the UI. The user has control over when to start the capture. When the asynchronous CaptureFileAsync operation completes, a StorageFile object is returned. 
+
+`` Note:
```
MikeHillberg, line 66: Is this still true? Or is this a leftover UWP-ism?

specs/CameraCaptureUI/CameraCaptureUI.md
```diff
@@ -0,0 +1,315 @@
...
+
+StorageFile file = await dialog.CaptureFileAsync(CameraCaptureUIMode.Photo);
+```
+Screenshot of the UI:
+
+
+![Screenshot](CameraCaptureUI_Snapshot.png)
+
+
+### 3.1.1 Constructor (The Change Introduced)
```
MikeHillberg, line 41: Is the constructor change and the new namespace the only changes compared to the existing UWP version?

specs/CameraCaptureUI/CameraCaptureUI.md
```diff
@@ -0,0 +1,315 @@
...
+
+
+## 3.1 CameraCaptureUI
+
+Provides a full window UI for capturing audio, video, and photos from a camera. As well as controls for trimming video, time delayed capture, and camera settings.
+
+This example shows how to use the CameraCaptureUI Class to take a picture. The dialog with the user occurs when you call the asynchronous method CaptureFileAsync.
+```c#
+// Using Microsoft.Windows.Media.Capture.CameraCaptureUI API to capture a photo
+CameraCaptureUI dialog = new CameraCaptureUI(windowId);
```
MikeHillberg, line 29: It would help to show this in the context of a WinUI3 app, specifically showing how to get the WindowId

specs/CameraCaptureUI/CameraCaptureUI.md
Saharsh979: Yes, in the new WinAppSDK implementation, the major changes include modifying the constructor and namespace. Additionally, there's a transition from explicit getter and setter methods in the OS version to using modern C++/WinRT's basic_property template. This template provides implicit getter and setter functionality through overloaded operators, and it adequately covers both getter and setter functionalities for all properties, ensuring parity with the methods.

specs/CameraCaptureUI/CameraCaptureUI.md
Saharsh979: The camera app is currently a UWP application, and there's a team working on migrating it to the Windows App SDK. (not sure about the timelines). From what I understand, even in the Windows App SDK version of the API, the camera app functions as a full-trust application with built-in access to the camera. It seems unnecessary to explicitly specify this access in the manifest.

As for the note about potential issues with the Windows Application Certification Kit related to specifying access, we'll need to test this to see if it's still **relevant today.**

Also, since the camera app migration is a work in progress at the moment, maybe we should hold off on adding specific notes about this issue for now? Thoughts? @AjitSurana

specs/CameraCaptureUI/CameraCaptureUI.md
```diff
@@ -0,0 +1,382 @@
...
+# Conceptual pages
+
+## Capture a photo (C++)
+
+A C++ example that tests whether a photo was successfully taken using the CaptureFileAsync method.
+
+```c++
+    // Initialize CameraCaptureUI with a window handle (Foreground Window in this example)
+    auto parentWindow = ::GetForegroundWindow();
+    winrt::Microsoft::UI::WindowId windowId{ reinterpret_cast<uint64_t>(parentWindow) };
```
david-risney, line 50: Is this the right way to convert HWND to WindowId? Should this be using [`winrt::Microsoft::UI::WindowId GetWindowIdFromWindow(const HWND & hwnd);`](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/win32/winrt-microsoft.ui.interop/nf-winrt-microsoft-ui-interop-getwindowidfromwindow)

specs/CameraCaptureUI/CameraCaptureUI.md
```diff
@@ -0,0 +1,382 @@
...
+    winrt::Microsoft::Windows::Media::Capture::CameraCaptureUI cameraUI(windowId);
+
+    // Configure Photo Settings
+    cameraUI.PhotoSettings().Format(CameraCaptureUIPhotoFormat::Jpeg);
+    cameraUI.PhotoSettings().AllowCropping(false);
+
+    // Capture a photo asynchronously
+    auto photoOperation = cameraUI.CaptureFileAsync(CameraCaptureUIMode::Photo);
+    auto photo = photoOperation.get(); // Get the result synchronously
```
david-risney, line 59: Usually the C++ & C# code samples should match. In this case C# handles this async and C++ does this sync. Should they both be sync or both be async?

specs/CameraCaptureUI/CameraCaptureUI.md
Saharsh979: To maintain consistency with C# sample, updated the approach of fetching the WindowId according to the official WinUI docs. 
https://learn.microsoft.com/en-us/windows/apps/develop/ui-input/retrieve-hwnd#winui-3-with-c-1

specs/CameraCaptureUI/CameraCaptureUI.md
Saharsh979: Agreed, updated the C++ sample to call async. Thanks!

specs/CameraCaptureUI/CameraCaptureUI.md
```diff
@@ -0,0 +1,387 @@
...
+    // Assert the result
+    if (photo != nullptr)
+    {
+        // Log success if photo capture was successful
+        Log::Comment(L"Photo capture was successful.");
+    }
+    else
+    {
+        // Log error if photo capture failed or was canceled
+        Log::Error(L"Photo capture failed or was canceled.");
```
dmachaj, line 75: If it failed then the IAsyncOperation should have thrown.  Cancel should have thrown a canceled exception.

specs/CameraCaptureUI/CameraCaptureUI.md
```diff
@@ -0,0 +1,387 @@
...
+
+
+## CameraCaptureUIMode enum
+
+Determines whether the user interface for capturing from the attached camera allows capture of photos, 
+videos, or both photos and videos.
+
+| Name | Value | Description |
+|-|-|-|
+| PhotoOrVideo | 0 | The user interface allows capture of photos and videos. |
```
dmachaj, line 240: Which value is the default if the developer doesn't explicitly specify one?  The same question applies to other fields such as PhotoFormat and VideoFormat.

specs/CameraCaptureUI/CameraCaptureUI.md
```diff
@@ -0,0 +1,387 @@
...
+
+## CameraCaptureUIMaxPhotoResolution enum
+Determines the highest resolution the user can select for capturing photos.
+
+| Name | Value | Description |
+|-|-|-|
+| HighestAvailable | 0 | The user can select any resolution. |
+| VerySmallQvga | 1 |The user can select resolutions up to 320 X 240, or a similar 16:9 resolution. |
+| SmallVga | 2 | The user can select resolutions up to 320 X 240, or a similar 16:9 resolution.|
+| MediumXga | 3 | The user can select resolutions up to 1024 X 768, or a similar 16:9 resolution. |
```
dmachaj, line 271: We probably can't change the names if we want UWP parity, but these names are kindof hard to read and understand :(.

specs/CameraCaptureUI/CameraCaptureUI.md
```diff
@@ -0,0 +1,387 @@
...
+
+## CameraCaptureUIMaxVideoResolution enum
+Determines the highest resolution the user can select for capturing video.
+
+| Name | Value | Description |
+|-|-|-|
+| HighestAvailable | 0 | The user can select any resolution. |
+| LowDefinition | 1 | The user can select resolutions up to low definition resolutions.|
+| StandardDefinition | 2 | The user can select resolutions up to standard definition resolutions.|
+| HighDefinition | 3 | The user can select resolutions up to high definition resolutions. |
```
dmachaj, line 284: Should these be defined in more specifics?  I assume SD means "480p" and HD is "720p" or possibly higher.  But that's a guess.

specs/CameraCaptureUI/CameraCaptureUI.md
```diff
@@ -0,0 +1,387 @@
...
+        HighestAvailable = 0,
+        VerySmallQvga,
+        SmallVga,
+        MediumXga,
+        Large3M,
+        VeryLarge5M,
+    } CameraCaptureUIMaxPhotoResolution;
+ 
+    [contract(CameraCaptureUIContract, 1)]
+    [default_interface]
```
dmachaj, line 349: `[default_interface]` is only needed if the runtimeclass has no methods.

specs/CameraCaptureUI/CameraCaptureUI.md
```diff
@@ -0,0 +1,387 @@
...
+ 
+        Boolean AllowTrimming;
+        Microsoft.Windows.Media.Capture.CameraCaptureUIVideoFormat Format;
+        Single MaxDurationInSeconds;
+        Microsoft.Windows.Media.Capture.CameraCaptureUIMaxVideoResolution MaxResolution;
+    }
+ 
+    [contract(CameraCaptureUIContract, 1)]
+    [default_interface]
+    [threading(sta),
```
dmachaj, line 375: So this threading model means that this API must be called from a UI thread?  What happens if an app has multiple islands with multiple UI threads?  Is that something they need to worry about or is the implementation threadsafe?

specs/CameraCaptureUI/CameraCaptureUI.md
MikeHillberg: Generally agree, but actually none of these have defaults. This one (`CameraCaptureUIMode`) is a required parameter to `CaptureFileAsync`. The others are properties on Settings classes you get from `CameraCaptureUI.PhotoSettings` and `VideoSettings`, and so the initial value is a feature of those APIs.

specs/CameraCaptureUI/CameraCaptureUI.md
MikeHillberg: That would be a fun and lengthy conversation but agreed we shouldn't create the incompatibility.

specs/CameraCaptureUI/CameraCaptureUI.md
MikeHillberg: Note that multi-UI-threaded is an existing scenario with the system API.

specs/CameraCaptureUI/CameraCaptureUI.md
Saharsh979: Does the cancel case also return null? 
For UWP apps using the UWP CameraCaptureUI API, it doesnt throw an exception and returns file as null in such cases.image

However, for WinUI apps using the WinAppSDK version of the CameraCaptureUI API, it returns null, along with throwing an exception E_INVALIDARG as well. Is this behavior okay as this behavior differs from UWP version.
image

Saharsh979: @dmachaj **"If it failed then the IAsyncOperation should have thrown. Cancel should have thrown a canceled exception."**

For the WinAppSDK version of the CameraCaptureUI API, it returns null and throws an E_INVALIDARG exception, which seems like to align with the behavior you mentioned and also sounds like a bug in my code, ideally my implementation should be handling this correct?

However, the key question is whether we should throw an exception for a cancel operation. In UWP, cancellation is handled by returning null, which is a straightforward way to manage this scenario.

specs/CameraCaptureUI/CameraCaptureUI.md
```diff
@@ -0,0 +1,382 @@
...
+ The code is from file
+ [CaptureVideo.xaml.cs](https://github.com/microsoftarchive/msdn-code-gallery-microsoft/tree/master/Official%20Windows%20Platform%20Sample/Windows%208%20app%20samples/%5BC%23%5D-Windows%208%20app%20samples/C%23/Windows%208%20app%20samples/CameraCaptureUI%20Sample%20(Windows%208)/C%23) of the Camera capture UI C# sample with a small modification to include the window ID in the CameraCaptureUI constructor.
+
+ ```c#
+// Retrieve the window handle (HWND) of the current WinUI 3 window.
+ var hWnd = WinRT.Interop.WindowNative.GetWindowHandle(this);
+
+// Get the WindowId for the window
+ Microsoft.UI.WindowId windowId;
+ windowId = Win32Interop.GetWindowIdFromWindow(hWnd);
```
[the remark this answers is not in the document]
Saharsh979: Updated the code to use AppWindow object to fetch the window.

specs/CameraCaptureUI/CameraCaptureUI.md
Saharsh979: So, the CameraCapture API will function even if you don't specify these properties. It only requires that you specify the mode, such as Photo, Video, or PhotoOrVideo. I agree with Mike's point.

specs/CameraCaptureUI/CameraCaptureUI.md
Saharsh979: I agree, @dmachaj. The terms VGA and XGA are unclear and don't provide much insight into their values.

Keeping it the same has the advantage of minimizing the effort required for developers who are using the CameraCapture API, transition their app from UWP to WinAppSDK.

specs/CameraCaptureUI/CameraCaptureUI.md
dmachaj: cppwinrt has a specific `winrt::hresult_canceled` exception type specifically for this case.  See https://github.com/microsoft/cppwinrt/blob/0deecf1ea4fcc6acd10602e2340e243fcdcf4f47/strings/base_error.h#L329.  That corresponds to `HRESULT_FROM_WIN32(ERROR_CANCELLED)`.

specs/CameraCaptureUI/CameraCaptureUI.md
dmachaj: What are the default values on those sub-classes, though?  If I fail to specify properties do I get the highest possible quality or the lowest possible quality?  Or something in the middle?

specs/CameraCaptureUI/CameraCaptureUI.md
dmachaj: That said, exceptions are supposed to be for exceptional circumstances.  The user deciding not to take a photo probably shouldn't count as exceptional.  As a result, S_OK+nullptr is probably the right approach.  (Other errors should still throw).

specs/CameraCaptureUI/CameraCaptureUI.md
MikeHillberg: I'm arguing that it would be inappropriate to document that in this case. The values you get from `CameraCaptureUI.PhotoSettings` is a feature of that API. Maybe the value you get back is a hard-coded value, or maybe it's something that the user sets in the Camera app settings.

specs/CameraCaptureUI/CameraCaptureUI.md
MikeHillberg: Agreed: user canceling is expected and should be S_OK+nullptr. An error case like the Camera app crashing probably should too.

The guideline is that apps using this API shouldn't be expected or recommended to have a try/catch; if an exception is thrown to the app it's a sign of an app bug.

specs/CameraCaptureUI/CameraCaptureUI.md
Saharsh979: I observed in the OS CameraCaptureUI Api, particularly in the CaptureFileAsync function, there is a locking mechanism implemented to only allow one operation at a time. 

I figure this is something I would need to implement for my API as well.

specs/CameraCaptureUI/CameraCaptureUI.md
dmachaj: I can buy that.  Perhaps it would also vary based on hardware (e.g. devices with more RAM default to higher resolution or something like that).

specs/CameraCaptureUI/CameraCaptureUI.md
Saharsh979: I agree the values do seem vague, but it is how the historic documentation defines the values. We can talk with the Camera team owners if they would like to take up this suggestion.

specs/CameraCaptureUI/CameraCaptureUI.md
```diff
@@ -0,0 +1,382 @@
...
+
+> Note: 
+You should not specify the webcam or microphone capabilities in your app manifest file if you are using CameraCaptureUI. 
+If you do so, your app will be displayed in the device's camera privacy settings, 
+but even if the user denies camera access to your app, 
+it will not prevent the CameraCaptureUI from capturing media. 
+This is because the Windows built-in camera app is a trusted first-party app that requires 
+the user to initiate photo, audio, and video capture with a button press. 
+Your app may fail Windows Application Certification Kit certification when submitted to 
+the Store if you specify the webcam or microphone capabilities when using CameraCaptureUI.
```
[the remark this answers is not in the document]
Saharsh979: Agreed, will test if this still holds true for win32/desktop apps. 
Also, we reached out to the current owners for the history, yet to get response from them.

specs/CameraCaptureUI/CameraCaptureUI.md
MikeHillberg: It might have been intentionally vague in the existing API. We've long had trouble deciding between flexibility like this vs having to make up names like xx-small. I don't think we should change the API from existing here, but if we can be more detailed in the doc like we are with MaxPhotoResolution that would be fine.

specs/CameraCaptureUI/CameraCaptureUI.md
AjitSurana: Saharsh and I was discussing about various cases and current API implementation.

**Success cases:**
* User captures a file [S_OK+File+No Exception]
* User presses the cancel button [S_OK+nullptr+No Exception]
**Failure cases:**
* If Camera app is being used already [Exception APP_NOT_AVAILABLE]
* If Camera app URI launch mapping unavailable [Exception PROTOCOL_NOT_AVAILABLE]
* Any other unknown error [Exception GENERAL_FAILURE]

Saharsh would look into more how the UWP API is implemented considering the above success and failure cases. 

In general, what is the recommendation to handle failure cases? is it recommended to throw an exception for failure cases (like mentioned above) or should some other approaches. If other approaches, I think following:
1. The return value of CaptureFileAsync API would be some other interface/class/structure than IAsyncOperation< StorageFile >. The disadvantage is API will be more complicated and non-uniform across UWP & WASDK.
2. Have separate APIs implemented to get the status of the error. This does not look like a clean way.

mominshaikhdevs:
This API should work both -
1. within AppContainer ([AppContainer for Win32 apps - #219](https://github.com/microsoft/windowsappsdk/issues/219) and [win32-app-isolation](https://github.com/microsoft/win32-app-isolation)).
2. and without AppContainer (MediumIL and HighIL).

specs/CameraCaptureUI/CameraCaptureUI.md
Saharsh979: Hey @MikeHillberg @dmachaj 
I've reviewed the behavior of the UWP API across these scenarios to understand how errors are being managed.

Here are the observations:

1) Successful Capture: Returns a storage object with S_OK.
2) Cancelled Capture: Returns null with S_OK. (No exceptions).
3) App Uninstalled: Displays a dialog box (Search for app in store) Yes/No? clicking on No returns null with S_OK.
4) Incorrect URI Mapping: Returns null with S_OK (No exceptions).

It seems the UWP API handles all these cases gracefully without throwing exceptions. Our goal should be to maintain consistency with this existing behavior.

specs/CameraCaptureUI/CameraCaptureUI.md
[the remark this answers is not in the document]
Saharsh979: I have tested this by running the WACK Tool on a UWP app with webcam capability, specified, the App Capabilities test passed, which contradicts this note. Seems like we will be better off without this note.

We are reaching out to the WACK team parallelly to get more insights on this.

specs/CameraCaptureUI/CameraCaptureUI.md
mominshaikhdevs: FWIW, https://www.windowslatest.com/2024/08/21/windows-11-24h2-will-finally-allow-multiple-apps-to-use-your-camera-at-the-same-time/ (since, winui3 apps _will be_ multi-UI-threaded.)

specs/CameraCaptureUI/CameraCaptureUI.md
Saharsh979: The values listed in the documentation were carried over directly from the existing documentation without any changes. We can reach out to the Camera team to discuss if they’d like to consider refining these definitions.

specs/CameraCaptureUI/CameraCaptureUI.md
Saharsh979: The locking mechanism in the UWP CameraCapture API operates at the object level, allowing only one capture operation per instance. I've added a task to track this internally and will implement a similar implementation for the WinAppSDK version of CameraCapture.

@mominshaikhdevs, the update that allows multiple apps to use the camera simultaneously seems to be outside the scope of this API. It doesn't appear that any changes are necessary at the CameraCaptureUI level. The primary function of CaptureFileAsync is to open the camera and return a storage file object. Thanks for bringing this to my attention.

specs/CameraCaptureUI/CameraCaptureUI.md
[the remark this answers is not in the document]
Saharsh979: Removing the note as discussed.

specs/CameraCaptureUI/CameraCaptureUI.md
Saharsh979: The API is designed to be used from a UI thread due to its STA threading model requirement. In a multi-UI-threaded environment, each thread should individually follow the STA model. If the API is called from different UI threads, it should work as long as each call is made from an STA-thread. I have tested the API in a WinUI3 packaged blank app multi-window scenario and it works as expected.

specs/CameraCaptureUI/CameraCaptureUI.md
```diff
@@ -0,0 +1,365 @@
...
+This example shows how to use the CameraCaptureUI Class to take a picture.
+The dialog with the user occurs when you call the asynchronous method CaptureFileAsync.
+
+```c#
+// Using Microsoft.Windows.Media.Capture.CameraCaptureUI API to capture a photo
+CameraCaptureUI dialog = new CameraCaptureUI(windowId);
+Size aspectRatio = new Size(16, 9);
+dialog.PhotoSettings.CroppedAspectRatio = aspectRatio;
+
+StorageFile file = await dialog.CaptureFileAsync(CameraCaptureUIMode.Photo);
```
MikeHillberg, line 134: Still confused here about the result that comes back. Do I need to do a null check? What happens if there's an error, or if the user clicks cancel? 

IIUC from your investigation, there shouldn't be errors, so no need to try/catch, so that's great.

If the user cancels, then this will return null, which the sample should show

```suggestion
StorageFile file = await dialog.CaptureFileAsync(CameraCaptureUIMode.Photo);
if(file != null)
{
    // The user didn't cancel
    this.ProcessFile(file);
}
```

MikeHillberg (approved):

dmachaj (approved):

specs/CameraCaptureUI/CameraCaptureUI.md
Saharsh979: Updated, thanks!

PR #4578, opened 2024-07-17
Reenable Installer Validation using Latest Official WindowsAppSDK builds

kythant:
Reenable Installer Validation where instead of downloading from the latest main branch, we download from the latest official build which is controlled by two pipeline variables.

This should give us more stability in the validation.


JesseCol:
Looks like the comments and Display Name are like "'Download WindowsAppSDK From Latest Nightly'".  Can we please change the comments and names to match the new behavior?  Thanks!

Scottj1s (approved):

JesseCol:
How do these vars get set?
  $(OfficialPipelineID) 
  $(LatestOfficialBuildID)

kythant:
> How do these vars get set? $(OfficialPipelineID) $(LatestOfficialBuildID)

you set these on the pipeline variable settings

PR #4582, opened 2024-07-18
Enable Validation for VSIX; Use Central Feed Management

kythant:
Similar to the Installer stage, the VSIX validation stage download from the latest official build which is controlled by two pipeline variables.

build/AzurePipelinesTemplates/WindowsAppSDK-BuildVSIX-Steps.yml is ported from WindowsAppSDKAggregator/build/AzurePipelinesTemplates/WindowsAppSDK-CreateVSIX-Steps.yml

build/AzurePipelinesTemplates/WindowsAppSDK-SignPackageContents-steps.yml is a direct copy of 
WindowsAppSDKAggregator/build/AzurePipelinesTemplates/WindowsAppSDK-SignPackageContents-steps.yml

Both files will be removed from the WindowsAppSDKAggregator repo in another PR


build/AzurePipelinesTemplates/WindowsAppSDK-SignPackageContents-steps.yml
alexlamtest, line 1: Nonblocking: I presume this file is basically the same as the corresponding one on the Agg repo?
If so, then assuming that it is beneficial to always keep these two in sync, would it be feasible to only keep a single shared copy? Say, but putting this on the config repo, or using other ways to share?

build/AzurePipelinesTemplates/WindowsAppSDK-BuildVSIX-Steps.yml
```diff
@@ -0,0 +1,236 @@
+parameters:
+- name: SignOutput
+  type: boolean
+  default: False
```
alexlamtest, line 4: If this yml file is only being called from build/AzurePipelinesTemplates/WindowsAppSDK-BuildVSIX-Stage.yml, then at least currently, this parameter can only remain False? I ask because if that is expected to remain the case at least in the short term, then we don't really need to port WindowsAppSDK-SignPackageContents-steps.yml over?

build/AzurePipelinesTemplates/WindowsAppSDK-BuildVSIX-Steps.yml
```diff
@@ -0,0 +1,236 @@
...
+- name: IsOfficial
+  type: boolean
+  default: False
```
alexlamtest, line 7: Similar question as above.
And IsOfficial seems to be unused?

build/AzurePipelinesTemplates/WindowsAppSDK-BuildVSIX-Steps.yml
```diff
@@ -0,0 +1,236 @@
...
+            Write-Host "Found " $version
+            Write-Host "##vso[task.setvariable variable=WindowsAppSDKPackageVersion]$version"
+            Write-Host "##vso[task.setvariable variable=WindowsAppSDKPackageVersion;isOutput=true]$version"
+            Exit 0
+          }
+        }
+      }
+      Exit 1
+
+- task: NuGetToolInstaller@1
```
alexlamtest, line 89: Do I understand it right that, from here on, the code is similar to of that in WindowsAppSDK-CreateVSIX-Steps.yml on the agg repo? If so, would it be feasible to refactor this file to only keep one copy of WindowsAppSDK-CreateVSIX-Steps.yml, say, on the config repo, to reduce the recurring burden of keeping 2 copies in sync?

alexlamtest (approved):
I have some suggestions, but no blocking concerns.

build/AzurePipelinesTemplates/WindowsAppSDK-SignPackageContents-steps.yml
kythant: Oh yea I should have made this clear in the description. We are removing it from the Aggregator repo.

build/AzurePipelinesTemplates/WindowsAppSDK-BuildVSIX-Steps.yml
kythant: Oh yea I should have made this clear in the description. We are removing it from the Aggregator repo.

build/AzurePipelinesTemplates/WindowsAppSDK-BuildVSIX-Steps.yml
kythant: It is being used to determine whether to publish artifacts manually

build/AzurePipelinesTemplates/WindowsAppSDK-BuildVSIX-Steps.yml
kythant: Oh yea I should have made this clear in the description. We are removing it from the Aggregator repo.

build/AzurePipelinesTemplates/WindowsAppSDK-BuildVSIX-Steps.yml
kythant: Oh whoops I see where the confusion is. This should be named "IsOneBranch" so that condition line 232 can kick in. Fixing

kythant: Oh wait no, it should be passed down to to the Publish Symbols yml template.

Directory.Packages.props
```diff
@@ -12,5 +12,24 @@
...
     <PackageVersion Include="Microsoft.SourceLink.GitHub" Version="$(MicrosoftSourceLinkGitHubVersion)" />
+    <PackageVersion Include="Microsoft.VisualStudio.SDK" Version="17.3.32804.24" />
+    <PackageVersion Include="Newtonsoft.Json" Version="13.0.2" />
+    <PackageVersion Include="NuGet.VisualStudio" Version="17.6.1" />
+    <PackageVersion Include="Microsoft.VisualStudio.TemplateWizardInterface" Version="17.5.33428.366" />
+    <PackageVersion Include="System.Net.Http" Version="4.3.4" />
+    <PackageVersion Include="System.Text.Json" Version="8.0.4" />
+    <PackageVersion Include="Microsoft.VSSDK.BuildTools" Version="17.4.2118" />
   </ItemGroup>
+  <ItemGroup Condition=" '$(WindowsAppSDKVersion)' != '' ">
```
Scottj1s, line 23: nit: I'd just put the condition on each PackageVersion so we have one ItemGroup

kythant: oh yea

kythant:
> ManagePackageVersionsCentrally

kythant:
> Can we take a run at fixing Microsoft.FrameworkUdk.PackageReference.csproj and MrtCoreManagedTest.csproj too here? (removing ManagePackageVersionsCentrally override)

I can take a look now but I'll do so in another PR

PR #4604, opened 2024-07-30
[MRTCore] Sync between WGA.PrimaryLanguageOverride and MWGA.PrimaryLanguageOverride

stefansjfw:
Sync between old and new API. If old one (Windows.Globalization.ApplicationLanguages.PrimaryLanguageOverride) was set more recently than the new one (Microsoft.Windows.Globalization.ApplicationLanguages.PrimaryLanguageOverride), also update the new one to the latest value.

Only do this for packaged apps, as Windows.Globalization.ApplicationLanguages.PrimaryLanguageOverride is not supported for unpackaged apps.

WARNING:

### DIFF
--- a/dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/src/ResourceContext.cpp
+++ b/dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/src/ResourceContext.cpp
@@ -6,6 +6,8 @@
 #include "ResourceContext.g.cpp"
 #include "winrt/Windows.Globalization.h"
 
+#include <AppModel.Identity.h>
+
 const wchar_t c_languageQualifierName[] = L"Language";
 
 #include "ApplicationLanguages.h"
@@ -99,6 +101,16 @@ void ResourceContext::Apply()
         if (!eachValue.Value().empty())
         {
             winrt::check_hresult(MrmSetQualifier(m_resourceContext, eachValue.Key().c_str(), eachValue.Value().c_str()));
+        }
+    }
+
+    // sync with Windows::Globalization::ApplicationLanguages::PrimaryLanguageOverride if it has been updated more recently
+    if (AppModel::Identity::IsPackagedProcess())
+    {
+        auto language = winrt::Windows::Globalization::ApplicationLanguages::PrimaryLanguageOverride();
+        if (language != ApplicationLanguages::PrimaryLanguageOverride())
+        {
+            ApplicationLanguages::PrimaryLanguageOverride(language);
         }
     }
     if (!ApplicationLanguages::PrimaryLanguageOverride().empty())


kmahone (approved):

dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/src/ResourceContext.cpp
```diff
@@ -101,6 +103,16 @@ void ResourceContext::Apply()
             winrt::check_hresult(MrmSetQualifier(m_resourceContext, eachValue.Key().c_str(), eachValue.Value().c_str()));
         }
     }
+
+    // sync with Windows::Globalization::ApplicationLanguages::PrimaryLanguageOverride if it has been updated more recently
+    if (AppModel::Identity::IsPackagedProcess())
+    {
+        auto language = winrt::Windows::Globalization::ApplicationLanguages::PrimaryLanguageOverride();
```
huichen123, line 110: do we need to check this every time loading resource? can we do it once in initialization, and if it's set, make MWGA.PrimaryLanguageOverride not settable?

evelynwu-msft: I think the primary obstacle with that approach is one of timing... unless the developer explicitly listens to the `Application.ResourceManagerRequested` event then WinUI is going to use MRT Core before custom user code runs. This would then result in WGA.PrimaryLanguageOverride (seemingly) never being respected.

Ideally there would be an event on WGA that we could listen to for changes to PrimaryLanguageOverride (and similarly for MWGA) but in the absence of such an event this seems to me the only way to ensure correct behavior.

evelynwu-msft (approved):
:shipit:

dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/src/ResourceContext.cpp
huichen123: https://learn.microsoft.com/en-us/uwp/api/windows.applicationmodel.resources.core.resourcequalifierobservablemap.mapchanged?view=winrt-26100 listens for language change, but we probably don't want a dependency on that. Or the override is essentially stored in a registry. Probably out of scope though.
MWGA.PrimaryLanguageOverride is persisted. so restarting app would solve that timing issue. not sure whether that's acceptable.
just some thoughts. i am ok with leaving it alone. Hopefully reading is fast enough.

evelynwu-msft: Well, developers _don't_ like restarting their apps in order for language changes to propagate. :)

We can always revisit if benchmarks/anecdata indicate that this is causing perf issues.

huichen123 (approved):

stefansjfw:
Closing in favor of #4608

PR #4663, opened 2024-08-20
New PackageDeploymentManager methods not tagged with new v2 contract

DrusTheAxe:
https://task.ms/53375777

```
static Boolean IsPackageDeploymentFeatureSupported(PackageDeploymentFeature feature);
Boolean IsPackageRegistrationPending(String packageFullName);
Boolean IsPackageRegistrationPendingForUser(String userSecurityId, String packageFullName);
```

Should be prefixed with [contract(PackageDeploymentContract, 2)]

NOTE: Also backporting to 1.6

### DIFF
--- a/dev/PackageManager/API/PackageManager.idl
+++ b/dev/PackageManager/API/PackageManager.idl
@@ -471,11 +471,13 @@ namespace Microsoft.Windows.Management.Deployment
         //-------------------------------------------------------------
         // IsRegistrationPending
 
-        [contract(PackageDeploymentContract, 2)]
-        Boolean IsPackageRegistrationPending(String packageFullName);
+        /// @warning The parameter is should be "packageFullName" but can't due to http://task.ms/53280356.
+        ///          Consider the current (wrong) parameter name deprecated until vFuture (2.0) when we can change to the new (right) parameter name.
+        Boolean IsPackageRegistrationPending(String packageFamilyName);
 
-        [contract(PackageDeploymentContract, 2)]
-        Boolean IsPackageRegistrationPendingForUser(String userSecurityId, String packageFullName);
+        /// @warning The parameter is should be "packageFullName" but can't due to http://task.ms/53280356.
+        ///          Consider the current (wrong) parameter name deprecated until vFuture (2.0) when we can change to the new (right) parameter name.
+        Boolean IsPackageRegistrationPendingForUser(String userSecurityId, String packageFamilyName);
     }
 
     [contract(PackageDeploymentContract, 1)]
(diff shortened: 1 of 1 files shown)


cwruss (approved):
LGTM

codendone (approved):

dev/PackageManager/API/PackageManager.idl
```diff
@@ -470,8 +471,10 @@ namespace Microsoft.Windows.Management.Deployment
         //-------------------------------------------------------------
         // IsRegistrationPending
 
+        [contract(PackageDeploymentContract, 2)]
```
brialmsft, line 474: IsPackageRegistrationPending and IsPackageRegistrationPendingForUser are present in 1.5-stable. The only difference I see is in their parameter names. IIRC parameter names can be significant, though I think that was mostly for out-param names (affects how they project for languages like Javascript which can have multiple return values).
In particular, I don't know offhand whether param names are considered in the algorithm that hashes together signature elements of the interface to generate an IID.

My recommendation would be to remove the contract attribute, and revert the param name to packageFamilyName just in case. If you want to validate that it has no impact on IID, it would probably be safe to change.

codendone: Related: http://task.ms/53280356

brialmsft: And to be clear: adding this contract attribute will _remove_ these methods from the existing IPackageDeploymentManager interface, and change its IID. So this would be a breaking change for 1.5 -> 1.6

DrusTheAxe: You're right. I misread the diff

Param name doesn't affect IID generation, right? Just method name, return values and parameter types?

brialmsft: I _think_ you're right. It's a MidlRT implementation detail, but I would expect node_param::GetInterfaceString to be the thing that contributes to the IID generation, and it doesn't factor in its own name, it just defers to the child node (which should represent the param type). Disclaimer: I'm a bit rusty in MidlRT code, and would still recommend building this and comparing the IID in the generated WinMD against an existing 1.5 WinMD.

brialmsft: Though per the bug that @codendone linked, parameter name changes are breaking for named parameters in C#. Obviously breaking the IID is a bigger deal, but now's the right time to decide whether correcting inaccurate param names is worth the minor break.

DrusTheAxe: #Sad now. OK, param names going back until the future mythical 2.0 :P

brialmsft (approved):
Did a full diff against stable-1.5, and all changes in the IDL now look safely versioned to me.

PR #4664, opened 2024-08-20
Fix ApplicationData.MachineFolder path

DrusTheAxe:
Per Chris Wall `MachineFolder` path should be `%ProgramData%\Microsoft\Windows\AppRepository\Families\ApplicationData\...pkgfamilyname...\Machine` (that ...\Families\... directory was missing)


cwruss (approved):
Looks good. Thanks for updating the tests to match.

PR #4668, opened 2024-08-21
1.6 Fix ApplicationData.MachineFolder path

DrusTheAxe:
cherry-pick [Fix ApplicationData.MachineFolder path](https://github.com/microsoft/WindowsAppSDK/pull/4664#top) #4664 

https://github.com/microsoft/WindowsAppSDK/pull/4664


bpulliam (changes requested):
This is targeting main and thus bringing in 42 commits. It needs to target release/1.6-stable

codendone (approved):

bpulliam (approved):

PR #4674, opened 2024-08-22
Fix centrally-versioned package references

JesseCol:
For the projects under the VSIX folder, the build stopped honoring the "Directory.Packages.props" file at the root of the repo.  This causes us to pick up old versions of all these packages, and hilarity ensues.

I'm not sure at the moment what caused this change in behavior, but this fix looks safe as we continue to investigate.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -2,10 +2,7 @@
     This is the project file used by CentralPackageVersions, and used to control all PackageReference's
     and make sure the version is only declared in a single place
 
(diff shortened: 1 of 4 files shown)


kythant (approved):

kythant:
Do you have one for the main branch coming?

JesseCol:
> Do you have one for the main branch coming?
Not yet but I'll do it soon.  Tracking with bug 53411362.

PR #4714, opened 2024-09-13
Create a DevCheck NuGet Package from which DevCheck.cmd in feeder repos can restore and utilize

kythant:
This PR setups a new pipeline to build a new Microsoft.WinAppSDK.DevCheck NuGet Package which will contain DevCheck.ps1
https://microsoft.visualstudio.com/ProjectReunion/_build?definitionId=144728

The plan is for DevCheck.cmd in feeder repos to restore the latest version of the Microsoft.WinAppSDK.DevCheck where it will find and run DevCheck.ps1


DrusTheAxe (approved):
LGTM

PR #4744, opened 2024-09-24
[WinAppSDK 1.5] Roll back build image to VS 17.9

kythant:
Our build image recently got updated to VS 17.10, which caused some changes and broke a few unit tests. This change rolls our image back to a previous version, that contains VS 17.9.

This is the "1.0.02629.40" image, described on this page:

https://eng.ms/docs/products/onebranch/infrastructureandimages/buildinfrastructureandimages/containerimages/windowsimages/windows2019vse2022/releasenotes


kythant:
Public PR is failing but it is validated here
https://microsoft.visualstudio.com/ProjectReunion/_build/results?buildId=107553706&view=results

JesseCol (approved):

PR #4747, opened 2024-09-25
Refactor Repo Mirroring from Two Pipelines to One Pipeline

kythant:
Before, we needed two pipelines to mirror our repos. One pipeline that is set on a CI trigger to trigger another internal pipeline that does the git actions. 

This changes makes it so we only need one pipeline on the github side.


build/WindowsAppSDK-SyncTrigger.yml
```diff
@@ -1,12 +1,45 @@
 name: $(BuildDefinitionName)_$(date:yyMM).$(date:dd)$(rev:rrr)
+
+resources:
+  repositories:
+    - repository: InternalWindowsAppSDK
+      type: git
+      name: ProjectReunion/WindowsAppSDK
+      ref: refs/heads/main
+    - repository: WindowsAppSDKConfig
```
JesseCol, line 9: Why do you need WindowsAppSDKConfig?

kythant: Oh wait, yea we don't

JesseCol (approved):

PR #4771, opened 2024-10-03
CameraCaptureUI API Implementation

Saharsh979:
**Validation Note**: A microsoft employee must use /azp run to validate using the pipelines below.
## API Design Overview

The CameraCaptureUI API allows desktop applications to effortlessly integrate native camera capture features across various Windows platforms. This proposal aims to address the limitations of the existing OS CameraCaptureUI, primarily designed for UWP, which poses challenges for desktop environments. Key issues include its reliance on CoreWindow and lack of support for IInitializeWithWindow, necessitating cumbersome workarounds.
 
This new API streamlines the integration of camera capture capabilities while ensuring feature parity across all Windows platforms supported by WinAppSDK, thus simplifying developers' workflows.
 
The CameraCaptureUI class provides a comprehensive UI for capturing audio, video, and photos. It allows customization of capture settings, including trimming video and adjusting camera settings.
 
### Constructor:

- CameraCaptureUI(windowId) - Initializes a new CameraCaptureUI object with the specified window ID.
 
### Properties 

- PhotoSettings: Manages photo capture settings.

- VideoSettings: Manages video capture settings.
 
### Methods

- CaptureFileAsync: Launches the capture UI and returns the resulting file.
 
This implementation ensures that the CameraCaptureUI API aligns closely with existing UWP functionality, only modifying the constructor to accept a WindowId.
 
## Example C++ Code

This example shows how to use the CameraCaptureUI Class to take a picture. 

```c++
    // Get the WindowId for the window
    Microsoft::UI::WindowId windowId = this->AppWindow().Id();
 
    // Initialize CameraCaptureUI with a window handle
    winrt::Microsoft::Windows::Media::Capture::CameraCaptureUI cameraUI(windowId);
 
    // Configure Photo Settings
    cameraUI.PhotoSettings().Format(CameraCaptureUIPhotoFormat::Jpeg);
    cameraUI.PhotoSettings().AllowCropping(false);
 
    // Capture a photo asynchronously
    auto photo = co_await cameraUI.CaptureFileAsync(CameraCaptureUIMode::Photo);
 
    // Assert the result
    if (photo != nullptr)
    {
        // Log success if photo capture was successful
        Log::Comment(L"Photo capture was successful.");
    }
    else
    {
        // Log error if photo capture failed or was canceled
        Log::Error(L"Photo capture failed or was canceled.");
    }
```
## Testing
- Successful local build
- Produced a experimental NuGet package leveraging the aggregator build pipeline, and tested it using a sample app by calling the CaptureFileAsync API, which functioned as intended.
- Executed unit tests with the mock framework package, and everything performed as expected.

#### **Note**: The Interactive Experiences package has been incorporated as a dependency in the Foundation repo to utilize Microsoft.UI API features.


dev/Common/PropertyHelper.h
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) Microsoft Corporation and Contributors. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#pragma once
+
+namespace winrt
+{
+    template<typename TStored> struct basic_property
```
dmachaj, line 8: Can you use [wil::single_threaded_property](https://github.com/microsoft/wil/blob/6f60a1b76fb812c6af5db1bc7abdec0001edd43f/include/wil/cppwinrt_authoring.h#L48) instead of defining a new simple property wrapper?

dev/Common/TelemetryHelper.h
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) Microsoft Corporation and Contributors.
+// Licensed under the MIT License.
+#pragma once
+class TelemetryHelper
+{
+public:
+    TelemetryHelper()
+    {
+        m_isPackagedApp = AppModel::Identity::IsPackagedProcess();
```
dmachaj, line 9: It seems like both of these fields can be marked `const` and initialized in the initializer list.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUIPhotoCaptureSettings.h
```diff
@@ -0,0 +1,31 @@
...
+#include "Microsoft.Windows.Media.Capture.CameraCaptureUIPhotoCaptureSettings.g.h"
+#include "PropertyHelper.h"
+
+namespace winrt::Microsoft::Windows::Media::Capture::implementation
+{
+    struct CameraCaptureUIPhotoCaptureSettings : CameraCaptureUIPhotoCaptureSettingsT<CameraCaptureUIPhotoCaptureSettings>
+    {
+        CameraCaptureUIPhotoCaptureSettings() = default;
+
+        void validate();
```
dmachaj, line 13: Should some of these methods, such as `validate()`, be private?

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUIPhotoCaptureSettings.h
```diff
@@ -0,0 +1,31 @@
...
+        basic_property<winrt::Windows::Foundation::Size> CroppedAspectRatio{ 0.0f, 0.0f };
+        basic_property<winrt::Windows::Foundation::Size> CroppedSizeInPixels{ 0.0f, 0.0f };
+        basic_property<winrt::Microsoft::Windows::Media::Capture::CameraCaptureUIPhotoFormat> Format{ winrt::Microsoft::Windows::Media::Capture::CameraCaptureUIPhotoFormat::Jpeg };
+        basic_property<winrt::Microsoft::Windows::Media::Capture::CameraCaptureUIMaxPhotoResolution> MaxResolution{ winrt::Microsoft::Windows::Media::Capture::CameraCaptureUIMaxPhotoResolution::HighestAvailable };
+    };
+}
+
+namespace winrt::Microsoft::Windows::Media::Capture::factory_implementation
+{
+    struct CameraCaptureUIPhotoCaptureSettings : CameraCaptureUIPhotoCaptureSettingsT<CameraCaptureUIPhotoCaptureSettings, implementation::CameraCaptureUIPhotoCaptureSettings>
```
dmachaj, line 28: Its been long enough that I forgot some details, but is this factory supposed to be non-agile so it has thread affinity?  Or is it fine to be agile and the API is multi-thread safe it just needs to be called from a UI thread?

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUITelemetry.h
```diff
@@ -0,0 +1,22 @@
...
+
+#pragma once
+#include "..\WindowsAppRuntime_Insights\WindowsAppRuntimeInsights.h"
+#include <TelemetryHelper.h>
+
+class CameraCaptureUITelemetry : public wil::TraceLoggingProvider
+{
+    IMPLEMENT_TRACELOGGING_CLASS_WITH_MICROSOFT_TELEMETRY(CameraCaptureUITelemetry, "Microsoft.WindowsAppSDK.CameraCaptureUITelemetry",
+        (0x8ab25b76, 0xfe88, 0x4887, 0xb9, 0x08, 0x24, 0xf8, 0xca, 0x38, 0x2b, 0x9d));
+    //{ 8ab25b76-fe88-4887-b908-24f8ca382b9d }
```
dmachaj, line 12: Was this GUID derived from the provider name using tlgguid.exe?

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUIVideoCaptureSettings.cpp
```diff
@@ -0,0 +1,45 @@
...
+#include "CameraCaptureUIVideoCaptureSettings.h"
+#include "Microsoft.Windows.Media.Capture.CameraCaptureUIVideoCaptureSettings.g.cpp"
+
+//using namespace winrt::Windows::Media::Capture;
+
+namespace winrt::Microsoft::Windows::Media::Capture::implementation
+{
+    void CameraCaptureUIVideoCaptureSettings::validate()
+    {
+        float durationInMs = MaxDurationInSeconds * 1000.0f;
```
dmachaj, line 13: Do these validate methods need STA thread type checking?

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI2.idl
```diff
@@ -0,0 +1,88 @@
...
+        Single MaxDurationInSeconds;
+        Microsoft.Windows.Media.Capture.CameraCaptureUIMaxVideoResolution MaxResolution;
+    }
+
+    [contract(CameraCaptureUIContract, 1)]
+    [threading(sta),
+    marshaling_behavior(none)]
+    runtimeclass CameraCaptureUI
+    {
+    //CameraCaptureUI(Windows.UI.WindowId window);
```
Hemantxk, line 80: nit: remove this

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI2.idl
```diff
@@ -0,0 +1,88 @@
...
+    CameraCaptureUI(Microsoft.UI.WindowId window);
+
+    CameraCaptureUIPhotoCaptureSettings PhotoSettings{ get; };
+    CameraCaptureUIVideoCaptureSettings VideoSettings{ get; };
+
+    Windows.Foundation.IAsyncOperation<Windows.Storage.StorageFile> CaptureFileAsync(Microsoft.Windows.Media.Capture.CameraCaptureUIMode mode);
+};
```
Hemantxk, line 87: nit: add space here

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.vcxitems
```diff
@@ -0,0 +1,30 @@
...
+    <ClCompile Include="$(MSBuildThisFileDirectory)CameraCaptureUIVideoCaptureSettings.cpp" />
+  </ItemGroup>
+  <ItemGroup>
+    <ClInclude Include="$(MSBuildThisFileDirectory)CameraCaptureUI.h" />
+    <ClInclude Include="$(MSBuildThisFileDirectory)CameraCaptureUIPhotoCaptureSettings.h" />
+    <ClInclude Include="$(MSBuildThisFileDirectory)CameraCaptureUITelemetry.h" />
+    <ClInclude Include="$(MSBuildThisFileDirectory)CameraCaptureUIVideoCaptureSettings.h" />
+  </ItemGroup>
+  <ItemGroup>
+    <Midl Include="$(MSBuildThisFileDirectory)CameraCaptureUI2.idl" />
```
Hemantxk, line 28: Why 2 and not just `CameraCaptureUI.idl`

dev/WindowsAppRuntime_DLL/pch.h
```diff
@@ -41,6 +41,8 @@
 #include <winrt/Windows.ApplicationModel.Activation.h>
 #include <winrt/Windows.ApplicationModel.Core.h>
 #include <winrt/Windows.ApplicationModel.AppExtensions.h>
+#include <winrt/Windows.ApplicationModel.DataTransfer.h>
+#include <winrt/Windows.ApplicationModel.h>
```
Hemantxk, line 45: duplicate include

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUITelemetry.h
```diff
@@ -0,0 +1,22 @@
...
+public:
+    DEFINE_COMPLIANT_MEASURES_EVENT_PARAM2(CaptureInitiated, PDT_ProductAndServicePerformance,
```
Hemantxk, line 14: You can create an Activity instead to mark the start and end of the capture operation. Consider adding mode of capture in these events.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.cpp
```diff
@@ -0,0 +1,145 @@
...
+#include <future>
+#include <ppltasks.h>
+#include <pplawait.h>
+#include <ShlObj.h>
+#include "CameraCaptureUI.h"
+#include "Microsoft.Windows.Media.Capture.CameraCaptureUI.g.cpp"
+#include "CameraCaptureUITelemetry.h"
+#include "TelemetryHelper.h"
+
+//using namespace winrt::Windows::Media::Capture;
```
Hemantxk, line 13: remove this

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+            if ((mode == CameraCaptureUIMode::PhotoOrVideo) || (mode == CameraCaptureUIMode::Video))
+            {
+                m_videoSettings->validate();
+                m_videoTokenFile = co_await CreateEmptyFileAndGetToken(m_videoSettings->GetFileExtension());
+                properties.Insert(L"VideoFileToken", box_value(m_videoTokenFile.token));
+                m_videoSettings->Serialize(properties);
+            }
+
+            if (addPicker)
```
Hemantxk, line 107: This is always false

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.cpp
```diff
@@ -0,0 +1,145 @@
...
+    {
+        LauncherOptions options;
+        if (targetWindow.Value)
+        {
+            check_hresult(options.as<IInitializeWithWindow>()->Initialize(reinterpret_cast<HWND>(targetWindow.Value)));
+        }
+
+        options.TreatAsUntrusted(false);
+        options.DisplayApplicationPicker(false);
+        options.TargetApplicationPackageFamilyName(L"Microsoft.WindowsCamera_8wekyb3d8bbwe");
```
Hemantxk, line 33: Consider having this application package name and other strings stored in static consts.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+        Uri launchUri{ L"microsoft.windows.camera.picker:" };
+        winrt::Windows::System::LaunchUriResult result = co_await Launcher::LaunchUriForResultsAsync(launchUri, options, properties);
+        if (!result.Result())
```
Hemantxk, line 37: Should we check results.Status() before this?

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+        co_return token_and_path{ token, tempFilePath };
+    }
+
+    IAsyncOperation<StorageFile> CameraCaptureUI::CaptureFileAsync(CameraCaptureUIMode mode)
+    {
+        bool isAppPackaged = m_telemetryHelper.IsPackagedApp();
+        PCWSTR appName = m_telemetryHelper.GetAppName().c_str();
+        try
+        {
```
Hemantxk, line 66: There should be a lock to prevent/hold parallel calls to CaptureFileAsync

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUIVideoCaptureSettings.cpp
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+#include "pch.h"
+#include "CameraCaptureUIVideoCaptureSettings.h"
+#include "Microsoft.Windows.Media.Capture.CameraCaptureUIVideoCaptureSettings.g.cpp"
+
+//using namespace winrt::Windows::Media::Capture;
```
Hemantxk, line 7: Remove this

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUIPhotoCaptureSettings.cpp
```diff
@@ -0,0 +1,55 @@
...
+        Size const aspect = CroppedAspectRatio;
+        auto hasPhotoFixedSizeConstraint = (size.Width != 0) || (size.Height != 0);
+        auto hasPhotoAspectRatioConstraint = (aspect.Width != 0) || (aspect.Height != 0);
+        auto hasPhotoSizeConstraint = hasPhotoFixedSizeConstraint || hasPhotoAspectRatioConstraint;
+
+        if (hasPhotoAspectRatioConstraint && hasPhotoAspectRatioConstraint)
+        {
+            throw hresult_invalid_argument(L"PhotoSettings can't have both a size and aspect ratio specified");
+        }
+        else if (AllowCropping && hasPhotoSizeConstraint)
```
Hemantxk, line 24: I think this check is incorrect, logic should be - If a size constraint is set then cropping must be enabled.

test/CameraCaptureUITests/wil.natvis
```diff
@@ -0,0 +1,3 @@
+<?xml version="1.0" encoding="utf-8"?>
```
Hemantxk, line 1: This file is needed? if not, delete.

test/CameraCaptureUITests/framework.h
Hemantxk, line 1: Remove if not needed

test/CameraCaptureUITests/CameraCaptureUITests.vcxproj
```diff
@@ -0,0 +1,274 @@
...
+    </ClCompile>
+  </ItemGroup>
+  <ItemGroup>
+    <Natvis Include="wil.natvis" />
+  </ItemGroup>
+  <ItemGroup>
+    <None Include="packages.config" />
+  </ItemGroup>
+  <ItemGroup>
+    <Reference Include="Microsoft.Windows.System.Power">
```
Hemantxk, line 233: nit:
```suggestion
    <Reference Include="Microsoft.Windows.Media.Capture">
```

test/DynamicDependency/data/Microsoft.WindowsAppRuntime.Framework/appxmanifest.xml
```diff
@@ -142,6 +142,12 @@
         <ActivatableClass ActivatableClassId="Microsoft.Windows.System.Power.PowerManager" ThreadingModel="both" />
       </InProcessServer>
     </Extension>
+    <Extension Category="windows.activatableClass.inProcessServer">
```
Hemantxk, line 145: Move this block up and keep in sorted order

test/CameraCaptureUITests/pch.h
```diff
@@ -0,0 +1,31 @@
...
+#include <windows.h>
+
+#include "WexTestClass.h"
+#include <wil\resource.h>
+#include <winrt\Windows.Foundation.h>
+#include <winrt\Windows.Foundation.Collections.h>
+#include <winrt\Windows.Management.Deployment.h>
+
+#include <WindowsAppRuntime.Test.Bootstrap.h>
+namespace TP = ::Test::Packages;
```
Hemantxk, line 29: This namespace is also being shortened in APITests.cpp

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUIPhotoCaptureSettings.h
Saharsh979: The validate() method is used in CaptureFileAsync to ensure the integrity of m_photoSettings and m_videoSettings before file capture. Since it's crucial for validating the state, keeping it public makes more sese and allows CaptureFileAsync to call it effectively.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUITelemetry.h
Saharsh979: Yes, thats right.

WindowsAppRuntime.sln
```diff
@@ -529,285 +529,378 @@ Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "PackageManager.Test.M.White
...
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "CameraCaptureUI", "CameraCaptureUI", "{0833D8EF-6E11-4133-B0EE-9B7625CD615E}"
+EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "CameraCaptureUI", "dev\Interop\CameraCaptureUI\CameraCaptureUI\CameraCaptureUI.vcxitems", "{95409D1E-843F-4316-8D8E-471B3E203F94}"
+EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "CameraCaptureUITests", "test\CameraCaptureUITests\CameraCaptureUITests.vcxproj", "{1DAA2342-CF55-48E5-B49C-982FA5C07014}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
+		Debug|ARM = Debug|ARM
```
AjitSurana, line 543: Adding ARM configuration is intentional?

dev/Common/PropertyHelper.h
Saharsh979: Sure, on the docs, it mentions "For settable properties, use single_threaded_rw_property<T> instead."
I understand that using this is preferable to creating a new property wrapper, does it offer any additional benefits as well?

Saharsh979: Updated!

dev/Common/TelemetryHelper.h
Saharsh979: Updated!

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.cpp
Saharsh979: Removed.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.cpp
Saharsh979: Makes sense, updated!

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUIVideoCaptureSettings.cpp
Saharsh979: Done.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUIPhotoCaptureSettings.cpp
Saharsh979: Thanks for pointing out, updated!

test/CameraCaptureUITests/framework.h
Saharsh979: Removed.

test/CameraCaptureUITests/wil.natvis
Saharsh979: It's not a hard requirement but it does enhance the debugging experience.

test/DynamicDependency/data/Microsoft.WindowsAppRuntime.Framework/appxmanifest.xml
Saharsh979: Done.

test/CameraCaptureUITests/pch.h
Saharsh979: Removed redundant code.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.cpp
Saharsh979: Done.

test/CameraCaptureUITests/CameraCaptureUITests.vcxproj
Saharsh979: Updated.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.vcxitems
Saharsh979: We can stick with "CameraCaptureUI.idl" for simplicity. However, if we expect more modifications or want to distinguish it from the original, adding this version number might help with clarity.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUIPhotoCaptureSettings.h
Saharsh979: Yes, the factory can be designed to be agile, meaning it can operate across multiple threads, as long as it adheres to the requirement of being called from a UI thread. 

Can you elaborate more on "thread affinity" in this context and your concerns regarding it.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI2.idl
```diff
@@ -0,0 +1,87 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.Windows.Media.Capture
+{
+    [contractversion(1)]
+    apicontract CameraCaptureUIContract {};
```
akanpatel2206, line 7: Better to add the contract "CameraCaptureUIContract" in specs/WinRT/WinRTAPIContracts.md as well.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUIVideoCaptureSettings.cpp
Saharsh979: The CameraCaptureUI API needs to be called from an STA thread. If it’s invoked from a non-STA thread, it will throw an exception: RPC_E_WRONG_THREAD. This occurs because accessing some UI elements or WinRT components which interact with UI is restricted to the thread that created it, which is typically the UI/STA thread.

I believe we don’t need an additional layer of checking or handling for this. The existing behavior sufficiently enforces the STA requirement.

dev/Common/TelemetryHelper.h
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) Microsoft Corporation and Contributors.
+// Licensed under the MIT License.
+#pragma once
+class TelemetryHelper
+{
+public:
+    TelemetryHelper()
```
AjitSurana, line 7: TelemetryHelper does not look like an appropriate name of this file/class here.
This class is helping in getting App information. AppInfo or AppProperties or similar name would be more appropriate.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.cpp
```diff
@@ -0,0 +1,142 @@
...
+            co_return L"";
+        }
+        else if (auto t = result.Result().TryLookup(L"SelectedTokens"))
+        {
+            co_return t.as<hstring>();
+        }
+        else
+        {
+            throw hresult_canceled();
```
AjitSurana, line 50: What is hresult_canceled() and why we need to throw it?

dev/Common/TelemetryHelper.h
Saharsh979: I wanted to write a general helper that could be used by others. This way, later more helpers for functionalities beyond just app information or properties could be added, which is why I kept it generic.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.cpp
```diff
@@ -0,0 +1,142 @@
...
+        }
+    }
+
+    concurrency::task<token_and_path> CreateEmptyFileAndGetToken(hstring const& fileExtension)
+    {
+        auto tempFolder = ApplicationData::Current().TemporaryFolder();
+        auto tempFile = co_await tempFolder.CreateFileAsync(L"CCapture" + fileExtension, CreationCollisionOption::GenerateUniqueName);
```
AjitSurana, line 57: Any reasoning why the "CCapture" name is used here? I assume CCapture stands for CameraCapture. Can we use the CameraCapture sort of name instead? the reasoning behind that is it is better human readable name. 

Having clear file name would help the program to determine the source of the file as well.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.cpp
AjitSurana: If two (or more) consecutive CaptureFileAsync async calls run well, no change required.

Interested to know if this is a real use case for the developers since existing UWP API had issues (with lock) itself but not aware if there is bug raised for the same. Great to know if such feedback comes from the preview program. Can be taken up once the preview feedback is received.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.cpp
```diff
@@ -0,0 +1,142 @@
...
+            {
+                file = co_await StorageFile::GetFileFromPathAsync(m_photoTokenFile.path);
+            }
+            else if (tokenValue == m_videoTokenFile.token)
+            {
+                file = co_await StorageFile::GetFileFromPathAsync(m_videoTokenFile.path);
+            }
+            else
+            {
+                file = co_await SharedStorageAccessManager::RedeemTokenForFileAsync(tokenValue);
```
AjitSurana, line 126: If you can get the file name from the tokenValue itself, why you need to manage m_videoTokenFile and m_photoTokenFile, it would simplify your implementation.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.cpp
```diff
@@ -0,0 +1,142 @@
...
+        }
+        catch (const hresult_error& ex)
+        {
+            CameraCaptureUITelemetry::CaptureError(isAppPackaged, appName, ex.code());
+            throw ex;
+        }
+        catch (const std::exception& ex)
+        {
+            CameraCaptureUITelemetry::CaptureError(isAppPackaged, appName, E_FAIL);
+            throw hresult_error(E_FAIL);
```
AjitSurana, line 139: In the pervious catch statement, throw ex is written. But in the current exception, hresult_error(E_FAIL) is written, curious to know the difference?

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.h
```diff
@@ -0,0 +1,51 @@
...
+{
+    struct token_and_path
+    {
+        hstring token;
+        hstring path;
+    };
+
+    struct CameraCaptureUI : CameraCaptureUIT<CameraCaptureUI>
+    {
+        CameraCaptureUI(winrt::Microsoft::UI::WindowId const& window) :
```
AjitSurana, line 20: Shouldn't window name be windowId? or Id? for improving code readability.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUIPhotoCaptureSettings.cpp
```diff
@@ -0,0 +1,54 @@
...
+{
+    void CameraCaptureUIPhotoCaptureSettings::validate()
+    {
+        Size const size = CroppedSizeInPixels;
+        Size const aspect = CroppedAspectRatio;
+        auto hasPhotoFixedSizeConstraint = (size.Width != 0) || (size.Height != 0);
+        auto hasPhotoAspectRatioConstraint = (aspect.Width != 0) || (aspect.Height != 0);
+        auto hasPhotoSizeConstraint = hasPhotoFixedSizeConstraint || hasPhotoAspectRatioConstraint;
+
+        if (hasPhotoAspectRatioConstraint && hasPhotoAspectRatioConstraint)
```
AjitSurana, line 19: Same condition twice

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.cpp
Hemantxk: The problem with current implementation without lock is that parallel calls to CaptureFileAsync API accidently overwrites the values of m_photoTokenFile and m_videoTokenFile member variables depending upon the mode of capture. That seems fundamentally wrong as the first call is not finished and tokens are already overwritten, and UWP API prevent that by lock (of-course that lock is not working as per expectation, needs to be fixed separately).
Either we can address this by modifying the API implementation and make it safe for parallel calls or write a lock and restrict it.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUIPhotoCaptureSettings.cpp
```diff
@@ -0,0 +1,54 @@
...
+        {
+            throw hresult_invalid_argument(L"PhotoSettings can't have both a size and aspect ratio specified");
+        }
+        else if (!AllowCropping && hasPhotoSizeConstraint)
+        {
+            throw hresult_invalid_argument(L"PhotoSettings can't have a ratio or size specified with cropping disabled");
```
AjitSurana, line 25: This as well we should capture in the spec doc.

AjitSurana:
Please add containment in the current PR or in a follow up PR.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.cpp
AjitSurana: If values are overridden, below piece of code will be executed and the result will be as expected.

`file = co_await SharedStorageAccessManager::RedeemTokenForFileAsync(tokenValue);`

Your concern is valid. However, as per the conversation with Saharsh, the API gives expected result so probably a follow up task or waiting to hear from preview feedback is ok.

If the below statement works well, why we need to have `m_photoTokenFile` and `m_videoTokenFile` member variables. Can we remove these two variable?

`file = co_await SharedStorageAccessManager::RedeemTokenForFileAsync(tokenValue);`

Saharsh979: I understand that the Camera app has certain limitations, and the team recognized this when deciding to implement a lock to ensure only one capture operation occurs at a time on the original API. Even if multiple capture operations are initiated, only the focused camera will render, while others will show a message stating "Camera is already in use by another app."
Also, for real world use cases, if people do actually want multiple camera capture instances up, they should go through a feature request process.
For now, I am going ahead with the Locking mechanism and restricting this behavior just as it's done for Original API.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.cpp
Saharsh979: It's simply an H_result error. It's not meant to occur; the purpose is to indicate that this situation shouldn't happen. We needed to throw an error, so we chose hresult_canceled to signify that something went wrong.

In my testing, even if I cancel the camera capture operation or force-terminate it from the task manager during capture, it returns a null result that the user can handle. If we encounter hresult_canceled, it indicates a significant and unexpected issue that ideally shouldn't occur.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.cpp
Saharsh979: That makes sense. For better readability, I'll go ahead and use the full name instead.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.cpp
AjitSurana: Another idea was discussed is can `m_photoTokenFile` and `m_videoTokenFile`  be converted to local variable to avoid lock?

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.cpp
Saharsh979: Updated!

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.cpp
Saharsh979: Note: winrt::hresult_error does not catch std::bad_alloc or other C++ standard exceptions.

In the first block, throw ex keeps the original error details, which helps with specific handling for WinRT issues. The second block catches more general standard exceptions and using throw hresult_error(E_FAIL) indicates a generic failure without providing the original error's context.

Saharsh979: Thanks for your comment! I was looking into best practices for handling exceptions and found the official guidelines. They mention using a catch-all handler for better error management. Updated the code to use the same!

https://github.com/microsoft/WindowsAppSDK/blob/743f4551dcd8eda8db1136d7f61a8ecc7f338325/docs/Coding-Guidelines/Languages-CPP.md

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.h
Saharsh979: Agreed, updated!

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUIPhotoCaptureSettings.cpp
Saharsh979: Updated.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUIPhotoCaptureSettings.cpp
```diff
@@ -0,0 +1,54 @@
...
+    {
+        Size const size = CroppedSizeInPixels;
+        Size const aspect = CroppedAspectRatio;
+        auto hasPhotoFixedSizeConstraint = (size.Width != 0) || (size.Height != 0);
+        auto hasPhotoAspectRatioConstraint = (aspect.Width != 0) || (aspect.Height != 0);
+        auto hasPhotoSizeConstraint = hasPhotoFixedSizeConstraint || hasPhotoAspectRatioConstraint;
+
+        if (hasPhotoAspectRatioConstraint && hasPhotoAspectRatioConstraint)
+        {
+            throw hresult_invalid_argument(L"PhotoSettings can't have both a size and aspect ratio specified");
```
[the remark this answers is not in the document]
Saharsh979: Acknowledged.
Created a task internally to monitor this.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUIPhotoCaptureSettings.cpp
Saharsh979: Agreed.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI2.idl
Saharsh979: Created a task to track this internally.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.cpp
Saharsh979: In addition to the camera app's limitations, the main concern is that the token can get overwritten if locking mechanism is not there. While suggesting that we make the token variables local could eliminate the need for a lock, this approach also presents risks. Without a lock, modifying properties during an ongoing operation—such as changing the extension type and calling CaptureFileAsync again—can lead to unpredictable behavior. The outcome often depends on which call reaches the camera app first, resulting in randomness that we should avoid.
 
Furthermore, testing revealed that while two parallel calls to the camera app may work fine, initiating three or more calls leads to a faulty state, ultimately requiring the app to be force-closed. This highlights a potential bug in Camera app's implementation, but it also shows why we need a locking mechanism.
 
To ensure stable and predictable user experiences, I will proceed with adding the lock to prevent multiple capture operations from happening at the same time.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.cpp
Saharsh979: I agree that we can retrieve the file using the token value provided by the camera. However, we still need to manage separate tokens for video and photo because the camera app’s implementation requires them. It relies on these distinct tokens to determine where to save the photo and video files.

If we don't provide these values, the camera capture operation fails; it won’t know which mode to initiate. This can lead to issues, such as not receiving the accept/decline button even after a successful capture. Therefore, maintaining these tokens is necessary.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.cpp
Saharsh979: We don't necessarily need to check result.Status() explicitly. If the operation is successful, we'll receive a file; in all other cases, we'll return null along with S_OK. This aligns with the UWP behavior, which handles similar scenarios gracefully without throwing exceptions.

UWP behavior which we are maintaining parity with - 
Successful Capture: Returns a storage object with S_OK.
Cancelled Capture: Returns null with S_OK. (No exceptions).
App Uninstalled: Displays a dialog box (Search for app in store) Yes/No? clicking on No returns null with S_OK.
Incorrect URI Mapping: Returns null with S_OK (No exceptions).

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUITelemetry.h
Saharsh979: Creating an activity doesn't seem necessary. I'll take on the task of adding the mode of capture to these events.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.cpp
Hemantxk: What I meant is to use LaunchUriStatus to know if the app was launch successful or not **as it is done in original winrt API**. I didn't mention anywhere that you need to necessarily throw exception in-case of failure to launch. You can simply log the case and return empty string to handle it gracefully and this is a suggestion only, feel free to skip it in this iteration.

build/NuSpecs/AppxManifest.xml
```diff
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
```
dmachaj, line 1: Why is this first line showing up as modified?  Did the Byte Order Mark (BOM) get changed unexpectedly?

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.vcxitems
dmachaj: This seems strange to me as well.  Beginning a numbering system with "2" is odd.  If you need to make further modifications you can just edit the IDL file as it already exists.  There shouldn't be a need to have many IDL files for many changes.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUIPhotoCaptureSettings.cpp
```diff
@@ -0,0 +1,54 @@
...
+            throw hresult_invalid_argument(L"PhotoSettings can't have a ratio or size specified with cropping disabled");
+        }
+    }
+
+    hstring CameraCaptureUIPhotoCaptureSettings::GetFileExtension()
+    {
+        switch (Format)
+        {
+        case CameraCaptureUIPhotoFormat::Jpeg:
+            return L".jpg";
```
dmachaj, line 34: The enum value is `jpeg` but the returned string is `jpg`.  Should these be consistent?  (As best as I can tell both are valid exceptions for a JPEG image).

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUIPhotoCaptureSettings.h
dmachaj: Thread affinity means that it can only be used by a specific thread.  It sounds like that is not what you want here, though.  Agile is fine if it can be used from multiple threads (assuming those threads are "UI" threads).

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUIVideoCaptureSettings.h
```diff
@@ -0,0 +1,28 @@
...
+    struct CameraCaptureUIVideoCaptureSettings : CameraCaptureUIVideoCaptureSettingsT<CameraCaptureUIVideoCaptureSettings>
+    {
+        CameraCaptureUIVideoCaptureSettings() = default;
+
+        void validate();
+
+        hstring GetFileExtension();
+        void Serialize(winrt::Windows::Foundation::Collections::ValueSet const& props);
+
+        wil::single_threaded_rw_property<bool> AllowTrimming{ true };
```
dmachaj, line 17: Is it ok for these properties to be single-threaded (and therefore unsafe to use from multiple threads)?  cppwinrt classes are Agile by default.

eng/PackageReference/ProjectReunion.InteractiveExperiences/Microsoft.ProjectReunion.InteractiveExperiences.TransportPackage.PackageReference.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <!-- This csproj is to facilitate restoring dependencies from Maestro. These versions are pulled from eng/version.details.xml -->
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
```
dmachaj, line 4: I thought WinAppSDK 1.6 added net8.0 support.  I'm surprised net6 is still being used in the main branch.

test/CameraCaptureUITests/wil.natvis
```diff
@@ -0,0 +1,170 @@
...
+<!--
+    Copyright (c) Microsoft. All rights reserved.
+    This code is licensed under the MIT License.
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF
+    ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED
+    TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+    PARTICULAR PURPOSE AND NONINFRINGEMENT.
+-->
+<AutoVisualizer xmlns="http://schemas.microsoft.com/vstudio/debugger/natvis/2010">
```
dmachaj, line 10: I'm not sure the WIL natvis file belongs in this project...

If you want it embedded in your PDB you should be able to pull it out of the WIL nuget package directory.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUIPhotoCaptureSettings.cpp
Saharsh979: That is how things are in the original API, kept it the same to maintain parity.

test/CameraCaptureUITests/wil.natvis
Saharsh979: I understand your point. However, I've noticed that every test project in this repo includes a wil.natvis file, and I figured it would be consistent to add it as well.

build/NuSpecs/AppxManifest.xml
Saharsh979: Fixed this!

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.vcxitems
Saharsh979: Changed it to CameraCaptureUI.idl!

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUIVideoCaptureSettings.h
dmachaj: I don't think these single-threaded-property helpers can be used in this PR.  Your objects are not single-threaded so property reads/writes must be protected by a lock.  There is unfortunately not a mechanism to add a lock to the WIL property helpers so the property getter/setter methods need to be hand-authored with locking included.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUIPhotoCaptureSettings.h
```diff
@@ -0,0 +1,30 @@
...
+    struct CameraCaptureUIPhotoCaptureSettings : CameraCaptureUIPhotoCaptureSettingsT<CameraCaptureUIPhotoCaptureSettings>
+    {
+        CameraCaptureUIPhotoCaptureSettings() = default;
+
+        void validate();
+
+        hstring GetFileExtension();
+        void Serialize(winrt::Windows::Foundation::Collections::ValueSet const& props);
+
+        wil::single_threaded_rw_property<bool> AllowCropping{ true };
```
jonwis, line 17: So on reflection, my original changes here - https://github.com/microsoft/WindowsAppSDK/blob/user/jonwis/camera-capture/dev/Interop/CameraCaptureUIVideoSettings.h - did not properly account for threading.

Windows APIs (excluding Windows.UI.*) are supposed to be free-threaded/multi-thread-apartment, meaning the objects themselves are responsible for ensuring correct locking behavior for modifications of their properties.

The WIL `wil::single_threaded...` types assume the caller is enforcing single-threaded (STA) behavior, which C++/WinRT does not.

My suggestion is to add appropriate locking, and _not_ use the `wil::single_threaded...` types here. Instead, something like:

```c++
struct CameraCaptureUIPhotoCaptureSettings ...
{
    winrt::srwlock m_lock;
    bool AllowCropping() { return read_with_lock(m_allowCropping); }
    void AllowCropping(bool b) { write_with_lock(m_allowCropping, b); }
    winrt::Windows::Foundation::Size CroppedAspectRatio() { return read_with_lock(m_croppedAspectRatio); }
    void CroppedAspectRatio(winrt::Windows::Foundation:Size s) { return write_with_lock(m_croppedAspectRatio, s); }
    // ...
private:
    template<typename T> auto read_with_lock(T& t) {
        return t;
    }
    template<typename T, typename R> void write_with_lock(T& t, R&& r) {
        t = std::forward<R>(r);     
    }
    bool m_allowCropping;
    winrt::Windows::Foundation::Size m_croppedAspectRatio{};
    winrt::Windows::Foundation::Size m_croppedSizeInPixels{};
    // ...
}
```

Then within the methods like GetFileExtension & Serialize, use `auto lock = m_lock.lock_shared()`.  Be sure to only use the `m_....` fields to read values from _inside_ the object's methods; you don't need to go through the locking wrappers again to just read them.

jonwis (changes requested):

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUITelemetry.h
```diff
@@ -0,0 +1,22 @@
...
+#include "..\WindowsAppRuntime_Insights\WindowsAppRuntimeInsights.h"
+#include <TelemetryHelper.h>
+
+class CameraCaptureUITelemetry : public wil::TraceLoggingProvider
+{
+    IMPLEMENT_TRACELOGGING_CLASS_WITH_MICROSOFT_TELEMETRY(CameraCaptureUITelemetry, "Microsoft.WindowsAppSDK.CameraCaptureUITelemetry",
+        (0x8ab25b76, 0xfe88, 0x4887, 0xb9, 0x08, 0x24, 0xf8, 0xca, 0x38, 0x2b, 0x9d));
+    //{ 8ab25b76-fe88-4887-b908-24f8ca382b9d }
+public:
+    DEFINE_COMPLIANT_MEASURES_EVENT_PARAM3(CaptureInitiated, PDT_ProductAndServicePerformance,
```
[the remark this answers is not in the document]
Saharsh979: Updated the code to use DEFINE_COMPLIANT_MEASURES_ACTIVITY now!

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUIVideoCaptureSettings.h
Saharsh979: I've implemented property-level srwlock to ensure that reads and writes are properly protected. This means we can safely access these properties from multiple threads, now.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUIPhotoCaptureSettings.h
Saharsh979: Thanks for the input! I’ve implemented property-level srwlock and also protected the member methods, Validate, GetFileExtension, and Serialize, for both PhotoSettings and VideoSettings using m_lock.lock_shared().

Tested my changes with a NuGet built from the aggregate pipelines on a test app in both STA and MTA environments.

To test this in a multi-threaded environment, I created a WinUI 3 blank app. I modified the photo format to PNG and called CaptureFileAsync for photo mode synchronously, without using co_await.

I then spun off a new thread to change the parameters, attempting to set the format to JPEG. However, I consistently observed that the result retained the PNG format. There’s no randomness this time, and we’ve effectively controlled the behavior that was previously causing unexpected outputs due to timing issues and the absence of property read/write locks.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUITelemetry.h
```diff
@@ -5,17 +5,41 @@
...
+                TraceLoggingWideString(appName, "AppName"),
+                TraceLoggingWideString(captureMode, "CaptureMode"));
+        }
+        CATCH_LOG()
+    END_ACTIVITY_CLASS();
 
-    DEFINE_COMPLIANT_MEASURES_EVENT_PARAM3(CaptureSuccessful, PDT_ProductAndServicePerformance,
-        bool, IsAppPackaged, PCWSTR, AppName, PCWSTR, CaptureMode);
+    BEGIN_COMPLIANT_MEASURES_ACTIVITY_CLASS(CaptureSuccessful, PDT_ProductAndServicePerformance);
+        DEFINE_ACTIVITY_START(bool isAppPackaged, PCWSTR appName, PCWSTR captureMode) noexcept try
```
Hemantxk, line 32: You only need one activity - to indicate the start and the end of capture operation.
You can name it - CameraCaptureOperation maybe?

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUIPhotoCaptureSettings.cpp
```diff
@@ -43,6 +45,7 @@ namespace winrt::Microsoft::Windows::Media::Capture::implementation
 
     void CameraCaptureUIPhotoCaptureSettings::Serialize(winrt::Windows::Foundation::Collections::ValueSet const& props)
     {
+        auto lock = m_lock.lock_shared();
         props.Insert(L"AllowCropping", box_value(AllowCropping()));
```
Hemantxk, line 49: Use private members like m_allowCropping, m_format etc. directly here as done in other places

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUIVideoCaptureSettings.cpp
```diff
@@ -36,6 +38,7 @@ namespace winrt::Microsoft::Windows::Media::Capture::implementation
 
     void CameraCaptureUIVideoCaptureSettings::Serialize(winrt::Windows::Foundation::Collections::ValueSet const& props)
     {
+        auto lock = m_lock.lock_shared();
         props.Insert(L"AllowTrimming", box_value(AllowTrimming()));
```
Hemantxk, line 42: Here too, use private member directly here.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUITelemetry.h
Saharsh979: That makes sense, updated!

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUIPhotoCaptureSettings.cpp
Saharsh979: Done!

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUIVideoCaptureSettings.cpp
Saharsh979: Done!

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.cpp
```diff
@@ -0,0 +1,137 @@
...
+            }
+            else if (mode == CameraCaptureUIMode::Video)
+            {
+                mediaType = L"video";
+            }
+            else
+            {
+                throw hresult_invalid_argument();
+            }
+            auto logCaptureOperation{ CameraCaptureUITelemetry::CameraCaptureOperation::Start(isAppPackaged, appName, mediaType) };
```
Hemantxk, line 96: You can move the start event up.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.cpp
AjitSurana: It is ERROR_CANCELLED HRESULT error code - https://learn.microsoft.com/en-us/uwp/cpp-ref-for-winrt/error-handling/hresult-error.

For HRESULT error code - https://learn.microsoft.com/en-us/uwp/cpp-ref-for-winrt/error-handling/hresult-error.

Why we need to throw hresult_canceled vs. hresult_error? Should we include `message` value so that developers understand what is the issue?

Hemantxk (approved):

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.cpp
```diff
@@ -0,0 +1,137 @@
...
+            }
+            StorageFile file = nullptr;
+
+            file = co_await SharedStorageAccessManager::RedeemTokenForFileAsync(tokenValue);
+            logCaptureOperation.Stop();
+            co_return file;
+        }
+        catch (...)
+        {
+            auto e = winrt::hresult_error(winrt::to_hresult(), winrt::take_ownership_from_abi);
```
jonwis, line 130: Do you need to do this? The default WIL activity system automatically grabs and logs thrown exceptions at point of failure. You shouldn't need this try/catch block here - just start the telemetry activity above, and errors will flow through the normal WIL->Telemetry provider pipeline.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.cpp
```diff
@@ -0,0 +1,137 @@
...
+            auto hr = e.code();
+            auto message = e.message().c_str();
+            CameraCaptureUITelemetry::CaptureError(isAppPackaged, appName, mediaType, hr, message);
+            throw e;
```
jonwis, line 134: Prefer re-throwing, via `throw;` rather than throwing a new hresult_error. C++/WinRT automatically catches exceptions at boundaries and produces the right exception types.

jonwis (approved):
Please check out the try/catch-logger comment. While it's not _wrong_ it shouldn't be strictly necessary either. Looks good - thank you for pushing forward on this!

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.cpp
Saharsh979: We explicitly create a string for the mode, so it makes more sense to use it. Also, if someone passes a value other than photo, video, or photo mode, the operation won't start. I think it functions correctly as is!

Hemantxk: You can pass mode as enum directly as received from caller in the activity, no need to wait for its conversion to string. Early you declare the activity better it is to capture all kinds of errors i.e. it should cover incorrect mode too.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.cpp
Saharsh979: Updated! Thanks for the suggestion.

dev/Interop/CameraCaptureUI/CameraCaptureUI/CameraCaptureUI.cpp
Saharsh979: Updated!

I wanted to make sure I get back the error telemetry for creating measures. I figured, the job gets done with the Telemetry activity provider pipeline as well. Got rid of the try/catch, relying on the telemetry activity now!

PR #4728, opened 2024-09-17
Delegate to OS RegFreeWinRT when available (>=24H1)

DrusTheAxe:
Delegate to OS RegFreeWinRT when available (>=24H1)

Don't detour APIs to our Undocked Reg-Free WinRT implementation if the OS supports Reg-Free WinRT (>=24H1)

https://task.ms/47326505

PR #4785, opened 2024-10-08
Variant of user/drustheaxe/no-urfw to test a pipeline issue

DrusTheAxe:
Variant of [Delegate to OS RegFreeWinRT when available (>=19H1)](https://github.com/microsoft/WindowsAppSDK/pull/4728#top)
 #4728 to test a pipeline issue


DrusTheAxe:
Obsolete so closing out

Issue #138, opened 2020-08-12
Proposal: Converged and improved badge APIs


SetTrend:
<table>
<tr>
<th>Summary</th>
<th>Status</th>
</tr>
<tr>
<td><p><strong>Customer / job-to-be-done</strong>: App developers need to keep their users informed about number of new items/etc via badges on the taskbar regardless of whether their app is currently open</p><p><strong>Problem</strong>: Win32 apps have a different API for using badges on taskbar, with completely different functionality than UWP apps. Win32 apps can't simply set a number (they have to use custom icons), and UWP apps can only use preset badge/number options (they can't use custom icons).</p>

<p><strong>How does our solution make things easier?</strong> All types of apps will benefit from a single, easy-to-use badge API, that supports straightforward numbers and common symbols in addition to custom icons.</p>

</td>

<td>❌&nbsp;Problem&nbsp;validation</br>
❌ Docs</br>
❌&nbsp;Dev&nbsp;spec</br>
❌&nbsp;Solution&nbsp;validation</br>
❌&nbsp;Implementation</br>
❌&nbsp;Samples
</tr>
</table>

## Target customer

<table>
<tr>
<th>Packaging</th>
<th>App types</th>
</tr>
<tr>
<td>✅ Packaged apps</br>
✅ Unpackaged apps
</td>
<td>
✅ WPF</br>
✅ Win32 (C++)</br>
✅ WinForms</br>
✅ Console</br>
✅ UWP</br>
⚠ Cross-plat (Electron, MAUI, React Native, etc)</br>
⚠ PowerShell
</td>
</tr>
</table>

While this work should be consumable by cross-platform platforms to improve experiences there, we should have a separate feature specifically for ensuring badges are built in "natively" and developers don't have to do additional custom work to access the Windows native layer.

## Customer's job-to-be-done

Developer needs to keep their users informed via badges regardless of whether their app is currently open. They may need advanced functionality like custom icons, or might just need simple functionality like setting a number.

## Problems that exist today

Problem | Validated?
--|--
**TODO**: Fill this out | ❌ TODO

## Summary

Converge badge APIs across Win32 and UWP, and also provide objects for the APIs rather than XML (bring in the Toolkit helpers into Reunion itself).

```c#
public class BadgeUpdater
{
  public void Update(byte newValue) {...}
  public void Update(BadgeGlyph newValue) {...}

  //...

  public enum BadgeGlyph
  {
    None,
    Activity,
    Alarm,
    //...
[...]

PR #4823, opened 2024-10-23
API Spec for Badge Notifications in WindowsAppSdk

SatwikKrSharma:
This is the API Spec for the Badge Notifications in WindowsAppSdk. Issue -  https://github.com/microsoft/WindowsAppSDK/issues/138

WARNING:


specs/BadgeNotifications/BadgeNotifications-spec.md
```diff
@@ -0,0 +1,192 @@
...
+
+    DateTime expirationTime = clock::now() + std::chrono::hours(24);
+    badgeNotification.Expiration(expirationTime);
+
+    winrt::BadgeNotificationManager manager = winrt::BadgeNotificationManager::Default();
+    manager.UpdateBadge(badgeNotification);
+}
+```
+
+# API pages
```
anupriya13, line 80: "Pages" can be renamed since it's a single page, something like "API Description"

specs/BadgeNotifications/BadgeNotifications-spec.md
```diff
@@ -0,0 +1,192 @@
...
+| Busy | A glyph that represents the user being busy or engaged in an activity, preventing interruptions. | ![Screenshot](badge-busy.png) |
+| NewMessage | A glyph that indicates the arrival of a new message, often used in messaging and email applications. | ![Screenshot](badge-newmessage.png) |
+| Paused | A glyph that signifies a pause in activity or content, such as media playback being paused. | ![Screenshot](badge-paused.png) |
+| Playing | A glyph that indicates media or content is currently playing. | ![Screenshot](badge-playing.png) |
+| Unavailable | A glyph that shows the user or a service is not currently available or offline. | ![Screenshot](badge-unavailable.png) |
+| Error | A glyph that denotes an error has occurred, which may require user action to resolve. | ![Screenshot](badge-error.png) |
+
+These glyphs are system-provided and standardized, ensuring a consistent look and feel across different applications that use them. By using these glyphs, developers can convey specific states or notifications to users in a visually intuitive manner.
+
```
anupriya13, line 136: Can add "You cannot provide your own badge image; only system-provided badge images can be used."

specs/BadgeNotifications/BadgeNotifications-spec.md
```diff
@@ -0,0 +1,192 @@
...
+# Description
+
+Badge notifications provide users with a visual indicator of new or important content in your application. This feature enhances user engagement by allowing them to quickly see updates without needing to open the app. Badge notifications are already supported for UWP apps. This initiative is to make badge notifications work in WindowsAppSdk.
+
+For more details see:
+
+-   [Badge Notification WinRT APIs](https://learn.microsoft.com/en-us/windows/apps/design/shell/tiles-and-notifications/badges)
+    Defines all the API constructs that we have for Badge Notifications in WinRT today.
+
```
anupriya13, line 28: reference https://github.com/Microsoft/Windows-universal-samples/tree/main/Samples/Notifications
This below part can be added too for clarity:

Badges enable the app to communicate whether the user has certain amount of notifications, or their attention is needed due to an event. They can also convey user status in a chat application.

Scenarios include:

- Setting a numerical badge onto a primary tile
- Setting a glyph badge onto a primary tile
- Clearing badges

specs/BadgeNotifications/BadgeNotifications-spec.md
```diff
@@ -0,0 +1,192 @@
...
+    manager.UpdateBadge(badgeNotification);
+}
+```
+
+## Create a Glyph badge (c++)
+
+ ```c++
+private void updateBadgeGlyph()
+{
+    winrt::BadgeGlyphValue badgeGlyphValue = winrt::BadgeGlyphValue::alert
```
jonwis, line 47: There's already a BadgeGlyph type. Consider `BadgeNotificationGlyph` as the enum type.

jonwis, line 47: For enum values, use `PascalCasing` for member names, so this becomes `BadgeNotificationGlyph::Alert`

specs/BadgeNotifications/BadgeNotifications-spec.md
```diff
@@ -0,0 +1,192 @@
...
+}
+```
+
+## Clear a badge
+
+ ```c++
+private void clearBadge()
+{
+    winrt::BadgeNotificationManager manager = winrt::BadgeNotificationManager::Default();
+    manager.ClearBadge();
```
jonwis, line 60: The verb methods on the `BadgeNotificationManager` don't need to include "Badge" in their method names. Instead, call this `Clear()` and then `UpdateBadge` becomes `Update`, etc.

specs/BadgeNotifications/BadgeNotifications-spec.md
```diff
@@ -0,0 +1,192 @@
...
+    runtimeclass BadgeNotificationManager
+    {
+        // Gets a Default instance of a BadgeNotificationManager
+        static BadgeNotificationManager Default{ get; };
+
+        // Applies a change to the badge's glyph or number.
+        void UpdateBadge(BadgeNotification notification);
+
+        // Removes the Badge Notifications for the App from Action Centre
+        Windows.Foundation.IAsyncAction ClearBadge();
```
jonwis, line 189: Why is "Clear" an async operation but Update is synchronous?

specs/BadgeNotifications/BadgeNotifications-spec.md
```diff
@@ -0,0 +1,192 @@
...
+        // This constructor initializes the BadgeNotification with a predefined glyph from the BadgeGlyphValue enum.
+        // @param value The BadgeGlyphValue representing a system-provided glyph to be displayed on the badge.
+        BadgeNotification(BadgeGlyphValue value);
+
+        // Gets or sets the time after which a Badge Notification should not be displayed.
+        Windows.Foundation.DateTime Expiration;
+    }
+
+    // The manager class which encompasses all Badge Notification API Functionality
+    runtimeclass BadgeNotificationManager
```
jonwis, line 180: What happens when one app has two top-level windows that don't "glom together" - can each top-level window get its own badge? Or is it applied to only the "main" window (whatever that is)?

specs/BadgeNotifications/BadgeNotifications-spec.md
```diff
@@ -0,0 +1,192 @@
...
+        Unavailable, // A glyph representing being unavailable
+        Error, // A glyph representing an error
+    };
+
+    runtimeclass BadgeNotification
+    {
+        // Constructs a BadgeNotification with a numeric badge value.
+        // This constructor initializes the BadgeNotification with a number that will be displayed on the badge.
+        // @param num The numeric value to be displayed on the badge.
+        BadgeNotification(UInt32 num);
```
jonwis, line 168: Parameter name should be `notificationCount` - "num" isn't enough information.

specs/BadgeNotifications/BadgeNotifications-spec.md
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    // The manager class which encompasses all Badge Notification API Functionality
+    runtimeclass BadgeNotificationManager
+    {
+        // Gets a Default instance of a BadgeNotificationManager
+        static BadgeNotificationManager Default{ get; };
+
+        // Applies a change to the badge's glyph or number.
+        void UpdateBadge(BadgeNotification notification);
```
jonwis, line 186: Thinking about this more, is the `BadgeNotification` class necessary? It seems like it's "in the way" - providing little additional value over just calling methods directly.  You could imagine a model like:

```c#
void SetNotificationCount(UInt32 notificationCount);
void SetNotificationCount(UInt32 notificationCount, DateTime expiration);
void SetNotificationGlyph(BadgeNotificationGlyph glyph);
void SetNotificationGlyph(BadgeNotificationGlyph glyph, DateTime expiration);
```

jonwis (changes requested):

specs/BadgeNotifications/BadgeNotifications-spec.md
SatwikKrSharma: It is only applied to the "main" window. Verified in the current behavior as well. When you create a new window of the outlook app only the main windows have the badge notification

specs/BadgeNotifications/BadgeNotifications-spec.md
SatwikKrSharma: 1. Encapsulation: By having a separate `BadgeNotification` class, I have encapsulate all the properties and behaviors related to the badge's appearance and data. This makes the class easier to understand and maintain. The `BadgeNotificationManager` then focuses solely on the management tasks, such as updating and clearing the badge.

2. Extensibility: In the future, the `BadgeNotification` class may need to support additional properties or methods (e.g., badge color, expiration time, or animation). Having a dedicated class for the badge notification makes it easier to extend its functionality without cluttering the management class or violating SRP.

3. Testability: With a separate `BadgeNotification` class, I can test the badge's functionality in isolation from the management logic. This leads to more focused and reliable unit tests, as each test can concentrate on a single aspect of the system.

4. Reusability: If the badge notification logic is encapsulated in its own class, it's possible to reuse the `BadgeNotification` class in different contexts within the application or across different applications, without needing to bring along the management logic.

5. Separation of Concerns: By keeping the `BadgeNotification` and `BadgeNotificationManager` classes separate, I have adhered to the principle of separation of concerns. Each class addresses a distinct concern: one for representing the badge notification data and another for handling the lifecycle of badge notifications.

6. Maintainability: As the application grows and evolves, having a clear separation between the representation of a badge notification and its management will make the codebase more maintainable. Changes to the notification appearance or data structure will not impact the management logic, and vice versa.

7. Collaboration: In a team environment, different developers might work on different aspects of the badge notifications. Having separate classes allows for clearer ownership and reduces the likelihood of merge conflicts when multiple developers are working on the same codebase.

In summary, keeping the `BadgeNotification` class separate from the `BadgeNotificationManager` class adheres to good object-oriented design principles and sets a strong foundation for a maintainable, extensible, and testable codebase.

specs/BadgeNotifications/BadgeNotifications-spec.md
SatwikKrSharma: The `Update` method is synchronous because displaying a notification is usually a quick operation that does not require significant processing or waiting for external resources. When an app calls `Update`, it typically expects the notification to appear immediately or nearly immediately, and there is no need to wait for a potentially long-running operation to complete. Therefore, making this method synchronous allows for a simple and straightforward usage pattern.

On the other hand, the `Clear` method is asynchronous because it involves more than just a simple display action. Removing a notification could potentially require interaction with the system's notification service, which involve I/O operations, updating the notification database, and ensuring that the removal is properly reflected in the Action Center. These operations can take more time and are subject to delays that are outside the control of the application.

By making `Clear` asynchronous, the Windows App SDK allows the calling thread to continue executing while the notification removal process is underway. This prevents the app from becoming unresponsive if the removal operation takes longer than expected. The async pattern also provides a way to handle the completion of the operation with a callback or continuation, which is useful for updating the app's state or UI after the notification has been successfully removed.

In summary, the decision to make `Update` synchronous and `Clear` asynchronous is based on the expected complexity and duration of the operations. 

Similar logic has been used for AppNotification in WindowsAppSdk - https://github.com/microsoft/WindowsAppSDK/blob/b0f4e3477ec6be65f674aee4624f993e1845560a/specs/AppNotifications/AppNotifications-spec.md

specs/BadgeNotifications/BadgeNotifications-spec.md
SatwikKrSharma: Done

specs/BadgeNotifications/BadgeNotifications-spec.md
SatwikKrSharma: Done

specs/BadgeNotifications/BadgeNotifications-spec.md
SatwikKrSharma: This information is already given in description and code examples

specs/BadgeNotifications/BadgeNotifications-spec.md
```diff
@@ -0,0 +1,192 @@
...
+
+# Examples
+
+## Create a numeric badge (c++)
+
+ ```c++
+private void setBadgeNumber(uint32_t num)
+{
+    winrt::BadgeNotification badgeNotification(num);
+    winrt::BadgeNotificationManager manager = winrt::BadgeNotificationManager::Default();
```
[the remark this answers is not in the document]
SatwikKrSharma: Done

specs/BadgeNotifications/BadgeNotifications-spec.md
SatwikKrSharma: Done

SatwikKrSharma: Done

specs/BadgeNotifications/BadgeNotifications-spec.md
SatwikKrSharma: Done

specs/BadgeNotifications/BadgeNotifications-spec.md
SatwikKrSharma: done

specs/BadgeNotifications/BadgeNotifications-spec.md
```diff
@@ -0,0 +1,192 @@
+Badge Notifications in Windows App SDK
+===
+
+This is the spec for proposal: [Issue #138](https://github.com/microsoft/WindowsAppSDK/issues/138)
+
+# Background
+
+A notification badge conveys summary or status information specific to your app. They can be numeric (1-99) or one of a set of system-provided glyphs. Examples of information best conveyed through a badge include network connection status in an online game, user status in a messaging app, number of unread mails in a mail app, and number of new posts in a social media app.
```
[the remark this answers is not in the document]
SatwikKrSharma: Hi Jon, prettier is not able to format the document. I have followed the same convention as the existing spec files in WindowsAppSdk and in the preview I am able to see the spec file correctly.

specs/BadgeNotifications/BadgeNotifications-spec.md
```diff
@@ -0,0 +1,213 @@
...
+Windows, where badges can be displayed on the app's tile. A badge can show a numeric value or a
+system-provided glyph, providing a visual indicator of some status or count directly on the app's
+icon.
+
+## BadgeNotification constructors
+
+| Name                                      | Description                                                                                                                                                                              |
+| ----------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| BadgeNotification(UInt32)                 | Creates a new `BadgeNotification` object with a numeric badge value. This is used to display a specific number on the app's badge.                                                       |
+| BadgeNotification(BadgeNotificationGlyph) | Creates a new `BadgeNotification` object with a glyph value. This uses a predefined glyph from the `BadgeNotificationGlyph` enum to display a system-provided symbol on the app's badge. |
```
FrankC-msft, line 102: Remind me, are overloads with the same arity OK in WinAppSDK?  I remember this causes problems for projections into weakly typed languages.

specs/BadgeNotifications/BadgeNotifications-spec.md
```diff
@@ -0,0 +1,213 @@
...
+## BadgeNotificationGlyph enum
+
+The `BadgeNotificationGlyph` enum defines a set of predefined glyphs that can be used to represent
+various statuses or notifications on an application's badge. These glyphs provide a visual cue on
+the app's tile, allowing users to quickly understand the app's status or to be alerted to new
+information without opening the app.
+
+| Name        | Description                                                                                                 | Glyph                                |
+| ----------- | ----------------------------------------------------------------------------------------------------------- | ------------------------------------ |
+| None        | No glyph is displayed. The badge will appear blank or not be shown at all.                                  | (No Badge Shown)                     |
```
FrankC-msft, line 139: Nit, this order is neither alphabetical after the None entry, nor the same as https://learn.microsoft.com/en-us/windows/apps/design/shell/tiles-and-notifications/badges .  (nor does it group Unavailable with the other colorfil statuses.)  Is there a reason to push Error down the list?

specs/BadgeNotifications/BadgeNotifications-spec.md
```diff
@@ -0,0 +1,213 @@
...
+The `BadgeNotificationManager` class is the central point for managing badge notifications for an
+application. It provides the functionality to update or clear the badge displayed on the app's tile.
+This class allows developers to modify the badge's appearance by setting a numeric value or a glyph
+and also to remove the badge when it is no longer needed.
+
+## BadgeNotificationManager Properties
+
+| Name    | Description                                                                                                        | Type                     |
+| ------- | ------------------------------------------------------------------------------------------------------------------ | ------------------------ |
+| Default | Provides a default instance of the `BadgeNotificationManager` for use in updating or clearing badge notifications. | BadgeNotificationManager |
```
FrankC-msft, line 121: Sample and API call this "Current", not Default.

specs/BadgeNotifications/BadgeNotifications-spec.md
```diff
@@ -0,0 +1,213 @@
...
+or notifications to users in a visually intuitive manner.
+
+> Note: Only system-provided badge images can be used.
+
+# API Details
+
+> Note: all of this new WinAppSDK API is to support Badge Notifications in WindowsAppSdk
+
+```c++ (but really MIDL3)
+namespace Microsoft.Windows.BadgeNotifications
```
FrankC-msft, line 164: Why create a new namespace for .BadgeNotifications instead of putting these enum and classes into Microsoft.Windows.AppNotifications?  Is there a use-case for these outside of apps' badges?

(Adding to the existing namespace looks analogous to [BadgeUpdater ](https://learn.microsoft.com/en-us/uwp/api/windows.ui.notifications.badgeupdater?view=winrt-26100) being in Windows.UI.Notifications Namespace which is described as "Contains classes that encapsulate tile, toast, and badge notifications.".  Is WinAppSDK intentionally more granular?)

specs/BadgeNotifications/BadgeNotifications-spec.md
```diff
@@ -0,0 +1,213 @@
...
+
+| Name   | Description                                                                                                                                                                                                      | Parameters                       | Returns                         |
+| ------ | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------------------------------- | ------------------------------- |
+| Update | Applies a change to the badge's glyph or number by providing a `BadgeNotification` object. This method is used to update the badge displayed on the app's tile to reflect a new status or count.                 | BadgeNotification `notification` | void                            |
+| Clear  | Removes the badge notifications for the app from the action center. This method is useful for clearing any badge notifications when they are no longer relevant or when the app wants to reset its badge status. |                                  | Windows.Foundation.IAsyncAction |
```
FrankC-msft, line 128: Why does Clear need to return IAsyncAction when Update() is void?  Avoid adding async to the API unless we have a specific requirement.

specs/BadgeNotifications/BadgeNotifications-spec.md
SatwikKrSharma: Corrected the spec

specs/BadgeNotifications/BadgeNotifications-spec.md
```diff
@@ -0,0 +1,222 @@
...
+   manager.Update(badgeNotification);
+}
+```
+
+## Create a Glyph badge (c++)
+
+```c++
+private void updateBadgeGlyph()
+{
+   winrt::BadgeNotificationGlyph badgeNotificationGlyph = winrt::BadgeNotificationGlyph::Alert
```
oldnewthing, line 64: TYPO
```suggestion
   winrt::BadgeNotificationGlyph badgeNotificationGlyph = winrt::BadgeNotificationGlyph::Alert;
```

specs/BadgeNotifications/BadgeNotifications-spec.md
```diff
@@ -0,0 +1,222 @@
...
+# Examples
+
+## Create a numeric badge (c++)
+
+```c++
+private void setBadgeNumber(uint32_t num)
+{
+   winrt::BadgeNotification badgeNotification(num);
+   winrt::BadgeNotificationManager manager = winrt::BadgeNotificationManager::Current();
+   manager.Update(badgeNotification);
```
oldnewthing, line 55: The general rule is that [if a method is a verb without a direct object, then the implied direct object is the object whose method is being called](https://devblogs.microsoft.com/oldnewthing/20240724-00/?p=110037).

So `manager.Update()` is updating the manager. `manager.Clear()` is clearing the manager.

Should these be
* manager.UpdateBadge(badgeNotification)
* manager.ClearBadge()

?

specs/BadgeNotifications/BadgeNotifications-spec.md
```diff
@@ -0,0 +1,222 @@
...
+## BadgeNotificationGlyph enum
+
+The `BadgeNotificationGlyph` enum defines a set of predefined glyphs that can be used to represent
+various statuses or notifications on an application's badge. These glyphs provide a visual cue on
+the app's tile, allowing users to quickly understand the app's status or to be alerted to new
+information without opening the app.
+
+| Name        | Description                                                                                                 | Glyph                                |
+| ----------- | ----------------------------------------------------------------------------------------------------------- | ------------------------------------ |
+| None        | No glyph is displayed. The badge will appear blank or not be shown at all.                                  | (No Badge Shown)                     |
```
oldnewthing, line 148: What's the difference between `manager.Update(winrt::BadgeNotification(winrt::BadgeNotificationGlyph::None))` and `manager.Clear()`? It seems that the only difference is that Update is sync, whereas Clear is async, so it seems that Update is easier to use.

specs/BadgeNotifications/BadgeNotifications-spec.md
```diff
@@ -0,0 +1,222 @@
...
+    // The manager class which encompasses all Badge Notification API Functionality
+    runtimeclass BadgeNotificationManager
+    {
+        // Gets a Default instance of a BadgeNotificationManager
+        static BadgeNotificationManager Current{ get; };
+
+        // Applies a change to the badge's glyph or number.
+        void Update(BadgeNotification notification);
+
+        // Removes the Badge Notifications for the App from Action Centre
```
oldnewthing, line 218: en-us spelling please.

specs/BadgeNotifications/BadgeNotifications-spec.md
```diff
@@ -0,0 +1,222 @@
...
+        Windows.Foundation.IAsyncAction Clear();
```
oldnewthing, line 219: Async methods should end in "Async". (But see earlier question of why this is async at all. Update is sync. Why can't Clear be sync?)

specs/BadgeNotifications/BadgeNotifications-spec.md
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    // The manager class which encompasses all Badge Notification API Functionality
+    runtimeclass BadgeNotificationManager
+    {
+        // Gets a Default instance of a BadgeNotificationManager
+        static BadgeNotificationManager Current{ get; };
+
+        // Applies a change to the badge's glyph or number.
+        void Update(BadgeNotification notification);
```
oldnewthing, line 216: If the app has multiple windows, which one gets the badge?

specs/BadgeNotifications/BadgeNotifications-spec.md
```diff
@@ -0,0 +1,222 @@
...
+
+| Name   | Description                                                                                                                                                                                                      | Parameters                       | Returns                         |
+| ------ | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------------------------------- | ------------------------------- |
+| Update | Applies a change to the badge's glyph or number by providing a `BadgeNotification` object. This method is used to update the badge displayed on the app's tile to reflect a new status or count.                 | BadgeNotification `notification` | void                            |
+| Clear  | Removes the badge notifications for the app from the action center. This method is useful for clearing any badge notifications when they are no longer relevant or when the app wants to reset its badge status. |                                  | Windows.Foundation.IAsyncAction |
```
oldnewthing, line 137: What happens if somebody does this:

```cpp
auto clearTask = manager.Clear(); // returns an IAsyncAction
//before the action completes
manager.Update(winrt::BadgeUpdateNotification(42)); // set the badge to 42
co_await clearTask; // wait for the Clear to complete
```
Does the 42 win because it happened after the Clear started? Or does it get wiped out when the Clear completes?

specs/BadgeNotifications/BadgeNotifications-spec.md
```diff
@@ -0,0 +1,222 @@
...
+        Unavailable, // A glyph representing being unavailable
+        Error, // A glyph representing an error
+    };
+
+    runtimeclass BadgeNotification
+    {
+        // Constructs a BadgeNotification with a numeric badge value.
+        // This constructor initializes the BadgeNotification with a number that will be displayed on the badge.
+        // @param num The numeric value to be displayed on the badge.
+        BadgeNotification(UInt32 notificationCount);
```
oldnewthing, line 198: Can I set the badge count to 5551212? Or does it follow the same limits as the WinRT badge API (values above 99 are reported as "99+").

oldnewthing, line 198: What happens if I set the notification count to zero? Does that show a zero badge? Or does that clear the badge?

specs/BadgeNotifications/BadgeNotifications-spec.md
```diff
@@ -0,0 +1,222 @@
...
+
+Example of a numeric badge in Taskbar (One new message in Whatsapp):
+
+![Screenshot](NumericBadgeExample.png)
+
+Example of a system-provided glyph in taskbar (New message in Outlook):
+
+![Screenshot](GlyphBadgeExample.png)
+
+This Badge notification functionality is missing in WinAppSDK, which not only blocks apps like
```
MikeHillberg, line 20: Would help to explain here why this API is different than the existing [BadgeNotification](https://docs.microsoft.com/uwp/api/Windows.UI.Notifications.BadgeNotification) API. (I'm guessing it's because the existing thing is more complex than is necessary anymore, but that should be documented here for the record.)

specs/BadgeNotifications/BadgeNotifications-spec.md
oldnewthing, line 216: Would it be better to get rid of the BadgeNotification entirely and just have three methods
* manager.SetBadgeAsGlyph(BadgeNotificationGlyph value, DateTime expiration)
* manager.SetBadgeAsNumber(UInt32 value, DateTime expiration)
* manager.ClearBadge()

specs/BadgeNotifications/BadgeNotifications-spec.md
```diff
@@ -0,0 +1,222 @@
...
+
+
+# Examples
+
+## Create a numeric badge (c++)
+
+```c++
+private void setBadgeNumber(uint32_t num)
+{
+   winrt::BadgeNotification badgeNotification(num);
```
jonwis, line 53: DISCUSSION: This could also be four methods:

* SetBadgeAsCount(int)
* SetBadgeAsCount(int, timeout)
* SetBadgeAsGlyph(Glyph)
* SetBadgeAsGlyph(Glyph, timeout)

The "set ... as" means you can either have one or the other, not both.

In the future, you can introduce that more complicated BadgeNotificationOptions object that wraps up other common properties to aid in notifying.

jonwis: RECOMMEND: Change to the overloaded methods, removing the BadgeNotification type

specs/BadgeNotifications/BadgeNotifications-spec.md
SatwikKrSharma: Note is added in the spec

specs/BadgeNotifications/BadgeNotifications-spec.md
SatwikKrSharma: Update with count 0, glyph None and clear has same end-user-visible behavior - verified using sample app

specs/BadgeNotifications/BadgeNotifications-spec.md
SatwikKrSharma: ClearBadge has been updated to Synchronous

specs/BadgeNotifications/BadgeNotifications-spec.md
SatwikKrSharma: BadgeNotification class has been removed from the Spec as per discussion

specs/BadgeNotifications/BadgeNotifications-spec.md
SatwikKrSharma: Clear badge has been updated to synchronous as per discussion

specs/BadgeNotifications/BadgeNotifications-spec.md
SatwikKrSharma: fixed

specs/BadgeNotifications/BadgeNotifications-spec.md
SatwikKrSharma: updated

specs/BadgeNotifications/BadgeNotifications-spec.md
SatwikKrSharma: updated

specs/BadgeNotifications/BadgeNotifications-spec.md
SatwikKrSharma: Clear badge updated to synchronous

specs/BadgeNotifications/BadgeNotifications-spec.md
SatwikKrSharma: updated the API spec with the note

SatwikKrSharma: count 0, glypth None and ClearBadge has some end user experience

specs/BadgeNotifications/BadgeNotifications-spec.md
SatwikKrSharma: Updated

specs/BadgeNotifications/BadgeNotifications-spec.md
SatwikKrSharma: fixed

jonwis (approved):
Changes applied from API review. Thanks folks!

PR #4822, opened 2024-10-22
API spec for BackgroundTaskBuilder API

godlytalias:
This is the API spec doc for BackgroundTaskBuilder API to support Background Task Registration for WinAppSDK apps.

Related Bug Report #3840

Issue #3840, opened 2023-09-05
Setting SystemTrigger or ApplicationTrigger as background task triggers throws exception when registering


Diegorro98:
### Describe the bug

I'm looking to create Win32 COM background task with Time, System and Application Triggers. Time works, but when System or Application triggers are set, calling [`BackgroundTaskBuilder.Register`](https://learn.microsoft.com/en-us/uwp/api/windows.applicationmodel.background.backgroundtaskbuilder.register) throws the following exception
```
System.Runtime.InteropServices.COMException: "Incompatible request (0x80070032)"
```

### Steps to reproduce the bug

In [BackgroundTaskWinMainComSample/CS](https://github.com/microsoft/DesktopBridgeToUWP-Samples/tree/master/Samples/BackgroundTaskWinMainComSample/CS) project from DesktopBridgeToUWP-Sample repository I'm able to replicate the issue. 

Only thing needed to update is at [MainWindow.xaml.cs:172](https://github.com/microsoft/DesktopBridgeToUWP-Samples/blob/7c7e66d38e200b19ab42ecc48153361f770bab39/Samples/BackgroundTaskWinMainComSample/CS/MainWindow.xaml.cs#L172), change the Time trigger by any of the others mentioned:
```C#
RegisterBackgroundTaskWithSystem(new SystemTrigger(SystemTriggerType.SessionConnected, false), typeof(TimeTriggeredTask).GUID, typeof(TimeTriggeredTask).Name);
```
```C#
RegisterBackgroundTaskWithSystem(new ApplicationTrigger(), typeof(TimeTriggeredTask).GUID, typeof(TimeTriggeredTask).Name);
```

### Expected behavior

It should be able to register the task and triggers should work.

### NuGet package version

Windows App SDK 1.4.0: 1.4.230822000

### Packaging type

Packaged (MSIX)

### Windows version

Insider Build (xxxxx)

### IDE

Visual Studio 2022

stevenbrix:
@bpulliam what's the status here?

tipa:
Same issue here, I am trying to use `new SystemTrigger(SystemTriggerType.TimeZoneChange, false)`

godlytalias:
Currently we support SystemTrigger or ApplicationTrigger with the newly added BackgroundTaskBuilder API in WinAppSDK 1.7 exp1 release, Sample : https://github.com/microsoft/WindowsAppSDK-Samples/pull/392

Diegorro98:
[...]

PR #4831, opened 2024-10-25
BackgroundTaskBuilder Implementation

godlytalias:
API Spec : #4822 
Issue : #3840 

Added implementation to support the registration of Background Tasks with full trust COM Components. Current [BackgroundTaskBuilder](https://learn.microsoft.com/en-us/uwp/api/windows.applicationmodel.background.backgroundtaskbuilder) is designed for UWP applications and many of the Background Task triggers are not supported for full trust COM Components and are supported only to be registered with Windows Runtime components that is to be launched with backgroundtaskhost process. Due to this, WinAppSDK desktop applications won’t be able to directly register the full trust COM components to be launched with background task triggers and have to do workaround of including the Windows Runtime components in the project.
In this implementation, a BackgroundTaskBuilder API internally registers the entry point of background task as the UniversalBGTask runtime component which then forwards the call to the full trust COM component registered by the application. 

**Testing**
- Tested with locally generated nuget packages

**Pending Action Items**
- In self-contained mode, manifest declaration for the UniversalBGTask class is to be added manually, this need to be fixed
- Support for C# applications to be verified

**Dependent Work Items**
- Wildcard usage by WinUI package affects the behavior which needs to be fixed, https://task.ms/54538137


build/CopyFilesToStagingDir.ps1
```diff
@@ -42,6 +42,12 @@ PublishFile $OverrideDir\PushNotifications-Override.json $FullPublishDir\
 PublishFile $FullBuildOutput\WindowsAppRuntime_DLL\Microsoft.WindowsAppRuntime.dll $FullPublishDir\Microsoft.WindowsAppRuntime\
 PublishFile $FullBuildOutput\WindowsAppRuntime_DLL\Microsoft.WindowsAppRuntime.pdb $FullPublishDir\Microsoft.WindowsAppRuntime\
 PublishFile $FullBuildOutput\WindowsAppRuntime_DLL\Microsoft.WindowsAppRuntime.lib $FullPublishDir\Microsoft.WindowsAppRuntime\
+PublishFile $FullBuildOutput\WindowsAppRuntime_UniversalBGTaskDLL\Microsoft.Windows.ApplicationModel.Background.UniversalBGTask.dll $FullPublishDir\Microsoft.Windows.ApplicationModel.Background.UniversalBGTask\
```
AjitSurana, line 45: Curious to know What is UniversalBGTask.dll and can its name be just BackgroundTask.dll

dev/BackgroundTask/BackgroundTaskBuilder/BackgroundTaskBuilder.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+    winrt::BackgroundTaskRegistration BackgroundTaskBuilder::Register()
+    {
+        winrt::BackgroundTaskBuilder builder;
+        builder.Name(m_name);
+        builder.SetTrigger(m_trigger);
+        builder.TaskEntryPoint(L"Microsoft.Windows.ApplicationModel.Background.UniversalBGTask.Task");
+        winrt::BackgroundTaskRegistration taskRegistration = builder.Register();
+
+        winrt::ApplicationDataContainer localSettings = winrt::ApplicationData::Current().LocalSettings();
```
AjitSurana, line 40: The background task will work for both packaged and unpackaged apps? Please check ApplicationData::Current() works for un-packaged app.

godlytalias: As of now it will support only packaged apps, If we need to support unpackaged apps we can support them in later iterations

dev/WindowsAppRuntime_UniversalBGTaskDLL/WindowsAppRuntime_UniversalBGTaskDLL.vcxproj
```diff
@@ -0,0 +1,167 @@
...
+    <MinimalCoreWin>true</MinimalCoreWin>
+    <ProjectGuid>{0c4405e6-029e-4363-8273-a9e1fcf057de}</ProjectGuid>
+    <RootNamespace>Microsoft.Windows.ApplicationModel.Background.UniversalBGTask</RootNamespace>
+    <DefaultLanguage>en-US</DefaultLanguage>
+    <MinimumVisualStudioVersion>16.0</MinimumVisualStudioVersion>
+    <AppContainerApplication>false</AppContainerApplication>
+    <ApplicationTypeRevision>10.0</ApplicationTypeRevision>
+    <WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>
+    <DesktopCompatible>true</DesktopCompatible>
+    <UseWinUI>true</UseWinUI>
```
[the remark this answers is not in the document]
godlytalias: Good catch, these got added by default by Visual Studio. Removed it

dev/BackgroundTask/BackgroundTaskBuilder/BackgroundTaskBuilder.cpp
ghost1372: @godlytalias there is a merged pr for ApplicationData (GetForUnpackaged), however it seems that it marked as contract v2, so mybe you can consider it.
https://github.com/microsoft/WindowsAppSDK/pull/4057

godlytalias: Thanks @ghost1372 for pointing out that API. I will need to see the working of background task feature with unpackaged apps and then add the support in an upcoming PR

mominshaikhdevs:
1. Is this API **_"MediumIL apps"_** only? (meaning supports only full trust applications?)
2. will this API work in **_"HighIL apps"_**?
3. will this API work in **_"AppContainer apps"_**? ([#219 - AppContainer for Win32 Apps](https://github.com/microsoft/windowsappsdk/issues/219) + [Win32 App Isolation repo](https://github.com/microsoft/win32-app-isolation)?)

mominshaikhdevs:
It will be best to have a proper test for them. otherwise we will be getting weird cases, where some portion of these APIs work in fine in MediumIL and HighIL but not in AppContainer.

AjitSurana (approved):

dev/WindowsAppRuntime_UniversalBGTaskDLL/packages.config
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="Microsoft.Windows.CppWinRT" version="2.0.230706.1" targetFramework="native" />
+  <package id="Microsoft.Windows.ImplementationLibrary" version="1.0.220914.1" targetFramework="native" />
+  <package id="Microsoft.Windows.SDK.BuildTools" version="10.0.22621.756" targetFramework="native" />
+  <package id="Microsoft.WindowsAppSDK" version="1.5.240311000" targetFramework="native" />
```
Hemantxk, line 6: Should we take latest version here or 1.6?

godlytalias: Further works are required here to dynamically update the versions

dev/WindowsAppRuntime_UniversalBGTaskDLL/readme.txt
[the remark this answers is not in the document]
godlytalias: Sure, will remove this in further PR. Thanks

godlytalias:
As we are approaching to the Experimental 1 release snap date, merging the PR. Please review and I will incorporate the comments in upcoming PRs. Thanks

dev/WindowsAppRuntime_UniversalBGTaskDLL/Task.cpp
```diff
@@ -0,0 +1,29 @@
...
+    {
+        winrt::hstring lookupStr = winrt::to_hstring(taskInstance.Task().TaskId());
+        ApplicationDataContainer localSettings = ApplicationData::Current().LocalSettings();
+        auto values = localSettings.Values();
+        auto lookupobj = values.Lookup(lookupStr);
+        winrt::guid comClsId = winrt::unbox_value<winrt::guid>(lookupobj);
+
+        winrt::Windows::ApplicationModel::Background::IBackgroundTask bgTask = nullptr;
+        winrt::hresult hr = CoCreateInstance(comClsId, nullptr, CLSCTX_LOCAL_SERVER, IID_IBackgroundTask, reinterpret_cast<void**>(&bgTask));
+        bgTask.Run(taskInstance);
```
jlaanstra, line 27: This is gonna AV if CoCreateInstance failed. and bgTask is still nullptr.

godlytalias: Thanks @jlaanstra for the review, This method gets called from Broker Infrastructure and not handle any error results, Isn't it fine to throw exception here in that case? It will kill the backgroundtaskhost.exe process that gets created by broker infrastructure in that case.

jlaanstra: This won't throw an exception. You'll run into a nullref which will terminate. If you want to throw you should throw the hr.

godlytalias: Yes, It may not make much difference from BI side as the BI will just get to know that the backgroundtaskhost process got disconnected as it runs down. However I see that it is better to throw the right exception which may help during debugging.

PR #4838, opened 2024-10-31
Fix for the winmd references

godlytalias:
Currently the references for winmd that comes from WindowsAppSDK nuget package is getting affected by the references mentioned in targets file of WinUI. This change to de-couple the definition of references between both the packages. 

Related Bug: https://task.ms/54538137

### DIFF
--- a/build/NuSpecs/WindowsAppSDK-Nuget-Native.targets
+++ b/build/NuSpecs/WindowsAppSDK-Nuget-Native.targets
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
--- a/build/NuSpecs/WindowsAppSDK-Nuget-Native.targets
+++ b/build/NuSpecs/WindowsAppSDK-Nuget-Native.targets
@@ -16,12 +16,6 @@
     </Reference>
   </ItemGroup>
 
-  <ItemGroup Condition="'$(AppxPackage)' == 'true'">
-    <Reference Include="$(MSBuildThisFileDirectory)..\..\lib\uap10.0\Microsoft.Windows.ApplicationModel.DynamicDependency.winmd">
-      <Private>false</Private>
-    </Reference>
-  </ItemGroup>
-
   <ItemGroup>
     <Reference Include="$(MSBuildThisFileDirectory)..\..\lib\uap10.0\Microsoft.Windows.ApplicationModel.WindowsAppRuntime.winmd">
       <Private>false</Private>
--- a/build/NuSpecs/WindowsAppSDK-Nuget-Native.targets
+++ b/build/NuSpecs/WindowsAppSDK-Nuget-Native.targets
@@ -16,6 +16,12 @@
     </Reference>
   </ItemGroup>
 
+  <ItemGroup Condition="'$(AppxPackage)' == 'true'">
+    <Reference Include="$(MSBuildThisFileDirectory)..\..\lib\uap10.0\Microsoft.Windows.ApplicationModel.DynamicDependency.winmd">
+      <Private>false</Private>
+    </Reference>
+  </ItemGroup>
+
   <ItemGroup>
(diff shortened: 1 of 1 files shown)


build/NuSpecs/WindowsAppSDK-Nuget-Native.targets
```diff
@@ -90,6 +90,12 @@
     </Reference>
   </ItemGroup>
 
+  <ItemGroup>
+    <Reference Include="$(MSBuildThisFileDirectory)..\..\lib\uap10.0\Microsoft.Windows.*.winmd">
```
JesseCol, line 94: Won't this make duplicate entries for the Microsoft.Windows.*.winmds we already added above?

JesseCol: I think we specifically want to add:
Microsoft.Windows.AppNotifications.winmd
Microsoft.Windows.Globalization.winmd
Microsoft.Windows.Management.Deployment.winmd

JesseCol (changes requested):

build/NuSpecs/WindowsAppSDK-Nuget-Native.targets
godlytalias: @JesseCol As I understand, as long as the `Private` property is not conflicting, it won't make issues. I applied it as wildcard to avoid any unforeseen issues because we were having a wildcard usage in WinUI that was getting applied to all winmd. But we can go with specific winmd as well.

JesseCol: Adding @kythant / @Scottj1s -- do you know who might want to review a change like this?

I'm afraid I don't have enough context to know if this change will be safe.  My advice is to review the binlogs for a WinAppSDK app and do a before/after comparison to see what changed -- and I'd expect we'd want to change as little as possible.

Scottj1s: I thought the issue was with WinUI eagerly referencing winmds that didn't belong to it - is this a workaround for that?  Shouldn't we fix that properly?

JesseCol: @Scottj1s I should have explained better.  Right now, the targets file in WinUI is pulling in *.winmd, and some of the winmds here are getting pulled in by that targets file.  I'm suggesting WinUI targets files only reference Microsoft.UI*.winmd, and this file speaks for the winmds in this repo.

Scottj1s: my apologies - wasn't paying close attention.  yes, I agree with that strategy - fewer and/or most narrowly scoped the wildcards, the better.

build/NuSpecs/WindowsAppSDK-Nuget-Native.targets
```diff
@@ -16,6 +16,12 @@
     </Reference>
   </ItemGroup>
 
+  <ItemGroup Condition="'$(AppxPackage)' == 'true'">
```
JesseCol, line 19: It looks like the owner of DynamicDependency.winmd wanted this to be included only when AppxPackage=='false'.  @DrusTheAxe, can you confirm that?  Thanks!

DrusTheAxe: Confirmed

DynamicDependency API is only supported in unpackaged processes

godlytalias: Yes it looks like that, however I kept it for being on safe side by keeping existing behavior.

godlytalias: Thanks @DrusTheAxe, Removed that entry

JesseCol (approved):

Scottj1s (approved):

build/NuSpecs/WindowsAppSDK-Nuget-Native.targets
DrusTheAxe: DynDep explicitly doesn't support Packaged apps. All the APIs in [dev/DynamicDependency/API/MsixDynamicDependency.cpp](https://github.com/microsoft/WindowsAppSDK/blob/main/dev/DynamicDependency/API/MsixDynamicDependency.cpp) have

```
    // WinAppSDK's Dynamic Dependencies requires a non-packaged process
    RETURN_HR_IF(HRESULT_FROM_WIN32(ERROR_NOT_SUPPORTED), AppModel::Identity::IsPackagedProcess());
```

PR #4845, opened 2024-11-03
EnsurePackage*Async() handling options.RegisterNewerIfAvailable(true). RegisterPackageByPackageFull/FamilyNameAsync crash. Test + Warning fixes

DrusTheAxe:
EnsurePackage*Async() not handling options.RegisterNewerIfAvailable(true)

`RegisterPackageByPackageFullNameAsync()` and `RegisterPackageByPackageFullNameAsync()` take their target parameter as `const hstring&` resulting in the _reference_ captured by their implementations' `co_await` and not the actual value. The caller deallocates the memory before `Register...` is done with it potentially leading to BadMojo(TM). Tests don't notice it as they're too simple to notice this use-after-free -- the memory may be deallocated but the memory's not overwritten (yet) in simple loads (like the tests). Larger scale (real world) use with more diverse memory patterns notice the problem (usually as a crash).

ApplicationDataTests_Elevated test cases are blocked state due to user context vs environmental state vs access control (ACLs whee!). Changed RunAs and associated test process setup to play nice together.

Fixed some PackageManagerTests that had incomplete state setup potentially causing false errors if a past previous test failed.

Fixed 400+ build warnings (mostly CS8305 due to [Experimental] attribute, almost all from OAuth generated C#/WinRT)

https://task.ms/54835036
https://task.ms/54858998
https://task.ms/54884960


wcheng-msft (approved):
:shipit:

PR #4852, opened 2024-11-05
Adding reference for DynamicDependency winmd

godlytalias:
Currently the references for winmd that comes from WindowsAppSDK nuget package is getting affected by the references mentioned in targets file of WinUI. This change to de-couple the definition of references between both the packages.

Related Bug: https://task.ms/54538137

### DIFF
--- a/build/NuSpecs/WindowsAppSDK-Nuget-Native.targets
+++ b/build/NuSpecs/WindowsAppSDK-Nuget-Native.targets
@@ -16,6 +16,12 @@
     </Reference>
   </ItemGroup>
 
+  <ItemGroup Condition="'$(AppxPackage)' == 'true'">
+    <Reference Include="$(MSBuildThisFileDirectory)..\..\lib\uap10.0\Microsoft.Windows.ApplicationModel.DynamicDependency.winmd">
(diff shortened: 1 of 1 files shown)


build/NuSpecs/WindowsAppSDK-Nuget-Native.targets
```diff
@@ -16,6 +16,12 @@
     </Reference>
   </ItemGroup>
 
+  <ItemGroup Condition="'$(AppxPackage)' == 'true'">
```
JesseCol, line 19: I'm not sure whey we don't just remove the "Condition" from the above entry for Microsoft.Windows.ApplicationModel.DynamicDependency.winmd?

JesseCol (approved):

build/NuSpecs/WindowsAppSDK-Nuget-Native.targets
godlytalias: To avoid any further unforeseen issues in unconditionally adding the 'implementation' parameter as well in the reference I did like this. WinUI was just adding 'private' for the references with its wildcard, so just replicating the same here.

PR #4855, opened 2024-11-07
Automatically generate password to avoid using Convert-To-SecureString

kythant:

WARNING:


tools/DevCheck/DevCheck.ps1
```diff
@@ -771,10 +771,10 @@ function Repair-DevTestPfx
 
     # -CertPassword <password> is a required parameter for this work
     $password = ''
-    $password_plaintext = $null
+    $passwordText = $null
```
codendone, line 774: >$passwordText = $null [](http://example.com/codeflow?start=4&length=21)

Unused?

kythant: Oh yes. Thanks for catching that.

PR #4866, opened 2024-11-08
DevCheck -CheckDependencies exitcode. Fix dependencies

DrusTheAxe:
`DevCheck -CheckDependencies` spots package dependencies that don't match the sanctioned list in `eng\Verison.*.xml` but it didn't set the exitcode if violators found so builds would detect violators, printf some message in a wall of text no one looks at and continue on.

Fixed to set exitcode so the build pipeline should abort if problem(s) are found

This turned up a few problems in the repo. Fixed (else this DevCheck fix couldn't be checked in)


kythant (approved):

godlytalias (dismissed):

Scottj1s (dismissed):

DrusTheAxe:
I think I understand why the build breaks. Fix is simple or less so, but either way invasive. Discussing with @godlytalias  to understand which way I should pivot. Update RSN

smnadircw:
yea

DrusTheAxe:
Obsolete as these changes have been incorporated thru other PRs

PR #4873, opened 2024-11-12
Fix incorrect cch param

Scottj1s:
PathCchCombineEx's cchPathOut is a character count, not a byte count.  Result was an occasional crash when PathCchCombineEx would use what it thought was available storage as scratch space, causing a buffer overrun.


Scottj1s:
internal tracking issue:  https://task.ms/55029408

JesseCol (approved):

PR #4895, opened 2024-11-18
Updating CameraCaptureUI Spec Doc

Saharsh979:
This PR includes the following changes:
1) The change updates the CameraCaptureUI specifications doc to include important notes on usage for PhotoSettings and VideoSettings.
2) Added the CameraCaptureUIContract in the WinRT contracts as well.
3) A small update by removing the marshalling=none and threading=STA attributes. These settings were relevant to the UWP version of CameraCapture, but they are no longer applicable in the context of WinAppSDK.

### DIFF
--- a/specs/CameraCaptureUI/CameraCaptureUI.md
+++ b/specs/CameraCaptureUI/CameraCaptureUI.md
@@ -190,6 +190,11 @@ Provides settings for capturing photos with CameraCaptureUI.
 The settings include aspect ratio, image size, format, resolution, 
 and whether or not cropping is allowed by the user interface (UI).
 
+**Important Note**:  
+- Don't specify both size and aspect ratio in the `PhotoSettings`, doing so will cause an invalid argument exception.  
+- `PhotoSettings` can't have a ratio or size specified with cropping disabled.
+
+
 ## CameraCaptureUIPhotoCaptureSettings Properties
 
 | Name | Description | Value |
@@ -206,6 +211,11 @@ and whether or not cropping is allowed by the user interface (UI).
 ## CameraCaptureUIVideoCaptureSettings Class
 Provides settings for capturing videos. 
 The settings include format, maximum resolution, maximum duration, and whether or not to allow trimming.
+
+**Important Note**:  
+- `MaxDurationInSeconds` must be a valid value (i.e., the duration should be in the range of 0 to `UINT32_MAX`).
+- `MaxDurationInSeconds` cannot be set if `AllowTrimming` is false.
+
 
 ## CameraCaptureUIVideoCaptureSettings Properties
 | Name | Description | Value |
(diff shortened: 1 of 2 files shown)


specs/CameraCaptureUI/CameraCaptureUI.md
```diff
@@ -190,6 +190,11 @@ Provides settings for capturing photos with CameraCaptureUI.
 The settings include aspect ratio, image size, format, resolution, 
 and whether or not cropping is allowed by the user interface (UI).
 
+**Important Note**:  
+- Don't specify both size and aspect ratio in the `PhotoSettings`, doing so will cause an invalid argument exception.  
+- `PhotoSettings` can't have a ratio or size specified with cropping disabled.
```
godlytalias, line 195: Should we specify on the behavior if ratio or size is specified when cropping is disabled? Whether API will throw exception?

specs/CameraCaptureUI/CameraCaptureUI.md
```diff
@@ -207,6 +212,11 @@ and whether or not cropping is allowed by the user interface (UI).
 Provides settings for capturing videos. 
 The settings include format, maximum resolution, maximum duration, and whether or not to allow trimming.
 
+**Important Note**:  
+- `MaxDurationInSeconds` must be a valid value (i.e., the duration should be in the range of 0 to `UINT32_MAX`).
+- `MaxDurationInSeconds` cannot be set if `AllowTrimming` is false.
```
godlytalias, line 217: Here also may be good to specify behavior if AllowTrimming is false

godlytalias (approved):

specs/CameraCaptureUI/CameraCaptureUI.md
Saharsh979: Yeah that makes sense, updated!

specs/CameraCaptureUI/CameraCaptureUI.md
Saharsh979: Yes, updated!

specs/CameraCaptureUI/CameraCaptureUI.md
```diff
@@ -190,6 +190,11 @@ Provides settings for capturing photos with CameraCaptureUI.
 The settings include aspect ratio, image size, format, resolution, 
 and whether or not cropping is allowed by the user interface (UI).
 
+**Important Note**:  
+- Don't specify both size and aspect ratio in the `PhotoSettings`, doing so will cause an invalid argument exception.  
+- `PhotoSettings` can't have a ratio or size specified with cropping disabled. Attempting to do so will result in an invalid argument exception.
```
Hemantxk, line 195: Related - should this statement also specify that by default AllowCropping is true? Similar for AllowTrimming?

Hemantxk (approved):

specs/CameraCaptureUI/CameraCaptureUI.md
Saharsh979: I think it’s a fair point to bring up, adding that clarification could help prevent confusion if the developer is unaware of the defaults.

Hemantxk (approved):

PR #4926, opened 2024-11-29
Local Badge Notification code changes

SatwikKrSharma:
Issue - https://github.com/microsoft/WindowsAppSDK/issues/138
API Spec - https://github.com/microsoft/WindowsAppSDK/pull/4823/files#diff-17dae42945b77909c6b45f1e7635b67969b8656882ccca5d20c56fd9ab653d6a

WARNING:


dev/Notifications/BaseNotifications/BaseNotification.h
```diff
@@ -0,0 +1,52 @@
...
+
+        BaseNotification(winrt::hstring const& payload);
+        winrt::hstring Tag() const noexcept;
+        void Tag(winrt::hstring const& value);
+        winrt::hstring Group() const noexcept;
+        void Group(winrt::hstring const& value);
+        uint32_t Id() const noexcept;
+        void Id(uint32_t id);
+        winrt::hstring Payload() const noexcept;
+        winrt::Microsoft::Windows::AppNotifications::AppNotificationProgressData Progress() const noexcept;
```
satkh, line 24: These properties are not relevant for raw or badge, should not be part of base notifications,
similar comment for other properties in this class.

satkh:
i think if we are bringing in the hierarchy of base notification -> badge notification
We should also add other classes for the notification types that we currently support (Toast, raw notifications), Or will be follow up PR ?

dev/Notifications/BadgeNotifications/BadgeNotificationManager.h
```diff
@@ -0,0 +1,58 @@
...
+
+namespace winrt::Microsoft::Windows::BadgeNotifications::implementation
+{
+    struct BadgeNotificationManager : BadgeNotificationManagerT<BadgeNotificationManager>, public ::Microsoft::Windows::BaseNotifications::BaseNotificationManager
+    {
+        BadgeNotificationManager();
+
+        static winrt::Microsoft::Windows::BadgeNotifications::BadgeNotificationManager Current();
+
+        void SetBadgeAsCount(_In_ uint32_t notificationCount);
```
satkh, line 17: nit : 
void SetBadgeAsCount(_In_ uint32_t notificationCount, _In_ winrt::Windows::Foundation::DateTime expiration = nullptr);  ?

dev/Common/TerminalVelocityFeatures-BadgeNotifications.xml
```diff
@@ -0,0 +1,23 @@
...
+
+     Licensed under the MIT License. See LICENSE in the project root for license information.
+
+-->
+
+  <!-- See docs/Coding-Guidelines/TerminalVelocity.md for more info.  -->
+
+  <!-- Enabled variants -->
+  <feature>
+    <name>Feature_BadgeNotifications</name>
```
satkh, line 14: nit : Feature_LocalBadgeNotifications ?

dev/Common/TerminalVelocityFeatures-BadgeNotifications.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0"?>
+<features xmlns="http://microsoft.com/windowsappsdk/TerminalVelocity/20210729/TerminalVelocityFeatures.xsd">
+
+  <!-- Copyright (c) Microsoft Corporation. All rights reserved.
+
+     Licensed under the MIT License. See LICENSE in the project root for license information.
```
satkh, line 6: move the copy right info to the top of the document

dev/Notifications/BaseNotifications/BaseNotification.h
SatwikKrSharma: We pass notification detail to FrameworkUdk as NotificationProperties. So NotificationProperties act as a adapter between WindowsAppSdk and FrameworUdk. And the INotificationProperties1 contains the Progress data so to use the existing functionality we should keep the progess data in the base class. Every other properties need to be part of base notifications because in WIndowsSdk also these are part of the base notifications

dev/Common/TerminalVelocityFeatures-BadgeNotifications.xml
SatwikKrSharma: This is the convention used for all the terminal velocity files in WindowsAppSdk

dev/Notifications/BaseNotifications/BaseNotification.h
satkh: I think, this can be managed through the specific notification classes like "ToastNotification" , "BadgeNotification", "RawNotification" ... The new Base Notification class need not be copy of AppNotifications class that we have currently, if we are modularizing this, specific properties should go to the relevant classes.

dev/Notifications/BadgeNotifications/BadgeNotificationManager.cpp
```diff
@@ -0,0 +1,157 @@
+// Copyright (c) Microsoft Corporation and Contributors.
+// Licensed under the MIT License.
+
+#include "pch.h"
+#include "BadgeNotificationManager.h"
+#include "Microsoft.Windows.BadgeNotifications.BadgeNotificationManager.g.cpp"
+#include "externs.h"
+#include <frameworkudk/toastnotifications.h>
```
satkh, line 8: BadgeNotificationManager should only delegate the call to BadgeNotification class.
the framework udk handling should go happen on BadgeNotification class.
BadgeNotificationManager, should only call the methods on BadgeNotification, to the functionality done.

SatwikKrSharma: BadgeNotification class is just a struct that carries information about the notification. It should not have any functionality. So all the operation happens on manager level and all the information is present in badgenotification class

dev/Notifications/BaseNotifications/BaseNotification.h
SatwikKrSharma: Yes i agree but the way INotificationProperties was constructed these are all the basic properties that needs to be here. Let's discuss this offline

dev/Notifications/BadgeNotifications/BadgeNotificationManager.cpp
```diff
@@ -0,0 +1,157 @@
...
+            {winrt::BadgeNotificationGlyph::Busy, L"busy"},
+            {winrt::BadgeNotificationGlyph::Error, L"error"},
+            {winrt::BadgeNotificationGlyph::NewMessage, L"newMessage"},
+            {winrt::BadgeNotificationGlyph::Paused, L"paused"},
+            {winrt::BadgeNotificationGlyph::Playing, L"playing"},
+            {winrt::BadgeNotificationGlyph::Unavailable, L"unavailable"}
+        };
+
+        *glyphString = enumMapping.at(glyphValue);
+        return;
```
satkh, line 135: nit : not required.

dev/Notifications/BadgeNotifications/BadgeNotificationManager.cpp
```diff
@@ -0,0 +1,157 @@
...
+            ::Microsoft::RoApi::ClearRoActivateInstanceCache();
+
+            // Store the BadgeNotificationManager in the COM static store
+            auto badgeNotificationManager{ winrt::make<BadgeNotificationManager>() };
+            appProperties.Insert(STORED_BADGENOTIFICATION_MANAGER_KEY, badgeNotificationManager);
+            return badgeNotificationManager;
+        }
+    }
+
+    void BadgeNotificationManager::SetBadgeAsCount(uint32_t notificationCount)
```
satkh, line 65: verify if valid count? handle cases when 0 ..

dev/Notifications/BadgeNotifications/BadgeNotificationManager.cpp
```diff
@@ -0,0 +1,157 @@
...
+    {
+        THROW_HR_IF_MSG(E_ILLEGAL_METHOD_CALL, !AppModel::Identity::IsPackagedProcess(), "Not applicable for unpackaged applications");
+
+        SetBadge(std::to_wstring(notificationCount), nullptr);
+        return;
+    }
+
+    void BadgeNotificationManager::SetBadgeAsCount(uint32_t notificationCount, winrt::DateTime expiration)
```
satkh, line 73: validate these arguments

dev/Notifications/BadgeNotifications/BadgeNotificationManager.cpp
```diff
@@ -0,0 +1,157 @@
...
+        SetBadge(glyphValueString, nullptr);
+        return;
+    }
+
+    void BadgeNotificationManager::SetBadgeAsGlyph(winrt::BadgeNotificationGlyph glyphValue, winrt::DateTime expiration)
+    {
+        THROW_HR_IF_MSG(E_ILLEGAL_METHOD_CALL, !AppModel::Identity::IsPackagedProcess(), "Not applicable for unpackaged applications");
+
+        PCWSTR glyphValueString;
+        GetBadgeNotificationGlyphToString(glyphValue, &glyphValueString);
```
satkh, line 96: handle winrt::BadgeNotificationGlyph::None case, call clear method ?

dev/Notifications/BadgeNotifications/BadgeNotificationManager.cpp
```diff
@@ -0,0 +1,157 @@
...
+        };
+
+        *glyphString = enumMapping.at(glyphValue);
+        return;
+    }
+
+    void BadgeNotificationManager::SetBadge(_In_ const std::wstring& value, _In_opt_ const winrt::Windows::Foundation::DateTime* expiration)
+    {
+        auto xmlResult = wil::str_printf<std::wstring>(L"<badge value='%ls'/>", value.c_str());
+        ::Microsoft::Windows::BadgeNotifications::BadgeNotification badgeNotifications(xmlResult.c_str());
```
satkh, line 141: badgeNotification

dev/Notifications/BadgeNotifications/BadgeNotificationManager.h
```diff
@@ -0,0 +1,58 @@
...
+#include "externs.h"
+#include "BaseNotificationManager.h"
+
+namespace winrt::Microsoft::Windows::BadgeNotifications::implementation
+{
+    struct BadgeNotificationManager : BadgeNotificationManagerT<BadgeNotificationManager>, public ::Microsoft::Windows::BaseNotifications::BaseNotificationManager
+    {
+        BadgeNotificationManager();
+
+        static winrt::Microsoft::Windows::BadgeNotifications::BadgeNotificationManager Current();
```
satkh, line 15: whats the purpose of this method?

dev/Notifications/BadgeNotifications/BadgeNotifications.idl
```diff
@@ -0,0 +1,52 @@
...
+    };
+
+    // The manager class which encompasses all Badge Notification API Functionality
+    [contract(BadgeNotificationsContract, 1), feature(Feature_BadgeNotifications)]
+    runtimeclass BadgeNotificationManager
+    {
+        // Gets a Default instance of a BadgeNotificationManager
+        static BadgeNotificationManager Current{ get; };
+
+        // Updates the badge on the app's tile to display a numeric value.
```
satkh, line 37: app's icon on task bar ?

dev/Notifications/BadgeNotifications/BadgeNotifications.idl
```diff
@@ -0,0 +1,52 @@
...
+        void SetBadgeAsCount(UInt32 notificationCount);
+
+        // Updates the badge on the app's tile to display a numeric value with an expiration time.
+        void SetBadgeAsCount(UInt32 notificationCount, Windows.Foundation.DateTime expiration);
+
+        // Updates the badge on the app's tile to display a glyph.
```
satkh, line 43: check if this can be merged into one method, using https://learn.microsoft.com/en-us/windows/win32/midl/defaultvalue

dev/Notifications/BadgeNotifications/BadgeNotificationManager.cpp
SatwikKrSharma: yes, when the count is 0, no notification will be displayed. Behavior is same as WindowsSdk

dev/Notifications/BadgeNotifications/BadgeNotificationManager.cpp
SatwikKrSharma: When we give None, shell automatically does not display the notification. Same behavior as WindowsSdk

dev/Notifications/BadgeNotifications/BadgeNotificationManager.h
SatwikKrSharma: To get the current instance of BadgeNotificationManager. Only 1 instance of manager will be created

dev/Notifications/BadgeNotifications/BadgeNotificationManager.cpp
satkh: in that case you can just return when 0 .

dev/Notifications/BadgeNotifications/BadgeNotificationManager.cpp
satkh: As per spec .. when this function is called with None, the notifications should be cleared. 
consider calling clear() in this case.

dev/Notifications/BaseNotifications/BaseNotificationManager.cpp
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    void BaseNotificationManager::RemoveAllNotification(ABI::Microsoft::Internal::ToastNotifications::NotificationType notificationType)
+    {
+        /*if (!IsSupported())
+        {
+            co_return;
+        }*/
+
+        //THROW_IF_FAILED(WnpNotifications_RemoveAllNotificationsForAppOfType(m_appId.c_str(), notificationType));
```
jonwis, line 49: Don't merge commented-out code, please. Delete or uncomment.

dev/Notifications/BadgeNotifications/BadgeNotificationManager.cpp
```diff
@@ -0,0 +1,157 @@
...
+        BaseNotifications::BaseNotificationManager::RemoveAllNotification(ToastABI::NotificationType::NotificationType_Badge);
+
+        logTelemetry.Stop();
+
+        return;
+    }
+
+    void BadgeNotificationManager::GetBadgeNotificationGlyphToString(_In_ winrt::BadgeNotificationGlyph glyphValue, _Out_ PCWSTR* glyphString)
+    {
+        static const std::unordered_map<winrt::BadgeNotificationGlyph, PCWSTR> enumMapping = {
```
jonwis, line 118: Don't build a static map like this - use an array, with lookup.

For example:

```c++
constexpr static std::pair<winrt::BadgeNotificationGlyph, PCWSTR> mapping[] = {
    { winrt::BadgeNotificationGlyph::None, L"none" },
    { winrt::BadgeNotificationGlyph::Activity, L"activity" },
    // ...
};
auto i = std::find_if(std::begin(mapping), std::end(mapping), [&](auto const& i) { return i.first == glyphValue; });
if (i != std::end(mapping)) {
    *glyphString = i->second;
}
```

That's compile-time initialized.  The linear scan of an array, amortized, along with the reduction in binary size, vastly outweighs the "O(1) lookup" promise.

Or use a `switch(...)` statement.  Just don't use a `static std::[unordered_]map`.

dev/Notifications/BadgeNotifications/BadgeNotification.cpp
```diff
@@ -0,0 +1,27 @@
...
+using namespace Microsoft::Windows::BaseNotifications;
+
+namespace ToastABI
+{
+    using namespace ::ABI::Microsoft::Internal::ToastNotifications;
+}
+
+namespace Microsoft::Windows::BadgeNotifications {
+    // Default constructor
+    BadgeNotification::BadgeNotification() : BaseNotification()
```
jonwis, line 17: Can the `Base-` type just be merged into this type, since the API review eliminated this type being handed out?

dev/Notifications/BadgeNotifications/BadgeNotificationManager.cpp
```diff
@@ -0,0 +1,157 @@
...
+            g_telemetryHelper,
+            m_appId,
+            badgeNotifications.Payload().c_str()) };
+
+        if (expiration != nullptr)
+        {
+            badgeNotifications.Expiration(*expiration);
+        }
+
+        BaseNotifications::BaseNotificationManager::Show(badgeNotifications);
```
jonwis, line 153: Feels like this interplay of the manager, base, and notification types needs to be simplified since the BadgeNotification types are no longer returned to anybody. Maybe this "show" method should also just take `xmlResult.c_str()` + `expiration ? *expiration : {}` ?

jonwis (commented):
Looking good. I think you can remove the BadgeNotification & BaseNotification types, though? If you need to update NotificationProperties to take a struct of parameters, or have a new ctor that knows "I'm being constructed with all defaults from the set badge notification path" then that's fine.

dev/Notifications/BadgeNotifications/BadgeNotificationManager.h
SatwikKrSharma: We have followed the same patter as AppNotificationManager and PushNotificationManager. Reason is - 
The reason for the AppNotificationManager having a Default API is that it registers with the PushNotificationsLongRunningTask (LRP) which is a singleton process routing ToastNotifications to WinAppSDK consumers. Having multiple AppNotificationManagers would cause multiple registrations with the LRP and would only route ToastNotifications to the most recently instantiated manager.
AppNotificationManager Foo registers with LRP
LRP holds mapping of processName->Foo
AppNotificationManager Bar registers with LRP
LRP overwrites Foo with processName->Bar
Foo is a stale manager doing nothing   
           Our implementation depends on the LRP singleton for managing routing notifications to registered handlers       (App/PushNotificationManager)

dev/Notifications/BadgeNotifications/BadgeNotifications.idl
SatwikKrSharma: We are following the API Spec suggestion

dev/Notifications/BadgeNotifications/BadgeNotificationManager.cpp
SatwikKrSharma: We don't need to validate these arguments because shell takes care of this. For e.g if the count is 0 no badge gets displayed if it is >100 only 99+ gets displayed. This behavior is as same as in current WindowsSdk implementation

dev/Common/TerminalVelocityFeatures-BadgeNotifications.xml
SatwikKrSharma: badge notification alone is considered local for push it will be completely different, and changes will be part of pushnotification directory rather than here

dev/Notifications/BaseNotifications/BaseNotification.h
SatwikKrSharma: changed done, Progress and conference config removed from base class

dev/Notifications/BadgeNotifications/BadgeNotificationManager.cpp
SatwikKrSharma: done

dev/Notifications/BaseNotifications/BaseNotificationManager.cpp
SatwikKrSharma: removed baseNotificationManager

dev/Notifications/BadgeNotifications/BadgeNotificationManager.cpp
SatwikKrSharma: changed to array

dev/Notifications/BadgeNotifications/BadgeNotificationManager.cpp
SatwikKrSharma: removed baseNotificationManager but keeping BaseNotification because it simplifies the already existing Notification code. E.g - AppNotification now inherits BaseNotification and BadgeNotification also inherits BaseNotification. Otherwise duplicity of code would have happened. Also this makes both AppNotification and BadgeNotification extensible and easy to add new attributes

dev/Notifications/BadgeNotifications/BadgeNotification.cpp
SatwikKrSharma: emoved baseNotificationManager but keeping BaseNotification because it simplifies the already existing Notification code. E.g - AppNotification now inherits BaseNotification and BadgeNotification also inherits BaseNotification. Otherwise duplicity of code would have happened. Also this makes both AppNotification and BadgeNotification extensible and easy to add new attributes

dev/AppNotifications/NotificationProperties.cpp
```diff
@@ -57,6 +57,36 @@ NotificationProperties::NotificationProperties(winrt::AppNotification const& toa
...
+
+    m_notificationType = ToastABI::NotificationType::NotificationType_Toast;
+}
+
+NotificationProperties::NotificationProperties(Microsoft::Windows::BaseNotifications::BaseNotification const& baseNotification)
+{
+    // Extract payload and convert it from XML to a byte array
+    auto payloadAsSimpleString = Helpers::WideStringToUtf8String(baseNotification.Payload());
+
+    m_payload = wil::unique_cotaskmem_array_ptr<byte>(static_cast<byte*>(CoTaskMemAlloc(payloadAsSimpleString.size())), payloadAsSimpleString.size());
```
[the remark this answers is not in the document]
SatwikKrSharma: code changes done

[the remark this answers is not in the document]
SatwikKrSharma: Hi Jon, I have kept the original format because, if the conversion happens in the constructor, we are sure that it will happen only once, but if we delay the conversion, it will happen twice one in get_Payload and one in get_PayloadSize which increases the cost. So, keeping the code as it is to handle the scenario

jonwis (approved):

PR #4942, opened 2024-12-08
[DRAFT PR] Storage Pickers Prototype PR for Dicussion

Hong-Xiang:
*NOTE* This is a draft PR for discussion, not for code review for production


prototype-workingdir/Microsoft.Storage.Pickers/FileOpenPicker.cpp
```diff
@@ -0,0 +1,154 @@
...
+    }
+    hstring FileOpenPicker::SettingsIdentifier()
+    {
+        return m_settingsIdentifier;
+    }
+    void FileOpenPicker::SettingsIdentifier(hstring const& value)
+    {
+        m_settingsIdentifier = value;
+    }
+    winrt::Microsoft::Storage::Pickers::PickerLocationId FileOpenPicker::SuggestedStartLocation()
```
whiskhub, line 33: As a WinUI 3 developer, I would appreciate it if we could provide any file path here. I understand that the goal is to keep it similar to the WinRT API, but the COM file picker doesn't have this limitation.   
Maybe by adding an additional `SuggestedStartLocationDirectory` that takes precedence if not set to null.

sungaila:
A file/folder picker that works for both packaged and unpackaged apps, evelated as admin or not, and inside an AppContainer or not ... that would be a dream come true.

Finally a chance to fix the current mess in the Windows App SDK.

prototype-workingdir/Microsoft.Storage.Pickers/FileOpenPicker.cpp
Hong-Xiang: @whiskhub  thank you for your suggestions. `SuggestedStartLocation` is one of the features currently under discussion. While the final API design, behavior, and release have not been finalized yet, we are actively working on trying to add it.

eduardobragaxz:
Why return something like PickFileResult instead of StorageFile? StorageFile is a lot more conveniant. If all you need is the path, the StorageFile already contains that, but if you need other properties you then have to do StorageFile.GetFileFromPathAsync. It can be a little cumbersome, specially when calling fileOpenPicker.PickMultipleFilesAsync(). It returns an IReadOnlyList<PickFileResult> instead of an IReadOnlyList<StorageFile>.

mominshaikhdevs:
> ...It returns an IReadOnlyList instead of an IReadOnlyList.

typo?

eduardobragaxz:
> > ...It returns an IReadOnlyList instead of an IReadOnlyList.
> 
> typo?
> 
> also,
> 
> * https://github.com/microsoft/WindowsAppSDK/discussions/5255.
> 
> 

GitHub eats the brackets 😩

I meant it returns an IReadOnlyList of PickFileResult instead of an IReadOnlyList of StorageFile.

Hong-Xiang:
Close this outdated PR since #5240 already merged.

PR #4965, opened 2024-12-12
Rewrote MrmGetFilePathFromName for clarity/robustness and expanded search to include exe path even with a baseDir

Scottj1s:
This work supports A/B runtime selection


dev/MRTCore/mrt/Core/src/MRM.cpp
```diff
@@ -969,96 +969,119 @@ STDAPI MrmGetFilePathFromName(_In_opt_ PCWSTR filename, _Outptr_ PWSTR* filePath
...
+        {
+            // Search, given FileName
+            case SearchPass::exeDirForFileName:
+                searchDir = exeDir.get();
+                searchDirCount = exeDirCount;
+                searchFilename = filename;
+                break;
+            case SearchPass::ParentPathForFileName:
+                // move to parent folder of previous search
+                RETURN_IF_FAILED(PathCchRemoveFileSpec(searchDir, searchDirCount));
```
JesseCol, line 1033: Seems like searchDir and searchDir count will always be zero here since they are both local to the loop body scope?

JesseCol (approved):

PR #4966, opened 2024-12-13
App-configurable compatibility options spec: RuntimeCompatibilityOptions

codendone:
Spec for app-configurable compatibility options, which can be set via `RuntimeCompatibilityOptions` APIs or via project properties.

### DIFF
--- a/specs/Compatibility/CompatibilityOptions.md
+++ b/specs/Compatibility/CompatibilityOptions.md
@@ -1,0 +1,220 @@
+App-configurable compatibility options
+===
+
+# Background
+
+_This spec adds a CompatibilityOptions class to control behavior of servicing changes._
+
+Apps using Windows App SDK have two choices today: use Windows App SDK as an
+automatically-updating framework package or use Windows App SDK in self-contained mode with no
+automatic updates. Apps which use the framework package automatically get fixes, which is great,
+but they have a chance of being broken by a bug or subtle behavior change in a servicing update.
+Some apps choose self-contained mode to avoid that risk of an automatic servicing update breaking
+them. But even when apps choose self-contained mode, they may find that they can't update to a
+newer servicing release due to a bug or behavior change. Also, self-contained apps can't use
+features which require shared services, such as some Notification features delivered via the
+Singleton package.
+
+App-configurable compatibility is intended to prevent these problems, enabling apps to choose to
+use the framework package with confidence they won't be broken. Also, apps which find a
+compatibility problem in a servicing release will still be able to move forward by temporarily
+disabling problematic changes. Even apps using self-contained can update to a new package with
+confidence that app-configurable compatibility will ensure a successful update.
+
+A new `CompatibilityOptions` class has APIs to control the behavior of servicing changes.
+There are also properties which can be set in the app's project file to automatically use the new
+APIs with the specified values.
+
+# API Pages
+
+## CompatibilityOptions class
+
+Configures which changes in Windows App SDK servicing releases are enabled. By default, all
+changes are enabled, but `CompatibilityOptions` can be used to lock the runtime behavior to a
+specified patch version or to disable specific changes:
+
+1. **Choose the patch version:** You can specify which servicing patch version's behavior you
+   want to use. For example, your app can specify that it wants the 1.7.2 version behavior, which
+   will have Windows App SDK run in that mode even if 1.7.3 or later is installed. This capability
+   allows you to control when your app gets new fixes or behavior changes, even when not using
+   self-contained mode.
+2. **Temporarily disable specific changes:** If your app encounters a problem with a specific
+   change in a servicing update, you can disable just that change while still benefiting from the
+   other changes or features in that update. (All changes are enabled by default for the patch
+   version in use.) Disabling a change is a temporary measure which gives time for a fix to be
+   released in a future Windows App SDK update or for you to implement an update in your app.
+
+Here's an example of using `CompatibilityOptions` to specify a patch version and disable a
+specific change:
+
+```c#
+void ApplyCompatibilityOptions() 
+{
+    var compatibilityOptions = new CompatibilityOptions();
+    compatibilityOptions.PatchMode1 = new WindowsAppRuntimeVersion(1,7,3);
+    compatibilityOptions.PatchMode2 = new WindowsAppRuntimeVersion(1,8,2);
+    compatibilityOptions.DisabledChanges.Add(CompatibilityChange.HypotheticalChange);
+    compatibilityOptions.Apply();
+}
+```
+
+Note that CompatibilityOptions must be applied early in the process before any other Windows App
+SDK APIs are called, or right after initializing the Windows App Runtime.
+
+### Specifying CompatibilityOptions in the app's project file
+
+You can also specify the patch version and disabled changes in the app's project file instead of
+directly calling the CompatibilityOptions APIs. This approach has the advantage of ensuring the
+options are applied early at the proper timing. Here's an example of how to specify the patch
+version and disabled changes in the project file (such as `.csproj` or `.vcxproj`):
+
+```xml
+  <PropertyGroup>
+    <WindowsAppSDKRuntimePatchMode>1.7.3</WindowsAppSDKRuntimePatchMode>
+    <WindowsAppSDKRuntimePatchMode2>1.8.2</WindowsAppSDKRuntimePatchMode2>
+    <WindowsAppSDKDisabledChanges>HypotheticalChange, OtherHypotheticalChange</WindowsAppSDKDisabledChanges>
+  </PropertyGroup>
+```
+
+The `WindowsAppSDKDisabledChanges` property is a comma-separated list of `CompatibilityChange`
+values to disable.
+
+### Behavior with no PatchMode specified
+
+If no `PatchMode1` or `PatchMode2` is specified, or if neither value matches the major.minor
+version of the runtime being used, then the runtime will use the latest patch version. In other
+words, the runtime will run with all servicing changes enabled, which is the same behavior as
+Windows App SDK 1.6 and earlier.
+
+## CompatibilityOptions.PatchMode1
+
+Optional patch mode to use if the runtime version matches the major.minor version. If the runtime
+version does not match the specified major.minor version, this value is ignored.
+
+Instead of directly using this API, this patch mode could be specified in the app's project file,
+like this:
+
+```xml
+  <PropertyGroup>
+    <WindowsAppSDKRuntimePatchMode>1.7.3</WindowsAppSDKRuntimePatchMode>
+  </PropertyGroup>
+```
+
+If your app is not in the process of transitioning to a new version of the Windows App SDK, you
+should only set this one patch mode.
+
+## CompatibilityOptions.PatchMode2
+
+Optional patch mode to use if the runtime version matches the major.minor version. If the runtime
+version does not match the specified major.minor version, this value is ignored.
+
+This property enables setting a second patch mode to help apps transitioning to a new version of
+the Windows App SDK. This is a convenience to allow the patch modes for both the old and new
+version to be specified during the transition. Apps not in the process of transitioning should
+only set the one patch mode they want to use.
+
+Setting both patch modes for the same major.minor version, such as 1.7.3 and 1.7.4, is not allowed
+and will generate an error during `Apply()`.
+
+Instead of directly using this API, this patch mode could be specified in the app's project file,
+like this:
+
+```xml
+  <PropertyGroup>
+    <WindowsAppSDKRuntimePatchMode2>1.8.2</WindowsAppSDKRuntimePatchMode2>
+  </PropertyGroup>
+```
+
+## CompatibilityOptions.DisabledChanges
+
+Optional list of specific servicing changes to disable.
+
+If you encounter a problem with a specific change in a servicing update, you can disable just that
+change by adding it to the `DisabledChanges` list before calling `Apply`, or by specifying it in
+your app's project file:
+```xml
+  <PropertyGroup>
+    <WindowsAppSDKDisabledChanges>HypotheticalChange, OtherHypotheticalChange</WindowsAppSDKDisabledChanges>
+  </PropertyGroup>
+```
+
+The [release notes](https://learn.microsoft.com/windows/apps/windows-app-sdk/stable-channel) for
+the Windows App SDK will list the name of each change that can be disabled.
+
+Note that disabling a change is a temporary measure which gives time for a fix to be released in a
+future Windows App SDK update or for you to implement an update in your app. You should report
+changes you disable if you believe it to be a bug in the Windows App SDK. This will help the
+Windows App SDK team to understand the impact of the change and prioritize a fix. You can report
+issues on the [WindowsAppSDK GitHub repo](https://github.com/microsoft/WindowsAppSDK/issues/new?template=bug-report.yaml).
+
+The capability to disable a change is **not** available across new Windows App SDK stable
+releases (such as 1.7.x to 1.8.x), so this does not allow you to permanently disable an
+intentional behavior change. Also, new stable releases, such as 1.8.0, start with no changes
+available to disable -- this capability only applies to disabling targeted servicing changes
+added in servicing releases.
+
+## CompatibilityOptions.Apply
+
+Apply the set compatibility options to the runtime.
+
+Note that CompatibilityOptions must be applied early in the process before any other Windows App
+SDK APIs are called, or right after initializing the Windows App Runtime. The options must be set
+early before the configuration is locked for the lifetime of the process. Since the Windows App
+Runtime needs to run with a consistent configuration, it will lock the configuration when it needs
+to ensure the configuration will no longer change. Calling `Apply` will also lock the
+configuration.
+
+If the configuration has already been locked, calling `Apply` will throw an `E_ILLEGAL_STATE_CHANGE`
+exception if the options differ from the locked configuration. It is okay to call `Apply` multiple
+times with the same configuration, such as if the process had initialized and used the Windows
+App Runtime earlier and is reinitializing for additional use.
+
+# API Details
+
+```c# (but really MIDL3)
+namespace Microsoft.Windows.ApplicationModel.WindowsAppRuntime
+{
+    /// The set of servicing changes that can be disabled.
+    enum CompatibilityChange
+    {
+        None = 0, /// do not use this value
+
+        // Add real changes here:
+
+        // 1.7.1
+        // HypotheticalChange,
+        // OtherHypotheticalChange,
+        // ...
+    };
+
+    /// Represents a version of the Windows App Runtime.
+    struct WindowsAppRuntimeVersion
+    {
+        UInt32 Major;
+        UInt32 Minor;
+        UInt32 Patch;
+    };
+
+    /// This object is used by the app to configure any desired compatibility options
+    /// for Windows App Runtime behavior of changes added in servicing updates. This
+    /// object is only used to set the runtime behavior and can't be used to query the
+    /// applied options.
+    runtimeclass CompatibilityOptions
+    {
+        CompatibilityOptions();
+
+        /// An optional patch mode to use if the runtime version matches the major.minor version.
+        WindowsAppRuntimeVersion PatchMode1 { get; set; };
+
+        /// An optional patch mode to use if the runtime version matches the major.minor version.
+        WindowsAppRuntimeVersion PatchMode2 { get; set; };
+
+        /// An optional list of specific servicing changes to disable.
+        IVector<CompatibilityChange> DisabledChanges{ get; };
+
+        /// Apply the compatibility options to the runtime.
+        void Apply();
+    }
+}
+```
+
--- a/specs/Compatibility/CompatibilityOptions.md
+++ b/specs/Compatibility/CompatibilityOptions.md
@@ -53,7 +53,7 @@ void ApplyCompatibilityOptions()
     var compatibilityOptions = new CompatibilityOptions();
     compatibilityOptions.PatchMode1 = new WindowsAppRuntimeVersion(1,7,3);
     compatibilityOptions.PatchMode2 = new WindowsAppRuntimeVersion(1,8,2);
-    compatibilityOptions.DisabledChanges.Add(CompatibilityChange.HypotheticalChange);
+    compatibilityOptions.DisabledChanges.Add(CompatibilityChange.SampleApiCrashFix);
     compatibilityOptions.Apply();
 }
 ```
@@ -78,7 +78,7 @@ version and disabled changes in the project file (such as `.csproj` or `.vcxproj
   <PropertyGroup>
     <WindowsAppSDKRuntimePatchMode>1.7.3</WindowsAppSDKRuntimePatchMode>
     <WindowsAppSDKRuntimePatchMode2>1.8.2</WindowsAppSDKRuntimePatchMode2>
-    <WindowsAppSDKDisabledChanges>HypotheticalChange, OtherHypotheticalChange</WindowsAppSDKDisabledChanges>
+    <WindowsAppSDKDisabledChanges>SampleApiCrashFix, OtherSampleApiCrashFix</WindowsAppSDKDisabledChanges>
   </PropertyGroup>
 ```
 
@@ -140,7 +140,7 @@ change by adding it to the `DisabledChanges` list before calling `Apply`, or by 
 your app's project file:
 ```xml
   <PropertyGroup>
-    <WindowsAppSDKDisabledChanges>HypotheticalChange, OtherHypotheticalChange</WindowsAppSDKDisabledChanges>
+    <WindowsAppSDKDisabledChanges>SampleApiCrashFix, OtherSampleApiCrashFix</WindowsAppSDKDisabledChanges>
   </PropertyGroup>
 ```
 
@@ -188,8 +188,8 @@ namespace Microsoft.Windows.ApplicationModel.WindowsAppRuntime
         // Add real changes here:
 
         // 1.7.1
-        // HypotheticalChange,
-        // OtherHypotheticalChange,
+        // SampleApiCrashFix,
+        // OtherSampleApiCrashFix,
         // ...
     };
 
--- a/specs/Compatibility/CompatibilityOptions.md
+++ b/specs/Compatibility/CompatibilityOptions.md
@@ -58,8 +58,14 @@ void ApplyCompatibilityOptions()
 }
 ```
 
-Note that CompatibilityOptions must be applied early in the process before any other Windows App
+CompatibilityOptions must be applied early in the process before any other Windows App
 SDK APIs are called, or right after initializing the Windows App Runtime.
+
+`PatchMode1` and `PatchMode2` are simply two fields to set relevant patch modes. These needn't
+match any specific version of the Windows App Runtime nor be in a specific order, so it is valid
+to set `PatchMode1` to 1.8.2 and `PatchMode2` to 1.7.3, for example. And, in the above example,
+when updating the app to 1.9, you may choose to simply update `PatchMode1` to 1.9.3 and leave
+`PatchMode2` as 1.8.2.
 
 ### Specifying CompatibilityOptions in the app's project file
 
@@ -83,8 +89,8 @@ values to disable.
 
 If no `PatchMode1` or `PatchMode2` is specified, or if neither value matches the major.minor
 version of the runtime being used, then the runtime will use the latest patch version. In other
-words, the runtime will run with all servicing changes enabled, which is the same behavior as
-Windows App SDK 1.6 and earlier.
+words, the runtime will run with all servicing changes enabled (just like how Windows App SDK
+worked before this API existed).
 
 ## CompatibilityOptions.PatchMode1
 
@@ -157,7 +163,7 @@ added in servicing releases.
 
 Apply the set compatibility options to the runtime.
 
-Note that CompatibilityOptions must be applied early in the process before any other Windows App
+CompatibilityOptions must be applied early in the process before any other Windows App
 SDK APIs are called, or right after initializing the Windows App Runtime. The options must be set
 early before the configuration is locked for the lifetime of the process. Since the Windows App
 Runtime needs to run with a consistent configuration, it will lock the configuration when it needs
--- a/specs/Compatibility/RuntimeCompatibilityOptions.md
+++ b/specs/Compatibility/RuntimeCompatibilityOptions.md
@@ -3,7 +3,7 @@ App-configurable compatibility options
 
 # Background
 
-_This spec adds a CompatibilityOptions class to control behavior of servicing changes._
+_This spec adds a RuntimeCompatibilityOptions class to control behavior of servicing changes._
 
 Apps using Windows App SDK have two choices today: use Windows App SDK as an
 automatically-updating framework package or use Windows App SDK in self-contained mode with no
@@ -21,16 +21,16 @@ compatibility problem in a servicing release will still be able to move forward 
 disabling problematic changes. Even apps using self-contained can update to a new package with
 confidence that app-configurable compatibility will ensure a successful update.
 
-A new `CompatibilityOptions` class has APIs to control the behavior of servicing changes.
+A new `RuntimeCompatibilityOptions` class has APIs to control the behavior of servicing changes.
 There are also properties which can be set in the app's project file to automatically use the new
 APIs with the specified values.
 
 # API Pages
 
-## CompatibilityOptions class
+## RuntimeCompatibilityOptions class
 
 Configures which changes in Windows App SDK servicing releases are enabled. By default, all
-changes are enabled, but `CompatibilityOptions` can be used to lock the runtime behavior to a
+changes are enabled, but `RuntimeCompatibilityOptions` can be used to lock the runtime behavior to a
 specified patch version or to disable specific changes:
 
 1. **Choose the patch version:** You can specify which servicing patch version's behavior you
@@ -44,21 +44,21 @@ specified patch version or to disable specific changes:
    version in use.) Disabling a change is a temporary measure which gives time for a fix to be
    released in a future Windows App SDK update or for you to implement an update in your app.
 
-Here's an example of using `CompatibilityOptions` to specify a patch version and disable a
+Here's an example of using `RuntimeCompatibilityOptions` to specify a patch version and disable a
 specific change:
 
 ```c#
-void ApplyCompatibilityOptions() 
+void ApplyRuntimeCompatibilityOptions() 
 {
-    var compatibilityOptions = new CompatibilityOptions();
+    var compatibilityOptions = new RuntimeCompatibilityOptions();
     compatibilityOptions.PatchMode1 = new WindowsAppRuntimeVersion(1,7,3);
     compatibilityOptions.PatchMode2 = new WindowsAppRuntimeVersion(1,8,2);
-    compatibilityOptions.DisabledChanges.Add(CompatibilityChange.SampleApiCrashFix);
+    compatibilityOptions.DisabledChanges.Add(RuntimeCompatibilityChange.SampleApiCrashFix);
     compatibilityOptions.Apply();
 }
 ```
 
-CompatibilityOptions must be applied early in the process before any other Windows App
+RuntimeCompatibilityOptions must be applied early in the process before any other Windows App
 SDK APIs are called, or right after initializing the Windows App Runtime.
 
 `PatchMode1` and `PatchMode2` are simply two fields to set relevant patch modes. These needn't
@@ -67,10 +67,10 @@ to set `PatchMode1` to 1.8.2 and `PatchMode2` to 1.7.3, for example. And, in the
 when updating the app to 1.9, you may choose to simply update `PatchMode1` to 1.9.3 and leave
 `PatchMode2` as 1.8.2.
 
-### Specifying CompatibilityOptions in the app's project file
+### Specifying RuntimeCompatibilityOptions in the app's project file
 
 You can also specify the patch version and disabled changes in the app's project file instead of
-directly calling the CompatibilityOptions APIs. This approach has the advantage of ensuring the
+directly calling the RuntimeCompatibilityOptions APIs. This approach has the advantage of ensuring the
 options are applied early at the proper timing. Here's an example of how to specify the patch
 version and disabled changes in the project file (such as `.csproj` or `.vcxproj`):
 
@@ -82,7 +82,7 @@ version and disabled changes in the project file (such as `.csproj` or `.vcxproj
   </PropertyGroup>
 ```
 
-The `WindowsAppSDKDisabledChanges` property is a comma-separated list of `CompatibilityChange`
+The `WindowsAppSDKDisabledChanges` property is a comma-separated list of `RuntimeCompatibilityChange`
 values to disable.
 
 ### Behavior with no PatchMode specified
@@ -92,7 +92,7 @@ version of the runtime being used, then the runtime will use the latest patch ve
 words, the runtime will run with all servicing changes enabled (just like how Windows App SDK
 worked before this API existed).
 
-## CompatibilityOptions.PatchMode1
+## RuntimeCompatibilityOptions.PatchMode1
 
 Optional patch mode to use if the runtime version matches the major.minor version. If the runtime
 version does not match the specified major.minor version, this value is ignored.
@@ -109,7 +109,7 @@ like this:
 If your app is not in the process of transitioning to a new version of the Windows App SDK, you
 should only set this one patch mode.
 
-## CompatibilityOptions.PatchMode2
+## RuntimeCompatibilityOptions.PatchMode2
 
 Optional patch mode to use if the runtime version matches the major.minor version. If the runtime
 version does not match the specified major.minor version, this value is ignored.
@@ -131,7 +131,7 @@ like this:
   </PropertyGroup>
 ```
 
-## CompatibilityOptions.DisabledChanges
+## RuntimeCompatibilityOptions.DisabledChanges
 
 Optional list of specific servicing changes to disable.
 
@@ -159,11 +159,11 @@ intentional behavior change. Also, new stable releases, such as 1.8.0, start wit
 available to disable -- this capability only applies to disabling targeted servicing changes
 added in servicing releases.
 
-## CompatibilityOptions.Apply
+## RuntimeCompatibilityOptions.Apply
 
 Apply the set compatibility options to the runtime.
 
-CompatibilityOptions must be applied early in the process before any other Windows App
+RuntimeCompatibilityOptions must be applied early in the process before any other Windows App
 SDK APIs are called, or right after initializing the Windows App Runtime. The options must be set
 early before the configuration is locked for the lifetime of the process. Since the Windows App
 Runtime needs to run with a consistent configuration, it will lock the configuration when it needs
@@ -181,7 +181,7 @@ App Runtime earlier and is reinitializing for additional use.
 namespace Microsoft.Windows.ApplicationModel.WindowsAppRuntime
 {
     /// The set of servicing changes that can be disabled.
-    enum CompatibilityChange
+    enum RuntimeCompatibilityChange
     {
         None = 0, /// do not use this value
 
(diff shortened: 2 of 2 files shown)


specs/Compatibility/CompatibilityOptions.md
```diff
@@ -0,0 +1,220 @@
...
+```xml
+  <PropertyGroup>
+    <WindowsAppSDKRuntimePatchMode>1.7.3</WindowsAppSDKRuntimePatchMode>
+  </PropertyGroup>
+```
+
+If your app is not in the process of transitioning to a new version of the Windows App SDK, you
+should only set this one patch mode.
+
+## CompatibilityOptions.PatchMode2
```
DrusTheAxe, line 106: `PatchMode2` name is misleading. It's a second meant for transition period. But couldn't I accomplish same with build logic and even more flexibly e.g.

```
<WindowsAppSDKRuntimePatchMode Condition="$([System.Text.RegularExpressions.Regex]::Match($(WindowsAppSDKVersion), '^\d+\.\d+')==1.7">1.7.3</WindowsAppSDKRuntimePatchMode>
<WindowsAppSDKRuntimePatchMode Condition="$([System.Text.RegularExpressions.Regex]::Match($(WindowsAppSDKVersion), '^\d+\.\d+')==1.8">1.8.2</WindowsAppSDKRuntimePatchMode>
<WindowsAppSDKRuntimePatchMode Condition="$([System.Text.RegularExpressions.Regex]::Match($(WindowsAppSDKVersion), '^\d+\.\d+')==1.9">1.9.5</WindowsAppSDKRuntimePatchMode>
```

or in code via similar e.g. in *.vcxproj

```
<PreprocessorDefinitions Condition="$([System.Text.RegularExpressions.Regex]::Match($(WindowsAppSDKVersion), '^\d+\.\d+')==1.7">MY_WINAPPSDK_PATCHMODE_MAJOR=1;MY_WINAPPSDK_PATCHMODE_MINOR=7;MY_WINAPPSDK_PATCHMODE_PATCH=3;%(PreprocessorDefinitions)</PreprocessorDefinitions>
<PreprocessorDefinitions Condition="$([System.Text.RegularExpressions.Regex]::Match($(WindowsAppSDKVersion), '^\d+\.\d+')==1.8">MY_WINAPPSDK_PATCHMODE_MAJOR=1;MY_WINAPPSDK_PATCHMODE_MINOR=8;MY_WINAPPSDK_PATCHMODE_PATCH=2;%(PreprocessorDefinitions)</PreprocessorDefinitions>
...
```

and in code

```
compatibilityOptions.PatchMode = new WindowsAppRuntimeVersion(MY_WINAPPSDK_PATCHMODE_MAJOR, MY_WINAPPSDK_PATCHMODE_MINOR, MY_WINAPPSDK_PATCHMODE_PATCH);
```

specs/Compatibility/CompatibilityOptions.md
```diff
@@ -0,0 +1,220 @@
...
+change by adding it to the `DisabledChanges` list before calling `Apply`, or by specifying it in
+your app's project file:
+```xml
+  <PropertyGroup>
+    <WindowsAppSDKDisabledChanges>HypotheticalChange, OtherHypotheticalChange</WindowsAppSDKDisabledChanges>
+  </PropertyGroup>
+```
+
+The [release notes](https://learn.microsoft.com/windows/apps/windows-app-sdk/stable-channel) for
+the Windows App SDK will list the name of each change that can be disabled.
```
DrusTheAxe, line 142: Can a list be provided in the SDK for build-time reference? e.g.

Add `AppConfigurableCompatibility.json` into the WinAppSDK nuget containing

```
{
  "AppConfigurableCompatibility": {
    "DisabledChanges": [
      {
        "DisabledChange": "HypotheticalChange",
        "Description": "Blah blah blah",
        "Version(s)": [
          "1.7.2+"
        ],
        "URL": "https://github.com/microsoft/WindowsAppSDK/docs/appconfigurablecompatibility/HypotheticalChange"
      },
      {
        "DisabledChange": "OtherHypotheticalChange",
        "Description": "Yadda yadda",
        "Version(s)": [
          "1.7.3 - 1.7.6"
        ],
        "URL": "https://github.com/microsoft/WindowsAppSDK/docs/appconfigurablecompatibility/OtherHypotheticalChange"
      },
      {
        "DisabledChange": "YetAnotherHypotheticalChange",
        "Description": "Booyah...",
        "Version(s)": [
          "1.7.3",
          "1.7.5+"
        ],
        "URL": "https://github.com/microsoft/WindowsAppSDK/docs/appconfigurablecompatibility/YetAnotherHypotheticalChange"
      }
    ]
  }
}
```

specs/Compatibility/CompatibilityOptions.md
```diff
@@ -0,0 +1,220 @@
...
+
+Note that disabling a change is a temporary measure which gives time for a fix to be released in a
+future Windows App SDK update or for you to implement an update in your app. You should report
+changes you disable if you believe it to be a bug in the Windows App SDK. This will help the
```
DrusTheAxe, line 146: Are disabled changes reported via insights?

specs/Compatibility/CompatibilityOptions.md
```diff
@@ -0,0 +1,220 @@
...
+Windows App SDK team to understand the impact of the change and prioritize a fix. You can report
+issues on the [WindowsAppSDK GitHub repo](https://github.com/microsoft/WindowsAppSDK/issues/new?template=bug-report.yaml).
+
+The capability to disable a change is **not** available across new Windows App SDK stable
+releases (such as 1.7.x to 1.8.x), so this does not allow you to permanently disable an
+intentional behavior change. Also, new stable releases, such as 1.8.0, start with no changes
+available to disable -- this capability only applies to disabling targeted servicing changes
+added in servicing releases.
```
DrusTheAxe, line 154: This implies `DisabledChanges` are version-specific?

Then shouldn't these be named `_1_7_HypotheticalChange` rather than `HypotheticalChange`? If similar option's provided in 1.8 then it would need a different name to specify to DisableChanges?

This gets more to _policy_ how we name these (Hypothetical)Changes but strong guidance from the start is best.

specs/Compatibility/CompatibilityOptions.md
```diff
@@ -0,0 +1,220 @@
...
+
+## CompatibilityOptions.Apply
+
+Apply the set compatibility options to the runtime.
+
+Note that CompatibilityOptions must be applied early in the process before any other Windows App
```
DrusTheAxe, line 160: SUGGESTION: **NOTE:** `CompatibilityOptions MUST be...

* Bold the **NOTE:** prefix instead of conversational english
* Capitalize MUST, much like IEEE MUST/SHOULD/MAY phrasing ([RFC 2119](https://www.rfc-editor.org/rfc/rfc2119))

specs/Compatibility/CompatibilityOptions.md
```diff
@@ -0,0 +1,220 @@
...
+SDK APIs are called, or right after initializing the Windows App Runtime. The options must be set
+early before the configuration is locked for the lifetime of the process. Since the Windows App
+Runtime needs to run with a consistent configuration, it will lock the configuration when it needs
+to ensure the configuration will no longer change. Calling `Apply` will also lock the
+configuration.
+
+If the configuration has already been locked, calling `Apply` will throw an `E_ILLEGAL_STATE_CHANGE`
```
DrusTheAxe, line 167: Is that a way to 'Reset` or 'Unapply' the configuration? OR no, 'till process death do you part?

specs/Compatibility/CompatibilityOptions.md
```diff
@@ -0,0 +1,220 @@
...
+exception if the options differ from the locked configuration. It is okay to call `Apply` multiple
+times with the same configuration, such as if the process had initialized and used the Windows
+App Runtime earlier and is reinitializing for additional use.
+
+# API Details
+
+```c# (but really MIDL3)
+namespace Microsoft.Windows.ApplicationModel.WindowsAppRuntime
```
DrusTheAxe, line 175: Missing API contract

SUGGESTION:
```
{
    [contractversion(1)]
    apicontract AppConfigurableCompatibilityContract{};

    /// The set of servicing changes that can be disabled.
    [contract(AppConfigurableCompatibilityContract, 1)]
    enum CompatibilityChange
...
```

specs/Compatibility/CompatibilityOptions.md
```diff
@@ -0,0 +1,220 @@
...
+
+    /// Represents a version of the Windows App Runtime.
+    struct WindowsAppRuntimeVersion
+    {
+        UInt32 Major;
+        UInt32 Minor;
+        UInt32 Patch;
+    };
+
+    /// This object is used by the app to configure any desired compatibility options
```
DrusTheAxe, line 198: By the "app" or "process"?

Can OOP WinRT servers use this API? Are OOP WinRT servers apps?

SUGGESTION: This object is used by the process...

specs/Compatibility/CompatibilityOptions.md
```diff
@@ -0,0 +1,220 @@
...
+    /// for Windows App Runtime behavior of changes added in servicing updates. This
+    /// object is only used to set the runtime behavior and can't be used to query the
+    /// applied options.
```
DrusTheAxe, line 201: Why no query API?

Let me guess: An app specifying this knows what it's doing so it doesn't have to query at runtime.

Question: What are class libraries supposed to do?

Lacking a query API creates issues for library authors. How does a library know current process' behavior? They don't control the process/environment they run in (not the app). Would that amount to...?

```
[contract(AppConfigurableCompatibilityContract, 1)]
runtimeclass CurrentCompatibilityOptions
{
    static CurrentCompatibilityOptions GetDefault();

    WindowsAppRuntimeVersion EffectiveVersion { get; }

    IVector<CompatibilityChange> DisabledChanges{ get; };
}
```

specs/Compatibility/CompatibilityOptions.md
```diff
@@ -0,0 +1,220 @@
+App-configurable compatibility options
+===
+
+# Background
+
+_This spec adds a CompatibilityOptions class to control behavior of servicing changes._
+
+Apps using Windows App SDK have two choices today: use Windows App SDK as an
+automatically-updating framework package or use Windows App SDK in self-contained mode with no
+automatic updates. Apps which use the framework package automatically get fixes, which is great,
```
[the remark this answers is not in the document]
codendone: Updates **are** automatic for apps using the framework package. And, as already in the text, apps using self-contained do not get automatic updates. I think the current text is fine for the context of this spec.

specs/Compatibility/CompatibilityOptions.md
```diff
@@ -0,0 +1,220 @@
...
+Here's an example of using `CompatibilityOptions` to specify a patch version and disable a
+specific change:
+
+```c#
+void ApplyCompatibilityOptions() 
+{
+    var compatibilityOptions = new CompatibilityOptions();
+    compatibilityOptions.PatchMode1 = new WindowsAppRuntimeVersion(1,7,3);
+    compatibilityOptions.PatchMode2 = new WindowsAppRuntimeVersion(1,8,2);
+    compatibilityOptions.DisabledChanges.Add(CompatibilityChange.HypotheticalChange);
```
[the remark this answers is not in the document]
codendone: There's nothing real I can use instead here. I could put something like `TextBoxDestructionCrash` or `RegisterPackageRequiresUri`, to name two fixes which were in 1.6.3, but then we're likely to debate those names here.

specs/Compatibility/CompatibilityOptions.md
```diff
@@ -0,0 +1,220 @@
...
+    compatibilityOptions.Apply();
+}
+```
+
+Note that CompatibilityOptions must be applied early in the process before any other Windows App
+SDK APIs are called, or right after initializing the Windows App Runtime.
```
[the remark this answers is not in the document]
codendone: The bootstrapper or other initialization code, whether from WinAppSDK auto-initializers for from explicit app calls, needs to happen first. This should happen right after that. I didn't have a better way to describe that timing that what I wrote here.

specs/Compatibility/CompatibilityOptions.md
codendone: What if I remove "Note that"? I think this is a weird place in the text to have "**NOTE:**". Maybe removing "Note that" would be better?

specs/Compatibility/CompatibilityOptions.md
codendone: There is no way to reset or unapply (unless WinAppSDK somehow gets entirely unloaded from the process). The intent is indeed 'til process death.

specs/Compatibility/CompatibilityOptions.md
codendone: Is that typically in the API spec? The actual code has that, which I think should be sufficient.

specs/Compatibility/CompatibilityOptions.md
codendone: I consider the servers to be apps as well. In this place, I prefer saying "app" as something which I believe will resonate better with developers.

specs/Compatibility/CompatibilityOptions.md
codendone: Class libraries don't get to choose. Yes, the app should know what it is doing, but shouldn't have any conditional code. The intent here is that no app or class library code conditions on whether the implementation of WinAppSDK code has a fix enabled or disabled.

We could later add a query API if a good reason and need for one is found.

specs/Compatibility/CompatibilityOptions.md
```diff
@@ -0,0 +1,220 @@
...
+automatically-updating framework package or use Windows App SDK in self-contained mode with no
+automatic updates. Apps which use the framework package automatically get fixes, which is great,
+but they have a chance of being broken by a bug or subtle behavior change in a servicing update.
+Some apps choose self-contained mode to avoid that risk of an automatic servicing update breaking
+them. But even when apps choose self-contained mode, they may find that they can't update to a
+newer servicing release due to a bug or behavior change. Also, self-contained apps can't use
+features which require shared services, such as some Notification features delivered via the
+Singleton package.
+
+App-configurable compatibility is intended to prevent these problems, enabling apps to choose to
```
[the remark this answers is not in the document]
codendone: I think 'App-configurable compatibility' is useful way for us to succinctly understand the functionality of this new feature, but I thought some form of `CompatibilityOptions` was a better name for the API called by apps.

specs/Compatibility/CompatibilityOptions.md
```diff
@@ -0,0 +1,220 @@
...
+
+# API Pages
+
+## CompatibilityOptions class
+
+Configures which changes in Windows App SDK servicing releases are enabled. By default, all
+changes are enabled, but `CompatibilityOptions` can be used to lock the runtime behavior to a
+specified patch version or to disable specific changes:
+
+1. **Choose the patch version:** You can specify which servicing patch version's behavior you
```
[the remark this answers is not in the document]
codendone: In my mind, this is consistent. The app developer chooses which WinAppSDK patch version they want to snap to, and then they set that as the patch mode to run in. Whether `PatchMode` is the right API name will, of course, be something considered in API review.

specs/Compatibility/CompatibilityOptions.md
```diff
@@ -0,0 +1,220 @@
+App-configurable compatibility options
+===
+
+# Background
+
+_This spec adds a CompatibilityOptions class to control behavior of servicing changes._
```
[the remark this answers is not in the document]
Scottj1s: Agreed - even though it's in a namespace, I'd prefer a more specific name.

specs/Compatibility/CompatibilityOptions.md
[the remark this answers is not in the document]
Scottj1s: Agree that a unified term would be clearer.  To me, "mode" doesn't quite capture it - I think of a mode as orthogonal to a given release/version/patch (like self-contained mode).  How about PatchLevel?

specs/Compatibility/CompatibilityOptions.md
```diff
@@ -0,0 +1,220 @@
...
+Configures which changes in Windows App SDK servicing releases are enabled. By default, all
+changes are enabled, but `CompatibilityOptions` can be used to lock the runtime behavior to a
+specified patch version or to disable specific changes:
+
+1. **Choose the patch version:** You can specify which servicing patch version's behavior you
+   want to use. For example, your app can specify that it wants the 1.7.2 version behavior, which
+   will have Windows App SDK run in that mode even if 1.7.3 or later is installed. This capability
+   allows you to control when your app gets new fixes or behavior changes, even when not using
+   self-contained mode.
+2. **Temporarily disable specific changes:** If your app encounters a problem with a specific
```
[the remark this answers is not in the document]
Scottj1s: I like the term "temporarily" as it conveys the temporary nature of the disable-ability (per major.minor).  But I do think it's a bit confusing that while this status only applies to a specific major.minor version, it's application here is 'global' (not scoped to a specific patch version).  That could be a sharp edge - maybe we should scope change disabling to a patch version.

specs/Compatibility/CompatibilityOptions.md
```diff
@@ -0,0 +1,220 @@
...
+
+Here's an example of using `CompatibilityOptions` to specify a patch version and disable a
+specific change:
+
+```c#
+void ApplyCompatibilityOptions() 
+{
+    var compatibilityOptions = new CompatibilityOptions();
+    compatibilityOptions.PatchMode1 = new WindowsAppRuntimeVersion(1,7,3);
+    compatibilityOptions.PatchMode2 = new WindowsAppRuntimeVersion(1,8,2);
```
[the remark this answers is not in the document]
Scottj1s: I think you would omit PatchMode2 altogether if you were happy with 1.8.anything

specs/Compatibility/CompatibilityOptions.md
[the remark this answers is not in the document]
Scottj1s: Agreed that a generic symbol is appropriate, but 'hypothetical' raises my eyebrows - maybe SampleChange?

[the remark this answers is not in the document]
Scottj1s: This is where scoping disabled changes to a patch level would make sense - error detected and reporting

specs/Compatibility/CompatibilityOptions.md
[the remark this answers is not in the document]
Scottj1s: I think "early in the process" is sufficient, given that an error would occur otherwise (the developer will learn how early is enough)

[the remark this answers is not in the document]
Scottj1s: Michael unified all the auto-initializer code to ensure that compatibility options are correctly sequenced.

specs/Compatibility/CompatibilityOptions.md
```diff
@@ -0,0 +1,226 @@
...
+### Specifying CompatibilityOptions in the app's project file
+
+You can also specify the patch version and disabled changes in the app's project file instead of
+directly calling the CompatibilityOptions APIs. This approach has the advantage of ensuring the
+options are applied early at the proper timing. Here's an example of how to specify the patch
+version and disabled changes in the project file (such as `.csproj` or `.vcxproj`):
+
+```xml
+  <PropertyGroup>
+    <WindowsAppSDKRuntimePatchMode>1.7.3</WindowsAppSDKRuntimePatchMode>
```
Scottj1s, line 79:
```suggestion
    <WindowsAppSDKRuntimePatchMode1>1.7.3</WindowsAppSDKRuntimePatchMode1>
```

Scottj1s: numeric suffixes should be consistent between xml and api

specs/Compatibility/CompatibilityOptions.md
```diff
@@ -0,0 +1,220 @@
...
+    <WindowsAppSDKRuntimePatchMode2>1.8.2</WindowsAppSDKRuntimePatchMode2>
+    <WindowsAppSDKDisabledChanges>HypotheticalChange, OtherHypotheticalChange</WindowsAppSDKDisabledChanges>
+  </PropertyGroup>
+```
+
+The `WindowsAppSDKDisabledChanges` property is a comma-separated list of `CompatibilityChange`
```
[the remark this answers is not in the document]
Scottj1s: not sure the flexible syntax warrants the parsing complexity.  presumably, most users would only ever apply a handful of changes.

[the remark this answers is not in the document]
Scottj1s: or even whitespace-only-delimited?
<WindowsAppSDKDisabledChanges>HypotheticalChange OtherHypotheticalChange</>

specs/Compatibility/CompatibilityOptions.md
Scottj1s: > `PatchMode2` name is misleading. It's a second meant for transition period. But couldn't I accomplish same with build logic and even more flexibly e.g.

I like the suggestion here - there's something mildly rigid about having the Mode1 and Mode2 properties.  Making the assignments dynamic allows for not just 2, but N possibilities.  

We could provide a WindowsAppSDKMajorMinorVersion property to make the conditions simpler:
`<WindowsAppSDKRuntimePatchMode Condition="'$(WindowsAppSDKMajorMinorVersion)'=='1.7'">1.7.3</WindowsAppSDKRuntimePatchMode>`

Scottj1s: Ah, I think Michael had in mind interactions with A/B Forwarding here, where the runtime version is selected dynamically (not tied to the SDK version built against).  So in that case, explicit properties are necessary.

specs/Compatibility/CompatibilityOptions.md
Scottj1s: Yes, change disabling should be patch-specific

specs/Compatibility/CompatibilityOptions.md
```diff
@@ -0,0 +1,226 @@
...
+The [release notes](https://learn.microsoft.com/windows/apps/windows-app-sdk/stable-channel) for
+the Windows App SDK will list the name of each change that can be disabled.
+
+Note that disabling a change is a temporary measure which gives time for a fix to be released in a
+future Windows App SDK update or for you to implement an update in your app. You should report
+changes you disable if you believe it to be a bug in the Windows App SDK. This will help the
+Windows App SDK team to understand the impact of the change and prioritize a fix. You can report
+issues on the [WindowsAppSDK GitHub repo](https://github.com/microsoft/WindowsAppSDK/issues/new?template=bug-report.yaml).
+
+The capability to disable a change is **not** available across new Windows App SDK stable
```
Scottj1s, line 156: Do we have automation in place (or planned), both to enforce containment of fixes, and to reset the containment "counter" for a new major.minor release?

specs/Compatibility/CompatibilityOptions.md
```diff
@@ -0,0 +1,226 @@
...
+    /// The set of servicing changes that can be disabled.
+    enum CompatibilityChange
+    {
+        None = 0, /// do not use this value
+
+        // Add real changes here:
+
+        // 1.7.1
+        // HypotheticalChange,
+        // OtherHypotheticalChange,
```
Scottj1s, line 192: How do we see multiple changelists being applied across patch versions, given the app only sees one version of the SDK?  i.e., where would the // 1.8.3 changes discovered?

specs/Compatibility/CompatibilityOptions.md
```diff
@@ -0,0 +1,226 @@
...
+CompatibilityOptions must be applied early in the process before any other Windows App
+SDK APIs are called, or right after initializing the Windows App Runtime. The options must be set
+early before the configuration is locked for the lifetime of the process. Since the Windows App
+Runtime needs to run with a consistent configuration, it will lock the configuration when it needs
+to ensure the configuration will no longer change. Calling `Apply` will also lock the
+configuration.
+
+If the configuration has already been locked, calling `Apply` will throw an `E_ILLEGAL_STATE_CHANGE`
+exception if the options differ from the locked configuration. It is okay to call `Apply` multiple
+times with the same configuration, such as if the process had initialized and used the Windows
```
Scottj1s, line 175: Is this a real scenario?  Is WinUI finally robust enough to support it?  If not, I'd just omit this sentence and allow that multiple calls are idempotent.

specs/Compatibility/CompatibilityOptions.md
DrusTheAxe: >The intent here is that no app or class library code conditions on whether the implementation of WinAppSDK code has a fix enabled or disabled.

Nice in theory but as they say, the only difference between theory and practice is in theory there's no difference whereas in practice...

Some behavioral changes can't be detected (w/o dipping into version checks). Lacking a query API leaves class library developers at a loss unable to know if they need to jump left or right. This _could_ be solved by dictating all CompatibilityChanges MUST be detectable via an `IsSupported` API e.g.

```
if (Microsoft.Stuff.Thingamajig.IsSupported("foo"))
{
    ...ComapatibilityChange.Foo is enabled...
}
else
{
    ...ComapatibilityChange.Foo is disabled or doesn't exist yet (e.g. checking in 1.7.2 when Foo was introduced in 1.7.3)...
}
```

If the feature doesn't already have an `IsFooSupported()` or `IsSupported(parameter_foo)` API they'll be required to add one _in a patch servicing release_, or if an `IsSupported(parm)` exists they'll need to invent a new parameter in a patch release ("easy" if parm=string, less so if enum).

This is a broad tax. It's not isolated to one feature, team or release.

A query API simplifies this problem.

specs/Compatibility/CompatibilityOptions.md
```diff
@@ -0,0 +1,226 @@
...
+    /// applied options.
+    runtimeclass CompatibilityOptions
+    {
+        CompatibilityOptions();
+
+        /// An optional patch mode to use if the runtime version matches the major.minor version.
+        WindowsAppRuntimeVersion PatchMode1 { get; set; };
+
+        /// An optional patch mode to use if the runtime version matches the major.minor version.
+        WindowsAppRuntimeVersion PatchMode2 { get; set; };
```
DrusTheAxe, line 216: The API is hardwired presuming an app will only work with 2 versions. That doesn't scale into the future if e.g. an app works with 1.7, 1.8 and 1.9. Or when WinAppSDK has LTSC releases for an app to work with e.g. LTSC 1.7, LTSC 1.10, current 1.11 and vNext 1.12. Any change to support &gt;2 patch modes will weird API shape changes (PatchMode3, PatchMode4... or more break-y changes)

RECOMMEND: Replace `PatchMode1` and `PatchMode2` with `IVector<WindowsAppRuntimeVersion> PatchModes{ get; };` Then the object supports an extensible list of versions, much like it supports an extensible list of `DisabledChanges`.

For example, the sample on line 55 changes from

```
void ApplyCompatibilityOptions()
{
    var compatibilityOptions = new CompatibilityOptions();
    compatibilityOptions.PatchMode1 = new WindowsAppRuntimeVersion(1,7,3);
    compatibilityOptions.PatchMode2 = new WindowsAppRuntimeVersion(1,8,2);
    compatibilityOptions.Apply();
}
```

to

```
void ApplyCompatibilityOptions()
{
    var compatibilityOptions = new CompatibilityOptions();
    compatibilityOptions.Add(new WindowsAppRuntimeVersion(1,7,3));
    compatibilityOptions.Add(new WindowsAppRuntimeVersion(1,8,2));
    compatibilityOptions.Apply();
}
```

specs/Compatibility/CompatibilityOptions.md
DrusTheAxe: Why would WinUI matter here? APIs shouldn't be calling `Apply()` only the app (the app/dev owns process context decisions).

`E_ILLEGAL_STATE_CHANGE` for 2nd+ `Apply()` call that would produce different outcomes (and S_OK if no different) seems righteous . Most apps will only call `Apply` once (and more often, nonce) but if devs make multiple calls the error seems appropriate

specs/Compatibility/CompatibilityOptions.md
```diff
@@ -0,0 +1,226 @@
...
+
+CompatibilityOptions must be applied early in the process before any other Windows App
+SDK APIs are called, or right after initializing the Windows App Runtime. The options must be set
+early before the configuration is locked for the lifetime of the process. Since the Windows App
+Runtime needs to run with a consistent configuration, it will lock the configuration when it needs
+to ensure the configuration will no longer change. Calling `Apply` will also lock the
+configuration.
+
+If the configuration has already been locked, calling `Apply` will throw an `E_ILLEGAL_STATE_CHANGE`
+exception if the options differ from the locked configuration. It is okay to call `Apply` multiple
```
DrusTheAxe, line 174: >if the options differ from the locked configuration

The options, or the effective options?

If an app winds up calling

```
var co = new CompatibilityOptions();
co.PatchMode1 = new WindowsAppRuntimeVersion(1,7,3);
co.PatchMode2 = new WindowsAppRuntimeVersion(1,8,2);
co.Apply();
...
var co2 = new CompatibilityOptions();
co2.PatchMode1 = new WindowsAppRuntimeVersion(1,7,3);
co2.Apply();
```

and it's running with 1.7 the _options_ to `Apply` differ but the `effective` options are identical.

Is the check for identical values for all properties of the object, or identical outcomes of the runtime decision?

specs/Compatibility/CompatibilityOptions.md
Scottj1s: WinUI has not supported reinitialization well (TLS cleanup issues), so better not to suggest this.

I have no problem with any subsequent call (whether identical or not) producing an E_ILLEGAL_STATE_CHANGE.  The app ought to be able to manage that requirement and it frees us from complex comparisons.

specs/Compatibility/CompatibilityOptions.md
Scottj1s: I had suggested the list before in the code review.  But it's not necessary to generalize beyond 2 - this is necessary to cover A/B Forward (otherwise one level would be sufficient).

specs/Compatibility/CompatibilityOptions.md
Scottj1s: yeah, seems like all kinds of interesting edge cases are possible.  I vote for a strict single Apply call pattern.

specs/Compatibility/CompatibilityOptions.md
DrusTheAxe: You assume "run with product version X and test version Y". That an app will only need 2 versions, but apps can be designed to work with any of multiple versions (especially when using WinAppSDK/MSIX and not SelfContained).

CompatibilityOptions is available to all WinAppSDK users, right? Not just if WindowsAppSdkSelfContained=true?

Scottj1s: Not an assumption - a hard limit.  A/B Forward will only support building with major.minor and running on that and major.minor-1.  CompatibilityOptions is designed primarily for FWP-dependent apps, not WindowsAppSdkSelfContained.

DrusTheAxe: Hard limit - sure. My point is, if "max 2 versions" needs &gt;2 in the future the API shape would need impactful changes. Plus `PatchMode1` and `PatchMode2` property names are kinda weird. If instead it's a list it wouldn't pose future API shape issues (and not be 'weird'), and there's no significant difference to the developer experience.

P.S. "designed primarily for" == SHOULD, not MUST. That makes it less the law than the Jack Sparrow Rule (There's what a man can do, and what a man can't do...). As John Wick taught us, a pencil can be used for more than designed primarily for... :-)

I didn't see any overt block disallowing or disabling this API if WinAppSDKSelfContained=true. Which is a good thing. Even if we expect more WinAppSDK/MSIX apps using this than WinAppSDK/SelfContained I can see value for the latter, albeit possibly less often. For instance, apps wanting _some_ perf or functional improvements in, say, 1.42.87 but the updated Fubar feature griefs them, so use 1.42.87 but disable CompatibleOption.Fubar until MS gets a fix out in 1.42.88(+). Works equally well for SelfContained as MSIX. Flexibility good. Flexibility without downside even better :P

specs/Compatibility/CompatibilityOptions.md
DrusTheAxe: >strict single Apply call pattern

As in, any 2nd+ `Apply()` errors (regardless the 2nd+ values)?

Seems reasonable. You've got my axe :-)

specs/Compatibility/CompatibilityOptions.md
Scottj1s: It certainly feels like a better API design to use a list and enforce an element count of [0..2] (throwing an exception on element 3).  I can get behind that.  

I do not understand your movie analogies.  But that's ok.

I say designed primarily for FWP-dependent, but not exclusively (per Michael's intro text).  Yes, WindowsAppSdkSelfContained could certainly also benefit - I assume for punching holes via disabled changes, without using the patch level feature (irrelevant).

specs/Compatibility/CompatibilityOptions.md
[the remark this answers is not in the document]
codendone: I think whitespace-only would be less easy to read. Think of the contents of this property like specifying the contents of an array.

codendone (commented):
Oops, I forgot to submit my comments from last week.

specs/Compatibility/CompatibilityOptions.md
codendone: CR and shiproom approval is the only verification we (already) have today. I think that should be sufficient.

There is no need to reset for each new major.minor release: the containment is never in the main branch and therefore never carries over to the next release.

specs/Compatibility/CompatibilityOptions.md
codendone: The second `Apply` will fail if it doesn't exactly match the first one. WinUI has long wanted to support an app being able to uninitialize and reinitialize, and this ability to call `Apply` again with the identical values enables that scenario. I don't see a need to prevent the scenario and make initialization potentially being more difficult for apps which manually initialize WinAppSDK.

specs/Compatibility/CompatibilityOptions.md
codendone: WinUI doesn't support it yet, but why should we make an app's attempt to ensure it is applied be an error?

specs/Compatibility/CompatibilityOptions.md
codendone: Across patch versions within a single major.minor release? Nothing special is required for that scenario.

If you mean specifying DisabledChanges, as an example, for both the 1.7.x version and 1.8.x version the app might use, that is not possible with the proposed design. The CompatibilityChange enum will only be defined with the values for the major.minor version the app compiles against. This adds another reason DisabledChanges should be temporary.

(Technically, a developer could copy the numeric value of the CompatibilityChange from the other major.minor SDK version and directly use that value. The intent is for that to not happen, or at least be so rare that the hack can be used if needed.)

specs/Compatibility/CompatibilityOptions.md
codendone: I started with a list for this property and I didn't like how it looked. It adds unnecessary complexity. As Scott said, our A/B design is specifically between two versions.

specs/Compatibility/CompatibilityOptions.md
codendone: We can add that if/when there is a need.

specs/Compatibility/CompatibilityOptions.md
codendone: One other scenario we've had with apps using XAML Islands is sometimes they are initializing multiple thread simultaneously. We could require apps to add a lock to ensure they only initialize once, which some apps will forget to do, or we can handle this in the `Apply` API using the mechanism described here. Allowing apps to simply Apply without a check is easier and safer for apps.

DrusTheAxe (approved):
LGTM

specs/Compatibility/CompatibilityOptions.md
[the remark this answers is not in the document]
codendone: `RuntimeCompatibilityOptions` sounds good to me. I'll wait for other feedback from API review. (If we go with that, then `CompatibilityChange` would similarly change to `RuntimeCompatibilityChange`.)

specs/Compatibility/CompatibilityOptions.md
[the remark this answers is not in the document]
codendone: I think the enum and a compile error if trying to use something not in the enum is sufficient. We also would not want to provide a runtime warning if DisabledChanges specified for 1.7 are still passed when the app is testing running against 1.8.

specs/Compatibility/CompatibilityOptions.md
```diff
@@ -0,0 +1,220 @@
...
+    compatibilityOptions.Apply();
+}
+```
+
+Note that CompatibilityOptions must be applied early in the process before any other Windows App
+SDK APIs are called, or right after initializing the Windows App Runtime.
+
+### Specifying CompatibilityOptions in the app's project file
+
+You can also specify the patch version and disabled changes in the app's project file instead of
```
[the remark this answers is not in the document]
codendone: Resolved by consistently renaming to "patch level"/"PatchLevel".

specs/Compatibility/CompatibilityOptions.md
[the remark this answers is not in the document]
codendone: Renamed everywhere to "PatchLevel"/"patch level"

specs/Compatibility/RuntimeCompatibilityOptions.md
```diff
@@ -0,0 +1,226 @@
...
+## RuntimeCompatibilityOptions.Apply
+
+Apply the set compatibility options to the runtime.
+
+RuntimeCompatibilityOptions must be applied early in the process before any other Windows App
+SDK APIs are called, or right after initializing the Windows App Runtime. The options must be set
+early before the configuration is locked for the lifetime of the process. Since the Windows App
+Runtime needs to run with a consistent configuration, it will lock the configuration when it needs
+to ensure the configuration will no longer change. Calling `Apply` will also lock the
+configuration.
```
oldnewthing, line 171: I think this text is trying to say "The Windows App Runtime will automatically lock the configuration at some point. Calling `Apply` also locks. Since you don't really know when the Windows App Runtime locks, you should call Apply as soon as possible, if you intend to call it at all."

specs/Compatibility/RuntimeCompatibilityOptions.md
```diff
@@ -0,0 +1,226 @@
...
+    enum RuntimeCompatibilityChange
+    {
+        None = 0, /// do not use this value
+
+        // Add real changes here:
+
+        // 1.7.1
+        // SampleApiCrashFix,
+        // OtherSampleApiCrashFix,
+        // ...
```
oldnewthing, line 193: What is the policy/process for retiring old RuntimeCompatibilityChange values?

is it that when (say) 1.5 goes out of support, then we comment out all the 1.5 compatibility values [but hold their values so they won't get reused]? Or do we leave them in place forever, even though they don't do anything?

What happens if we have
```cpp
enum RuntimeCompatibilityChange
{
  None = 0,
  // 1.7.1
  ResetPolarityOnWidgetChange, // does specifying this disable the reset?
  WidgetColorLeak, // specifying this restores the leak?

  // 1.8.1
  GadgetCrashWhenShownWhileDisabled,
}
```
and now we have a 1.7.2 compatibility change. Do we add it at the end, so the numbers go 1.7.1 to 1.8.1 to 1.7.2?

or do we reserve 100 change numbers between 1.7.1 and 1.8.1 so we can insert more 1.7.x values later?

specs/Compatibility/RuntimeCompatibilityOptions.md
```diff
@@ -0,0 +1,226 @@
...
+    /// The set of servicing changes that can be disabled.
+    enum RuntimeCompatibilityChange
+    {
+        None = 0, /// do not use this value
+
+        // Add real changes here:
+
+        // 1.7.1
+        // SampleApiCrashFix,
+        // OtherSampleApiCrashFix,
```
oldnewthing, line 192: Can you give a more realistic example of an enum name?

For example, suppose we made a change to reset the polarity of a widget when it changes. What would we call the enum that suppresses the auto-reset?

Suppose we made a change to fix a crash when a gadget is shown while it is disabled. What would we call the enum that re-enables the crash?

specs/Compatibility/RuntimeCompatibilityOptions.md
```diff
@@ -0,0 +1,226 @@
...
+    /// applied options.
+    runtimeclass RuntimeCompatibilityOptions
+    {
+        RuntimeCompatibilityOptions();
+
+        /// An optional patch level to use if the runtime version matches the major.minor version.
+        WindowsAppRuntimeVersion PatchLevel1 { get; set; };
+
+        /// An optional patch level to use if the runtime version matches the major.minor version.
+        WindowsAppRuntimeVersion PatchLevel2 { get; set; };
```
oldnewthing, line 216: Would we ever need a third patch level, if somebody wants to run on 3 different versions of the Windows App Runtime?

specs/Compatibility/RuntimeCompatibilityOptions.md
```diff
@@ -0,0 +1,226 @@
...
+
+Setting both patch levels for the same major.minor version, such as 1.7.3 and 1.7.4, is not allowed
+and will generate an error during `Apply()`.
+
+Instead of directly using this API, this patch level could be specified in the app's project file,
+like this:
+
+```xml
+  <PropertyGroup>
+    <WindowsAppSDKRuntimePatchLevel2>1.8.2</WindowsAppSDKRuntimePatchLevel2>
```
oldnewthing, line 130: Why not
```xml
<WindowsAppSDKRuntimePatchLevels>1.7.1,1.8.2</WindowsAppSDKRuntimePatchLevels>
```
?

specs/Compatibility/RuntimeCompatibilityOptions.md
```diff
@@ -0,0 +1,226 @@
...
+
+Optional patch level to use if the runtime version matches the major.minor version. If the runtime
+version does not match the specified major.minor version, this value is ignored.
+
+Instead of directly using this API, this patch level could be specified in the app's project file,
+like this:
+
+```xml
+  <PropertyGroup>
+    <WindowsAppSDKRuntimePatchLevel>1.7.3</WindowsAppSDKRuntimePatchLevel>
```
oldnewthing, line 105: What is our capitalization convention on `WindowsAppSDK`? By the standard capitalization rules, it would be `WindowsAppSdk`.

specs/Compatibility/RuntimeCompatibilityOptions.md
codendone: Correct.

specs/Compatibility/RuntimeCompatibilityOptions.md
codendone: The current plan is that only 1.7.x builds will contain the 1.7.x changes. When moving to 1.8.0, all entries (other than `None`) will be removed -- apps will need to evaluate compile errors at that time if they were disabling 1.7.x changes. The code for 1.8 will never include any containment checks from 1.7 (those containment checks stay in that version's release branch).

Regarding the numbers, it isn't shown here, but each enum will actually be assigned to the internal ADO ID of the servicing change. (Or some other unique identifier in the future, such as if we choose to ensure there are issues logged on GitHub for all servicing fixes and use those issue numbers.)

specs/Compatibility/RuntimeCompatibilityOptions.md
codendone: For those examples, I'd propose these names:
* WidgetPolarityReset
* ShowDisabledGadgetCrash

Using the list of fixes in 1.6.4, we might have used these names:
* MultiLineTextBoxSelection
* DDLMPackageInstall
* MissingDetoursExport
* CommandBarMenuPositioning
* OldGraphicsHardwareCrash
* PointerEventWindowCloseCrash
* UIAInitReentrancyCrash
* CompositionCapabilitiesCrash
* ExpandedUnicodeSupportForTextBoxes
* PackageManagerRegisterNewerHandling

We may need a separate discussion on the naming scheme to use.

specs/Compatibility/RuntimeCompatibilityOptions.md
codendone: No, for "A/B Forward" we specifically intend only to support two adjacent versions, such as for an app to move from 1.7 to 1.8.

specs/Compatibility/RuntimeCompatibilityOptions.md
codendone: The intent is to stay close to the API, which could then make the question the same as the question for line 216, where we want to limit to just two versions.

specs/Compatibility/RuntimeCompatibilityOptions.md
codendone: In the WinAppSDK .props/.targets files, everything currently uses capital "DK" except for a single `<WindowsAppSdkFoundation>` property which is for internal-only testing (and should ideally be reworked to not be in the official WinAppSDK package).

Issue #4923, opened 2024-11-28
Enhancing iconography for project and item templates


niels9001:
With the recent WinUI logo refresh, we need to update the project and item templates icon. 

- For all templates, we currently use the same icon: this makes it more difficult to recognize, at first glance, what the template is for. From an accessibility and cognitive load point of view, this requires the user to read the text/description. For novice users, a small icon could help understand what the template is for. And lastly, we do not differentiate between the coding language (C# vs. C++).
- “Custom Control” is not the standard that XAML developers are used to: Templated Control would be the better.
- Project template titles do not follow the WPF/UWP standard for their title: WPF Blank App, Universal Blank App vs. Blank App, Packaged (WinUI 3 in Desktop)
- Titles + descriptions are pretty verbose and can be shortened.
- The old WinUI icon is used vs. the new icon.

Therefore, we've worked together with the design team to create a consistent, yet unique icon set for project and item templates:

**Project templates**
Original (left) vs. proposed (right)
![Image](https://github.com/user-attachments/assets/7125b31d-210a-4901-abcd-48531f4668ac)

**Item templates**
Original (top) vs. proposed (bottom)
![Image](https://github.com/user-attachments/assets/d9ecec27-7e2b-4f7c-a540-45eb0ee4d5cd)

PR #4924, opened 2024-11-28
Updating project / item template icons and descriptions

niels9001:
Closes: #4923

The following changes are included:

**Updates to the Windows App SDK and WinUI logo**
The default Windows icon that was used to represent WASDK is now replaced with a dedicated Windows App SDK logo. The old WinUI logo has been updated to the latest version.

**Updates to project and item template titles and descriptions**
- Replacing `Windows UI Library (WinUI 3)` with `WinUI 3`, as this is the product naming we are using across our docs and communication.
- Following the WPF / UWP template naming conventions starting with the framework name and then the project title. E.g. `Blank App, Packaged (WinUI 3 in Desktop)` => `WinUI Blank App (Packaged)`
- Changing `Custom Control` to `Templated Control`, which is the correct name for this item template.
- Making descriptions less verbose.

**Updates to project and item template icons**
Instead of using the default WinUI logo for every template, we are now using individual icons following the Visual Studio design guidelines for iconography (and similar to what WPF and UWP have been using). This helps the developer to understand what the template is about and what language it's using.

Mock-up: original vs. new
image

Mock-up: original vs. new
image

PR #4986, opened 2024-12-19
Updating project / item template icons and descriptions

niels9001:
ORIGINAL PR (with comments + feedback): #4924

Closes: #4923

The following changes are included:

**Updates to the Windows App SDK and WinUI logo**
The default Windows icon that was used to represent WASDK is now replaced with a dedicated Windows App SDK logo. The old WinUI logo has been updated to the latest version.

**Updates to project and item template titles and descriptions**
- Replacing `Windows UI Library (WinUI 3)` with `WinUI 3`, as this is the product naming we are using across our docs and communication.
- Following the WPF / UWP template naming conventions starting with the framework name and then the project title. E.g. `Blank App, Packaged (WinUI 3 in Desktop)` => `WinUI Blank App (Packaged)`
- Changing `Custom Control` to `Templated Control`, which is the correct name for this item template.
- Making descriptions less verbose.

**Updates to project and item template icons**
Instead of using the default WinUI logo for every template, we are now using individual icons following the Visual Studio design guidelines for iconography (and similar to what WPF and UWP have been using). This helps the developer to understand what the template is about and what language it's using.

Mock-up: original vs. new
image

Mock-up: original vs. new
Frame 15


Lamparter:
"WinUI Blank App" isn't quite grammatically correct as it isn't the correct order of words. It should be "Blank WinUI App" to maintain readability and conciseness - [adj., (noun)] (Blank WinUI App) is the correct word order rather than the proposed pattern [noun, adj., noun] (WinUI Blank App)

jbgski:
Great work on the icons, they look fantastic, especially in the Visual Studio's redesigned UI. New project template names (e.g. WinUI Blank App) are nice too :)

However, as someone who does both UWP and WinUI dev at the same time, I'd really avoid removing the "(WinUI 3)" suffixes. The icons are not enough to separate between UWP and WinUI item types like Blank Page, especially when using search, and this will cause extra confusion for new developers

(also, there's a typo in the screenshot, maybe it's fixed already - Resources File says `.rews` instead of `.resw`)

Sergio0694:
> "I'd really avoid removing the "(WinUI 3)" suffixes"

We've updated all UWP templates in Visual Studio (17.13), so this is actually consistent. We will have:
- WinUI Blank App
- UWP Blank App
- UWP Blank App (.NET Native)

So it'll be pretty easy to tell them apart 🙂

---

@niels9001 I like the new icons, but they look very different than the ones for all other various project types in Visual Studio. Are all other default icons also being updated in Visual Studio to match this new thinner/rounded style? Otherwise I feel like they would look a bit out of place. It'd be nice if we could ensure they're all consistent with each other. Is this being considered?

niels9001:
> > "I'd really avoid removing the "(WinUI 3)" suffixes"
> 
> We've updated all UWP templates in Visual Studio (17.13), so this is actually consistent. We will have:
> 
> * WinUI Blank App
> * UWP Blank App
> * UWP Blank App (.NET Native)
> 
> So it'll be pretty easy to tell them apart 🙂
> 
> @niels9001 I like the new icons, but they look very different than the ones for all other various project types in Visual Studio. Are all other default icons also being updated in Visual Studio to match this new thinner/rounded style? Otherwise I feel like they would look a bit out of place. It'd be nice if we could ensure they're all consistent with each other. Is this being considered?

Correct, with the naming we tried to align to UWP/WPF as much as possible to be consistent and recognizable.

@Sergio0694 With regards to the icon style; these icons were designed by the VS design team. With the [recent Fluent Design update for VS2022](https://devblogs.microsoft.com/visualstudio/the-visual-studio-ui-refresh-preview-is-here/), iconography is being updated as well. My assumption is that the style that these new WinUI icons are using, is the new style for (new) icons going forward for VS.

jbgski:
> We've updated all UWP templates in Visual Studio (17.13), so this is actually consistent.

@Sergio0694 I was talking about file templates like "Blank Page", "Resource Dictionary". AFAIK those aren't cross-compatible with UWP due to the different namespaces in .xaml.cs files (WUX/MUX), so there could be some issues when someone accidentally creates a WinUI 3 Blank Page in an UWP project or vice versa

Sergio0694:
@jbgski you should only get compatible item templates showing up, based on project types. For UWP projects, you'll only get UWP item templates, and for WinUI projects you'll only get WinUI item templates. Not sure I'm following what the issue is here exactly. You mentioned that mixed scenarios would be confusing, but using both UWP and WinUI XAML in the same project is not a supported configuration (and it won't even work, because CsWinRT can only project one mode at a time).

jbgski:
@Sergio0694 Sorry if it wasn't clear, I wasn't talking about mixed project scenarios.

As you can see here in my UWP project, in the file template selector, when I search for "page" (strona) I see both UWP page templates and WinUI page templates:
image

If the (WinUI 3) text is removed from the item templates, the only difference between UWP and WinUI templates will be the icon, which isn't clear enough.

sylveon:
> you should only get compatible item templates showing up, based on project types. For UWP projects, you'll only get UWP item templates, and for WinUI projects you'll only get WinUI item templates.

This is definitely incorrect. Here's what I see when I try adding a new item to a C++ WinUI project.

image

The ones with the Windows logo are UWP templates. I made this exact mistake earlier today of adding a UWP item to a WinUI project accidentally. Notice that the UWP blank page is selected *by default*.

sylveon:
Here's the same for a brand new cppwinrt UWP app:
image

I also see WinUI 3 templates. @Sergio0694

zinniaa (approved):

assets/WindowsAppSDK240.svg
Saharsh979, line 1: In this file, the dimensions of the icon have changed, was this size change intentional?

niels9001: Yes. The name implies it's 240x240 but it wasn't. The new version does use those dimensions

dev/VSIX/ProjectTemplates/Desktop/CSharp/PackagedApp/BlankApp/WinUI.Desktop.Cs.BlankApp.vstemplate
```diff
@@ -2,8 +2,8 @@
 <VSTemplate Version="3.0.0" xmlns="http://schemas.microsoft.com/developer/vstemplate/2005" Type="Project">
   <TemplateData>
     <Hidden>true</Hidden>
-    <Name>Blank App (WinUI 3 in Desktop)</Name>
-    <Description>A project for creating a Desktop app based on the Windows UI Library (WinUI 3).</Description>
+    <Name>WinUI Blank App</Name>
```
npuvvada, line 5: Can we use resx file for Name, description in these templates as well ?

dev/VSIX/ProjectTemplates/Desktop/CppWinRT/PackagedApp/BlankApp/WinUI.Desktop.CppWinRT.BlankApp.vstemplate
```diff
@@ -2,8 +2,8 @@
 <VSTemplate Version="3.0.0" xmlns="http://schemas.microsoft.com/developer/vstemplate/2005" Type="Project">
   <TemplateData>
     <Hidden>true</Hidden>
-    <Name>Blank App (WinUI 3 in Desktop)</Name>
-    <Description>A project for creating a Desktop app based on the Windows UI Library (WinUI 3).</Description>
+    <Name>WinUI Blank App</Name>
```
npuvvada, line 5: Same here, Can we use resx file for Name, description in these templates as well ?

npuvvada (approved):
Looks good to me. Approving with minor suggestions for using resx for CS blankapp and cppwinrt blank app template

dev/VSIX/Extension/Cpp/Common/VSPackage.resx
```diff
@@ -118,31 +118,31 @@
...
-    <value>Blank Window (WinUI 3 in Desktop)</value>
+    <value>Blank Window</value>
   </data>
   <data name="1002" xml:space="preserve">
-    <value>[Experimental] Blank Window (WinUI 3 in Desktop)</value>
+    <value>[Experimental] Blank Window</value>
   </data>
   <data name="1003" xml:space="preserve">
-    <value>A blank window for Desktop apps based on the Windows UI Library (WinUI 3).</value>
+    <value>A blank window (WinUI 3).</value>
```
Scottj1s, line 127: we should be consistent - either "WinUI 3" or just "WinUI"

dev/VSIX/Extension/Cpp/Common/VSPackage.resx
```diff
@@ -151,58 +151,58 @@
...
-    <value>[Experimental] User Control (WinUI 3)</value>
+    <value>[Experimental] User Control</value>
   </data>
   <data name="1023" xml:space="preserve">
-    <value>A user control for apps based on the Windows UI Library (WinUI 3).</value>
+    <value>A blank user control (WinUI 3).</value>
   </data>
   <data name="1025" xml:space="preserve">
-    <value>Blank App, Packaged with Windows Application Packaging Project (WinUI 3 in Desktop)</value>
+    <value>WinUI Blank App (Packaged with Windows Application Packaging Project)</value>
```
Scottj1s, line 175: I like the preceding strings that just have WinUI in parentheses
```suggestion
    <value>Blank App (WinUI, Packaged with Windows Application Packaging Project)</value>
```

dev/VSIX/ProjectTemplates/Desktop/CSharp/PackagedApp/BlankApp/WinUI.Desktop.Cs.BlankApp.vstemplate
Scottj1s: good find. we should be using this construct for all item and project templates:
```
    <Name ID="XXX" Package="FIXME-PACKAGEGUID" />
    <Description ID="YYY" Package="FIXME-PACKAGEGUID" />

```
these three need to be localized:
ProjectTemplates/Desktop/CppWinRT/PackagedApp/BlankApp/WinUI.Desktop.CppWinRT.BlankApp.vstemplate
ProjectTemplates/Desktop/CSharp/PackagedApp/WapProj/WinUI.Desktop.Cs.WapProj.vstemplate
ProjectTemplates/Desktop/CppWinRT/PackagedApp/WapProj/WinUI.Desktop.CppWinRT.WapProj.vstemplate

niels9001: ProjectTemplates/Desktop/CSharp/PackagedApp/BlankApp/WinUI.Desktop.Cs.BlankApp.vstemplate too

niels9001: @npuvvada @Scottj1s Fixed!

dev/VSIX/ProjectTemplates/Desktop/CppWinRT/PackagedApp/BlankApp/WinUI.Desktop.CppWinRT.BlankApp.vstemplate
niels9001: Great catch. Fixed!

dev/VSIX/Extension/Cpp/Common/VSPackage.resx
Lamparter: Why not just WinUI App?

Scottj1s: I'm fine with either WinUI Foo ... and Foo ... (WinUI) - just would like consistency

Scottj1s (approved):

PR #4988, opened 2024-12-19
Remove MRT Core .props/.targets files

Sergio0694:
This PR removes the MRT Core .props/.targets files, as they're now in https://www.nuget.org/packages/Microsoft.Windows.SDK.BuildTools.MSIX.


0x5bfa:
You may want to remove these lines except native ones.

https://github.com/microsoft/WindowsAppSDK/blob/ad538768f704aeed816dcb794f3283829c6f18e7/BuildAll.ps1#L323-L330

manodasanW (approved):

PR #4994, opened 2024-12-20
Added TaskGroup property support to BackgroundTaskBuilder API

godlytalias:
TaskGroup property support has been added to the BackgroundTaskBuilder API.

### DIFF
--- a/dev/BackgroundTask/BackgroundTaskBuilder/BackgroundTaskBuilder.idl
+++ b/dev/BackgroundTask/BackgroundTaskBuilder/BackgroundTaskBuilder.idl
@@ -24,6 +24,7 @@ namespace Microsoft.Windows.ApplicationModel.Background
             );
 
         String Name{ set; get; };
+        Windows.ApplicationModel.Background.BackgroundTaskRegistrationGroup TaskGroup{ set; get; };
 
         [return_name("task")]
             Windows.ApplicationModel.Background.BackgroundTaskRegistration Register();
--- a/specs/BackgroundTask/BackgroundTaskBuilder.md
+++ b/specs/BackgroundTask/BackgroundTaskBuilder.md
@@ -105,6 +105,7 @@ namespace Microsoft.Windows.ApplicationModel.Background
             );
 
         String Name{ set; get; };
+        Windows.ApplicationModel.Background.BackgroundTaskRegistrationGroup TaskGroup { set; get; };
 
         Windows.ApplicationModel.Background.BackgroundTaskRegistration Register(
             );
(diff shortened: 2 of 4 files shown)


AjitSurana:
@godlytalias, please elaborate the purpose of the TaskGroup property.

godlytalias:
> @godlytalias, please elaborate the purpose of the TaskGroup property.

Its used for grouping the background tasks. So that apps can get all the tasks under that particular group using AllTasks API (https://learn.microsoft.com/en-us/uwp/api/windows.applicationmodel.background.backgroundtaskregistrationgroup.alltasks?view=winrt-26100)

dev/BackgroundTask/BackgroundTaskBuilder/BackgroundTaskBuilder.idl
```diff
@@ -24,6 +24,7 @@ namespace Microsoft.Windows.ApplicationModel.Background
             );
 
         String Name{ set; get; };
+        Windows.ApplicationModel.Background.BackgroundTaskRegistrationGroup TaskGroup{ set; get; };
```
Hemantxk, line 27: Are we going to expose other properties as well from UWP's BackgroundTaskBuilder? Like we have exposed AddCondition method but do not allow modifications to CancelOnConditionLoss property. What's the plan for other remaining properties.

specs/BackgroundTask/BackgroundTaskBuilder.md
```diff
@@ -105,6 +105,7 @@ namespace Microsoft.Windows.ApplicationModel.Background
             );
 
         String Name{ set; get; };
+        Windows.ApplicationModel.Background.BackgroundTaskRegistrationGroup TaskGroup { set; get; };
```
Hemantxk, line 108: Do we need another round of API review for this, or the prior one is sufficient.

dev/BackgroundTask/BackgroundTaskBuilder/BackgroundTaskBuilder.idl
godlytalias: CancelOnConditionLoss will not help as the win32 process won't be killed if condition is lost. If required, apps can have listeners and close win32 process if required. Other properties are also not meaningful for COM BG Task.

specs/BackgroundTask/BackgroundTaskBuilder.md
godlytalias: This one was in spec earlier also, just implementation was missing

AjitSurana:
@godlytalias, One follow up question - is it possible the same background task registration group has both Windows and Microsoft namespace background tasks? If so, AllTasks sort of properties return correct value?

godlytalias:
> @godlytalias, One follow up question - is it possible the same background task registration group has both Windows and Microsoft namespace background tasks? If so, AllTasks sort of properties return correct value?

It returns tasks registered by both namespace APIs as internally WinAppSDK API is calling platform SDK API only. Id, Name etc.. of the tasks will be the same

AjitSurana (approved):

Hemantxk (approved):

PR #5003, opened 2025-01-03
Switch auto initializer targets to add via BeforeClCompileTargets

codendone:
Add auto initializer targets via `<BeforeClCompileTargets>` (rather than `BeforeTargets="ClCompile"`) to get better timing for picking up added compiler options. This helps to ensure they are added after dynmically-added compiler options, such as `/arm64ec`.

A separate `AddUndockedRegFreeWinRTCppDefines` target was created to ensure those global defines are still added with the same timing they had previously. (I'm not sure if those global defines are required, but that is a question for another day.)

### DIFF
--- a/build/NuSpecs/WindowsAppSDK-Nuget-Native.Bootstrap.targets
+++ b/build/NuSpecs/WindowsAppSDK-Nuget-Native.Bootstrap.targets
@@ -1,6 +1,6 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
-    <Target Name="GenerateBootstrapCpp" BeforeTargets="ClCompile">
+    <Target Name="GenerateBootstrapCpp">
         <ItemGroup>
             <ClCompile Include="$(MSBuildThisFileDirectory)..\..\include\MddBootstrapAutoInitializer.cpp">
                 <PrecompiledHeader>NotUsing</PrecompiledHeader>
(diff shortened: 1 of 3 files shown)


Scottj1s (approved):

build/NuSpecs/WindowsAppSDK-Nuget-Native.Bootstrap.targets
```diff
@@ -1,6 +1,6 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
-    <Target Name="GenerateBootstrapCpp" BeforeTargets="ClCompile">
+    <Target Name="GenerateBootstrapCpp">
         <ItemGroup>
             <ClCompile Include="$(MSBuildThisFileDirectory)..\..\include\MddBootstrapAutoInitializer.cpp">
                 <PrecompiledHeader>NotUsing</PrecompiledHeader>
```
Scottj1s, line 6: PrecompiledHeader=NotUsing? was this a workaround for racy addition of compile flags? can we remove it?

Scottj1s: Oh I remember this - it was a compiler error based on dependencies.  We should keep it.

codendone: I think it is because we don't know what the name of the precompiled header is, if any, to #include the right file in the cpp, particularly for these pre-defined, non-generated files.

PR #5065, opened 2025-01-23
Onboard Static Validation Tests 

kythant:
These set of tests will check for whether experimental attributes are present in a stable and whether all the binaries are correctly signed in the TransportPackage for an Official build.

The code in test/StaticValidationTests/PackageInspectionTest.csproj is copied from the Aggregator

Validated:
Success case with experimental mode
https://microsoft.visualstudio.com/ProjectReunion/_build/results?buildId=114733850&view=results

Failure case with stable mode but with experimental APIs
https://microsoft.visualstudio.com/ProjectReunion/_build/results?buildId=114667865&view=results


build/AzurePipelinesTemplates/WindowsAppSDK-Publish-Stage.yml
```diff
@@ -9,7 +9,7 @@ stages:
 - stage: Publish
   dependsOn:
   - Test
-  - Pack
```
alexlamtest, line 12: It feels a bit non-intuitive that "Publish" can happen without waiting for "Pack". Maybe add a comment to elaborate?

alexlamtest: Oh, wait, StaticValidationTest now depends on Pack for Publish no longer needs to wait for Pack directly. I see.

kythant: Yea let me rework this a little

kythant: This was before
image

This is the picture now after the latest commit
image

build/AzurePipelinesTemplates/WindowsAppSDK-StaticValidationTest-Stage.yml
```diff
@@ -0,0 +1,118 @@
...
+          }
+          else
+          {
+              Write-Host "Error: Found more than one Foundation TransportPackage Found"
+              Write-Host "Please make sure only one package is present"
+              Write-Host "##vso[task.complete result=Failed;]DONE"
+          }
+
+    - task: VSBuild@1
+      displayName: 'Build PackageInspectionTest.sln'
```
alexlamtest, line 70: Nonblocking suggestion: consider adding "(restore)" to distinguish between these 2 tasks?

kythant: Good idea

test/StaticValidationTests/Program.cs
```diff
@@ -0,0 +1,297 @@
...
+                {
+                    ValidationTestCommon.LogLine("ERROR: SignTool verify returned error.");
+                    ValidationTestCommon.LogLine($"SignTool command line: signtool {arguments}");
+                    ValidationTestCommon.LogLine("SignTool output:");
+                    ValidationTestCommon.LogLine("=====================");
+                    ValidationTestCommon.LogLine(string.Join("\r\n", cmd.StandardOut));
+                    ValidationTestCommon.LogLine("=====================");
+                }
+
+                if (extraValidation != null)
```
alexlamtest, line 254: Is this block applicable even when the digital signature is invalid?

test/StaticValidationTests/Program.cs
```diff
@@ -0,0 +1,297 @@
...
+                {
+                    ValidationTestCommon.LogLine(file);
+                    signError |= CheckDigitalSignature(options.SignTool, file);
+                    isNotDeterministic |= !CheckDeterministicProperty(file);
+                    ValidationTestCommon.LogLine("");
+                }
+
+                foreach (var validation in CertValidations)
+                {
+                    if (validation.ValidationCount == 3) // 3 times because we are checking for amd64, x86 and arm64.
```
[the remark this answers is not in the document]
kythant: Currently not enabled for Foundation yet so this should still be 3

test/StaticValidationTests/Program.cs
kythant: So this is a vestigial from the Aggregator code where it had to check for different signatures for Microsoft.Internal.FrameworkUdk.dll but we don't need that in Foundation hence CertValidations in line 63 is empty.

But I am keeping this logic just incase.

PR #5070, opened 2025-01-26
Removing terminal velocity from CameraCaptureUI

Saharsh979:
This PR removes the TerminalVelocity mechanism from the CameraCaptureUI API, which was previously used to control its visibility across different release channels (Preview, Stable, Experimental). As the feature is now moving towards broader availability (**Preview Release**) and no longer needs the experimental feature flagging, we are removing the TerminalVelocity-related checks.

The API has gone through spec-review:  https://github.com/microsoft/WindowsAppSDK/pull/4721

### DIFF
--- a/test/CameraCaptureUITests/APITests.cpp
+++ b/test/CameraCaptureUITests/APITests.cpp
@@ -9,7 +9,6 @@
 #include <mutex>
 #include <winrt/Windows.Storage.h>
 #include <winrt/Windows.Foundation.h>
-#include <TerminalVelocityFeatures-CameraCaptureUI.h>
 
 using namespace std::chrono_literals;
 using namespace WEX::Common;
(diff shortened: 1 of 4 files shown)


test/CameraCaptureUITests/APITests.cpp
```diff
@@ -9,7 +9,6 @@
 #include <mutex>
 #include <winrt/Windows.Storage.h>
 #include <winrt/Windows.Foundation.h>
-#include <TerminalVelocityFeatures-CameraCaptureUI.h>
```
Hemantxk, line 12: Can you confirm there are no changes in this header that needs to be checked-in?

Hemantxk (approved):

hiteshkrmsft (approved):

test/CameraCaptureUITests/APITests.cpp
Hemantxk: Now that I think of it, we can remove the xml, header and pipeline entry once this feature is out in stable release.

PR #5091, opened 2025-01-31
Adding IXP & FrameworkUDK to Central Package Management

Saharsh979:
The hardcoded versions are going to get stale since Maestro isnt able to update these!

This PR adds FrameworkUDK and IXP to Central Package Management by adding them to Directory.Packages.props. Additionally, hardcoded package references have been removed from packages.config to align with best practices.

https://task.ms/55990470

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -6,7 +6,8 @@
 -->
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <ItemGroup>
-    <PackageVersion Include="Microsoft.ProjectReunion.InteractiveExperiences.TransportPackage" Version="1.8.0-CI-26107.1720.241003-1631.3" />
+    <PackageVersion Include="Microsoft.ProjectReunion.InteractiveExperiences.TransportPackage" Version="$(MicrosoftProjectReunionInteractiveExperiencesTransportPackagePackageVersion)" />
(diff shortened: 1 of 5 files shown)


kythant (approved):

hiteshkrmsft (approved):

akanpatel2206 (approved):

kythant:
Please cherry-pick this change to release/1.7-stable but please use the existing version numbers from the release/1.7-stable branch.

There is also a matter of reliably restoring these packages which involve creating a barebones csproj that is only there to do the restore for us and it should probably be in DevCheck.ps1.

We don't want a random csproj to be restoring these packages for a vcxproj

kythant:
> Please cherry-pick this change to release/1.7-stable but please use the existing version numbers from the release/1.7-stable branch.
> 
> There is also a matter of reliably restoring these packages which involve creating a barebones csproj that is only there to do the restore for us and it should probably be in DevCheck.ps1.
> 
> We don't want a random csproj to be restoring these packages for a vcxproj. This maybe happening by happenstance if the validation succeeds. Please do so in this PR or in the next PR.

Looks like both
eng\PackageReference\FrameworkUdk\Microsoft.FrameworkUdk.PackageReference.csproj
eng\PackageReference\IXP\IXP.TransportPackage.PackageReference.csproj
covered these already.

Your PR is good to go

PR #4949, opened 2024-12-10
Reapply "Delegate to OS RegFreeWinRT when available (>=24H1) (#4728)" (#4935)

DrusTheAxe:
Delegate to OS RegFreeWinRT when available (>=24H2)

Don't detour APIs to our Undocked Reg-Free WinRT implementation if the OS supports Reg-Free WinRT (>=24H2)

https://task.ms/47326505

* WinAppSDK Dynamic Dependencies delegates to OS Dynamic Dependencies on Win11 (>=24H2)
* Updated and expanded Dynamic Dependencies test suite
* Undocked RegFree WinRT (URFW) disabled when WinAppSDK Dynamic Dependencies delegates to OS Dynamic Dependencies (in this case OS RegFree WinRT applies)
* NOTE: This required changing some test package names e.g. foo-x.y => foo.x.y
* Merge with latest in main
* `DevCheck -SyncDependencies` to fixup dependencies
* Updated `Microsoft.Taef dependency` to 10.95.240918004
* Added `Microsoft.Windows.SDK.BuildTools` to master dependency list (`eng\Version.*.xml`)
* Fixed bad project files (`WindowsAppRuntime_UniversalBGTaskDLL.vcxproj`, `BackgroundTaskTests.vcxproj`
, `OAuth2ManagerTests.vcxproj`, `OAuthTestApp.vcxproj` e.g. `<Import...Microsoft.Cpp.Default.props>` at top of file (line 3), use `$(NugetPackageDirectory)` (not .. path'ing), use dependency `$(...Version)` macros (not hardcoded version numbers)
* Fixed WindowsAppRuntime.sln wasn't building some test MSIX packages on ARM64
* Fixed WindowsAppRuntime.sln some tests missing Build Dependencies on test package projects they depend on
* DevCheck: minor bug fixes
* Pipeline: `WindowsAppSDK-RuntTests-Steps.yml` adds 'Dump services' task for troubleshooting
* Pipeline: Add `-CheckVisualStudio` to build and test phases (`WindowsAppSDK-SetupBuildEnvironment-Steps.yml` + `WindowsAppSDK-RuntTests-Steps.yml`)
* GettingStarted docs: Updated Win11 24H2 SDK link

This reverts commit c0883363d81752ed030c6cb23fedaaca0f3d74b2.

Revert "Revert "Delegate to OS RegFreeWinRT when available (>=24H1) (#4728)""

[...]

PR #5106, opened 2025-02-06
Bootstrapper formatting bad DDLM package name in FindDDLMViaEnumeration()

DrusTheAxe:
Bootstrapper built bad DDLM package name in FindDDLMViaEnumeration() thus looking for DDLM packages named `...ddlm-version...` rather than `...ddlm.version...`

Local tests passed as they were named with dash (-) but product DDLMs are named with dot (.) causing toplevel tests using production packages to fail. Fixed the Bootstrapper and updated local tests to use naming consistent with product to prevent future confusion

Problem repro'd using nightly winappsdk nuget and locally built tests (manually running tests plus walking through debugger). Replaced the local tests' bootstrapper DLL with the fixed one built per this PR and verified success as expected (manually run tests plus walking through debugger).

This PR complements [Reapply "Delegate to OS RegFreeWinRT when available (>=24H1) (#4728)" (#4935) #4949](https://github.com/microsoft/WindowsAppSDK/pull/4949)

NOTE: This change was previously in #4949 but lost through a bad merge.


kythant (approved):

joshusto (approved):

cwruss (approved):
LGTM

PR #5110, opened 2025-02-06
[Cherry-pick to 1.7] Removed TerminalVelocity usages from IDL

godlytalias:
Cherry-pick of #5107

### DIFF
--- a/dev/BackgroundTask/BackgroundTaskBuilder/BackgroundTaskBuilder.idl
+++ b/dev/BackgroundTask/BackgroundTaskBuilder/BackgroundTaskBuilder.idl
@@ -1,6 +1,5 @@
 // Copyright (c) Microsoft Corporation and Contributors.
 // Licensed under the MIT License.
-#include <TerminalVelocityFeatures-BackgroundTask.h>
 
 namespace Microsoft.Windows.ApplicationModel.Background
 {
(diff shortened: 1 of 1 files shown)


codendone (approved):

DrusTheAxe:
If  TerminalVelocityFeatures-BackgroundTask.h is obsolete do you plan to delete dev\common\TerminalVelocityFeatures-BackgroundTask.xml and .h ?

godlytalias:
> If TerminalVelocityFeatures-BackgroundTask.h is obsolete do you plan to delete dev\common\TerminalVelocityFeatures-BackgroundTask.xml and .h ?

@DrusTheAxe As of now, the usage is still there in implementation files, and yeah it need to be cleaned up. Should it be just before the stable release? Or is there any guideline?

DrusTheAxe:
> @DrusTheAxe As of now, the usage is still there in implementation files, and yeah it need to be cleaned up. Should it be just before the stable release? Or is there any guideline?

Ideally 'dead code' is cleaned up when it becomes 'dead'. As the say, the best time to act is yesterday. The 2nd best time to act is today :-)

You can make the change at any time in main. Whether it's worthy to bring to the vNext release branch before stable (.0) release is a question for @codendone, shiproom, office hours, etc.

If inert it's not worthy to bring to release/x.y-stable servicing updates. If it's harmful then it's a bug worthy of servicing (though in this case it's hard to see how it would a problematic worthy of servicing).

@codendone would know about documented guidelines

codendone:
I'm not aware of any written guidance currently about cleanup.

+1 to the comments from @DrusTheAxe on risky/benefit. Since this is just cleanup which should not cause any issues for 1.7, I suggest doing the cleanup only in main--it isn't worth the cherry-pick and additional churn in the release branch.

Issue #4844, opened 2024-11-02
StorageFile.GetFileFromPathAsync


hu13:
I am have been trying to debug an issue i have with StorageFile.GetFileFromPathAsync.

In my use case, i am accessing this API via a winrt-go [(https://github.com/hu13/winrt-go).](https://github.com/hu13/winrt-go/blob/hangkun/poc/windows/storage/storagefile.go#L262)

I have a test that creates the test file in a os.TempDir's location which belongs in: `~/AppData/Local/Temp/myDir/myTestFile.txt`, then i call the getfilepath api and it keeps returning UnauthorizedAccessException error aka 0x80070005.
Some stacktrace:
Exception 0xc0000005 0x1 0x50 0x7ffe99921c58
PC=0x7ffe99921c58

runtime.cgocall(0x795080, 0xc00007a330)
	C:/hostedtoolcache/windows/go/1.23.2/x64/src/runtime/cgocall.go:167 +0x3e fp=0xc000043c60 sp=0xc000043bf8 pc=0x78439e
syscall.SyscallN(0xc00003caa0?, {0xc00020e318?, 0xc00013f160?, 0xb4f765?})
	C:/hostedtoolcache/windows/go/1.23.2/x64/src/runtime/syscall_windows.go:519 +0x46 fp=0xc000043c80 sp=0xc000043c60 pc=0x78f506
github.com/saltosystems/winrt-go/windows/storage.StorageFileGetFileFromPathAsync({0xc00003caa0, 0x53})

This only happens in github action VM that runs windows server 2022.  The same test passes on my physical windows PC.
From my understanding, the StorageApi should have access to [TemporaryFolder](https://learn.microsoft.com/en-us/uwp/api/windows.storage.applicationdata.temporaryfolder) by default.

Any ideas? Thanks

whiskhub:
The `ApplicationData.TemporaryFolder` property only works in (MSIX) packaged apps. It returns a special temporary directory for your MSIX package, like `%LOCALAPPDATA%\Packages\PackageName_CompanyName\TempState`.

But you seem to have an issue with `StorageFile.GetFileFromPathAsync`. I know on Windows Server, there is often just one Administrator user that is always elevated. And these WinRT APIs often have difficulties with elevated users. But the StorageFile APIs should normally work.

[...]

PR #5155, opened 2025-02-14
Public Spec for Microsoft.Windows.Storage.Pickers

DinahK-2SO:
The public Spec for `Microsoft.Windows.Storage.Pickers`

Preview: [Microsoft.Windows.Storage.Pickers.md](https://github.com/microsoft/WindowsAppSDK/blob/user/DinahK-2SO/storage_pickers_public_spec/specs/StoragePickers/Microsoft.Windows.Storage.Pickers.md)

The code is also open for review. Find PR: #5240

### DIFF
--- a/specs/StoragePickers/FileOpenPicker.md
+++ b/specs/StoragePickers/FileOpenPicker.md
@@ -11,16 +11,25 @@ Supports specifying the initial location, extension filters, and text on commit 
 
 # API Pages
 
+## Definition
+
+```C#
+runtimeclass FileOpenPicker
+{
+    FileOpenPicker(Microsoft.UI.WindowId windowId);
+
+    string CommitButtonText;
+    string SettingsIdentifier;
+    IVector<string> FileTypeFilter{ get; };
+    PickerLocationId SuggestedStartLocation;
+    PickerViewMode ViewMode;
+
+    Windows.Foundation.IAsyncOperation<PickFileResult> PickSingleFileAsync();
+    Windows.Foundation.IAsyncOperation<IVectorView<PickFileResult>> PickMultipleFilesAsync();
+}
+```
+
 ## Constructor
-
-### Attributes
-
-| **Attribute**            | **Type**           | **Description**                                                          |
-|--------------------------|--------------------|--------------------------------------------------------------------------|
-| `CommitButtonText`       | `string`           | Gets or sets the text displayed on the commit button of the file picker. |
-| `FileTypeFilter`         | `IList<string>`    | Gets the collection of file types that the file picker displays.         |
-| `SuggestedStartLocation` | [Microsoft.Windows.Storage.Pickers.PickerLocationId](./PickerLocationId.md)| Gets or sets the initial location where the file picker looks for files.  |
-| `ViewMode`               | [Microsoft.Windows.Storage.Pickers.PickerViewMode](./PickerViewMode.md)    | Gets or sets the view mode that the file picker is using to present items.|
 
 ### Examples
 C#
--- a/specs/StoragePickers/FileOpenPicker.md
+++ b/specs/StoragePickers/FileOpenPicker.md
@@ -1,0 +1,172 @@
+FileOpenPicker Class
+===
+
+# Background
+
+Namespace: [Microsoft.Windows.Storage.Pickers](./Microsoft.Windows.Storage.Pickers.md)
+
+Represents a UI element that lets the user choose and open files.
+
+Supports specifying the initial location, extension filters, and text on commit button.
+
+# API Pages
+
+## Constructor
+
+### Attributes
+
+| **Attribute**            | **Type**                                                | **Description**                                                          |
+|--------------------------|---------------------------------------------------------|--------------------------------------------------------------------------|
+| `ViewMode`               | [Microsoft::Windows::Storage::Pickers::PickerViewMode](./PickerViewMode.md)  | Gets or sets the view mode that the file picker is using to present items. |
+| `SuggestedStartLocation` | [Microsoft::Windows::Storage::Pickers::PickerLocationId](./PickerLocationId.md)| Gets or sets the initial location where the file picker looks for files. |
+| `CommitButtonText`       | `winrt::hstring`                                        | Gets or sets the text displayed on the commit button of the file picker. |
+| `FileTypeFilter`         | `Windows::Foundation::Collections::IVector<hstring>`    | Gets the collection of file types that the file picker displays.         |
+
+### Examples
+C#
+
+```csharp
+using Microsoft.Windows.Storage.Pickers;
+
+var openPicker = new FileOpenPicker(this.AppWindow.Id)
+{
+    // (Optional) specify the initial location. If not specified, use system default:
+    SuggestedStartLocation = PickerLocationId.DocumentsLibrary,
+    
+    // (Optional) specify the text displayed on commit button. If not specified, use system default:
+    CommitButtonText = "Choose selected files",
+
+    // (Optional) specify file extensions filters. If not specified, default to all (*.*)
+    FileTypeFilter = { ".txt", ".pdf", ".doc", ".docx" },
+};
+```
+
+C++
+
+```cpp
+#include <winrt/Microsoft.Windows.Storage.Pickers.h>
+using namespace winrt::Microsoft::Windows::Storage::Pickers;
+
+FileOpenPicker openPicker(AppWindow().Id());
+
+// (Optional) specify the initial location. If not specified, use system default:
+openPicker.SuggestedStartLocation(PickerLocationId::DocumentsLibrary);
+
+// (Optional) specify the text displayed on commit button. If not specified, use system default:
+openPicker.CommitButtonText(L"Choose selected files");
+
+// (Optional) specify file extensions filters. If not specified, default to all (*.*)
+openPicker.FileTypeFilter().Append(L".txt");
+openPicker.FileTypeFilter().Append(L".pdf");
+openPicker.FileTypeFilter().Append(L".doc");
+openPicker.FileTypeFilter().Append(L".docx");
+```
+
+## FileOpenPicker.PickSingleFilesAsync
+
+Displays a UI element that allows user to choose and open one file.
+
+### Definition
+```cpp
+winrt::Windows::Foundation::IAsyncOperation<winrt::Windows::Storage::StorageFile> PickSingleFileAsync();
+```
+Return null if the file dialog was cancelled or closed without selection.
+
+### Examples
+
+C#
+
+```csharp
+using Microsoft.Windows.Storage.Pickers;
+
+var openPicker = new FileOpenPicker(this.AppWindow.Id);
+
+var file = await openPicker.PickSingleFileAsync();
+if (file is not null)
+{
+    var content = System.IO.File.ReadAllText(file.Path);
+}
+else
+{
+    // error handling.
+}
+```
+
+C++
+```cpp
+#include <winrt/Microsoft.Windows.Storage.Pickers.h>
+#include <fstream>
+#include <string>
+using namespace winrt::Microsoft::Windows::Storage::Pickers;
+
+FileOpenPicker openPicker(AppWindow().Id());
+auto& file = co_await openPicker.PickSingleFileAsync();
+if (file != nullptr)
+{
+    std::ifstream fileReader(file.Path().c_str());
+    std::string text((std::istreambuf_iterator<char>(fileReader)), std::istreambuf_iterator<char>());
+    winrt::hstring hText = winrt::to_hstring(text);
+}
+else
+{
+    // error handling.
+}
+```
+
+## FileOpenPicker.PickMultipleFilesAsync
+
+Displays a UI element that allows user to choose and open multiple files.
+
+### Definition
+```cpp
+winrt::Windows::Foundation::IAsyncOperation<winrt::Windows::Foundation::Collections::IVectorView<winrt::Windows::Storage::StorageFile>> PickMultipleFilesAsync();
+```
+Return an empty list (Count = 0) if the file dialog's cancelled or closed.
+
+### Examples
+
+C#
+
+```csharp
+using Microsoft.Windows.Storage.Pickers;
+
+var openPicker = new FileOpenPicker(this.AppWindow.Id);
+
+var files = await openPicker.PickMultipleFilesAsync();
+if (files is not null)
+{
+    var pickedFilePaths = files.Select(f => f.Path);
+    foreach (var path in pickedFilePaths)
+    {
+        var content = System.IO.File.ReadAllText(path);
+    }
+}
+else
+{
+    // error handling.
+}
+```
+
+C++
+```cpp
+#include <winrt/Microsoft.Windows.Storage.Pickers.h>
+#include <fstream>
+#include <string>
+using namespace winrt::Microsoft::Windows::Storage::Pickers;
+
+FileOpenPicker openPicker(AppWindow().Id());
+auto& files = co_await openPicker.PickMultipleFilesAsync();
+if (files.Size() > 0)
+{
+    for (auto const& file : files)
+    {
+        std::ifstream fileReader(file.Path().c_str());
+        std::string text((std::istreambuf_iterator<char>(fileReader)), std::istreambuf_iterator<char>());
+        winrt::hstring hText = winrt::to_hstring(text);
+    }
+}
+else
+{
+    // error handling.
+}
+```
--- a/specs/StoragePickers/FileOpenPicker.md
+++ b/specs/StoragePickers/FileOpenPicker.md
@@ -15,25 +15,26 @@ Supports specifying the initial location, extension filters, and text on commit 
 
 ### Attributes
 
-| **Attribute**            | **Type**                                                | **Description**                                                          |
-|--------------------------|---------------------------------------------------------|--------------------------------------------------------------------------|
-| `ViewMode`               | [Microsoft::Windows::Storage::Pickers::PickerViewMode](./PickerViewMode.md)  | Gets or sets the view mode that the file picker is using to present items. |
-| `SuggestedStartLocation` | [Microsoft::Windows::Storage::Pickers::PickerLocationId](./PickerLocationId.md)| Gets or sets the initial location where the file picker looks for files. |
-| `CommitButtonText`       | `winrt::hstring`                                        | Gets or sets the text displayed on the commit button of the file picker. |
-| `FileTypeFilter`         | `Windows::Foundation::Collections::IVector<hstring>`    | Gets the collection of file types that the file picker displays.         |
+| **Attribute**            | **Type**           | **Description**                                                          |
+|--------------------------|--------------------|--------------------------------------------------------------------------|
+| `CommitButtonText`       | `string`           | Gets or sets the text displayed on the commit button of the file picker. |
+| `FileTypeFilter`         | `IList<string>`    | Gets the collection of file types that the file picker displays.         |
+| `SuggestedStartLocation` | [Microsoft.Windows.Storage.Pickers.PickerLocationId](./PickerLocationId.md)| Gets or sets the initial location where the file picker looks for files.  |
+| `ViewMode`               | [Microsoft.Windows.Storage.Pickers.PickerViewMode](./PickerViewMode.md)    | Gets or sets the view mode that the file picker is using to present items.|
 
 ### Examples
 C#
 
-```csharp
+```C#
 using Microsoft.Windows.Storage.Pickers;
 
 var openPicker = new FileOpenPicker(this.AppWindow.Id)
 {
-    // (Optional) specify the initial location. If not specified, use system default:
+    // (Optional) specify the initial location.
+    //     If not specified, using PickerLocationId.Unspecified by default.
     SuggestedStartLocation = PickerLocationId.DocumentsLibrary,
     
-    // (Optional) specify the text displayed on commit button. If not specified, use system default:
+    // (Optional) specify the text displayed on commit button. If not specified, use system default.
     CommitButtonText = "Choose selected files",
 
     // (Optional) specify file extensions filters. If not specified, default to all (*.*)
@@ -43,16 +44,17 @@ var openPicker = new FileOpenPicker(this.AppWindow.Id)
 
 C++
 
-```cpp
+```C++
 #include <winrt/Microsoft.Windows.Storage.Pickers.h>
 using namespace winrt::Microsoft::Windows::Storage::Pickers;
 
 FileOpenPicker openPicker(AppWindow().Id());
 
-// (Optional) specify the initial location. If not specified, use system default:
+// (Optional) specify the initial location.
+//     If not specified, using PickerLocationId.Unspecified by default.
 openPicker.SuggestedStartLocation(PickerLocationId::DocumentsLibrary);
 
-// (Optional) specify the text displayed on commit button. If not specified, use system default:
+// (Optional) specify the text displayed on commit button. If not specified, use system default.
 openPicker.CommitButtonText(L"Choose selected files");
 
 // (Optional) specify file extensions filters. If not specified, default to all (*.*)
@@ -62,21 +64,19 @@ openPicker.FileTypeFilter().Append(L".doc");
 openPicker.FileTypeFilter().Append(L".docx");
 ```
 
-## FileOpenPicker.PickSingleFilesAsync
+## FileOpenPicker.PickSingleFileAsync
 
 Displays a UI element that allows user to choose and open one file.
 
-### Definition
-```cpp
-winrt::Windows::Foundation::IAsyncOperation<winrt::Windows::Storage::StorageFile> PickSingleFileAsync();
-```
+Returns the picked file.
+
 Return null if the file dialog was cancelled or closed without selection.
 
 ### Examples
 
 C#
 
-```csharp
+```C#
 using Microsoft.Windows.Storage.Pickers;
 
 var openPicker = new FileOpenPicker(this.AppWindow.Id);
@@ -93,15 +93,15 @@ else
 ```
 
 C++
-```cpp
+```C++
 #include <winrt/Microsoft.Windows.Storage.Pickers.h>
 #include <fstream>
 #include <string>
 using namespace winrt::Microsoft::Windows::Storage::Pickers;
 
 FileOpenPicker openPicker(AppWindow().Id());
-auto& file = co_await openPicker.PickSingleFileAsync();
-if (file != nullptr)
+auto& file{ co_await openPicker.PickSingleFileAsync() };
+if (file)
 {
     std::ifstream fileReader(file.Path().c_str());
     std::string text((std::istreambuf_iterator<char>(fileReader)), std::istreambuf_iterator<char>());
@@ -117,17 +117,15 @@ else
 
 Displays a UI element that allows user to choose and open multiple files.
 
-### Definition
-```cpp
-winrt::Windows::Foundation::IAsyncOperation<winrt::Windows::Foundation::Collections::IVectorView<winrt::Windows::Storage::StorageFile>> PickMultipleFilesAsync();
-```
-Return an empty list (Count = 0) if the file dialog's cancelled or closed.
+Returns a collection of picked files.
+
+Return an empty list (Count = 0) if the file dialog's cancelled or closed without selection.
 
 ### Examples
 
 C#
 
-```csharp
+```C#
 using Microsoft.Windows.Storage.Pickers;
 
 var openPicker = new FileOpenPicker(this.AppWindow.Id);
@@ -148,14 +146,14 @@ else
 ```
 
 C++
-```cpp
+```C++
 #include <winrt/Microsoft.Windows.Storage.Pickers.h>
 #include <fstream>
 #include <string>
 using namespace winrt::Microsoft::Windows::Storage::Pickers;
 
 FileOpenPicker openPicker(AppWindow().Id());
-auto& files = co_await openPicker.PickMultipleFilesAsync();
+auto& files{ co_await openPicker.PickMultipleFilesAsync() };
 if (files.Size() > 0)
 {
     for (auto const& file : files)
--- a/specs/StoragePickers/FileOpenPicker.md
+++ b/specs/StoragePickers/FileOpenPicker.md
@@ -68,7 +68,7 @@ openPicker.FileTypeFilter().Append(L".docx");
 
 Displays a UI element that allows user to choose and open one file.
 
-Returns the picked file.
+Returns a light weight object that has the path of the picked file.
 
 Return null if the file dialog was cancelled or closed without selection.
 
@@ -81,10 +81,10 @@ using Microsoft.Windows.Storage.Pickers;
 
 var openPicker = new FileOpenPicker(this.AppWindow.Id);
 
-var file = await openPicker.PickSingleFileAsync();
-if (file is not null)
+var result = await openPicker.PickSingleFileAsync();
+if (result is not null)
 {
-    var content = System.IO.File.ReadAllText(file.Path);
+    var content = System.IO.File.ReadAllText(result.Path);
 }
 else
 {
@@ -100,10 +100,10 @@ C++
 using namespace winrt::Microsoft::Windows::Storage::Pickers;
 
 FileOpenPicker openPicker(AppWindow().Id());
-auto& file{ co_await openPicker.PickSingleFileAsync() };
-if (file)
+auto& result{ co_await openPicker.PickSingleFileAsync() };
+if (result)
 {
-    std::ifstream fileReader(file.Path().c_str());
+    std::ifstream fileReader(result.Path().c_str());
     std::string text((std::istreambuf_iterator<char>(fileReader)), std::istreambuf_iterator<char>());
     winrt::hstring hText = winrt::to_hstring(text);
 }
@@ -117,7 +117,7 @@ else
 
 Displays a UI element that allows user to choose and open multiple files.
 
-Returns a collection of picked files.
+Returns a collection of light weight objects that has the path of picked files.
 
 Return an empty list (Count = 0) if the file dialog's cancelled or closed without selection.
 
@@ -130,10 +130,10 @@ using Microsoft.Windows.Storage.Pickers;
 
 var openPicker = new FileOpenPicker(this.AppWindow.Id);
 
-var files = await openPicker.PickMultipleFilesAsync();
-if (files is not null)
+var results = await openPicker.PickMultipleFilesAsync();
+if (results.Count > 0)
 {
-    var pickedFilePaths = files.Select(f => f.Path);
+    var pickedFilePaths = results.Select(f => f.Path);
     foreach (var path in pickedFilePaths)
     {
         var content = System.IO.File.ReadAllText(path);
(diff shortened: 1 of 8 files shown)


specs/StoragePickers/FileOpenPicker.md
```diff
@@ -0,0 +1,172 @@
...
+# API Pages
+
+## Constructor
+
+### Attributes
+
+| **Attribute**            | **Type**                                                | **Description**                                                          |
+|--------------------------|---------------------------------------------------------|--------------------------------------------------------------------------|
+| `ViewMode`               | [Microsoft::Windows::Storage::Pickers::PickerViewMode](./PickerViewMode.md)  | Gets or sets the view mode that the file picker is using to present items. |
+| `SuggestedStartLocation` | [Microsoft::Windows::Storage::Pickers::PickerLocationId](./PickerLocationId.md)| Gets or sets the initial location where the file picker looks for files. |
```
whiskhub, line 21: Is there a chance `SuggestedStartLocationDirectory` will be added, where you can provide any `string` file path (e.g. if `SuggestedStartLocation == PickerLocationId.Unspecified`)? (see https://github.com/microsoft/WindowsAppSDK/pull/4942#discussion_r1890837161)

It would be a bummer if this limitation, that only applies to the UWP picker and UWP apps and not the COM picker, would be retained. From history, if it's not being addressed now, it will forever stay like that :/

benstevens48:
Is the general principle in WindowsAppSDK to continue using the StorageFile and StorageFolder types? I mean if writing in C#, using the .Net IO functions is more convenient, in which case it's preferable just to have file paths and not have the overhead of the StorageFile/Folder instances. However, I have to admit, I'm not full up to date with all the possible security models that can be used with WindowsAppSDK, so maybe there's a reason for using StorageFile and StorageFolder instead of just returning the file path?

mominshaikhdevs:
https://github.com/microsoft/WindowsAppSDK/issues/8#issuecomment-2256000803

#

 a new picker API that _must have these two **bare minimum**_ features:

1. works both in [AppContainer](https://github.com/microsoft/windowsappsdk/issues/219)/[win32-app-isolation](https://github.com/microsoft/win32-app-isolation) and outside of AppContainer. 

2. as fast as raw win32 apis.

specs/StoragePickers/FolderPicker.md
```diff
@@ -0,0 +1,100 @@
...
+```cpp
+winrt::Windows::Foundation::IAsyncOperation<winrt::Windows::Storage::StorageFolder> PickSingleFolderAsync();
+```
+Return null if the file dialog was cancelled or closed without selection.
+
+### Examples
+
+C#
+
+```csharp
```
DrusTheAxe, line 70: I think `c#` is preferable to `csharp` as it's supported by more environments (GitHub, ADO, VSCode, ...)

@codendone do you recall that conversation we had. Did it end with `c#` is the best choice or am I misremembering?

specs/StoragePickers/FolderPicker.md
```diff
@@ -0,0 +1,100 @@
...
+// (Optional) specify the text displayed on commit button. If not specified, use system default:
+folderPicker.CommitButtonText(L"Select Folder");
+```
+
+## FolderPicker.PickSingleFolderAsync
+
+Displays a UI element that allows the user to choose a folder.
+
+### Definition
+```cpp
```
DrusTheAxe, line 61: **WinRT APIs** should be defined as MIDL3.

**Examples** are written in specific languages (usually C#). Non-WinRT APIs are defined in their specific language, in the rare event a WinAppSDK API isn't WinRT.

RECOMMEND: Define as MIDL3, not any specific language

```c# (but really MIDL3)
namespace Windows.Storage.Pickers
{
    Windows.Foundation.IAsyncOperation<Windows.Storage.StorageFolder> PickSingleFolderAsync();
}
```

specs/StoragePickers/FileOpenPicker.md
```diff
@@ -0,0 +1,172 @@
+FileOpenPicker Class
```
DrusTheAxe, line 1: The spec is split across multiple files, with their definitions (IDL) likewise split across multiple files. This complicates reviewing (and holistically understanding) the design and details.

Do you intend to have multiple .idl files in the implementation?

CONSIDER: Moving all the definitions to a single file, ideally as an `# API` section in Microsoft.Windows.Storage.Pickers.md

```
...
# API

```c# (but really MIDL3)
namespace Windows.Storage.Pickers
{
    ...
}
```

That will also address gaps e.g. where's the API contract details?

specs/StoragePickers/PickerLocationId.md
```diff
@@ -0,0 +1,39 @@
...
+
+Specifies the default folder or library that a picker starts from or displays to the user.
+
+Namespace: [Microsoft.Windows.Storage.Pickers](./Microsoft.Windows.Storage.Pickers.md)
+
+# Fields
+
+| **Name**           | **Value** | **Description**                        |
+|--------------------|-----------|----------------------------------------|
+| DocumentsLibrary   | 0         | The Documents library.                 |
```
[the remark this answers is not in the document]
DinahK-2SO: Yes and no. For the enum itself, indeed it is 0 when not specified.
However, in the code of our new pickers, all 3 pickers set the default location as `Unspecified`.
We will implement:
```
struct FileOpenPicker : FileOpenPickerT<FileOpenPicker>
{
    private:
        // ... 
        PickerLocationId m_suggestedStartLocation{ PickerLocationId::Unspecified };
```
You can find detailed code on dev branch: https://github.com/microsoft/WindowsAppSDK/blob/user/xianghong/storage-pickers-develop/dev/Interop/StoragePickers/FileOpenPicker.h

specs/StoragePickers/PickerLocationId.md
```diff
@@ -0,0 +1,39 @@
+PickerLocationId Enum
+===
+
+Specifies the default folder or library that a picker starts from or displays to the user.
+
+Namespace: [Microsoft.Windows.Storage.Pickers](./Microsoft.Windows.Storage.Pickers.md)
+
+# Fields
+
+| **Name**           | **Value** | **Description**                        |
```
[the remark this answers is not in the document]
DinahK-2SO: > Where's the enum for `ApplicationData.Current.LocalFolder`?
> 
> Where's the enum for `ApplicationData.Current.LocalCacheFolder`?
> 
> Where's the enum for `ApplicationData.Current.TemporaryFolder`?

Hi @DrusTheAxe  They're part of the `Microsoft.Windows.Storage.ApplicationData` namespace, not in scope of this spec `Microsoft.Windows.Storage.Pickers`

DinahK-2SO:
> Is the general principle in WindowsAppSDK to continue using the StorageFile and StorageFolder types? I mean if writing in C#, using the .Net IO functions is more convenient, in which case it's preferable just to have file paths and not have the overhead of the StorageFile/Folder instances. However, I have to admit, I'm not full up to date with all the possible security models that can be used with WindowsAppSDK, so maybe there's a reason for using StorageFile and StorageFolder instead of just returning the file path?

Hello @benstevens48 , The new API returns `StorageFile` and `StorageFolder` primarily to maintain consistency with the APIs in `Windows.Storage.Pickers` that it is intended to replace. This approach allows users currently utilizing `Windows.Storage.Pickers` to migrate to Microsoft.Windows.Storage.Pickers with minimal changes.

At the current stage, once we have the StorageFile, we can use StorageFile.Path for subsequent operations (as demonstrated in the examples). In future, we may consider adding more methods and return types.

benstevens48:
> > Is the general principle in WindowsAppSDK to continue using the StorageFile and StorageFolder types? I mean if writing in C#, using the .Net IO functions is more convenient, in which case it's preferable just to have file paths and not have the overhead of the StorageFile/Folder instances. However, I have to admit, I'm not full up to date with all the possible security models that can be used with WindowsAppSDK, so maybe there's a reason for using StorageFile and StorageFolder instead of just returning the file path?
> 
> Hello @benstevens48 , The new API returns `StorageFile` and `StorageFolder` primarily to maintain consistency with the APIs in `Windows.Storage.Pickers` that it is intended to replace. This approach allows users currently utilizing `Windows.Storage.Pickers` to migrate to Microsoft.Windows.Storage.Pickers with minimal changes.
> 
> At the current stage, once we have the StorageFile, we can use StorageFile.Path for subsequent operations (as demonstrated in the examples). In future, we may consider adding more methods and return types.

@DinahK-2SO Thanks for the reply. That makes sense. I'm just concerned about the ongoing usage of StorageFile in WIndowsAppSDK since it always had very poor performance in UWP (I mean just creating 1 StorageFile could consume 100kB of RAM and loading many StorageFiles was hundreds of times slower than using Win32/.Net APIs). Although I haven't tested in in WIndowsAppSDK , so maybe it behaves differently. Probably in a file picker it doesn't matter too much, although in some cases the user may pick 1000s of files (like all photos in a large folder of photos etc), so it could start to become an issue.

specs/StoragePickers/PickerLocationId.md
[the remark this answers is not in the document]
whiskhub: A current "UWP (AppContainer) app" may not have the `Videos Library` capability, but can still open a file picker in the `Videos Library` using this enum. Setting a `SuggestedStartFolder` directly using `KnownFolder.VideosLibrary` would throw `access denied` however.
 
So, probably some kind of enum must stay for AppContainer apps.

specs/StoragePickers/FileOpenPicker.md
DinahK-2SO: Hi WinAppSDK friends @whiskhub , @MartyIX , @benstevens48 . 

The string input of `SuggestedStartLocation` feature is on our wishlist for future milestones. For the first version of `Microsoft.Windows.Storage.Pickers`, we aim to implement method signatures consistent with `Windows.Storage.Pickers` and will not add new signatures at this time.

specs/StoragePickers/PickerLocationId.md
```diff
@@ -0,0 +1,39 @@
...
+| DocumentsLibrary   | 0         | The Documents library.                 |
+| ComputerFolder     | 1         | The Computer folder.                   |
+| Desktop            | 2         | The Windows desktop.                   |
+| Downloads          | 3         | The Downloads folder.                  |
+| HomeGroup          | 4         | The HomeGroup.                         |
+| MusicLibrary       | 5         | The Music library.                     |
+| PicturesLibrary    | 6         | The Pictures library.                  |
+| VideosLibrary      | 7         | The Videos library.                    |
+| Objects3D          | 8         | The Objects library.                   |
+| Unspecified        | 9         | An unspecified location.               |
```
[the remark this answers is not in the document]
DinahK-2SO: Thanks @DrusTheAxe , I will add more details in the spec.
An unspecified location means that the file picker does not preset any specific storage location, allowing the user to choose the storage location themselves.

PS: The values here are consistent with [Windows.Storage.Pickers.PickerLocationId](https://learn.microsoft.com/en-us/uwp/api/windows.storage.pickers.pickerlocationid)

specs/StoragePickers/FolderPicker.md
```diff
@@ -0,0 +1,100 @@
...
+{
+    var path = folder.Path;
+}
+else
+{
+    // error handling.
+}
+```
+
+C++
```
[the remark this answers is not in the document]
DinahK-2SO: Hi @DrusTheAxe 

The reason for including both C# and C++ examples in the documentation is to cater to a broader audience. While the examples may only differ in syntax, providing both versions can help developers who are more comfortable with one language over the other. 

This approach is not unique to our specs. We have several other documents that provide examples in both languages, such as the [CameraCaptureUI.md](https://github.com/microsoft/WindowsAppSDK/blob/main/specs/CameraCaptureUI/CameraCaptureUI.md)

specs/StoragePickers/PickerLocationId.md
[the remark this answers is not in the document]
mominshaikhdevs: 1. UWPs/(aka `CoreWindow`s). 
2. AppContainer. 
3. MSIX.

are totally three independent things.

(https://github.com/microsoft/WindowsAppSDK/issues/4844#issuecomment-2455310626 + https://github.com/microsoft/winget-cli/pull/4260#issuecomment-2037458576)

specs/StoragePickers/PickerLocationId.md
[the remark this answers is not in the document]
DrusTheAxe: If `Unspecified` is the default why is the value for `Unspecified` != 0

Defaults don't have to be `0`/`false`/`null` but it's convenient. As this is new APIs with new types seems the default could have a value of `0`. Is there a specific reason it doesn't or can't?

specs/StoragePickers/PickerLocationId.md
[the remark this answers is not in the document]
DrusTheAxe: >but can still open a file picker in the Videos Library using this enum

Oh? That's interesting. Why?

Because the picker is interactive so OK on the prompt implicitly has user consent? If the app wants to access a location which normally requires an AppContainer to declare a capability but the user interactively navigates there that's OK as it implies user consent?

DrusTheAxe:
> > Is the general principle in WindowsAppSDK to continue using the StorageFile and StorageFolder types? I mean if writing in C#, using the .Net IO functions is more convenient, in which case it's preferable just to have file paths and not have the overhead of the StorageFile/Folder instances. However, I have to admit, I'm not full up to date with all the possible security models that can be used with WindowsAppSDK, so maybe there's a reason for using StorageFile and StorageFolder instead of just returning the file path?
> 
> Hello @benstevens48 , The new API returns `StorageFile` and `StorageFolder` primarily to maintain consistency with the APIs in `Windows.Storage.Pickers` that it is intended to replace. This approach allows users currently utilizing `Windows.Storage.Pickers` to migrate to Microsoft.Windows.Storage.Pickers with minimal changes.

Can you provide a variant which returns the file as a string e.g. `string PickSingleFileNameAsync(...)`?

Properly implementing that would require more than skin deep differences but ahort term that could be implemented as

```
var storageFile = picker.PickSingleFileAsync();
return storageFile != null ? storageFile.Path : null;
```

with the deeper implementation optimized to return the string w/o any `StorageFile` in the callstack. Even if only syntactic candy it lays the foundation for the improved dev experience while deferring the deeper optimized work for later when time permits.

DrusTheAxe:
@DinahK-2SO Performance can be a concern if you have the path/file as a string. The cost to create a StorageFile/Folder object can be non-trivial if your sole purpose is to rip the string out of the returned object and free the object.

Is the sole purpose of this spec is to provide WinAppSDK alternatives supporting elevation to Windows.* APIs that don't?

Seems it's maybe that and _less_ given what's spec'd is a subset of Windows.* pickers (no PickMultipleFilesAsync, CreateForUser, etc). It's hard assess the spec without an understanding of the desired endgame. Is that longer-term clear to you but intentionally deferred for now, 'cause, reasons?

specs/StoragePickers/FileOpenPicker.md
whiskhub: I am just afraid that "wishlist for future milestones" practically means "never" 😅 which means I'd "never" be able to use this picker :/

specs/StoragePickers/FileSavePicker.md
```diff
@@ -0,0 +1,125 @@
...
+| `ViewMode`                 | [Microsoft::Windows::Storage::Pickers::PickerViewMode](./PickerViewMode.md)    | Gets or sets the view mode that the file picker is using to present items.         |
+| `SuggestedStartLocation`   | [Microsoft::Windows::Storage::Pickers::PickerLocationId](./PickerLocationId.md)| Gets or sets the initial location where the file picker looks for files.           |
+| `SuggestedFileName`        | `winrt::hstring`                       | Gets or sets the file name displayed in the file name input box on launching the dialog.                                   |
+| `DefaultFileExtension`     | `winrt::hstring`                       | Gets or sets the file extension tailing the suggested file name in the file name input box on launching the dialog.        |
+| `CommitButtonText`         | `winrt::hstring`                       | Gets or sets the text displayed on the commit button of the file picker.                                                   |
+| `FileTypeChoices`          | `winrt::Windows::Foundation::Collections::IMap<hstring, winrt::Windows::Foundation::Collections::IVector<hstring>>`    | The file extensions categorized by purpose.|
```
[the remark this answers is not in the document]
DinahK-2SO: Hi @DrusTheAxe 

This is the document for FileSavePicker.
The existing property is [FileSavePicker.FileTypeChoices](https://learn.microsoft.com/en-us/uwp/api/windows.storage.pickers.filesavepicker.filetypechoices)

specs/StoragePickers/FileOpenPicker.md
DinahK-2SO: Hello @DrusTheAxe , 
1. we do have `PickerMultipleFilesAsync`, please find it here: [FileOpenPicker.PickMultipleFilesAsync](https://github.com/microsoft/WindowsAppSDK/blob/user/DinahK-2SO/storage_pickers_public_spec/specs/StoragePickers/FileOpenPicker.md#fileopenpickerpickmultiplefilesasync)

2. Regarding the omission of `CreateForUser`, thank you for bringing this up. This method is primarily used for the [the Xbox multi-user model](https://learn.microsoft.com/en-us/previous-versions/windows/uwp/xbox-apps/multi-user-applications). Since the new pickers are mainly designed for Desktop Apps and do not currently include Xbox support, we decided to omit it in this milestone.

specs/StoragePickers/PickerLocationId.md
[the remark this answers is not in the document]
DinahK-2SO: Thank you @DrusTheAxe  for your insightful suggestions! At this stage, we do not support specifying start locations outside of the PickerLocationId enum. However, your insights are invaluable and represent excellent design considerations! We will certainly discuss them in future milestones.

specs/StoragePickers/PickerLocationId.md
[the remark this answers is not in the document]
DinahK-2SO: Hi @DrusTheAxe, Thank you for your valuable feedback.

The value for Unspecified is set to 9, and all values in this enum are aligned with [Windows.Storage.Pickers.PickerLocationId](https://learn.microsoft.com/en-us/uwp/api/windows.storage.pickers.pickerlocationid) to ensure backward compatibility.

I added more explanations in the latest code push to clarify the default behavior, please fine it here: https://github.com/microsoft/WindowsAppSDK/blob/user/DinahK-2SO/storage_pickers_public_spec/specs/StoragePickers/PickerLocationId.md#pickerlocationidunspecified

specs/StoragePickers/PickerLocationId.md
```diff
@@ -0,0 +1,39 @@
...
+|--------------------|-----------|----------------------------------------|
+| DocumentsLibrary   | 0         | The Documents library.                 |
+| ComputerFolder     | 1         | The Computer folder.                   |
+| Desktop            | 2         | The Windows desktop.                   |
+| Downloads          | 3         | The Downloads folder.                  |
+| HomeGroup          | 4         | The HomeGroup.                         |
+| MusicLibrary       | 5         | The Music library.                     |
+| PicturesLibrary    | 6         | The Pictures library.                  |
+| VideosLibrary      | 7         | The Videos library.                    |
+| Objects3D          | 8         | The Objects library.                   |
```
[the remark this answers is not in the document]
DinahK-2SO: @DrusTheAxe  Thanks for your insights. I added more explanations in the latest code push to introduce these folders.

specs/StoragePickers/PickerLocationId.md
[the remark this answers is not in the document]
DinahK-2SO: Hi @DrusTheAxe , 
The current milestone does not include an enum for the application's source location (pkgdir for packaged apps, exedir for unpackaged apps). 

However, we can consider adding this in future updates. Details will be addressed in other replies.

specs/StoragePickers/PickerViewMode.md
```diff
@@ -0,0 +1,23 @@
...
+# Fields
+
+| **Name**   | **Value** | **Description**                |
+|------------|-----------|--------------------------------|
+| List       | 0         | Displays items in a list view. |
+| Thumbnail  | 1         | Displays items as thumbnails.  |
+
+# Definition
+
+```cpp
```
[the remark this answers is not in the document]
DinahK-2SO: @DrusTheAxe Updated in latest code push.

specs/StoragePickers/FolderPicker.md
```diff
@@ -0,0 +1,100 @@
...
+```cpp
+#include <winrt/Microsoft.Windows.Storage.Pickers.h>
+using namespace winrt::Microsoft::Windows::Storage::Pickers;
+
+FolderPicker folderPicker(AppWindow().Id());
+auto& folder = co_await folderPicker.PickSingleFolderAsync();
```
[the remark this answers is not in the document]
DinahK-2SO: @DrusTheAxe Updated. Thanks.

specs/StoragePickers/FolderPicker.md
```diff
@@ -0,0 +1,100 @@
...
+if (folder != nullptr)
```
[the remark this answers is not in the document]
DinahK-2SO: Hi @DrusTheAxe I understand you were suggestion the modern C++ style with `if (folder)`. I've updated the code sample. Please review the update

specs/StoragePickers/FolderPicker.md
```diff
@@ -0,0 +1,100 @@
...
+{
+    auto path = folder.Path();
```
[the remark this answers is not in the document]
DinahK-2SO: Updated. Thanks @DrusTheAxe

specs/StoragePickers/FolderPicker.md
DinahK-2SO: Updated. @DrusTheAxe Please find them here: https://github.com/microsoft/WindowsAppSDK/blob/user/DinahK-2SO/storage_pickers_public_spec/specs/StoragePickers/Microsoft.Windows.Storage.Pickers.md#definition

I pull all the API definitions into the main page Microsoft.Windows.Storage.Pickers.md, as recommended in another comment.

specs/StoragePickers/FileOpenPicker.md
DinahK-2SO: Updated. Please find their definitions here: https://github.com/microsoft/WindowsAppSDK/blob/user/DinahK-2SO/storage_pickers_public_spec/specs/StoragePickers/Microsoft.Windows.Storage.Pickers.md#definition

codendone: If you keep the files separate, I recommend adding an `# API Details` section at the bottom of each file and with a link to "Microsoft.Windows.Storage.Pickers.md#definition".

specs/StoragePickers/Microsoft.Windows.Storage.Pickers.md
```diff
@@ -0,0 +1,116 @@
...
+        HomeGroup = 4,
+        MusicLibrary = 5,
+        PicturesLibrary = 6,
+        VideosLibrary = 7,
+        Objects3D = 8,
+        Unspecified = 9,
+    }
+
+    runtimeclass PickFileResult {
+        PickFileResult(string path);
```
Hong-Xiang, line 63: We should not expose constructor of PickFileResult, and developer should not construct PickFileResult from path

specs/StoragePickers/Microsoft.Windows.Storage.Pickers.md
```diff
@@ -0,0 +1,116 @@
...
+
+        string Path { get; }
+
+        IAsyncOperation<Windows.Storage.StorageFile> ToStorageFileAsync();
+    }
+
+    runtimeclass PickFolderResult {
+        PickFolderResult(string path);
```
Hong-Xiang, line 71: Same as above, we should remove these constructors

specs/StoragePickers/Microsoft.Windows.Storage.Pickers.md
```diff
@@ -0,0 +1,116 @@
...
+        Windows.Foundation.IAsyncOperation<Windows.Foundation.Collections.IVectorView<PickFileResult>> PickMultipleFilesAsync();
+    }
+
+    runtimeclass FileSavePicker
+    {
+        FileSavePicker(Microsoft.UI.WindowId windowId);
+
+        string CommitButtonText;
+        string DefaultFileExtension;
+        IDictionary<string, IList<string>> FileTypeChoices;
```
Hong-Xiang, line 97: Maybe we should keep consistency on collection types

specs/StoragePickers/Microsoft.Windows.Storage.Pickers.md
```diff
@@ -0,0 +1,116 @@
...
+
+        IAsyncOperation<Windows.Storage.StorageFile> ToStorageFolderAsync();
+    }
+
+    runtimeclass FileOpenPicker
+    {
+        FileOpenPicker(Microsoft.UI.WindowId windowId);
+
+        string CommitButtonText;
+        IList<string> FileTypeFilter;
```
Hong-Xiang, line 83: Use correct collection type

specs/StoragePickers/Microsoft.Windows.Storage.Pickers.md
```diff
@@ -0,0 +1,116 @@
...
+
+        Windows.Foundation.IAsyncOperation<PickFileResult> PickSaveFileAsync()
+    }
+
+    runtimeclass FolderPicker
+    {
+        FolderPicker(Microsoft.UI.WindowId windowId);
+
+        string CommitButtonText;
+        IList<string> FileTypeFilter;
```
Hong-Xiang, line 109: Use correct collection type

specs/StoragePickers/PickFileResult.md
```diff
@@ -0,0 +1,88 @@
...
+|---------------|----------|-----------------|
+| `Path`        | `string` | Gets the path of the picked file. |
+
+### Examples
+C#
+
+```csharp
+using Microsoft.Windows.Storage.Pickers;
+
+var result = new PickFileResult("C:\\example\\file.txt");
```
Hong-Xiang, line 28: PickFileResult should not and can not be constructed from path, the only way to get PickFileResult is from pickers

specs/StoragePickers/PickFolderResult.md
```diff
@@ -0,0 +1,89 @@
...
+|---------------|----------|-----------------|
+| `Path`        | `string` | Gets the path of the picked folder. |
+
+### Examples
+C#
+
+```C#
+using Microsoft.Windows.Storage.Pickers;
+
+var result = new PickFolderResult("C:\\example\\folder");
```
Hong-Xiang, line 28: Remove this example

ghost1372:
Excuse me for sending messages among the elders.
But can you clarify and consider the following questions/requests?
1.Why you are using WindowId instead of hwnd? later you get hwnd like this:
https://github.com/microsoft/WindowsAppSDK/blob/da6a44e7f305b52d3a3092bf4e5f0e6a072d7fd1/dev/Interop/StoragePickers/FileOpenPicker.cpp#L57 

So why don't you get hwnd directly?
2.Why are there different properties for extensions?
FileOpenPicker-> **FileTypeFilter**
FileSavePicker-> **FileTypeChoices**

3.When we use **Windows.Storage.Picker**, we must set FileTypeFilter or FileTypeChoices, otherwise picker will not run. and get exception:

> System.Runtime.InteropServices.COMException: 'Unspecified error


I hope the new picker doesn't have this problem.

4.we can set details for extension in FileSavePicker like this:
`picker.FileTypeChoices.Add("Images", new string[] {".jpg", ".png" });`
but we can not add details and multiple extensions for FileOpenPicker:
`picker.FileTypeFilter.Add(".png");`

specs/StoragePickers/Microsoft.Windows.Storage.Pickers.md
```diff
@@ -0,0 +1,116 @@
...
+
+        Windows.Foundation.IAsyncOperation<PickFileResult> PickSaveFileAsync()
+    }
+
+    runtimeclass FolderPicker
+    {
+        FolderPicker(Microsoft.UI.WindowId windowId);
+
+        string CommitButtonText;
+        IVector<string> FileTypeFilter{ get; };
```
ghost1372, line 107: Does FolderPicker need FileTypeFilter? I think it's a copy-paste problem.

specs/StoragePickers/FolderPicker.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Definition
+
+```C#
+runtimeclass FolderPicker
+{
+    FolderPicker(Microsoft.UI.WindowId windowId);
+
+    string CommitButtonText;
+    IVector<string> FileTypeFilter{ get; };
```
ghost1372, line 22: same as https://github.com/microsoft/WindowsAppSDK/pull/5155/files#r1996312694

DinahK-2SO:
Hi @ghost1372 , thanks for raising these questions and sharing your insights! This is a long answer list following a long question list, so I'll address each part separately.

> 1.Why you are using WindowId instead of hwnd? later you get hwnd like this:
> 
> https://github.com/microsoft/WindowsAppSDK/blob/da6a44e7f305b52d3a3092bf4e5f0e6a072d7fd1/dev/Interop/StoragePickers/FileOpenPicker.cpp#L57
> 
> So why don't you get hwnd directly?

The reason for using `WindowId` instead of `hwnd` is to simplify the code and improve the developer experience. Using the existing `Windows.Storage.Pickers` with window handle requires additional steps, as the example shown below:

```csharp
using Windows.Storage.Pickers;

var hwnd = WinRT.Interop.WindowNative.GetWindowHandle(this);
var picker = new FileOpenPicker();
WinRT.Interop.InitializeWithWindow.Initialize(picker, hwnd);
var file = await picker.PickerSingleFileAsync();
string path = file.Path;
```

In the new `Microsoft.Windows.Storage.Pickers`, this process will streamlined by using `WindowId`:

```csharp
using Microsoft.Windows.Storage.Pickers;

var picker = new FileOpenPicker(this.AppWindow.Id);
var result = await picker.PickerSingleFileAsync();
string path = result.Path;
```

We can see the new approach reduces more boilerplate code and makes the API easier to use.

---

> 2.Why are there different properties for extensions?
> 
> FileOpenPicker-> **FileTypeFilter** FileSavePicker-> **FileTypeChoices**
> 
> Why not use **FileTypeChoices** for both?
> 

Also for the question 4: It's an inspiring request to add `FileTypeChoices` for the `FileOpenPicker`. While it might not be implemented in this release, if it can be helpful, do please suggest it in our [feedback wanted section](https://github.com/microsoft/WindowsAppSDK/discussions/categories/feedback-wanted). 

---
> 3.When we use **Windows.Storage.Picker**, we must set FileTypeFilter or FileTypeChoices, otherwise picker will not run. and get exception:
> 
> > System.Runtime.InteropServices.COMException: 'Unspecified error
> 
> I hope the new picker doesn't have this problem.

First, the answer for `FileOpenPicker.FileTypeFilter`:

Yes, we are making it optional. And with `FileTypeFilter` not specified, the `FileOpenPicker` would defalt to `*` - meaning all files.

This is also demonstrated in our [example code](https://github.com/microsoft/WindowsAppSDK/blob/user/DinahK-2SO/storage_pickers_public_spec/specs/StoragePickers/FileOpenPicker.md#examples) .

For the `FileTypeChoices` property of `FileSavePicker`, I can't make it optional for now because it can be essential in some cases: 

When the file picked by the end user doesn't exist, `FileSavePicker` creates the file. And if the user hasn't specified the new file's extension at this point, FileSavePicker will default to using the first extension in the current choice. In this context, `FileTypeChoices` is essential.

---
> 4.We can set details for extension in FileSavePicker like this:
> 
> `picker.FileTypeChoices.Add("Images", new string[] {".jpg", ".png" });`
> 
> but we can not add details and multiple extensions for FileOpenPicker:
> 
> `picker.FileTypeFilter.Add(".png");`

Please refer to the previous answer 2.

ghost1372:
Thank you @DinahK-2SO 
i created a request for it.
https://github.com/microsoft/WindowsAppSDK/discussions/5255

specs/StoragePickers/PickerLocationId.md
```diff
@@ -0,0 +1,110 @@
...
+It is often mapped to `%USERPROFILE%\Desktop`.
+
+## PickerLocationId.Downloads
+The `Downloads` value represents the user's Downloads folder.
+
+This is the default location where web browsers and other applications save downloaded files. 
+It is often mapped to `%USERPROFILE%\Downloads`.
+
+## PickerLocationId.HomeGroup
+! Note that [HomeGroup has been removed from Windows 10 (version 1803) and later](https://support.microsoft.com/en-us/windows/homegroup-removed-from-windows-10-version-1803-07ca5db1-7bca-4d11-68a3-a31ff4a09979),
```
codendone, line 68: What does it map to? If this isn't meaningful, then it is probably better to just remove this enum value.

codendone:
Iteration 5 looks good to me.

yeelam-gordon (approved):
Please clarify do we need to indicate it is for "experimental only" or not need.
Also, for Microsoft doc, do we need to do it ourselves or not? I assume the doc reviewer is need.

mominshaikhdevs:
@DinahK-2SO clarify in the docs if it works within [AppContainers](https://github.com/microsoft/issues/219) | [win32-app-isolation](https://github.com/microsoft/win32-app-isolation) or not?

HEIC-to-JPEG-Dev:
So, are we saying that using the pickers will work in WinUI 3 LowIL mode at some release of the SDK ?

charlesroddie:
@DinahK-2SO can you clarify the above two comments? The incompatibility of pickers with LowIL/appcontainer environments has been the biggest problem of WinUI for years. Is the current status that's it's a way to distribute apps to users who are happy with less security and (as of this PR) to system administrators? Or is the fact that it doesn't work with LowIL/appcontainer something to be resolved before the feature is released?

specs/StoragePickers/Microsoft.Windows.Storage.Pickers.md
```diff
@@ -0,0 +1,117 @@
...
+===
+
+The `Microsoft.Windows.Storage.Pickers` API in the Windows App SDK lets desktop applications (like 
+WinUI3) present a streamlined UI for selecting files or folders, modifying filenames and extensions, 
+and accessing paths — all while integrating seamlessly across Windows desktops.
+
+# Background
+
+The standard OS file and folder picker APIs [Windows.Storage.Pickers](https://learn.microsoft.com/en-us/uwp/api/windows.storage.pickers) 
+do not work when running as an administrator. To address this gap, here we're introducing the new
```
charlesroddie, line 11: Why does the spec mention "running as administrator" but not inside appcontainer? @DinahK-2SO @DrusTheAxe @codendone ?

HEIC-to-JPEG-Dev: Because it will **only** run in HighIL (Admin mode) and MediumIL (WinUI default) - it will not run in LowIL (think UWP security model). They have said it may be added in future, but there is no roadmap for that.

charlesroddie: Surely that can't be true @DinahK-2SO @DrusTheAxe @codendone that a whole new spec for pickers is done but the central problem of pickers in winui is ignored?

PR #5166, opened 2025-02-19
Update IsCallingPreviewSupported API

anupriya13:
### **What Changed?**

IsCallingPreviewSupported API from returning false to returning WindowsVersion::IsWindows11_22H2OrGreater(). All OS changes are completed.

**Note**: This the work around we are doing for now. Actual fix for this would be, to have an API on Framework Udk which would tell if feature were enabled or not for more accuracy which we are currently working on. TO DO in future PRs.

### **Why is this change required?**

We need to create sample test apps and test the feature end to end.

### **What happens when App calls the Interfaces and OS changes are not present?**

If the feature is not available in OS then default behaviour is handled from OS where new UI elements "Settings Button" and "Camera Preview" won't be visible, rest all UI elements will be visible and the notification pops successfully.

### DIFF
--- a/dev/AppNotifications/AppNotificationConferencingConfig.cpp
+++ b/dev/AppNotifications/AppNotificationConferencingConfig.cpp
@@ -43,10 +43,9 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
     }
 
     ///Checks if the calling preview feature is supported on the current OS version
-    ///TO DO - This method needs implementation on framework UDK, for now it always returns false
     bool AppNotificationConferencingConfig::IsCallingPreviewSupported()
     {
-        return false;
+        return WindowsVersion::IsWindows11_22H2OrGreater();
     }
 }
 
@@ -57,7 +56,7 @@ namespace winrt::Microsoft::Windows::AppNotifications
 {
     bool AppNotificationConferencingConfig::IsCallingPreviewSupported()
     {
-        return false;
+        return WindowsVersion::IsWindows11_22H2OrGreater();
     }
 }
 #endif
--- a/dev/AppNotifications/AppNotificationConferencingConfig.cpp
+++ b/dev/AppNotifications/AppNotificationConferencingConfig.cpp
@@ -43,6 +43,7 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
     }
 
     ///Checks if the calling preview feature is supported on the current OS version
+    ///TO DO - This method needs implementation on framework UDK for more accuracy, for now it always returns IsWindows11_22H2OrGreater
     bool AppNotificationConferencingConfig::IsCallingPreviewSupported()
     {
         return WindowsVersion::IsWindows11_22H2OrGreater();


satkh:
Add in description,
1. What is this change about.
2. What happens when App calls the Interfaces and OS changes are not present.
3. any other info

satkh (approved):

dev/AppNotifications/AppNotificationConferencingConfig.cpp
```diff
@@ -43,10 +43,9 @@ namespace winrt::Microsoft::Windows::AppNotifications::implementation
     }
 
     ///Checks if the calling preview feature is supported on the current OS version
-    ///TO DO - This method needs implementation on framework UDK, for now it always returns false
     bool AppNotificationConferencingConfig::IsCallingPreviewSupported()
     {
-        return false;
+        return WindowsVersion::IsWindows11_22H2OrGreater();
```
godlytalias, line 48: This check will return true even for 22621.0 os version right? Whether the feature is supported in that OS Version?

satkh: No this will not be available, in all the versions.
This the work around we are doing, for now actual fix for this would be, to have an API on Framework Udk which would tell if feature were enabled or not.

dev/AppNotifications/AppNotificationConferencingConfig.cpp
```diff
@@ -57,7 +56,7 @@ namespace winrt::Microsoft::Windows::AppNotifications
 {
     bool AppNotificationConferencingConfig::IsCallingPreviewSupported()
     {
-        return false;
+        return WindowsVersion::IsWindows11_22H2OrGreater();
```
codendone, line 59: > WindowsVersion::IsWindows11_22H2OrGreater() [](http://example.com/codeflow?start=14&length=44)

This one should not need to be changed. If the feature is disabled, that means this is a stable release and this experimental API should never be enabled.

dev/AppNotifications/AppNotificationConferencingConfig.cpp
codendone: So this is a temporary and not-actually-correct change? If so, there should at least be a big comment on that and still a TODO to fix this in a future change.

Is there a reason the correct change can't be done now?

dev/AppNotifications/AppNotificationConferencingConfig.cpp
satkh: @codendone , 
So, this will remain disabled for entire winappsdk 1.7 timeline.
To enable this, we need to add this change on 1.8 experimental 1?

dev/AppNotifications/AppNotificationConferencingConfig.cpp
anupriya13: Updated the TODO Comment, please review @codendone

dev/AppNotifications/AppNotificationConferencingConfig.cpp
codendone: No, this change (line 60) should not be necessary to enable the function for 1.8 experimental builds. The TerminalVelocity for that case should have the #define on line 55 set to 1, which means this code won't be used. This #ifdef'ed function should only be used when the feature is completely disabled for a stable build.

Issue #3871, opened 2023-09-20
AppInstance.GetActivatedEventArgs() doesn't get IFileActivatedEventArgs if file has Unicode chars


peerpalo:
### Describe the bug

This was found out in microsoft/PowerToys/issues/28147. 

If application is registered in shell\open\command with a registry event args like this `"----ms-protocol:ms-encodedlaunch:App?ContractId=Windows.File&Verb=open&File=%1"` and filename is, for instance, `新建 文本文档.reg`, `GetActivatedEventArgs()` will return a `ExtendedActivationKind.Protocol` instead of `ExtendedActivationKind.File` and no file could be retrieved.

### Steps to reproduce the bug

1. Register an application in HKEY_CLASSES_ROOT with shell\open\command `"----ms-protocol:ms-encodedlaunch:App?ContractId=Windows.File&Verb=open&File=%1"`
2. Try to open a file with name `新建 文本文档.reg`

### Screenshots

![image](https://github.com/microsoft/WindowsAppSDK/assets/6475341/f9081d42-d89c-4102-8e76-49e08feea8b8)
![image](https://github.com/microsoft/WindowsAppSDK/assets/6475341/b2954f45-da98-4de2-bed7-46723e13639b)

### NuGet package version

Windows App SDK 1.4.0: 1.4.230822000

### Packaging type

Unpackaged

### Windows version

Windows 11 version 22H2 (22621, 2022 Update)

### IDE

Visual Studio 2022

asklar:
there's 2 bugs here:
cc @aeloros fyi

1. The deserialization does not handle punycode, and compares against the contractId directly, here: https://github.com/microsoft/WindowsAppSDK/blob/a4935d2b7a8d21fcc0f0b0c010a6b083766a2cef/dev/AppLifecycle/ExtensionContract.h#L47

2. The file argument is not unescaped, here: https://github.com/microsoft/WindowsAppSDK/blob/a4935d2b7a8d21fcc0f0b0c010a6b083766a2cef/dev/AppLifecycle/FileActivatedEventArgs.h#L52

MuyuanMS:
I have created https://github.com/microsoft/WindowsAppSDK/pull/5175 on this issue.
`uri.ParseQuery()` would have empty result when uri contains unicode characters, which is what caused the problem.
During investigation, besides files with unicode characters, I also found out that current Uri parsing logic also would cause issues (app could crash) when file contains special characters like `&` (Just for awareness).
[...]

PR #5175, opened 2025-02-21
[Issue #3871] File activation for unpackaged App would fail when file path contains unicode characters or special characters

MuyuanMS:
## Background
For unpackaged App, when opening a file with type registered with [RegisterForFileTypeActivation()](https://learn.microsoft.com/windows/windows-app-sdk/api/winrt/microsoft.windows.applifecycle.activationregistrationmanager.registerforfiletypeactivation?view=windows-app-sdk-1.6), current design is to execute the app with encoded protocol as argument (`----ms-protocol:ms-encodedlaunch:App?ContractId=Windows.File&Verb=open&File=%1`)

E.g.
`"C:\MyApp.exe" "----ms-protocol:ms-encodedlaunch:App?ContractId=Windows.File&Verb=open&File=C:\myfile.foo"`

## Expectation
The expectation is when the app calls `GetActivatedEventArgs()`, the returned value should has a type of `ExtendedActivationKind.File` with file path/name in the data.

## Problem
However, when the file path contains unicode characters or special characters like `&`, the file activation would fail as the encoded protocol is not correctly parsed. E.g.
### With unicode characters:
`"C:\MyApp.exe" "----ms-protocol:ms-encodedlaunch:App?ContractId=Windows.File&Verb=open&File=C:\你好.foo"`
This would result in activation type being parsed as `ExtendedActivationKind.Protocol`

### With special character:
`"C:\MyApp.exe" "----ms-protocol:ms-encodedlaunch:App?ContractId=Windows.File&Verb=open&File=C:\a&b.foo"`
This would crash the app on calling `GetActivatedEventArgs()`

## Root cause
The query in `winrt::Windows::Foundation::Uri` is parsed with `QueryParsed()`, which would split on `&` and cause problem when having unicode characters.

## Solution
The best way to address the issue, is to change the activation protocol design, such that file path, which may contain unicode characters or special characters, is not placed in the protocol uri not designed to handle file paths. 
E.g. instead of `----ms-protocol:ms-encodedlaunch:App?ContractId=Windows.File&Verb=open&File=%1`, we put the file path as separate parameter like:
`"----ms-protocol:ms-encodedlaunch:App?ContractId=Windows.File&Verb=open" "%1"`

However, this would require design changes in multiple components including registering, serializing and deserializing protocols. For now, this PR is a simple fix just on parsing logic.

Instead of using Uri.QueryParsed(), we use simple string operations to parse the uri query. (E.g. substring from "&File=" to get the file path). 

## Test
During the fix I locally created a sample app launched with below methods to test the change:

1. activation from file `abc.txt`
2. activation from file `你好.txt`
3. activation from file `a&b.txt`
4. activation from file `😀.txt`

Previously, 2 would give protocol type, and 3 would crash the app.
After fix, all three cases give file type, with correct file path contained in the returned value.

I have also added a functional test activating unpackaged app via a file with unicode characters and special character `&`

# Auto Generated PR Description

WARNING:

### DIFF
--- a/dev/AppLifecycle/ExtensionContract.h
+++ b/dev/AppLifecycle/ExtensionContract.h
@@ -37,17 +37,6 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
 
     inline std::tuple<ExtendedActivationKind, winrt::Windows::Foundation::IInspectable> DecodeActivatedEventArgs(winrt::Windows::Foundation::Uri const& uri)
     {
-        // If the Uri strictly follows uri format from GenerateEncodedLaunchUri, first query parameter is ContractId
-        // This is to avoid calling uri.QueryParsed() as it can cause issues when uri contains unicode characters or special characters like &.
-        for (const auto& extension : c_extensionMap)
-        {
-            std::wstring contractIdQuery = L"?" + std::wstring(c_contractIdKeyName) + L"=" + std::wstring(extension.contractId);
-            if (CompareStringOrdinal(uri.Query().c_str(), (int)wcslen(contractIdQuery.c_str()), contractIdQuery.c_str(), -1, TRUE) == CSTR_EQUAL)
-            {
-                return { extension.kind, extension.factory(uri) };
-            }
-        }
-
         for (auto const& pair : uri.QueryParsed())
         {
             if (CompareStringOrdinal(pair.Name().c_str(), -1, c_contractIdKeyName, -1, TRUE) == CSTR_EQUAL)
@@ -61,6 +50,14 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
                     }
                 }
             }
+        }
+
+        // QueryParsed() function would return empty when it's a file contract with unicode characters in file path
+        // Thus following additional check for file contract is needed
+        auto fileContractUri = GenerateEncodedLaunchUri(L"App", c_fileContractId);
+        if (CompareStringOrdinal(uri.AbsoluteUri().c_str(), fileContractUri.length(), fileContractUri.c_str(), -1, TRUE) == CSTR_EQUAL)
+        {
+            return { ExtendedActivationKind::File, FileActivatedEventArgs::Deserialize(uri) };
         }
 
         return { ExtendedActivationKind::Protocol, nullptr };
--- a/dev/AppLifecycle/ExtensionContract.h
+++ b/dev/AppLifecycle/ExtensionContract.h
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation and Contributors.
+// Copyright (c) Microsoft Corporation and Contributors.
 // Licensed under the MIT License.
 #pragma once
 
@@ -37,6 +37,17 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
 
     inline std::tuple<ExtendedActivationKind, winrt::Windows::Foundation::IInspectable> DecodeActivatedEventArgs(winrt::Windows::Foundation::Uri const& uri)
     {
+        // If the Uri strictly follows uri format from GenerateEncodedLaunchUri, first query parameter is ContractId
+        // This is to avoid calling uri.QueryParsed() as it can cause issues when uri contains unicode characters or special characters like &.
+        for (const auto& extension : c_extensionMap)
+        {
+            std::wstring contractIdQuery = L"?" + std::wstring(c_contractIdKeyName) + L"=" + std::wstring(extension.contractId);
+            if (CompareStringOrdinal(uri.Query().c_str(), (int)wcslen(contractIdQuery.c_str()), contractIdQuery.c_str(), -1, TRUE) == CSTR_EQUAL)
+            {
+                return { extension.kind, extension.factory(uri) };
+            }
+        }
+
         for (auto const& pair : uri.QueryParsed())
         {
             if (CompareStringOrdinal(pair.Name().c_str(), -1, c_contractIdKeyName, -1, TRUE) == CSTR_EQUAL)
--- a/dev/AppLifecycle/FileActivatedEventArgs.h
+++ b/dev/AppLifecycle/FileActivatedEventArgs.h
@@ -64,7 +64,7 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
             auto verbBegin = query.find(L"&Verb=");
             if (verbBegin == std::wstring::npos)
             {
-                throw winrt::hresult_invalid_argument();
+                throw winrt::hresult_invalid_argument(L"Query of encoded file protocol should contain 'Verb'");
             }
             verbBegin += 6; // Length of "&Verb="
 
@@ -77,7 +77,7 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
             auto fileBegin = query.find(L"&File=");
             if (fileBegin == std::wstring::npos)
             {
-                throw winrt::hresult_invalid_argument();
+                throw winrt::hresult_invalid_argument(L"Query of encoded file protocol should contain 'File'");
             }
             fileBegin += 6; // Length of "&File="
 
--- a/dev/AppLifecycle/FileActivatedEventArgs.h
+++ b/dev/AppLifecycle/FileActivatedEventArgs.h
@@ -48,15 +48,15 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
         static winrt::Windows::Foundation::IInspectable Deserialize(winrt::Windows::Foundation::Uri const& uri)
         {
             auto parsedQuery = uri.QueryParsed();
+
+            // By design parsed query should have a size of 3 (ContractId, Verb, File), in which case result from QueryParsed() can be used directly.
+            // However, it may have a size of 0 when uri contains unicode characters, or more than 3 when file path contains "&", which requires manual parsing with string functions.
             if (parsedQuery.Size() == 3)
             {
                 auto verb = parsedQuery.GetFirstValueByName(L"Verb");
                 auto file = parsedQuery.GetFirstValueByName(L"File");
                 return make<FileActivatedEventArgs>(verb, file);
             }
-
-            // parsed query may have a size of 0 when uri contains unicode characters, or more than 3 when file contains "&"
-            // In such cases, manually parse the query with string functions
 
             const std::wstring query = uri.Query().c_str();
             auto queryLength = query.length();
--- a/dev/AppLifecycle/FileActivatedEventArgs.h
+++ b/dev/AppLifecycle/FileActivatedEventArgs.h
@@ -47,27 +47,46 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
 
         static winrt::Windows::Foundation::IInspectable Deserialize(winrt::Windows::Foundation::Uri const& uri)
         {
-            // Standard serialize has a query of the form: "?ContractId=Windows.File&Verb=<verb>&File=<file_path>"
-            // If the query has the exact same format, then we can deserialize it with simple string operations,
-            // as QueryParsed() can cause issues when file path contains unicode characters or specical characters like &.
-            std::wstring serializedQueryHeader = L"?" + std::wstring(c_contractIdKeyName) + L"=" + std::wstring(c_fileContractId) + L"&Verb=";
-            int headerLen = (int)wcslen(serializedQueryHeader.c_str());
-            if (CompareStringOrdinal(uri.Query().c_str(), headerLen, serializedQueryHeader.c_str(), -1, TRUE) == CSTR_EQUAL)
+            auto parsedQuery = uri.QueryParsed();
+            if (parsedQuery.Size() == 3)
             {
-                int queryLen = (int)wcslen(uri.Query().c_str());
-                const std::wstring query = uri.Query().c_str();
-                int verbEnd = query.find(L"&", headerLen);
-                auto verb = winrt::to_hstring(query.substr(headerLen, verbEnd - headerLen).c_str());
-
-                int filePos = query.find(L"&File=");
-                auto file = winrt::to_hstring(query.substr(filePos + 6, queryLen - filePos - 6).c_str());
-
+                auto verb = parsedQuery.GetFirstValueByName(L"Verb");
+                auto file = parsedQuery.GetFirstValueByName(L"File");
                 return make<FileActivatedEventArgs>(verb, file);
             }
 
-            auto query = uri.QueryParsed();
-            auto verb = query.GetFirstValueByName(L"Verb");
-            auto file = query.GetFirstValueByName(L"File");
+            // parsed query may have a size of 0 when uri contains unicode characters, or more than 3 when file contains "&"
+            // In such cases, manually parse the query with string functions
+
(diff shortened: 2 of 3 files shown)


dev/AppLifecycle/FileActivatedEventArgs.h
```diff
@@ -47,6 +47,24 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
...
         {
+            // Standard serialize has a query of the form: "?ContractId=Windows.File&Verb=<verb>&File=<file_path>"
+            // If the query has the exact same format, then we can deserialize it with simple string operations,
+            // as QueryParsed() can cause issues when file path contains unicode characters or specical characters like &.
+            std::wstring serializedQueryHeader = L"?" + std::wstring(c_contractIdKeyName) + L"=" + std::wstring(c_fileContractId) + L"&Verb=";
+            int headerLen = (int)wcslen(serializedQueryHeader.c_str());
+            if (CompareStringOrdinal(uri.Query().c_str(), headerLen, serializedQueryHeader.c_str(), -1, TRUE) == CSTR_EQUAL)
+            {
+                int queryLen = (int)wcslen(uri.Query().c_str());
+                const std::wstring query = uri.Query().c_str();
```
Scottj1s, line 58: consider hoisting this up to reuse on lines 55, 57

dev/AppLifecycle/FileActivatedEventArgs.h
```diff
@@ -47,6 +47,24 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
...
+                int verbEnd = query.find(L"&", headerLen);
```
Scottj1s, line 59: need error checking for [std::string::npos](https://en.cppreference.com/w/cpp/string/basic_string/npos)

dev/AppLifecycle/FileActivatedEventArgs.h
```diff
@@ -47,6 +47,24 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
...
+                auto verb = winrt::to_hstring(query.substr(headerLen, verbEnd - headerLen).c_str());
+
+                int filePos = query.find(L"&File=");
```
Scottj1s, line 62: need error checking for [std::string::npos](https://en.cppreference.com/w/cpp/string/basic_string/npos)

dev/AppLifecycle/FileActivatedEventArgs.h
Scottj1s: also, isn't this going to find the first '&' in the query string, prior to Verb?  it might be better to search remaining substrings as you go - strip the query header off to create a std::wstring verb = "Verb=<verb>&File=<file_path>", and strip that off to create a std::wstring file = "File=<file_path>"

dev/AppLifecycle/ExtensionContract.h
```diff
@@ -37,6 +37,17 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
 
     inline std::tuple<ExtendedActivationKind, winrt::Windows::Foundation::IInspectable> DecodeActivatedEventArgs(winrt::Windows::Foundation::Uri const& uri)
     {
+        // If the Uri strictly follows uri format from GenerateEncodedLaunchUri, first query parameter is ContractId
```
dhoehna, line 40: What if this is not true?  If the Uri does not strictly follow the uri format from GenerateEncodedLaunchUri?

dev/AppLifecycle/ExtensionContract.h
```diff
@@ -37,6 +37,17 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
...
+        // This is to avoid calling uri.QueryParsed() as it can cause issues when uri contains unicode characters or special characters like &.
+        for (const auto& extension : c_extensionMap)
+        {
+            std::wstring contractIdQuery = L"?" + std::wstring(c_contractIdKeyName) + L"=" + std::wstring(extension.contractId);
+            if (CompareStringOrdinal(uri.Query().c_str(), (int)wcslen(contractIdQuery.c_str()), contractIdQuery.c_str(), -1, TRUE) == CSTR_EQUAL)
```
dhoehna, line 45: `CompareStringOrdinal` looks for binary equality.  This looks like a linguistic comparison and should use the appropriate method.  Please use [comparestringex](https://learn.microsoft.com/windows/win32/api/stringapiset/nf-stringapiset-comparestringex)

dhoehna, line 45: For the 2nd parameter, why the length of `contractIdQuery` and not `uri`?

dev/AppLifecycle/ExtensionContract.h
```diff
@@ -37,6 +37,17 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
 
     inline std::tuple<ExtendedActivationKind, winrt::Windows::Foundation::IInspectable> DecodeActivatedEventArgs(winrt::Windows::Foundation::Uri const& uri)
     {
+        // If the Uri strictly follows uri format from GenerateEncodedLaunchUri, first query parameter is ContractId
+        // This is to avoid calling uri.QueryParsed() as it can cause issues when uri contains unicode characters or special characters like &.
+        for (const auto& extension : c_extensionMap)
+        {
+            std::wstring contractIdQuery = L"?" + std::wstring(c_contractIdKeyName) + L"=" + std::wstring(extension.contractId);
```
dhoehna, line 44: I'm having a hard time understanding what this method does.  The difference between this loop, and the loop below is comparing the string with both the contract name and contract id.  Further, it uses the same methods to compare the strings.

dev/AppLifecycle/ExtensionContract.h
```diff
@@ -37,6 +37,17 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
...
+            if (CompareStringOrdinal(uri.Query().c_str(), (int)wcslen(contractIdQuery.c_str()), contractIdQuery.c_str(), -1, TRUE) == CSTR_EQUAL)
+            {
+                return { extension.kind, extension.factory(uri) };
+            }
+        }
+
```
dhoehna, line 50: What is true if code reaches here?

dev/AppLifecycle/FileActivatedEventArgs.h
```diff
@@ -47,6 +47,24 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
 
         static winrt::Windows::Foundation::IInspectable Deserialize(winrt::Windows::Foundation::Uri const& uri)
         {
+            // Standard serialize has a query of the form: "?ContractId=Windows.File&Verb=<verb>&File=<file_path>"
+            // If the query has the exact same format, then we can deserialize it with simple string operations,
```
dhoehna, line 51: What will happen if the query is not the same format?

dhoehna, line 51: NIT: Don't use `we` in code comments.

dev/AppLifecycle/FileActivatedEventArgs.h
dhoehna: +1 considering the duplicate string code is in a different file.

dev/AppLifecycle/FileActivatedEventArgs.h
```diff
@@ -47,6 +47,24 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
...
+            // as QueryParsed() can cause issues when file path contains unicode characters or specical characters like &.
+            std::wstring serializedQueryHeader = L"?" + std::wstring(c_contractIdKeyName) + L"=" + std::wstring(c_fileContractId) + L"&Verb=";
+            int headerLen = (int)wcslen(serializedQueryHeader.c_str());
```
dhoehna, line 54: auto.

dev/AppLifecycle/FileActivatedEventArgs.h
dhoehna, line 62: is `.find()` case-sensitive?

dhoehna:
Please add tests.

dev/AppLifecycle/ExtensionContract.h
MuyuanMS: Uri parsing and checking for exact match sounds more like ordinal comparison, like existing code for parsing contract id - Or are there cases where protocols to be put in registry could be localized?

dev/AppLifecycle/ExtensionContract.h
MuyuanMS: Sorry for the poor readability in logic - I have refactored it a little bit.
Point is when dealing with encoded protocols for files, when file contains special characters, `QueryParsed()` function would not parse correctly (it would return empty if unicode characters are involved) and the current loop may not run at all.

dev/AppLifecycle/FileActivatedEventArgs.h
MuyuanMS: This encoded launch is for WinAppSDK apps only - which means all registration should be from the code here. There's no public documentation suggesting this encoded protocol can be customized.
Without the assumption, there's no way to guarantee correct parsing, as file path could contain special characters like `&`
Never the less, I have refactored a little bit to loosen the assumption to be the query should contain 3 and only 3 pairs, for ContractId, Verb, and File.

MuyuanMS: Got it, thanks!

dev/AppLifecycle/FileActivatedEventArgs.h
MuyuanMS: The second argument in `find()` indicates the index to start the search, so it finds the first `&` after `Verb=`

dev/AppLifecycle/FileActivatedEventArgs.h
MuyuanMS: Yes, and the existing logic also treats protocol query as case sensitive

dev/AppLifecycle/ExtensionContract.h
dhoehna: I see.  I would suggest to add the reasoning why the assumption is safe to make.

dev/AppLifecycle/ExtensionContract.h
dhoehna: I see URI and I think of a user making the URI, or a part of it.  Language comparison.  Further, unless I am incorrect, UNICODE characters can be represented with a different number of bytes.  

If the uri is generated by code, I believe a binary comparison would suffice.

dev/AppLifecycle/ExtensionContract.h
dhoehna: Coolio.

dev/AppLifecycle/FileActivatedEventArgs.h
dhoehna: I apologize.  I should be more clear in what I'm asking for.  Can you add an assert, or some other code, that checks to make sure the queries have the exact same format?

I'm only thinking ahead by 1-5 years when this code breaks because the assumption is not true anymore.

dev/AppLifecycle/ExtensionContract.h
```diff
@@ -52,6 +52,14 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
             }
         }
 
+        // QueryParsed() function would return empty when it's a file contract with unicode characters in file path
+        // Thus following additional check for file contract is needed
+        auto fileContractUri = GenerateEncodedLaunchUri(L"App", c_fileContractId);
```
dhoehna, line 57: NIT: I can't comment on the code above here.  I'll leave the comment here instead.  Line 40-53.  Prefer to check for the incorrect states and `continue` early instead of having nested scopes.

dhoehna:
The code looks good to me.  Please add automated tests.

dev/AppLifecycle/FileActivatedEventArgs.h
```diff
@@ -47,9 +47,46 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
 
         static winrt::Windows::Foundation::IInspectable Deserialize(winrt::Windows::Foundation::Uri const& uri)
         {
-            auto query = uri.QueryParsed();
-            auto verb = query.GetFirstValueByName(L"Verb");
-            auto file = query.GetFirstValueByName(L"File");
+            auto parsedQuery = uri.QueryParsed();
+            if (parsedQuery.Size() == 3)
```
dhoehna, line 51: Why 3?  Is this a safe upper limit, or is 3 the magic number of queries with UNICODE characters and special characters?

dhoehna: I just now read the comment.  I suggest to move it up to explain the number 3.

dev/AppLifecycle/ExtensionContract.h
MuyuanMS: The first `if` statement on Line 42 does the job, if I'm understanding this comment correctly. It will only enter nested loops if the query name is `ContractId`

dev/AppLifecycle/FileActivatedEventArgs.h
MuyuanMS: Done

dev/AppLifecycle/FileActivatedEventArgs.h
```diff
@@ -47,9 +47,46 @@ namespace winrt::Microsoft::Windows::AppLifecycle::implementation
...
+                return make<FileActivatedEventArgs>(verb, file);
+            }
+
+            const std::wstring query = uri.Query().c_str();
+            auto queryLength = query.length();
+
+            auto verbBegin = query.find(L"&Verb=");
+            if (verbBegin == std::wstring::npos)
+            {
+                throw winrt::hresult_invalid_argument();
```
dhoehna, line 67: for this, and all other `winrt::hresult_invalid_argument` please add a message indicating the incorrect argument and why it is incorrect.

test/AppLifecycle/FunctionalTests.cpp
```diff
@@ -128,6 +131,42 @@ namespace Test::AppLifecycle
...
+
+            // Cleanup any leftover data from previous runs i.e. ensure we running with a clean slate
+            try
+            {
+                Execute(L"AppLifecycleTestApp.exe", L"/UnregisterFile", g_deploymentDir);
+                WaitForEvent(event, m_failed);
+            }
+            catch (...)
+            {
+                //TODO:Unregister should not fail if ERROR_FILE_NOT_FOUND | ERROR_PATH_NOT_FOUND
```
dhoehna, line 147: Instead of a comment, add the check here.  That way, if Unregister does fail if ERROR_FILE_NOT_FOUND or ERROR_PATH_NOT_FOUND developers will be notified.

dhoehna:
Thanks for adding the test. :)

dhoehna (approved):

test/AppLifecycle/FunctionalTests.cpp
MuyuanMS: This is copied from other testcases unregistering protocol or files - If this would have failed due to file not found or path not found, the following execute statements would fail as well anyway and developer can indeed know it.

dev/AppLifecycle/FileActivatedEventArgs.h
MuyuanMS: Added

dhoehna (approved):

PR #5182, opened 2025-02-25
EnsurePackage*ReadyAsync() doesn't register v2 if v1 is registered and caller asks for minVersion=v1

DrusTheAxe:
EnsurePackage*ReadyAsync() doesn't register v2 if v1 is registered and caller asks for minVersion=v1

AddPackage(foo-v1)
StagePackage(foo-v2)
==> v1 is registered, v2 is staged
pdm.RegisterNewerIfAvailable(true)
pdm.EnsurePackageReadyAsync(foo-v1)
ACTUAL=> v1 is registered, v2 is staged
EXPECTED> v2 is registered

https://task.ms/55967280 

Also fixed *PackageOptions.ExpectedDigests() didn't JIT-create a map on access


dev/PackageManager/API/M.W.M.D.PackageDeploymentManager.cpp
```diff
@@ -3575,6 +3598,61 @@ namespace winrt::Microsoft::Windows::Management::Deployment::implementation
...
 
+    winrt::Microsoft::Windows::Management::Deployment::RegisterPackageOptions PackageDeploymentManager::ToRegisterOptions(winrt::Microsoft::Windows::Management::Deployment::EnsureReadyOptions const& options) const
+    {
+        winrt::Microsoft::Windows::Management::Deployment::RegisterPackageOptions toOptions;
+        const auto addOptions{ options.AddPackageOptions() };
+
+        // RegisterPackageOptions has no TargetVolume -- no need (the packge is already staged however it's staged on the machine)
+        for (const auto uri : addOptions.DependencyPackageUris())
+        {
+            toOptions.DependencyPackageUris().Append(uri);
```
letao-msft, line 3609: This question doesn't affect your changes and I do need more sleep, but can you remind me in which scenarios Register needs dependencies to be specified?  Same thing with OptionalPackageFamilyNames.  I feel like DependencyResolver would be able to figure them out implicitly.

DrusTheAxe: One example
* A.msix containing `<PackageDepedency Name="XYZ"...>` in appxmanifest.xml
* Stage(A.msix)
* Register(A) => EROR_DEPENDENCY_NOT_RESOLVED 'cause can't satisfy XYZ dependency

Another
* C:\A containing `<PackageDepedency Name="XYZ"...>` in appxmanifest.xml
* Register(C:\A\appxmanifest.xml) => EROR_DEPENDENCY_NOT_RESOLVED 'cause can't satisfy XYZ dependency

Dependency resolution uses what's on the machine plus provided as parameters to the API call (AddPackageAsync etc). Since `Add*Async()` us effectively Stage+Register dependencies must be resolved to succeed. But if you don't use Add*() the additional info in the *Options object can be necessary.

Personally it was easy - I xref'd what's in `EnsureReadyOptions` (mostly `AddPackageOptions`) with `RegisterPackageOptions` and mapped everything over that could be. We may want to add more options in the future e.g. if you specify EnsureReadyOptions.AddPackageOptions.Foo and really do need it to be honored but .Foo doesn't exist in RegisterPackageOptions, then perhaps when we do Register we should fail because the "required" option isn't honored.

This largely isn't true today - you can specify lots of values in e.g. `AddPackageOptions` but not all permutations of `AddPackageAsync(string,options)` use every value in options. So I didn't feel a strong need to invent any such required/enforcement/error'ing today. If/when it becomes a significant need we can always add more knobs into the options objects relatively quickly.

letao-msft (approved):
:shipit:

PR #5208, opened 2025-03-06
56138352: Use the global default version of LKG compiler

alexlamtest:
Use the version of LKG compiler specified by the Agg repo as the global default (file build\AzurePipelinesTemplates\WindowsAppSDK-Global-LKGVersions.yml), unless a different version is specified as a repo-local override (in local file WindowsAppSDK-CommonVariables.yml).

Here's a summary of how the branch being used on this WindowsAppSDK repo determines the branch being used on the Agg repo for accessing the file WindowsAppSDK-Global-LKGVersions.yml:

- WindowsAppSDK : Main --> Agg: Main, this is likely the majority case.
- WindowsAppSDK : release-1.x --> Agg: release-1.x, when x is >= 8, i.e., for servicing branches new enough to support this mechanism, the servicing branch continue to use the LKG compiler version specified by the WindowsAppSDK-Global-LKGVersions.yml file on the Agg repo to build participating feeder repos.
- WindowsAppSDK : release-1.x --> Agg: Main, when x is <8. For those old servicing branches on Agg that don't contain have the WindowsAppSDK-CommonVariables.yml file, redirect to Agg: Main to avoid "file not found" error during yml parsing. At runtime, the local WindowsAppSDK-CommonVariables.yml file on the WindowsAppSDK repo specifies the LKG compiler version to use, ignoring the global default set in the Agg: Main branch.
- WindowsAppSDK : private branch --> Agg: Main, by default. That's because for most private branches on the WindowsAppSDK repo, the developer would not have a corresponding private branch on the Agg repo. But in case the developer does want to use a corresponding private branch on the Agg repo, in the private branch on the WindowsAppSDK repo, changing
build/AzurePipelinesTemplates/WindowsAppSDK-Global-LKGVersions.yml@WindowsAppSDKAggregatorMainBranch to
build/AzurePipelinesTemplates/WindowsAppSDK-Global-LKGVersions.yml@WindowsAppSDKAggregatorSourceBranch should get the desired behavior.

//////////////////////////////////////

WARNING:

### DIFF
--- a/build/AzurePipelinesTemplates/WindowsAppSDK-Build-Stage.yml
+++ b/build/AzurePipelinesTemplates/WindowsAppSDK-Build-Stage.yml
@@ -57,18 +57,52 @@ stages:
           buildConfiguration: 'release'
           normalizedConfiguration: 'fre'
     variables:
-      ob_outputDirectory: '$(REPOROOT)\out'
-      ob_sdl_codeSignValidation_excludes: '-|**\Release\**;-|**\packages\**'
-      ob_artifactBaseName: "FoundationBinaries_$(buildConfiguration)_$(buildPlatform)"
-      ${{ if parameters.runStaticAnalysis }}:
-        ob_sdl_apiscan_enabled: true
-      ${{ if not( parameters.runStaticAnalysis ) }}:
-        ob_sdl_apiscan_enabled: false
-      ob_sdl_apiscan_softwareFolder: '$(build.SourcesDirectory)\APIScanTarget'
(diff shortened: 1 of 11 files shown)


alexlamtest:
Putting this PR on hold because we are exploring the option of creating servicing branches on the Config repo, which will enable us to move the yml file which specifies the global default LKG compiler version from the Agg repo to the Config repo. If we indeed opt to go that way, this PR will require updates.

lauren-ciha:
We have closed this PR because the fix is no longer needed.

PR #5228, opened 2025-03-13
Enable MSIXTooling if DotNet CLI

kythant:

WARNING:

### DIFF
--- a/build/NuSpecs/Microsoft.WindowsAppSDK.Foundation.props
+++ b/build/NuSpecs/Microsoft.WindowsAppSDK.Foundation.props
@@ -6,6 +6,12 @@
         If true, this is the WindowsAppSdk deployment of Foundation, versus the internal deployment for test only.
         This property is defined by deployment-specific props imports, and should not be overridden by clients.-->
     <WindowsAppSdkFoundation>true</WindowsAppSdkFoundation>
+
+    <!-- 
+        WindowsAppSDK apps cannot build without MSIX tooling enabled because MRTCore Tooling will failed to 
+        https://github.com/microsoft/WindowsAppSDK/issues/3939 
+    -->
+    <EnableMsixTooling Condition="'$(MSBuildRuntimeType)'=='Core'">true</EnableMsixTooling>
   </PropertyGroup>
 
   <!--


dev/MRTCore/packaging/MrtCore.targets
```diff
@@ -4,6 +4,8 @@
   <!-- We want MRT Core to NOT be used in WAP or UWP projects. Instead, we want the in-box PRI generation tooling to be used. -->
   <PropertyGroup>
     <EnableCoreMrtTooling Condition="'$(EnableCoreMrtTooling)'=='' and '$(WindowsAppContainer)'!='true' and '$(MSBuildProjectExtension)' != '.wapproj'">true</EnableCoreMrtTooling>
+    <!-- If MSBuildRuntimeType is Core then we are using dotnet build because it will fail since ExpandPriContent is not available in the dotnet enviroment -->
```
jonwis, line 7: Do we know why "expandpricontent" isn't available there?

dev/MRTCore/packaging/MrtCore.targets
```diff
@@ -4,6 +4,8 @@
...
+    <EnableCoreMrtTooling Condition="'$(MSBuildRuntimeType)'=='Core'">false</EnableCoreMrtTooling>
```
jonwis, line 8: Should this be combined into the prior line? Extend the list of `and` values.

Feels like "enable this toolset" makes more sense based on the content of the project.

Can someone build a C# WinUI3 app with resources in `dotnet build`? Should they be able to?

dev/MRTCore/packaging/MrtCore.targets
kythant: Technical answer:
In DotNet Build,
MSBuildExtensionsPath = C:\Program Files\dotnet\sdk\9.0.201\
and Microsoft.Build.Packaging.Pri.Tasks.dll cannot be found there.

General answer:
I am not sure why. Maybe, Microsoft.Build.Packaging.Pri.Tasks.dll is just not provided by dotnet.

dev/MRTCore/packaging/MrtCore.targets
kythant: Should this be combined into the prior line? Extend the list of and values.
We could but I was worried the line will become too long. 

Can someone build a C# WinUI3 app with resources in dotnet build? Should they be able to?
I don't know if this is a valid scenario. I presume we would only build with dotnet build if we are on a dotnet app?
@DrusTheAxe @jamespik @MikeHillberg What do you all think?

DrusTheAxe: Can - today? I thought there were issues with `dotnet build` and work needed to make it work. @Scottj1s likely knows more but not currently available.

>I presume we would only build with dotnet build if we are on a dotnet app?

dotnet build and msbuild equivalent-ish, but that -ish can be significant in syntax, functionality and environment factors. There are differences, both obvious and subtle. Search for "dotnet build vs msbuild" for some of the details.

It's desirable to fully support `dotnet build` but it always sounded like feature-level work (not a simple bugfix).

Scottj1s (changes requested):
Chatted with Will.  We need to RCA the failure here and seek a general approach that works with both NET Core (via dotnet, MSBuildRuntimeType == "Core") and NET FX (via msbuild, MSBuildRuntimeType == "Full").  A good starting point is to target both in a single Microsoft.Build.Msix.dll assembly via `<TargetFramework>netstandard2.0</TargetFramework>`, versus [multi-targeting for NET Core and NET FX](https://dev.azure.com/microsoft/ProjectReunion/_git/WindowsAppSDKClosed?path=/src/Microsoft-WinUI-Sdk/Microsoft.Build.Msix/Microsoft.Build.Msix.csproj&version=GBmain&line=4&lineEnd=5&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents).  Having done that, we can experiment with dotnet and msbuild executions to see whether we have complete support for both .NET implementations (and not, e.g., a partial NET Core dotnet environment that relies on VS NET FX assemblies).  Worst case, we can provide a complete lifted .NET Standard 2.0 environment.

dev/MRTCore/packaging/MrtCore.targets
Scottj1s: This definitely smells: 
`    <AppxMSBuildToolsPath Condition="'$(AppxMSBuildToolsPath)' == ''">$(MSBuildExtensionsPath)\Microsoft\VisualStudio\v$(VisualStudioVersion)\AppxPackage\</AppxMSBuildToolsPath>`

If we can't find a way to load the entire packaging (PRI, MRT, Appx) environment from dotnet, without relying on potentially unavailable VS dlls, we may need to consider lifting that entire environment and eliminating the hybrid mixture.

PR #5240, opened 2025-03-18
Experimental Microsoft.Windows.Storage.Pickers.* API

Hong-Xiang:
Draft implementation for `Microsoft.Windows.Storage.Pickers.*` APIs following spec #5155.


dev/Interop/StoragePickers/StoragePickersTelemetry.h
```diff
@@ -0,0 +1,29 @@
...
+    "Microsoft.WindowsAppSDK.StoragePickersTelemetry",
+    // {6ddc5826-bf0a-522e-cc84-0e70eda439ed}
+    (0x6ddc5826,0xbf0a,0x522e,0xcc,0x84,0x0e,0x70,0xed,0xa4,0x39,0xed));
+
+class StoragePickersTelemetry : public wil::TraceLoggingProvider
+{
+    IMPLEMENT_TELEMETRY_CLASS(StoragePickersTelemetry, StoragePickersTelemetryProvider);
+
+public:
+    BEGIN_COMPLIANT_MEASURES_ACTIVITY_CLASS(StoragePickersOperation, PDT_ProductAndServicePerformance);
```
AgneLukoseviciute, line 18: Consider adding separate events for PickSingleFile and PickMultipleFiles similar to how App Notification Telemetry has separate events for Unregister and UnregisterAll: https://github.com/microsoft/WindowsAppSDK/blob/638865280167cb73b28cf55c1247d92f0d39a753/dev/AppNotifications/AppNotificationTelemetry.h#L46

dev/Interop/StoragePickers/FileOpenPicker.cpp
```diff
@@ -0,0 +1,161 @@
...
+        PCWSTR appName = m_telemetryHelper.GetAppName().c_str();
+        auto logCaptureOperation{ StoragePickersTelemetry::StoragePickersOperation::Start(isAppPackaged, appName) };
+
+        PickerCommon::PickerParameters parameters{};
+
+        CaptureParameters(parameters);
+
+        co_await winrt::resume_background();
+
+        auto cancellationToken = co_await winrt::get_cancellation_token();
```
yeelam-gordon, line 81: In this repo, I see other code are using have these two lines of code:
            auto cancellation = co_await winrt::get_cancellation_token();
            cancellation.enable_propagation();
Should we do the same?
This line may be OK, since the dialog haven't popup, but I see we have another line near the same, which we already finished all the dialog operation, but still return nullptr, which is strange.

dev/Interop/StoragePickers/FileOpenPicker.cpp
```diff
@@ -0,0 +1,161 @@
...
+#include <windows.h>
+#include <shobjidl.h>
+#include <shobjidl_core.h>
+#include <winrt/Microsoft.UI.Interop.h>
+#include "PickerCommon.h"
+#include "PickFileResult.h"
+
+namespace winrt::Microsoft::Windows::Storage::Pickers::implementation
+{
+    FileOpenPicker::FileOpenPicker(winrt::Microsoft::UI::WindowId const& windowId)
```
yeelam-gordon, line 17: Is it possible people (or is it allowed) to pass null or empty windowId?
If not allow, where do we do the checking?

And how about if I can an non-relevant id? any where we should catch any result for throwing correct exception in case?

dev/Interop/StoragePickers/FileOpenPicker.cpp
```diff
@@ -0,0 +1,161 @@
...
+    {
+        bool isAppPackaged = m_telemetryHelper.IsPackagedApp();
+        PCWSTR appName = m_telemetryHelper.GetAppName().c_str();
+        auto logCaptureOperation{ StoragePickersTelemetry::StoragePickersOperation::Start(isAppPackaged, appName) };
+
+        PickerCommon::PickerParameters parameters{};
+
+        CaptureParameters(parameters);
+
+        co_await winrt::resume_background();
```
yeelam-gordon, line 79: I see this line usually after get_cancellation_token();
Not sure if there is a practice there.

dev/Interop/StoragePickers/FileOpenPicker.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+    winrt::Windows::Foundation::IAsyncOperation<winrt::Windows::Foundation::Collections::IVectorView<winrt::Microsoft::Windows::Storage::Pickers::PickFileResult> > FileOpenPicker::PickMultipleFilesAsync()
+    {
+        // capture parameters to avoid using get strong referece of picker
+        PickerCommon::PickerParameters parameters{};
+        CaptureParameters(parameters);
+
+        co_await winrt::resume_background();
+
+        winrt::Windows::Foundation::Collections::IVector<winrt::Microsoft::Windows::Storage::Pickers::PickFileResult> results{ winrt::single_threaded_vector<winrt::Microsoft::Windows::Storage::Pickers::PickFileResult>() };
```
yeelam-gordon, line 118: I assume we can create this "results" object after we know how many objects "are picked.

I would like to pass in the initial size of the vector, so that it doesn't need to keep "expanding" when we add the "result" one by one to results.

dev/Interop/StoragePickers/FileSavePicker.cpp
```diff
@@ -0,0 +1,167 @@
...
+#include <wil/resource.h>
+#include <winrt/Microsoft.UI.Interop.h>
+#include <winrt/Windows.Foundation.Collections.h>
+#include "PickerCommon.h"
+#include "PickFileResult.h"
+
+namespace winrt::Microsoft::Windows::Storage::Pickers::implementation
+{
+
+    FileSavePicker::FileSavePicker(winrt::Microsoft::UI::WindowId const& windowId)
```
yeelam-gordon, line 22: Sorry for my strange question.
Can we create a "FileOpenPicker" inside this FileSavePicker, and as the result many similar code can be gone?

dev/Interop/StoragePickers/FileSavePicker.cpp
```diff
@@ -0,0 +1,167 @@
...
+        auto cancellationToken = co_await winrt::get_cancellation_token();
+        if (cancellationToken())
+        {
+            co_return nullptr;
+        }
+
+        auto dialog = create_instance<IFileSaveDialog>(CLSID_FileSaveDialog, CONTEXT_ALL);
+        parameters.ConfigureDialog(dialog);
+
+        if (!PickerCommon::IsHStringNullOrEmpty(defaultFileExtension))
```
yeelam-gordon, line 110: Are you usually put these logic into parameters.ConfigureDialog?

dev/Interop/StoragePickers/FolderPicker.cpp
```diff
@@ -0,0 +1,103 @@
...
+#include "Microsoft.Windows.Storage.Pickers.FolderPicker.g.cpp"
+#include <shobjidl.h>
+#include <shobjidl_core.h>
+#include <winrt/Microsoft.UI.Interop.h>
+#include "PickerCommon.h"
+#include "PickFolderResult.h"
+
+namespace winrt::Microsoft::Windows::Storage::Pickers::implementation
+{
+    FolderPicker::FolderPicker(winrt::Microsoft::UI::WindowId const& windowId)
```
yeelam-gordon, line 15: Again, the same strange question.

Let's say there is an "internal picker", which you have most of the common implementation there, will it helps on many of similar code across 3 picker now?

yeelam-gordon: I see you have some "Common logic" in Common's cpp

dev/Interop/StoragePickers/PickerCommon.cpp
```diff
@@ -0,0 +1,213 @@
...
+#include "PickerCommon.h"
+#include <wil/resource.h>
+#include "ShObjIdl.h"
+#include <KnownFolders.h>
+
+namespace {
+
+    GUID HashHStringToGuid(winrt::hstring const& input)
+    {
+        auto algorithm = winrt::Windows::Security::Cryptography::Core::HashAlgorithmProvider::OpenAlgorithm(winrt::Windows::Security::Cryptography::Core::HashAlgorithmNames::Md5());
```
yeelam-gordon, line 14: Even for this like "hash case", MD5 is considered as not secure and static scanner will create security bug.
We need to use like HashAlgorithmNames::Sha256()

dev/Interop/StoragePickers/PickerCommon.cpp
```diff
@@ -0,0 +1,213 @@
...
+        guid.Data3 = (guid.Data3 & 0x0FFF) | 0x3000; // Set the version to 3
+        guid.Data4[0] = (guid.Data4[0] & 0x3F) | 0x80; // Set variant to RFC 4122
+
+        return guid;
+    }
+
+    winrt::com_ptr<IShellItem> GetKnownFolderFromId(winrt::Microsoft::Windows::Storage::Pickers::PickerLocationId pickerLocationId)
+    {
+        KNOWNFOLDERID knownFolderId;
+        switch (pickerLocationId)
```
yeelam-gordon, line 39: Can we create any dictionary or map can just directly look up?

dev/Interop/StoragePickers/FileOpenPicker.cpp
```diff
@@ -0,0 +1,161 @@
...
+        parameters.SettingsIdentifierId = m_settingsIdentifier;
+        parameters.PickerLocationId = m_suggestedStartLocation;
+        parameters.FileTypeFilterPara = PickerCommon::CaptureFilterSpec(parameters.FileTypeFilterData, m_fileTypeFilter.GetView());
+    }
+
+    winrt::Windows::Foundation::IAsyncOperation<winrt::Microsoft::Windows::Storage::Pickers::PickFileResult> FileOpenPicker::PickSingleFileAsync()
+    {
+        bool isAppPackaged = m_telemetryHelper.IsPackagedApp();
+        PCWSTR appName = m_telemetryHelper.GetAppName().c_str();
+        auto logCaptureOperation{ StoragePickersTelemetry::StoragePickersOperation::Start(isAppPackaged, appName) };
```
yeelam-gordon, line 73: I do see people called "logCaptureOperation.Stop();"
Which make sense.

And not sure if we can added one more parameter to indicate:
a. No problem (include cancel)
b. User exception, i.e. nothing we can do
c. Our Unknown/Code fault exception, i.e. something we need to improve.

codendone:
>       <AdditionalDependencies>onecore.lib;onecoreuap.lib;%(AdditionalDependencies)</AdditionalDependencies>

I'd expect shell32.lib to be needed for this case as well. (I'm guessing we just aren't generally building x86 debug.)

---
Refers to: dev/WindowsAppRuntime_DLL/WindowsAppRuntime_DLL.vcxproj:157 in c19343f. [](commit_id = c19343fc4856d534bc185f8a39b1a8deac102304, deletion_comment = False)

dev/WindowsAppRuntime_DLL/WindowsAppRuntime_DLL.vcxproj
```diff
@@ -171,7 +172,7 @@
     <Link>
       <SubSystem>Windows</SubSystem>
       <EnableUAC>false</EnableUAC>
-      <AdditionalDependencies>onecore.lib;onecoreuap.lib;%(AdditionalDependencies)</AdditionalDependencies>
+      <AdditionalDependencies>onecore.lib;onecoreuap.lib;shell32.lib;%(AdditionalDependencies)</AdditionalDependencies>
       <ModuleDefinitionFile>WindowsAppRuntime.def</ModuleDefinitionFile>
       <DelayLoadDLLs>Microsoft.Internal.FrameworkUdk.dll;%(DelayLoadDLLs)</DelayLoadDLLs>
```
codendone, line 177: >DelayLoadDLLs [](http://example.com/codeflow?start=7&length=13)

shell32.lib presumably brings in shell32.dll. If so, I think that dll should be added to the DelayLoadDlls (for all flavors).

dev/Projections/CS/Microsoft.Windows.Storage.Pickers.Projection/Microsoft.Windows.Storage.Pickers.Projection.csproj
```diff
@@ -0,0 +1,59 @@
...
+    <IsTrimmable>true</IsTrimmable>
+  </PropertyGroup>
+
+  <!-- Suppress CS8305: Feature is for evaluation purposes only and is subject to change or removal in future updates. -->
+  <PropertyGroup>
+    <NoWarn>8305</NoWarn>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.ProjectReunion.InteractiveExperiences.TransportPackage" />
```
codendone, line 21: nit: I think there is no need for the IXP package for the pickers?

dev/Interop/StoragePickers/PickerCommon.cpp
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    winrt::hstring GetPathFromShellItem(winrt::com_ptr<IShellItem> shellItem)
+    {
+        wil::unique_cotaskmem_string filePath;
+        check_hresult(shellItem->GetDisplayName(SIGDN_FILESYSPATH, filePath.put()));
+        return winrt::hstring{ filePath.get() };
+    }
+
+    std::vector<COMDLG_FILTERSPEC> CaptureFilterSpec(std::vector<winrt::hstring>& buffer, winrt::Windows::Foundation::Collections::IVectorView<winrt::hstring> filters)
```
codendone, line 145: >buffer [](http://example.com/codeflow?start=82&length=6)

Why pass in this buffer? It looks like `parameters.FileTypeFilterData` is what gets passed in, which I don't see anything using. Why not have the buffer just be a local, or skip it entirely to directly build result in a single loop?

dev/Interop/StoragePickers/PickerCommon.cpp
```diff
@@ -0,0 +1,213 @@
...
+            result.push_back({ L"All Files", L"*.*" });
+        }
+        return result;
+    }
+
+    std::vector<COMDLG_FILTERSPEC> CaptureFilterSpec(std::vector<winrt::hstring>& buffer, winrt::Windows::Foundation::Collections::IMapView<winrt::hstring, winrt::Windows::Foundation::Collections::IVector<hstring>> filters)
+    {
+        std::vector<COMDLG_FILTERSPEC> result(filters.Size());
+        buffer.clear();
+        buffer.reserve(filters.Size() * 2ull);
```
codendone, line 172: >2ull [](http://example.com/codeflow?start=40&length=4)

nit: curious that this is different than the other CaptureFilterSpec

dev/Interop/StoragePickers/PickerCommon.cpp
```diff
@@ -0,0 +1,213 @@
...
+        check_hresult(shellItem->GetDisplayName(SIGDN_FILESYSPATH, filePath.put()));
+        return winrt::hstring{ filePath.get() };
+    }
+
+    std::vector<COMDLG_FILTERSPEC> CaptureFilterSpec(std::vector<winrt::hstring>& buffer, winrt::Windows::Foundation::Collections::IVectorView<winrt::hstring> filters)
+    {
+        std::vector<COMDLG_FILTERSPEC> result(filters.Size());
+        buffer.clear();
+        buffer.reserve(filters.Size() * (size_t)2);
+        for (auto filter : filters)
```
codendone, line 150: >auto [](http://example.com/codeflow?start=13&length=4)

nit: `const auto&` like the other CaptureFilterSpec?

dev/Interop/StoragePickers/PickerCommon.cpp
```diff
@@ -0,0 +1,213 @@
...
+    void PickerParameters::ConfigureDialog(winrt::com_ptr<IFileDialog> dialog)
+    {
+        if (!IsHStringNullOrEmpty(CommitButtonText))
+        {
+            check_hresult(dialog->SetOkButtonLabel(CommitButtonText.c_str()));
+        }
+
+        if (!IsHStringNullOrEmpty(SettingsIdentifierId))
+        {
+            auto guid = HashHStringToGuid(SettingsIdentifierId);
```
codendone, line 201: >SettingsIdentifierId [](http://example.com/codeflow?start=42&length=20)

Is SettingsIdentifierId just the Id passed by the app, or is it combined in some way with an app identify (such as package name or exe name if unpackaged)? If it isn't combined with an app identifier, we might want to change that, so you don't get two apps both specifying something simple like "SaveDialog" and slamming each other's settings. Or do packaged apps have this info by default get stored in virtualized registry, such that this might only be an issue for unpackaged apps (and apps which disable registry virtualization)?

(This can be logged as an issue for later follow-up. No need to block initial checkin.)

dev/Interop/StoragePickers/Microsoft.Windows.Storage.Pickers.idl
```diff
@@ -0,0 +1,85 @@
+// Copyright (c) Microsoft Corporation and Contributors.
+// Licensed under the MIT License.
+
+#include <TerminalVelocityFeatures-StoragePickers.h>
+namespace Microsoft.Windows.Storage.Pickers
+{
+    [feature(Feature_StoragePickers)]
```
codendone, line 7: >[feature(Feature_StoragePickers)] [](http://example.com/codeflow?start=4&length=33)

(not blocking initial checkin:) Note: Before these APIs go stable and remove "[feature(...)]", these will need to have a "[contract(...)]" added.

dev/Interop/StoragePickers/PickerCommon.h
```diff
@@ -0,0 +1,33 @@
...
+namespace PickerCommon {
+    winrt::hstring GetPathFromShellItem(winrt::com_ptr<IShellItem> shellItem);
+
+    std::vector<COMDLG_FILTERSPEC> CaptureFilterSpec(std::vector<winrt::hstring>& buffer, winrt::Windows::Foundation::Collections::IVectorView<winrt::hstring> filters);
+    std::vector<COMDLG_FILTERSPEC> CaptureFilterSpec(std::vector<winrt::hstring>& buffer, winrt::Windows::Foundation::Collections::IMapView<winrt::hstring, winrt::Windows::Foundation::Collections::IVector<winrt::hstring>> filters);
+
+    bool IsHStringNullOrEmpty(winrt::hstring value);
+
+    struct PickerParameters {
+        HWND HWnd;
```
codendone, line 24: >; [](http://example.com/codeflow?start=17&length=1)

Add `{}` to ensure HWnd is initialized.

dev/Interop/StoragePickers/FolderPicker.cpp
```diff
@@ -0,0 +1,113 @@
...
+            auto hr = dialog->Show(parameters.HWnd);
+            if (FAILED(hr) || cancellationToken())
+            {
+                logTelemetry.Stop(m_telemetryHelper, false);
+                co_return nullptr;
+            }
+        }
+
+        winrt::com_ptr<IShellItem> shellItem{};
+        check_hresult(dialog->GetResult(shellItem.put()));
```
codendone, line 99: >check_hresult [](http://example.com/codeflow?start=8&length=13)

(not blocking:) If this throws, logTelemetry.Stop() will not be called -- is that okay?

codendone: Same question for the other pickers.

dev/Interop/StoragePickers/FileSavePicker.cpp
```diff
@@ -0,0 +1,176 @@
...
+        wil::unique_cotaskmem_string fileName;
+        check_hresult(shellItem->GetDisplayName(SIGDN_NORMALDISPLAY, fileName.put()));
+        std::wstring fileNameStr(fileName.get());
+
+        // Check if the file name has an extension
+        if (fileNameStr.find_last_of(L".") == std::wstring::npos)
+        {
+            // If the user defined file name doesn't have an extension,
+            //     automatically use the first extension from the first category in fileTypeChoices.
+            auto firstCategory = fileTypeChoices.First().Current();
```
codendone, line 157: >fileTypeChoices.First() [](http://example.com/codeflow?start=33&length=23)

Is it legal to not pass in any FileTypeChoices? I think we decided it is, in which case some handling is required here.

dev/Interop/StoragePickers/FileSavePicker.cpp
```diff
@@ -0,0 +1,176 @@
...
+            auto firstExtension = firstCategory.Value().GetAt(0);
+            pathStr = pathStr + firstExtension;
```
codendone, line 159: >firstExtension [](http://example.com/codeflow?start=32&length=14)

Is there validation anywhere, perhaps in the base IFileDialog code, that the extension starts with a '.'?

test/StoragePickersTests/StoragePickersTests.cpp
```diff
@@ -0,0 +1,211 @@
...
+    class StoragePickersTests
+    {
+    public:
+        BEGIN_TEST_CLASS(StoragePickersTests)
+            TEST_CLASS_PROPERTY(L"ThreadingModel", L"MTA") // MTA is required for ::Test::Bootstrap::SetupPackages()
+            TEST_CLASS_PROPERTY(L"RunFixtureAs:Class", L"RestrictedUser")
+            //TEST_CLASS_PROPERTY(L"RunFixtureAs:Class", L"UAP")
+            //TEST_CLASS_PROPERTY(L"RunAs", L"UAP")
+
+            TEST_CLASS_PROPERTY(L"IsolationLevel", L"Method") // each test sets its own CompatibilityOptions
```
codendone, line 29: >each test sets its own CompatibilityOptions [](http://example.com/codeflow?start=65&length=43)

Bad copy/paste. You could probably remove this line entirely, since I think method-level isolation should not be required by these tests.

.gitignore
```diff
@@ -395,3 +395,6 @@ build/override/
 # WindowsAppSDK specific files
 Microsoft.WinUI.AppX.targets
 !dev/vsix/**/*.pubxml
+
+# TAEF related
+redist/
```
codendone, line 400: Why is this needed? It doesn't look like you changed anything about TAEF install, and this doesn't seem to have been an issue for others in the past.

dev/Interop/StoragePickers/FileOpenPicker.cpp
Hong-Xiang: If we pass empty handle here (e.g. simply using `default` in csharp code for this parameter), the underlying API Common Item Dialog is actually accepting empty/nullptr HWND, and can successfully pick files/folders, however it will create an isolated windows in that case, not blocking the triggering UI, which might not be expected in normal WinUI applications.
Maybe we could consider propose new APIs in following interations that adding a new constructor without requiring a WindowId, which is not indented to be used in normal WinUI applications, but could be used in framework libraries or console applications. If we decide this is not a indented behaviour, we can add check for nullptr later.
The WindowsId is simply a value type as a wrapper to HWND, if we constructured manually with arbitary value, it might work like passing a non-relevant pointer to a API, which might be a undefined behaviour. In practice, it seems Common Item Dialog has underlying check on this, and an non-relevant value will behaviour like passing nullptr, does not throw a exception, can succefully pick a file, but for this case, Win32 API IsWindows might be able to used as check for this.

dev/Interop/StoragePickers/FolderPicker.cpp
Hong-Xiang: Since those types are implemented to work with Cpp/WinRT, personally I found duplicating similar code for those field/property definitions actually making things easier to maintain due to Cpp/WinRT's helper code generation mechanism. Some common logic does not related to Cpp/WinRT's interface are abstracted into Common 's cpp.

dev/Interop/StoragePickers/PickerCommon.cpp
Hong-Xiang: Thank you for your suggestion! A task was created for later follow-up,

dev/Interop/StoragePickers/PickerCommon.cpp
```diff
@@ -0,0 +1,213 @@
...
+        if (FAILED(hr))
+        {
+            return nullptr;
+        }
+
+        return defaultFolder;
+    }
+
+
+    winrt::hstring FormatExtensionWithWildcard(winrt::hstring extension)
```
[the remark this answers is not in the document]
Hong-Xiang: Thank you for your suggests! We've found similar feature request from developers in issues. We've created task for this including some other features not compatible with current UWP APIs.

dev/Interop/StoragePickers/Microsoft.Windows.Storage.Pickers.idl
Hong-Xiang: Thank you for your comment, we will adding it later before move to stable release.

dev/Interop/StoragePickers/FileSavePicker.cpp
Hong-Xiang: Thank you very much for your for catching this mistake, a guard for empty fileTypeChoices is necessary here.
For file save pickers, filter behavior might be different for open ones. for open picker we've decide to use wildcard fallback when calling pickers with empty filters. But for save pickers, since we are creating a new file, a concrete extension is required, thus we can not fall back to use wildcard files.

dev/Projections/CS/Microsoft.Windows.Storage.Pickers.Projection/Microsoft.Windows.Storage.Pickers.Projection.csproj
Hong-Xiang: Those project files are auto generated, I think this references might due to reference to `winrt::Microsoft::UI::WindowId` type in picker constructors

dev/Interop/StoragePickers/FileSavePicker.cpp
Hong-Xiang: We've added common logic into ConfigureDialog, and leave save pickers specific code here

.gitignore
Hong-Xiang: Thank you for your observations. Removed this change.

dev/Interop/StoragePickers/FileOpenPicker.cpp
Hong-Xiang: Thank you for your suggestions, updated all usage of cancellation token and resume background to the suggested pattern

dev/Interop/StoragePickers/PickerCommon.cpp
Hong-Xiang: Thank you for your questions. 
The buffer and return result are both "out" values. The problem here is the returned Win32 styles `COMDLG_FILTERSPEC` only contains raw pointers (get from `c_str()` method of `hstring`s), which does not automatically expand lifetime of strings dynamically constructed in this function. The buffer is used to carry those strings and is expected to expand lifetime of those string to the lifetime of setting Common Item Dialog options.
Comments has been updated to hopefully reduce confusion on this

dev/Interop/StoragePickers/FileOpenPicker.cpp
Hong-Xiang: Thank you for your questions, we've added a task to track on this, we need to decide the way of reporting exceptions, some times a Start without Stop might be a natural way of recording a unexpected exception, while some cases we may want to record those details. Updates on telemetry will be added later

dev/WindowsAppRuntime_DLL/WindowsAppRuntime_DLL.vcxproj
Hong-Xiang: Thank you for your guidance. The modifications related to shell32.lib and shell32.dll have been added to all configurations

dev/Interop/StoragePickers/FileOpenPicker.cpp
Hong-Xiang: Thank you for your suggestions! updated cancellation token usage in all async methods with calling enable_propagation

dev/Interop/StoragePickers/FileOpenPicker.cpp
Hong-Xiang: Thank you for your suggestions, updated code to following common usage styles

dev/Interop/StoragePickers/PickerCommon.cpp
Hong-Xiang: thank you for your suggestions, fixed to use same `(size_t)` conversion

dev/Interop/StoragePickers/PickerCommon.cpp
Hong-Xiang: Thank you for your suggestion, fixed to use const reference

dev/Interop/StoragePickers/PickerCommon.h
Hong-Xiang: Thank you for your suggestions, fixed.

test/StoragePickersTests/StoragePickersTests.cpp
Hong-Xiang: Thank you for your suggestion, fixed.

test/StoragePickersTests/StoragePickersTests.cpp
```diff
@@ -0,0 +1,211 @@
...
+                VERIFY_FAIL(L"Standard exception occurred during file save picker.");
+            }
+        }
+
+       */
+
+
+        TEST_METHOD(VerifyFileOpenPickerOptionsAreReadCorrectly)
+        {
+            auto parentWindow = ::GetForegroundWindow();
```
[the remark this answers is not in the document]
Hong-Xiang: Thank you for your suggestions, fixed to use default constructor for window id.

dev/Interop/StoragePickers/FileSavePicker.cpp
```diff
@@ -0,0 +1,176 @@
...
+        }
+
+        // Create a file. If the file already exists, will prompt to let user select cancele or override.
```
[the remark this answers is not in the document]
Hong-Xiang: Thank you for your question.
The misleading comments has been updated. Common Item Dialog will prompt to let user confirm override on existing file. If they cancelled then we would got Failed HResult and return nullptr earlier. If we reach code here, the file is expected to be truncated.

dev/Interop/StoragePickers/FileOpenPicker.cpp
```diff
@@ -0,0 +1,161 @@
...
+#include "pch.h"
+#include "FileOpenPicker.h"
+#include "Microsoft.Windows.Storage.Pickers.FileOpenPicker.g.cpp"
+#include "StoragePickersTelemetry.h"
+#include <windows.h>
+#include <shobjidl.h>
+#include <shobjidl_core.h>
+#include <winrt/Microsoft.UI.Interop.h>
+#include "PickerCommon.h"
+#include "PickFileResult.h"
```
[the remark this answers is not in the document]
Hong-Xiang: A code clean task is created to track this.

dev/Interop/StoragePickers/Microsoft.Windows.Storage.Pickers.idl
```diff
@@ -0,0 +1,85 @@
...
+    runtimeclass FolderPicker
+    {
+        FolderPicker(Microsoft.UI.WindowId windowId);
+
+        Microsoft.Windows.Storage.Pickers.PickerViewMode ViewMode;
+        String SettingsIdentifier;
+        Microsoft.Windows.Storage.Pickers.PickerLocationId SuggestedStartLocation;
+        String CommitButtonText;
+        Windows.Foundation.Collections.IVector<String> FileTypeFilter{ get; };
```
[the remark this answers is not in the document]
Hong-Xiang: Thank you very much for your question! It seems does not change behavior of folder picker, we'll follow investigation on this.

dev/Interop/StoragePickers/FileSavePicker.cpp
Hong-Xiang: Thank you very much for your question! This guides me to do some deeper investigation of UWP and our current experimental APIs behavior, and found out that there should be some refinement on our spec/implementation and we could resolve some existing limitations of UWP pickers. We'll continue improve this in following updates.

dev/Projections/CS/Microsoft.Windows.Storage.Pickers.Projection/Microsoft.Windows.Storage.Pickers.Projection.csproj
codendone: Ah, yes, WindowId makes sense. Thanks!

codendone (approved):

dev/Interop/StoragePickers/StoragePickersTelemetry.h
DinahK-2SO: Thanks Agnė for reminding! I added telemetries for the 3 pickers in recent iterations.

dev/Interop/StoragePickers/PickerCommon.cpp
DinahK-2SO: Thanks @codendone , Now we're refactoring the `CaptureFilterSpec` methods here #5434

dev/Interop/StoragePickers/FileSavePicker.cpp
DinahK-2SO: Hello @codendone  and @Hong-Xiang , please review #5619 

Adding validations.

PR #5260, opened 2025-03-21
API spec review: UserActivityHistory

llongley:
This adds an API spec for UserActivityHistory, which can be used to query user activity history in supported apps from the past 28 days.

### DIFF
--- a/specs/UserActivityHistory/UserActivityHistoryAPI.md
+++ b/specs/UserActivityHistory/UserActivityHistoryAPI.md
@@ -1,0 +1,243 @@
+UserActivityHistory
+===
+
+# Background
+
+The UserActivity class can be used to note down and preserve a record of activities that the user
+is currently doing on their computer - e.g., browsing a website, reading a Word document, etc.
+
+To record user activity, you use UserActivityChannel to retrieve a UserActivity object via the
+API GetOrCreateUserActivityAsync. If a UserActivity with the given ID already exists, it will be
+returned; otherwise, a new UserActivity object will be created and returned. You can then call
+the API GetSession to return a UserActivitySession object that tracks how long the user is engaged
+in that activity. This structure allows multiple sessions to be associated with the same activity,
+representing the case where the user completes that activity a bit at a time - e.g., beginning to
+watch a movie, then pausing, then watching more later. These will be treated as the same singular
+user activity that spans multiple sessions.
+
+UserActivityHistory is a new set of APIs that allow you to query the past 28 days of the user's
+activity history, which will enable you to bring back content that the user has previously been
+interacting with.
+
+# API Pages
+
+## UserActivityHistory class
+
+This class provides static methods that enable you to query the user's activity history.
+This activity history is stored in a database managed by a local service, and these APIs
+call out to that service to retrieve data from the database.
+
+Here is an example usage of the class that will enable you to bring back the webpage for a
+Korean recipe that the user had previously interacted with within the last day:
+
+```c#
+UserActivityHistoryQuery query = new();
+query.Keywords = new string[] { "Korean", "recipe" };
+query.LatestStartTime = DateTime.Now.AddDays(-1);
+
+IList<UserActivityHistoryItem> results = await UserActivityHistory.SearchAsync(
+    new UserActivityHistoryQuery[] { query },
+    UserActivityHistoryOrderBy.DwellTime,
+    maxResults: 1);
+
+UserActivityHistoryItem item = results.FirstOrDefault();
+
+if (item != null)
+{
+    // Now we can use item.ActivationUri to bring back the webpage in the state in which the user
+    // was last viewing it.
+}
+```
+
+## UserActivityHistory.Search method
+
+This method synchronously queries the user's activity history and returns a list of items matching
+the criteria specified in the `queries` parameter. The results are ordered in descending order
+according by the `orderBy` parameter: either by the most recent start times, the most recent
+end times, or the longest time spent on the activity.
+
+Each parameter in `queries` is ORed together in the resulting database query, whereas the contents
+of a single object in the `queries` array are ANDed together.  For example, if you provided two
+queries, each of which contained the keywords "tax", one of which had a content type of
+"application/pdf" and the other had a content type of "image/*", the resulting database query
+would be something along these lines:
+
+```sql
+SELECT * FROM UserActivityHistory WHERE
+    (CONTAINS (DisplayText, '"tax"')) AND
+    ((CONTAINS (ContentType, '"application/pdf"')) OR
+     (CONTAINS (ContentType, '"image/*"')));
+```
+
+## UserActivityHistory.SearchAsync method
+
+This method is an asynchronous version of the `Search` method.
+
+## UserActivityHistory.GetAppsWithUserActivity method
+
+This method synchronously retrieves a list of all the app names with data in the user's activity
+history database. You can use this, for example, to show the user the list of apps that are being
+queried against, so the user can understand why an app that is not recording user activity is not
+showing up in the results.
+
+## UserActivityHistory.GetAppsWithUserActivityAsync method
+
+This method is an asynchronous version of the `GetAppsWithUserActivity` method.
+
+## UserActivityHistoryItem class
+
+This class represents a single item in the user's activity history. It contains properties that
+describe in what app the activity occurred, what the nature of the activity was, the URI of the
+resource involved in the activity (e.g., a document, a webpage, a video, etc.), the URI that
+can be used to bring back the state the user left the activity in, and the times when the user
+started the activity and ended the activity.
+
+If the user performed the same activity multiple times, there will be multiple
+`UserActivityHistoryItem` objects returned, each with different start and end times.
+
+## UserActivityHistoryItem.AppName property
+
+This property contains the name of the app in which the activity occurred.
+
+## UserActivityHistoryItem.ActivityId property
+
+This property contains the ID of the activity, which can be used to collate multiple sessions
+of the same activity. For example, if the user watched a movie in multiple sessions, the
+`ActivityId` property can be used to identify how long in total the user spent watching that movie.
+
+## UserActivityHistoryItem.DisplayText property
+
+This property contains a string that is how the app chose to describe the activity. For example,
+if the activity was reading the contents of a webpage, this property might contain the webpage's title.
+
+## UserActivityHistoryItem.ContentType property
+
+This property contains the MIME type of the content being interacted with. For example, if the user
+was looking at a PNG image, this property would contain the string "image/png".
+
+## UserActivityHistoryItem.ContentUri property
+
+This property contains the URI of the content being interacted with. For example, if the user was
+looking at a webpage, this property would contain the URI of that webpage.
+
+## UserActivityHistoryItem.ActivationUri property
+
+This property contains the URI that can be used to bring back the state the user left the activity in.
+For example, if the user was looking at a webpage, this property would contain the URI of that webpage
+with additional information such as what page the user was on, what their scroll position was, etc.
+
+## UserActivityHistoryItem.StartTime property
+
+This property contains the time when the user started the activity session.
+
+## UserActivityHistoryItem.EndTime property
+
+This property contains the time when the user ended the activity session.
+
+## UserActivityHistoryQuery class
+
+This class is used to specify criteria for what portion of the user's activity history you want to
+retrieve. It allows you to specify keywords to search for, content types to filter by, and time ranges
+to filter by.
+
+## UserActivityHistoryQuery.Keywords property
+
+This property is an array of keywords, each of which is used to lexically search against the
+DisplayText column in the database. Keywords are case-insensitive, and results returned will
+be those that contain all of the keywords in the array.
+
+## UserActivityHistoryQuery.ContentType property
+
+This property is a string that specifies the content type associated with the activity you want
+to retrieve. It allows the inclusion of an asterisk as a wildcard - e.g., "image/*" will match
+all content types beginning with "image/", such as "image/png", "image/jpeg", etc.
+This property is case-insensitive.
+
+## UserActivityHistoryQuery.EarliestStartTime property
+
+This is a nullable DateTime property that specifies the earliest start time of the activity
+you want to retrieve. Any activities with a StartTime property earlier than this will be excluded.
+If this property is left as null, it will be ignored.
+
+## UserActivityHistoryQuery.EarliestEndTime property
+
+This is a nullable DateTime property that specifies the earliest end time of the activity
+you want to retrieve. Any activities with an EndTime property earlier than this will be excluded.
+If this property is left as null, it will be ignored.
+
+## UserActivityHistoryQuery.LatestStartTime property
+
+This is a nullable DateTime property that specifies the latest start time of the activity
+you want to retrieve. Any activities with a StartTime property later than this will be excluded.
+If this property is left as null, it will be ignored.
+
+## UserActivityHistoryQuery.LatestEndTime property
+
+This is a nullable DateTime property that specifies the latest end time of the activity
+you want to retrieve. Any activities with an EndTime property later than this will be excluded.
+If this property is left as null, it will be ignored.
+
+## UserActivityHistoryOrderBy enum
+
+This enum specifies what property the results should be ordered by.  The options are as follows:
+
+| Name | Description |
+|-|-|
+| StartTime | Results will be in descending order of their StartTime property |
+| EndTime | Results will be in descending order of their EndTime property |
+| DwellTime | Results will be in descending order of the difference between their EndTime and StartTime properties |
+
+# API Details
+
+```c# (but really MIDL3)
+namespace Microsoft.Windows.ApplicationModel.UserActivities
+{
+    runtimeclass UserActivityHistory
+    {
+        static IVector<UserActivityHistoryItem> Search(
+            UserActivityHistoryQuery[] queries,
+            UserActivityHistoryOrderBy orderBy,
+            UInt32 maxResults); 
+
+        static IAsyncOperation<IVector<UserActivityHistoryItem> > SearchAsync( 
+            UserActivityHistoryQuery[] queries, 
+            UserActivityHistoryOrderBy orderBy, 
+            UInt32 maxResults);
+
+        static IVector<String> GetAppsWithUserActivity();
+
+        static IAsyncOperation<IVector<String>> GetAppsWithUserActivityAsync();
+    }
+
+    runtimeclass UserActivityHistoryItem
+    {
+        String AppName { get; };
+        String ActivityId { get; };
+        String DisplayText { get; };
+        String ContentType { get; };
+        String ContentUri { get; };
+        String ActivationUri { get; };
+        DateTime StartTime { get; };
+        DateTime EndTime { get; };
+    }
+
+    runtimeclass UserActivityHistoryQuery
+    {
+        UserActivityHistoryQuery();
+
+        String[] Keywords;
+        String ContentType;
+        IReference<DateTime> EarliestStartTime;
+        IReference<DateTime> EarliestEndTime;
+        IReference<DateTime> LatestStartTime;
+        IReference<DateTime> LatestEndTime;
+    }
+
+    enum UserActivityHistoryOrderBy
+    {
+        StartTime,
+        EndTime,
+        DwellTime 
+    };
+}
+```
--- a/specs/UserActivityHistory/UserActivityHistoryAPI.md
+++ b/specs/UserActivityHistory/UserActivityHistoryAPI.md
@@ -3,21 +3,43 @@ UserActivityHistory
 
 # Background
 
-The UserActivity class can be used to note down and preserve a record of activities that the user
+The [UserActivity](https://learn.microsoft.com/uwp/api/windows.applicationmodel.useractivities.useractivity)
+class can be used to note down and preserve a record of activities that the user
 is currently doing on their computer - e.g., browsing a website, reading a Word document, etc.
+This allows Windows to have insight into the application state, enabling smart experiences that are
+built around the semantics of the app. For example, a document editor can give Windows information
+about the document that the user is editing, so that Recall can later take the user to the document
+at the same location.
 
-To record user activity, you use UserActivityChannel to retrieve a UserActivity object via the
-API GetOrCreateUserActivityAsync. If a UserActivity with the given ID already exists, it will be
-returned; otherwise, a new UserActivity object will be created and returned. You can then call
-the API GetSession to return a UserActivitySession object that tracks how long the user is engaged
-in that activity. This structure allows multiple sessions to be associated with the same activity,
-representing the case where the user completes that activity a bit at a time - e.g., beginning to
-watch a movie, then pausing, then watching more later. These will be treated as the same singular
-user activity that spans multiple sessions.
+To record user activity, an app uses [UserActivityChannel](https://learn.microsoft.com/en-us/uwp/api/windows.applicationmodel.useractivities.useractivitychannel?view=winrt-26100)
+to retrieve a UserActivity object via the API [GetOrCreateUserActivityAsync](https://learn.microsoft.com/en-us/uwp/api/windows.applicationmodel.useractivities.useractivitychannel.getorcreateuseractivityasync?view=winrt-26100#windows-applicationmodel-useractivities-useractivitychannel-getorcreateuseractivityasync(system-string)).
+If a UserActivity with the given ID already exists, it will be returned; otherwise, a new UserActivity
+object will be created and returned. You can then call the API [GetSession](https://learn.microsoft.com/en-us/uwp/api/windows.applicationmodel.useractivities.useractivity.createsession?view=winrt-26100#windows-applicationmodel-useractivities-useractivity-createsession)
+to return a [UserActivitySession](https://learn.microsoft.com/en-us/uwp/api/windows.applicationmodel.useractivities.useractivitysession?view=winrt-26100)
+object that tracks how long the user is engaged in that activity. This structure allows multiple
+sessions to be associated with the same activity, representing the case where the user completes
+that activity a bit at a time - e.g., beginning to watch a movie, then pausing, then watching more later.
+These will be treated as the same singular user activity that spans multiple sessions.
 
-UserActivityHistory is a new set of APIs that allow you to query the past 28 days of the user's
-activity history, which will enable you to bring back content that the user has previously been
-interacting with.
+UserActivityHistory is a new set of APIs that allow you to query up to the past 28 days of the
+user's activity history, which will enable you to bring back content that the user has previously
+been interacting with.
+
+# Conceptual pages (How To)
+
+The intended use case of this API is to allow you to make search queries against the user's activity history
+on their local computer.  The string matching in the query parameters in this API is lexical in nature,
+meaning that it is expected that any natural language semantic parsing of the user's input will be done by
+your app prior to calling this API.
+
+For example, if a user types in something along the lines of, "Please find the Korean recipe I was looking at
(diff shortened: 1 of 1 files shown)


specs/UserActivityHistory/UserActivityHistoryAPI.md
```diff
@@ -0,0 +1,243 @@
+UserActivityHistory
+===
+
+# Background
+
+The UserActivity class can be used to note down and preserve a record of activities that the user
```
asklar, line 6:
```suggestion
The [UserActivity](https://learn.microsoft.com/uwp/api/windows.applicationmodel.useractivities.useractivity
) class can be used to record activities that the user
```

specs/UserActivityHistory/UserActivityHistoryAPI.md
```diff
@@ -0,0 +1,243 @@
...
+is currently doing on their computer - e.g., browsing a website, reading a Word document, etc.
+
```
asklar, line 8: "This allows Windows to have insight into the application state, enabling smart experiences that are built around the semantics of the app. For example, a document editor can give Windows information about the document that the user is editing, so that Recall can later take the user to the document at the same location."

specs/UserActivityHistory/UserActivityHistoryAPI.md
```diff
@@ -0,0 +1,243 @@
...
+To record user activity, you use UserActivityChannel to retrieve a UserActivity object via the
```
asklar, line 9: nit, change "you" to "an app"

specs/UserActivityHistory/UserActivityHistoryAPI.md
```diff
@@ -0,0 +1,243 @@
...
+API GetOrCreateUserActivityAsync. If a UserActivity with the given ID already exists, it will be
+returned; otherwise, a new UserActivity object will be created and returned. You can then call
+the API GetSession to return a UserActivitySession object that tracks how long the user is engaged
+in that activity. This structure allows multiple sessions to be associated with the same activity,
+representing the case where the user completes that activity a bit at a time - e.g., beginning to
+watch a movie, then pausing, then watching more later. These will be treated as the same singular
+user activity that spans multiple sessions.
+
+UserActivityHistory is a new set of APIs that allow you to query the past 28 days of the user's
```
asklar, line 18: 28 days --> is this a parameter of how the history is stored? e.g. can I say how long to keep this for? (maybe with a system-defined upper limit)?

specs/UserActivityHistory/UserActivityHistoryAPI.md
```diff
@@ -0,0 +1,243 @@
...
+This class provides static methods that enable you to query the user's activity history.
+This activity history is stored in a database managed by a local service, and these APIs
+call out to that service to retrieve data from the database.
+
+Here is an example usage of the class that will enable you to bring back the webpage for a
+Korean recipe that the user had previously interacted with within the last day:
+
+```c#
+UserActivityHistoryQuery query = new();
+query.Keywords = new string[] { "Korean", "recipe" };
```
asklar, line 35: can I just pass "Korean recipe" instead? what is the criteria for choosing to wordbreak or not?

specs/UserActivityHistory/UserActivityHistoryAPI.md
```diff
@@ -0,0 +1,243 @@
...
+query.LatestStartTime = DateTime.Now.AddDays(-1);
+
+IList<UserActivityHistoryItem> results = await UserActivityHistory.SearchAsync(
```
asklar, line 38: this says SearchAsync but below says Search?

specs/UserActivityHistory/UserActivityHistoryAPI.md
```diff
@@ -0,0 +1,243 @@
...
+```
+
+## UserActivityHistory.Search method
+
+This method synchronously queries the user's activity history and returns a list of items matching
+the criteria specified in the `queries` parameter. The results are ordered in descending order
+according by the `orderBy` parameter: either by the most recent start times, the most recent
+end times, or the longest time spent on the activity.
+
+Each parameter in `queries` is ORed together in the resulting database query, whereas the contents
```
asklar, line 59: I think the db is an implementation detail so I'd recommend focusing more on "can I just enter a full natural language sentence, or just keywords"? 
For example, if the API took a single NL string, then the system could extract keywords from it, and then do the lookup in the database with those keywords as an implementation detail.

specs/UserActivityHistory/UserActivityHistoryAPI.md
asklar, line 9: Conceptual Pages section?

specs/UserActivityHistory/UserActivityHistoryAPI.md
```diff
@@ -0,0 +1,243 @@
...
+of a single object in the `queries` array are ANDed together.  For example, if you provided two
```
asklar, line 60: I think we need a paragraph (and sample) before this that describes content type, because without it this sentence is hard to follow.

specs/UserActivityHistory/UserActivityHistoryAPI.md
```diff
@@ -0,0 +1,243 @@
...
+```sql
+SELECT * FROM UserActivityHistory WHERE
+    (CONTAINS (DisplayText, '"tax"')) AND
+    ((CONTAINS (ContentType, '"application/pdf"')) OR
+     (CONTAINS (ContentType, '"image/*"')));
+```
+
+## UserActivityHistory.SearchAsync method
+
+This method is an asynchronous version of the `Search` method.
```
asklar, line 74: do we even need this? apps can just spin up a thread, call the sync version and wait for that.

asklar: here and others -- if we don't have a good reason for async versions, consider removing.

specs/UserActivityHistory/UserActivityHistoryAPI.md
```diff
@@ -0,0 +1,243 @@
...
+
+## UserActivityHistory.GetAppsWithUserActivity method
+
+This method synchronously retrieves a list of all the app names with data in the user's activity
+history database. You can use this, for example, to show the user the list of apps that are being
+queried against, so the user can understand why an app that is not recording user activity is not
+showing up in the results.
```
asklar, line 81: shouldn't there be a way to get the list of apps for a particular content type? for example if I want to search for documents, it doesn't help me to know that my music player has 20'000 songs added to my user history.

specs/UserActivityHistory/UserActivityHistoryAPI.md
```diff
@@ -0,0 +1,243 @@
...
+
+## UserActivityHistory.GetAppsWithUserActivityAsync method
+
+This method is an asynchronous version of the `GetAppsWithUserActivity` method.
+
+## UserActivityHistoryItem class
+
+This class represents a single item in the user's activity history. It contains properties that
+describe in what app the activity occurred, what the nature of the activity was, the URI of the
```
asklar, line 90: can you clarify what you mean by "nature of the activity"?

specs/UserActivityHistory/UserActivityHistoryAPI.md
```diff
@@ -0,0 +1,243 @@
...
+resource involved in the activity (e.g., a document, a webpage, a video, etc.), the URI that
+can be used to bring back the state the user left the activity in, and the times when the user
+started the activity and ended the activity.
```
asklar, line 93: nit, "started and ended the activity"

specs/UserActivityHistory/UserActivityHistoryAPI.md
```diff
@@ -0,0 +1,243 @@
...
+
+If the user performed the same activity multiple times, there will be multiple
+`UserActivityHistoryItem` objects returned, each with different start and end times.
+
+## UserActivityHistoryItem.AppName property
+
+This property contains the name of the app in which the activity occurred.
```
asklar, line 100: what kind of name is this? a PFN? an AUMID? display name? exe path?
Is this something Windows infers from the caller, or the app provided this when recording the UserActivity?

specs/UserActivityHistory/UserActivityHistoryAPI.md
```diff
@@ -0,0 +1,243 @@
...
+
+## UserActivityHistoryItem.DisplayText property
+
+This property contains a string that is how the app chose to describe the activity. For example,
+if the activity was reading the contents of a webpage, this property might contain the webpage's title.
+
+## UserActivityHistoryItem.ContentType property
+
+This property contains the MIME type of the content being interacted with. For example, if the user
+was looking at a PNG image, this property would contain the string "image/png".
```
asklar, line 116: is this be "anything the app gave Windows", or "one from a list of known MIME types"?
e.g. can Contoso say "contoso/bloop"?

specs/UserActivityHistory/UserActivityHistoryAPI.md
```diff
@@ -0,0 +1,243 @@
...
+
+## UserActivityHistoryItem.ContentUri property
+
+This property contains the URI of the content being interacted with. For example, if the user was
+looking at a webpage, this property would contain the URI of that webpage.
```
asklar, line 121: what happens if the content is not URI backed?
For example, you were creating a new doc -- autosave was off so no URI. Or, the app generated an image with stable diffusion. Is it a requirement that all UserActivity content can be referenced via URI?
And also I'm assuming that if the content goes away, we don't necessarily know about it?

specs/UserActivityHistory/UserActivityHistoryAPI.md
```diff
@@ -0,0 +1,243 @@
...
+
+## UserActivityHistoryItem.ActivationUri property
+
+This property contains the URI that can be used to bring back the state the user left the activity in.
+For example, if the user was looking at a webpage, this property would contain the URI of that webpage
+with additional information such as what page the user was on, what their scroll position was, etc.
```
asklar, line 127: what page the user was on --> that's the content uri already, no?

specs/UserActivityHistory/UserActivityHistoryAPI.md
```diff
@@ -0,0 +1,243 @@
...
+
+## UserActivityHistoryItem.StartTime property
+
+This property contains the time when the user started the activity session.
+
+## UserActivityHistoryItem.EndTime property
+
+This property contains the time when the user ended the activity session.
```
asklar, line 135: can a UA exist with a start and no end?

specs/UserActivityHistory/UserActivityHistoryAPI.md
```diff
@@ -0,0 +1,243 @@
...
+
+## UserActivityHistoryQuery class
+
+This class is used to specify criteria for what portion of the user's activity history you want to
+retrieve. It allows you to specify keywords to search for, content types to filter by, and time ranges
+to filter by.
+
+## UserActivityHistoryQuery.Keywords property
+
+This property is an array of keywords, each of which is used to lexically search against the
```
asklar, line 145: there is no semantic search? If there's a misspelling or alternate spelling of a keyword, the system won't find it?
e.g. en-US "vacation", en-UK "holiday"

specs/UserActivityHistory/UserActivityHistoryAPI.md
```diff
@@ -0,0 +1,243 @@
...
+DisplayText column in the database. Keywords are case-insensitive, and results returned will
+be those that contain all of the keywords in the array.
```
asklar, line 147: I think this means apps are incentivized to ask for the least number of keywords, which seems counterintuitive?

specs/UserActivityHistory/UserActivityHistoryAPI.md
```diff
@@ -0,0 +1,243 @@
...
+
+## UserActivityHistoryQuery.ContentType property
+
```
asklar, line 150: can the asterisk appear anywhere or only as the value of the subtype?

specs/UserActivityHistory/UserActivityHistoryAPI.md
```diff
@@ -0,0 +1,243 @@
...
+This property is a string that specifies the content type associated with the activity you want
+to retrieve. It allows the inclusion of an asterisk as a wildcard - e.g., "image/*" will match
+all content types beginning with "image/", such as "image/png", "image/jpeg", etc.
+This property is case-insensitive.
+
+## UserActivityHistoryQuery.EarliestStartTime property
+
+This is a nullable DateTime property that specifies the earliest start time of the activity
```
asklar, line 158: conceptually, nullable -> optional

asklar: here and below

specs/UserActivityHistory/UserActivityHistoryAPI.md
```diff
@@ -0,0 +1,243 @@
...
+        static IAsyncOperation<IVector<String>> GetAppsWithUserActivityAsync();
+    }
+
+    runtimeclass UserActivityHistoryItem
+    {
+        String AppName { get; };
+        String ActivityId { get; };
+        String DisplayText { get; };
+        String ContentType { get; };
+        String ContentUri { get; };
```
asklar, line 218: what can the app do with this though? in general this won't be useful to apps other than the originator of this content, unless it's public, due to needing to authenticate.

specs/UserActivityHistory/UserActivityHistoryAPI.md
asklar, line 18: - can any app query any other app's UserActivity?
- can any apps call this api or is there some sort of capability?

llongley: 28 days is an arbitrarily-defined length of time that seemed like an amount that would allow users to accomplish everything they want to do without storing more than would be required.  There isn't anything in the current design that would allow users to set it to a different amount, though if we wanted to add something to settings, there also isn't anything that would prevent us from doing that.

I can add an "up to" comment to cover the possibility of us adding a setting later.

specs/UserActivityHistory/UserActivityHistoryAPI.md
llongley: "Korean recipe" would match those exact words in that order with the space in the middle.  Passing in the two separately will match any display text containing "Korean" and "recipe".  So "A recipe for Korean food", "A Korean bibimbap recipe" would match "Korean" and "recipe", but would not match "Korean recipe".

specs/UserActivityHistory/UserActivityHistoryAPI.md
llongley: Search and SearchAsync are both part of the API spec.  One is just an async version of the other, for convenience.  I can change this to use the sync version of the API if that creates confusion, though.

specs/UserActivityHistory/UserActivityHistoryAPI.md
llongley: I wasn't sure if this needed one, given that the template said it was optional, but I can add one.

specs/UserActivityHistory/UserActivityHistoryAPI.md
llongley: I've updated the spec to ignore the existence of the database in lieu of just talking about behavior.  NL processing is out of scope for this API - it's expected that the app will perform that and will then pass the result into this API.

specs/UserActivityHistory/UserActivityHistoryAPI.md
llongley: Good point, I'll add an explanation.

specs/UserActivityHistory/UserActivityHistoryAPI.md
llongley: You could do that with the Search API - e.g., if you want to find all apps interacting with PNG images, you could do something like

```c#
UserActivityHistoryQuery query = new();
query.ContentType = "image/png";

IList<UserActivityHistoryItem> imageActivity = UserActivityHistory.Search(
    new UserActivityHistoryQuery[] { query },
    UserActivityHistoryOrderBy.DwellTime,
    maxResults: 50);

List<string> appNames = imageActivity.Select(item => item.AppName).Distinct().ToList();
```

The intention for this API is to be able to allow users to see that an app they might have been expecting to be searched (e.g. Adobe Photoshop, perhaps) is not actually reporting user activity.

specs/UserActivityHistory/UserActivityHistoryAPI.md
llongley: Updated to explicitly call out what's in a user activity object.

specs/UserActivityHistory/UserActivityHistoryAPI.md
llongley: My current implementation contains the exe path, but it would be even better if there were a way to get the display name.  I can't immediately find one.  We have to infer this from the caller; the UserActivity object does not have this property anywhere.

specs/UserActivityHistory/UserActivityHistoryAPI.md
llongley: It's a string property, so it can be anything.  I can make that clear in the description.

specs/UserActivityHistory/UserActivityHistoryAPI.md
llongley: The primary purpose of this API is to be able to retrieve something that the user was previously working with, so if there isn't an ActivationUri populated, then we won't add that UserActivity to the database.

llongley: I'll add a comment to that end.

specs/UserActivityHistory/UserActivityHistoryAPI.md
llongley: Probably, yeah.  I'll remove that.

specs/UserActivityHistory/UserActivityHistoryAPI.md
llongley: No - we save a UserActivity when a session for it ends, and it won't have ended without having previously started.

specs/UserActivityHistory/UserActivityHistoryAPI.md
llongley: Correct - we're expecting that the app will have done the heavy lifting of semantic parsing, etc.  The specific scenario we're picturing for this API is an agentic AI app that will construct the parameters to pass into this API.

We certainly might be able to add semantic search later, but that is currently not in scope for this API.

specs/UserActivityHistory/UserActivityHistoryAPI.md
llongley: Not necessarily - if you want to search for keyword A or keyword B, you can pass in two separate queries.  The motivation behind requiring that we match all of the keywords is to make it so that you can be more specific.  For example, you can search for both "Korean" and "recipe" to avoid getting, for example, a webpage you visited on Korean history.

specs/UserActivityHistory/UserActivityHistoryAPI.md
llongley: Anywhere.

specs/UserActivityHistory/UserActivityHistoryAPI.md
llongley: Changed.

specs/UserActivityHistory/UserActivityHistoryAPI.md
llongley: As long as the resource is something that can be accessed via a URI, e.g. a webpage, a document, etc., that URI can be used to bring it up to the user, which is the primary idea behind this API.

specs/UserActivityHistory/UserActivityHistoryAPI.md
llongley: Oh, good point, I forgot to include a discussion on this.  An app can query any user activity that has been recorded on the same computer for the same user, but only if a) the user grants permission to access user activity history via a consent dialog, and b) the app is Microsoft logo certified, which basically means that we know what the app is and have given it the OK.  A random exe downloaded off the internet will not be able to access this data.

specs/UserActivityHistory/UserActivityHistoryAPI.md
```diff
@@ -3,21 +3,43 @@ UserActivityHistory
...
+on their local computer.  The string matching in the query parameters in this API is lexical in nature,
+meaning that it is expected that any natural language semantic parsing of the user's input will be done by
+your app prior to calling this API.
+
+For example, if a user types in something along the lines of, "Please find the Korean recipe I was looking at
+earlier today", your app might have an agentic AI parse that input and determine that the user is looking
+for a webpage that contains the keywords "Korean" and "recipe", and construct a query with those keywords,
+a content type of "text/html", and an access time within the last 24 hours.
+
+In order for an app to make use of this API, it must be Windows logo certified, and the user must provide
```
asklar, line 40: link to "how do I get the Windows logo certification" page?

specs/UserActivityHistory/UserActivityHistoryAPI.md
asklar: we can use CallerIdentity or similar (e.g. CoGetCallContext) to get an AUMID. When an app receives this AUMID they can find the display name of that AUMID for display purposes. We could choose to store the display name too, because the app might get uninstalled sometime after capturing the user activity and before querying it (what happens with that app's user activity history, does it get deleted?)

In any case let's take an action item to update the wording here once we have a solid caller id implementation

specs/UserActivityHistory/UserActivityHistoryAPI.md
asklar: thanks! that's fine for a v1, but tagging @andrewleader to consider this as an improvement

specs/UserActivityHistory/UserActivityHistoryAPI.md
```diff
@@ -0,0 +1,266 @@
...
+
+```c# (but really MIDL3)
+namespace Microsoft.Windows.ApplicationModel.UserActivities
+{
+    runtimeclass UserActivityHistory
+    {
+        static IVector<UserActivityHistoryItem> Search(
+            UserActivityHistoryQuery[] queries,
+            UserActivityHistoryOrderBy orderBy,
+            UInt32 maxResults);
```
ptorr-msft, line 223: Is there a material benefit to passing maxResults (e.g. perf? or making the user feel better if we include this as part of the consent prompt?).

specs/UserActivityHistory/UserActivityHistoryAPI.md
```diff
@@ -0,0 +1,266 @@
...
+            UserActivityHistoryQuery[] queries,
+            UserActivityHistoryOrderBy orderBy,
+            UInt32 maxResults); 
+
+        static IAsyncOperation<IVector<UserActivityHistoryItem> > SearchAsync( 
+            UserActivityHistoryQuery[] queries, 
+            UserActivityHistoryOrderBy orderBy, 
+            UInt32 maxResults);
+
+        static IVector<String> GetAppsWithUserActivity();
```
ptorr-msft, line 230: Should this be a list of ProductIds rather than just strings? What data do we have from unpackaged apps recording activities (like Office)?

specs/UserActivityHistory/UserActivityHistoryAPI.md
ptorr-msft: Also, need to take into account whatever consent model we end up with - one time, every time, does it include context, etc?

specs/UserActivityHistory/UserActivityHistoryAPI.md
```diff
@@ -0,0 +1,266 @@
...
+| Name | Description |
+|-|-|
+| StartTime | Results will be in descending order of their StartTime property |
+| EndTime | Results will be in descending order of their EndTime property |
+| DwellTime | Results will be in descending order of the difference between their EndTime and StartTime properties |
+
+# API Details
+
+```c# (but really MIDL3)
+namespace Microsoft.Windows.ApplicationModel.UserActivities
```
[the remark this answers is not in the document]
DrusTheAxe: Beat me to it.

`Microsoft.Windows.AI.UserActivities` seems more apt

specs/UserActivityHistory/UserActivityHistoryAPI.md
```diff
@@ -0,0 +1,266 @@
...
+Here is an example usage of the class that will enable you to bring back the webpage for a
+Korean recipe that the user had previously interacted with within the last day:
+
+```c#
+UserActivityHistoryQuery query = new();
+query.Keywords = new string[] { "Korean", "recipe" };
+query.LatestStartTime = DateTime.Now.AddDays(-1);
+
+IList<UserActivityHistoryItem> results = UserActivityHistory.Search(
+    new UserActivityHistoryQuery[] { query },
```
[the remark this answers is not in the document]
DrusTheAxe: RECOMMEND: Add an overload for 1 query

specs/UserActivityHistory/UserActivityHistoryAPI.md
```diff
@@ -0,0 +1,266 @@
...
+    UserActivityHistoryOrderBy.DwellTime,
+    maxResults: 1);
+
+UserActivityHistoryItem item = results.FirstOrDefault();
+
```
DrusTheAxe, line 66: RECOMMEND: Remove blank line

specs/UserActivityHistory/UserActivityHistoryAPI.md
```diff
@@ -0,0 +1,266 @@
...
+which is the [MIME type](https://docs.w3cub.com/http/basics_of_http/mime_types/complete_list_of_mime_types.html)
+of the resource the user was interacting with.  The search will match if the ContentType of the item matches
+the ContentType property of the query.  The ContentType supports using an asterisk as a wildcard
+to match a range of content types - e.g., "image/*" will match "image/png", "image/jpeg", etc.
+
+## UserActivityHistory.GetAppsWithUserActivity method
+
+This method synchronously retrieves a list of all the app names with data in the user's activity
+history database. You can use this, for example, to show the user the list of apps that are being
+queried against, so the user can understand why an app that is not recording user activity is not
```
DrusTheAxe, line 101: >...why an app not recording isn't showing up in the list of apps that recorded...

Typo? Should be "...what an app <strike>that is not</strike> recording user activity is not showing up..." ?

specs/UserActivityHistory/UserActivityHistoryAPI.md
DrusTheAxe: >My current implementation contains the exe path

Is this an unpackaged app or packaged?

If packaged you'll want to record the package full name. Given that you can lookup its DisplayName (and Logo) localized for the current user to view.

Is this _historical_? Does UserActivityHistory retain information recorded by apps after they're uninstalled? If so then you can't guarantee looking up the DisplayName. If so there are options but they have caveats so I'll wait to hear if relevant before saying more.

>but it would be even better if there were a way to get the display name

```
p = packageManager.FindPackageForUser("", pkgfullname)
string displayName = p.DisplayName
```

returns the package's DisplayName localized for the calling user. Are there cases where the package isn't registered for the calling user?

specs/UserActivityHistory/UserActivityHistoryAPI.md
```diff
@@ -0,0 +1,266 @@
...
+            UInt32 maxResults);
+
+        static IVector<String> GetAppsWithUserActivity();
+
+        static IAsyncOperation<IVector<String>> GetAppsWithUserActivityAsync();
+    }
+
+    runtimeclass UserActivityHistoryItem
+    {
+        String AppName { get; };
```
DrusTheAxe, line 237: What is `AppName`?

If a packaged app is this the app's AUMID (programmatic id) or DisplayName (localized string for human consumption)?

Does this API support unpackaged apps?

lauren-ciha:

PR #5261, opened 2025-03-22
[1.7] 56439870: A/B Containment for EnsurePackage*ReadyAsync()

DrusTheAxe:
56439870: A/B Containment for EnsurePackage*ReadyAsync() doesn't register v2 if v1 is registered and caller asks for minVersion=v1

Adds A/B Containment for [[1.7] EnsurePackage*ReadyAsync() doesn't register v2 if v1 is registered and caller asks for minVersion=v1 #5182 #5191](https://github.com/microsoft/WindowsAppSDK/pull/5191)

https://task.ms/56439870

### DIFF
--- a/dev/PackageManager/API/M.W.M.D.PackageDeploymentManager.cpp
+++ b/dev/PackageManager/API/M.W.M.D.PackageDeploymentManager.cpp
@@ -34,6 +34,8 @@ namespace ABI::Windows::Management::Deployment
 
 #include <IsWindowsVersion.h>
 
+#include "PackageManager.Containment.h"
+
 static_assert(static_cast<int>(winrt::Microsoft::Windows::Management::Deployment::StubPackageOption::Default) == static_cast<int>(winrt::Windows::Management::Deployment::StubPackageOption::Default),
               "winrt::Microsoft::Windows::Management::Deployment::StubPackageOption::Default != winrt::Windows::Management::Deployment::StubPackageOption::Default");
 static_assert(static_cast<int>(winrt::Microsoft::Windows::Management::Deployment::StubPackageOption::InstallFull) == static_cast<int>(winrt::Windows::Management::Deployment::StubPackageOption::InstallFull),
--- a/dev/PackageManager/API/M.W.M.D.PackageDeploymentManager.cpp
+++ b/dev/PackageManager/API/M.W.M.D.PackageDeploymentManager.cpp
@@ -1930,7 +1930,7 @@ namespace winrt::Microsoft::Windows::Management::Deployment::implementation
         activityId = winrt::guid{};
 
         winrt::Microsoft::Windows::Management::Deployment::PackageReadyOrNewerAvailableStatus readyOrNewerStatus{};
-        if (options.RegisterNewerIfAvailable())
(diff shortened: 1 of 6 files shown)


dev/PackageManager/API/M.W.M.D.PackageDeploymentManager.cpp
```diff
@@ -1930,7 +1930,7 @@ namespace winrt::Microsoft::Windows::Management::Deployment::implementation
         activityId = winrt::guid{};
 
         winrt::Microsoft::Windows::Management::Deployment::PackageReadyOrNewerAvailableStatus readyOrNewerStatus{};
-        if (options.RegisterNewerIfAvailable())
+        if (WinAppSdk::Containment::IsChangeEnabled<WINAPPSDK_CHANGEID_56439870>() && options.RegisterNewerIfAvailable())
```
codendone, line 1933: >WinAppSdk::Containment::IsChangeEnabled<WINAPPSDK_CHANGEID_56439870>() [](http://example.com/codeflow?start=12&length=70)

Is just doing the else{} block the same as before #5191? In other words, `IsReady()` will return always return true at the same times `IsReadyOrNewerAvailable()` will return Ready? The code is different enough that I can't easily tell.

DrusTheAxe: Yes

Code used to be (essentially)

```
bool isReady{ IsReady(p) }
if (isReady) { return S_OK }
...
m_packageManager.AddPackageByUriAsync(p)
```

Then with ReadyOrNewerAvailable it effectively became

```
ReadyOrNewerAvailableStatus /*NotReady|Ready|NewerAvailable/* readyStatus{}
if (options.ReadyOfNewerAvailable)
{
    readyStatus = IsReadyOrNewerAvailable(p)
}
else
{
    readyStatus = IsReady(p) ? Ready | NotReady;
}
bool isReady{ readyStatus == Ready || readyStatys == readyOrNewerAvailable }
if (isReady) { return S_OK }
...
if (readyStatus == NewerAvailable)
{
    co_return m_packageManager.RegisterPackageByPackageFamilyNameAsync(pkgfamilyname)
}
...
m_packageManager.AddPackageByUriAsync(p)
```

1.7.0 behavior = ignore options.ReadyOfNewerAvailable (i.e. treat as always false). This way we only call IsReady() and only have 2 possible answers -- NotReady or Ready -- and we always hit the AddPackage call. Since we never call IsReadyOrNewerAvailable() in this case we can never get the 3rd status -- NewerAvailable -- and thus will never hit the RegisterPackage call.

Simplest way to guarantee this behavior is to only check `options.RegisterNewerIfAvailable()` if 1.7.1+ behavior is desired, and thus

```
if (options.RegisterNewerIfAvailable())
```

becomes

```
if (WinAppSdk::Containment::IsChangeEnabled<WINAPPSDK_CHANGEID_56439870>() && options.RegisterNewerIfAvailable())
```

cwruss (approved):

SaChinta-MSFT (approved):

codendone (approved):

codendone (approved):

PR #5278, opened 2025-03-27
PackageDeploymentManager - Fix .Stop() events not being generated. Include PackageReadyOrNewerAvailableStatus in .Stop() events.

DrusTheAxe:
Fixed .Stop() events to be called before co_return...CompletedSuccess (rather than after and thus not actually be called).

Add PackageReadyOrNewerAvailableStatus to .Stop() events for EnsurePackageReadyAsync and EnsurePackageSetReadyAsync.

NOTE: EnsurePackageReadyByUriAsync calls EnsurePackageSetReadyAsync to do its heavy lifting so see the nested event for this data.

Fix compiler warnings


dabhattimsft:
>         {

Can we have an event here to tell if packageSet was already ready?

---
Refers to: dev/PackageManager/API/M.W.M.D.PackageDeploymentManager.cpp:472 in 0562da1. [](commit_id = 0562da1f0ed3a71dea30c5205324756bdd745977, deletion_comment = False)

DrusTheAxe:
> > ```
> >                 PackageDeploymentStatus::CompletedFailure, activityId, error, extendedError, errorText);
> > ```
> 
> can we expand activity stop event to have package deployment result?
> 
> Refers to: dev/PackageManager/API/M.W.M.D.PackageDeploymentManager.cpp:504 in [0562da1](https://github.com/microsoft/WindowsAppSDK/commit/0562da1f0ed3a71dea30c5205324756bdd745977). [](commit_id = [0562da1](https://github.com/microsoft/WindowsAppSDK/commit/0562da1f0ed3a71dea30c5205324756bdd745977), deletion_comment = False)

This is a thin wrapper over Add|Ensure calls so those will have the results you're looking for. Adding that info here is limited value and high complexity so the ROI wasn't worthy. You'll have an event with a nested event so later analysis has the data.

If this is problematic come hunt me up to discuss.

PR #5336, opened 2025-04-15
Add ABForward Tests

Dreynor87:
Add ABForward Tests.

The goal of these tests are to show that code that is built with version N, but wants to run with version N-1, as long as it only uses objects that are available in N-1, the code will still work as it would have had the code built with version N-1.

The tests build with the version in the repo, but are set to dynamically load the runtime version 1.6. We then attempt to create multiple objects. If those objects are available in 1.6, then the object creation is succeeding (as shown by the lack of thrown exception). If those objects are not available in 1.6, we do see the thrown exception. We are using those objects mostly to show that we are indeed running with a different version than we are building with.

The tests succeed when running locally:
image


Dreynor87:
@microsoft-github-policy-service agree [company="{Microsoft}"]

Dreynor87:

Dreynor87:
> Is there a plan to add a self-contained AB test as well? See [here](https://github.com/microsoft/WinUI-Gallery/compare/main...jessecol/ab-attempt) for inspiration.

I expect we would want to eventually add AB Forward Tests for all types possible. The Dynamic Dependency one is just the first one we are doing.

test/ABForward/FunctionalTests.cpp
```diff
@@ -0,0 +1,134 @@
...
+namespace Test::ABForward
+{
+    class FunctionalTests
+    {
+    private:
+        void AddPackages()
+        {
+            auto packageManager{ Windows::Management::Deployment::PackageManager() };
+
+            auto frameworkPackages{ packageManager.FindPackages(L"Microsoft.WindowsAppRuntime.1.6_8wekyb3d8bbwe") };
```
Scottj1s, line 30: here too - could pass version number msbuild property as cl.exe additional definition

test/ABForward/FunctionalTests.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+            if (!frameworkPackages.First().HasCurrent())
+            {
+                auto frameworkUri{ Windows::Foundation::Uri(Test::Packages::GetMsixPackagePath(L"Microsoft.WindowsAppRuntime.1.6").c_str()) };
+                auto frameworkOp{ packageManager.AddPackageAsync(frameworkUri, nullptr, DeploymentOptions::None) };
+                auto frameworkResult{ frameworkOp.get() };
+            }
+
+            auto ddlmUri{ Windows::Foundation::Uri(Test::Packages::GetMsixPackagePath(L"Microsoft.WindowsAppRuntime.DDLM.1.6").c_str()) };
+            auto ddlmOp{ packageManager.AddPackageAsync(ddlmUri, nullptr, DeploymentOptions::None) };
```
Scottj1s, line 40: curious why we don't condition the AddPackageAsync on FindPackages, as we do for the FWP above

test/ABForward/FunctionalTests.cpp
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        TEST_METHOD(AppNotificationsBuilder)
+        {
+            winrt::Microsoft::Windows::AppNotifications::Builder::AppNotificationBuilder();
+        }
+
+        TEST_METHOD(BackgroundTaskBuilder)
+        {
+            // BackgroundTaskBuilder is not available in 1.6
```
Scottj1s, line 84: what's the plan for rolling the test forward to the next release?

test/ABForward/arm64/Microsoft.WindowsAppRuntime.1.6.msix
Scottj1s, line 1: why are the arm64 msixes checked in?

test/ABForward/pch.h
```diff
@@ -0,0 +1,45 @@
...
+#include <winrt/Microsoft.Windows.ApplicationModel.Background.h>
+#include <winrt/Microsoft.Windows.BadgeNotifications.h>
+#include <winrt/Microsoft.Windows.Management.Deployment.h>
+#include <winrt/Microsoft.Windows.Security.AccessControl.h>
+#include <winrt/Microsoft.Windows.Storage.h>
+#include <winrt/Microsoft.Windows.System.h>
+#include <winrt/Microsoft.Windows.System.Power.h>
+
+#include <WindowsAppRuntime.Test.Bootstrap.h>
+namespace TP = ::Test::Packages;
```
Scottj1s, line 45: namespace aliases in headers are generally frowned upon

Dreynor87: This was copied from another test. I can move it for this PR.

test/ABForward/FunctionalTests.cpp
Scottj1s: that's a reasonable plan - too much trouble to attempt automation.  and really not necessary if we're snapshotting into each release branch.

test/ABForward/arm64/Microsoft.WindowsAppRuntime.1.6.msix
Scottj1s: my point is that these can be retrieved at build time, so shouldn't be checked in

test/ABForward/FunctionalTests.cpp
Dreynor87: AddPackage will quickly succeed if it is already present.

test/ABForward/arm64/Microsoft.WindowsAppRuntime.1.6.msix
Dreynor87: I will look into this route.

Dreynor87: No longer using these. Forgot to delelet. Will delete them now.

test/ABForward/FunctionalTests.cpp
```diff
@@ -0,0 +1,134 @@
...
+        BEGIN_TEST_CLASS(FunctionalTests)
+            TEST_CLASS_PROPERTY(L"ThreadingModel", L"MTA")
+            TEST_CLASS_PROPERTY(L"RunAs:Class", L"RestrictedUser")
+        END_TEST_CLASS()
+
+        TEST_CLASS_SETUP(ClassInit)
+        {
+            AddPackages();
+
+            const UINT32 c_Version_MajorMinor{ 0x00010006 };
```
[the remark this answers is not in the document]
Dreynor87: Using the preprocessor definitions to create this now...

test/ABForward/FunctionalTests.cpp
Scottj1s: in that case, why not always call it from the FWP as well?  just trying to understand the difference in logic

test/ABForward/packages.config
```diff
@@ -0,0 +1,7 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="Microsoft.Taef" version="10.95.240918004" targetFramework="native" />
+  <package id="Microsoft.Windows.CppWinRT" version="2.0.230706.1" targetFramework="native" />
+  <package id="Microsoft.Windows.ImplementationLibrary" version="1.0.240803.1" targetFramework="native" />
+  <package id="Microsoft.WindowsAppSDK" version="1.6.250408002" targetFramework="native" />
```
Scottj1s, line 6: probably want to add the transitive Microsoft.Web.WebView2 reference as well

test/ABForward/pch.h
Scottj1s: still planning to move this?

Dreynor87: I "moved" it, by adding to the cpp file, and forgot to remove here... finishing.

test/ABForward/packages.config
Dreynor87: I don't see any other projects referencing that one, and I don't seem to need it. Why would I add that here?

Scottj1s: Because it's a dependency of WinAppSDK as of 1.6.  Using tooling (nuget.exe, devenv.exe) to add WinAppSDK would pull in WebView2 as well, vs manual project edits.  Without it, you might (someday) get projection build errors due to missing metadata, whether tests call WebView2 APIs or not.

Scottj1s (approved):

test/ABForward/packages.config
Dreynor87: Added that here. I do need a re-approval from you before I can submit.

PR #5349, opened 2025-04-18
[1.7 servicing] Fix Bootstrapper only use Win11 DynamicDependency API if available AND WinAppSDK target >= 1.7 (#5321)

DrusTheAxe:
Bootstrapper uses the Win11 OS DynamicDependency API when available but should only do so for versions of WinAppSDK using the OS API, i.e. WinAppSDK >= 1.7

Also fixed a dependency build order issue

!cherry-pick fc5962d

This is a backport of [Fix Bootstrapper only use Win11 DynamicDependency API if available AND WinAppSDK target >= 1.7 #5321](https://github.com/microsoft/WindowsAppSDK/pull/5321) to release/1.7-stable

https://task.ms/57143286


dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -466,6 +479,8 @@ void FirstTimeInitialization(
         }
 
         // Track our initialized state
+        g_usingWin11Support = UsingWin11Support::No;
+        //
```
SaChinta-MSFT, line 483: nit: empty comment ?

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
```diff
@@ -394,6 +405,8 @@ void FirstTimeInitialization(
         }
 
         // Track our initialized state
+        g_usingWin11Support = UsingWin11Support::Yes;
+        //
```
SaChinta-MSFT, line 409: nit: empty comment ?

SaChinta-MSFT (approved):

dev/WindowsAppRuntime_BootstrapDLL/MddBootstrap.cpp
DrusTheAxe: Intentional whitespace

PR #5358, opened 2025-04-23
Supplements Missing Functionalities for FileOpenPicker.FileTypeFilter and FileSavePicker.FilterTypeChoices

DinahK-2SO:
## Description
This PR addresses the following missing functionalities for `FileOpenPicker.FileTypeFilter` and `FileSavePicker.FilterTypeChoices`:

### FileOpenPicker.FileTypeFilter
Test code (C++):
```C++
winrt::Microsoft::Windows::Storage::Pickers::FileOpenPicker picker{ AppWindow().Id() };
picker.FileTypeFilter().Append(L".jpg");
picker.FileTypeFilter().Append(L".png");
auto& file = co_await picker.PickSingleFileAsync();
```
1. When `FileTypeFilters` are defined, append a file type filter that includes all the already defined file types.
2. On launching the Open Dialog, auto select the unioned "All Files" category.

    **With this PR:**
    add-unioned-auto-select

    **Without this PR:**
    image

    
### FileSavePicker.FilterTypeChoices
Test code (C++):
```C++
winrt::Microsoft::Windows::Storage::Pickers::FileSavePicker picker{ AppWindow().Id() };
picker.FileTypeChoices().Insert(L"Documents", single_threaded_vector<hstring>({ L".txt", L".doc", L".docx", L".pdf"}));
picker.FileTypeChoices().Insert(L"Pictures", single_threaded_vector<hstring>({ L".jpg", L"jpeg", L".png", L".bmp"}));
auto& file = co_await picker.PickSaveFileAsync();
```

1. When `FileTypeChoices` are defined and a file is saved with an extension not listed in the chosen file type filters, automatically append the first extension from the choices (utilizing the built-in option of the COM file dialog - [FOS_STRICTFILETYPES](https://learn.microsoft.com/en-us/previous-versions/bb775856(v=vs.85))), unless the file already exists, in which why the FileSavePicker will allow saving a file with extension not listed in the chosen file type filters.
    **With this PR:**
    force-known-extensions
    **Without this PR:**
    force-known-extensions-beforefix

3. When multiple categories are defined in `FileTypeChoices`, the suggested file extension should change according to the selected category.
    **With this PR:**
    alter-extension-with-choices
    **Without this PR:**
    alter-extension-with-choices-beforefix


DinahK-2SO:
> @DinahK-2SO Wouldn't it be better if the All Files option was the first option? Currently, All Files is the last option.(I mean it should be placed before all types, not after them.)

Hi @ghost1372 , thanks for sharing your insights. We considered about this change. Given it might look strange to users who are already used to see "All Files" at the bottom of other filter types, we will not move it to the top for now.

However, showing user all the files by default will provide better user experience. (Thanks for bring up this topic!)

**Therefore, we're making the "All Files" option selected by default.** 
Here is a demo:
on-open-select-all

DrusTheAxe:
No Tests?

RECOMMEND: Add tests under `test\Interop\StoragePickers`.

Automated tests are desired as they prevent future regressions.

If this can't be automated for some reason Manual tests can be acceptable. In those cases a `README.md` with detailed instructions is necessary -- define any prerequisites, how to setup, how to execute, how to verify expected behavior, etc. Please very detailed in these cases and assume someone with no idea about your feature will follow the README and can easily verify the outcome.

dev/Interop/StoragePickers/PickerCommon.cpp
```diff
@@ -152,29 +152,54 @@ namespace PickerCommon {
     /// </returns>
     std::vector<COMDLG_FILTERSPEC> CaptureFilterSpec(std::vector<winrt::hstring>& buffer, winrt::Windows::Foundation::Collections::IVectorView<winrt::hstring> filters)
```
DrusTheAxe, line 153: `buffer` is a rather nondescript name.

SUGGESTION: `formattedFilters`? `filtersForCommonDialog`?

dev/Interop/StoragePickers/PickerCommon.cpp
```diff
@@ -152,29 +152,54 @@ namespace PickerCommon {
...
     {
-        std::vector<COMDLG_FILTERSPEC> result(filters.Size());
+        int resultSize = filters.Size() + 1;
         buffer.clear();
-        buffer.reserve(filters.Size() * (size_t)2);
+        buffer.reserve(resultSize * (size_t)2);
```
DrusTheAxe, line 157: Why the `(size_t)` cast? And if you do need a cast, you shouldn't use C style casts but rather C++ new style cast notation e.g. `static_cast<size_t>(2)`

SUGGESTION:

```
buffer.clear();`
buffer.reserve((filters.Size() + 1) * 2);
```

If the compiler gives you warning about mixing unsigned values from .Size() and signed values from the constants you can make the constants unsigned e.g. `buffer.reserve((filters.Size() + 1u) * 2u);`

dev/Interop/StoragePickers/PickerCommon.cpp
```diff
@@ -152,29 +152,54 @@ namespace PickerCommon {
...
+
+        std::wstring allFilesExtensionList;
         for (const auto& filter : filters)
         {
             auto ext = FormatExtensionWithWildcard(filter);
             buffer.push_back(filter);
             buffer.push_back(ext);
+            allFilesExtensionList += ext + L";";
         }
```
DrusTheAxe, line 166: You worry about growing `buffer` above with more capacity than you immediately need (size = add 1 and double that) but not here for allFilesExtensionList? This line will cause extra allocation amounting to

```
wstring temp ctor = ext.operator+(L";")
allFilesExtensionList += 
temp dtor
```

SUGGESTION: 
```
allFilesExtensionList.reserve(ext.length() + 1);
allFilesExtensionList += ext;
allFilesExtensionList += L";";
```

String+String creating a temporary's not perfectly efficient but not the worst thing in the world. Doing it in a loop gets more interesting as it magnifies the impact. And since you're already being somewhat meticulous elsewhere to efficiently minimize memory allocation (and deallocation) and copying data it seems worthy to do here too.

dev/Interop/StoragePickers/PickerCommon.cpp
```diff
@@ -152,29 +152,54 @@ namespace PickerCommon {
...
+            buffer[1] = L"*.*";
+            resultSize = 1;
         }
+        else
+        {
+            buffer.push_back(L"All Files");
+            buffer.push_back(allFilesExtensionList.c_str());
+        }
+
+        std::vector<COMDLG_FILTERSPEC> result(resultSize);
```
DrusTheAxe, line 192: NIT: `...result{ resultSize };`

DrusTheAxe (changes requested):

dev/Interop/StoragePickers/PickerCommon.cpp
```diff
@@ -152,29 +152,54 @@ namespace PickerCommon {
...
-            result.at(i) = { buffer.at(i * 2).c_str(), buffer.at(i * 2 + 1).c_str() };
+            allFilesExtensionList.pop_back(); // Remove the last semicolon
         }
 
-        if (result.size() == 0)
+        if (filters.Size() == 0)
         {
-            result.push_back({ L"All Files", L"*.*" });
+            // when filters not defined, set filter to All Files *.*
+            buffer.push_back(L"All Files");
```
codendone, line 176: `"All Files"` needs to come from a localized string resource.

dev/Interop/StoragePickers/PickerCommon.cpp
```diff
@@ -152,29 +152,54 @@ namespace PickerCommon {
...
+            buffer.push_back(L"*.*");
+        }
+        else if (filters.Size() == 1 && allFilesExtensionList == L"*")
```
codendone, line 179: If the app chose "*" as the extension, it seems wrong for us to override that the be "*.*" with the "All Files" name we chose.

dev/Interop/StoragePickers/PickerCommon.cpp
```diff
@@ -152,29 +152,54 @@ namespace PickerCommon {
...
+        {
+            // when there're only one filter "*", set filter to All Files *.* (override the values pushed above)
+            buffer[0] = L"All Files";
+            buffer[1] = L"*.*";
+            resultSize = 1;
         }
+        else
+        {
```
codendone, line 187: Is this new automatic "All Files" behavior something the UWP pickers do? Conceptually, the automatic behavior is nice, but I worry we might be forcing something on apps which don't want it or might have issues in some cases. Two potential problems I'm thinking of:
1. What if the app wants "All Images" rather than "All Files"? The app can add that themselves, but then we override the default with this additional option.
2. What if the app has its own categories, such as "Word Documents (.doc, .docx)", "Excel Documents (.xls, .xlsx)", (category:) "Office Documents (.doc, .docx, .xls, .xlsx)", where extensions are duplicated? Does "All Files" end up duplicating the extensions from the "Office Documents" category?

My feeling is that to handling (1) we might need an explicit API for the app to say if it wants an automatic "All Files", perhaps with an optional string if they want a different string used.

Handling (2) likely requires extra de-duplicate code.

codendone (changes requested):

dev/Interop/StoragePickers/PickerCommon.cpp
DinahK-2SO: Hi @codendone , yes, all the changes in description are consistent with the behavior of UWP pickers.

dev/Interop/StoragePickers/FileSavePicker.cpp
```diff
@@ -118,9 +118,17 @@ namespace winrt::Microsoft::Windows::Storage::Pickers::implementation
...
+            // the default extension has a ".", like ".txt". remove the "." (a.k.a "txt") then set it to the dialog
+            //      otherwise, the diaplayed file name would have two "."
+            check_hresult(dialog->SetDefaultExtension(defaultFileExtension.c_str() + 1));
+        }
+        else
+        {
+            check_hresult(dialog->SetDefaultExtension(L""));
         }
 
+        check_hresult(dialog->SetOptions(FOS_STRICTFILETYPES));
```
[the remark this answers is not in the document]
DinahK-2SO: Hi @codendone , The UWP pickers do have this restriction. In UWP pickers, if the "jpeg" is not listed in FileTypeChoices, user cannot save a file with ".jpeg" extension. 

For UWP pickers, this restriction can be bypassed when developers set one FileTypeChoices to be `"."` - meaning developers really allow the end user to save a file with any file extension. In the existing new pickers, this bypass can be conducted by `".*"`

dev/Interop/StoragePickers/PickerCommon.cpp
DinahK-2SO: Hi @codendone , as we tested locally, the UWP FileOpenPicker doesn't support a FileTypeFilter set to be "" or ".*" , which would throw error " _File extensions must begin with '.' and contain no wildcards.'_ "

The good new is that UWP pickers support "*" to be the FileOpenPicker filter - and it would create a filter "All Files (*)". This is the behavior we are trying to be consistent with.

dev/Interop/StoragePickers/FileSavePicker.cpp
```diff
@@ -118,9 +118,17 @@ namespace winrt::Microsoft::Windows::Storage::Pickers::implementation
...
         if (!PickerCommon::IsHStringNullOrEmpty(defaultFileExtension))
         {
-            check_hresult(dialog->SetDefaultExtension(defaultFileExtension.c_str()));
+            // the default extension has a ".", like ".txt". remove the "." (a.k.a "txt") then set it to the dialog
+            //      otherwise, the diaplayed file name would have two "."
+            check_hresult(dialog->SetDefaultExtension(defaultFileExtension.c_str() + 1));
+        }
+        else
+        {
+            check_hresult(dialog->SetDefaultExtension(L""));
```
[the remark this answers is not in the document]
DinahK-2SO: Yes, this is for the second fix for FileSavePicker. We discovered that only when DefaultExtension was set (even if set to an empty string), can the suggested file extension should change according to the selected category.

dev/Interop/StoragePickers/FileSavePicker.cpp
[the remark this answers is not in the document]
Hong-Xiang: @DinahK-2SO the workaround by "." may not be desired, I thought we have a separate task for the "no extension" scenarios. by pass with ".*" maybe not be ideal.

dev/Interop/StoragePickers/PickerCommon.cpp
Hong-Xiang: @DinahK-2SO  I think we need to take more care on handling "" here (and some other related code to handle parsing extensions), we've got this good review suggestion from @codendone 's previous review suggestions, we need to check and add proper handling. AFAIK, common file dialog will not handle this automatically for us

dev/Interop/StoragePickers/FileOpenPicker.cpp
```diff
@@ -90,6 +90,7 @@ namespace winrt::Microsoft::Windows::Storage::Pickers::implementation
         auto dialog = create_instance<IFileOpenDialog>(CLSID_FileOpenDialog, CONTEXT_ALL);
 
         parameters.ConfigureDialog(dialog);
+        check_hresult(dialog->SetFileTypeIndex(parameters.FileTypeFilterPara.size()));
```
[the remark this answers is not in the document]
DinahK-2SO: Yes, it is one-based! I am also surprised to find out it 😃 
https://learn.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-ifiledialog-setfiletypeindex
image

dev/Interop/StoragePickers/FileOpenPicker.cpp
```diff
@@ -142,7 +143,7 @@ namespace winrt::Microsoft::Windows::Storage::Pickers::implementation
         auto dialog = create_instance<IFileOpenDialog>(CLSID_FileOpenDialog, CONTEXT_ALL);
 
         parameters.ConfigureDialog(dialog);
-
+        check_hresult(dialog->SetFileTypeIndex(parameters.FileTypeFilterPara.size()));
```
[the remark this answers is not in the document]
DinahK-2SO: Yes, one-based, not zero-based. 😃 
https://learn.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-ifiledialog-setfiletypeindex

dev/Interop/StoragePickers/FileSavePicker.cpp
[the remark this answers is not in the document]
codendone: In a quick test, I see the UWP FileSavePicker force an extension if I specify a file which doesn't exist. But if I point it at a file which already exists with some other file extension, it lets me use it.

dev/Interop/StoragePickers/PickerCommon.cpp
codendone: Note: This was a pre-existing issue. Feel free to log a bug and tackle this in a later change.

(Aside: UWP appears to use "All files" with a lowercase 'f'. I don't know which is better.)

dev/Interop/StoragePickers/PickerCommon.cpp
codendone: I didn't escape my \*'s properly in my original comment.

@DinahK-2SO, I think what you're saying is if the app specifies only one filter as "\*" (just the wildcard), UWP would show "All files (\*)". But the change on line 182 will make the WinAppSDK FileOpenPicker show "All Files (\*.\*)", with the extra ".\*" added, correct? That's an intentional change in behavior?

dev/Interop/StoragePickers/PickerCommon.cpp
codendone: Ah, so the FileOpenPicker doesn't allow categories/grouping, so the duplication doesn't apply. And the UWP picker similarly doesn't allow customization of the "All files" text. These feel like odd choices for the UWP picker, but being consistent with that, at least as a starting point, makes sense. Resolving this comment.

dev/Interop/StoragePickers/PickerCommon.cpp
DinahK-2SO: Thanks @DrusTheAxe and @Hong-Xiang . Given this part will be moved in future, I am not changing the variable name in this PR.

dev/Interop/StoragePickers/FileSavePicker.cpp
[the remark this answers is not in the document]
DinahK-2SO: > In a quick test, I see the UWP FileSavePicker force an extension if I specify a file which doesn't exist. But if I point it at a file which already exists with some other file extension, it lets me use it.

Thanks @codendone for bringing up this point. From our local test, this is also the behavior of the `FOS_STRICTFILETYPES` option (although the behavior "saving an existing file" is not mentioned in public doc).

Please kindly find below a quick demo about this behavior:
1. There an existing file in test folder, the file is "HelloDevelopers.jpg"
2. We launch the FileSavePicker dialog with the `FOS_STRICTFILETYPES` option, and Document File Types (.txt, .pdf and so on)
3. if we try to save a file named as "HelloDevelopers.jpg", it prompts that "HelloDevelopers.jpg" already exists. And if we insist to save "HelloDevelopers.jpg", it overrides the existing file.
4. If we try to save a file named as "HelloDevelopers2.jpg", it creates "HelloDevelopers2.jpg.txt"

dev/Interop/StoragePickers/PickerCommon.cpp
DinahK-2SO: Thanks @codendone for your patient explanation, now I get the point!

Indeed there're no good reason to from "All Files (*)" to "All Files (*.*)", I will fix in code.

DinahK-2SO: > @DinahK-2SO I think we need to take more care on handling "" here (and some other related code to handle parsing extensions), we've got this good review suggestion from @codendone 's previous review suggestions, we need to check and add proper handling. AFAIK, common file dialog will not handle this automatically for us

Thanks @Hong-Xiang , the empty string "" will be formatted to "*" in above method `FormatExtensionWithWildcard`, therefore we don't need to handle "" here.

dev/Interop/StoragePickers/PickerCommon.cpp
DinahK-2SO: Thanks @DrusTheAxe for sharing this guidance! Updated in code.

dev/Interop/StoragePickers/PickerCommon.cpp
DinahK-2SO: Thanks @DrusTheAxe for sharing this improvement! I updated the code.

dev/Interop/StoragePickers/PickerCommon.cpp
DinahK-2SO: Thanks @codendone  to pin-point this issue. Given we don't yet have a stable localization design, I hide the "All files" string in the latest push. We're only displaying the union extension list as a temporary solution. Adding @yeelam-gordon and @Hong-Xiang here for your reference.

image

image

dev/Interop/StoragePickers/PickerCommon.cpp
DinahK-2SO: Thanks @DrusTheAxe . I updated in code.

DinahK-2SO:
> Automated tests are desired as they prevent future regressions.
> 
> If this can't be automated for some reason Manual tests can be acceptable. In those cases a `README.md` with detailed instructions is necessary -- define any prerequisites, how to setup, how to execute, how to verify expected behavior, etc. Please very detailed in these cases and assume someone with no idea about your feature will follow the README and can easily verify the outcome.

Hi @DrusTheAxe , Thank you for your comment and recommendations regarding the tests and appreciate your attention to detail and the importance of automated tests in preventing future regressions!

Currently, our testing framework is unable to capture the behavior after the File Dialog pops up, which makes it challenging to write test cases in time for 1.8-experimental2. Given this limitation, could we proceed with merging these fixes first and then work on the unit test automations afterward?

For the fixes targeted in this PR, we've uploaded some recordings of our manual tests in the PR description. We will supplement the test automation later to ensure comprehensive coverage.

Thanks and regards,

dev/Interop/StoragePickers/PickerCommon.cpp
codendone: I think a "Feedback wanted" discussion is not the right place for that, at least as currently phrased. Localizing strings shown in the UI is a requirement of released features, so there is no feedback needed for that, nor is it a standalone feature request. If this issue were about whether we want "All files" text shown for that item, that might be fine as a separate feature request if we believe it isn't a requirement.

codendone (approved):

dev/Interop/StoragePickers/PickerCommon.cpp
```diff
@@ -152,29 +152,57 @@ namespace PickerCommon {
...
         {
-            result.at(i) = { buffer.at(i * 2).c_str(), buffer.at(i * 2 + 1).c_str() };
+            allFilesExtensionList.pop_back(); // Remove the last semicolon
+        }
+
+        if (filters.Size() == 0)
+        {
+            // when filters not defined, set filter to All Files *.*
+            buffer.push_back(L"");
+            buffer.push_back(L"*");
```
yeelam-gordon, line 180: Should it be *.*? Since input like .txt will become *.txt as well

yeelam-gordon: Same for line 186

yeelam-gordon (approved):
Thanks for the continuous update on the feedback. Last comments I have that we discussed yestersday.

dev/Interop/StoragePickers/PickerCommon.cpp
DinahK-2SO: When displayed, the `"*"` will be automatically wrapped into a pair of brackets. That's why we will not add "." as key name. If the key name is ".", the filter will be displayed as `.(*)`, which looks strange.

DinahK-2SO:
Hello @DrusTheAxe , we added the document for manual test cases in the PR. Please take a look when you got a moment.

yeelam-gordon (approved):
Dinah talked with me offline on the comments. Looks good.

DinahK-2SO:
> ...then work on the unit test automations afterward?

Adding the unit tests here: #5434

DinahK-2SO:
Removing the completed PR from reviewer's task assignments.

PR #5363, opened 2025-04-25
User/haonanttt/test stage separation

haonanttt:
A microsoft employee must use /azp run to validate using the pipelines below.

WARNING:
Comments made by azure-pipelines bot maybe inaccurate.
Please see pipeline link to verify that the build is being ran.

For status checks on the main branch, please use TransportPackage-Foundation-PR
(https://microsoft.visualstudio.com/ProjectReunion/_build?definitionId=81063&_a=summary)
and run the build against your PR branch with the default parameters.

PR #5364, opened 2025-04-25
Pipeline Performance Optimization: Make Static Analysis Run Parallelly with Build

MuyuanMS:
Currently largest bottleneck building foundation is static analysis, which can be run independently but currently blocks dependent stage from running. After this change, now during build, two build stage will start in parallel - one without static analysis to allow dependent steps like Test or Pack to run, the other with static analysis

WARNING:


build/WindowsAppSDK-Foundation-Nightly.yml
```diff
@@ -113,6 +113,12 @@ extends:
 
     stages:
     - template: AzurePipelinesTemplates\WindowsAppSDK-Build-Stage.yml@self
+      parameters:
+        SignOutput: ${{ parameters.SignOutput }}
+        runStaticAnalysis : false
+    
+    - ${{ if eq(parameters.runStaticAnalysis, 'true') }}:
+      - template: AzurePipelinesTemplates\WindowsAppSDK-Build-Stage.yml@self
       parameters:
```
haonanttt, line 122: Incorrect indentation. The `parameters` section should add a `tab` and move a grid to right

build/WindowsAppSDK-Foundation-Official.yml
```diff
@@ -98,6 +98,12 @@ extends:
 
     stages:
     - template: AzurePipelinesTemplates\WindowsAppSDK-Build-Stage.yml@self
+      parameters:
+        SignOutput: ${{ parameters.SignOutput }}
+        runStaticAnalysis : false
+    
+    - ${{ if eq(parameters.runStaticAnalysis, 'true') }}:
+      - template: AzurePipelinesTemplates\WindowsAppSDK-Build-Stage.yml@self
       parameters:
```
haonanttt, line 107: Incorrect indentation. The `parameters` section should add a `tab` and move a grid to right

build/WindowsAppSDK-Foundation-PR.yml
```diff
@@ -92,7 +92,13 @@ extends:
     - template: AzurePipelinesTemplates\WindowsAppSDK-Build-Stage.yml@self
       parameters:
         SignOutput: false
-        runStaticAnalysis : ${{ parameters.runStaticAnalysis }}
+        runStaticAnalysis: false
+    
+    - ${{ if eq(parameters.runStaticAnalysis, 'true') }}:
+      - template: AzurePipelinesTemplates\WindowsAppSDK-Build-Stage.yml@self
+      parameters:
```
haonanttt, line 99: Incorrect indentation. The `parameters` section should add a `tab` and move a grid to right

build/WindowsAppSDK-Foundation-Nightly.yml
```diff
@@ -113,6 +113,12 @@ extends:
...
         SignOutput: ${{ parameters.SignOutput }}
         runStaticAnalysis : ${{ parameters.runStaticAnalysis }}
```
haonanttt, line 124: no error here. Only a question: how do you think if we directly use `runStaticAnalysis : true` here? the same to below files

build/WindowsAppSDK-Foundation-Nightly.yml
MuyuanMS: Good catch! fixed

build/WindowsAppSDK-Foundation-Nightly.yml
MuyuanMS: Right! maybe hardcode it to be true makes the code easier to read - I have changed it

haonanttt (approved):

yeelam-gordon (approved):
Thanks. Basically we build twice, and use the one without statistic analysis to save the time for running test earlier (As the result overall finished earlier as well)
One side effect, if the build for static analysis is failed, we may not want to mark this pipeline as failure given in the past the run for static analysis will continue on error, but it is a minor comments and I don't know how frequent it will happen.

Hong-Xiang (approved):

LegendaryBlair (approved):

alexlamtest (changes requested):
This is great work.
I think there is one thing we need to address, though.
The existing way to avoid spending too much time on PREfast is to simply run PREfast only for x64fre.
Please see this code in WindowsAppSDK-BuildMRT-Steps.yml and WindowsAppSDK-BuildFoundation-Steps.yml:

  - task: SDLNativeRules@3
...
    condition: and(succeeded(), eq(variables['buildPlatform'], 'x64'), eq(variables['buildConfiguration'], 'Release'))

Therefore, the Build_x86_StaticAnalysis stage is basically doing the same work as the Build_x86 stage, and that explains why they take similar time to finish. The same story applies to the Build_arm64_StaticAnalysis stage and the Build_arm64 stage.  In contrast, the Build_arm64_StaticAnalysis stage takes much longer to run than Build_arm64 due to running PREfast, as expected.

I see 2 basic options: 1) continue to only run PREfast on x64fre only, that means we skip the Build_x86_StaticAnalysis and Build_arm64_StaticAnalysis stages, which hopefully will allow the pipeline to finish even sooner, 2) adjust the code cited above to enable running PREfast for x86 and arm64 in the Build_x86_StaticAnalysis and Build_arm64_StaticAnalysis stages. This is expected to increase the "total new workload" of the pipeline and is thus likely to reduce the performance gain.
IMHO, #1 is a legitimate choice because it does not regress today's pipeline behavior. 
But if we prefer to data to support the decision, we can also first prototype #2 to confirm that running PREfast only for x64 only is "good enough" (we initially did verify our PREfast results are basically the same for x86, x64, and arm64, but that was a couple of years ago, things might have changed, probably worth another check). If the PREfast results are good, we should feel more comfortable going for option #1. Just my 2 cents.
Hence, I expect this PR to take more changes. If so, I suggest rebasing this PR on the branch with the merged PR [Performance Optimization for Foundation Build Pipeline: Break down build and test stage per platform by haonanttt · Pull Request #5363 · microsoft/WindowsAppSDK](https://github.com/microsoft/WindowsAppSDK/pull/5363) in place, to simplify this PR.
Thank you very much!

MuyuanMS:
> This is great work. I think there is one thing we need to address, though. The existing way to avoid spending too much time on PREfast is to simply run PREfast only for x64fre. Please see this code in WindowsAppSDK-BuildMRT-Steps.yml and WindowsAppSDK-BuildFoundation-Steps.yml:
> 
> * task: SDLNativeRules@3
>   ...
>   condition: and(succeeded(), eq(variables['buildPlatform'], 'x64'), eq(variables['buildConfiguration'], 'Release'))
> 
> Therefore, the Build_x86_StaticAnalysis stage is basically doing the same work as the Build_x86 stage, and that explains why they take similar time to finish. The same story applies to the Build_arm64_StaticAnalysis stage and the Build_arm64 stage. In contrast, the Build_arm64_StaticAnalysis stage takes much longer to run than Build_arm64 due to running PREfast, as expected.
> 
> I see 2 basic options: 1) continue to only run PREfast on x64fre only, that means we skip the Build_x86_StaticAnalysis and Build_arm64_StaticAnalysis stages, which hopefully will allow the pipeline to finish even sooner, 2) adjust the code cited above to enable running PREfast for x86 and arm64 in the Build_x86_StaticAnalysis and Build_arm64_StaticAnalysis stages. This is expected to increase the "total new workload" of the pipeline and is thus likely to reduce the performance gain. IMHO, #1 is a legitimate choice because it does not regress today's pipeline behavior. But if we prefer to data to support the decision, we can also first prototype #2 to confirm that running PREfast only for x64 only is "good enough" (we initially did verify our PREfast results are basically the same for x86, x64, and arm64, but that was a couple of years ago, things might have changed, probably worth another check). If the PREfast results are good, we should feel more comfortable going for option #1. Just my 2 cents. Hence, I expect this PR to take more changes. If so, I suggest rebasing this PR on the branch with the merged PR [Performance Optimization for Foundation Build Pipeline: Break down build and test stage per platform by haonanttt · Pull Request #5363 · microsoft/WindowsAppSDK](https://github.com/microsoft/WindowsAppSDK/pull/5363) in place, to simplify this PR. Thank you very much!



> This is great work. I think there is one thing we need to address, though. The existing way to avoid spending too much time on PREfast is to simply run PREfast only for x64fre. Please see this code in WindowsAppSDK-BuildMRT-Steps.yml and WindowsAppSDK-BuildFoundation-Steps.yml:
> 
> * task: SDLNativeRules@3
>   ...
>   condition: and(succeeded(), eq(variables['buildPlatform'], 'x64'), eq(variables['buildConfiguration'], 'Release'))
> 
> Therefore, the Build_x86_StaticAnalysis stage is basically doing the same work as the Build_x86 stage, and that explains why they take similar time to finish. The same story applies to the Build_arm64_StaticAnalysis stage and the Build_arm64 stage. In contrast, the Build_arm64_StaticAnalysis stage takes much longer to run than Build_arm64 due to running PREfast, as expected.
> 
> I see 2 basic options: 1) continue to only run PREfast on x64fre only, that means we skip the Build_x86_StaticAnalysis and Build_arm64_StaticAnalysis stages, which hopefully will allow the pipeline to finish even sooner, 2) adjust the code cited above to enable running PREfast for x86 and arm64 in the Build_x86_StaticAnalysis and Build_arm64_StaticAnalysis stages. This is expected to increase the "total new workload" of the pipeline and is thus likely to reduce the performance gain. IMHO, #1 is a legitimate choice because it does not regress today's pipeline behavior. But if we prefer to data to support the decision, we can also first prototype #2 to confirm that running PREfast only for x64 only is "good enough" (we initially did verify our PREfast results are basically the same for x86, x64, and arm64, but that was a couple of years ago, things might have changed, probably worth another check). If the PREfast results are good, we should feel more comfortable going for option #1. Just my 2 cents. Hence, I expect this PR to take more changes. If so, I suggest rebasing this PR on the branch with the merged PR [Performance Optimization for Foundation Build Pipeline: Break down build and test stage per platform by haonanttt · Pull Request #5363 · microsoft/WindowsAppSDK](https://github.com/microsoft/WindowsAppSDK/pull/5363) in place, to simplify this PR. Thank you very much!

Currently I tried to go with option 2 (Results: https://dev.azure.com/microsoft/ProjectReunion/_build/results?buildId=121725042&view=results), given that PREfast has been moved to a separate stage and they can run parallelly if resource allows. If we agree on running prefast for all three archs, I'll proceed to make the code changes in building MRT steps as well. Do you know how I can confirm whether PREfast results are the same for x86, x64, and arm64? (Just by build log, or do I actually have to manually make code changes to fail some PREfast checks and compare results?)

MuyuanMS:
> I see 2 basic options: 1) continue to only run PREfast on x64fre only, that means we skip the Build_x86_StaticAnalysis and Build_arm64_StaticAnalysis stages, which hopefully will allow the pipeline to finish even sooner, 2) adjust the code cited above to enable running PREfast for x86 and arm64 in the Build_x86_StaticAnalysis and Build_arm64_StaticAnalysis stages. 

@alexlamtest I have updated current PR with option 1. Below are some of the results while making the decision:

- [Original flow](https://dev.azure.com/microsoft/ProjectReunion/_build/results?buildId=121579347&view=results)
  - Built Time: 2h 26m 32s
  - Sum of All Jobs' Time: 582 min
- [Run PREfast for all 3 archs](https://dev.azure.com/microsoft/ProjectReunion/_build/results?buildId=121725042&view=results)
  - Built Time: 1h 40m 25s (-31.5%)
  - Sum of All Jobs' Time: 793 min (+36.3%)
- [Run PREfast for x64](https://dev.azure.com/microsoft/ProjectReunion/_build/results?buildId=122009703&view=results)
  - Built Time: 1h 27m 36s (-40%)
  - Sum of All Jobs' Time: 607 min (+4.3%)

yeelam-gordon:
> > I see 2 basic options: 1) continue to only run PREfast on x64fre only, that means we skip the Build_x86_StaticAnalysis and Build_arm64_StaticAnalysis stages, which hopefully will allow the pipeline to finish even sooner, 2) adjust the code cited above to enable running PREfast for x86 and arm64 in the Build_x86_StaticAnalysis and Build_arm64_StaticAnalysis stages.
> 
> @alexlamtest I have updated current PR with option 1. Below are some of the results while making the decision:
> 
> * [Original flow](https://dev.azure.com/microsoft/ProjectReunion/_build/results?buildId=121579347&view=results)
>   
>   * Built Time: 2h 26m 32s
>   * Sum of All Jobs' Time: 582 min
> * [Run PREfast for all 3 archs](https://dev.azure.com/microsoft/ProjectReunion/_build/results?buildId=121725042&view=results)
>   
>   * Built Time: 1h 40m 25s (-31.5%)
>   * Sum of All Jobs' Time: 793 min (+36.3%)
> * [Run PREfast for x64](https://dev.azure.com/microsoft/ProjectReunion/_build/results?buildId=122009703&view=results)
>   
>   * Built Time: 1h 27m 36s (-40%)
>   * Sum of All Jobs' Time: 607 min (+4.3%)

@alexlamtest, I will suggest let Muyuan to checkin the one only improve performance first, which the result is promising and without increase much build agent cost (i.e. 4.3% only)
And if we believe it is necessary to run statistic analysis against all the build architect, it should be an easy change later on.

alexlamtest (approved):

alexlamtest:
_>how I can confirm whether PREfast results are the same for x86, x64, and arm64?
You can use a OneBranch pipeline: https://microsoft.visualstudio.com/ProjectReunion/_build/results?buildId=121978283&view=logs&j=096205eb-4599-56da-affa-13246cc49eac. If you look at the end of each build job, there is a "Guardian: Post Analysis" task there. PREfast results can be found in there (non-OneBranch pipelines, such as the validation pipeline run for this PR, does not have that Guardian task). But note that for the pipeline run I cited above, even for the x64 build, there is no PREfast results shown in the Guardian task. That's simply because the cited pipeline run's parameter has runStaticAnalysis=false. If you set that parameter to true when you manually trigger the pipeline to run, or you use yml code changes to set the parameter to true, then you should be able to find PREfast results there, at least in the x64 cause. You would want to enable PREFast to run also for x86 and amd64. Then you should get PREfast results for all 3 architectures and you can confirm that their results are very similar. From my experience many months ago, the difference was 5% or less, and x64 usually found the most PREfast issues. Hope this helps!

Yes, I agree with Gordon that we can submit the changes for option #1. The changes LGTM, I already approved the PR. Thank you for this great work!

PR #5375, opened 2025-04-28
[Experimental] Decimal data type

DrusTheAxe:
Decimal data type (base-10 fixed-point math!)

This is implemented over the Win32 [Decimal](https://learn.microsoft.com/windows/win32/api/wtypes/ns-wtypes-decimal-r1) structure as the common data representation. C#'s `System.Decimal` uses the same representation in memory (different but equivalent implementation)

Recommendation is use your language's native decimal data type, if you've got one (C#, Python, ...). WinAppSDK supplements some languages to aid interop with the common data format ([Decimal](https://learn.microsoft.com/windows/win32/api/wtypes/ns-wtypes-decimal-r1)).

For some languages without a decimal data type WinAppSDK provides a native language data type for decimal operations and interop with the common data format. The C++ `decimal` class in `decimal.h` is an example of this.

For languages without a decimal data type WinAppSDK provides a runtimeclass `Microsoft.Windows.Foundation.Decimal`. This is only recommended when no other language native decimal type is available.

This is explained in more detail in `specs\decimal\decimal.md`. NOTE: This pre-draft is for early adopters to better understand the offering here. These APIs will stay EXPERIMENTAL until API Review is complete and whatever changes fall out of that are incorporated (could be minor, could be radical and majorly impactful).

NOTE: API Review will act on separate PR reviewing a future (review-ready) form of `specs\decimal\decimal.md`. This file is purely informational at this time and not to be taken today as definitive or API Review worthy.


dev/Decimal/decimalcppwinrt.h
```diff
@@ -0,0 +1,114 @@
...
+
+//----------------------------------------------------------------------
+// Conversion functions
+
+namespace winrt::Microsoft::Windows::Foundation
+{
+/// Return value as DECIMAL.
+inline DECIMAL to_DECIMAL(winrt::Microsoft::Windows::Foundation::DecimalValue const& value)
+{
+    return *reinterpret_cast<DECIMAL const*>(&value);
```
sylveon, line 62: Recommend avoiding UB here and using either std::bit_cast or memcpy.

dev/Projections/CS/Microsoft.Windows.Foundation/DecimalExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+            // meaning negative.
+            //   [0] = (int)d.Low;
+            //   [1] = (int)d.Mid;
+            //   [2] = (int)d.High;
+            //   [3] = d._flags;
+            // where d._flags is 2 bytes reserved, 1 byte sign, 1 byte scale
+            //
+            // @see https://github.com/dotnet/runtime/blob/1d1bf92fcf43aa6981804dc53c5174445069c9e4/src/libraries/System.Private.CoreLib/src/System/Decimal.cs#L581C13-L581C80
+
+            int[] bits = System.Decimal.GetBits(d);
```
Sergio0694, line 39: This is very inefficient, as it allocates an array every time. A better way would be to stack-allocate a buffer:
```suggestion
Span<int> bits = stackalloc int[4];
decimal.GetBits(d, bits);
```
The rest can stay the same.

dev/Projections/CS/Microsoft.Windows.Foundation/DecimalExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+using System.Runtime.CompilerServices;
+
+using Microsoft.Windows.Foundation;
+
+namespace Microsoft.Windows.Foundation
+{
+    public static class DecimalExtensions
+    {
+        /// Return a WinRT Decimal object.
+        public static Microsoft.Windows.Foundation.Decimal ToDecimal(this System.Decimal d)
```
Sergio0694, line 14: Should use the keyword in source, not the fully qualified type name, as it's a primitive in C# (same below)
```suggestion
        public static Microsoft.Windows.Foundation.Decimal ToDecimal(this decimal d)
```

dev/Decimal/decimalcppwinrt.h
DrusTheAxe: UB = ?

sylveon: Undefined behavior

sylveon:
One thing I noticed is that Decimal is a boxed version of DecimalValue. This seems counter to the current practice where we have, for example Color, ColorHelper for static methods, and IReference<Color> for boxing. Why the divergence here?

DrusTheAxe:
> One thing I noticed is that `Decimal` is a boxed version of `DecimalValue`. This seems counter to the current practice where we have, for example `Color`, `ColorHelper` for static methods, and `IReference<Color>` for boxing. Why the divergence here?

Unclear if/how that works given it must answer "How is `DecimalValue` handled by `PropertyValue`?"

https://learn.microsoft.com/en-us/uwp/api/windows.foundation.ireference-1

>IReference<T> inherits [IPropertyValue](https://learn.microsoft.com/en-us/uwp/api/windows.foundation.ipropertyvalue?view=winrt-26100). Types that implement IReference<T> also implement the interface members of [IPropertyValue](https://learn.microsoft.com//api/windows.foundation.ipropertyvalue).

Perhaps `Color` / `ColorHelper` / `IReference<Color>` is the better answer for `DecimalValue` / `Decimal` handling. The upside to experimental is one can experiment :-) You're the 2nd person to mention `ColorHelper` today so color &lt;g&gt; me intrigued. Need to do some more reading and perhaps throw together a `DecimalHelper` and related tests to answer some questions.

Can you think of _another_ reference besides `Color`/`ColorHelper` using this pattern?

sylveon:
> Can you think of another reference besides Color/ColorHelper using this pattern?

WF.GuidHelper, MUX.DurationHelper, MUX.CornerRadiusHelper, MUX.GridLengthHelper, MUX.PointHelper, MUX.RectHelper, MUX.SizeHelper, MUX.ThicknessHelper MUX.Media.MatrixHelper, etc.

As far as `IReference<T>` goes, the typical way to get one in C++ is to use `winrt::box_value`. I suppose a WASDK header would ship with the overload needed. @Sergio0694 would know best for cswinrt.

Sergio0694:
For CsWinRT, we implement `IReference<T>` and `IPropertyValue` for every WinRT struct or delegate that exists. If you have any projected WinRT type (struct, enum, or delegate), or any custom mapped type (struct, enum, or delegate), whenever you marshal a boxed version of it we also add `IPropertyValue` and `IReference<T>` for it in the CCW vtable. The only difference compared to primitive types that `IReferenceValue` supports, is that it will report `OtherType` when you ask it "what are you?".

sylveon:
Yeah, afaik cppwinrt has a basic implementation for other types, my main line of thought was the IPropertyValue member methods. If that's not necessary we can easily get away with using the regular implementation in cppwinrt instead of shipping a header.

Sergio0694:
To clarify I was agreeing with you that it'd be doable to do the same in CsWinRT if we just went with `DecimalValue` (and boxed), and `DecimalHelpers`. That was also my suggestion anyway, as it would avoid that "two boxed versions" issue you mentioned 🙂

sylveon:
If we go that way, I would rename it Decimal, but yeah :)

test/Decimal/CPP/DecimalTests.cpp
```diff
@@ -0,0 +1,1563 @@
...
+            Microsoft::Windows::Foundation::decimal object2;
+            object2 = data;
+            const auto to2{ object.to_string() };
+            VERIFY_ARE_EQUAL(0, wcscmp(data, to.c_str()), std::format(L"'{}' == '{}'", data, to.c_str()).c_str());
+        }
+
+        TEST_METHOD(ctor_to_assign_pcwstr_lcid)
+        {
+            PCWSTR data{ L"-12.345" };
+            const Microsoft::Windows::Foundation::decimal object(data, GetSystemDefaultLCID());
```
codendone, line 300: Using `GetSystemDefaultLCID()` or `LOCALE_SYSTEM_DEFAULT` (or other non-specific values) could cause issues for anyone running the test on a machine with different number formats (such as '.' as the grouping symbol [for thousands/millions/etc.]).

dev/Decimal/decimal.h
```diff
@@ -0,0 +1,703 @@
...
+
+    decimal& operator=(unsigned long value)
+    {
+        THROW_IF_FAILED(::VarDecFromUI4(value, &m_decimal));
+        return *this;
+    }
+
+    decimal& operator=(PCWSTR value)
+    {
+        THROW_IF_FAILED(::VarDecFromStr(value, LOCALE_INVARIANT, 0, &m_decimal));
```
codendone, line 300: Is LOCALE_INVARIANT always the correct option here? Might be a topic for eventual API review.

codendone (approved):

test/Decimal/CPP/DecimalTests.cpp
DrusTheAxe: Hmmm. Good point. The pipeline had no difficulty despite the myriad of environs we run on so yay but I've made a note to revise the tests in the next PR

dev/Decimal/decimal.h
DrusTheAxe: It's the most inncouous neutral option but yes, the whole localization string conversion thing is a topic for API Review as noted in `specs/decimal/decimal.md` section 6.1. API Review, item `1. String conversions - How to handle` :-)

test/Decimal/DecimalCalcuator/CS/DecimalCalculator_CS.csproj
```diff
@@ -0,0 +1,36 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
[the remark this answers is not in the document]
DrusTheAxe: Doh! Thanks. Will fix in next PR

PR #5413, opened 2025-05-11
Decimal spec

DrusTheAxe:
Spec for decimal support

### DIFF
--- a/specs/decimal/decimal.md
+++ b/specs/decimal/decimal.md
@@ -1,0 +1,467 @@
+# 1. Decimal
+
+THIS FEATURE IS CURRENTLY EXPERIMENTAL.
+
+This feature provides a common decimal representation and aids for use within and across programming languages.
+
+Some languages have a native decimal data type, e.g. [Decimal struct in C#](https://learn.microsoft.com/dotnet/api/system.decimal)
+and [decimal in Python](https://docs.python.org/3/library/decimal.html). These are recommended for
+use in their languages. Windows App SDK provides supplemental aids for interoperability with the
+common data type.
+
+Other languages have no native decimal data type e.g. C++, Rust and Javascript. Windows App SDK
+provides language native support for some languages including interoperability with the common data
+type:
+
+* C++ -- `Microsoft::Windows::Foundation::decimal`
+
+Windows App SDK provides a decimal WinRT runtimeclass `Microsoft.Windows.Foundation.Decimal` as a
+fallback for languages without a native decimal data type (whether provided by the language or
+Windows App SDK). This is fully functional but has performance overhead relative to language native
+data types which are commonly stack allocated and offer better language integration. The WinRT type
+is recommended when your language has no native type or other comparable solution.
+
+# 2. Common Decimal Representation
+
+The Win32 [DECIMAL](https://learn.microsoft.com/windows/win32/api/wtypes/ns-wtypes-decimal-r1)
+structure is the common decimal format, stored as a 96-bit (12-byte) unsigned integer scaled by a
+variable power of 10. The power of 10 scaling factor specifies the number of digits to the right of
+the decimal point, and ranges from 0 to 28. This is expressed in WinRT as the following structure:
+
+```c# (but really midl3)
+    struct DecimalValue
+    {
+        UInt16 Reserved;
+        Byte Scale;         // Allowed values: 0-28
+        Byte Sign;          // Allowed values: 0x80 = negative, 0x00 = zero or positive
+        UInt32 Hi32;
+        UInt64 Lo64;
+    }
+```
+
+This matches the memory layout and use of the Win32 [DECIMAL](https://learn.microsoft.com/windows/win32/api/wtypes/ns-wtypes-decimal-r1)
+structure and the [C# Decimal struct](https://learn.microsoft.com/dotnet/api/system.decimal). This
+data structure may be re-expressed in each language using language-specific constructs e.g. WinRT
+defines the `DecimalValue` structure as the Win32 `DECIMAL` definition syntax is incompatible with
+WinRT IDL.
+
+# 3. WinRT API
+
+Windows App SDK provides a `Decimal` WinRT runtimeclass in addition to the `DecimalValue` structure.
+
+```c# (but really MIDL3)
+namespace Microsoft.Windows.Foundation
+{
+    [contractversion(1)]
+    apicontract DecimalContract{};
+
+    /// WinRT representation of the Win32 DECIMAL structure.
+    /// @note This is the identical memory layout and encoding of the Win32 DECIMAL structure.
+    ///       The latter's definition is valid for COM but not WinRT making this equivalent structure necessary.
+    /// @see https://learn.microsoft.com/windows/win32/api/wtypes/ns-wtypes-decimal-r1
+    [feature(Feature_Decimal)]
+    [contract(DecimalContract, 1)]
+    struct DecimalValue
+    {
+        UInt16 Reserved;
+        UInt8 Scale;        // Allowed values: 0-28
+        UInt8 Sign;         // Allowed values: 0x80 = negative, 0x00 = zero or positive
+        UInt32 Hi32;
+        UInt64 Lo64;
+    };
+
+    [feature(Feature_Decimal)]
+    [contract(DecimalContract, 1)]
+    runtimeclass DecimalHelper
+    {
+        static DecimalValue FromBoolean(Boolean value);
+        static DecimalValue FromInt16(Int16 value);
+        static DecimalValue FromInt32(Int32 value);
+        static DecimalValue FromInt64(Int64 value);
+        static DecimalValue FromUInt8(UInt8 value);
+        static DecimalValue FromUInt16(UInt16 value);
+        static DecimalValue FromUInt32(UInt32 value);
+        static DecimalValue FromUInt64(UInt64 value);
+        static DecimalValue FromSingle(Single value);
+        static DecimalValue FromDouble(Double value);
+        static DecimalValue FromString(String value);                          // LCID=LOCALE_INVARIANT
+        static DecimalValue FromStringWithSystemDefaultLocale(String value);   // LCID=GetSystemDefaultLCID()
+        static DecimalValue FromStringWithUserDefaultLocale(String value);     // LCID=GetUserDefaultLCID()
+        static DecimalValue FromStringWithThreadLocale(String value);          // LCID=GetThreadLocale()
+
+        static Boolean ToBoolean(DecimalValue value);
+        static Int16 ToInt16(DecimalValue value);
+        static Int32 ToInt32(DecimalValue value);
+        static Int64 ToInt64(DecimalValue value);
+        static UInt8 ToUInt8(DecimalValue value);
+        static UInt16 ToUInt16(DecimalValue value);
+        static UInt32 ToUInt32(DecimalValue value);
+        static UInt64 ToUInt64(DecimalValue value);
+        static Single ToSingle(DecimalValue value);
+        static Double ToDouble(DecimalValue value);
+        static String ToString(DecimalValue value);                         // LCID=LOCALE_INVARIANT
+        static String ToStringWithSystemDefaultLocale(DecimalValue value);  // LCID=GetSystemDefaultLCID()
+        static String ToStringWithUserDefaultLocale(DecimalValue value);    // LCID=GetUserDefaultLCID()
+        static String ToStringWithThreadLocale(DecimalValue value);         // LCID=GetThreadLocale()
+
+        /// Return true if (left == right).
+        static Boolean Equals(DecimalValue left, DecimalValue right);
+
+        /// Compare decimal values.
+        /// @return 0 if left and right are equal, <0 if left is less than right or >0 if left is greater than right.
+        static Int32 Compare(DecimalValue left, DecimalValue right);
+
+        /// Return true if the value is valid.
+        static Boolean IsValid(DecimalValue value);
+
+        /// Return the scaling factor of the value (the number of decimal digits).
+        /// @return the scaling factor, ranging from 0 to max_scale().
+        static UInt32 Scale(DecimalValue value);
+
+        /// Return the sign of the value.
+        /// @return 0 if value is zero, <0 if value is less than zero or >0 if value is greater than zero.
+        static Int32 Sign(DecimalValue value);
+
+        /// Return the maximum scaling factor
+        static UInt32 MaxScale();
+
+        /// Return the maximum value (79,228,162,514,264,337,593,543,950,335).
+        static DecimalValue MaxValue();
+
+        /// Return the minimum value (-79,228,162,514,264,337,593,543,950,335).
+        static DecimalValue MinValue();
+
+        /// Return a decimal whose value is (-value).
+        static DecimalValue Negate(DecimalValue value);
+
+        /// Return the absolute value.
+        static DecimalValue Abs(DecimalValue value);
+
+        /// Return the integral digits; any fractional digits are discarded.
+        static DecimalValue Truncate(DecimalValue value);
+
+        /// Return the integral digits rounded down to -infinity; any fractional digits are discarded.
+        static DecimalValue Floor(DecimalValue value);
+
+        /// Return the integral digits rounded up to +infinity; any fractional digits are discarded.
+        static DecimalValue Ceiling(DecimalValue value);
+
+        /// Return the value rounded to the specific number of decimal places.
+        static DecimalValue Round(DecimalValue value, Int32 decimalPlaces);
+
+        /// Return value clamped to the inclusive range of min and max.
+        /// @return value if min <= value <= max, or min if value < min, or max if max < value.
+        static DecimalValue Clamp(DecimalValue value, DecimalValue min, DecimalValue max);
+
+        /// Returns a DecimalValue whose value is (left + right).
+        static DecimalValue Add(DecimalValue left, DecimalValue right);
+
+        /// Returns a DecimalValue whose value is (left - right).
+        static DecimalValue Subtract(DecimalValue left, DecimalValue right);
+
+        /// Returns a DecimalValue whose value is (left * right).
+        static DecimalValue Multiply(DecimalValue left, DecimalValue right);
+
+        /// Returns a DecimalValue whose value is (left / right).
+        static DecimalValue Divide(DecimalValue left, DecimalValue right);
+
+        /// Returns a DecimalValue whose value is (left % right).
+        static DecimalValue Modulo(DecimalValue left, DecimalValue right);
+    }
+}
+```
+
+# 4. C++ API
+
+Windows App SDK provides a native language decimal data type for C++ as the
+`Microsoft::Windows::Foundation::decimal` class in `decimal.h`. This class has the following features:
+
+* Constructor and assignment (operator=) overloads to define a decimal object from various data types
+* to_xxx() methods converting a decimal object's value to various data types
+* Relational operations: `compare()` `==` `!=` `<` `<=` `>` `>=`
+* Unary operations: + - ++ --
+* Binary operations: + += - -= * *= / /= % %=
+* Properties: `sign()` `scale()`
+* Mathematical operations: `abs()` `truncate()` `floor()` `ceil()` `round()` `clamp()`
+* Constants: `max_scale()` `max_value()` `min_value()`
+
+Errors are expressed via thrown exceptions e.g. `decimal{1} / decimal{0}` will throw a divide-by-zero exception
+
+## 4.1. decimal.h
+
+```c++
+#if !defined(__WindowsAppSDK_Microsoft_Windows_Foundation_decimal_)
+#define __WindowsAppSDK_Microsoft_Windows_Foundation_decimal_
+
+#include <oleauto.h>
+
+#if !defined(_VC_NODEFAULTLIB)
+#pragma comment(linker, "/defaultlib:oleaut32.lib")
+#endif
+
+inline int compare(const DECIMAL& left, const DECIMAL& right);
+
+inline bool operator==(const DECIMAL& left, const DECIMAL& right);
+
+inline bool operator!=(const DECIMAL& left, const DECIMAL& right);
+
+inline bool operator< (const DECIMAL& left, const DECIMAL& right);
+
+inline bool operator<=(const DECIMAL& left, const DECIMAL& right);
+
+inline bool operator> (const DECIMAL& left, const DECIMAL& right);
+
+inline bool operator>=(const DECIMAL& left, const DECIMAL& right);
+
+namespace Microsoft::Windows::Foundation
+{
+class decimal
+{
+public:
+    decimal() = default;
+    ~decimal() = default;
+
+    decimal(const decimal& value);
+    decimal(decimal&& value);
+    constexpr decimal(const DECIMAL& value);
+    decimal(bool value)l
+    decimal(char value);
+    decimal(std::int16_t value);
+    decimal(std::int32_t value);
+    decimal(std::int64_t value);
+    decimal(std::uint8_t value);
+    decimal(std::uint16_t value);
+    decimal(std::uint32_t value);
+    decimal(std::uint64_t value);
+    decimal(float value);
+    decimal(double value);
+    decimal(long value);
+    decimal(unsigned long value);
+    decimal(PCWSTR value);
+    decimal(PCWSTR value, const LCID locale);
+    decimal(const std::wstring& value);
+    decimal(const std::wstring& value, const LCID locale);
+#if defined(__WINSTRING_H_)
+    decimal(const HSTRING& value);
+    decimal(const HSTRING& value, const LCID locale);
+#endif // defined(__WINSTRING_H_)
+#if defined(WINRT_BASE_H)
+    decimal(const winrt::hstring& value);
+    decimal(const winrt::hstring& value, const LCID locale);
+#endif // defined(WINRT_BASE_H)
+
+    decimal& operator=(const decimal& value);
+    decimal& operator=(decimal&& value);
+    decimal& operator=(const DECIMAL& value);
+    decimal& operator=(bool value);
+    decimal& operator=(char value);
+    decimal& operator=(std::int16_t value);
+    decimal& operator=(std::int32_t value);
+    decimal& operator=(std::int64_t value);
+    decimal& operator=(std::uint8_t value);
+    decimal& operator=(std::uint16_t value);
+    decimal& operator=(std::uint32_t value);
+    decimal& operator=(std::uint64_t value);
+    decimal& operator=(float value);
+    decimal& operator=(double value);
+    decimal& operator=(long value);
+    decimal& operator=(unsigned long value);
+    decimal& operator=(PCWSTR value);
+    decimal& operator=(const std::wstring& value);
+#if defined(__WINSTRING_H_)
+    decimal& operator=(const HSTRING& value);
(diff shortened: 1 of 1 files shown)


ptorr-msft:
For comment in meeting invite around `ToString` -- consider leveraging ICU. It's been in the OS since RS5 or something and is the preferred cross-platform way of doing i18n. Not sure how well it maps to WinRT anymore; Ron or Axel would know (if they haven't forgotten LOL).

specs/decimal/decimal.md
```diff
@@ -68,125 +229,106 @@ namespace Microsoft.Windows.Foundation
...
+        static DecimalValue FromInt32(Int32 value);
+        static DecimalValue FromInt64(Int64 value);
+        static DecimalValue FromUInt8(UInt8 value);
+        static DecimalValue FromUInt16(UInt16 value);
+        static DecimalValue FromUInt32(UInt32 value);
+        static DecimalValue FromUInt64(UInt64 value);
+        static DecimalValue FromSingle(Single value);
+        static DecimalValue FromDouble(Double value);
+        static DecimalValue FromString(String value);                          // LCID=LOCALE_INVARIANT
+        static DecimalValue FromStringWithSystemDefaultLocale(String value);   // LCID=GetSystemDefaultLCID()
```
jonwis, line 245: Is there a version that the app can pass in a parameter?

jonwis, line 245: There was heated debate in API design in 2018ish about how to represent locales in API parameters.  Our public documentation - https://learn.microsoft.com/en-us/windows/win32/intl/locale-identifiers - suggests using "locale names" (like https://learn.microsoft.com/en-us/windows/win32/intl/locale-names) as the preferred currency.  The API group landed on "it's just a string with a structure" rather than introducing some custom type.  See also https://learn.microsoft.com/en-us/uwp/api/windows.globalization.language.languagetag?view=winrt-26100 for where that peeks out.

specs/decimal/decimal.md
```diff
@@ -68,125 +229,106 @@ namespace Microsoft.Windows.Foundation
...
+        static UInt64 ToUInt64(DecimalValue value);
+        static Single ToSingle(DecimalValue value);
+        static Double ToDouble(DecimalValue value);
+        static String ToString(DecimalValue value);                         // LCID=LOCALE_INVARIANT
+        static String ToStringWithSystemDefaultLocale(DecimalValue value);  // LCID=GetSystemDefaultLCID()
+        static String ToStringWithUserDefaultLocale(DecimalValue value);    // LCID=GetUserDefaultLCID()
+        static String ToStringWithThreadLocale(DecimalValue value);         // LCID=GetThreadLocale()
+
+        /// Return true if (left == right).
+        static Boolean Equals(DecimalValue left, DecimalValue right);
```
jonwis, line 265: Does this need to take an "epsilon" value for comparison?

specs/decimal/decimal.md
```diff
@@ -68,125 +229,106 @@ namespace Microsoft.Windows.Foundation
...
+
+        /// Compare decimal values.
+        /// @return 0 if left and right are equal, <0 if left is less than right or >0 if left is greater than right.
+        static Int32 Compare(DecimalValue left, DecimalValue right);
+
+        /// Return true if the value is valid.
+        static Boolean IsValid(DecimalValue value);
```
jonwis, line 272: What does "valid" mean here?

specs/decimal/decimal.md
```diff
@@ -68,125 +229,106 @@ namespace Microsoft.Windows.Foundation
...
 
         /// Return the scaling factor of the value (the number of decimal digits).
         /// @return the scaling factor, ranging from 0 to max_scale().
-        UInt32 Scale { get; };
+        static UInt32 Scale(DecimalValue value);
```
jonwis, line 276: We got a request from @jeprice and @jevansa to consider using signed integers - Int32 et al - rather than unsigned, as other languages (like C#) default to signed and unsigned is the outlier.

specs/decimal/decimal.md
```diff
@@ -197,66 +339,307 @@ Windows App SDK provides a native language decimal data type for C++ as the
...
+class decimal
+{
+public:
+    decimal() = default;
+    ~decimal() = default;
+
+    decimal(const decimal& value);
+    decimal(decimal&& value);
+    constexpr decimal(const DECIMAL& value);
+    decimal(bool value)l
```
jonwis, line 384: Needs semicolon

specs/decimal/decimal.md
```diff
@@ -45,7 +57,152 @@ data structure may be re-expressed in each language using language-specific cons
 defines the `DecimalValue` structure as the Win32 `DECIMAL` definition syntax is incompatible with
```
jonwis, line 57: Might link to the set of APIs this is related to - the `VarDec*` helpers, like https://learn.microsoft.com/en-us/windows/win32/api/oleauto/nf-oleauto-vardecfromstr ?

specs/decimal/decimal.md
DrusTheAxe: No, that's for floating point math. No need in base-10 math -- 1.33 is exactly that, and not 1.32, 1.34, or 1.333.

specs/decimal/decimal.md
DrusTheAxe: `DecimalValue` defines `Scale` as a byte/uint8 but only valid values are 0-28, and `Sign` is byte/uint8 but only valid values are 0 and 0x80.

Interestingly, at lest some of the `Var*Dec*()` functions detect this and error if invalid, and some don't. We'll be more rigorous in checking our inputs of course :-)

specs/decimal/decimal.md
DrusTheAxe: C# `Decimal.Scale` returns `Byte`

https://learn.microsoft.com/en-us/dotnet/api/system.decimal.scale?view=net-9.0

specs/decimal/decimal.md
```diff
@@ -197,66 +339,307 @@ Windows App SDK provides a native language decimal data type for C++ as the
...
+    decimal(std::uint64_t value);
+    decimal(float value);
+    decimal(double value);
+    decimal(long value);
+    decimal(unsigned long value);
+    decimal(PCWSTR value);
+    decimal(PCWSTR value, const LCID locale);
+    decimal(const std::wstring& value);
+    decimal(const std::wstring& value, const LCID locale);
+#if defined(__WINSTRING_H_)
```
oldnewthing, line 401: Correct guard is `__hstring_h__` since it is `hstring.h` that defines `HSTRING`, not `winstring.h`.

This is technically an ODR violation if some components include hstring and others don't.

Can explicitly declare.

```cpp
struct HSTRING__;
typedef struct HSTRING__* HSTRING;
```

similarly for `winrt::hstring`.

specs/decimal/decimal.md
```diff
@@ -197,66 +339,307 @@ Windows App SDK provides a native language decimal data type for C++ as the
...
+    decimal truncate() const;
+
+    /// Return the integral digits rounded down to -infinity; any fractional digits are discarded.
+    decimal floor() const;
+
+    /// Return the integral digits rounded up to +infinity; any fractional digits are discarded.
+    decimal ceil() const;
+
+    /// Return this clamped to the inclusive range of min and max.
+    /// @return this if min <= this <= max, or min if this < min, or max if max < this.
```
oldnewthing, line 515: Three more cases not covered: What happens if `min > max`?

specs/decimal/decimal.md
```diff
@@ -45,7 +57,152 @@ data structure may be re-expressed in each language using language-specific cons
 defines the `DecimalValue` structure as the Win32 `DECIMAL` definition syntax is incompatible with
 WinRT IDL.
 
-# 3. WinRT API
+# 3. Examples
```
MikeHillberg, line 60: It would be helpful to have examples of an API that takes/returns a Decimal type. Like a CurrencyTextBox control that has a Decimal-typed Amount property.

specs/decimal/decimal.md
```diff
@@ -45,7 +57,152 @@ data structure may be re-expressed in each language using language-specific cons
...
+        var d = Decimal.Parse("1.23");
+        var e = Decimal.Parse("4567.089");
+        var f = new Decimal((long)-4);
+        var g = new Decimal((uint)1967);
+        var h = new Decimal(1001.0);
+
+        var x = Decimal.Round(((a - b) + (c % d)) * e, 5) / f;
+        var y = Math.Clamp(Decimal.Round(x++, 3), f, g);
+        var yfloor = Decimal.Floor(y);
+        var xceil = Decimal.Ceiling(x);
```
idigdoug, line 118: These temporary values are not present in the otherwise-aligned C++ example. Should probably make the examples line up as closely as possible, so either remove the temporaries or use them in both examples.

specs/decimal/decimal.md
```diff
@@ -45,7 +57,152 @@ data structure may be re-expressed in each language using language-specific cons
...
+    const Microsoft::Windows::Foundation::decimal e{ L"4567.089" };
+    const Microsoft::Windows::Foundation::decimal f{ -4ll };
+    const Microsoft::Windows::Foundation::decimal g{ 1967u };
+    const Microsoft::Windows::Foundation::decimal h{ 1001.0f };
+
+    auto x = (((a - b) + (c % d)) * e).round(5) / f;
+    auto y = x++.round(3).clamp(f, g);
+    auto z = --y.floor() * ++x.ceil() * -y.truncate();
+
+    WEX::Logging::Log::Comment(WEX::Common::String().Format(L"x = %ls", x.to_string().c_str()));
```
idigdoug, line 154: >W [](http://example.com/codeflow?start=4&length=1)

I don't think WEX would be a well-known concept. Maybe just stick with the printfs?

specs/decimal/decimal.md
```diff
@@ -45,7 +57,152 @@ data structure may be re-expressed in each language using language-specific cons
...
+    //   y = x++.round(3).clamp(f, g)
+    //   z = --y.floor() * ++x.ceil() * -y.truncate()
+
+    auto x = DH::Divide(DH::Round(DH::Multiply(DH::Add(DH::Subtract(a, b), (DH::Modulo(c, d))), e), 5), f);
+    auto one = DH::FromInt32(1);
+    auto y = DH::Clamp(DH::Round(x, 3), f, g);
+    x = DH::Add(x, one);
+    auto z = DH::Multiply(DH::Multiply(DH::Subtract(DH::Floor(y), one), DH::Add(DH::Ceiling(x), one)), DH::Negate(DH::Truncate(y)));
+
+    printf(L"x = %ls\n", DH::ToString(x).c_str());
```
idigdoug, line 198: >x [](http://example.com/codeflow?start=13&length=1)

Isn't this going to produce different results than above since we no longer have the side-effects of `++x` and `--y` in play?

specs/decimal/decimal.md
```diff
@@ -68,125 +229,106 @@ namespace Microsoft.Windows.Foundation
...
+        static DecimalValue FromInt16(Int16 value);
+        static DecimalValue FromInt32(Int32 value);
+        static DecimalValue FromInt64(Int64 value);
+        static DecimalValue FromUInt8(UInt8 value);
+        static DecimalValue FromUInt16(UInt16 value);
+        static DecimalValue FromUInt32(UInt32 value);
+        static DecimalValue FromUInt64(UInt64 value);
+        static DecimalValue FromSingle(Single value);
+        static DecimalValue FromDouble(Double value);
+        static DecimalValue FromString(String value);                          // LCID=LOCALE_INVARIANT
```
idigdoug, line 244: >S [](http://example.com/codeflow?start=32&length=1)

Lots of our APIs tend to use CurrentLocale for the simply-named method, and make the InvariantLocale the long-named method (e.g. C#). While I wish it were the other way around, I'm hesitant to go against this precedent.

specs/decimal/decimal.md
```diff
@@ -197,66 +339,307 @@ Windows App SDK provides a native language decimal data type for C++ as the
...
+
+namespace Microsoft::Windows::Foundation
+{
+class decimal
+{
+public:
+    decimal() = default;
+    ~decimal() = default;
+
+    decimal(const decimal& value);
```
idigdoug, line 381: Implicit conversions require care because sometimes they can be surprising or mask typos. Given `void Foo(DECIMAL n)`, are we ok with all of the following compiling without warning?

Foo(false);
Foo('A');
Foo(L"Hello");

For at least these types, we might consider making the constructors explicit.

specs/decimal/decimal.md
```diff
@@ -197,66 +339,307 @@ Windows App SDK provides a native language decimal data type for C++ as the
...
+    decimal(std::int64_t value);
+    decimal(std::uint8_t value);
+    decimal(std::uint16_t value);
+    decimal(std::uint32_t value);
+    decimal(std::uint64_t value);
+    decimal(float value);
+    decimal(double value);
+    decimal(long value);
+    decimal(unsigned long value);
+    decimal(PCWSTR value);
```
idigdoug, line 397: Conspicuous lack of PCSTR, which people are going to complain about.

specs/decimal/decimal.md
```diff
@@ -197,66 +339,307 @@ Windows App SDK provides a native language decimal data type for C++ as the
...
+#if defined(WINRT_BASE_H)
+    decimal(const winrt::hstring& value);
+    decimal(const winrt::hstring& value, const LCID locale);
+#endif // defined(WINRT_BASE_H)
+
+    decimal& operator=(const decimal& value);
+    decimal& operator=(decimal&& value);
+    decimal& operator=(const DECIMAL& value);
+    decimal& operator=(bool value);
+    decimal& operator=(char value);
```
idigdoug, line 414: Similar issues here -- do we really want the following to compile?

dec = 'A';
dec = false;
dec = L"Hello";

specs/decimal/decimal.md
```diff
@@ -197,66 +339,307 @@ Windows App SDK provides a native language decimal data type for C++ as the
...
+#if defined(__WINSTRING_H_)
+    decimal& operator=(const HSTRING& value);
+#endif // defined(__WINSTRING_H_)
+#if defined(WINRT_BASE_H)
+    decimal& operator=(const winrt::hstring& value);
+#endif // defined(WINRT_BASE_H)
+
+    const DECIMAL& to_decimal() const;
+    bool to_bool() const;
+    char to_char() const;
```
idigdoug, line 437: Are people really asking to convert a decimal to a bool or char? What does it mean to do so?

specs/decimal/decimal.md
```diff
@@ -197,66 +339,307 @@ Windows App SDK provides a native language decimal data type for C++ as the
...
+    std::int64_t to_int64() const;
+    std::uint8_t to_uint8() const;
+    std::uint16_t to_uint16() const;
+    std::uint32_t to_uint32() const;
+    std::uint64_t to_uint64() const;
+    float to_float() const;
+    double to_double() const;
+    long to_long() const;
+    unsigned long to_ulong() const;
+    std::wstring to_string() const;
```
idigdoug, line 449: Consider `to_wstring` to match the naming and be explicit. Also, it seems likely that somebody is going to ask for something that returns CHAR-based strings.

specs/decimal/decimal.md
jonwis: For parsing, https://learn.microsoft.com/en-us/dotnet/api/system.decimal.tostring?view=net-9.0&redirectedfrom=MSDN#System_Decimal_ToString_System_String_ FromString has overloads for (string) and (string, IFormatProvider). That follows our usual rule of adding overload parameters after the base parameter list.

For formatting, C# goes the other way https://learn.microsoft.com/en-us/dotnet/api/system.decimal.tostring?view=net-9.0 and says that there's a four-way overload, with IFormatProvider being first.

specs/decimal/decimal.md
jonwis: `Foo(MWF::decimal const&)` would compile with `Foo(L"Hello")`, but `Foo(DECIMAL)` should not - https://godbolt.org/z/axEb41T4e ... I agree that adding `explicit` to the nonobvious ctors is good practice, though.

specs/decimal/decimal.md
```diff
@@ -197,66 +339,307 @@ Windows App SDK provides a native language decimal data type for C++ as the
...
+    decimal(std::uint16_t value);
+    decimal(std::uint32_t value);
+    decimal(std::uint64_t value);
+    decimal(float value);
+    decimal(double value);
+    decimal(long value);
+    decimal(unsigned long value);
+    decimal(PCWSTR value);
+    decimal(PCWSTR value, const LCID locale);
+    decimal(const std::wstring& value);
```
jonwis, line 399: Use `std::wstring_view value` in case someone doesn't have a std::wstring handy.

Does require you to potentially copy to a null-terminated string to call VarDecFromString, though.

Consider detecting `wil::zwstring_view`'s presence?

specs/decimal/decimal.md
```diff
@@ -197,66 +339,307 @@ Windows App SDK provides a native language decimal data type for C++ as the
...
+C# provides rich support via its [Decimal struct](https://learn.microsoft.com/dotnet/api/system.decimal).
+Windows App SDK adds a small C# API in an assembly for interop with `DecimalValue`.
+
+```c#
+namespace Microsoft.Windows.Foundation
+{
+    public static class DecimalExtensions
+    {
+        /// Return a WinRT DecimalValue structure.
+        public static Microsoft.Windows.Foundation.DecimalValue ToDecimalValue(this decimal d);
```
jonwis, line 639: How about `ToWindowsDecimal` ? "Decimal" as a standalone noun still seems to mean "the .NET Decimal type"

jonwis: Maybe make this the other way around?

```c#
public static decimal ToDecimal(this Microsoft.Windows.Foundation.DecimalValue value);
```

So you say:

```c#
DecimalValue dv = /* ... */;
decimal d = dv.ToDecimal();
```

Instead of the strange:

```c#
DecimalValue dv = /* ... */;
decimal d;
d = d.FromDecimalValue(d, dv);
```

specs/decimal/decimal.md
```diff
@@ -45,7 +57,152 @@ data structure may be re-expressed in each language using language-specific cons
...
+    printf(L"x = %ls\n", DH::ToString(x).c_str());
+    printf(L"y = %ls\n", DH::ToString(y).c_str());
+    printf(L"z = %ls\n", DH::ToString(z).c_str());
+    return DH::ToInt32(z);
+}
+```
+
+# 7. WinRT API
 
 Windows App SDK provides a `Decimal` WinRT runtimeclass in addition to the `DecimalValue` structure.
```
sylveon, line 207: It looks like the Decimal runtime class was removed, but this text still needs to be updated to reflect that.

Also I suggest renaming DecimalValue to Decimal in this case, since the runtime class is gone.

specs/decimal/decimal.md
idigdoug: While technically it's an ODR violation, practically this is one of the "safer" violations. Omitting a method based on ifdef isn't too bad. Omitting an overload might be a bit more risky if it is in any way possible that code might successfully compile to use the overload when present and successfully compile to use a different overload when absent. That probably isn't possible here.

I would consider defining this as

`decimal(struct HSTRING__ const& value)`

specs/decimal/decimal.md
DrusTheAxe, line 397: RECOMMEND: No constructor or operator= taking strings. Always named methods because we have the 3-function-problem e.g.
* ctor()
* ctor(string localeName)
* string invariant ???

Always use named methods for to/from string

DrusTheAxe: CONSIDER: Add narrow-string support

specs/decimal/decimal.md
DrusTheAxe: CONSIDER: Add overloads for string_view

specs/decimal/decimal.md
DrusTheAxe: RECOMMEND: Change the symbol detection

CONSIDER: Reject overload if header not recognized e.g.
```
#if defined(__hstring_h_)
decimal(const HSTRING& value);
#else
decimal(const HSTRING& value) = delete;
#endif
```

(not that we're keeping the constructor, point is to explicitly block unsupported overloads use the `...=delete` trick)

specs/decimal/decimal.md
DrusTheAxe: CONSIDER: Dropping `bool` conversions

specs/decimal/decimal.md
DrusTheAxe: DISCUSSION: `min > max` throws an exception

CONSIDER: Review `std::clamp`

specs/decimal/decimal.md
DrusTheAxe: RECOMMEND: Change to `ToDecimal`

i.e. `returntype To<returntype>(sourcetype)`

specs/decimal/decimal.md
DrusTheAxe: RECOMMEND: Replace ctor(string) with explicit named methods, drop to/from(char) and consider dropping to/from(bool) per previous comment

specs/decimal/decimal.md
DrusTheAxe: RECOMMEND: Fix

specs/decimal/decimal.md
DrusTheAxe: RECOMMEND: Replace oper=(string) with explicit named methods, drop to/from(char) and consider dropping to/from(bool) per previous comment

specs/decimal/decimal.md
DrusTheAxe: RECOMMEND: +1 rename to `to_wstring()`

specs/decimal/decimal.md
DrusTheAxe: CONSIDER: Add doccomment `@throw...` for the methods that can throw exceptions (only a handful)

specs/decimal/decimal.md
```diff
@@ -197,66 +339,307 @@ Windows App SDK provides a native language decimal data type for C++ as the
...
+
+    decimal operator+(const decimal& value) const;
+    decimal operator-(const decimal& value) const;
+    decimal operator*(const decimal& value) const;
+    decimal operator/(const decimal& value) const;
+
+    /// Modulo operation using the Truncated method.
+    /// @note The % operator in C, C#, Rust and other languages use this method.
+    /// @note The result's sign will match the current value's sign.
+    /// @see https://en.wikipedia.org/wiki/Modulo
```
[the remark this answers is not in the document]
DrusTheAxe: C++ calls it the **modulo operator**. Wikipedia has the [plain(ish) explanation](https://en.wikipedia.org/wiki/Modulo):

In [computing](https://en.wikipedia.org/wiki/Computing) and [mathematics](https://en.wikipedia.org/wiki/Mathematics), the modulo operation returns the [remainder](https://en.wikipedia.org/wiki/Remainder) or signed remainder of a [division](https://en.wikipedia.org/wiki/Division_(mathematics)), after one number is divided by another, the latter being called the [modulus](https://en.wikipedia.org/wiki/Modular_arithmetic) of the operation.

Operator=modulo, Result=remainder.

>C# calls it "[remainder operator](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#12104-remainder-operator)"

I see your quote and raise you => https://learn.microsoft.com/en-us/dotnet/api/system.decimal?view=net-9.0

I love Wikipedia's explanation of the multiple common(!) ways to perform modulo arithmetic and how C etc use truncation vs Knuth made a compelling case for rounding.


Rust gets even weirder with 'Euclidean arithmetic' vs 'regular arithmetic'. I'm sure there's a formal definition but this is a reasonably clear one => https://www.reddit.com/r/rust/comments/u7jt83/what_is_the_difference_between_euclidean/ TL;DR `a % b` vs `a.rem_euclid(b)` if you do/not want potentially negative results.

Math is weird :P

specs/decimal/decimal.md
```diff
@@ -197,66 +339,307 @@ Windows App SDK provides a native language decimal data type for C++ as the
...
+
+private:
+    DECIMAL m_decimal{};
+};
+}
+
+#endif // !defined(__WindowsAppSDK_Microsoft_Windows_Foundation_decimal_)
+```
+
+## 8.2. decimalcppwinmrt.h
```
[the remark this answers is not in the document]
DrusTheAxe: RECOMMEND: Fix typo

specs/decimal/decimal.md
DrusTheAxe: RECOMMEND: Per other comment DON'T expose LCID. Replace ctor and oper= with the trio of named methods

* ToString()
* ToStringInvariant()
* ToStringWithLocale(string localeName)

(and the inverse From... equivalents)

DrusTheAxe: RECOMMEND: Per other comment DON'T expose LCID. Replace ctor and oper= with the trio of named methods

* ToString()
* ToStringInvariant()
* ToStringWithLocale(string localeName)
* 
(and the inverse From... equivalents)

Thanks for the link re local names, will be useful for the implementation

DrusTheAxe: DISCUSSION: `LOCALE_NAME_USER_DEFAULT`, `..._INVARIANT` and `...SYSTEM_DEFAULT` are defined in winnls.h

```
// String based NLS APIs
//

#define LOCALE_NAME_USER_DEFAULT            NULL
#define LOCALE_NAME_INVARIANT               L""
#define LOCALE_NAME_SYSTEM_DEFAULT          L"!x-sys-default-locale"
```

`NULL` vs `""` makes this problematic for WinRT HSTRINGs so we chose UserDefault for our default (e.g. `ToString()`) for consistency with C# (and other languages where 'current locale` / etc rules commonly apply) and an explicitly named `...Invariant` method for the invariant locale.

specs/decimal/decimal.md
oldnewthing: Conditionally `delete`ing a constructor is arguably an even worse ODR violation. If choosing between the two, I would prefer to conditionally exclude the constructor rather than conditionally delete it.

specs/decimal/decimal.md
```diff
@@ -332,7 +344,7 @@ Windows App SDK provides a native language decimal data type for C++ as the
 
 Errors are expressed via thrown exceptions e.g. `decimal{1} / decimal{0}` will throw a divide-by-zero exception
 
-## 4.1. decimal.h
+## 8.1. decimal.h
```
oldnewthing, line 373: Having both an HSTRING constructor and a PCWSTR constructor means that `decimal(nullptr)` is ambiguous. We should have an explicit constructor for nullptr to resolve the ambiguity.
```cpp
decimal(std::nullptr_t);
```

specs/decimal/decimal.md
```diff
@@ -45,6 +57,13 @@ data structure may be re-expressed in each language using language-specific cons
 defines the `DecimalValue` structure as the Win32 `DECIMAL` definition syntax is incompatible with
```
sylveon, line 57: I'd still suggest renaming `DecimalValue` to just `Decimal` now that the name is not used by the runtime class anymore.

MikeHillberg (approved):

specs/decimal/decimal.md
oldnewthing: (But if constructing from nullptr is invalid, then no need to bother.)

DrusTheAxe: Invalid/meaningless/not-supported. There's (now) no ctors (now) taking a pointer - only way to pass in a pointer is named methods e.g. from_string(PCWSTR) vs from_HSTRING(HSTRING)

PR #5414, opened 2025-05-11
Revised Decimal data type

DrusTheAxe:
Revised decimal support incorporating feedback
* Replaced WinRT `Decimal` runtimeclass with `DecimalHelper`
* Renamed `fix` to `truncate`
* Renamed `integer` to `floor`
* Added `ceil` / `Ceiling`
* Added `clamp` / `Clamp`
* Added `is_valid` / `IsValid`
* Renamed WinRT abbreviated to full names (`Sub` / `Mul` / `Div` / `Mod` to `Subtract` / `Multiply` / `Divide` / `Mod`)
* Added validation checks
* Fixed string->decimal conversion bug due to `VarDecFromStr` not iniitalizing `DECIMAL.wReserved`
* Expanded tests


jeffstall (approved):
Thanks for taking the feedback! This will be much better for C++ and C# users immediately, and there's a good fallback for languages that don't yet support decimal type natively. It's much more consistent with how we're handling numerics with Vector3 (C#) and float3 (C++).

jeffstall (approved):

PR #5446, opened 2025-05-23
Fix FolderPicker crash when selecting "This PC" or special paths

DinahK-2SO:
## Issue

When users attempt to select "This PC" or other special paths in the FolderPicker, the dialog closes unexpectedly and crashes instead of providing guidance.

**gif for the behavior before fix**
new-picker-pick-pc

## Expect
After end user have selected "This PC" in the folder picker dialog, the dialog shows a message to inform user this is not valid, instead of crashing

## Fix

* [`dev/Interop/StoragePickers/FolderPicker.cpp`](diffhunk://#diff-fafd9b82668f2ebe3d318ec048dab1bf54074c05dd6f56dc392b0d3e5c0294c1L91-R91): Updated the `SetOptions` method to include the `FOS_FORCEFILESYSTEM` flag in addition to `FOS_PICKFOLDERS`, ensuring that the folder picker dialog enforces the selection of filesystem objects.

**gif for the behavior after fix**
after-fix-pick-pc


yeelam-gordon (approved):
Looks Good

PR #5495, opened 2025-06-06
[Draft] Adding Localization To Foundation

Hong-Xiang:
A Draft PR to added localization related prototype code to run e2e pipelines for testing.

WARNING:


eng/Version.Dependencies.props
```diff
@@ -14,7 +14,10 @@
     <MicrosoftProjectReunionInteractiveExperiencesTransportPackagePackageVersion>$([System.Text.RegularExpressions.Regex]::Match($(VersionDetailsXml), 'Name="Microsoft.ProjectReunion.InteractiveExperiences.TransportPackage"\s+Version="(.*?)"').Groups[1].Value)</MicrosoftProjectReunionInteractiveExperiencesTransportPackagePackageVersion>
     <MicrosoftWinAppSDKEngCommonPackageVersion>$([System.Text.RegularExpressions.Regex]::Match($(VersionDetailsXml), 'Name="Microsoft.WinAppSDK.EngCommon"\s+Version="(.*?)"').Groups[1].Value)</MicrosoftWinAppSDKEngCommonPackageVersion>
     <MicrosoftWindowsAppSDKAppLicensingInternalTransportPackagePackageVersion>$([System.Text.RegularExpressions.Regex]::Match($(VersionDetailsXml), 'Name="Microsoft.WindowsAppSDK.AppLicensingInternal.TransportPackage"\s+Version="(.*?)"').Groups[1].Value)</MicrosoftWindowsAppSDKAppLicensingInternalTransportPackagePackageVersion>
+    <MicrosoftWindowsAppSDKBasePackageVersion>$([System.Text.RegularExpressions.Regex]::Match($(VersionDetailsXml), 'Name="Microsoft.WindowsAppSDK.Base"\s+Version="(.*?)"').Groups[1].Value)</MicrosoftWindowsAppSDKBasePackageVersion>
```
Hong-Xiang, line 17: While MSIXPackageVersion below is used in appruntime, the BasePackgeVersion and InteractiveExperiencePackageVersion seems auto generated by DevCheck.cmd -SyncDependencies by following notes in Version.Details.xml. Not sure if we should remove those

dev/Interop/StoragePickers/PickerLocalization.cpp
```diff
@@ -0,0 +1,20 @@
...
+#include <winrt\Microsoft.Windows.ApplicationModel.Resources.h>
+#include <iostream>
+
+namespace PickerCommon {
+    const winrt::hstring priPath = L"Microsoft.WindowsAppRuntime.pri";
+    const winrt::hstring resourceName = L"Microsoft.WindowsAppRuntime/StoragePickers/All Files";
+    void UpdateAllTextLocalization(PickerCommon::PickerParameters& parameter)
+    {
+        winrt::Microsoft::Windows::ApplicationModel::Resources::ResourceManager manager = winrt::Microsoft::Windows::ApplicationModel::Resources::ResourceManager(priPath);
+        auto text = manager.MainResourceMap().GetValue(resourceName).ValueAsString();
```
Hong-Xiang: [Question] Should we use try get value related API to allow fall back text or fail early to make it easier to find that pri file is not found (which indicates WinAppSDK runtime may not be properly packaged into consumer app)
Current implementation is using simple fail early way.

dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/src/Microsoft.Windows.ApplicationModel.Resources.vcxproj
```diff
@@ -21,7 +21,7 @@
     <WindowsTargetPlatformMinVersion>10.0.17134.0</WindowsTargetPlatformMinVersion>
     <!-- Need to add this otherwise the ADO build won't copy mrm.dll to output directory, and eventually the file will miss in ManagedTest.build.appxrecipe and make
          the tests fail to run in ADO pipeline -->
-    <BuildingInsideVisualStudio>true</BuildingInsideVisualStudio>
+    <!-- <BuildingInsideVisualStudio>true</BuildingInsideVisualStudio> -->
```
Hong-Xiang: [Note & Question] 
In current pipeline test it seems that no test is failing after this change.
If we don't comment this out, this AdHoc BuilidingInsideVisualStudio config will case pipeline errors,
by binlog it seems that adding this will prevent pipeline to build its dependencies,
thus mrm.lib will not build.
Question: is there concern on this modification?

eng/Version.Details.xml
```diff
@@ -47,6 +47,10 @@
       <Uri>https://dev.azure.com/microsoft/LiftedIXP/_git/DCPP</Uri>
       <Sha>b430accfb06428e6d5f4a4306f6acd07c84fc7d6</Sha>
     </Dependency>
+    <Dependency Name="Microsoft.Windows.SDK.BuildTools.MSIX" Version="1.8.0-main.20250602.0">
```
Hong-Xiang: [Question] Would maestro automatically start to pushing future version update to this package?
Although it might be safe to merge first and monitor maestro behavior.

dev/Interop/StoragePickers/PickerLocalization.cpp
```diff
@@ -0,0 +1,20 @@
...
+        parameter.AllFilesText = text;
+    }
```
yeelam-gordon, line 20: This code is highly coupling on all the localization logic with PickerParameters.

Two comments and see if it make sense:
a. We usually define localization "Key" as constant, and having helper method to load the localization resource at the place they should consume. 
e.g. the code PickerCommon.cpp, it try to add FilterData "AllFilesText", it should be like:
GetLocalizedResource(AllFilesTextKey)

b. What you wrote on how loading resources now are like loading resource from an file, instead of EMBEDDED resource as part of the Microsoft.WindowsAppRuntime?
Should we always load with our dll so that we don't need to worry it? And probably the resourceName may be simpler in case.

dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/src/Microsoft.Windows.ApplicationModel.Resources.vcxproj
yeelam-gordon: If there is no need, the above comment should be removed together.

dev/Interop/StoragePickers/FileOpenPicker.cpp
```diff
@@ -75,6 +76,7 @@ namespace winrt::Microsoft::Windows::Storage::Pickers::implementation
         auto logTelemetry{ StoragePickersTelemetry::FileOpenPickerPickSingleFile::Start(m_telemetryHelper) };
 
         PickerCommon::PickerParameters parameters{};
+        PickerCommon::UpdateAllTextLocalization(parameters);
 
         CaptureParameters(parameters);
```
yeelam-gordon, line 81: Please see my comment below. I think loading localized resource should be happening at CaptureParameters

yeelam-gordon (changes requested):

test/DynamicDependency/data/Microsoft.WindowsAppRuntime.Framework/appxmanifest.xml
```diff
@@ -155,6 +155,18 @@
         <ActivatableClass ActivatableClassId="Microsoft.Windows.Storage.Pickers.FolderPicker" ThreadingModel="both" />
       </InProcessServer>
     </Extension>
+
+    <Extension Category="windows.activatableClass.inProcessServer">
+      <InProcessServer>
```
DrusTheAxe: MRT defines 8 runtimeclasses. Why only list 4 here?

`dev\MRTCore\mrt\Microsoft.Windows.ApplicationModel.Resources\src\Microsoft.Windows.ApplicationModel.Resources.idl`

RECOMMEND: Add all MRT runtimeclasses. Missing: `ResourceNotFoundEventArgs`, `ResourceMap`, `ResourceContext`, `ApplicationLanguages`

test/DynamicDependency/data/Microsoft.WindowsAppRuntime.Framework/appxmanifest.xml
```diff
@@ -155,6 +155,18 @@
...
+        <Path>Microsoft.Windows.ApplicationModel.Resources.dll</Path>
+        <ActivatableClass ActivatableClassId="Microsoft.Windows.ApplicationModel.Resources.ResourceManager" ThreadingModel="both" />
+        <ActivatableClass ActivatableClassId="Microsoft.Windows.ApplicationModel.Resources.IResourceManagerFactory" ThreadingModel="both" />
```
DrusTheAxe: What is `IResourceManagerFactory`? I don't see it listed in any file under `dev\MRTCore`

eng/Version.Dependencies.props
```diff
@@ -14,7 +14,10 @@
...
+    <MicrosoftWindowsAppSDKInteractiveExperiencesPackageVersion>$([System.Text.RegularExpressions.Regex]::Match($(VersionDetailsXml), 'Name="Microsoft.WindowsAppSDK.InteractiveExperiences"\s+Version="(.*?)"').Groups[1].Value)</MicrosoftWindowsAppSDKInteractiveExperiencesPackageVersion>
```
DrusTheAxe, line 18: Is your branch stale? I see this line already exists in main

eng/Version.Dependencies.props
```diff
@@ -14,7 +14,10 @@
...
     <MicrosoftWindowsCsWinRTPackageVersion>$([System.Text.RegularExpressions.Regex]::Match($(VersionDetailsXml), 'Name="Microsoft.Windows.CsWinRT"\s+Version="(.*?)"').Groups[1].Value)</MicrosoftWindowsCsWinRTPackageVersion>
+    <MicrosoftWindowsSDKBuildToolsMSIXPackageVersion>$([System.Text.RegularExpressions.Regex]::Match($(VersionDetailsXml), 'Name="Microsoft.Windows.SDK.BuildTools.MSIX"\s+Version="(.*?)"').Groups[1].Value)</MicrosoftWindowsSDKBuildToolsMSIXPackageVersion>
```
DrusTheAxe: How'd this get here? I don't see it in Version.Dependencies.xml

This is a generated file. Never edit it directly -- edit Version.Dependencies.xml or Version.Details.xml and run `DevCheck -SyncDependencies` to propogate your updates (here and elsewhere)

See the huge comment at the top of `eng\Version.Dependendencies.xml` or `eng\Version.Details.xml` for more details

eng/Version.Dependencies.props
DrusTheAxe: This filer is entirely generated by DevCheck. Don't edit this file directly.

This is derived from `Microsoft.WindowsAppSDK.Base` in Version.Details.xml -- which present in main but not in Version.Dependencies.props. That means someone previously edited Version.Details.xml and DIDN'T run `DevCheck -SyncDependencies` to propogate the change. Thank you for correcting this previous gap.

I've got a change which adds this and other crosschecks, but hasn't been incorporated into the build yet. Once added it will make these oversights impossible

dev/WindowsAppRuntime_DLL/packages.config
```diff
@@ -5,4 +5,5 @@
   <package id="Microsoft.SourceLink.GitHub" version="1.1.1" targetFramework="native" developmentDependency="true" />
   <package id="Microsoft.Windows.CppWinRT" version="2.0.230706.1" targetFramework="native" />
   <package id="Microsoft.Windows.ImplementationLibrary" version="1.0.240803.1" targetFramework="native" />
-</packages>
+  <package id="Microsoft.Windows.SDK.BuildTools.MSIX" version="1.8.0-main.20250602.0" targetFramework="native" developmentDependency="true" />
```
DrusTheAxe: This isn't listed in eng\Version.Dependencies.xml. That's a problem -- all dependencies must be defined in Version.*.xml, and those files require approval to add new dependencies.

(Probably not a problem ***in this case***, but still)

dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/src/Microsoft.Windows.ApplicationModel.Resources.vcxproj
DrusTheAxe: There's subtle differences between building inside VS vs outside. Do you know why it was added in the first place?

@Scottj1s any comment?

DrusTheAxe: P.S. +1 if this line is obsolete so too the comment above it -- keep or delete both

dev/Interop/StoragePickers/PickerLocalization.cpp
```diff
@@ -0,0 +1,20 @@
...
+// Licensed under the MIT License.
+
+#include "pch.h"
+#include "PickerLocalization.h"
+#include <winrt\base.h>
+#include <winrt\Microsoft.Windows.ApplicationModel.Resources.h>
+#include <iostream>
+
+namespace PickerCommon {
+    const winrt::hstring priPath = L"Microsoft.WindowsAppRuntime.pri";
```
DrusTheAxe, line 15: WinAppSDK has multiple .pri files e.g. `Microsoft.UI.pri` and `Microsoft.UI.Xaml.Controls.pri`.

Given these resources are specifically for the Pickers this should be a picker-specific filename e.g. `Microsoft.Windows.Storage.Pickers.pri`

dev/Interop/StoragePickers/PickerLocalization.cpp
DrusTheAxe: Where does it look?

dev/Interop/StoragePickers/PickerLocalization.cpp
```diff
@@ -0,0 +1,20 @@
...
+    const winrt::hstring resourceName = L"Microsoft.WindowsAppRuntime/StoragePickers/All Files";
```
DrusTheAxe: Use a Picker-specific name e.g. `Microsoft.Windows.Storage.Pickers`

see other comment

test/DynamicDependency/data/Microsoft.WindowsAppRuntime.Framework/appxmanifest.xml
DrusTheAxe: P.S. Please sort the <ActivatableClass> lines alphabetically e.g. ...KnownResourceQualifierName shouldn't appear after ...ResourceCandidate

DrusTheAxe (dismissed):

dev/Interop/StoragePickers/PickerLocalization.cpp
Hong-Xiang: Hi @DrusTheAxe , thanks for your comments!

I've spent some time researching the proper naming conventions for .pri files. It seems that each .pri file typically corresponds to a .dll file—for example, we have both Microsoft.UI.dll and Microsoft.UI.pri, as well as Microsoft.UI.Xaml.Controls.dll and Microsoft.UI.Xaml.Controls.pri. So I’m assuming this pairing might be a convention.

For StoragePickers, there’s no Microsoft.Windows.Storage.Pickers.dll; instead, we only have Microsoft.WindowsAppRuntime.dll. While the name might be a bit confusing, Microsoft.WindowsAppRuntime.dll is actually just the foundation component of the released WinAppSDK. It includes multiple components under different namespaces—like CameraCaptureUI and Notification—but it’s not the WinAppSDK itself, just a collection of APIs implemented in one place.

Because of this, I feel that introducing a separate .pri file specifically for StoragePickers might cause additional confusion, since it wouldn’t follow the usual .dll-.pri pairing convention.

Would love to hear your thoughts on this—do you think this reasoning makes sense?

eng/Version.Dependencies.props
Hong-Xiang: Thanks for your comments! This result was generated by running the sync dependencies command, as you suggested, not manually modified.
The version originates from the Microsoft.Windows.SDK.BuildTools.MSIX dependency listed in Version.Details.xml. Since this package is provided through internal repositories, I’ve followed the aggregator repo’s usage pattern and placed it in Version.Details.xml instead of Version.Dependencies.xml.

eng/Version.Dependencies.props
Hong-Xiang: Currently in main they are listed in Version.Details.xml, but unfortunately it seems they are not generated in this props files https://github.com/microsoft/WindowsAppSDK/blob/1a71198a79e3ccdc6671b972177d893f7d7b7660/eng/Version.Dependencies.props#L10-L17

test/DynamicDependency/data/Microsoft.WindowsAppRuntime.Framework/appxmanifest.xml
Hong-Xiang: Hi @DrusTheAxe  Thanks for your suggestions! The XML changes you noticed were part of earlier, unsuccessful attempts to enable localization in unit tests. I’ve now removed those unused activation configurations entirely—apologies for the confusion, they should have been cleaned up earlier.

Currently, current unit test still doesn’t work with the newly added localization-related code. It requires not only WinRT activation changes but also additional setup to include MRT Core in the unit test project. For now, unfortunately localization functionality can only be verified through manual testing.
Given the recent metapackage refactor, I’d like to propose a future update to this unit test to use the actual component package instead of current mocked one. That should resolve the activation issues automatically by leveraging the activation definition file shipped with the real component package. Would you mind us to separate this into a separate task instead of in this localization PR?

test/DynamicDependency/data/Microsoft.WindowsAppRuntime.Framework/appxmanifest.xml
Hong-Xiang: We've removed all modifications from the test AppxManifest.xml.
Thanks for your thorough review and great catch!

dev/Interop/StoragePickers/PickerLocalization.cpp
Hong-Xiang: Thanks for your suggestions. You're absolutely right—this logic is tightly coupled with PickerParameters, and ideally, it should reside in CaptureParameters or PickerCommon.cpp.
However, due to current limitations in the unit test setup for the picker, we've separated this file to avoid impacting existing tests in PickerCommon.cpp with localization-related logic. The separation allows us to at least keep unit tests for PickerCommon logic working.

dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/src/Microsoft.Windows.ApplicationModel.Resources.vcxproj
Hong-Xiang: Thanks for your suggestions! I've removed both the configuration and the related comment.

eng/Version.Dependencies.props
Hong-Xiang: Thank you for your explanation! The change was generated by running the sync command, not manually edited. Your explanation clarifies the current situation well—it seems appropriate to keep this change and proceed with merging it into main to resolve the issue.

dev/WindowsAppRuntime_DLL/packages.config
Hong-Xiang: Microsoft.Windows.SDK.BuildTools.MSIX is listed in Version.Details.xml, it seems that if I don't list it here there would be build error.

dev/Interop/StoragePickers/FileOpenPicker.cpp
Hong-Xiang: Although we couldn’t fully integrate this into CaptureParameters due to unit test constraints, we’ve introduced abstraction to reduce coupling

dev/Interop/StoragePickers/PickerLocalization.cpp
Hong-Xiang: As noted in the comment above, we currently can't modify the name of the WindowsAppRuntime DLL or separate StoragePickers into its own DLL. Given these constraints, should we proceed with changing to this namespace

dev/Interop/StoragePickers/PickerLocalization.cpp
Hong-Xiang: By design, it reads data from Microsoft.WindowsAppRuntime.pri. Although this PRI file is expected to be deployed with WinAppSDK, this question is about should we consider adding special handling to fall back to predefined text in case the pri file is unexpectedly missing?

dev/Interop/StoragePickers/PickerLocalization.cpp
```diff
@@ -0,0 +1,22 @@
...
+// MRT Core dependencies in unit tests. Once unit tests are refactored to use component
+// packages instead of mocked packages, this can be moved into CaptureParameters method.
+
+#include "pch.h"
+#include "PickerLocalization.h"
+#include <winrt\base.h>
+#include <winrt\Microsoft.Windows.ApplicationModel.Resources.h>
+#include <iostream>
+
+namespace PickerCommon {
```
Scottj1s: given the association of namespace PickerCommon with PickerCommon.h,cpp, should the namespace here by PickerLocalization? seems a bit clearer

dev/Interop/StoragePickers/PickerLocalization.cpp
```diff
@@ -0,0 +1,22 @@
...
+    const winrt::hstring priPath = L"Microsoft.WindowsAppRuntime.pri";
+    winrt::hstring GetStoragePickersLocalizationText(winrt::hstring key)
+    {
+        winrt::Microsoft::Windows::ApplicationModel::Resources::ResourceManager manager = winrt::Microsoft::Windows::ApplicationModel::Resources::ResourceManager(priPath);
```
Scottj1s: consider:
```suggestion
        auto manager = winrt::Microsoft::Windows::ApplicationModel::Resources::ResourceManager(priPath);
```

dev/Interop/StoragePickers/PickerLocalization.h
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation and Contributors.
+// Licensed under the MIT License.
+
+#pragma once
+#include "PickerCommon.h"
+
+namespace PickerCommon {
```
Scottj1s: PickerLocalization namespace?

eng/Version.Details.xml
Scottj1s: No, Microsoft.Windows.SDK.BuildTools.MSIX is considered an external dependency, like Microsoft.Web.WebView2.  Maestro won't update, and I don't think we need to track it in Version.Details.xml - only Version.Dependencies.xml,props.

dev/WindowsAppRuntime_DLL/packages.config
Scottj1s: what build error are you resolving with a Microsoft.Windows.SDK.BuildTools.MSIX reference?  this seems odd.

dev/Interop/StoragePickers/PickerLocalization.cpp
DrusTheAxe: >should we consider adding special handling to fall back

No.

You should expect to find the resources you added to WinAppSDK for this express purpose.

The only reason this would fail is unavoidable runtime errors (e.g. E_OUTOFMEMORY) -- in which case anything _else_ would be equally subject to failure if not making a bad situation worse -- or incorrect installation of their app and/or winappsdk, or likewise bugged app configuration -- which shouldn't be a problem.

Any 'fallback' would be trying to solve a non-concern while also creating more complexity i.e. increased cost to development, maintain, troubleshoot, yadda yadda.

TL;DR don't workaround your peers in WinAppSDK. Rely on them :-)

test/DynamicDependency/data/Microsoft.WindowsAppRuntime.Framework/appxmanifest.xml
DrusTheAxe: > Given the recent metapackage refactor, I’d like to propose a future update to this unit test to use the actual component package instead of current mocked one

That doesn't solve the inherent chicken-egg issue where you'd need to build the repo to then update and build tests of the thing now previously checked in. Bad mojo.

You should consider adding (some) tests to WindowsAppSDKAggregator using the built product. The tests we have there are Integration tests (akin to Windows BVTs) that use the built product to verify working as expected. But these integration tests don't - can't - replace Unit or Functional tests in the repo here.

DrusTheAxe: You do raise an interest point. As we move ahead with the Metapackage we should add integration tests for each Component Package (i.e. sub-Aggregator level) -- new pipeline(s) in this repo to build and run new integration tests checked into this repo to verify the built product Component Package.

As a Unit/Functional test I can have tests verifying the built content in the same repo with real dependencies but a mocked config of the current repo. In Foundation's case it's simple given there's no lower layer Component Packages, but think about e.g. WinUI repo will make a WinUI Component Package including WinUI's built bits and dependent on Foundation, DWriteCore, etc.

@Scottj1s this is already part of the Metapackage / Component Package plan?

dev/Interop/StoragePickers/PickerLocalization.cpp
DrusTheAxe: This is a good question, and it relates to a (very) recent discussion about Microsoft.WindowsAppRuntime.dll's size.

Today it's 1.8M and contains a wide variety of behavior. The question's been raised perhaps the time has come to split into multiple DLLs e.g. Pickers, OAuth, Notifications and AppLifecycle aren't all used by all audiences and probably should not all live in 1 DLL, but rather 2+ DLLs. Exact factoring (and base principles guiding any such factoring) is TBD, so far the discussion is at the 'has the time come' level.

Larger DLLs mean more work at load time - more pages to allocate/map, more imports to resolve and load, more global constructors, yadda yadda. That's fine when you use 80%+ of all APIs, but if not...

@RDMacLachlan @Scottj1s any decisions yet on this?

dev/Interop/StoragePickers/PickerLocalization.cpp
DrusTheAxe: >or separate StoragePickers into its own DLL

Why not?

It's fair to question if there's merit today in StoragePickers being in a separate DLL, but I'm not aware of any reason why it _couldn't_.

dev/Interop/StoragePickers/PickerLocalization.cpp
Hong-Xiang: I’d like to share my personal learning during on adding localization related to this here.

Splitting the runtime into multiple DLLs could benefit not only size and performance, but also improve our engineering system.

If we can separate the runtime DLL into individual functional DLLs—and one step more to make each of them a fully functional NuGet package—it would greatly simplify and standardize tasks like this localization PR. 
For example, imagine we have separate NuGet packages (and DLLs) for MRTCore (already has separate dll), StoragePicker, and the maybe Runtime.dll for other component, and allowing references between them. Then adding localization to StoragePicker—or any other component in this repo in the future—would become much easier. It could be as straightforward as how a typical library or app, by just referencing MRTCore package.

Additionally, we could still maintain a foundation NuGet package that depends on these individual packages. This way, consumer packages of the foundation wouldn’t need to change, similar to how the metadata package works in the aggregator repo.

dev/Interop/StoragePickers/PickerLocalization.cpp
Hong-Xiang: Sorry for the confusion here—when I said “we can't,” I actually meant that I can't move StoragePicker into a separate DLL without broader agreement from the team.

If we do reach consensus on splitting the DLLs (which I fully support and would be happy to help implement, more comments on it in this [comment](https://github.com/microsoft/WindowsAppSDK/pull/5495#discussion_r2160908043) ), I’d suggest handling that in a separate PR. It would be a fairly large change, and I’m hoping to get the localization work into 1.8 exp 4 if possible.

test/DynamicDependency/data/Microsoft.WindowsAppRuntime.Framework/appxmanifest.xml
Hong-Xiang: Just wanted to share some additional context. I’ve been manually testing the component package locally (thanks to the metapackage refactor), and it’s already working well. So at this point, we don’t need to wait for a full build of the final aggregator repo to get a functioning feeder component.

In my local test app, I’m able to consume the picker APIs without any special configuration for MRTCore dependencies—just adding the Foundation NuGet package (in self-contained mode) is enough. Based on that, I believe if we refactor the current unit tests to use the generated Foundation component NuGet package, it should work without needing to handle all the MRTCore dependency issues.

One note: I’m using the Foundation Component NuGet package generated from our Azure PR pipeline, so I assume we can also use it in pipeline/local unit tests. Although don't know how to properly change unit test to use component package yet, and I haven’t been able to use the locally built Foundation NuGet package in manual tests—only the pipeline-generated one works so far. But I believe that can be resolved.

Hope this helps clarify things!

eng/Version.Details.xml
Scottj1s: I think we still need an entry in eng/Version.Dependencies.xml

dev/Interop/StoragePickers/PickerLocalization.cpp
```diff
@@ -0,0 +1,22 @@
...
+#include "PickerLocalization.h"
+#include <winrt\base.h>
+#include <winrt\Microsoft.Windows.ApplicationModel.Resources.h>
+#include <iostream>
+
+namespace PickerLocalization {
+    const winrt::hstring priPath = L"Microsoft.WindowsAppRuntime.pri";
+    winrt::hstring GetStoragePickersLocalizationText(winrt::hstring key)
+    {
+        auto manager = winrt::Microsoft::Windows::ApplicationModel::Resources::ResourceManager(priPath);
```
yeelam-gordon, line 18: Should this be "static auto manager = xxxx", so we don't need to recreate this resource manager everytime.
Not sure if it support in our compiler now, copilot said it support in latest VC version (i.e. static variable inside a method, it will only initialize once like putting it outside, but only this method can use it)

yeelam-gordon: While, I am not sure if the "Culture" change will impact this.
I assume if we keep recreating the Manager, it may be aware the Culture change within a process cycle.

dev/Interop/StoragePickers/PickerLocalization.cpp
```diff
@@ -0,0 +1,22 @@
...
+        return manager.MainResourceMap().GetValue(key).ValueAsString();
```
yeelam-gordon, line 19: Let's say my culture doesn't have localization string.
I assume this one:
1. Will not throw exception. (I hope not)
2. But will it fallback to use English? i.e. at least not empty

In theory, there should be fallback by MRT, which my hope if we define the Resource file correct?
The following has clear fall back hierachy for specific => neutral (common fallback => invariant (everything fallback)
strings\
├── de\resources.resw        ← neutral German
├── de-DE\resources.resw      ← specific German in Germany
├── zh\resources.resw        ← neutral Chinese
├── zh-CN\resources.resw      ← specific Chinese in China
├── en-US\resources.resw     ← specific US English
└── resources.resw           ← invariant (often default, typically English)

dev/Interop/StoragePickers/PickerLocalization.cpp
yeelam-gordon: Last but not least, let's say this helper method throws exception for whatever reason, but we don't want to fail the app.

Should we like:
1. In this method signature, provide a parameter for default return value (e.g. "All Files") in case any failure
2. Try catch, and return the default value in finally. Exception logging for sure.

dev/Interop/StoragePickers/StoragePickers.vcxitems
```diff
@@ -46,4 +48,20 @@
   <ItemGroup>
     <Midl Include="$(MSBuildThisFileDirectory)Microsoft.Windows.Storage.Pickers.idl" />
   </ItemGroup>
+  <ItemGroup>
```
yeelam-gordon, line 51: Does it support like:
```md
<ItemGroup>
  <PRIResource Include="$(MSBuildThisFileDirectory)Strings\**\*.resw" />
</ItemGroup>
```

yeelam-gordon (approved):
See my feedback if those are making senses.

dev/Interop/StoragePickers/PickerLocalization.cpp
DrusTheAxe: MRT's mostly bolt-on to Foundation. Handling it via an MRT Component Package would be appropos and beneficial as noted, and relatively easy.

Anything currently in Foundation using MRT would need to split out of WindowsAppRuntime.sln to a separate build flow and separate Component Package, to avoid circular dependencies.

The recent Pickers use MRT. Anything else in Foundation today?

dev/Interop/StoragePickers/PickerLocalization.cpp
DrusTheAxe: MRT's got logic to pick 'best fit' given the available resources and machine/user/process properties.

'best fit' ideally is 'exact match' but if not you still get _something_ - worst case the default specified in appxmanifest.xml.

dev/Interop/StoragePickers/PickerLocalization.cpp
Hong-Xiang: Hi @DrusTheAxe , thanks for your comments.
Pickers is the first component in Foundation to use MRT.
However, there is actually no real circular dependencies in WindowsAppRuntime.sln. Currently foundation has a interesting pipeline/build script, the MRT Core project has its own solution file, and in WindowsAppRuntime.sln, we reference its project only to access the winmd files. While this setup works, it does introduce some confusion, and we plan to refactor it in the next release iteration.
Unfortunately, this kind of refactor on build pipeline isn’t a straightforward change, as it would require updates to our pipeline and build scripts. Would you be open to letting this PR focus solely on adding minimal localization support, and we can address the project structural improvements in a dedicated follow-up PR?

dev/Interop/StoragePickers/PickerLocalization.cpp
Hong-Xiang: Hi @DrusTheAxe , thanks for your guidance! I’ve also done some testing across different languages. It seems the system applies a fairly complex set of rules to determine the best-fit match, and as you pointed out, it always falls back to some text.

dev/Interop/StoragePickers/StoragePickers.vcxitems
Hong-Xiang: Thanks for your suggestions, added in #5557

dev/Interop/StoragePickers/PickerLocalization.cpp
Hong-Xiang: Thanks for your suggestion, added in #5557

PR #5498, opened 2025-06-09
Upload TAEF test results

haonanttt:
# Goal
Upload TAEF test results to ADO artifacts and ADO "Test" tab

# Changes in this PR
1. In `TestAll.ps1`, modified the command calling TE.exe to log the test and take screenshots. For the logs and screenshots, copy them to `TestOutput` folder that will be uploaded to artifacts
2. Add a `test/AppModelProviders.wprp` file. Currently this file is directly copied from Aggregatore repo
3. In `build/AzurePipelinesTemplates/WindowsAppSDK-RunTests-Steps.yml`, call a `WindowsAppSDK-ConvertWttLogToXUnit-Steps` from Config repo to convert the .wtl file to a xunit format. The conversion task is another PR in Config repo
4. In `build/AzurePipelinesTemplates/WindowsAppSDK-RunTests-Steps.yml`, upload the test result and screenshots to ADO artifacts

# Insights needed from reviewers
1. The `test/AppModelProviders.wprp` file is directly copied from Aggregator repo, is this meet the requirement in Foundation?
2. Since the "convert the .wtl log to xunit format, therefore ADO test tab could accept the format" scenario is useful and needed in several different repo (at least for Foundation, Aggregator, WinUI). Please help me review my PR in Config repo which extract the conversion as a templated yml task
3. Currently the Test stages continues to fail, which is actually blocking. My current proposal is to "add a white list that will bypass the existing test failure", then the existing failure will not block and new coming failures will still block. How do you think about this proposal?

WARNING:


eng/Version.Details.xml
```diff
@@ -35,6 +35,10 @@
       <Uri>https://dev.azure.com/microsoft/ProjectReunion/_git/WindowsAppSDKClosed</Uri>
       <Sha>f52676312ebf1d66ef59731feb4fad29a713a897</Sha>
     </Dependency>
+    <Dependency Name="Microsoft.ProjectReunion.WinUI.TransportPackage" Version="3.0.0-zmain.250527.0-nightly">
```
DrusTheAxe: Why is Foundation depending on WinUI? That's an inversion of layering

haonanttt: In short, I'm trying to bring the `ConvertWttLogToXUnitLog` from WinUI-lift to foundation, making our foundation's pipeline could upload the TAEF test results to ADO page. 

Current problem is that "Foundation pipeline will run the TAEF tests, but will neither publish the result as artifact, nor convert the result to xunit format and upload to ADO test tab. (Aggregator will both publish test result to artifact and convert the .wtl file into xunit format and upload to ADO test tab. In this PR I'm trying to make Foundation also handle test results like Aggregator)"
Alex have told me in his memory Aggregator might introduces `ConvertWttLogToXUnitLog` from WinUI, so that's the reason why i also depend on WinUI-lift here. I'm trying to use the same file for this conversion.
This PR is still a draft one, I'll modify the title and add a [draft] tag into it. Thank you for sharing your insight! I'll reach out to you offline for more discussion here!

build/AzurePipelinesTemplates/WindowsAppSDK-RunTests-Steps.yml
```diff
@@ -168,3 +172,41 @@ steps:
...
+  - task: PowerShell@2
+    displayName: Edit the artifactAttempt variable when the job has failed
+    condition: failed()
+    inputs:
+      targetType: 'inline'
+      script: |
+        Write-Host "##vso[task.setvariable variable=artifactAttempt]_$(System.JobAttempt)"
+
+  - task: PublishPipelineArtifact@1
+    condition: succeededOrFailed()
```
alexlamtest, line 210: Is it intentional that this task is using a different condition from the PublishPipelineArtifact task? I expect the "Edit artifactAttempt" task above to hardly ever introduce a failure. That led me think PublishTestResults should also be succeededOrFailed()?

build/AzurePipelinesTemplates/WindowsAppSDK-RunTests-Steps.yml
```diff
@@ -1,11 +1,15 @@
...
   buildPlatform: ''
   buildConfiguration: ''
+  ImageName: ''
   TaefSelect: '*'
   BinaryCompatSwitch: ''
   testLocale: ''
 
 steps:
+  - checkout: self
+    path: s
```
alexlamtest, line 11: Isn't this the default, so we may not really need this?

build/AzurePipelinesTemplates/WindowsAppSDK-RunTests-Steps.yml
haonanttt: No, I just copied them from Aggregator, didn't change the condition for any of them. I think it make sense to align there 2 tasks, so changed the `PublishTestResults` to `succeededOrFailed()` as well

build/AzurePipelinesTemplates/WindowsAppSDK-RunTests-Steps.yml
haonanttt: If we don't explicitly use `checkout` task, the `checkout: self` is default. But as we added a `checkout: WindowsAppSDKConfig` task at line 176, we need to also add `checkout: self`. Otherwise the current branch will not be checked out. This could be seen at [this run](https://microsoft.visualstudio.com/ProjectReunion/_build/results?buildId=125015579&view=logs&s=2e1a7036-387b-52ab-3a04-7cccd5380c97&j=ac72351a-ad63-5c2a-8d9c-5ffee596248d). Comparing with original flow, the checkout self task is missed
Before this PR:
image
This PR without `checkout: self`:
image

If we don't add the checkout path, the `checkout: self` task somehow will checkout the code to a wrong path, see [this run](https://microsoft.visualstudio.com/ProjectReunion/_build/results?buildId=125016890&view=logs&j=ac72351a-ad63-5c2a-8d9c-5ffee596248d&t=0e10c163-6f40-5750-d49e-212242fca467):
Before this PR:
image
This PR without `path: s`:
image

According to the above results, we need to explicitly write checkout task here

alexlamtest: I understand the checkout, but am unaware that the "path: s" is needed. Thanks for explaining!

PR #5538, opened 2025-06-18
One page design for `FileSavePicker.SuggestedSaveFilePath`

DinahK-2SO:
One page design for `FileSavePicker.SuggestedSaveFilePath`

Code implementation: #5547


specs/StoragePickers/SuggestedSaveFile.md
```diff
@@ -0,0 +1,338 @@
...
+    `Windows.Storage.StorageFile` to `Microsoft.Windows.Storage.Pickers.SuggestedSaveFile`.
+
+## 4. Detailed Design
+
+The `SuggestedSaveFile` class will provide the necessary information for a suggested save file.
+
+```cs
+namespace Microsoft.Windows.Storage.Pickers {
+    runtimeclass SuggestedSaveFile {
+        // Constructor is not public
```
Hong-Xiang: So the next static method is not related constructor. Maybe adding a comment here would help to reduce confusion.

specs/StoragePickers/SuggestedSaveFile.md
```diff
@@ -0,0 +1,338 @@
...
+        static SuggestedSaveFile CreateFromPath(String path);
+    }
+
+    runtimeclass FileSavePicker {
+        // ... other members
+        SuggestedSaveFile SuggestedSaveFile;
```
Hong-Xiang: We still need to find out that if we accepting a WinRT runtimeclass from this property, how could we get its underlying cpp object to retrieve its path field.

specs/StoragePickers/SuggestedSaveFile.md
```diff
@@ -0,0 +1,338 @@
...
+the `SuggestedSaveFile`. 
+
+    This takes precedence over the `FileSavePicker.SuggestedFileName` property if both are set.
+
+In essence, `SuggestedSaveFile` provides a comprehensive hint for both the target save location 
+(directory) and the proposed file name. 
+
+The file's full path, provided when creating the `SuggestedSaveFile` object via its static factory 
+method, contains all the necessary information (directory and file name) for the picker to implement 
+these behaviors internally.
```
asklar: this explains the 'what' but not the 'why', e.g. why does this need to be a runtimeclass, rather than just using the string path? why does the design explicitly require the path to not be readable after it is created/set?

jonwis: This type doesn't need an intermediate object.  Win32 apps think in paths, not storagefiles. Everywhere a Win32 app might want to save something is representable by a "regular string path." While it's not as strongly typed as an object, the naming makes it clear - "this is the path where you should save the file."

So the updated API would be:

```c#
runtimeclass FileSavePicker
{
    FileSavePicker(Microsoft.UI.WindowId windowId);

    string CommitButtonText;

    string SettingsIdentifier;
    string DefaultFileExtension;
    string SuggestedFileName;
    string SuggestedPath;
    IMap<string, IVector<string>> FileTypeChoices{ get; };

    PickerLocationId SuggestedStartLocation;

    Windows.Foundation.IAsyncOperation<PickFileResult> PickSaveFileAsync()
}
```

A win32 app then says this (assuming `m_config` is a type that stores data for the app):

```c++
auto picker = FileSavePicker{this->m_windowId};
picker.DefaultFileExtension(L".kitten");
picker.SuggestedFileName(cleanup_filename(this->m_currentDocumentTitle));
picker.SuggestedPath(this->m_config.user_specified_save_path());
picker.SuggestedStartLocation = PickerLocationId::Custom;
auto result = co_await picker.PickSaveFileAsync();
if (result) {
    this->m_writeDocumentToFile(result.Path());
}
```

An app that has code still using StorageFile can interop with this via:

```c++
// as above, but...
winrt::StorageFolder folder = get_default_storage();
picker.SuggestedPath(folder.Path());
// and then...
std::filesystem::path savePath{result.Path()};
auto folder = co_await winrt::StorageFolder::GetFolderFromPathAsync(savePath.parent_path().native());
auto file = co_await folder.CreateFileAsync(savePath.filename());
// Write the content to "file"
```

The changes suggested are then:

* Remove the SuggestedSaveFile _type_ and use `String` instead
* Add a new `PickerLocationId` which is `.Custom` meaning that the picker should use the `.SuggestedPath`

specs/StoragePickers/SuggestedSaveFile.md
jonwis: See reply to @asklar - sidestep that by just returning the path.

jonwis (changes requested):

specs/StoragePickers/SuggestedSaveFile.md
```diff
@@ -0,0 +1,346 @@
...
+    *   **Current Design:**
+        In this design, we're adopting the 3rd option - static factory methods:
+        ```cs
+        runtimeclass SuggestedSaveFile {
+            // Constructor is not public
+            static SuggestedSaveFile CreateFromPath(String path);
+        }
+        ```
+
+*   **8.2.3 Rationale for Encapsulating the Path Property**
```
DinahK-2SO: Hi @jonwis  and @asklar , thanks for joining this review! Here are the reasons of creating a class for this input. It's mainly reserved for potential future updates.

specs/StoragePickers/SuggestedSaveFile.md
DinahK-2SO: Thanks @jonwis for joining this review! We need to encapsulate the Path Property, please find and continue the discussion in **8.2.3 Rationale for Encapsulating the Path Property**.

I am closing this thread :)

benstevens48:
Why have a suggested file instead of folder? What if we want the initial file name to be blank, in case there is no good default name or we want to force the user to choose a name?

specs/StoragePickers/SuggestedSaveFile.md
jonwis: StorageFile _should not be used_ as an input/output type for any new APIs.  The Windows App SDK has limited support for UWP clients. In API design, we try to avoid "low value intermediates" - the SuggestedSaveFile here provides limited value compared to taking a `String Path` property diectly.

All related languages & runtimes provide "path based" primary/base operations:

* [System.IO.File](https://learn.microsoft.com/en-us/dotnet/api/system.io.file?view=net-9.0) is a static class operating on file path strings
* [CreateFileW](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew) is a static Win32 function for opening files by path strings
* [fs.open](https://nodejs.org/api/fs.html#fsopenpath-flags-mode-callback) in Rust operates on file path strings
* [os.path](https://docs.python.org/3/library/os.path.html) in Python operates on file path strings

I can see a world where [PickSaveFileAsync](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.windows.storage.pickers.filesavepicker.picksavefileasync?view=windows-app-sdk-1.8) returns a `PickFileSaveResult` object (rather than the generic `PickFileResult` object) with additional properties, like "did the user cancel the operation" or "did the user click some other affordance in the file-save UI", etc.  That object, then, contains a `String Path` + whatever other properties you want.  But don't introduce an object whose sole purpose is to proxy a StorageFile, which is a type we want to avoid in the future.

Hong-Xiang: Hi @jonwis , thanks for your suggestions! Using a path would definitely simplify both the interface and implementation. We understand that we are trying to avoid use StorageFile in new API, actually the main motivation behind introducing the new proxy type is to avoid any explicit use of StorageFile in our new API.

Our concern is that relying solely on paths might not fully cover all scenarios where StorageFile could be used. That’s why we created this new type—to act as an abstraction layer. If we ever need to bring back StorageFile support in the future, we can do so without disrupting the API.

From your suggestions, can I confirm that using paths is sufficient for all the scenarios we plan to support? Would this mean we may have limited support for container scenarios—where I assume StorageFile might carry security-related tokens that a string path can’t? (In such case Win32 isolation could be a workaround)

DinahK-2SO: Offline discussion concluded to use a string type.

codendone (approved):
Looks good to me. Per @benstevens48's question, if `SuggestedSaveFile` only points at a folder instead of a file in a folder, will that have the effect of setting the start folder with no default filename? Or is it required to include a suggested filename?

specs/StoragePickers/FileSavePicker.SuggestedSaveFilePath.md
```diff
@@ -0,0 +1,264 @@
...
+
+*   **Initial Directory:** The dialog defaults to the directory of the `SuggestedSaveFile`. 
+
+    This behavior overrides any picker-specific remembered folder settings and ensures the dialog 
+    opens in the suggested file's parent folder.
+
+*   **Pre-filled File Name:** The file name field in the dialog is pre-populated with the name of 
+the `SuggestedSaveFile`. 
+
+    This takes precedence over the `FileSavePicker.SuggestedFileName` property if both are set.
```
DinahK-2SO: Thanks @codendone for reviewing both the spec and code! Here we explained that the Path of SuggestedSaveFile should take precedence over the `SuggestedFileName` #Closed

DinahK-2SO:
> Looks good to me. Per @benstevens48's question, if `SuggestedSaveFile` only points at a folder instead of a file in a folder, will that have the effect of setting the start folder with no default filename? Or is it required to include a suggested filename?

Thanks @codendone for reviewing this design! 

First, The SuggestedSaveFile of UWP FileSavePicker doesn't accept a StorageFolder object - the compiler would show this error
> Cannot implicitly convert type 'Windows.Storage.StorageFolder' to 'Windows.Storage.StorageFile'

image

In the scope of our new APIs, I would strongly recommend the logic of `TrySuggestedSaveFilePath` to be simple as below:

`TrySuggestedSaveFilePath` takes whatever after the last slash to fill in the file name box,
as long as the part before last slash is a valid and existing folder.
And return false without changing any value if the input doesn't even contain a slash, or the folder doesn't exist.
[Test cases here for clarification](https://github.com/microsoft/WindowsAppSDK/pull/5547/files#r2183261542), will update the content in this design next.

The input would be largely depends on API consumer, but above logic is concrete - The method itself doesn't decide if the input path is a file path or folder path.

This is to reduce the complexity and avoid more undefined behaviors (e.g. should `.git` in `C:\temp\MyProject\.git` be treated as a filename or a folder name? what if the input string is `C:\temp\1.txt` but there's folder `1.txt` under `C:\temp`? etc. )

For the questioned scenario, we can add another method in future, to set the suggested folder without file name, e.g. introducing a `SuggestedFolder` attribute and `TrySetSuggestedFolder`, and look for the folder of input path string, no matter the format.

codendone:
> `TrySuggestedSaveFilePath` takes whatever after the last slash to fill in the file name box, as long as the part before last slash is a valid and existing folder. And return false without changing any value if the input doesn't even contain a slash, or the folder doesn't exist.

I agree with that design.

> This is to reduce the complexity and avoid more undefined behaviors (e.g. should `.git` in `C:\temp\MyProject\.git` be treated as a filename or a folder name? what if the input string is `C:\temp\1.txt` but there's folder `1.txt` under `C:\temp`? etc. )

This relates to the question "if SuggestedSaveFile only points at a folder instead of a file in a folder, will that have the effect of setting the start folder with no default filename? Or is it required to include a suggested filename?". 

From the implementation, I believe the current design is if `C:\temp\MyProject\.git\` is specified with the trailing `\`, then a folder is specified and will be used (assuming it exists) and no filename will be set (allowing `SuggestedSaveFile` to be used, if that is set). I think that design is good.

But, as you said, what happens if `C:\temp\MyProject\.git` is set? I think the current implementation will use `.git` as the default filename even if that exists as a folder. That seems wrong. It may be necessary to check if the full specified path points to a folder and treat it as a folder with no filename.

Splitting out a separate `SuggestedFolder` attribute and `TrySetSuggestedFolder` seems unnecessary (and potentially confusing) if the handling of SuggestedSaveFilePath simply does that extra "the path is just a folder" check.

DinahK-2SO:
> But, as you said, what happens if `C:\temp\MyProject\.git` is set? I think the current implementation will use `.git` as the default filename even if that exists as a folder. That seems wrong. It may be necessary to check if the full specified path points to a folder and treat it as a folder with no filename.

Thanks @codendone for raising this point! I also discussed this topic with @yeelam-gordon yesterday, and we concluded the current design could be more beneficial.

It's great to know that we have reached consensus on the first case:
* `C:\temp\MyProject\.git\` sets folder `C:\temp\MyProject\.git`, with empty file name; (Thanks @benstevens48 for raising the use scenario, with this tailing `\`, developer can set folder while leaving the initial file name blank)

The question is the second case:
* `C:\temp\MyProject\.git` sets folder `C:\temp\MyProject`, with `.git` filled into file name, even when there's a folder `C:\temp\MyProject\.git`.

While it might seem counter-intuitive at first, this behavior is based on a very strict logic. It makes the API's behavior 100% **predictable** and gives developers clearer control over the code.

* Via `TrySetSuggestedSaveFilePath`, the developer attempts to pass 2 suggestions to the end user: a location and a file name.

  With the current design, if the folder `C:\temp\MyProject\.git` exists, end user can see:
  1. the save dialog opens `C:\temp\MyProject` (passing the 1st suggestion).
  1. it is also suggested to save the new file as `.git` (the 2nd suggestion), although they cannot save as there's already a folder named `.git` 
  
  In this case, the end user can decide:
  1. rename the folder `.git` to another name and save file
  1. change the new file's name
  1. or choose any other location with any other file name
  
  All based on 2 suggestions from the developer.
  
* In contrast, if we check the folder existence of the input string and jump into the deeper folder, the API's behavior would become more complex and unpredictable. In above example, end user would see:
  1. the save file dialog opens `C:\temp\MyProject\.git`, which was not the intentioned folder of developer, 
  1. no suggested file name - another information failed to pass to the end user.
  
  This would be like a second-guessing the developer's intent based on the unpredictable state of the end-user's file system, bringing an unreliable API's behavior

specs/StoragePickers/FileSavePicker.md
```diff
@@ -21,7 +21,7 @@ runtimeclass FileSavePicker
     string SettingsIdentifier;
     string DefaultFileExtension;
     string SuggestedFileName;
-    Windows.Storage.StorageFile SuggestedSaveFile;
+    string SuggestedSaveFilePath;
```
codendone: I think the public docs will need some details for this property (and will ideally have at least a sentence, or possibly more) for all properties). Ideally, this doc will include some of those proposed details, similar to the property definitions for RuntimeCompatibilityOptions (https://github.com/microsoft/WindowsAppSDK/blob/8c6bc94b144a1e3b56953ef84dbd7c00fc755683/specs/Compatibility/RuntimeCompatibilityOptions.md?plain=1#L95).

Part of what I'd like to see included:
* SuggestedSaveFilePath will override a filename set in SuggestedFileName
* If the app wants to set a folder without a filename, the path must end in `\`.

(This could be a future PR.)

codendone: `TrySetSuggestedSaveFilePath` is not listed here, and the property is now read-only

codendone (approved):

specs/StoragePickers/FileSavePicker.md
DinahK-2SO: Thanks @codendone ! I updated FileSavePicker.md

DinahK-2SO: Thanks @codendone ! I updated FileSavePicker.md with more details.

codendone (approved):

PR #5445, opened 2025-05-22
WinAppSdk 1.7 installer not able to install Singleton package when WinAppSdk 1.6 is installed and 1.6 Singleton is being used in the background

ssparach:
A microsoft employee must use /azp run to validate using the pipelines below.

WARNING:
Comments made by azure-pipelines bot maybe inaccurate.
Please see pipeline link to verify that the build is being ran.

For status checks on the main branch, please use TransportPackage-Foundation-PR
(https://microsoft.visualstudio.com/ProjectReunion/_build?definitionId=81063&_a=summary)
and run the build against your PR branch with the default parameters.

---------------------------------------------------------------------------------------------------------------
**Description:** 
The current installer behavior does not allow a newer version of the Singleton package to install if an older version is running in the background. This PR fixes this issue by always calling AddPackage with ForceTargetAppShutdown for the Singleton, and then restarting the NotificationLongRunningPlatform.

**Validation:** 
Manual confirmation on VM with repro
Functional tests and pipeline validation

PR #5540, opened 2025-06-18
[1.7 servicing] WinAppSdk 1.7 installer not able to install Singleton package when WinAppSdk 1.6 is installed and 1.6 Singleton is being used in the background

ssparach:

WARNING:

--------------------------------------------------------------------------------------------------------------------------------------
This is a back-port of #5445 to release/1.7-stable. 
[https://task.ms/58022587](https://task.ms/58022587)

### DIFF
--- a/dev/Deployment/DeploymentManager.cpp
+++ b/dev/Deployment/DeploymentManager.cpp
@@ -543,12 +543,7 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
         initializeActivity.SetInstallStage(::WindowsAppRuntime::Deployment::Activity::DeploymentStage::AddPackage);
         for (auto package : c_targetPackages)
         {
-            auto isSingleton{ false };
-            if (CompareStringOrdinal(package.identifier.c_str(), -1, WINDOWSAPPRUNTIME_PACKAGE_SUBTYPENAME_SINGLETON, -1, TRUE) == CSTR_EQUAL)
-            {
-                isSingleton = true;
-            }
-
+            auto isSingleton{ CompareStringOrdinal(package.identifier.c_str(), -1, WINDOWSAPPRUNTIME_PACKAGE_SUBTYPENAME_SINGLETON, -1, TRUE) == CSTR_EQUAL };
             initializeActivity.Reset();
             initializeActivity.SetCurrentResourceId(package.identifier);
 
@@ -573,7 +568,7 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
 
             // If the current application has runFullTrust capability, then Deploy the target package in a Breakaway process.
             // Otherwise, call PackageManager API to deploy the target package.
-             // The Singleton package will always set true for forceDeployment and the running process will be terminated to update the package.
+            // The Singleton package will always set true for forceDeployment and the running process will be terminated to update the package.
             if (initializeActivity.GetIsFullTrustPackage())
             {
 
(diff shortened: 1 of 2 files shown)


dev/Deployment/DeploymentManager.cpp
```diff
@@ -543,6 +543,12 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
         initializeActivity.SetInstallStage(::WindowsAppRuntime::Deployment::Activity::DeploymentStage::AddPackage);
         for (auto package : c_targetPackages)
         {
+            auto isSingleton{ false };
+            if (CompareStringOrdinal(package.identifier.c_str(), -1, WINDOWSAPPRUNTIME_PACKAGE_SUBTYPENAME_SINGLETON, -1, TRUE) == CSTR_EQUAL)
+            {
+                isSingleton = true;
+            }
```
Scottj1s: nit- more compactly:
```suggestion
            auto isSingleton{ CompareStringOrdinal(package.identifier.c_str(), -1, WINDOWSAPPRUNTIME_PACKAGE_SUBTYPENAME_SINGLETON, -1, TRUE) == CSTR_EQUAL };
```

dev/Deployment/DeploymentManager.cpp
```diff
@@ -567,19 +573,19 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
 
             // If the current application has runFullTrust capability, then Deploy the target package in a Breakaway process.
             // Otherwise, call PackageManager API to deploy the target package.
+             // The Singleton package will always set true for forceDeployment and the running process will be terminated to update the package.
             if (initializeActivity.GetIsFullTrustPackage())
             {
```
Scottj1s, line 574: empty line

installer/dev/install.cpp
```diff
@@ -265,6 +265,12 @@ namespace WindowsAppRuntimeInstaller
         const auto forceDeployment{ WI_IsFlagSet(options, WindowsAppRuntimeInstaller::Options::ForceDeployment) };
```
Scottj1s: minor: would it be clearer based on the comment below about singleton forcing deployment, to just roll it in?
```suggestion
        const auto forceDeployment{ WI_IsFlagSet(options, WindowsAppRuntimeInstaller::Options::ForceDeployment) || isSingleton };
```

installer/dev/install.cpp
```diff
@@ -265,6 +265,12 @@ namespace WindowsAppRuntimeInstaller
...
         auto& installActivityContext{ WindowsAppRuntimeInstaller::InstallActivity::Context::Get() };
 
+        auto isSingleton{false};
+        if (CompareStringOrdinal(resource.id.c_str(), static_cast<int>(resource.id.size()), applicableSingletonResourceID.c_str(), static_cast<int>(applicableSingletonResourceID.size()), TRUE) == CSTR_EQUAL)
```
Scottj1s: curious - why do we compare to WINDOWSAPPRUNTIME_PACKAGE_SUBTYPENAME_SINGLETON above and to applicableSingletonResourceID here?  maybe consider factoring a common utility method?

dev/Deployment/DeploymentManager.cpp
ssparach: I have incorporated these suggestions and will first commit them to the main branch before cherry-picking them into this servicing PR. Here is the link to the PR for the main branch for your review: #5549. Thanks!

dev/Deployment/DeploymentManager.cpp
ssparach: Incorporated suggestions inside #5549.

installer/dev/install.cpp
ssparach: Incorporated suggestions inside #5549

installer/dev/install.cpp
ssparach: The value differs in both instances. The DeploymentManager focuses on the package name, while the installer executable is concerned with the package's full name, including the architecture, to deploy the appropriate resource package. The DeploymentManager only considers the type of the package i.e. whether it is "main"/ "singleton" etc. It uses this to determine if an existing higher version exists. If so, it reregisters that. If not, it uses this to build the path to the MSIX package inside the framework package's folder for deployment.

Scottj1s (approved):

PR #5541, opened 2025-06-18
Use WindowsAppSDK-Versions.yml from WindowsAppSDKConfig

kythant:

WARNING:

### DIFF
--- a/build/ProjectReunion-BuildFoundation.yml
+++ b/build/ProjectReunion-BuildFoundation.yml
@@ -6,9 +6,9 @@ name: $(version)
 trigger: none
 
 variables:
-- template: WindowsAppSDK-Versions.yml
-- template: WindowsAppSDK-CommonVariables.yml
 - template: WindowsAppSDK-Foundation-TestConfig.yml@WindowsAppSDKConfig
+- template: AzurePipelinesTemplates\WindowsAppSDK-Versions.yml@WindowsAppSDKConfig
+- template: WindowsAppSDK-CommonVariables.yml
 
 resources:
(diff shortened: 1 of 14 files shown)


build/ProjectReunion-BuildFoundation.yml
```diff
@@ -6,17 +6,17 @@ name: $(version)
 trigger: none
 
 variables:
-- template: WindowsAppSDK-Versions.yml
-- template: WindowsAppSDK-CommonVariables.yml
 - template: WindowsAppSDK-Foundation-TestConfig.yml@WindowsAppSDKConfig
+- template: AzurePipelinesTemplates\WindowsAppSDK-Versions.yml@WindowsAppSDKVersionConfig
```
alexlamtest, line 10: For my understanding, aside from the conceptual separation of "test" and "version" stuffs for clarity, are there other reasons for declaring a separate new "config repo" versus using the existing "WindowsAppSDKConfig" declaration?

alexlamtest (approved):

build/ProjectReunion-BuildFoundation.yml
kythant: It is because we need the Version Config to be on a per branch basis and if we have one declaration, then we would negate the advantages of all our other templates being on the main branch of the Config repo. 

Hence we need two resources defined for WindowsAppSDkConfig: one for Versions and one for Tests and other templates

alexlamtest: Thanks!

PR #5547, opened 2025-06-20
FileSavePicker.SuggestedSaveFilePath

DinahK-2SO:
## Description

This PR changes the property of `FileSavePicker.SuggestedSaveFile` into a light-weight string property `FileSavePicker.SuggestedSaveFilePath`, which is the implementation of design spec: #5538

This PR also adds contract for the storage pickers, and excludes the unnecessary property (FolderPicker.FileTypeFilter) from the contract.

## Code Details

This pull request introduces several changes to the `StoragePickers` module, focusing on enhancing the `FileSavePicker` functionality by replacing `SuggestedSaveFile` with `SuggestedSaveFilePath`, removing the `FileTypeFilter` from `FolderPicker`, and adding new API contracts for better versioning. Additionally, it includes updates to test cases and internal configurations for improved functionality and reliability.

### Enhancements to `FileSavePicker`:

* Replaced `SuggestedSaveFile` with `SuggestedSaveFilePath` in `FileSavePicker` to allow specifying file paths directly, including updates to both the implementation (`FileSavePicker.cpp`) and header (`FileSavePicker.h`). [[1]](diffhunk://#diff-59e025f2c0bb98d7a12bb6b5d5bdcc1f63b0ee4f150ef4fde1338a5de4650fc4L65-R71) [[2]](diffhunk://#diff-31e8543f64c4b9c27fdc095f3d4ac245d2c03be85eed061919cb6cd6c3de3ca1L31-R32)
* Added logic to `PickerParameters::ConfigureFileSaveDialog` to handle `SuggestedSaveFilePath` for setting the starting directory and file name in the save dialog, with fallback to `SuggestedFileName`.
* Updated the `.idl` file to reflect the `SuggestedSaveFilePath` change in the `FileSavePicker` runtime class.

### Simplifications to `FolderPicker`:

* Removed the `FileTypeFilter` property from `FolderPicker`, simplifying its API and associated logic in both the implementation (`FolderPicker.cpp`) and header (`FolderPicker.h`). [[1]](diffhunk://#diff-fafd9b82668f2ebe3d318ec048dab1bf54074c05dd6f56dc392b0d3e5c0294c1L54-L65) [[2]](diffhunk://#diff-17cbee7c1a27ec54bd9fddbae53d09b13a7b17428b0423b0337144934fea2c36L27-L28)

### API Versioning and Contracts:

* Introduced a new `StoragePickersContract` (version 1.8) in the `.idl` file to enable better API versioning and feature gating for `StoragePickers`.

### Test Suite Updates:

* Added new test cases for `PickerParameters::ConfigureFileSaveDialog` to verify behavior when `SuggestedSaveFilePath` is set and its precedence over `SuggestedFileName`.
* Updated tests for `FileSavePicker` to validate the `SuggestedSaveFilePath` property.

### Internal Configuration Updates:

* Enhanced `PickerCommon.cpp` with additional includes (`shellapi.h`, `shobjidl_core.h`, etc.) to support new functionality for file path handling.


DinahK-2SO:
Good morning @jonwis  and @asklar , thanks for reviewing the design spec #5538.

This is a implementation PR, it contains the changes raised during the design spec review and discussion.

Please also review this PR when you're available.

test/DynamicDependency/data/Microsoft.WindowsAppRuntime.Framework/appxmanifest.xml
```diff
@@ -153,6 +153,7 @@
         <ActivatableClass ActivatableClassId="Microsoft.Windows.Storage.Pickers.FileOpenPicker" ThreadingModel="both" />
         <ActivatableClass ActivatableClassId="Microsoft.Windows.Storage.Pickers.FileSavePicker" ThreadingModel="both" />
         <ActivatableClass ActivatableClassId="Microsoft.Windows.Storage.Pickers.FolderPicker" ThreadingModel="both" />
+        <ActivatableClass ActivatableClassId="Microsoft.Windows.Storage.Pickers.SuggesteeSaveFile" ThreadingModel="both" />
```
MartyIX: Is this a typo
```suggestion
        <ActivatableClass ActivatableClassId="Microsoft.Windows.Storage.Pickers.SuggestedSaveFile" ThreadingModel="both" />
```
?

DinahK-2SO: Nice catch! And also, this class no longer exists. I am removing it from the config file.
Thanks a lot! @MartyIX

dev/Interop/StoragePickers/Microsoft.Windows.Storage.Pickers.idl
```diff
@@ -57,18 +65,20 @@ namespace Microsoft.Windows.Storage.Pickers
         String CommitButtonText;
         Windows.Foundation.Collections.IMap<String, Windows.Foundation.Collections.IVector<String> > FileTypeChoices{ get; };
         String DefaultFileExtension;
-        Windows.Storage.StorageFile SuggestedSaveFile;
+        String SuggestedSaveFilePath;
         String SuggestedFileName;
```
codendone, line 68: I apparently missed this in the spec PR: How do `SuggestedSaveFilePath` and `SuggestedFileName` interact? (The answer appears to be in PickerCommon.cpp.)

dev/Interop/StoragePickers/PickerCommon.cpp
```diff
@@ -249,4 +251,50 @@ namespace PickerCommon {
...
+            auto suggestedPath = std::filesystem::path{ SuggestedSaveFilePath.c_str() };
+            if (!suggestedPath.empty())
+            {
+                // Set the directory of SuggestedSaveFile as the starting location of save picker Dialog
+                auto dirPath = suggestedPath.parent_path();
+                if (!dirPath.empty())
+                {
+                    winrt::com_ptr<IShellItem> shellItem;
+                    HRESULT hr = SHCreateItemFromParsingName(dirPath.c_str(), nullptr, IID_PPV_ARGS(shellItem.put()));
+                    if (SUCCEEDED(hr))
```
codendone: I assume a failure HR is returned if the specified path is invalid? And we just quietly ignore the value in that case??

dev/Interop/StoragePickers/PickerCommon.cpp
```diff
@@ -249,4 +251,50 @@ namespace PickerCommon {
...
+                    {
+                        check_hresult(dialog->SetFolder(shellItem.get()));
+                    }
+                }
+
+                // Extract filename and use it (takes precedence over SuggestedFileName)
```
codendone: >// Extract filename and use it (takes precedence over SuggestedFileName) [](http://example.com/codeflow?start=16&length=72)

Was this mentioned in the spec?

DinahK-2SO: Yes! It's on the second section: https://github.com/microsoft/WindowsAppSDK/pull/5538/files#r2165368768

dev/Interop/StoragePickers/Microsoft.Windows.Storage.Pickers.idl
DinahK-2SO: Yes! The path of SuggestedSaveFilePath takes precedence over the SuggestedFileName. The behavior consists with that of the UWP FileSavePicker. https://github.com/microsoft/WindowsAppSDK/pull/5538/files#r2165368768

dev/Interop/StoragePickers/PickerCommon.cpp
```diff
@@ -249,4 +251,50 @@ namespace PickerCommon {
...
+                std::wstring fileName = suggestedPath.filename().wstring();
```
Hong-Xiang: Just curious, when getting path above, you are using hstring -> c_str, and when getting filename, you are using wstring -> hstring, is there any concern here? Does those path/filename working on unicode values? Would you mind to test some basic scenarios e.g. path/filename with unicode e.g. chinese characters, check if you an properly get those string in C# calling these APIs?

DinahK-2SO: Thanks @Hong-Xiang , the fileNameToSet is eventually an hstring, the wstring is for the intermediate `std::filesystem::path` object.

I tested to send a path from the open picker result - with Chinese characters in file name, and the file name can be displayed correctly.
image

codendone (changes requested):
Just the remaining question about how to handle failure from SHCreateItemFromParsingName(). I'm not sure it is good to just ignore the failure. I expect at least some failures would have previously been caught in app code, such as if the string doesn't have valid path syntax. Would the old model have also required that the path exist?

I wonder if we should throw an invalid argument error (with an error string to explain which arg is invalid). What does the old syntax or the UWP picker do with an invalid path syntax or missing path?

DinahK-2SO:
> Just the remaining question about how to handle failure from SHCreateItemFromParsingName(). I'm not sure it is good to just ignore the failure. I expect at least some failures would have previously been caught in app code, such as if the string doesn't have valid path syntax. Would the old model have also required that the path exist?
> 
> I wonder if we should throw an invalid argument error (with an error string to explain which arg is invalid). What does the old syntax or the UWP picker do with an invalid path syntax or missing path?

Hi @codendone , Thanks for raising this!

In most cases, the old model (the UWP FileSavePicker) wouldn't encounter this issue, because the UWP picker APIs typically take a `StorageFile` object to set this value. When a `StorageFile` is created, the system already ensures that this file and its folder exists - so the UWP `FileSavePicker` doesn't need to do any extra validation.

That said, it *is* possible to simulate this scenario in UWP with a specifically designed code and steps:
Use the code 2 launch 2 open pickers, and one save picker:
1. At the first `FileOpenPicker`, manually pick a `StorageFile`.
2. On launching the second picker, manually delete the folder containing the picked file from step 1.
3. Then, before the launching a `FileSavePicker`, setting the previously picked `StorageFile` as `SuggestedSaveFile` in code.

In this case, the save file dialog will still launch successfully because the original `StorageFile` object remains valid for setting the folder - even though the actual folder on file system no longer exists. 

After having launched, the dialog will show an error message : **"Location is not available"**, because the folder that previous existed is gone. **Please note that this message box doesn't block the user from continuing to use the save dialog. The user can dismiss the message box and proceed.**  @asklar , in summary a similar messagebox is shown in the UWP pickers.

And the SuggestedSaveFile's file name is still automatically filled in the file name input box.
image

In contrast, the new picker model uses a string path to set `SuggestedSaveFilePath`, and `SaveFileDialog` requires a valid `ShellItem` to call `SetFolder`. So we must create the shell item from the folder string and validate whether the folder item has been successfully created before setting it to the dialog. 

Even if the folder item creation fails (e.g., the folder doesn't exist), as a suggested value, it should not block users from using the dialog - that's the reason why we initially chose to ignore the error and only to set the folder when it exists.

**On second thought, we could show a message box *before* launching the dialog** - similar to how UWP shows an error *after* launch. This would inform the user about the non-existing but expected folder, without blocking user from using the dialog

1. But the timing differs slightly from UWP's behavior.
1. I'm displaying the message box with warning icon (instead of the error icon of UWP picker), as it won't block user from using the dialog.
1. What's more, this message doesn't have localization yet (pending on a PR for localization #5495 , we can add its localization in future releases)

How do you like this update? Here's a demo of the latest iteration:
path-not-exist

dev/Interop/StoragePickers/PickerCommon.cpp
DinahK-2SO: Thanks @codendone for raising this question! Adding a message box in the latest iteration, which won't prevent the end user from using the picker, as the UWP picker does. Please find more comparison in details here: https://github.com/microsoft/WindowsAppSDK/pull/5547#issuecomment-3006868996

codendone: A few thoughts on a message box solution:
* If the path contains invalid path characters, this just becomes a message to the user and not feedback to the app that it was invalid. This is presumably a loss in fidelity compared to UWP apps hitting an exception when trying to create a StorageItem. I wonder if there is an easy way to check "could this be a valid path" to throw InvalidArg if it could never be valid.
* If it *could* be a valid path but some of the path is missing, maybe just a dialog is okay.
* I wonder if the setter for SuggestedSaveFilePath should check if the path is valid and throw an exception if it isn't. That timing would provide optimal visibility and clarity for apps. It may be desirable to remove the public setter and add a method to set it to make it clearer that it could fail for an invalid path. (If the path is removed after the setter/method, then the message box would more directly mirror UWP behavior.)
* Showing the message box before the save dialog seems more confusing to me than the UWP behavior.
* What code/stack shows the message box for the UWP save dialog? Was that part of the UWP save dialog code, or something below it which we could be getting but maybe haven't configured for?

dev/Interop/StoragePickers/PickerCommon.cpp
```diff
@@ -249,4 +253,62 @@ namespace PickerCommon {
...
+                    if (SUCCEEDED(hr))
+                    {
+                        check_hresult(dialog->SetFolder(shellItem.get()));
+                    }
+                    else
+                    {
+                        // The messagebox won't block user from using the FileSavePicker Dialog.
+                        // TODO: the message and label here needs globalization.
+                        MessageBoxW(
+                            nullptr,
```
codendone: >nullptr [](http://example.com/codeflow?start=28&length=7)

If we end up showing a message box, the parent HWND should be correctly set.

DinahK-2SO: Thanks @codendone for catching this. Added HWnd in the latest iteration.

dev/Interop/StoragePickers/PickerCommon.cpp
```diff
@@ -249,4 +253,62 @@ namespace PickerCommon {
...
+                    {
+                        check_hresult(dialog->SetFolder(shellItem.get()));
+                    }
+                    else
+                    {
+                        // The messagebox won't block user from using the FileSavePicker Dialog.
+                        // TODO: the message and label here needs globalization.
+                        MessageBoxW(
+                            this->HWnd,
+                            std::format(L"{} is unavailable. If the location is on this PC, make sure the device or drive is connected or the disc is inserted, and then try again. If the location is on a network, make sure you\x2019re connected to the network or Internet, and then try again. If the location still can\x2019t be found, it might have been moved or deleted.",
```
asklar: 1. this is not localizable
2. why are we using messagebox inside the app sdk?
3. which activation context is this going to use to ensure this uses latest comctl?
4. more importantly, why show an error at all? either fail the call, fail silently (pick another folder), or create the folder

DinahK-2SO: Hi @asklar , Thanks for reviewing the PR. 

1. About localization, this PR is targeted for exp4 and we will add localization in the future releases (The message here is localizable with the new localization mechanism developed at the same time in #5495 )!

2 and 4. About the reason of adding the message box here, please refer to the previous conversation: https://github.com/microsoft/WindowsAppSDK/pull/5547#issuecomment-3006868996

3. about the activation context, thanks for raising this new point! We can configure actctx in future updates. Here's an example of the look of current message box:
    image

codendone (approved):

dev/Interop/StoragePickers/PickerCommon.cpp
```diff
@@ -110,6 +115,53 @@ namespace PickerCommon {
...
+        {
+            // If no slashes were found, return nullptr as we cannot determine a parent folder
+            return { nullptr, L""};
+        }
+
+        // If the parent folder does not exist or is not a directory, we cannot set folder.
+        if (!std::filesystem::exists(folderPath) || !std::filesystem::is_directory(folderPath))
+        {
+            return { nullptr, L"" };
+        }
```
codendone: Why this change to manual path validation? Does std::filesystem::path.parent_path() plus SHCreateItemFromParsingName not provide sufficient validation?

test/StoragePickersTests/PickerCommonTests.cpp
```diff
@@ -193,5 +197,108 @@ namespace Test::PickerCommonTests
...
+            dialog->GetFileName(fileName.put());
+            VERIFY_ARE_EQUAL(L"MyFile2.txt", std::wstring(fileName.get()));
+        }
+
+        TEST_METHOD(VerifyParseFolderItemAndFileName)
+        {
+            // Arrange.
+            std::wstring testFolder = L"C:\\temp\\testsuggestedSaveFilePath";
+
+            std::vector<std::tuple<bool, std::wstring, std::wstring, bool>> test_cases {
```
DinahK-2SO, line 245: #5538

test/StoragePickersTests/PickerCommonTests.cpp
```diff
@@ -193,5 +197,101 @@ namespace Test::PickerCommonTests
...
+
+            wil::unique_cotaskmem_string fileName{};
+            dialog->GetFileName(fileName.put());
+            VERIFY_ARE_EQUAL(L"MyFile2.txt", std::wstring(fileName.get()));
+        }
+
+        TEST_METHOD(VerifyParseFolderItemAndFileName)
+        {
+            // Arrange.
+            std::wstring testFolder = L"C:\\temp\\testsuggestedSaveFilePath";
```
codendone, line 243: This may not be a valid creatable folder when run in the build pipelines.

test/StoragePickersTests/PickerCommonTests.cpp
```diff
@@ -193,5 +197,101 @@ namespace Test::PickerCommonTests
...
+
+            std::vector<std::tuple<bool, std::wstring, std::wstring, bool>> test_cases {
+            // test folder exists      folder        file              expect non-empty result
+                {true,              testFolder,      L"MyFile.txt",     true },
+                {true,              testFolder,      L"MyFile",         true },
+                {true,              testFolder,      L".source",        true },
+                {true,              testFolder,      L"",               true },
+                {true,              L"C:\\",         L".source",        true },
```
codendone: I'm not certain if build pipelines are guaranteed to have a C drive. It might be okay to leave this as is if it works. Otherwise, it may need to pull an environment variable to in some other way use a confirmed valid drive.

test/StoragePickersTests/StoragePickersTests.cpp
```diff
@@ -169,6 +169,10 @@ namespace Test::StoragePickersTests
             picker.SuggestedStartLocation(winrt::Microsoft::Windows::Storage::Pickers::PickerLocationId::DocumentsLibrary);
             VERIFY_ARE_EQUAL(picker.SuggestedStartLocation(), winrt::Microsoft::Windows::Storage::Pickers::PickerLocationId::DocumentsLibrary);
 
+            bool isSuccess = picker.TrySetSuggestedSaveFilePath(L"C:\\MyFile.txt");
```
codendone, line 172: Same "assumes existence of C drive" issue here.

codendone (changes requested):

dev/Interop/StoragePickers/PickerCommon.cpp
DinahK-2SO: Offline discussion decided not to throw an error box for `suggested` value.

dev/Interop/StoragePickers/PickerCommon.cpp
DinahK-2SO: Thanks @codendone for raising this topic! I changed to manual parsing, thinking a single pass to extract both folder and file paths would be more efficient. 

But on second thought, it maybe not - so I reverted the change back to use filesystem methods.

To answer the question: `parent_path` is sufficient. It fulfills our needs.
image

test/StoragePickersTests/StoragePickersTests.cpp
DinahK-2SO: Thanks @codendone , looks our pipeline runs successfully~
image

test/StoragePickersTests/PickerCommonTests.cpp
DinahK-2SO: Thanks @codendone , looks our pipeline run's successful~
image

test/StoragePickersTests/PickerCommonTests.cpp
DinahK-2SO: Thanks @codendone for raising this concern! This looks like a same question as below (about if the C: drive exists or not) . Therefore, I am closing this thread. If you have more concerns, please feel free to re-activate this comment.

codendone (approved):

DinahK-2SO:
Removing the completed PR from reviewers assignment list.

PR #5553, opened 2025-06-23
Removed *string_view overloads. Added wil::z[w]string_view tests

DrusTheAxe:
Removed *string_view overloads as there's no good way for us to handle it if not null-terminated (callers can easily deal when needed and efficiently best suiting their need). If a *string_view is null-terminated use the PC[W]STR overloads e.g.

```
std::string_view s{ "12.345" };
auto d{ Microsoft::Windows::Foundation::decimal::from_string(s.data());
```

Added wil::zstring_view and wil::zwstring_view tests

Added ActivatableClass to Transport Package manifest fragment for reference in Aggregator


jonwis:
Drat, I commented on a commit - see here https://github.com/microsoft/WindowsAppSDK/commit/89c5be8946277e77c609fcf67215e29193f87f80#r160642740

PR #5570, opened 2025-06-27
Add init_seg to ensure auto initialize runs before user defined static objects

nieubank:

WARNING:

### DIFF
--- a/dev/Common/WindowsAppRuntimeAutoInitializer.cpp
+++ b/dev/Common/WindowsAppRuntimeAutoInitializer.cpp
@@ -59,6 +59,8 @@ namespace Microsoft::Windows::ApplicationModel::WindowsAppRuntime::Common
 // Use init_seg(lib) to ensure this static object is initialized during the "lib" phase,
 // which occurs before user-defined static objects. This guarantees that WindowsAppRuntime
 // auto-initialization completes before any application static globals that might depend on it.
+#pragma warning(disable : 4073)
 #pragma init_seg(lib)
     static AutoInitialize g_autoInitialize;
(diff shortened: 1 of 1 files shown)


dev/Common/WindowsAppRuntimeAutoInitializer.cpp
```diff
@@ -55,5 +55,10 @@ namespace Microsoft::Windows::ApplicationModel::WindowsAppRuntime::Common
 #endif
         }
     };
+
+// Use init_seg(lib) to ensure this static object is initialized during the "lib" phase,
+// which occurs before user-defined static objects. This guarantees that WindowsAppRuntime
+// auto-initialization completes before any application static globals that might depend on it.
+#pragma init_seg(lib)
```
Scottj1s, line 63: is there any potential race now between lib phase object init?  i.e., what if the auto-initializer relies on the CRT?

nieubank: yes and no, technically speaking the CRT is not fully initialized in this state but everything that we *should* be depending on is initialized in the "compiler" section. "lib" is reserved for exactly these purposes, so I'm not too concerned about any potential races here. I have confirmed at least that our existing auto initializers appear to be functioning correctly.

See https://github.com/MicrosoftDocs/cpp-docs/blob/main/docs/preprocessor/init-seg.md
compiler
Reserved for Microsoft C run-time library initialization. Objects in this group are constructed first.

lib
Available for third-party class-library vendors' initializations. Objects in this group are constructed after the ones marked as compiler, but before any others.

PR #5577, opened 2025-06-30
[Fix] Add try-catch to Storage Picker's Localization Code

Hong-Xiang:
Hot fix (mitigate) for pri file not found.

This happens when 

1 using WinAppSDK meta package, using component package works (already has fix PR on WinAppSDK aggregator Repo)
2 in non-self contained scenario - currently using this as a mitigation

example normal behavior (in zh-cn)
image

fall back behaviour
image

### DIFF
--- a/dev/Interop/StoragePickers/PickerLocalization.cpp
+++ b/dev/Interop/StoragePickers/PickerLocalization.cpp
@@ -9,13 +9,13 @@
 #include "PickerLocalization.h"
 #include <winrt\base.h>
 #include <winrt\Microsoft.Windows.ApplicationModel.Resources.h>
-#include "StoragePickersTelemetry.h"
 
 namespace PickerLocalization {
     const winrt::hstring priPath = L"Microsoft.WindowsAppRuntime.pri";
     winrt::hstring GetStoragePickersLocalizationText(winrt::hstring key, winrt::hstring fallback)
     {
-        // adding try-catch to prevent localization error break picker experience
+        // adding try-catch to prevent localization error break picker experience on first shipping
+        // TODO: remove try-catch after stabilization period
         try
         {
             auto manager = winrt::Microsoft::Windows::ApplicationModel::Resources::ResourceManager(priPath);
@@ -23,7 +23,7 @@ namespace PickerLocalization {
         }
         catch (...)
         {
-            StoragePickersTelemetry::StoragePickerLocalizationLookupError(key.c_str());
+            LOG_CAUGHT_EXCEPTION();
             return fallback;
         }
     }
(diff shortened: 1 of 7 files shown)


dev/Interop/StoragePickers/PickerLocalization.cpp
```diff
@@ -9,14 +9,23 @@
...
-        return manager.MainResourceMap().GetValue(key).ValueAsString();
+        // adding try-catch to prevent localization error break picker experience
+        try
+        {
+            auto manager = winrt::Microsoft::Windows::ApplicationModel::Resources::ResourceManager(priPath);
+            return manager.MainResourceMap().GetValue(key).ValueAsString();
+        }
+        catch (...)
+        {
+            StoragePickersTelemetry::StoragePickerLocalizationLookupError(key.c_str());
```
yeelam-gordon: Will it log the "exception" detail?

yeelam-gordon: Since without the exception, we don't know why it failed.

yeelam-gordon (commented):
? Please see my comments.

yeelam-gordon:
> Can we hold off on this until we determine root cause of the missing WAR PRI file in FWP-dependent deployments? This mitigation may not be necessary and would be ideal to avoid it.

@Scottj1s , Xiang know about it, it is about in Aggregator during meta-packaging re-packaging, there is explicitly a step to "Specify what pri file" is need.
See the fix here: https://dev.azure.com/microsoft/ProjectReunion/_git/WindowsAppSDKAggregator/pullrequest/13215819

What I talked with Xiang last week is that given this is FIRST time having localization, we should:
a. Ensure the code (without localization) are still usable if there is any failure for whatever reason that unpredictable.
That's why this PR is here.
b. We can remove those Try-catch-logging later on, when we feel confident given loading localization resource should be reliable enough on running smooth.

Scottj1s:
> > Can we hold off on this until we determine root cause of the missing WAR PRI file in FWP-dependent deployments? This mitigation may not be necessary and would be ideal to avoid it.
> 
> @Scottj1s , Xiang know about it, it is about in Aggregator during meta-packaging re-packaging, there is explicitly a step to "Specify what pri file" is need. See the fix here: https://dev.azure.com/microsoft/ProjectReunion/_git/WindowsAppSDKAggregator/pullrequest/13215819
> 
> What I talked with Xiang last week is that given this is FIRST time having localization, we should: a. Ensure the code (without localization) are still usable if there is any failure for whatever reason that unpredictable. That's why this PR is here. b. We can remove those Try-catch-logging later on, when we feel confident given loading localization resource should be reliable enough on running smooth.

My understanding from Xiang was that even with the Aggregator fix which you linked, the pri file was still not being found, hence this fallback fix.  @Hong-Xiang can you clarify the current state, including the Aggregator fix?  Why are things currently "unpredictable"?

Hong-Xiang:
Closing this PR to reduce confusion, we will track fix on #5582

PR #5582, opened 2025-07-01
[Fix] Locating Microsoft.WindowsAppRuntime.pri in framework usage

Hong-Xiang:
In framework deployment scenario, the Microsoft.WindowsAppRuntime.pri is installed in the WinodowsAppRuntime's msix installation folder, not the folder of app using WinAppSDK, the resource manager need to handle that case.
Current code is copy from WinUI's code (with minimum modification) handling this same problem from
* https://github.com/microsoft/microsoft-ui-xaml/blob/main/src/controls/dev/dll/SharedHelpers.cpp#L77-L158
* [microsoft-ui-xaml/src/controls/dev/ResourceHelper/ResourceAccessor.cpp at main · microsoft/microsoft-ui-xaml](https://github.com/microsoft/microsoft-ui-xaml/blob/main/src/controls/dev/ResourceHelper/ResourceAccessor.cpp#L32-L40)

With this fix localization would work in both self-contained and framework sceario.

### DIFF
--- a/dev/Interop/StoragePickers/PickerLocalization.cpp
+++ b/dev/Interop/StoragePickers/PickerLocalization.cpp
@@ -9,30 +9,41 @@
 #include "PickerLocalization.h"
 #include <winrt\base.h>
 #include <winrt\Microsoft.Windows.ApplicationModel.Resources.h>
+#include <vector>
 
 
 const winrt::hstring priFileName = L"Microsoft.WindowsAppRuntime.pri";
 winrt::hstring GetPriFilePath()
 {
     HMODULE hModule = nullptr;
+    // use GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT, the module handle does not need to be released
+    // GetPriFilePath should only be called once to initialize the static priFilePath variable, thus thread safe
     if (GetModuleHandleExW(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS | GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT,
         reinterpret_cast<LPCWSTR>(&GetPriFilePath), &hModule))
     {
-        wchar_t path[MAX_PATH];
-        if (GetModuleFileNameW(hModule, path, MAX_PATH) > 0)
+        auto buffer = std::make_unique<wchar_t[]>(MAX_PATH + 1);
+        auto pathLength = GetModuleFileName(hModule, buffer.get(), MAX_PATH + 1);
+        if (pathLength > 0)
         {
-            std::filesystem::path dllPath(path);
+            auto dllPathString = std::wstring(buffer.get(), pathLength);
+            std::filesystem::path dllPath(dllPathString);
+            if (!std::filesystem::is_regular_file(dllPath))
+            {
+                return priFileName;
+            }
             std::filesystem::path parentDir = dllPath.parent_path();
             auto priFilePath = parentDir / std::wstring{ priFileName };
-            auto priFilePathHstring = winrt::hstring{ priFilePath.wstring() };
-            return priFilePathHstring;
+            if (std::filesystem::exists(priFilePath))
+            {
+                return winrt::hstring{ priFilePath.wstring() };
+            }
         }
     }
     return priFileName;
 }
 
 namespace PickerLocalization {
-    
+
     winrt::hstring GetStoragePickersLocalizationText(winrt::hstring key, winrt::hstring fallback)
     {
         static winrt::hstring priPath = GetPriFilePath();
--- a/dev/Interop/StoragePickers/PickerLocalization.cpp
+++ b/dev/Interop/StoragePickers/PickerLocalization.cpp
@@ -9,13 +9,13 @@
 #include "PickerLocalization.h"
 #include <winrt\base.h>
 #include <winrt\Microsoft.Windows.ApplicationModel.Resources.h>
-#include "StoragePickersTelemetry.h"
 
 namespace PickerLocalization {
     const winrt::hstring priPath = L"Microsoft.WindowsAppRuntime.pri";
     winrt::hstring GetStoragePickersLocalizationText(winrt::hstring key, winrt::hstring fallback)
     {
-        // adding try-catch to prevent localization error break picker experience
+        // adding try-catch to prevent localization error break picker experience on first shipping
+        // TODO: remove try-catch after stabilization period
         try
         {
             auto manager = winrt::Microsoft::Windows::ApplicationModel::Resources::ResourceManager(priPath);
@@ -23,7 +23,7 @@ namespace PickerLocalization {
         }
         catch (...)
         {
-            StoragePickersTelemetry::StoragePickerLocalizationLookupError(key.c_str());
+            LOG_CAUGHT_EXCEPTION();
             return fallback;
         }
     }
--- a/dev/Interop/StoragePickers/PickerLocalization.cpp
+++ b/dev/Interop/StoragePickers/PickerLocalization.cpp
@@ -9,7 +9,7 @@
 #include "PickerLocalization.h"
 #include <winrt\base.h>
 #include <winrt\Microsoft.Windows.ApplicationModel.Resources.h>
-#include <vector>
+#include <array>
 
 
 const winrt::hstring priFileName = L"Microsoft.WindowsAppRuntime.pri";
@@ -21,11 +21,11 @@ winrt::hstring GetPriFilePath()
     if (GetModuleHandleExW(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS | GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT,
         reinterpret_cast<LPCWSTR>(&GetPriFilePath), &hModule))
     {
-        auto buffer = std::make_unique<wchar_t[]>(MAX_PATH + 1);
-        auto pathLength = GetModuleFileName(hModule, buffer.get(), MAX_PATH + 1);
+        std::array<wchar_t, MAX_PATH + 1> buffer;
+        auto pathLength = GetModuleFileName(hModule, buffer.data(), MAX_PATH + 1);
         if (pathLength > 0)
         {
-            auto dllPathString = std::wstring(buffer.get(), pathLength);
+            auto dllPathString = std::wstring(buffer.data(), pathLength);
             std::filesystem::path dllPath(dllPathString);
             if (!std::filesystem::is_regular_file(dllPath))
             {
--- a/dev/Interop/StoragePickers/PickerLocalization.cpp
+++ b/dev/Interop/StoragePickers/PickerLocalization.cpp
@@ -9,106 +9,6 @@
 #include "PickerLocalization.h"
 #include <winrt\base.h>
 #include <winrt\Microsoft.Windows.ApplicationModel.Resources.h>
-#include <iostream>
-
-// find framework package related code from https://github.com/microsoft/microsoft-ui-xaml/blob/main/src/controls/dev/ResourceHelper/ResourceAccessor.cpp
-// and https://github.com/microsoft/microsoft-ui-xaml/blob/main/src/controls/dev/dll/SharedHelpers.cpp
-
-namespace {
-    constexpr wchar_t c_reunionPackageNamePrefix[] = L"Microsoft.WindowsAppRuntime";
-    constexpr int c_reunionPackageNamePrefixLength = ARRAYSIZE(c_reunionPackageNamePrefix) - 1;
-
-    // Tries to retrieve the current package graph. Will return true if a package graph exists
-    // and is non-empty, else false.
-    bool TryGetCurrentPackageGraph(
-        const std::uint32_t flags,
-        std::uint32_t& packageCount,
-        const PACKAGE_INFO*& packageGraph,
-        std::unique_ptr<BYTE[]>& buffer)
-    {
-        std::uint32_t bufferLength{};
-        LONG rc{ ::GetCurrentPackageInfo(flags, &bufferLength, nullptr, nullptr) };
-        if ((rc == APPMODEL_ERROR_NO_PACKAGE) || (rc == ERROR_SUCCESS))
-        {
-            // No/empty package graph
-            return false;
-        }
-        else if (rc != ERROR_INSUFFICIENT_BUFFER)
-        {
-            if (FAILED(HRESULT_FROM_WIN32(rc)))
-            {
-                return false;
-            }
-        }
-
-        buffer = std::make_unique<BYTE[]>(bufferLength);
-        if (SUCCEEDED(HRESULT_FROM_WIN32(::GetCurrentPackageInfo(flags, &bufferLength, buffer.get(), &packageCount))))
-        {
-            packageGraph = reinterpret_cast<PACKAGE_INFO*>(buffer.get());
-            return true;
-        }
-
-        return false;
-    }
-
-    bool IsInFrameworkPackageImpl(winrt::hstring& frameworkInstallLocation)
-    {
-        // We could be running in an app that is using WinUI "In-App" (i.e. the WinUI dlls are deployed as part of the app). 
-        // Or we could be running in the Project Reunion Framework Package that an app takes a dependency on.
-        // We determine if we are in a Framework Package by looking at the App's package dependencies. If the app has a Project Reunion dependency
-        // then we must be running in a Framework Package.
-        const UINT32 c_filter{ PACKAGE_FILTER_HEAD | PACKAGE_FILTER_DIRECT | PACKAGE_FILTER_STATIC | PACKAGE_FILTER_DYNAMIC | PACKAGE_INFORMATION_BASIC };
-        std::uint32_t packageCount{};
-        const PACKAGE_INFO* packageGraph{};
-        std::unique_ptr<BYTE[]> packageBuffer;
-        std::cout << "calling is in framework package" << std::endl;
-        if (TryGetCurrentPackageGraph(c_filter, packageCount, packageGraph, packageBuffer))
-        {
-            if (packageGraph)
-            {
-                for (std::uint32_t index = 0; index < packageCount; index++)
-                {
-                    const PACKAGE_INFO& packageInfo{ packageGraph[index] };
-
-                    // We use FamilyName here and not DisplayName because DisplayName is a localized string, but FamilyName is a stable identifier.
-                    auto packageFamilyName = packageInfo.packageFamilyName;
-                    int nameLength{ static_cast<int>(wcslen(packageFamilyName)) };
-
-                    if (nameLength < c_reunionPackageNamePrefixLength)
-                    {
-                        continue;
-                    }
-
-                    if (
-                       CompareStringOrdinal(
-                           packageFamilyName,
-                           c_reunionPackageNamePrefixLength,
-                           c_reunionPackageNamePrefix,
-                           c_reunionPackageNamePrefixLength,
-                           TRUE) == CSTR_EQUAL)
-                    {
-                        frameworkInstallLocation = winrt::hstring(packageInfo.path);
-                        std::cout << "calling is in framework package result true" << std::endl;
-                        return true;
-                    }
-                }
-            }
-        }
-        std::cout << "calling is in framework package result false" << std::endl;
-        return false;
-    }
-}
-
-
-
-
-bool PickerLocalization::IsInFrameworkPackage(winrt::hstring& frameworkPackageInstallLocation)
-{
-    static winrt::hstring s_frameworkInstallLocation;
-    static bool isInFrameworkPackage = IsInFrameworkPackageImpl(s_frameworkInstallLocation);    
-    frameworkPackageInstallLocation = s_frameworkInstallLocation;
-    return isInFrameworkPackage;
-}
 
 
 const winrt::hstring priFileName = L"Microsoft.WindowsAppRuntime.pri";
@@ -121,15 +21,10 @@ winrt::hstring GetPriFilePath()
         wchar_t path[MAX_PATH];
         if (GetModuleFileNameW(hModule, path, MAX_PATH) > 0)
         {
-            // Use std::filesystem to get the parent directory
             std::filesystem::path dllPath(path);
             std::filesystem::path parentDir = dllPath.parent_path();
-            // Combine parentDir with Microsoft.WindowsAppRuntime.pri
             auto priFilePath = parentDir / std::wstring{ priFileName };
-            auto pathHstring = winrt::hstring{ parentDir.wstring() };
             auto priFilePathHstring = winrt::hstring{ priFilePath.wstring() };
-            std::wcout << L"Parent directory: " << pathHstring.c_str() << std::endl;
-            std::wcout << L"PRI file path: " << priFilePathHstring.c_str() << std::endl;
             return priFilePathHstring;
         }
     }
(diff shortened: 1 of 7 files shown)


dev/Interop/StoragePickers/PickerLocalization.cpp
```diff
@@ -9,14 +9,119 @@
...
+            packageGraph = reinterpret_cast<PACKAGE_INFO*>(buffer.get());
+            return true;
+        }
+
+        return false;
+    }
+
+    bool IsInFrameworkPackageImpl(winrt::hstring& frameworkInstallLocation)
+    {
+        // We could be running in an app that is using WinUI "In-App" (i.e. the WinUI dlls are deployed as part of the app).
```
Scottj1s:
```suggestion
        // We could be running in an app that is using WinAppSDK Self-Contained (i.e. the WinAppSDK dlls are deployed as part of the app). 
```

dev/Interop/StoragePickers/PickerLocalization.cpp
```diff
@@ -9,14 +9,119 @@
...
 #include <winrt\Microsoft.Windows.ApplicationModel.Resources.h>
-#include <iostream>
 
+// find framework package related code from https://github.com/microsoft/microsoft-ui-xaml/blob/main/src/controls/dev/ResourceHelper/ResourceAccessor.cpp
+// and https://github.com/microsoft/microsoft-ui-xaml/blob/main/src/controls/dev/dll/SharedHelpers.cpp
+
+namespace {
+    constexpr wchar_t c_reunionPackageNamePrefix[] = L"Microsoft.WindowsAppRuntime";
+    constexpr int c_reunionPackageNamePrefixLength = ARRAYSIZE(c_reunionPackageNamePrefix) - 1;
+
```
Scottj1s: ProjectReunion is the old name - let's use WindowsAppSDK for build time considerations and WindowsAppRuntime for runtime.
```suggestion
    constexpr wchar_t c_WindowsAppRuntimePackageNamePrefix[] = L"Microsoft.WindowsAppRuntime";
    constexpr int c_WindowsAppRuntimePackageNamePrefixLength = ARRAYSIZE(c_WindowsAppRuntimePackageNamePrefix) - 1;

```

dev/Interop/StoragePickers/PickerLocalization.cpp
```diff
@@ -9,14 +9,119 @@
...
+                    {
+                        continue;
+                    }
+
+                    if (
+                       CompareStringOrdinal(
+                           packageFamilyName,
+                           c_reunionPackageNamePrefixLength,
+                           c_reunionPackageNamePrefix,
+                           c_reunionPackageNamePrefixLength,
```
Scottj1s:
```suggestion
        // Or we could be running in the WindowsAppRuntime Framework Package that an app takes a dependency on.
        // We determine if we are in a Framework Package by looking at the App's package dependencies. If the app has a WindowsAppRuntime dependency
        // then we must be running in a Framework Package.
        const UINT32 c_filter{ PACKAGE_FILTER_HEAD | PACKAGE_FILTER_DIRECT | PACKAGE_FILTER_STATIC | PACKAGE_FILTER_DYNAMIC | PACKAGE_INFORMATION_BASIC };
        std::uint32_t packageCount{};
        const PACKAGE_INFO* packageGraph{};
        std::unique_ptr<BYTE[]> packageBuffer;
        if (TryGetCurrentPackageGraph(c_filter, packageCount, packageGraph, packageBuffer))
        {
            if (packageGraph)
            {
                for (std::uint32_t index = 0; index < packageCount; index++)
                {
                    const PACKAGE_INFO& packageInfo{ packageGraph[index] };

                    // We use FamilyName here and not DisplayName because DisplayName is a localized string, but FamilyName is a stable identifier.
                    auto packageFamilyName = packageInfo.packageFamilyName;
                    int nameLength{ static_cast<int>(wcslen(packageFamilyName)) };

                    if (nameLength < c_reunionPackageNamePrefixLength)
                    {
                        continue;
                    }

                    if (
                       CompareStringOrdinal(
                           packageFamilyName,
                           c_WindowsAppRuntimePackageNamePrefixLength,
                           c_WindowsAppRuntimePackageNamePrefix,
                           c_WindowsAppRuntimePackageNamePrefixLength,
```

dev/Interop/StoragePickers/PickerLocalization.h
```diff
@@ -5,6 +5,8 @@
 #include "PickerCommon.h"
 
 namespace PickerLocalization {
-    winrt::hstring GetStoragePickersLocalizationText(winrt::hstring key);
+    winrt::hstring GetStoragePickersLocalizationText(winrt::hstring key, winrt::hstring fallback);
```
Scottj1s: it's still not clear to me why we need a fallback, if we're confident we've fixed root cause of the missing pri.

dev/Interop/StoragePickers/PickerLocalization.cpp
```diff
@@ -9,14 +9,119 @@
...
+                }
+            }
+        }
+        return false;
+    }
+}
+bool PickerLocalization::IsInFrameworkPackage(winrt::hstring& frameworkPackageInstallLocation)
+{
+    static winrt::hstring s_frameworkInstallLocation;
+    static bool isInFrameworkPackage = IsInFrameworkPackageImpl(s_frameworkInstallLocation);
```
Scottj1s: Have we tested with unpackaged, FWP-dependent apps (DynDeps)?

dev/Interop/StoragePickers/PickerLocalization.cpp
codendone: Also: Change `WinUI` to `WinAppSDK` (in both references)

dev/Interop/StoragePickers/PickerLocalization.cpp
Hong-Xiang: Hi @Scottj1s, thank you very much for your comment, it starts my deep dive over the past weeks into MSIX framework packages, dynamic dependencies, and testing with modified MSIX setups. I’ve resolved many long-standing questions around framework-dependent and bootstrap scenarios during learning those concepts!

The strange behavior is relates to the GetCurrentPackageInfo API. 
According to the documentation, it should return APPMODEL_ERROR_NO_PACKAGE in unpackaged scenarios. Here's what I observed:

* When using an unpackaged app, directly calling GetCurrentPackageInfo correctly returns no package info as expected.
* When a custom DLL (which calls GetCurrentPackageInfo) is packed into a framework MSIX and installed, and then used by an unpackaged app via dynamic dependency APIs like PackageDependency.Create, the API still behaves as expected—no package info is returned, since it's process-based.
* However, when using an unpackaged app with the WinAppSDK framework, calling the same APIs from WindowsAppRuntime.dll does resolve the package path successfully, which is strange, although since WinUI is using similar code, it's successfully resolution is some how expected, and meet my previous tests. It seems WinAppSDK’s framework dependency behaves differently from a custom msix framework package.

Since I can't really understand the special behavior, and to simplify things, I’ve changed to a new implementation to locate the WindowsAppRuntime.dll directly, since our method is compiled into it. This way, we can reliably find the PRI file in the same folder, regardless of whether it's a framework-dependent or self-contained scenario. This approach avoids hardcoding paths and ensures the WindowsAppRuntime.dll is already loaded since it is inside calling GetPriFilePath which is a method in WindowsAppRuntime.dll itself.

I’ve tested this in both component and dynamic unpackaged framework dependency scenarios. 

Would you mind reviewing this new method for resolving the PRI file location? Thanks again!

dev/Interop/StoragePickers/PickerLocalization.cpp
```diff
@@ -9,14 +9,119 @@
...
+// find framework package related code from https://github.com/microsoft/microsoft-ui-xaml/blob/main/src/controls/dev/ResourceHelper/ResourceAccessor.cpp
+// and https://github.com/microsoft/microsoft-ui-xaml/blob/main/src/controls/dev/dll/SharedHelpers.cpp
+
+namespace {
+    constexpr wchar_t c_reunionPackageNamePrefix[] = L"Microsoft.WindowsAppRuntime";
+    constexpr int c_reunionPackageNamePrefixLength = ARRAYSIZE(c_reunionPackageNamePrefix) - 1;
+
+    // Tries to retrieve the current package graph. Will return true if a package graph exists
+    // and is non-empty, else false.
+    bool TryGetCurrentPackageGraph(
```
DrusTheAxe: Delete this function and use the existing `AppModel::PackageGraph::GetCurrentPackageGraph()` in `dev\Common\AppModel.PackageGraph.h`

(you're not the first to need this. Not even the 2nd :P)

dev/Interop/StoragePickers/PickerLocalization.cpp
```diff
@@ -9,14 +9,119 @@
...
+    {
+        // We could be running in an app that is using WinUI "In-App" (i.e. the WinUI dlls are deployed as part of the app). 
+        // Or we could be running in the Project Reunion Framework Package that an app takes a dependency on.
+        // We determine if we are in a Framework Package by looking at the App's package dependencies. If the app has a Project Reunion dependency
+        // then we must be running in a Framework Package.
+        const UINT32 c_filter{ PACKAGE_FILTER_HEAD | PACKAGE_FILTER_DIRECT | PACKAGE_FILTER_STATIC | PACKAGE_FILTER_DYNAMIC | PACKAGE_INFORMATION_BASIC };
+        std::uint32_t packageCount{};
+        const PACKAGE_INFO* packageGraph{};
+        std::unique_ptr<BYTE[]> packageBuffer;
+        if (TryGetCurrentPackageGraph(c_filter, packageCount, packageGraph, packageBuffer))
```
DrusTheAxe: Replace with call to `AppModel::PackageGraph::GetCurrentPackageGraph()`

See earlier comment

dev/Interop/StoragePickers/PickerLocalization.cpp
```diff
@@ -9,14 +9,119 @@
...
+        if (SUCCEEDED(HRESULT_FROM_WIN32(::GetCurrentPackageInfo(flags, &bufferLength, buffer.get(), &packageCount))))
+        {
+            packageGraph = reinterpret_cast<PACKAGE_INFO*>(buffer.get());
+            return true;
+        }
+
+        return false;
+    }
+
+    bool IsInFrameworkPackageImpl(winrt::hstring& frameworkInstallLocation)
```
DrusTheAxe: The purpose here is to find the WinAppSDK framework package in the current process' package graph (if present)?

This logic already largely exists in WinAppSDK. Add this to `dev\Common\WindowsAppRuntime.PackageGraph.h`

```
namespace WindowsAppRuntime::PackageGraph
{
/// Return the package information for the Windows App Runtime runtime (framework) package in the current process' package graph (if any)
/// @note This is a variant of WindowsAppRuntime_IsSelfContained() but returns package details instead of mere existence.
///       You can think of WindowsAppRuntime_IsSelfContained() as a misnomer for `bool WindowsAppRuntime::PackageGraph::IsFrameworkPresent()`.
/// @param packageGraphCount number of packages in packageGraph.
/// @param packageGraph current packageGraph.
/// @return pointer to the Windows App SDK runtime (framework) package in the package graph, or NULL if not found.
inline PACKAGE_INFO* GetFramework(uint32_t& packageGraphCount, wil::unique_cotaskmem_ptr<PACKAGE_INFO>& packageGraph)
{
    const UINT32 flags{ PACKAGE_FILTER_HEAD | PACKAGE_FILTER_DIRECT | PACKAGE_FILTER_STATIC | PACKAGE_FILTER_DYNAMIC | PACKAGE_INFORMATION_BASIC };
    uint32_t packageGraphCount{};
    const PACKAGE_INFO* packageInfo{};
    RETURN_IF_FAILED(::AppModel::PackageGraph::GetCurrentPackageGraph(flags, packageGraphCount, packageInfo, packageGraph));
    const auto frameworkPackageFamilyName{ ::WindowsAppRuntime::VersionInfo::Framework::GetPackageFamilyName() };
    for (uint32_t index=0; index < packageGraphCount; ++index)
    {
        if (CompareStringOrdinal(packageInfo[index].packageFamilyName, -1, frameworkPackageFamilyName.c_str(), -1, TRUE) == CSTR_EQUAL)
        {
            // Found the Windows App SDK framework package in the package graph
            return packageInfo + index;
        }
    }

    // Didn't find the Windows App SDK framework package in the package graph
    packageGraphCount = 0;
    packageGraph.reset();
    return nullptr;
}

/// Return the package information for the Windows App Runtime runtime (framework) package in the current process' package graph (if any)
/// @note This is a variant of WindowsAppRuntime_IsSelfContained() but returns package details instead of mere existence.
///       You can think of WindowsAppRuntime_IsSelfContained() as a misnomer for `bool WindowsAppRuntime::PackageGraph::IsFrameworkPresent()`.
/// @param buffer memory for the package graph information.
/// @return pointer to the Windows App SDK runtime (framework) package in the package graph, or NULL if not found.
inline PACKAGE_INFO* GetFramework(wil::unique_cotaskmem_ptr<PACKAGE_INFO>& buffer)
{
    uint32_t packageGraphCount{};
    return GetFramework(packageGraphCount, buffer);
}
}
```

dev/Interop/StoragePickers/PickerLocalization.cpp
DrusTheAxe: Moot. Replace all this with GetFramework() per comment below

dev/Interop/StoragePickers/PickerLocalization.cpp
DrusTheAxe: Moot. Replace all this with GetFramework() per comment below

dev/Interop/StoragePickers/PickerLocalization.cpp
DrusTheAxe: Moot. Replace all this with GetFramework() per comment below

dev/Interop/StoragePickers/PickerLocalization.cpp
DrusTheAxe: Replace all this with GetFramework() per my comment

DrusTheAxe (changes requested):

dev/Interop/StoragePickers/PickerLocalization.cpp
```diff
@@ -9,14 +9,47 @@
 #include "PickerLocalization.h"
 #include <winrt\base.h>
 #include <winrt\Microsoft.Windows.ApplicationModel.Resources.h>
-#include <iostream>
+#include <array>
+
+
```
DrusTheAxe: Remove extra blank line

dev/Interop/StoragePickers/PickerLocalization.cpp
```diff
@@ -9,14 +9,47 @@
...
+const winrt::hstring priFileName = L"Microsoft.WindowsAppRuntime.pri";
```
DrusTheAxe: Don't declare global constructor as it adds to load time perf hit for anyone loading the DLL, whether or not they ever using this functionality

dev/Interop/StoragePickers/PickerLocalization.cpp
```diff
@@ -9,14 +9,47 @@
...
+        std::array<wchar_t, MAX_PATH + 1> buffer;
+        auto pathLength = GetModuleFileName(hModule, buffer.data(), MAX_PATH + 1);
+        if (pathLength > 0)
+        {
+            auto dllPathString = std::wstring(buffer.data(), pathLength);
+            std::filesystem::path dllPath(dllPathString);
+            if (!std::filesystem::is_regular_file(dllPath))
+            {
+                return priFileName;
+            }
```
DrusTheAxe: Use WIL, not raw Win32 APIs

Replace lines 15-43 using a variant of `DeploymentManager::GenerateDeploymentAgentPath()`

```
winrt::hstring GetPriFilePath()
{
    wil::unique_hmodule module;
    THROW_IF_WIN32_BOOL_FALSE(GetModuleHandleEx(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS, reinterpret_cast<PCWSTR>(&GetPriFilePath), &module));

    std::filesystem::path modulePath{ wil::GetModuleFileNameW<std::wstring>(module.get()) };
    const auto priFilename{ modulePath.parent_path() / priFileName };
    if (std::filesystem::exists(priFilePath))
    {
        return winrt::hstring{ priFilename.c_str() };
    }

    PCWSTR c_WindowsAppRuntimePriFilename{ L"Microsoft.WindowsAppRuntime.pri" };
    return winrt::hstring{ c_WindowsAppRuntimePriFilename };
}
```

dev/Interop/StoragePickers/PickerLocalization.cpp
```diff
@@ -9,14 +9,47 @@
...
+    }
+    return priFileName;
+}
 
 namespace PickerLocalization {
-    const winrt::hstring priPath = L"Microsoft.WindowsAppRuntime.pri";
+
     winrt::hstring GetStoragePickersLocalizationText(winrt::hstring key)
     {
+        static winrt::hstring priPath = GetPriFilePath();
```
DrusTheAxe: NIT: `static winrt::hstring priPath{ GetPriFilePath() };`

dev/Interop/StoragePickers/PickerLocalization.cpp
DrusTheAxe: >However, when using an unpackaged app with the WinAppSDK framework, calling the same APIs from WindowsAppRuntime.dll does resolve the package path successfully, which is strange

Not so strange. Unpackaged processes using Dynamic Dependencies work as expected. On Ge that leverages the OS DynDep APIs and infrastructure (as of 1.7), and downlevel relies on WinAppSDK's DynDep functionality which includes altering several Windows APIs involving the package graph via Detours magic :-)

>It seems WinAppSDK’s framework dependency behaves differently from a custom msix framework package

From an unpackaged process or packaged? The latter typically use a static dependency declared in appxmanifest.xml which ultimately uses the machine core machinery (the package graph), but Windows has that baked in since Win8. It's the `unpackaged processes using packaged bits` that's relatively new

dev/Interop/StoragePickers/PickerLocalization.cpp
Hong-Xiang: Thank you for your suggestion, updated using initializer syntax.

dev/Interop/StoragePickers/PickerLocalization.cpp
Hong-Xiang: Thank you for your thorough review. We've updated the implementation to align with DeploymentManager's approach, and the related code has been removed accordingly.

dev/Interop/StoragePickers/PickerLocalization.cpp
Hong-Xiang: Thank you for your thorough review. We've updated the implementation to align with DeploymentManager's approach, and the related code has been removed accordingly.

dev/Interop/StoragePickers/PickerLocalization.cpp
Hong-Xiang: Thank you for your thorough review. We've updated the implementation to align with DeploymentManager's approach, and the related code has been removed accordingly.

dev/Interop/StoragePickers/PickerLocalization.h
Hong-Xiang: Thank you for your comment, we've removed fallback related code, and currently using a code similar to existing DeploymentManager

dev/Interop/StoragePickers/PickerLocalization.cpp
Hong-Xiang: Thank you for your thorough review. We've updated the implementation to align with DeploymentManager's approach, and the related code has been removed accordingly.

dev/Interop/StoragePickers/PickerLocalization.cpp
Hong-Xiang: Thank you for your thorough review. We've updated the implementation to align with DeploymentManager's approach, and the related code has been removed accordingly.

dev/Interop/StoragePickers/PickerLocalization.cpp
Hong-Xiang: Thank you for your thorough review. We've updated the implementation to align with DeploymentManager's approach, and the related code has been removed accordingly.

dev/Interop/StoragePickers/PickerLocalization.cpp
Hong-Xiang: Thank you for your carefully review! Related empty line was removed.

dev/Interop/StoragePickers/PickerLocalization.cpp
Hong-Xiang: Thank you for your guidance and suggestions! Having a reference implementation was really helpful. I've updated the code to closely follow the approach used in GenerateDeploymentAgentPath, using WIL libraries and hmodule life time management.

dev/Interop/StoragePickers/PickerLocalization.cpp
Hong-Xiang: Thank you for the suggestion! I've updated it to use a raw static char pointer, following the implementation in GenerateDeploymentAgentPath.

dev/Interop/StoragePickers/PickerLocalization.cpp
Hong-Xiang: Hi @DrusTheAxe  Thank you for your clarification! The test app is 
* unpackaged console app
* using winappsdk framework, which includes bootstrap code automatically
* uses `PackageDependency.Create` APIs to add dynamic package dependencies
What's puzzling is that within the same process, when my custom DLL (packaged in to a msix framework package) calls `GetCurrentPackageInfo`-related APIs, it returns no package. However, the same call from `WindowsAppRuntime.dll` successfully resolves the WinAppSDK package location. Is this expected behavior?

dev/Interop/StoragePickers/PickerLocalization.cpp
```diff
@@ -9,44 +9,20 @@
...
-            {
-                return winrt::hstring{ priFilePath.wstring() };
-            }
-        }
-    }
-    return priFileName;
-}
 
 namespace PickerLocalization {
+    static PCWSTR c_WindowsAppRuntimeLocalizationPRIFilename{ L"Microsoft.WindowsAppRuntime.pri" };
```
Scottj1s, line 16: Or consider:
```
// Linker-provided symbol for current module handle
EXTERN_C IMAGE_DOS_HEADER __ImageBase;
...
std::filesystem::path modulePath{ wil::GetModuleFileNameW<std::wstring>((HINSTANCE)&__ImageBase) };
```

dev/Interop/StoragePickers/PickerLocalization.cpp
DrusTheAxe: Please share the code (or link)

What version of WinAppSDK are you using?

What edition and version of Windows (Desktop 24H1, Server 2022, etc)?

>when my custom DLL (packaged in to a msix framework package) calls GetCurrentPackageInfo-related APIs, it returns no package. However, the same call from WindowsAppRuntime.dll successfully resolves the WinAppSDK package location. Is this expected behavior?

Depends on the version of WinAppSDK + Windows in play. If you're using WinAppSDK >=1.7 on >= Ge (24H2) then GetCurrentPackageInfo gives you the package graph (or subset you filtered by). I'd need to more about the environment and code to try to disagnose any further

dev/Interop/StoragePickers/PickerLocalization.cpp
Hong-Xiang: Thanks so much for the suggestions—felt like a ninja move! 
I checked WIL's library and found an existing method that wraps the exact logic using `__ImageBase`: [GetModuleInstanceHandle](https://github.com/microsoft/wil/blob/0250c6c328224aa623c569004874204dc96f1874/include/wil/win32_helpers.h#L788-L791) Now the final code is just:
```
std::filesystem::path modulePath{ wil::GetModuleFileNameW<std::wstring>(wil::GetModuleInstanceHandle()) };
```
Really happy to replace that long chunk of package graph API code with something so clean, just 6 lines include 2 lines for `{}`, and fully modern cpp wil code without need to worrying about lifetime management 🎉 .
Just updated code and testing it now.

Hong-Xiang:
Hi @DrusTheAxe, the code has been updated to use WIL APIs, and all calls to package graph-related APIs have been removed. Could you please check if there are still any requested changes?

dev/Interop/StoragePickers/PickerLocalization.cpp
Hong-Xiang: Thanks @DrusTheAxe for the clarification! 
I created a new repo to reproduce the issue and looked deeper into the msix framework package/package graph API usage. I can now get GetCurrentPackageInfo work correctly in an unpackaged console app on a Windows 11 desktop with WinAppSDK 1.7.250606001. It turns out I wasn’t using the API correctly before—apologies for the confusion.
Although we're not using this API in the current PR since we’ve opted for a simpler solution, I still learned a lot through this process. Really appreciate your helpful insights!

PR #5597, opened 2025-07-04
fix: MSB8027, two or more files with the same name

qiutongMS:
**Fix MSB4120 and MSB8027 warnings in MSBuild targets files

Problem:
MSBuild was generating multiple warnings when building the project:

1. MSB4120: "Item 'ClCompile' definition within target references itself via (qualified or unqualified) metadatum 'PreprocessorDefinitions'. This can lead to unintended expansion and cross-applying of pre-existing items."

2. MSB8027: Multiple targets were being executed repeatedly due to cross-target interference caused by the self-referencing item definitions.

Root Cause:
The issue occurred because multiple PreprocessorDefinitions were being set conditionally within ClCompile items inside targets, each referencing %(PreprocessorDefinitions) to append to existing definitions. This created several problems:

1. **Self-reference pattern**: MSBuild flags this as potentially problematic due to unexpected behavior during item evaluation and expansion
2. **Cross-target interference**: The repeated references to %(PreprocessorDefinitions) caused targets to influence each other, leading MSBuild to re-execute the same target multiple times
3. **Build performance impact**: Redundant target execution slowed down the build process
**

WARNING:

### DIFF
--- a/build/NuSpecs/WindowsAppSDK-Nuget-Native.AutoInitializer.targets
+++ b/build/NuSpecs/WindowsAppSDK-Nuget-Native.AutoInitializer.targets
@@ -1,21 +1,28 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-
-  <Target Name="WindowsAppRuntimeAutoInitializer">
-    <!-- Build the preprocessor definitions based on conditions -->
+  
+  <Target Name="SetAutoInitializerPreprocessorDefinitions" BeforeTargets="BeforeClCompile" >
+    <!-- Build the AutoInitializer preprocessor definitions based on conditions -->
     <PropertyGroup>
       <AutoInitializerPreprocessorDefinitions></AutoInitializerPreprocessorDefinitions>
       <AutoInitializerPreprocessorDefinitions Condition="'$(WindowsAppSdkBootstrapInitialize)'=='true'">$(AutoInitializerPreprocessorDefinitions);MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_BOOTSTRAP</AutoInitializerPreprocessorDefinitions>
       <AutoInitializerPreprocessorDefinitions Condition="'$(WindowsAppSdkDeploymentManagerInitialize)'=='true'">$(AutoInitializerPreprocessorDefinitions);MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_DEPLOYMENTMANAGER</AutoInitializerPreprocessorDefinitions>
       <AutoInitializerPreprocessorDefinitions Condition="'$(WindowsAppSdkUndockedRegFreeWinRTInitialize)'=='true'">$(AutoInitializerPreprocessorDefinitions);MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_UNDOCKEDREGFREEWINRT</AutoInitializerPreprocessorDefinitions>
       <AutoInitializerPreprocessorDefinitions Condition="'$(WindowsAppSdkCompatibilityInitialize)'=='true'">$(AutoInitializerPreprocessorDefinitions);MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_COMPATIBILITY</AutoInitializerPreprocessorDefinitions>
+      <!-- Inherit the previous definitions based on condition -->
+      <AutoInitializerPreprocessorDefinitions Condition="'$(PreProcessorDefinitionsInheritance)'!='true'">$(AutoInitializerPreprocessorDefinitions);%(ClCompile.PreprocessorDefinitions)</AutoInitializerPreprocessorDefinitions>
+      <PreProcessorDefinitionsInheritance Condition="'$(PreProcessorDefinitionsInheritance)'!='true'">true</PreProcessorDefinitionsInheritance>
+      <!-- Set PreprocessorDefinitions final value for following target -->
+      <PreprocessorDefinitions>$(AutoInitializerPreprocessorDefinitions);$(PreprocessorDefinitions)</PreprocessorDefinitions>
     </PropertyGroup>
-    
+  </Target>
+
+  <Target Name="WindowsAppRuntimeAutoInitializer">
     <ItemGroup>
       <ClCompile Include="$(MSBuildThisFileDirectory)..\..\include\WindowsAppRuntimeAutoInitializer.cpp">
         <PrecompiledHeader>NotUsing</PrecompiledHeader>
-        <PreprocessorDefinitions>$(AutoInitializerPreprocessorDefinitions)</PreprocessorDefinitions>
+        <PreprocessorDefinitions>$(PreprocessorDefinitions)</PreprocessorDefinitions>
       </ClCompile>
     </ItemGroup>
   </Target>
--- a/build/NuSpecs/WindowsAppSDK-Nuget-Native.AutoInitializer.targets
+++ b/build/NuSpecs/WindowsAppSDK-Nuget-Native.AutoInitializer.targets
@@ -3,13 +3,19 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <Target Name="WindowsAppRuntimeAutoInitializer">
+    <!-- Build the preprocessor definitions based on conditions -->
+    <PropertyGroup>
+      <AutoInitializerPreprocessorDefinitions></AutoInitializerPreprocessorDefinitions>
+      <AutoInitializerPreprocessorDefinitions Condition="'$(WindowsAppSdkBootstrapInitialize)'=='true'">$(AutoInitializerPreprocessorDefinitions);MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_BOOTSTRAP</AutoInitializerPreprocessorDefinitions>
+      <AutoInitializerPreprocessorDefinitions Condition="'$(WindowsAppSdkDeploymentManagerInitialize)'=='true'">$(AutoInitializerPreprocessorDefinitions);MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_DEPLOYMENTMANAGER</AutoInitializerPreprocessorDefinitions>
+      <AutoInitializerPreprocessorDefinitions Condition="'$(WindowsAppSdkUndockedRegFreeWinRTInitialize)'=='true'">$(AutoInitializerPreprocessorDefinitions);MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_UNDOCKEDREGFREEWINRT</AutoInitializerPreprocessorDefinitions>
+      <AutoInitializerPreprocessorDefinitions Condition="'$(WindowsAppSdkCompatibilityInitialize)'=='true'">$(AutoInitializerPreprocessorDefinitions);MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_COMPATIBILITY</AutoInitializerPreprocessorDefinitions>
+    </PropertyGroup>
+    
     <ItemGroup>
       <ClCompile Include="$(MSBuildThisFileDirectory)..\..\include\WindowsAppRuntimeAutoInitializer.cpp">
         <PrecompiledHeader>NotUsing</PrecompiledHeader>
-        <PreprocessorDefinitions Condition="'$(WindowsAppSdkBootstrapInitialize)'=='true'">MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_BOOTSTRAP;%(PreprocessorDefinitions)</PreprocessorDefinitions>
-        <PreprocessorDefinitions Condition="'$(WindowsAppSdkDeploymentManagerInitialize)'=='true'">MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_DEPLOYMENTMANAGER;%(PreprocessorDefinitions)</PreprocessorDefinitions>
-        <PreprocessorDefinitions Condition="'$(WindowsAppSdkUndockedRegFreeWinRTInitialize)'=='true'">MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_UNDOCKEDREGFREEWINRT;%(PreprocessorDefinitions)</PreprocessorDefinitions>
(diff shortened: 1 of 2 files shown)


build/NuSpecs/WindowsAppSDK-Nuget-Native.AutoInitializer.targets
```diff
@@ -3,13 +3,19 @@
...
+    </PropertyGroup>
+    
     <ItemGroup>
       <ClCompile Include="$(MSBuildThisFileDirectory)..\..\include\WindowsAppRuntimeAutoInitializer.cpp">
         <PrecompiledHeader>NotUsing</PrecompiledHeader>
-        <PreprocessorDefinitions Condition="'$(WindowsAppSdkBootstrapInitialize)'=='true'">MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_BOOTSTRAP;%(PreprocessorDefinitions)</PreprocessorDefinitions>
-        <PreprocessorDefinitions Condition="'$(WindowsAppSdkDeploymentManagerInitialize)'=='true'">MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_DEPLOYMENTMANAGER;%(PreprocessorDefinitions)</PreprocessorDefinitions>
-        <PreprocessorDefinitions Condition="'$(WindowsAppSdkUndockedRegFreeWinRTInitialize)'=='true'">MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_UNDOCKEDREGFREEWINRT;%(PreprocessorDefinitions)</PreprocessorDefinitions>
-        <PreprocessorDefinitions Condition="'$(WindowsAppSdkCompatibilityInitialize)'=='true'">MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_COMPATIBILITY;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+        <PreprocessorDefinitions>$(AutoInitializerPreprocessorDefinitions)</PreprocessorDefinitions>
```
codendone, line 22: Shouldn't this still include `;%(PreprocessorDefinitions)` to pick up any global definitions (such as building debug or release)? (Same in the other file.)

I'm unclear how (per the PR description) the PreprocessorDefinitions in these two targets files were influencing each other. Are they not both set relative to the single file of each `<ClCompile>`?

codendone: (Creating the new definitions in a property group makes sense, and that part looks good.)

qiutongMS: Good questions! Let me address this systematically:

**Why the warnings needed to be fixed:**

The MSB4120/MSB8027 warnings weren't just cosmetic - they indicated a real problem where preprocessor definitions from different targets were "leaking" into each other through the `%(PreprocessorDefinitions)` references. 

When Target A executes and adds definitions via `%(PreprocessorDefinitions)`, and then Target B executes with its own `%(PreprocessorDefinitions)` reference, Target B can inadvertently pick up definitions that Target A added. This creates unpredictable behavior and forces MSBuild to re-execute targets to try to reach a consistent state. We can see this clearly in the sample-app binlog in the aggregator build: 
image
image
Target `WindowsAppRuntimeAutoInitializer` added ClCompile twice, and the second one got an extra "MICROSOFT_WINDOWSAPPSDK_BOOTSTRAP_AUTO_INITIALIZE=1" which was added in Target `GenerateBootstrapCpp`.

**The fix:**

We moved the conditional logic into `PropertyGroup` elements to build the definitions deterministically, then apply them once to each `ClCompile` item. This eliminates the cross-target leakage and makes each target's behavior independent and predictable.

**Potential impact of the fix:**

You're absolutely right to question the removal of `%(PreprocessorDefinitions)`. Our current fix changes two behaviors:

1. **Global inheritance**: The auto-initializer files no longer inherit global preprocessor definitions (like `_DEBUG`, platform defines, etc.), but I didn't find any use in those two .cpp files.
2. **Definition propagation**: The WindowsAppSDK-specific definitions are now isolated to just the injected `.cpp` files, rather than propagating to subsequent compilation units. 

I notice that, in aggregator repo, there is a Todo, to move `MICROSOFT_WINDOWSAPPSDK_BOOTSTRAP_AUTO_INITIALIZE=1` to `windowsappsdk\build\NuSpecs\WindowsAppSDK-Nuget-Native.Bootstrap.targets`. Considering that the todo will be pushed forward, the modification in WindowsAppSDK-Nuget-Native.Bootstrap.targets should add `;%(PreprocessorDefinitions)`(which I did in the [commit](https://github.com/microsoft/WindowsAppSDK/pull/5597/commits/394f980474b03de17e564096d3645e71c4961b7b)); but the original modification in WindowsAppSDK-Nuget-Native.AutoInitializer.targets should be kept unchanged.

codendone: Why can't WindowsAppSDK-Nuget-Native.AutoInitializer.targets also add back `%(PreprocessorDefinitions)`? There isn't a clear need to include global definitions, but from a code perspective there also isn't a need to block global definitions, which may also be useful in the future for checks for things like `_DEBUG` (as you mentioned). Does adding back `%(PreprocessorDefinitions)` cause the problem again, or is the problem fixed by the new use of PropertyGroups to only set PreprocessorDefinitions once for each `<ClCompile>`?

codendone:
Why does the PR now show changes committed from other PRs?

qiutongMS:
> Why does the PR now show changes committed from other PRs?

Thanks for pointing that out! I had rebased the branch earlier, which temporarily pulled in commits from other PRs. I've since cleaned it up—now only my actual changes remain. Appreciate your review!

build/NuSpecs/WindowsAppSDK-Nuget-Native.AutoInitializer.targets
qiutongMS: Adding back` %(PreprocessorDefinitions)` in both files will cause the problem again.
This is because when two `<ClCompile>` targets are added to `<BeforeClCompileTargets>`, the `%(PreprocessorDefinitions`) can be considered as a placeholder that is continuously monitored by MSBuild. Therefore, when Target `GenerateBootstrapCpp` is executed first, MSBuild will detect the change of `PreprocessorDefinitions`. If there is still `%(PreprocessorDefinitions`) in the subsequent ClCompile tasks such as `WindowsAppRuntimeAutoInitializer`, it will be considered as two compile tasks that need to be added and executed repeatedly under different conditions, and cause the MSB8027(Two or more files with the name of WindowsAppRuntimeAutoInitializer.cpp will produce outputs to the same location) warning.
Creating the new definitions in a property group aimed at fixing MSB4120(Item 'ClCompile' definition within target references itself via (qualified or unqualified) metadatum 'PreprocessorDefinitions'. This can lead to unintended expansion and cross-applying of pre-existing items. More info: https://aka.ms/msbuild/metadata-self-ref) and making the code clearer.

codendone: Would a solution be to add `%(PreprocessorDefinitions)` as the default value of `AutoInitializerPreprocessorDefinitions`? Then the `PreprocessorDefinitions` in this file is never self-referencing. (That still assumes that setting `PreprocessorDefinitions` inside a `ClCompile` only affects the single source file specified in that ClCompile, though, which is seemingly not the case or there wouldn't have been the original issue here. Sounds like an msbuild bug.)

Anyway, thanks for talking this through. I'm fine with what you have.

codendone (approved):

qiutongMS:
Explain for the newest commit.
For each targets in the two files:
1. add a"SetXXXPreprocessorDefinitions" target with condition `BeforeTargets="BeforeClCompile"`, this condition will make the target executed before the original target
2. in the "SetXXXPreprocessorDefinitions" target, create the new definitions like "XXXPreprocessorDefinitions", which only contains definitions with conditions in the original target.
3.Defined a PreProcessorDefinitionsInheritance property, if it is not true, we need to append ClCompile.PreprocessorDefinitions to "XXXPreprocessorDefinitions", and set the property as true. This could make sure we got the definitions from previous jobs.
4.At the end of the "SetXXXPreprocessorDefinitions" target, `<PreprocessorDefinitions>$(XXXPreprocessorDefinitions);$(PreprocessorDefinitions)</PreprocessorDefinitions>` will change the PreprocessorDefinitions value, which will be used in the following targets.
In this way, all the worries we talked before can be solved.

build/NuSpecs/WindowsAppSDK-Nuget-Native.AutoInitializer.targets
```diff
@@ -1,15 +1,28 @@
...
+      <AutoInitializerPreprocessorDefinitions Condition="'$(WindowsAppSdkUndockedRegFreeWinRTInitialize)'=='true'">$(AutoInitializerPreprocessorDefinitions);MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_UNDOCKEDREGFREEWINRT</AutoInitializerPreprocessorDefinitions>
+      <AutoInitializerPreprocessorDefinitions Condition="'$(WindowsAppSdkCompatibilityInitialize)'=='true'">$(AutoInitializerPreprocessorDefinitions);MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_COMPATIBILITY</AutoInitializerPreprocessorDefinitions>
+      <!-- Inherit the previous definitions based on condition -->
+      <AutoInitializerPreprocessorDefinitions Condition="'$(PreProcessorDefinitionsInheritance)'!='true'">$(AutoInitializerPreprocessorDefinitions);%(ClCompile.PreprocessorDefinitions)</AutoInitializerPreprocessorDefinitions>
+      <PreProcessorDefinitionsInheritance Condition="'$(PreProcessorDefinitionsInheritance)'!='true'">true</PreProcessorDefinitionsInheritance>
+      <!-- Set PreprocessorDefinitions final value for following target -->
+      <PreprocessorDefinitions>$(AutoInitializerPreprocessorDefinitions);$(PreprocessorDefinitions)</PreprocessorDefinitions>
```
codendone: Why set `PreprocessorDefinitions` here? Because I'm not sure how msbuild does things, I'd be far more concerned about this polluting other targets than the "localized" setting inside the ClCompile below. I suggest removing this and keeping the change to directly use AutoInitializerPreprocessorDefinitions down in ClCompile.

build/NuSpecs/WindowsAppSDK-Nuget-Native.AutoInitializer.targets
```diff
@@ -1,15 +1,28 @@
...
+      <AutoInitializerPreprocessorDefinitions Condition="'$(WindowsAppSdkDeploymentManagerInitialize)'=='true'">$(AutoInitializerPreprocessorDefinitions);MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_DEPLOYMENTMANAGER</AutoInitializerPreprocessorDefinitions>
+      <AutoInitializerPreprocessorDefinitions Condition="'$(WindowsAppSdkUndockedRegFreeWinRTInitialize)'=='true'">$(AutoInitializerPreprocessorDefinitions);MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_UNDOCKEDREGFREEWINRT</AutoInitializerPreprocessorDefinitions>
+      <AutoInitializerPreprocessorDefinitions Condition="'$(WindowsAppSdkCompatibilityInitialize)'=='true'">$(AutoInitializerPreprocessorDefinitions);MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_COMPATIBILITY</AutoInitializerPreprocessorDefinitions>
+      <!-- Inherit the previous definitions based on condition -->
+      <AutoInitializerPreprocessorDefinitions Condition="'$(PreProcessorDefinitionsInheritance)'!='true'">$(AutoInitializerPreprocessorDefinitions);%(ClCompile.PreprocessorDefinitions)</AutoInitializerPreprocessorDefinitions>
+      <PreProcessorDefinitionsInheritance Condition="'$(PreProcessorDefinitionsInheritance)'!='true'">true</PreProcessorDefinitionsInheritance>
```
codendone: PreProcessorDefinitionsInheritance set here in a .targets won't interfere with the same property name in the other .targets? Should we make the name unique to ensure there isn't any interference?

build/NuSpecs/WindowsAppSDK-Nuget-Native.AutoInitializer.targets
qiutongMS: Thanks for the suggestion! You're absolutely right—setting PreprocessorDefinitions at the property level can affect other targets, which is something I should have been more cautious about.

The reason I initially placed it there is because I noticed some TODOs in the aggregator repo indicating that similar logic might eventually be moved up to this level, so I assumed this was the intended direction. Also, this change was meant to align with how the logic was handled in the previous version.
image

That said, if the goal is to scope the definition strictly to the current ClCompile target, I can definitely change it and apply it locally as suggested.

build/NuSpecs/WindowsAppSDK-Nuget-Native.AutoInitializer.targets
qiutongMS: Setting PreProcessorDefinitionsInheritance at the global level in a .targets file can interfere with the same property name used elsewhere. That interference is actually intentional in this case: the goal is to ensure that we only append the inherited definitions once, even if multiple targets run in other sequence.

While the current test conditions guarantee a clear execution order between the two targets, I wanted to make the logic more robust in case that assumption changes in the future.

build/NuSpecs/WindowsAppSDK-Nuget-Native.AutoInitializer.targets
codendone: The intent in both of these targets, at least, is to only set PreprocessorDefinitions for the single file specified in the ClCompile in that file. Given that, I recommend changing to a unique variable name like `AutoInitializerPreprocessorDefinitions` and using just that final and complete value down in the ClCompile. (And similar in the other file.)

build/NuSpecs/WindowsAppSDK-Nuget-Native.AutoInitializer.targets
codendone: Given that, it sounds like this should change from `PreProcessorDefinitionsInheritance` to something unique like `AutoInitializerPreProcessorDefinitionsInheritance`, to ensure this file maintains the intent of only inheriting once while also ensuring it doesn't affect any other targets. And same in the other file?

build/NuSpecs/WindowsAppSDK-Nuget-Native.AutoInitializer.targets
qiutongMS: This has been modified in the latest commit.

build/NuSpecs/WindowsAppSDK-Nuget-Native.AutoInitializer.targets
qiutongMS: Since we have already used unique 'XXXPreprocessorDefinitions' in each files, We can append directly without the Inheritance check(They both happen independently in their respective set target). So I delete the varibles.

PR #5634, opened 2025-07-18
Official API Review for Microsoft.Windows.Storage.Pickers

DinahK-2SO:
**Description**
This PR contains all design specs of the `Microsoft.Windows.Storage.Pickers` API set.
Please start with `specs/Storage.Pickers/Microsoft.Windows.Storage.Pickers.md`

**Note:**
This PR is created for the official API review of `Microsoft.Windows.Storage.Pickers` API set.

Since there're some documents on the main branch already, the files are copied from [dev/Interop/StoragePickers](https://github.com/microsoft/WindowsAppSDK/tree/main/dev/Interop/StoragePickers) and updated for official review.

After the review is completed, we will remove [dev/Interop/StoragePickers](https://github.com/microsoft/WindowsAppSDK/tree/main/dev/Interop/StoragePickers) and check in this officially reviewed version.


specs/Storage.Pickers/Microsoft.Windows.Storage.Pickers.md
```diff
@@ -0,0 +1,152 @@
...
+pattern.*
+1. The new pickers allow developers to use pickers without requiring `FileTypeFilter` or 
+`FileTypeChoices` to be specified. While UWP pickers throw exceptions when these properties are not 
+set, the new pickers default to show all file when developers don't explicitly configure these 
+attributes.
+1. *HomeGroup location has been excluded from the new `PickerLocationId` enum as it's no longer 
+supported in Windows 10 and later*
+1. *FolderPicker.FileTypeFilter has been exluded as it was not functionally appropriate for folder 
+selection*
+1. *Excluding methods and attributes that are already marked as deprecated, or "Do not use".
```
kmahone: >attributes [](http://example.com/codeflow?start=26&length=10)

Throughout this document and others in this PR the word 'attribute' is used instead of 'property'. 

Property is the correct term here. Attributes are something different in WinRT.

Update all usages of 'attribute' to 'property'. #Closed

specs/Storage.Pickers/Microsoft.Windows.Storage.Pickers.md
```diff
@@ -0,0 +1,152 @@
...
+
+        Windows.Foundation.IAsyncOperation<PickFileResult> PickSaveFileAsync();
+    }
+
+    runtimeclass FolderPicker
+    {
+        FolderPicker(Microsoft.UI.WindowId windowId);
+
+        string CommitButtonText;
+        IVector<string> FileTypeFilter{ get; };
```
oldnewthing: Presumably this should be deleted, given discussion above. #Closed

specs/Storage.Pickers/Microsoft.Windows.Storage.Pickers.md
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    runtimeclass FileOpenPicker
+    {
+        FileOpenPicker(Microsoft.UI.WindowId windowId);
+
+        string CommitButtonText;
+        string SettingsIdentifier;
+        IVector<string> FileTypeFilter{ get; };
+        PickerLocationId SuggestedStartLocation;
```
oldnewthing, line 112: I suspect a common request would be to allow the suggested start location to be a specific folder of the app's choosing. We offer this for the save picker (SuggestedSaveFilePath) but not the open or folder picker. #Closed

specs/Storage.Pickers/PickFileResult.md
```diff
@@ -0,0 +1,66 @@
+PickFileResult Class
+===
+
+# Background
+
+Namespace: [Microsoft.Windows.Storage.Pickers](./Microsoft.Windows.Storage.Pickers.md)
+
+Represents the result of a file picking operation. This is a lightweight class that contains a 
+string attribute representing the file path.
```
oldnewthing:
```suggestion
string property representing the file path.
``` #Closed

File: specs/Storage.Pickers/PickFolderResult.md [hunk 5]
```diff
@@ -0,0 +1,67 @@
...
+```C++
+#include <winrt/Microsoft.Windows.Storage.Pickers.h>
+using namespace winrt::Microsoft::Windows::Storage::Pickers;
+
+FolderPicker picker;
+auto& result{ co_await openPicker.PickSingleFolderAsync() };
+if (result)
+{
+    // Perform this conversion if you have business logic that uses StorageFolder
+    auto& storageFolder{ co_await winrt::Windows::Storage::StorageFolder::GetFolderFromPathAsync(result.Path) }
```
oldnewthing:
```suggestion
    auto storageFolder{ co_await winrt::Windows::Storage::StorageFolder::GetFolderFromPathAsync(result.Path) }
```
Compiler error:
> Cannot convert from winrt::Windows::Storage::StorageFolder to winrt::Windows::Storage::StorageFolder&
> A non-const reference may only be bound to an lvalue. #Closed

specs/Storage.Pickers/PickerLocationId.md
```diff
@@ -0,0 +1,107 @@
...
+}
+```
+
+## PickerLocationId.DocumentsLibrary
+The `DocumentsLibrary` value represents the user's Documents library.
+
+This is the default location where many applications save documents, and it is often mapped 
+to the `%USERPROFILE%\Documents` folder on the user's system. 
+If the user has configured OneDrive for documents, this location would be redirected to a OneDrive 
+folder, such as `%USERPROFILE%\OneDrive - ...\Documents`.
```
oldnewthing:
```suggestion
folder, such as `%USERPROFILE%\OneDrive\Documents`.
```

Most users have only one OneDrive, so it's just called "OneDrive" with no suffix.

Similarly for the other examples. #Closed

specs/Storage.Pickers/PickerLocationId.md
```diff
@@ -0,0 +1,107 @@
...
+
+## PickerLocationId.VideosLibrary
+The `VideosLibrary` value represents the user's Videos library.
+
+This is the default location where many applications save video files, and it is often mapped 
+to `%USERPROFILE%\Videos`. If the user has configured OneDrive for the videos library, this location 
+would be redirected to a OneDrive folder, such as `%USERPROFILE%\OneDrive - ...\Videos`.
+
+## PickerLocationId.Objects3D
+! Note: This location is being phased out, so it is recommended to avoid using this value.
```
oldnewthing, line 92: What happens when the phaseout is complete and somebody passes Objects3D? #Closed

specs/Storage.Pickers/PickerLocationId.md
```diff
@@ -0,0 +1,107 @@
...
+
+## PickerLocationId.Unspecified
+The `Unspecified` value represents an unspecified location.
+
+When this value is used, the file picker will use the system default location, which may vary 
+depending on the user's configuration and the context in which the picker is used.
+
+When the `SuggestedStartLocation` was not specified in [FileOpenPicker](./FileOpenPicker.md), 
+[FileSavePicker](./FileSavePicker.md), or [FolderPicker](./FolderPicker.md), 
+the default value is `PickerLocationId.Unspecified`.
```
oldnewthing, line 107: Not sure what this means. There is no SuggestedStartLocation parameter to the FileOpenPicker, FileSavePicker, or FolderPicker constructor, so there's nowhere to specify it.

If you're saying that the default value of the SuggestedStartLocation property is Unspecified, then that remark should go in the documentation of that property in FileOpenPicker, FileSavePicker, and FolderPicker.

specs/Storage.Pickers/FileOpenPicker.md
```diff
@@ -0,0 +1,183 @@
...
+openPicker.SuggestedStartLocation(PickerLocationId::DocumentsLibrary);
+
+// (Optional) specify the text displayed on commit button. If not specified, use system default.
+openPicker.CommitButtonText(L"Choose selected files");
+
+// (Optional) specify file extensions filters. If not specified, default to all (*.*)
+openPicker.FileTypeFilter().Append(L".txt");
+openPicker.FileTypeFilter().Append(L".pdf");
+openPicker.FileTypeFilter().Append(L".doc");
+openPicker.FileTypeFilter().Append(L".docx");
```
oldnewthing: Quicker is
```cpp
openPicker.FileTypeFilter().ReplaceAll({ L".txt", L".pdf", L".doc", L".docx" }); #Closed

File: specs/Storage.Pickers/FileOpenPicker.md [hunk 10]
```diff
@@ -0,0 +1,183 @@
...
+#include <fstream>
+#include <string>
+using namespace winrt::Microsoft::Windows::Storage::Pickers;
+
+FileOpenPicker openPicker(AppWindow().Id());
+auto& result{ co_await openPicker.PickSingleFileAsync() };
+if (result)
+{
+    std::ifstream fileReader(result.Path().c_str());
+    std::string text((std::istreambuf_iterator<char>(fileReader)), std::istreambuf_iterator<char>());
```
oldnewthing: Can be simplified to

```suggestion
    std::string text(std::istreambuf_iterator(fileReader), {});
``` #Closed

File: specs/Storage.Pickers/FileOpenPicker.md [hunk 11]
```diff
@@ -0,0 +1,183 @@
...
+        std::string text((std::istreambuf_iterator<char>(fileReader)), std::istreambuf_iterator<char>());
+        winrt::hstring hText = winrt::to_hstring(text);
+    }
+}
+else
+{
+    // error handling.
+}
+```
+
```
oldnewthing, line 186: Need some discussion of the properties not demonstrated by the samples.

#### CommitButtonText (String)

Label text of the FileOpenPicker's commit button. If this is the empty string, then the system uses a default label of "Open" (suitably translated).

Default: Empty string.

#### SettingsIdentifier (String)

An arbitrary string to associate with the state of the picker. 

If your application uses multiple instances of the file open picker, you can use this property to identify the individual instances.

To give the user a sense of consistency, the file open picker remembers the last location that the user navigated to and will generally start at that location. Normally, this remembered location is shared by all file open pickers used by the app. To remember a different location for different pickers, choose a different SettingsIdentifier for each picker.

For example, if you use one file open picker for picking flowers, and another for picking fruit, you can set the SettingsIdentifier for the first picker to "Flowers" and the SettingsIdentifier for the second picker to "Fruit". That way, each time the user picks flowers, they default to the last location they used for picking flowers; and each time the user picks fruit, they default to the last location they used for picking fruit.

Default: Empty string

#### ViewMode (PickerViewMode)

The view mode for displaying items.

Default: List <- is this true?

specs/Storage.Pickers/FileSavePicker.md
```diff
@@ -0,0 +1,208 @@
...
+`FileSavePicker.SuggestedSaveFilePath` is a read-only attribute, meaning it can only be set via 
+above method `FileSavePicker.TrySetSuggestedSaveFilePath`.
+
+### Examples
+C#
+```C#
+using Microsoft.Windows.Storage.Pickers;
+
+var savePicker = new FileSavePicker(this.AppWindow.Id);
+var isSuccess = savePicker.TrySetSuggestedSaveFilePath(@"C:\temp\MyProject\MyFile.txt");
```
oldnewthing: Can I pass an empty string to cancel a previous TrySetSuggestedSaveFilePath and return to the system default? #Closed

specs/Storage.Pickers/FileSavePicker.md
```diff
@@ -0,0 +1,208 @@
...
+}
+```
+
+### The Parsing Logic of TrySuggestedSaveFilePath
+
+Method `TrySuggestedSaveFilePath` takes whatever after **the last slash** to fill in the file name 
+box, as long as the part before the last slash is a valid and existing folder. 
+
+And return `false` without writing any value if the input doesn't even contain a slash, 
+or the folder doesn't exist on the end user's environment.
```
oldnewthing: On the other hand, the directory might be valid at the time the Try method is called, but becomes invalid before you call PickSaveFileAsync(). What happens in that case? Do we throw an error? Do we just ignore the folder but keep the file name? Do we ignore both the folder and file name?

What happens if SuggestedSaveFilePath and SuggestedFileName conflict?

Maybe we should just let people set SuggestedSaveFilePath to whatever they want and do validation at PickSaveFileAsync time, seeing as we have to do late validation anyway.

specs/Storage.Pickers/FileOpenPicker.md
```diff
@@ -0,0 +1,183 @@
...
+# See Also
```
oldnewthing, line 187: What is the scope of SettingsIdentifier? For UWP, they are scoped to the package. What is the scope for WinAppSDK? Does it depend on whether the app is packaged? #Closed

specs/Storage.Pickers/FileOpenPicker.md
kmahone: RECOMMENDED: Do that.

specs/Storage.Pickers/FileOpenPicker.md
kmahone: RECOMMENDED: do that.

specs/Storage.Pickers/PickFolderResult.md
Scottj1s: same for result above, and for similar references throughout

specs/Storage.Pickers/FileOpenPicker.md
kmahone: RECOMMENDED: 
Set these properties from a sample.
Add the descriptive text that Raymond provided to the spec.
Find out what is the default value for ViewMode and add it to the doc.

specs/Storage.Pickers/FileOpenPicker.md
kmahone: DISCUSSION: If I use the same SettingsIdentifier for FileOpenPicker and FileSavePicker do they share the same state?
Answer: Currently yes, but it might change in the future.

What happens if two unrelated unpackaged apps use the same SettingsIdentifier string?
Currently, all apps will share the same global state for the same SettingsIdentifier string.

How do we avoid unintentional collisions between unrelated unpackaged apps?

Should this be typed as GUID instead of string?
Could we hash it with the executable name?

RECOMMENDED: Cut this SettingsIdentifier property for v1. 
In the future we can add it back when the behavior is more understood.

specs/Storage.Pickers/FileSavePicker.md
kmahone: DISCUSSION: No. It will throw an error.
If I call TrySetSuggestedSaveFilePath, I cannot undo that.

specs/Storage.Pickers/FileSavePicker.md
kmahone: DISCUSSION:
Do we throw when you call PickSaveFileAsync? Do you ignore the error? That is confusing.
We want to do some validation upfront.

In UWP, we used StorageFolder, so we got some validation upfront for free from that.


RECOMMENDED:
Remove TrySetSuggestedSaveFilePath method.
SuggestedSaveFilePath is a read-write property.
On set, validate that the string is a well formed path. Throw an invalid arg exception if the string is not a well-formed path string. The previous value is preserved.
If it is a well-formed path that does not exist, we should NOT throw.
If is is set to empty string, it should reset back to default. 
At pick time, if the folder is not there, fall back to a known good path. Don't show any error message.

specs/Storage.Pickers/Microsoft.Windows.Storage.Pickers.md
kmahone: RECOMMENDED: Use "property" as the name, not attribute.

specs/Storage.Pickers/Microsoft.Windows.Storage.Pickers.md
kmahone: DISCUSSION: In plans for v2 of this api.

specs/Storage.Pickers/Microsoft.Windows.Storage.Pickers.md
kmahone: RECOMMENDED: Remove this.

specs/Storage.Pickers/PickFileResult.md
kmahone: RECOMMENDED: Use property.

specs/Storage.Pickers/PickFolderResult.md
kmahone: RECOMMENDED: Fix this. Also validate that the other sample code compiles.

specs/Storage.Pickers/PickerLocationId.md
kmahone: RECOMMENDED: Simplify the sample.

specs/Storage.Pickers/PickerLocationId.md
kmahone: DISCUSSION: It will fall back to the documents folder.

specs/Storage.Pickers/PickerLocationId.md
kmahone: DISCUSSION: Clarify the doc here.

specs/Storage.Pickers/Microsoft.Windows.Storage.Pickers.md
DinahK-2SO: Yes, there're keen requests from the community, (e.g. [this comment](https://github.com/microsoft/WindowsAppSDK/discussions/5368#discussioncomment-12949910) ). 

We're planning to achieve this function in v2.

specs/Storage.Pickers/PickerLocationId.md
DinahK-2SO: Fallback to the Documents folder.

Including the Objects3D, any location listed in the `PickerLocationId` can fallback to the `Documents` folder when the location doesn't exists on end user's machine.

specs/Storage.Pickers/PickerLocationId.md
DinahK-2SO: Yes, this point is also documented in the 3 pickers too. 

I presume developers might search for the PickerLocationId to get the values they can use and what does each value mean, so I also added some details here.

specs/Storage.Pickers/FileOpenPicker.md
DinahK-2SO: Updated the document.

Answer to the last question: yes, it is true that ViewMode defaults to List.

specs/Storage.Pickers/FileOpenPicker.md
DinahK-2SO: Thanks @oldnewthing for reviewing this design.

I have removed the contents about `SettingsIdentifier` as we have decided.

specs/Storage.Pickers/FileSavePicker.md
DinahK-2SO: Thanks @oldnewthing for raising this topic here!
I removed the method `TrySetSuggestedSaveFilePath`, using the `SuggestedSaveFilePath` property's own setter to update the values, as we concluded in during the review.

specs/Storage.Pickers/FileSavePicker.md
```diff
@@ -0,0 +1,195 @@
...
+
+```C#
+runtimeclass FileSavePicker
+{
+    FileSavePicker(Microsoft.UI.WindowId windowId);
+
+    string CommitButtonText;
+    string DefaultFileExtension;
+    string SuggestedFileName;
+    string SuggestedSaveFilePath;
```
Scottj1s, line 22: We had discussed TrySuggestedSaveFilePath overriding SuggestedFileName.  But seeing these two properties now side by side, the redundancy is jarring.  I propose that we remove SuggestedFileName and just keep the existing parsing behavior that we had for TrySuggestedSaveFilePath, applied to SuggestedSaveFilePath.  I.e., you can provide a folder and optionally a file name to initialize the dialog with.  We could of course deterministically define the behavior of these two properties, such as "last one wins" - but it just seems like a confusing design to have both.

specs/Storage.Pickers/FileSavePicker.md
```diff
@@ -0,0 +1,195 @@
...
+
+*   **Initial Directory:** The dialog defaults to the directory of the `SuggestedSaveFilePath`. 
+
+    This behavior overrides any picker-specific remembered folder settings and ensures the dialog 
+    opens in the suggested file's parent folder.
+
+*   **Pre-filled File Name:** The file name field in the dialog is pre-populated with the name of 
+the `SuggestedSaveFilePath`. 
+
+    This file name overrides the `FileSavePicker.SuggestedFileName` property if both are set.
```
Scottj1s: This helps.  Looking at Tempo for other examples, I see we use *FilePath properties generally to capture fully qualified path+name.  I suppose the simplicity of SuggestedFileName means we should preserve it, even if it overlaps.  So I would be okay with just renaming SuggestedSaveFilePath to SuggestedFilePath (Save being implicit).  The rename would also set us up better for adding a SuggestedFilePath to FileOpenPicker and/or FolderPicker.

specs/Storage.Pickers/FileSavePicker.md
```diff
@@ -0,0 +1,195 @@
...
+
+### Examples
+C#
+```C#
+using Microsoft.Windows.Storage.Pickers;
+
+var savePicker = new FileSavePicker(this.AppWindow.Id);
+savePicker.SuggestedSaveFilePath(@"C:\temp\MyProject\MyFile.txt");
```
Scottj1s: Update all C# sample snippets to use property assignment syntax, since this is no longer a method

specs/Storage.Pickers/FileSavePicker.md
Scottj1s: Should it 'override' or simply set the property?  I.e., shouldn't this be the behavior?

```
SuggestedFileName = "FirstName.txt";
SuggestedSaveFilePath = @"C:\temp\LastName.txt";
// now, SuggestedFileName = "LastName.txt";

```

This "last one wins" strategy is simple to reason about - users could order their code as convenient:
```
SuggestedSaveFilePath = SomeFilePathVariable;
// Keep the initial folder, but change the file name
SuggestedFileName = "LastName.txt";

```

specs/Storage.Pickers/Microsoft.Windows.Storage.Pickers.md
```diff
@@ -0,0 +1,146 @@
...
+1. *All new pickers are designed specifically for desktop apps and uses a `WindowId` property to 
+link the picker to its host window, replacing the `WinRT.Interop.InitializeWithWindow.Initialize` 
+pattern.*
+1. The new pickers allow developers to use pickers without requiring `FileTypeFilter` or 
+`FileTypeChoices` to be specified. While UWP pickers throw exceptions when these properties are not 
+set, the new pickers default to show all file when developers don't explicitly configure these 
+properties.
+1. *HomeGroup location has been excluded from the new `PickerLocationId` enum as it's no longer 
+supported in Windows 10 and later*
+1. *FolderPicker.FileTypeFilter has been exluded as it was not functionally appropriate for folder
```
Scottj1s:
```suggestion
1. *FolderPicker.FileTypeFilter has been excluded as it was not functionally appropriate for folder 
```

specs/Storage.Pickers/PickerLocationId.md
```diff
@@ -0,0 +1,107 @@
+PickerLocationId Enum
+===
+
+Specifies the default folder or library that a picker starts from or displays to the user.
```
Scottj1s, line 4: Maybe mention that this is a best effort, which won't present an error to the user on failure to resolve to a physical folder (e.g., if Objects3D doesn't exist), but will default to some other reasonable location.

specs/Storage.Pickers/FileSavePicker.md
DinahK-2SO: The file name of SuggestedSaveFilePath takes precedence over the SuggestedFileName. This behavior is consistent to that of the UWP FileSavePicker.

Updated the spec and code to remove TrySetSuggestedSaveFilePath method, adding the setter of SuggestedSaveFilePath property with validations.

specs/Storage.Pickers/FileSavePicker.md
```diff
@@ -0,0 +1,195 @@
...
+bool isSuccess = savePicker.SuggestedSaveFilePath(L"C:\\temp\\MyProject\\MyFile.txt");
+```
+
+### The Parsing Logic of Setting SuggestedSaveFilePath
+
+`SuggestedSaveFilePath` takes whatever after **the last slash** to fill in the file name box, 
+as long as the part before the last slash is a valid and existing folder. 
+
+It allows the value to be set to an empty string. However, it raises invalid argument exception 
+if the non-empty input doesn't contain have a folder path.
```
codendone: "doesn't contain have a folder path" -- perhaps "has invalid path syntax"?

specs/Storage.Pickers/Microsoft.Windows.Storage.Pickers.md
```diff
@@ -0,0 +1,146 @@
...
+===
+
+The `Microsoft.Windows.Storage.Pickers` API in the Windows App SDK provides a modernized file and 
+folder picker experience for desktop applications. This API is based on the existing 
+[Windows.Storage.Pickers](https://learn.microsoft.com/en-us/uwp/api/windows.storage.pickers) API 
+design, but with key improvements for desktop scenarios.
+
+# Background
+
+The new `Microsoft.Windows.Storage.Pickers` API addresses three critical limitations of the UWP file
```
codendone: Should `three` be `two`, given that only two bullets are listed?

specs/Storage.Pickers/FileSavePicker.md
oldnewthing: Windows Runtime design principles are that properties can be set in any order with no change in effect. So last-writer-wins breaks this rule. The rule exists because in scenarios like XAML binding, you don't explicitly control the order in which the properties are set.

It should say that "this file name **takes priority** over SuggestedFileName" rather than overrides it.

They are independent properties, and you can set them differently if you like, but we will choose SuggestedSaveFilePath if there is a conflict with SuggestedFileName.

Scottj1s: That principle makes sense for independent properties, but does it apply here?  Do we have another API example of overlapping properties?  To take the binding example, what if you had two bindings setting a common property - which would take effect?  My concern is that users might set SuggestedFileName and be confused that it's not being applied.

DinahK-2SO: Hi @Scottj1s , thanks raising this point! I am closing this conversation since our offline discussion has decided not to rename this property now. 

The reason is: The `FileOpenPicker` and `FolderPicker` might need a `SuggestedFolderPath` more than a `SuggestedFilePath` - the `FolderPicker` Dialog doesn't display files; and the `FileOpenPicker`, unlike the `FileSavePicker` that can create a new file, it is a little strange to let developers presume a file exists on user's machine. 

Closing this conversation, we will have more discussion for the new functionalites of `FileOpenPicker` and `FolderPicker` in future.

specs/Storage.Pickers/FileSavePicker.md
DinahK-2SO: Looks this part was not written clearly. I am rephrasing it.

specs/Storage.Pickers/PickerLocationId.md
DinahK-2SO: Thanks @Scottj1s for raising this point! I added this info in the pages of pickers.

specs/Storage.Pickers/FileSavePicker.md
yeelam-gordon: Ah... @Scottj1s, that's a good perspective. I used to think about it differently.
What surprised me initially was that there are two separate properties controlling something related, even though they don't carry exactly the same information.

If there were only one property—for example, just SuggestSaveFileInfo—and multiple methods like SetPath and SetFileName to update it, then I wouldn't see it as a 'last-one-wins' situation. Instead, I’d view it as simply updating parts of the same file information."

Scottj1s: That's fine, but I'd just offer that independent of what we do for FileOpenPicker/FolderPicker, it doesn't make a lot of sense for FileSavePicker to have any properties with "Save" in the name.  Worse, it's inconsistent - why don't all properties have "Save" in the name?  (I'm not advocating that - think none should, it's implied).

Scottj1s: @yeelam-gordon, yes, I originally added a comment to that effect, but later deleted it:  why not just have a single property to serve everything?  Then, Raymond's concern would be moot, since the overlap would be eliminated. SuggestedFilePath could refer to a folder, a filename, or both.  But I thought the filename-only use might not be as clear as the existing dedicated property, SuggestedFileName.  If we want to retain both, I'd suggest we make SuggestedFilePath either a) only be used to set a path, so as to not overlap with SuggestedFileName; or b) have SuggestedFilePath set SuggestedFileName indirectly.  Either of those would be a better developer experience than the override behavior (IMO).

Scottj1s: What did we decide?  I think the main concern was property overlap, in turn leading to concerns/questions about overriding vs setting.  So if we eliminate the overlap, we eliminate those concerns.  Shall we go with SuggestedFileName and SuggestedFolderPath?

DinahK-2SO: Offline discussion with API reviewers (including @Scottj1s, @oldnewthing, @codendone , @yeelam-gordon) decided to:
1. remove `FileSavePicker.SuggestedSaveFile`
2. add new property `FileSavePicker.SuggestedFolder`

In future (1.9), we can add `FileSavePicker.SuggestedDefaultFolder` to allow for configs like `SuggestedStartLocation` with string type.

Hello @Scottj1s , thanks for raising this topic and sharing your insights! I have updated the spec. Please take a look back.

charlesroddie:
How is there a whole review process and no mention of the fact that you can't use pickers in AppContainers? All comments from community (in previous PRs) and all the existing issues about pickers get ignored in creating this spec? App developers that want to distribute to end users with reasonable security need to wait another 5 years for a V3 spec?

@DinahK-2SO @oldnewthing @kmahone @codendone @yeelam-gordon @Scottj1s

DinahK-2SO:
> How is there a whole review process and no mention of the fact that you can't use pickers in AppContainers? All comments from community (in previous PRs) and all the existing issues about pickers get ignored in creating this spec? App developers that want to distribute to end users with reasonable security need to wait another 5 years for a V3 spec?
> 
Hi @charlesroddie , thanks for highlighting this important topic! 

This spec primarily targets enabling picker functionality when running in elevated mode (i.e., as administrator). While AppContainer scenarios are not currently in the scope of this spec, I deeply appreciate your interest and enthusiasm. 

The new pickers support containerized [AppContainer with win32-isolation](https://github.com/microsoft/win32-app-isolation), which is also in preview. Please stay tuned for future updates as we continue to expand storage picker capabilities.

PR #5635, opened 2025-07-18
Set WinUISDKReferences=false in Templates

lauren-ciha:
### Motivation for the change:
* The `WinUISDKReferences` property was added to allow developers with pre-existing projects to opt-in to Hybrid CRT. New projects can use Hybrid CRT (`WinUISDKReferences=false`) by default.

### Changes description:
Sets `WinUISDKReferences` to `false` in the template files. See diff for more details.

### Verification Steps:
Manually verified by:
* Building the VsixExtension solution
* Removing the component template vsixes from the Visual Studio Installer
* Double clicking the Standalone vsix file for each project (which installs them on Visual Studio)
* Built a sample project with each template
* Verified that the C# project `AppxManifest.xml` does not have a reference to `Microsoft.VCLibs.140.00.Debug.UWPDesktop`, while the C++ project does

WARNING:

### DIFF
--- a/dev/VSIX/ProjectTemplates/Desktop/CSharp/ClassLibrary/ProjectTemplate.csproj
+++ b/dev/VSIX/ProjectTemplates/Desktop/CSharp/ClassLibrary/ProjectTemplate.csproj
@@ -5,6 +5,7 @@
     <RootNamespace>$safeprojectname$</RootNamespace>
     <RuntimeIdentifiers>win-x86;win-x64;win-arm64</RuntimeIdentifiers>
     <UseWinUI>true</UseWinUI>
+    <WinUISDKReferences>false</WinUISDKReferences>
   </PropertyGroup>
   
 </Project>
--- a/dev/VSIX/ProjectTemplates/Desktop/CSharp/PackagedApp/BlankApp/ProjectTemplate.csproj
+++ b/dev/VSIX/ProjectTemplates/Desktop/CSharp/PackagedApp/BlankApp/ProjectTemplate.csproj
@@ -8,6 +8,7 @@
     <Platforms>x86;x64;ARM64</Platforms>
     <RuntimeIdentifiers>win-x86;win-x64;win-arm64</RuntimeIdentifiers>
     <UseWinUI>true</UseWinUI>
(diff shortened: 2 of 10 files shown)


dev/VSIX/ProjectTemplates/Desktop/CSharp/PackagedApp/BlankApp/ProjectTemplate.csproj
```diff
@@ -8,6 +8,7 @@
     <Platforms>x86;x64;ARM64</Platforms>
     <RuntimeIdentifiers>win-x86;win-x64;win-arm64</RuntimeIdentifiers>
     <UseWinUI>true</UseWinUI>
+    <WinUISDKReferences>false</WinUISDKReferences>
```
Scottj1s, line 11: looks like we'll need to add to all wapproj files as well, since each *.WapProj.vstemplate adds a WindowsAppSDK package reference, which would then (by default) add the VCLibs SDKReferences

dev/VSIX/ProjectTemplates/Desktop/CSharp/ClassLibrary/ProjectTemplate.csproj
```diff
@@ -5,6 +5,7 @@
     <RootNamespace>$safeprojectname$</RootNamespace>
     <RuntimeIdentifiers>win-x86;win-x64;win-arm64</RuntimeIdentifiers>
     <UseWinUI>true</UseWinUI>
+    <WinUISDKReferences>false</WinUISDKReferences>
```
Scottj1s, line 8: good catch

dev/VSIX/ProjectTemplates/Desktop/CSharp/PackagedApp/BlankApp/ProjectTemplate.csproj
Scottj1s: Based on the finding that wapproj independently adds VCLibs references, it might seem unnecessary to opt out of WinAppSDK adding them too.  But I'd still like to keep this in place, just in case wapproj decides to add their own opt out support.

PR #5671, opened 2025-08-01
Fixing configuration for MRT

### LINKED ISSUE #5034: BuildAll.ps1 failed on Debug configuration due to missing MRM.dll
### Describe the bug

`Build.ps1` hardcoded configuration to `Release` by `configurationForMrtAndAnyCPU`, which caused `BuildAll.ps1` with configuration `Debug` failed on clean build, it failed with error:
```
Cannot find path '<WindowsAppSDKPath>\BuildOutput\Release\x64\mrm\MRM.dll' because it does not exist.

at <ScriptBlock>, <WindowsAppSDKPath>\BuildAll.ps1: line 304
```

Run a clean build (all release output need to be removed) of `BuildAll.cmd -Configuration:Debug`.

It seems that there are a special configuration variable for MRT called [configurationForMrtAndAnyCPU](https://github.com/microsoft/WindowsAppSDK/blob/ad538768f704aeed816dcb794f3283829c6f18e7/BuildAll.ps1#L71), which is hardcoded to "Release",
and when build MRT, it uses [configurationToRun](https://github.com/microsoft/WindowsAppSDK/blob/ad538768f704aeed816dcb794f3283829c6f18e7/BuildAll.ps1#L216).
If we run `BuildAll.cmd` in `Debug` mode, after build MRT, when copying output, it will look result from `Release` directory, and those dll is not found.

### Expected behavior

_No response_

### Screenshots

_No response_

### NuGet package version

None

### Packaging type

_No response_

### Windows version

_No response_

### IDE

_No response_

### Additional context

_No response_

guimafelipe:
Fixes #5034 

WARNING:

### DIFF
--- a/BuildAll.ps1
+++ b/BuildAll.ps1
@@ -220,9 +220,9 @@ Try {
             {
                 foreach($platformToRun in $platform.Split(","))
                 {
-                    write-host "Building MrtCore.sln for configuration $configurationToRun and platform:$platformToRun"
+                    write-host "Building MrtCore.sln for configuration $configurationForMrtAndAnyCPU and platform:$platformToRun"
                     & $msBuildPath /restore "$MRTSourcesDirectory\mrt\MrtCore.sln" `
-                                    /p:Configuration=$configurationToRun `
+                                    /p:Configuration=$configurationForMrtAndAnyCPU `
                                     /p:Platform=$platformToRun `
                                     /p:PGOBuildMode=$PGOBuildMode `
                                     /binaryLogger:"BuildOutput/binlogs/MrtCore.$platformToRun.$configurationToRun.binlog"


BuildAll.ps1
```diff
@@ -220,9 +220,9 @@ Try {
             {
                 foreach($platformToRun in $platform.Split(","))
                 {
-                    write-host "Building MrtCore.sln for configuration $configurationToRun and platform:$platformToRun"
+                    write-host "Building MrtCore.sln for configuration $configurationForMrtAndAnyCPU and platform:$platformToRun"
```
ssparach: Using $configurationForMrtAndAnyCPU here results in the MRTCore solution being built only with the Release configuration, as the value appears to be hardcoded. Is building MRTCore with the Debug configuration an unsupported or intentionally disabled scenario?

guimafelipe: From my understanding, the entire code uses the Release configuration for MrtCore solution. This was the only place where it was different. This change is coherent with the rest of the script. I would like to confirm that with @kythant as the intended behavior.

kythant: $configurationForMrtAndAnyCPU was used in staging the files over to the NuGet. While it's valid that we are throwing away the Debug binaries after building them, we don't really make use of the debug binaires from MRTCore. Hence, either we keep building Debug just to build Debug or we also delete the foreach loop for the configuration on line 219.

guimafelipe: The problem is that when we only passed `Debug` configuration when running the script, it will build 
`Debug` and then try to use `Release`. I am for now removing the foreach loop.

PR #5699, opened 2025-08-11
Add 'All Files' label for the summarizing filter

DinahK-2SO:
### 
This PR is a minor fix - adding the "All Files" label for the summarizing filter of FileOpenPicker.

This label was removed before due to the lack of localization, but didn't add back when the localization is achieved.

### Before fix
image

### After fix
image

### DIFF
--- a/dev/Interop/StoragePickers/PickerCommon.cpp
+++ b/dev/Interop/StoragePickers/PickerCommon.cpp
@@ -320,7 +320,7 @@ namespace PickerCommon {
         }
         else
         {
-            FileTypeFilterData.push_back(L"");
+            FileTypeFilterData.push_back(AllFilesText);
             FileTypeFilterData.push_back(allFilesExtensionList.c_str());
         }
 


Hong-Xiang (approved):

dev/Interop/StoragePickers/PickerCommon.cpp
```diff
@@ -320,7 +320,7 @@ namespace PickerCommon {
         }
         else
         {
-            FileTypeFilterData.push_back(L"");
+            FileTypeFilterData.push_back(AllFilesText);
```
yeelam-gordon, line 323: I see your screenshot.
Will this kind of "push_back" data, instead of using a localization template for formatting, causing localization problem for Right to left languages (Like Arabic) or other language their order are just different?

yeelam-gordon: Talked with Dinah, it is added as part of spec instead of string.
Thus, there should be not a problem.

yeelam-gordon (approved):

PR #5703, opened 2025-08-12
[FixBuildBreak][1.6] Adding explicit install of Windows SDK 22000 to the BuildInstaller and StaticValidation stages

alexlamtest:
Updating 1.6-stable's BuildInstaller stage to explicitly install Windows SDK 22000 in a pipeline run, because this version is no longer included in VS 2022.
The script windows-sdk.ps1, which already installs another version of the Windows SDK, is now upgraded to support installing 2 different version of the Windows SDK.A microsoft employee must use /azp run to validate using the pipelines below.

WARNING:


alexlamtest:
Should target 1.8.

PR #5718, opened 2025-08-18
Fix race with /external github trigger

ssparach:

WARNING:


ssparach:
> It looks like this will immediately close all area-External issues (currently 11), regardless of how that label was set. Is that intentional?

Nope, it wasn't intentional. I've updated the change so it now only closes open external issues where the external bot comment is present. Thanks!

PR #5719, opened 2025-08-18
Fixing and enabling AppLifecycle tests

### LINKED ISSUE #5710: Enable AppLifecycle tests
The AppLifecycle tests are broken and are currently disabled. We need to fix and enable them to have more robustness in the project and prevent regressions. This is similar to the issue #5654, and can be enhanced to add new tests to assure changes like #5696.

Currently, the tests are breaking here: 
`[Error]: AreNotEqual(waitResult, static_cast<DWORD>(WAIT_TIMEOUT)) - Values (258, 258) [File \test\AppLifecycle\Shared.cpp Line: 105]`

guimafelipe:
This Pull Request aims to fix and enable the AppLifecycle tests.
Fixes #5710.

The changes are:

- Removing the manual installation of `Microsoft.VCLibs.140.00.UWPDesktop` as recommended by the [documentation](https://learn.microsoft.com/en-us/troubleshoot/developer/visualstudio/cpp/libraries/c-runtime-packages-desktop-bridge). Now adding it as a package reference in the appx manifest for the test app with version `0.0.0.0` so it always gets the best match for the different images and configuration at the testing pipeline.
- Implemented the `TODO` for making the file and protocol registrations to not fail if the file is not found or the path is not existent. This was making some tests fail by not signalizing success when they were the first being run. For example, the file registration test tries to unregister first to clean up possible other failing tests. If it were the first one, it would not find the file extension in the registry and throw an exception. Now, we catch this exception and check if it was because the file or the path were inexistent, which is an expected situation for the test.
- Fixed the `BootstrapInitialize()` method in `Helpers.cpp` on the test app to correctly use the prefixes instead of the complete family names for the packages. The version and the publisher ID are separated inputs.
- Disabled the AppLifecycle Tests from running on the image `Win10_rs5_DC_x64chk` as it is not currently being able to install the `VCLibs` correctly. They are now in the `BypassTests.json` and will be considered skipped in this image. All the other images, Windows 11 and 10, arm64 or x64, are working properly.

With these changes, the tests are expected to run correctly and therefore the `.testdef` is changed to be `Enabled`.

WARNING:


test/BypassTests.json
lauren-ciha, line 1: Consider updating the pull request description to include why these tests aren't being run.

guimafelipe: Done!

ssparach (approved):

PR #5729, opened 2025-08-20
Correct Package Family Name Construction for Main and Singleton Packages

ssparach:
Summary
----------------------------------------------------------------------------------------------------------

This change corrects the construction of the packageFamilyName for the main and singleton packages in DeploymentManager.cpp. Previously, the package family names were being built incorrectly, which caused the DeploymentManager to incorrectly determine that these packages were not installed. As a result, the deployment status was reported as PackageInstallRequired instead of Ok.

Details
•	Fixed the logic that formats the version tag and constructs the packageFamilyName for main and singleton packages.
•	With the corrected names, DeploymentManager now accurately detects installed packages and returns the correct deployment status.

Testing
-------------------------------------------------------------------------------------------------------------------
The fix was validated in the debugger: after the change, the deployment status is Ok when the packages are present. 

WARNING:

### DIFF
--- a/dev/Deployment/DeploymentManager.cpp
+++ b/dev/Deployment/DeploymentManager.cpp
@@ -112,23 +112,72 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
         {
             // Build package family name based on the framework naming scheme.
             std::wstring packageFamilyName{};
+            // The main and singleton packages contain a shortened version tag for the package family name.
+            std::wstring formattedVersionTag{};
+
+            auto ExtractFormattedVersionTag = [](const std::wstring& versionTag, bool addDashPrefix = false) -> std::wstring
+            {
+                wchar_t firstLetter = versionTag.size() > 1 ? versionTag[1] : L'\0';
+                std::wstring numberBeforeUnderscore;
+                size_t i = 2;
+                while (i < versionTag.size() && iswdigit(versionTag[i]))
+                {
+                    numberBeforeUnderscore += versionTag[i];
+                    ++i;
+                }
+                std::wstring result;
+                if (firstLetter != L'\0')
+                {
+                    if (addDashPrefix)
+                    {
+                        result += L"-";
+                    }
+                    result += firstLetter;
+                }
+                result += numberBeforeUnderscore;
+                return result;
+            };
+
+            // PackageFamilyName = Prefix + Subtypename + VersionTag + Suffix
+            // On WindowsAppSDK 1.1+, Main and Singleton packages are sharing same Package Name Prefix.
             if (package.versionType == PackageVersionType::Versioned)
             {
-                // PackageFamilyName = Prefix + SubTypeName + VersionIdentifier + Suffix
-                // On WindowsAppSDK 1.1+, Main and Singleton packages are sharing same Package Name Prefix.
-                packageFamilyName = WINDOWSAPPRUNTIME_PACKAGE_NAME_MAINPREFIX WINDOWSAPPRUNTIME_PACKAGE_SUBTYPENAME_DELIMETER + package.identifier + packageNameVersionIdentifier + WINDOWSAPPRUNTIME_PACKAGE_NAME_SUFFIX;
+                // Versioned: keep "Major.Minor-" and append first letter after '-' and number (if any) before '_'
+                // Example: packageNameVersionIdentifier = "1.8-experimental4", packageNameVersionTag = "-experimental4"
+                // formattedVersionTag = "1.8-e4"
+
+                size_t dashPos = packageNameVersionIdentifier.find('-');
+                std::wstring majorMinor = (dashPos != std::wstring::npos) ? packageNameVersionIdentifier.substr(0, dashPos) : packageNameVersionIdentifier; // "1.8"
+                if (dashPos != std::wstring::npos)
+                {
+                    formattedVersionTag = majorMinor + ExtractFormattedVersionTag(packageNameVersionTag, /*addDashPrefix*/true);
+                }
+                else
+                {
+                    formattedVersionTag = majorMinor;
+                }
             }
             else if (package.versionType == PackageVersionType::Unversioned)
             {
-                // PackageFamilyName = Prefix + Subtypename + VersionTag + Suffix
-                // On WindowsAppSDK 1.1+, Main and Singleton packages are sharing same Package Name Prefix.
-                packageFamilyName = WINDOWSAPPRUNTIME_PACKAGE_NAME_MAINPREFIX WINDOWSAPPRUNTIME_PACKAGE_SUBTYPENAME_DELIMETER + package.identifier + packageNameVersionTag + WINDOWSAPPRUNTIME_PACKAGE_NAME_SUFFIX;
+                // Unversioned: append first letter after '-' and number (if any) before '_'
+                // Example: packageNameVersionTag = "-experimental4"
+                // formattedVersionTag = "-e4"
+                if (packageNameVersionTag.find(L'-') != std::wstring::npos)
+                {
+                    formattedVersionTag = ExtractFormattedVersionTag(packageNameVersionTag, /*addDashPrefix*/true);
+                }
+                else
+                {
+                    formattedVersionTag = L"";
+                }
             }
             else
             {
                 // Other version types not supported.
                 FAIL_FAST_HR(HRESULT_FROM_WIN32(ERROR_UNSUPPORTED_TYPE));
             }
+
+            packageFamilyName = WINDOWSAPPRUNTIME_PACKAGE_NAME_MAINPREFIX WINDOWSAPPRUNTIME_PACKAGE_SUBTYPENAME_DELIMETER + package.identifier + formattedVersionTag + WINDOWSAPPRUNTIME_PACKAGE_NAME_SUFFIX;
 
             // Get target version based on the framework.
             auto targetPackageVersion{ frameworkPackageInfo.Package(0).packageId.version };
@@ -181,7 +230,7 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
             initializeActivityContext.SetIsFullTrustPackage();
         }
 
-            ::WindowsAppRuntime::Deployment::Activity::Context::Get().SetIsFullTrustPackage();
+        ::WindowsAppRuntime::Deployment::Activity::Context::Get().SetIsFullTrustPackage();
         initializeActivityContext.GetActivity().Start(deploymentInitializeOptions.ForceDeployment(),
                                                       Security::IntegrityLevel::IsElevated(),
                                                       isPackagedProcess,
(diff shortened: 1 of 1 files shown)


dev/Deployment/DeploymentManager.cpp
```diff
@@ -112,24 +112,73 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
         {
             // Build package family name based on the framework naming scheme.
             std::wstring packageFamilyName{};
+            // The main and singleton packages contain a shortened version tag for the package family name.
+            std::wstring formattedVersionTag{};
+
+            auto ExtractFormattedVersionTag = [](const std::wstring& versionTag, bool addDashPrefix = false) -> std::wstring
+            {
+                wchar_t firstLetter = versionTag.size() > 1 ? versionTag[1] : L'\0';
```
manodasanW: Can you add some comments for the format of these variables like for versionTag

dev/Deployment/DeploymentManager.cpp
```diff
@@ -112,24 +112,73 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
...
+                std::wstring numberBeforeUnderscore;
+                size_t i = 2;
+                while (i < versionTag.size() && iswdigit(versionTag[i]))
```
manodasanW: Assuming the number comes after -experimental, it is going to stop on e at the start, isn't it?

Scottj1s: agreed - maybe consider fscanf?

dev/Deployment/DeploymentManager.cpp
```diff
@@ -112,24 +112,73 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
         {
             // Build package family name based on the framework naming scheme.
             std::wstring packageFamilyName{};
+            // The main and singleton packages contain a shortened version tag for the package family name.
+            std::wstring formattedVersionTag{};
+
+            auto ExtractFormattedVersionTag = [](const std::wstring& versionTag, bool addDashPrefix = false) -> std::wstring
```
Scottj1s: remove addDashPrefix (always true)

dev/Deployment/DeploymentManager.cpp
```diff
@@ -112,24 +112,73 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
...
+                size_t dashPos = packageNameVersionIdentifier.find('-');
+                std::wstring majorMinor = (dashPos != std::wstring::npos) ? packageNameVersionIdentifier.substr(0, dashPos) : packageNameVersionIdentifier; // "1.8"
+                if (dashPos != std::wstring::npos)
+                {
+                    formattedVersionTag = majorMinor + ExtractFormattedVersionTag(packageNameVersionTag, /*addDashPrefix*/true);
+                }
+                else
+                {
+                    formattedVersionTag = majorMinor;
+                }
```
Scottj1s: simplification:
```suggestion
                if (!packageNameVersionTag.empty())
                {
                    formattedVersionTag = packageNameVersionIdentifier.substr(0, versionTagPos) + ExtractFormattedVersionTag(packageNameVersionTag, /*addDashPrefix*/true);
                }
                else
                {
                    formattedVersionTag = packageNameVersionIdentifier; // "1.8"
                }
```

dev/Deployment/DeploymentManager.cpp
```diff
@@ -112,24 +112,73 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
...
             }
             else if (package.versionType == PackageVersionType::Unversioned)
             {
-                // PackageFamilyName = Prefix + Subtypename + VersionTag + Suffix
-                // On WindowsAppSDK 1.1+, Main and Singleton packages are sharing same Package Name Prefix.
-                packageFamilyName = WINDOWSAPPRUNTIME_PACKAGE_NAME_MAINPREFIX WINDOWSAPPRUNTIME_PACKAGE_SUBTYPENAME_DELIMETER + package.identifier + packageNameVersionTag + WINDOWSAPPRUNTIME_PACKAGE_NAME_SUFFIX;
+                // Unversioned: append first letter after '-' and number (if any) before '_'
+                // Example: packageNameVersionTag = "-experimental4"
+                // formattedVersionTag = "-e4"
+                if (packageNameVersionTag.find(L'-') != std::wstring::npos)
```
Scottj1s:
```suggestion
                if (!packageNameVersionTag.empty())
```

dev/Deployment/DeploymentManager.cpp
DrusTheAxe: Don't parse it this way. Use `swscanf_s` like in MddBootstrap.cpp and MddLifetimeManagement.cpp

This greatly simplifies the code to a few lines.

dev/Deployment/DeploymentManager.cpp
```diff
@@ -112,24 +112,73 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
...
+
+            // PackageFamilyName = Prefix + Subtypename + VersionTag + Suffix
+            // On WindowsAppSDK 1.1+, Main and Singleton packages are sharing same Package Name Prefix.
             if (package.versionType == PackageVersionType::Versioned)
             {
-                // PackageFamilyName = Prefix + SubTypeName + VersionIdentifier + Suffix
-                // On WindowsAppSDK 1.1+, Main and Singleton packages are sharing same Package Name Prefix.
-                packageFamilyName = WINDOWSAPPRUNTIME_PACKAGE_NAME_MAINPREFIX WINDOWSAPPRUNTIME_PACKAGE_SUBTYPENAME_DELIMETER + package.identifier + packageNameVersionIdentifier + WINDOWSAPPRUNTIME_PACKAGE_NAME_SUFFIX;
+                // Versioned: keep "Major.Minor-" and append first letter after '-' and number (if any) before '_'
+                // Example: packageNameVersionIdentifier = "1.8-experimental4", packageNameVersionTag = "-experimental4"
```
DrusTheAxe: This is inaccurate and will age over time

Replace this comment with a reference to see `3. Package Naming` in specs\Deployment\MSIXPackages.md

dev/Deployment/DeploymentManager.cpp
```diff
@@ -112,24 +112,73 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
...
+                    formattedVersionTag = L"";
+                }
             }
             else
             {
                 // Other version types not supported.
                 FAIL_FAST_HR(HRESULT_FROM_WIN32(ERROR_UNSUPPORTED_TYPE));
             }
 
+            packageFamilyName = WINDOWSAPPRUNTIME_PACKAGE_NAME_MAINPREFIX WINDOWSAPPRUNTIME_PACKAGE_SUBTYPENAME_DELIMETER + package.identifier + formattedVersionTag + WINDOWSAPPRUNTIME_PACKAGE_NAME_SUFFIX;
```
DrusTheAxe: use `std::format()` rather than string concatenation e.g.

```
packageFamilyName = std::format(WINDOWSAPPRUNTIME_PACKAGE_NAME_MAINPREFIX WINDOWSAPPRUNTIME_PACKAGE_SUBTYPENAME_DELIMETER L"{}{}" WINDOWSAPPRUNTIME_PACKAGE_NAME_SUFFIX, package.identifier + formattedVersionTag);
```

DrusTheAxe (changes requested):

dev/Deployment/DeploymentManager.cpp
```diff
@@ -113,24 +113,60 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
...
+                    result = std::wstring(L"-") + versionTag[1];
+                }
+
+                unsigned int numberBeforeUnderscore;
+                if (swscanf_s(versionTag.c_str(), L"%*[^0-9]%u", &numberBeforeUnderscore) == 1)
+                {
+                    result += std::to_wstring(numberBeforeUnderscore);
+                }
+                return result;
+            };
```
guimafelipe: Does this being a lambda could affect performance? Could this just be a regular function?

dev/Deployment/DeploymentManager.cpp
```diff
@@ -113,24 +113,60 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
...
+            {
+                //versionTag = "-experimental", result = "-e"
+                //verstionTag = "-experimental4" result = "-e4"
+                std::wstring result{};
+                if (versionTag.size() > 1)
+                {
+                    result = std::wstring(L"-") + versionTag[1];
+                }
+
+                unsigned int numberBeforeUnderscore;
```
DrusTheAxe: NIT: `...score{};`

dev/Deployment/DeploymentManager.cpp
DrusTheAxe: Shouldn't affect perf but symbols will be cryptic if you ever look at a callstack (e.g. in a dump)

I'd make this a function, for easier reader comprehension and maintainability. I mean sure, I can read it, but a function's easier/quicker/more-obvious for more people to grok, and code is read far more often than written, so I always recommend favoring maintainability over sexy new language features if you have a choice.

dev/Deployment/DeploymentManager.cpp
```diff
@@ -113,24 +113,60 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
...
+                if (swscanf_s(versionTag.c_str(), L"%*[^0-9]%u", &numberBeforeUnderscore) == 1)
```
DrusTheAxe: What's the `[^0-9]` do here? Wondering how this differs from "%u"

dev/Deployment/DeploymentManager.cpp
```diff
@@ -113,24 +113,60 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
...
+                }
+                else
+                {
+                    formattedVersionTag = L"";
+                }
             }
             else
             {
                 // Other version types not supported.
                 FAIL_FAST_HR(HRESULT_FROM_WIN32(ERROR_UNSUPPORTED_TYPE));
```
DrusTheAxe: FAIL_FAST will crash the process. Seems pretty harsh considering this is an inproc API running in some app's process space, and giving them no chance to note the issue or recover.

Why not `THROW_HR_MSG(HRESULT_FROM_WIN32(ERROR_UNSUPPORTED_TYPE), "%d", static_cast<int>(package.versionType));`?

guimafelipe (approved):

dev/Deployment/DeploymentManager.cpp
ssparach: `[^0-9]` skips over all non-digit characters before trying to parse the number. This pattern will ensure that the first numeric value is extracted that appears after any prefix.

dev/Deployment/DeploymentManager.cpp
ssparach: Thanks for raising this—totally agree that crashing the process with FAIL_FAST feels overly aggressive and that the host app should be given a chance to respond.
Switching to `THROW_HR_MSG` makes sense and aligns better with expected error handling patterns. The change is small, so I’ve gone ahead and included it in this PR.

Appreciate the review Howard!

PR #5734, opened 2025-08-21
Microsoft.Windows.Foundation.winmd built but not shipped

DrusTheAxe:
Microsoft.Windows.Foundation.winmd built but not shipped (not include…d in Transport package or nuget logic)

See https://github.com/microsoft/WindowsAppSDK/discussions/5250#discussioncomment-14170283


build/CopyFilesToStagingDir.ps1
```diff
@@ -55,6 +55,7 @@ PublishFile $FullBuildOutput\WindowsAppRuntime_DLL\StrippedWinMD\Microsoft.Windo
 PublishFile $FullBuildOutput\WindowsAppRuntime_DLL\StrippedWinMD\Microsoft.Windows.AppNotifications.Builder.winmd $FullPublishDir\Microsoft.WindowsAppRuntime\
 PublishFile $FullBuildOutput\WindowsAppRuntime_DLL\StrippedWinMD\Microsoft.Windows.AppNotifications.winmd $FullPublishDir\Microsoft.WindowsAppRuntime\
 PublishFile $FullBuildOutput\WindowsAppRuntime_DLL\StrippedWinMD\Microsoft.Windows.BadgeNotifications.winmd $FullPublishDir\Microsoft.WindowsAppRuntime\
+PublishFile $FullBuildOutput\WindowsAppRuntime_DLL\StrippedWinMD\Microsoft.Windows.Foundation.winmd $FullPublishDir\Microsoft.WindowsAppRuntime\
```
kythant, line 58: I do feel like we keep running into this issue where we forget to include binaries.

I think we need to modify these scripts, so it includes everything and then warn us if it detects anything new from the baseline.

DrusTheAxe: That would be a neat idea. At least, automatically include *.winmd from WindowsAppRuntime_DLL

Other projects in the solution can be handled more mechanically like today (if there are any producing *.winmd)

PR #5759, opened 2025-08-27
Enabling LRP tests

guimafelipe:
This PR enables LRP, with some changes to achieve that:
- Added the tests to the bypass list for old windows 10 versions.
- Changes `RunAs:Class` to `RunAs` to enable inheritance and run the test methods in the same context.
- Updated the bootstrap initialization code to work with both Win10 and Win11.
- Using case-insensitive string comparison.
- Increase timeout limit to avoid flakiness.
- Added the `.testdef` file.
- Updated proj file to include bootstrap DLLs.

WARNING:


test/LRPTests/APITests.cpp
```diff
@@ -40,13 +41,28 @@ namespace Test::LRP
         TEST_CLASS_SETUP(ClassInit)
         {
 
-            ::Test::Bootstrap::SetupPackages(Test::Bootstrap::Packages::Framework | Test::Bootstrap::Packages::Singleton);
+            ::Test::Bootstrap::Setup();
```
ssparach, line 44: Is there a specific reason why `::Test::Bootstrap::Setup `and `::Test::Bootstrap::Cleanup()` for LRPTests are now being invoked for all default packages, rather than just for Singleton, Framework as originally implemented?

guimafelipe: The bootstrapping requires the main package for Win 10 machines. It was safer to setup everything this way and the cost is not that high.

ssparach (approved):

PR #5764, opened 2025-08-28
DeploymentManager.Initialize improvements. Fix project dependencies. DevCheck enhancements

DrusTheAxe:
* DeploymentManager
  * Fixed concurrency bug if 2 threads concurrently called Initialize
  * Adjusted exception message to clearly explain Initialize() cannot be called again once successful - previous text accounted for AutoIntializer + explicit Intiialize() call but not 2 explicit Iniitalize() calls, and advice only made sense for those using MSBuild-based build systems.
  * Minor perf optimizations
  * Improved error reporting
* Windows SDK references
  * Changed the handful of Windows SDK 18362 and 2200 uses to 26100
  * Updated build pipelines to check/install Windows SDK 10.0.26100 if necessary
* VisualStudio2022.vsconfig
  * VS 2022 recently removed Windows SDKs 18362 and 22000 (in a patch update!) causing errors if you import this config file. Removed the (now)broken components
  * Added Windows 11 SDK 26100 component
* DevCheck
  * -CheckDependencies complained the generated eng/VersionDependencies.props was wrong. Ran -SyncDependencies to correct (equivalent content, no material change)
  * Updated Install-WindowsSDK to also install 10.0.26100.4654
  * Fixed Install-WindowsSDK to correctly use an SDK's URL to download for installation

### DIFF
--- a/dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/unittests/MrtCoreManagedTest.csproj
+++ b/dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/unittests/MrtCoreManagedTest.csproj
@@ -12,7 +12,7 @@
     <AssemblyName>MrtCoreManagedTest</AssemblyName>
     <DefaultLanguage>en-US</DefaultLanguage>
     <TargetPlatformIdentifier>UAP</TargetPlatformIdentifier>
-    <TargetPlatformVersion>10.0.26100.4654</TargetPlatformVersion>
+    <TargetPlatformVersion>10.0.26100.0</TargetPlatformVersion>
     <TargetPlatformMinVersion>10.0.17763.0</TargetPlatformMinVersion>
     <MinimumVisualStudioVersion>14</MinimumVisualStudioVersion>
     <FileAlignment>512</FileAlignment>
--- a/installer/dev/WindowsAppRuntimeInstall.vcxproj
+++ b/installer/dev/WindowsAppRuntimeInstall.vcxproj
@@ -11,7 +11,7 @@
     <ProjectGuid>{e6e59b30-9f55-4550-aa73-3b3b3dc89872}</ProjectGuid>
     <Keyword>Win32Proj</Keyword>
     <RootNamespace>WindowsAppRuntimeInstall</RootNamespace>
-    <WindowsTargetPlatformVersion Condition=" '$(WindowsTargetPlatformVersion)' == '' ">10.0.22000.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion Condition=" '$(WindowsTargetPlatformVersion)' == '' ">10.0.26100.0</WindowsTargetPlatformVersion>
     <WindowsTargetPlatformMinVersion>10.0.17763.0</WindowsTargetPlatformMinVersion>
     <TargetName>WindowsAppRuntimeInstall-</TargetName>
   </PropertyGroup>


build/AzurePipelinesTemplates/WindowsAppSDK-BuildInstaller-Stage.yml
```diff
@@ -75,4 +75,16 @@ stages:
         arguments: >
           -SdkVersion "10.0.22000"
 
+    # The MMS2022 image used to come with Windows SDK 10.0.26100, but not any more.
```
Scottj1s, line 66: why do we need both 22000 and 26100?

build/AzurePipelinesTemplates/WindowsAppSDK-SetupBuildEnvironment-Steps.yml
```diff
@@ -64,8 +64,8 @@ steps:
     inputs:
       targetType: filePath
       filePath: build\scripts\Install-WindowsSdkISO.ps1
-      arguments: 18362
-    displayName: 'Install Insider SDK (18362)'
+      arguments: 26100
+    displayName: 'Install Insider SDK (26100)'
```
Scottj1s: why do we need both 17763 and 26100?

Scottj1s: nit: not an Insider SDK - latest published version on aka.ms/windowssdk, for 24h2

build/AzurePipelinesTemplates/WindowsAppSDK-StaticValidationTest-Stage.yml
```diff
@@ -77,6 +77,18 @@ stages:
         arguments: >
           -SdkVersion "10.0.22000"
 
+    # The MMS2022 image used to come with Windows SDK 10.0.26100, but not any more.
```
Scottj1s, line 68: why do we need both 22000 and 26100?

build/NuSpecs/AppxManifest.xml
```diff
@@ -12,7 +12,7 @@
     <Resource Language="en-US" />
   </Resources>
   <Dependencies>
-    <TargetDeviceFamily Name="Windows.Universal" MinVersion="10.0.17763.0" MaxVersionTested="10.0.18362.0" />
+    <TargetDeviceFamily Name="Windows.Universal" MinVersion="10.0.17763.0" MaxVersionTested="10.0.26100.4654" />
```
Scottj1s: we talked about this before - the only externally visible effects of maxversiontested are:

- potential quirking behavior (not implemented)
- available appx manifest schema elements (not currently used)

so what's the purpose here?

build/scripts/Install-WindowsSdkISO.ps1
```diff
@@ -265,14 +265,14 @@ if ($InstallWindowsSDK)
     # Static(ish) link for Windows SDK
     # Note: there is a delay from Windows SDK announcements to availability via the static link
     $uri = ""
-    if ($buildNumber -eq "18362")
+    if ($buildNumber -eq "26100")
```
Scottj1s, line 268: curious scripting - basically an assert that we can/should only build with 26100?

build/scripts/windows-sdk.ps1
```diff
@@ -120,6 +120,21 @@ elseif ($SdkVersion -eq "10.0.22000")
         exit $sdkExitCode
     }
 }
+elseif ($SdkVersion -eq "10.0.26100")
```
Scottj1s, line 108: are lines 108-122 cruft now?  (remove)

dev/Deployment/DeploymentManager.cpp
```diff
@@ -165,12 +166,12 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
         winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::DeploymentInitializeOptions const& deploymentInitializeOptions,
         bool isRepair)
     {
-        if (isAutoInitialized)
+        auto lock{ m_isInitializedLock.lock_exclusive() };
```
Scottj1s, line 169: I don't think this is required in practice, given entry point startup is single-threaded.  But a little paranoia never hurts.  (i.e, the scenario we were seeing was auto-init + explicit call.  this locking would protect against no auto-init and multiple threads racing with an explicit call, which seems... very unlikely)

dev/Deployment/DeploymentManager.cpp
```diff
@@ -165,12 +166,12 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
...
+
+        if (s_isInitialized)
         {
-            HRESULT hr = HRESULT_FROM_WIN32(E_UNEXPECTED);
-            winrt::hstring message = L"DeploymentManager can only be initialized once and has already been initialized by default. "
-                L"Please either remove your explicit initialization call, or disable auto-initialization by adding <WindowsAppSdkDeploymentManagerInitialize>false</WindowsAppSdkDeploymentManagerInitialize> to your .csproj file.";
-            throw winrt::hresult_error(hr, message);
+            throw winrt::hresult_error(HRESULT_FROM_WIN32(ERROR_INVALID_STATE),
+                                       L"DeploymentManager is already initialized (can only be initialized once). See https://learn.microsoft.com/windows/windows-app-sdk/api/winrt/microsoft.windows.applicationmodel.windowsappruntime.deploymentmanager.initialize?view=windows-app-sdk-1.7#microsoft-windows-applicationmodel-windowsappruntime-deploymentmanager-initialize");
```
Scottj1s, line 174: This comment is not as helpful/actionable as the original.  A user can only get into this situation (which, btw, resulted in a failfast due to a WIL tracelogging bug) by combining auto-init with an explicit call.  It's useful to point that out.  

The URL is a good addition.

build/AzurePipelinesTemplates/WindowsAppSDK-BuildInstaller-Stage.yml
DrusTheAxe: We shouldn't need 22000. I didn't remove that as it wasn't blocking but sure, it's unnecessary so may as well cut it.

build/AzurePipelinesTemplates/WindowsAppSDK-SetupBuildEnvironment-Steps.yml
DrusTheAxe: 17763 is referenced more widely across winappsdk, devcheck already checks and can install it so it wasn't a blocker

Given the repo's unbuild-able if you don't already have 18365+22000 SDKs on your machine I'd like to get this PR in as-is and we can make a follow up to purge 17763 and only need vLatest (26100)

DrusTheAxe: Lolz 18362 = 19H1 which hasn't been insider for years (2019...)

I'll update the comment

build/AzurePipelinesTemplates/WindowsAppSDK-StaticValidationTest-Stage.yml
DrusTheAxe: See previous comment for why and will change

build/NuSpecs/AppxManifest.xml
DrusTheAxe: Was swept up in the global search for 18365

On a more rational point, this is our test and we routinely tell people to set MaxVersionTested to their highest known version (to avoid quirking etc, and more importantly, prevent newer manifest features from behind hidden/blocked from the dev).

On the gripping hand <g> all our test packages should probably use MaxVerisonTested=26100(new-ish).

Fourthly, one could make the argument we should do both -- run all tests with Max=Min and again with Max=10.0.65535.0. But this is a lot of work for little gain, given how Quirk's aren't really a thing anymore (haven't been in years - we have better ways to solve those problems) so it's really the manifest justification - but the developer can figure out something that works and, long as the manifest doesn't change ('cause new development) then it's fine.

I'll revert this change since it shouldn't matter here, and leave it to you to chew on the larger strategic question.

build/scripts/Install-WindowsSdkISO.ps1
DrusTheAxe: Got me, it was there before me, and the purpose of this script is just to install the specific PSDK version if need be

DevCheck should handle this (now) so we can probably scrap this whole script and instead use `DevCheck -CheckWindowsSDK -InstallWindowsSDK`. Then build pipeline and dev inner loop use same solution. But my focus right now was to unblock builds that right now devs cannot build. Refactoring / cleanup / sanitizing the repo for the longer term is a good idea for a future PR and not the urgent problem

build/scripts/windows-sdk.ps1
DrusTheAxe: Probably. Arguably the whole file should be and we switch to leveraging DevCheck

I'm focused on urgently unblocking the blocker so this cleanup should be in a later PR

dev/Deployment/DeploymentManager.cpp
DrusTheAxe: >given entry point startup is single-threaded

What entry point? Nothing stops 2 threads from instantiating DeploymentManager and calling Initialize concurrently. Hence the standard guidance to protect globals against concurrency misuse.

Unlikely a problem, for well intentioned devs, but these are the sorts of things malware looks for. Best to follow the usual guidance than learn later someone intentionally forced the dice to come up snakeeyes for fun and profit.

dev/Deployment/DeploymentManager.cpp
DrusTheAxe: A user gets in this situation if Initialize is called twice. AutoInit+Initialize() or 2xInitialize

The MSBuild property is good but only applies to _some_ depending on your build process. Better to point folks to the docs (the URL) and _the docs_ can (and should) talk about the MSBuild properties (or better, a link to a page about the AutoInitializer that also mentions the MSBuild property) and wax expansive

Scottj1s (approved):

tools/DevCheck/DevCheck.ps1
```diff
@@ -707,7 +708,7 @@ function Test-WindowsSDKInstall
     }
     else
     {
-        Write-Host "...ERROR: Windows SDK $($version) not found or valid. See [Getting Started doc's Tooling Prerequisites](https://github.com/microsoft/WindowsAppSDK/blob/main/docs/Coding-Guidelines/GettingStarted.md#tooling-prerequisites)" -ForegroundColor Red -BackgroundColor Black
+        Write-Host "...ERROR: Windows SDK $($version) not found or valid. Run with -InstallWindowsSDK to download and install missing SDK(s). For more information see https://github.com/microsoft/WindowsAppSDK/blob/main/docs/Coding-Guidelines/GettingStarted.md#tooling-prerequisites" -ForegroundColor Red -BackgroundColor Black
```
guimafelipe, line 711: Could we install it automatically from this script in case it is missing instead of needing the -InstallWindowsSDK flag?

DrusTheAxe: We could, and probably will at a future date, but that has implications I didn't want to try to wrangle just yet. There's some other oddities to cleanup in the future, e.g. the build pipeline downloads the SDK .iso but DevCheck uses setup.exe.

The urgent issue is to unblock devs who can't build if they don't already happen to have the needed SDK(s) on their machine. A future PR will add further cleanup and improvements.

Scottj1s (approved):

haonanttt:
Since we didn't keep the `10.0.22000` case in [windows-sdk.ps1](https://github.com/microsoft/WindowsAppSDK/blob/eb97d4d2b4e740e2daa5467c7ebbc6281c281de4/build/scripts/windows-sdk.ps1#L108). We should also modify [WindowsAppSDK-StaticValidationTest-Stage.yml](https://github.com/microsoft/WindowsAppSDK/blob/eb97d4d2b4e740e2daa5467c7ebbc6281c281de4/build/AzurePipelinesTemplates/WindowsAppSDK-StaticValidationTest-Stage.yml#L101) as well? Currently it actually continue to fail in our Nightly pipeline (although in pipeline it somehow shows as warning)

PR #5771, opened 2025-08-29
API review for new properties in Storage.Pickers - SuggestedDefaultFolder, FileTypeChoices

DinahK-2SO:
This is a design spec for adding new functionality to the `Microsoft.Windows.Storage.Pickers` APIs, including 2 main changes:
1. the `SuggestedFolder` and `SuggestedStartFolder` properties for all three picker types, 
2. the new `FileTypeChoices` property for `FileOpenPicker`. The documentation and code samples are updated to reflect these changes, with detailed explanations of how each property affects picker behavior.

**New and enhanced picker properties:**

* Added `SuggestedFolder` and `SuggestedStartFolder` properties to `FileOpenPicker`, `FileSavePicker`, and `FolderPicker`, allowing developers to control both the persistent and initial folder shown when the picker dialog opens. [[1]](diffhunk://#diff-5cfa035ddfdddfd5f0ea035dc2e78ac86e1b55c323ed0dd06dc9c8bcbf3b5869R22-R29) [[2]](diffhunk://#diff-bcc29af476dc86f30a02d3d5b1dd7590c05e0c8bbae63e7c8bdc9a10feb6f3ecL22-R26) [[3]](diffhunk://#diff-e80bc34d75ca387aa052733c5c44bf7a036fe36ada3021dbb0f3371bcdb4f181R23-R26) [[4]](diffhunk://#diff-bba148e5db9446f11863ced91e43406bb3c998bf5475e491d7086a5ed47b9fe0R122-R129) [[5]](diffhunk://#diff-bba148e5db9446f11863ced91e43406bb3c998bf5475e491d7086a5ed47b9fe0L126-R147) [[6]](diffhunk://#diff-bba148e5db9446f11863ced91e43406bb3c998bf5475e491d7086a5ed47b9fe0R159-R189)
* Updated documentation and code samples to demonstrate usage and precedence rules for `SuggestedFolder` and `SuggestedStartFolder`, including detailed comments and a new explanatory note on their differences and interaction with user navigation history. [[1]](diffhunk://#diff-5cfa035ddfdddfd5f0ea035dc2e78ac86e1b55c323ed0dd06dc9c8bcbf3b5869R47-R58) [[2]](diffhunk://#diff-5cfa035ddfdddfd5f0ea035dc2e78ac86e1b55c323ed0dd06dc9c8bcbf3b5869R91-R102) [[3]](diffhunk://#diff-5cfa035ddfdddfd5f0ea035dc2e78ac86e1b55c323ed0dd06dc9c8bcbf3b5869R68-R74) [[4]](diffhunk://#diff-5cfa035ddfdddfd5f0ea035dc2e78ac86e1b55c323ed0dd06dc9c8bcbf3b5869R112-R117) [[5]](diffhunk://#diff-bcc29af476dc86f30a02d3d5b1dd7590c05e0c8bbae63e7c8bdc9a10feb6f3ecR43-R54) [[6]](diffhunk://#diff-bcc29af476dc86f30a02d3d5b1dd7590c05e0c8bbae63e7c8bdc9a10feb6f3ecL50-L53) [[7]](diffhunk://#diff-bcc29af476dc86f30a02d3d5b1dd7590c05e0c8bbae63e7c8bdc9a10feb6f3ecR87-R98) [[8]](diffhunk://#diff-bcc29af476dc86f30a02d3d5b1dd7590c05e0c8bbae63e7c8bdc9a10feb6f3ecL86-L89) [[9]](diffhunk://#diff-e80bc34d75ca387aa052733c5c44bf7a036fe36ada3021dbb0f3371bcdb4f181R48-R59) [[10]](diffhunk://#diff-e80bc34d75ca387aa052733c5c44bf7a036fe36ada3021dbb0f3371bcdb4f181R82-R93) [[11]](diffhunk://#diff-e80bc34d75ca387aa052733c5c44bf7a036fe36ada3021dbb0f3371bcdb4f181R154-R157) [[12]](diffhunk://#diff-bba148e5db9446f11863ced91e43406bb3c998bf5475e491d7086a5ed47b9fe0R31-R32) [[13]](diffhunk://#diff-bba148e5db9446f11863ced91e43406bb3c998bf5475e491d7086a5ed47b9fe0R51-R59) [[14]](diffhunk://#diff-bba148e5db9446f11863ced91e43406bb3c998bf5475e491d7086a5ed47b9fe0R159-R189)

**File type filtering improvements:**

* Introduced the `FileTypeChoices` property to `FileOpenPicker`, enabling grouped and labeled file type filters in the picker dialog. When both `FileTypeChoices` and `FileTypeFilter` are set, `FileTypeChoices` takes precedence. [[1]](diffhunk://#diff-5cfa035ddfdddfd5f0ea035dc2e78ac86e1b55c323ed0dd06dc9c8bcbf3b5869R22-R29) [[2]](diffhunk://#diff-5cfa035ddfdddfd5f0ea035dc2e78ac86e1b55c323ed0dd06dc9c8bcbf3b5869R68-R74) [[3]](diffhunk://#diff-5cfa035ddfdddfd5f0ea035dc2e78ac86e1b55c323ed0dd06dc9c8bcbf3b5869R112-R117) [[4]](diffhunk://#diff-bba148e5db9446f11863ced91e43406bb3c998bf5475e491d7086a5ed47b9fe0R51-R59) [[5]](diffhunk://#diff-bba148e5db9446f11863ced91e43406bb3c998bf5475e491d7086a5ed47b9fe0R122-R129)

**Documentation and API clarification:**

* Expanded API documentation to clarify the behaviors and precedence of the new properties, including a note distinguishing `SuggestedFolder` (persistent folder) from `SuggestedStartFolder`/`SuggestedStartLocation` (initial folder before user navigation). [[1]](diffhunk://#diff-bba148e5db9446f11863ced91e43406bb3c998bf5475e491d7086a5ed47b9fe0R159-R189) [[2]](diffhunk://#diff-e80bc34d75ca387aa052733c5c44bf7a036fe36ada3021dbb0f3371bcdb4f181R154-R157)
* Updated conceptual and migration notes to highlight the new properties and their intended scenarios, such as supporting persistent folder selection across picker sessions. [[1]](diffhunk://#diff-bba148e5db9446f11863ced91e43406bb3c998bf5475e491d7086a5ed47b9fe0R31-R32) [[2]](diffhunk://#diff-bba148e5db9446f11863ced91e43406bb3c998bf5475e491d7086a5ed47b9fe0R51-R59)

These changes make it much easier to control folder selection behavior in desktop apps, provide better user experiences, and support more advanced filtering scenarios.

### DIFF
--- a/specs/Storage.Pickers/FileOpenPicker.md
+++ b/specs/Storage.Pickers/FileOpenPicker.md
@@ -23,7 +23,7 @@ runtimeclass FileOpenPicker
     IMap<String, IVector<String>> FileTypeChoices{ get; };
     IVector<string> FileTypeFilter{ get; };
 
-    String SuggestedDefaultFolder;
+    String SuggestedStartFolder;
     PickerLocationId SuggestedStartLocation;
     
     PickerViewMode ViewMode;
@@ -46,7 +46,7 @@ var openPicker = new FileOpenPicker(this.AppWindow.Id)
     // (Optional) set an initial folder by absolute path. 
     //     Takes precedence over SuggestedStartLocation when both defined.
     //     If this folder is not found, falls back to SuggestedStartLocation.
-    SuggestedDefaultFolder = @"C:\\MyFiles",
+    SuggestedStartFolder = @"C:\\MyFiles",
 
     // (Optional) Specify the initial location for the picker. 
     //     If the specified location doesn't exist on the user's machine, it falls back to the DocumentsLibrary.
@@ -84,7 +84,7 @@ FileOpenPicker openPicker(AppWindow().Id());
 // (Optional) set an initial folder by absolute path. 
 //     Takes precedence over SuggestedStartLocation when both defined.
 //     If not found, falls back to SuggestedStartLocation.
-openPicker.SuggestedDefaultFolder(L"C:\\MyFiles");
+openPicker.SuggestedStartFolder(L"C:\\MyFiles");
 
 // (Optional) Specify the initial location for the picker. 
 //     If the specified location doesn't exist on the user's machine, it falls back to the DocumentsLibrary.
--- a/specs/Storage.Pickers/FileOpenPicker.md
+++ b/specs/Storage.Pickers/FileOpenPicker.md
@@ -23,6 +23,7 @@ runtimeclass FileOpenPicker
     IMap<String, IVector<String>> FileTypeChoices{ get; };
     IVector<string> FileTypeFilter{ get; };
 
+    string SuggestedFolder;
     String SuggestedStartFolder;
     PickerLocationId SuggestedStartLocation;
     
--- a/specs/Storage.Pickers/FileOpenPicker.md
+++ b/specs/Storage.Pickers/FileOpenPicker.md
@@ -46,7 +46,7 @@ var openPicker = new FileOpenPicker(this.AppWindow.Id)
     // (Optional) set an initial folder by absolute path. 
     //     Takes precedence over SuggestedStartLocation when both defined.
     //     If this folder is not found, falls back to SuggestedStartLocation.
-    SuggestedStartFolder = @"C:\\MyFiles",
+    SuggestedStartFolder = @"C:\MyFiles",
 
     // (Optional) Specify the initial location for the picker. 
     //     If the specified location doesn't exist on the user's machine, it falls back to the DocumentsLibrary.
@@ -59,8 +59,7 @@ var openPicker = new FileOpenPicker(this.AppWindow.Id)
 
     // (Optional) group file types into labeled choices
     //     FileTypeChoices takes precedence over FileTypeFilter when both defined.
-    FileTypeChoices = new Dictionary<string, IList<string>>
-    {
+    FileTypeChoices = {
         { "Documents", new List<string> { ".txt", ".doc", ".docx" } },
         { "Pictures", new List<string> { ".png", ".jpg", ".jpeg", ".bmp" } }
     },
--- a/specs/Storage.Pickers/FileOpenPicker.md
+++ b/specs/Storage.Pickers/FileOpenPicker.md
@@ -19,8 +19,13 @@ runtimeclass FileOpenPicker
     FileOpenPicker(Microsoft.UI.WindowId windowId);
 
     string CommitButtonText;
+
+    IMap<String, IVector<String>> FileTypeChoices{ get; };
     IVector<string> FileTypeFilter{ get; };
+
+    String SuggestedDefaultFolder;
     PickerLocationId SuggestedStartLocation;
+    
     PickerViewMode ViewMode;
 
     Windows.Foundation.IAsyncOperation<PickFileResult> PickSingleFileAsync();
@@ -38,6 +43,11 @@ using Microsoft.Windows.Storage.Pickers;
 
 var openPicker = new FileOpenPicker(this.AppWindow.Id)
 {
+    // (Optional) set an initial folder by absolute path. 
+    //     Takes precedence over SuggestedStartLocation when both defined.
+    //     If this folder is not found, falls back to SuggestedStartLocation.
+    SuggestedDefaultFolder = @"C:\\MyFiles",
+
     // (Optional) Specify the initial location for the picker. 
     //     If the specified location doesn't exist on the user's machine, it falls back to the DocumentsLibrary.
     //     If not set, it defaults to PickerLocationId.Unspecified, and the system will use its default location.
@@ -46,6 +56,14 @@ var openPicker = new FileOpenPicker(this.AppWindow.Id)
     // (Optional) specify the text displayed on the commit button. 
     //     If not specified, the system uses a default label of "Open" (suitably translated).
     CommitButtonText = "Choose selected files",
+
+    // (Optional) group file types into labeled choices
+    //     FileTypeChoices takes precedence over FileTypeFilter when both defined.
+    FileTypeChoices = new Dictionary<string, IList<string>>
+    {
+        { "Documents", new List<string> { ".txt", ".doc", ".docx" } },
+        { "Pictures", new List<string> { ".png", ".jpg", ".jpeg", ".bmp" } }
+    },
 
     // (Optional) specify file extension filters. If not specified, defaults to all files (*.*).
     FileTypeFilter = { ".txt", ".pdf", ".doc", ".docx" },
@@ -63,6 +81,11 @@ using namespace winrt::Microsoft::Windows::Storage::Pickers;
 
 FileOpenPicker openPicker(AppWindow().Id());
 
+// (Optional) set an initial folder by absolute path. 
+//     Takes precedence over SuggestedStartLocation when both defined.
+//     If not found, falls back to SuggestedStartLocation.
+openPicker.SuggestedDefaultFolder(L"C:\\MyFiles");
+
 // (Optional) Specify the initial location for the picker. 
 //     If the specified location doesn't exist on the user's machine, it falls back to the DocumentsLibrary.
 //     If not set, it defaults to PickerLocationId.Unspecified, and the system will use its default location.
@@ -71,6 +94,12 @@ openPicker.SuggestedStartLocation(PickerLocationId::DocumentsLibrary);
 // (Optional) specify the text displayed on the commit button. 
 //     If not specified, the system uses a default label of "Open" (suitably translated).
 openPicker.CommitButtonText(L"Choose selected files");
+
+// (Optional) group file types into labeled choices
+//     FileTypeChoices takes precedence over FileTypeFilter when both defined.
+auto choices = openPicker.FileTypeChoices();
+choices.Insert(L"Documents", winrt::single_threaded_vector<winrt::hstring>({ L".txt", L".doc", L".docx" }));
+choices.Insert(L"Pictures", winrt::single_threaded_vector<winrt::hstring>({ L".png", L".jpg", L".jpeg", L".bmp" }));
 
 // (Optional) specify file extension filters. If not specified, defaults to all files (*.*).
 openPicker.FileTypeFilter().ReplaceAll({ L".txt", L".pdf", L".doc", L".docx" });
--- a/specs/Storage.Pickers/FileOpenPicker.md
+++ b/specs/Storage.Pickers/FileOpenPicker.md
@@ -43,7 +43,14 @@ using Microsoft.Windows.Storage.Pickers;
 
 var openPicker = new FileOpenPicker(this.AppWindow.Id)
 {
-    // (Optional) set an initial folder by absolute path. 
+    // (Optional) Sets the folder that the file dialog always tries to display when it opens.
+    //     SuggestedFolder will not be overriden by the last picked folder.
+    //     If not specified, or the specified path doesn't exist, defaults to the last folder the user picked.
+    //     On first launch of the picker, SuggestedFolder takes precedence over the SuggestedStartFolder if both set.
+    SuggestedFolder = @"C:\MyFiles",
+
+    // (Optional) Sets an initial folder path shown when the picker is first launched.
+    //     Once the user has picked from a directory, SuggestedStartFolder will be silently ignored.
     //     Takes precedence over SuggestedStartLocation when both defined.
     //     If this folder is not found, falls back to SuggestedStartLocation.
     SuggestedStartFolder = @"C:\MyFiles",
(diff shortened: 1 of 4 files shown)


specs/Storage.Pickers/FileOpenPicker.md
```diff
@@ -19,8 +19,13 @@ runtimeclass FileOpenPicker
     FileOpenPicker(Microsoft.UI.WindowId windowId);
 
     string CommitButtonText;
+
+    IMap<String, IVector<String>> FileTypeChoices{ get; };
```
Scottj1s, line 23: Consider renaming to FileTypeFilterGroups to clarify the hierarchical structure, and the override connection with FileTypeFilter.

specs/Storage.Pickers/FileOpenPicker.md
```diff
@@ -38,6 +43,11 @@ using Microsoft.Windows.Storage.Pickers;
 
 var openPicker = new FileOpenPicker(this.AppWindow.Id)
 {
+    // (Optional) set an initial folder by absolute path. 
+    //     Takes precedence over SuggestedStartLocation when both defined.
+    //     If this folder is not found, falls back to SuggestedStartLocation.
+    SuggestedDefaultFolder = @"C:\\MyFiles",
```
Scottj1s: Consider renaming to SuggestedStartFolder, to reenforce its connection with, and generalization of, SuggestedStartLocation.  "Default" and "Start" could convey different behaviors, rather than one being an override of the other.

specs/Storage.Pickers/FileOpenPicker.md
DinahK-2SO: Thanks @Scottj1s , the name `FileTypeChoices` aligns with the **existing** property `FileSavePicker.FileTypeChoices`. Wouldn't it be more intuitive for developers to see the same property name?

specs/Storage.Pickers/FileOpenPicker.md
DinahK-2SO: Thanks @Scottj1s for this good idea! the name `SuggestedStartFolder` also aligns with the property name `SuggestedStartLocation`. I will update.

specs/Storage.Pickers/FileOpenPicker.md
Scottj1s: Ah I missed that - yes, let's stick with FileTypeChoices for continuity

specs/Storage.Pickers/FileOpenPicker.md
DinahK-2SO: updated.

specs/Storage.Pickers/FileOpenPicker.md
```diff
@@ -38,6 +43,11 @@ using Microsoft.Windows.Storage.Pickers;
 
 var openPicker = new FileOpenPicker(this.AppWindow.Id)
 {
+    // (Optional) set an initial folder by absolute path. 
+    //     Takes precedence over SuggestedStartLocation when both defined.
+    //     If this folder is not found, falls back to SuggestedStartLocation.
+    SuggestedStartFolder = @"C:\\MyFiles",
```
codendone: I think `@"C:\\MyFiles"` should have a single backslash, given the '@'. (Same comment also for the other pickers.)

specs/Storage.Pickers/FileOpenPicker.md
```diff
@@ -47,6 +57,14 @@ var openPicker = new FileOpenPicker(this.AppWindow.Id)
     //     If not specified, the system uses a default label of "Open" (suitably translated).
     CommitButtonText = "Choose selected files",
 
+    // (Optional) group file types into labeled choices
+    //     FileTypeChoices takes precedence over FileTypeFilter when both defined.
+    FileTypeChoices = new Dictionary<string, IList<string>>
+    {
+        { "Documents", new List<string> { ".txt", ".doc", ".docx" } },
+        { "Pictures", new List<string> { ".png", ".jpg", ".jpeg", ".bmp" } }
```
oldnewthing, line 72: This design is tricky because the app is providing the IList, and it could provide one that behaves erratically. (E.g., reports the wrong Size, or one that generates different extensions each time you ask, or one that makes a cross-process COM call when you ask for an item.) We have to be careful how we access the list elements to be resilient to misbehaving list implementations.

```csharp
class WeirdList : IList<string>
{
    public int Count => Math.Random(0, 10);
    public bool IsReadOnly => true;
    public Item[int index] => GenerateRandomString();
}
```

Also, how do you set this in JavaScript or Python? I don't think they have a way to create IList implementations.

But it seems that this already exists in the FileSavePicker, so hopefully we can use the same solution.

specs/Storage.Pickers/FileOpenPicker.md
```diff
@@ -63,6 +81,11 @@ using namespace winrt::Microsoft::Windows::Storage::Pickers;
 
 FileOpenPicker openPicker(AppWindow().Id());
 
+// (Optional) set an initial folder by absolute path. 
+//     Takes precedence over SuggestedStartLocation when both defined.
+//     If not found, falls back to SuggestedStartLocation.
+openPicker.SuggestedStartFolder(L"C:\\MyFiles");
```
oldnewthing, line 101: Maybe `SuggestedStartFolderPath` to emphasize that it's a path, and not an `IStorageFolder`.

specs/Storage.Pickers/FileSavePicker.md
```diff
@@ -85,7 +96,7 @@ savePicker.SuggestedFileName(L"NewDocument");
 
 // (Optional) Sets the folder that the file save dialog displays when it opens.
 //     If not specified or the specified path doesn't exist, defaults to the last folder the user visited.
-savePicker.SuggestedFolder = L"C:\\MyFiles",
+savePicker.SuggestedFolder(L"C:\\MyFiles");
```
oldnewthing: What is the difference between FileSavePicker.SuggestedStartFolder and FileSavePicker.SuggestedFolder? They seem to do the same thing.

Should we simply continue using FileSavePicker.SuggestedFolder (and not add FileSavePicker.SuggestedStartFolder), and add FileOpenPicker.SuggestedFolder + FolderPicker.SuggestedFolder?

DinahK-2SO: Hi @oldnewthing , Thanks for raising this point!

`SuggestedFolder` means the path will be always tried to open, regardless of user's operation. ( Under the hood, it uses [the `SetFolder` method of COM APIs](https://learn.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-ifiledialog-setfolder))

Meanwhile, `SuggestedStartFolder` is the path shown the first time the user launches the picker - usually when the app is newly installed. After user has picked a file, if the picker is launched again later, it will open the user's last selected folder, and `SuggestedStartFolder` will be silent. ( It corresponds to [the `SetDefaultFolder` method in the COM API](https://learn.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-ifiledialog-setdefaultfolder) . )

The effective time span of `SuggestedStartFolder` is the same as that of `SuggestedStartLocation`.

specs/Storage.Pickers/FileOpenPicker.md
DinahK-2SO: My mistake - I'll update it. Thanks for pointing it out! @codendone

specs/Storage.Pickers/FileOpenPicker.md
DinahK-2SO, line 23: Hi @oldnewthing , the values of FileTypeChoices are `IVector` instead of `IList`. Would `IVector` help avoid the issues you mentioned below with `IList`? Just wanted to confirm if this design is safe. `IVector` is also the interface for existing `FileSavePicker.FileTypeChoices`

specs/Storage.Pickers/FileOpenPicker.md
DinahK-2SO: Hi @oldnewthing , thanks for pointing out. The API here should be `IVector` instead of `IList`. The sample code is wrong.

specs/Storage.Pickers/FileOpenPicker.md
DinahK-2SO: Hi @oldnewthing , `FileSavePicker` already includes a `SuggestedFolder` property, which is also of type string. To maintain consistency, would it make sense to name the new property `SuggestedStartFolder` instead of `SuggestedStartFolderPath`? Happy to hear your thoughts!

specs/Storage.Pickers/FileSavePicker.md
DinahK-2SO: Adding this explanation in example code comments.

specs/Storage.Pickers/FileOpenPicker.md
DinahK-2SO: Fixed in latest iteration.

specs/Storage.Pickers/FileOpenPicker.md
DinahK-2SO: Fixed in latest iteration.

specs/Storage.Pickers/FileOpenPicker.md
kmahone: RECOMMENDED: Keep FileTypeChoices

specs/Storage.Pickers/FileOpenPicker.md
kmahone: RECOMMENDED: fix.

specs/Storage.Pickers/FileOpenPicker.md
kmahone: RECOMMENDED: SuggestedStartFolder

specs/Storage.Pickers/FileSavePicker.md
kmahone: I think this is a pretty subtle distinction that would not be obvious to someone casually reading the descriptions. I would update the Notes section with a paragraph explaining the functional difference between the two.

RECOMMENDED: Clarify distinction between SuggestedStartFolder and SuggestedFolder.

specs/Storage.Pickers/FileOpenPicker.md
kmahone: Since we have already shipped with SuggestedFolder, let's maintain consistency here.

RECOMMENDED: keep SuggestedStartFolder.

specs/Storage.Pickers/FileOpenPicker.md
kmahone: Dinah: note that IVector<T> is a WinRT interface and will appear in the definition in the runtimeclass.
IList<T> is a .net type and it is used as the projection of the WinRT IVector<T> into C#.

So in your sample C# code you will use IList<T> (or an implementation of that interface such as List<T>).

specs/Storage.Pickers/FileOpenPicker.md
kmahone: The proposed api is:

```
IMap<string, IVector<string>> FileTypeChoices{ get; };
```
This matches the existing Windows.Storage.Pickers.FileSavePicker. 
We could diverge from what is in the system type if there was a strong motivation to do so but the current proposal seems reasonable.

RECOMMENDED: stick with IVector

specs/Storage.Pickers/FileOpenPicker.md
DinahK-2SO: Thanks @kmahone , it is fixed.

specs/Storage.Pickers/FileSavePicker.md
```diff
@@ -23,12 +23,26 @@ runtimeclass FileSavePicker
...
+    String SuggestedStartFolder;
     PickerLocationId SuggestedStartLocation;
 
     Windows.Foundation.IAsyncOperation<PickFileResult> PickSaveFileAsync();
 }
 ```
 
+Notes: **Understanding SuggestedStartFolder/SuggestedStartLocation vs SuggestedFolder:**
+
+These two properties have fundamentally different behaviors in terms of when and how they affect the picker:
```
kmahone: >These [](http://example.com/codeflow?start=0&length=5)

Add line breaks to wrap these lines at a reasonable width to improve readability.

kmahone (approved):
:shipit:

DinahK-2SO:
Hi @kmahone and API reviewers,
 
Good day! I came across [a request (#5836 comment-3291107774)](https://github.com/microsoft/WindowsAppSDK/issues/5836#issuecomment-3291107774) for the SuggestedFolder functionality in the open picker too (currently, this property is only available in the save picker). To address this need, I’m adding the `SuggestedFolder` property in both `FileOpenPicker` and `FolderPicker`.

Please take a moment to review the updated proposal when you get a chance - your feedback is much appreciated!

specs/Storage.Pickers/Microsoft.Windows.Storage.Pickers.md
```diff
@@ -138,10 +156,34 @@ namespace Microsoft.Windows.Storage.Pickers
...
 ```
+
+Note: **Understanding SuggestedStartFolder/SuggestedStartLocation vs SuggestedFolder:**
+
+These two kinds of properties have fundamentally different behaviors in terms of when and how they 
+affect the picker:
+
+- `SuggestedFolder` sets the path that will always be tried when opening the picker, regardless of 
+    the user's previous operations. This uses the [SetFolder](https://learn.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-ifiledialog-setfolder) 
+    method of the underlying COM APIs and takes precedence over any user navigation history.
```
codendone, line 177: This was discussed before, but I'm forgetting the details. If the `SuggestedFolder` exists, are `SuggestedStartFolder` and `SuggestedStartLocation` both ignored/unused?

mominshaikhdevs: @DinahK-2SO

Marv51:
Just curious, do the changes to FileTypeChoices here already address #5827? 

Basically, the developer seems to have no control over the order and the default selection of FileTypeChoices in FileSavePicker, which is a regression from the WinRT pickers.

DinahK-2SO:
> Just curious, do the changes to FileTypeChoices here already address #5827?
> 
> Basically, the developer seems to have no control over the order and the default selection of FileTypeChoices in FileSavePicker, which is a regression from the WinRT pickers.

Thanks @Marv51 for reporting this issue. It hasn’t been resolved yet. It is being looked into and I will share updates as soon as we have them.

PR #5781, opened 2025-09-02
Enabling app notification tests

guimafelipe:
This PR enables App notification tests.

The changes on this PR include:
- Removing the scope `:Class` from the `RunAs` property for the unpackaged tests. This scope prevents the inheritance of the property value by the methods, making the methods being executed in the same scope as the main test, which is `System`. App notifications needs to be run in a non elevated process, which requires integrity level of medium or lower. For that reason, the previous configuration was breaking the tests, so now the test methods are run in `RestrictedUser` scope alongside the class methods, making everything run in the same process.
- Adding the `.testdef` file and copying it as other tests.
- Adding code to skip the tests in older Win10 versions.
- Rewriting the bootstrap initialization code.

WARNING:

### DIFF
--- a/test/AppNotificationTests/AppNotificationTests.testdef
+++ b/test/AppNotificationTests/AppNotificationTests.testdef
@@ -1,0 +1,18 @@
+{
+    "Tests": [
+        {
+            "Description": "App Notification Tests Packaged",
+            "Filename": "AppNotificationTests.dll",
+            "Parameters": "/name:PackagedTests*",
+            "Architectures": ["x64", "x86", "arm64"],
+            "Status": "Enabled"
+        },
+        {
+            "Description": "App Notification Tests Unpackaged",
+            "Filename": "AppNotificationTests.dll",
+            "Parameters": "/name:UnpackagedTests*",
+            "Architectures": ["x64", "x86", "arm64"],
+            "Status": "Enabled"
+        }
+    ]
+}
--- a/test/AppNotificationTests/AppNotificationTests.testdef
+++ b/test/AppNotificationTests/AppNotificationTests.testdef
@@ -12,7 +12,7 @@
             "Filename": "AppNotificationTests.dll",
             "Parameters": "/name:UnpackagedTests*",
             "Architectures": ["x64", "x86", "arm64"],
-            "Status": "Enabled"
+            "Status": "Disabled"
         }
     ]
 }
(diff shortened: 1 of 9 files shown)


test/AppNotificationTests/AppNotificationTests.testdef
```diff
@@ -0,0 +1,18 @@
+{
+    "Tests": [
+        {
+            "Description": "App Notification Tests Packaged",
+            "Filename": "AppNotificationTests.dll",
+            "Parameters": "/name:PackagedTests*",
+            "Architectures": ["x64", "x86"],
```
ssparach, line 7: Are these tests not configured to run for ARM64? Just checking if there's a gap in coverage or if it's intentional.

test/AppNotificationTests/BaseTestSuite.cpp
```diff
@@ -369,7 +384,16 @@ void BaseTestSuite::VerifyRemoveWithIdentifierAsyncUsingNonActiveToastIdentifier
...
 
-    VERIFY_ARE_EQUAL(removeNotificationAsync.wait_for(c_timeout), winrt::Windows::Foundation::AsyncStatus::Completed);
+    // On x86, the wait_for may return Error instead of Completed.
+    // We check if the HR is E_INVALIDARG in that case.
+    auto status = removeNotificationAsync.wait_for(c_timeout);
+    VERIFY_IS_TRUE(status == winrt::Windows::Foundation::AsyncStatus::Completed || status == winrt::Windows::Foundation::AsyncStatus::Error);
+
+    if (status == winrt::Windows::Foundation::AsyncStatus::Error)
+    {
+        VERIFY_THROWS_HR(removeNotificationAsync.GetResults(), E_INVALIDARG);
```
ssparach, line 394: Has there been any investigation into why x86 is exhibiting different behavior or returning an error compared to x64? Catching the error and rethrowing only on INVALID_ARG might inadvertently mask other underlying issues.

test/AppNotificationTests/AppNotificationTests.testdef
guimafelipe: Yes, the tests are currently not being built for arm64.

test/AppNotificationTests/BaseTestSuite.cpp
guimafelipe: I don't know why, but this was happening in two x86 machines in a consistent manner. We can also disable this test if that is desired.

ssparach: Let's keep the current workaround in place to ensure we have a basic test case running. However, we should also create a work item to track investigation for x86 support and add a TODO comment to remove this workaround once resolved. As part of this work item, let's also include looking into building tests for arm64 to ensure broader coverage.

PR #5685, opened 2025-08-05
Fix msbuild warnings

ssparach:
A microsoft employee must use /azp run to validate using the pipelines below.

WARNING:
Comments made by azure-pipelines bot maybe inaccurate.
Please see pipeline link to verify that the build is being ran.

For status checks on the main branch, please use TransportPackage-Foundation-PR
(https://microsoft.visualstudio.com/ProjectReunion/_build?definitionId=81063&_a=summary)
and run the build against your PR branch with the default parameters.

PR #5782, opened 2025-09-02
Fix Warning MSB8027 by replacing PreprocessorDefinitions with AdditionalOptions in Bootstrap target

qiutongMS:
This change addresses MSBuild warning MSB8027 that occurs when multiple targets in BeforeClCompileTargets modify the same parameter.

**Technical Details:**

1. Semantic Equivalence: PreprocessorDefinitions and AdditionalOptions are semantically equivalent for defining preprocessor macros. Both /D MACRO_NAME (AdditionalOptions) and MACRO_NAME;%(PreprocessorDefinitions) achieve the same compilation result.

2. Root Cause: When two targets in BeforeClCompileTargets both modify the same parameter (e.g., PreprocessorDefinitions), the second target to execute may cause duplicate entries to be added to the cl compile command due to parameter modification conflicts.
 
3. Solution: By using different but equivalent parameters (PreprocessorDefinitions vs AdditionalOptions), we can avoid this MSBuild parameter collision issue while maintaining the same functionality. 

**Scope:**

This commit specifically addresses the Bootstrap target. The GenerateUndockedRegFreeWinRTCpp target has the same underlying issue, but modifications to it cause arm64EC build failures in WinUI projects. Further investigation is needed before addressing that target, so this change is submitted as an incremental fix for the Bootstrap scenario.

Testing: This change has been tested to ensure no functional regression while eliminating the MSBuild warning.

WARNING:

### DIFF
--- a/build/NuSpecs/WindowsAppSDK-Nuget-Native.Bootstrap.targets
+++ b/build/NuSpecs/WindowsAppSDK-Nuget-Native.Bootstrap.targets
@@ -4,14 +4,14 @@
         <ItemGroup>
             <ClCompile Include="$(MSBuildThisFileDirectory)..\..\include\MddBootstrapAutoInitializer.cpp">
                 <PrecompiledHeader>NotUsing</PrecompiledHeader>
-                <PreprocessorDefinitions Condition="'$(WindowsAppSDKBootstrapAutoInitializeOptions_Default)'=='true'">MICROSOFT_WINDOWSAPPSDK_BOOTSTRAP_AUTO_INITIALIZE_OPTIONS_DEFAULT;%(PreprocessorDefinitions)</PreprocessorDefinitions>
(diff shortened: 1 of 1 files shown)


Scottj1s:
I'm not confident this is a complete fix.  It's important to understand why modifying PreprocessorDefinitions causes this warning.  It's due to MSBuild batching behavior.  Any time "%" metadata is referenced in a target or task, that target/task is executed for each distinct batch defined by that metadata.  This could occur as easily with AdditionalOptions.  [My recommendation](https://github.com/microsoft/WindowsAppSDK/pull/5685#issuecomment-3229606953) was to ensure that any dynamic preprocessor definitions be applied **_after all sources had been added_**.  This will prevent batching and the subsequent warning.

chwarr:
In the proposed change, the ClCompile item is added inside a target. It doesn't look like any target logic is being used here, so another way to fix this might be to add the ClCompile item outside a target. Then, the [batching multiplicative behavior](https://learn.microsoft.com/visualstudio/msbuild/msbuild-batching?view=vs-2022#item-batching-on-self-referencing-metadata) won't kick in.

PR #5783, opened 2025-09-03
Lowering Build Sample Apps and Test Sample Apps from Aggregator to Foundation

haonanttt:
# Proposal
This PR aims to lower the Sample App Tests (Build sample apps and test launch them) to Foundation to find the Foundation's code error earlier.

Note: This PR's change will bring much running time to pipeline. So our default setting is we don't run Build and Test Sample Apps in PR build, but run them in Nightly and Official

# Changes in this PR
### Build Sample Apps
1. `build/AzurePipelinesTemplates/WindowsAppSDK-BuildSamplesCompat-Job.yml` - Mainly copied from [Aggregator](https://microsoft.visualstudio.com/ProjectReunion/_git/WindowsAppSDKAggregator?path=/build/AzurePipelinesTemplates/WindowsAppSDK-BuildSamplesCompat-Job.yml), the major changes are:
    - Add `IsOneBranch` param (**line 4**). Using this param to decide the pool (**line 37**) and publish build artifact (**line 377**)
    - Separate `UseDotNet@2` task to not using global json (**line 142**)
    - Remove `WindowsAppSDKVerifyTransitiveDependencies` for VSBuild command (**line 233** and **line 249**) since there is no need for ML package
    - Remove `SDLNativeRules@3` except cpp-console-unpackaged (**line 276**) since other .sln will not be built at Foundation
2. `build/WindowsAppSDK-CommonVariables.yml`:
    - Specify Samples Branch
3. `test/ModifySampleAppsReferences.ps1` - Modified from [Samples](https://github.com/microsoft/WindowsAppSDK-Samples/blob/release/experimental/UpdateVersions.ps1), the major changes are:
    - Hardcode the packages to remove (packages that are depended by WinAppSDK but are not depended by Foundation)

### Test Sample Apps
1. `TestAll.ps1`:
    - Add `callingStage` param and filter the tests to run
2. `WindowsAppRuntime.sln`
3. `test/WindowsAppSDK.Test.NetCore/` folder and its contents - Mainly copied from [Aggregator](https://microsoft.visualstudio.com/ProjectReunion/_git/WindowsAppSDKAggregator?path=/test/WindowsAppSDK.Test.NetCore)
4. `test/WindowsAppSDK.Test.SampleTests/` folder and its contents - Mainly copied from [Aggregator](https://microsoft.visualstudio.com/ProjectReunion/_git/WindowsAppSDKAggregator?path=/test/WindowsAppSDK.Test.SampleTests)
    - **2, 3, 4** together generates the test sample apps TAEF test dll
    - For **4**, the major difference compared with Aggregator is: 1) in Foundation, we removed sample apps that cannot be built only with Foundation package from `WindowsAppSDKSampleAppTests.cs` and 2) In `WindowsAppSDK.Test.SampleTests.csproj` we set `<ManagePackageVersionsCentrally>false</ManagePackageVersionsCentrally>` to be `false` as Foundation originally uses centralized version control. And we removed the package reference for WindowsAppSDK
5. `build/AzurePipelinesTemplates/WindowsAppSDK-RunTests-Steps.yml` - Mainly copied from [Aggregator](https://microsoft.visualstudio.com/ProjectReunion/_git/WindowsAppSDKAggregator?path=/build/AzurePipelinesTemplates/WindowsAppSDK-RunTests-Steps.yml)(This PR's **line 97 - line 157** matches Aggregator's **line 117 - line 200**), the major changes are:
    - Manually handles `net6` folder generated by `WindowsAppSDK.Test.NetCore` at **line 218**, to make the `TestAssembly.cs` property `[TestProperty("CoreClrProfile", "net6")]` work
6. `build/AzurePipelinesTemplates/WindowsAppSDK-RunTestsInPipeline-Job.yml`:
    - Add params

### Other Files
1. `build/AzurePipelinesTemplates/WindowsAppSDK-BuildAndTestSampleApps-Stage.yml`:
    - Organize the stages of `BuildSampleApps` and `TestSampleApps`
    - The param `sampleFeatureAreasList` is slightly different compared with [Aggregator](https://microsoft.visualstudio.com/ProjectReunion/_git/WindowsAppSDKAggregator?path=/build/AzurePipelinesTemplates/WindowsAppSDK-BuildWindowsAppSDK-Stages.yml). The major difference is in Foundation we only keep the folder that all of its content could be successfully bulilt
2. `build/AzurePipelinesTemplates/WindowsAppSDK-Publish-Stage.yml`
    - Add params to decide whether the `PublishToMaestro` stage will be blocked by `TestSampleApps` stage. By default the TestSampleApps will not affect the PublishToMaestro, if pipeline user want to have such dependency, he could manually tick the  `MaestroDependOnTestSamples`
3. `build/ProjectReunion-BuildFoundation.yml`
4. `build/WindowsAppSDK-Foundation-Nightly.yml`
5. `build/WindowsAppSDK-Foundation-Official.yml`
6. `build/WindowsAppSDK-Foundation-PR.yml`

# Pending discussion
1. In Foundation, when building the Sample App, package reference changed from `WindowsAppSDK` to `WindowsAppSDK.Foundation`. Therefore the dependency of package `WindowsAppSDK.Runtime` is removed. This causes [MicrosoftWindowsAppSDKPackageDir](https://microsoft.visualstudio.com/ProjectReunion/_search?action=contents&type=code&text=MicrosoftWindowsAppSDKPackageDir&filters=ProjectFilters%7BProjectReunion%7D&result=DefaultCollection/ProjectReunion/WindowsAppSDKAggregator/GBmain//build/NuSpecs/RuntimePackage/build/Microsoft.WindowsAppSDK.Runtime.props&includeFacets=false) still empty, then cause [WindowsAppSDKSelfContained](https://microsoft.visualstudio.com/ProjectReunion/_search?action=contents&type=code&text=WindowsAppSDKSelfContained&filters=ProjectFilters%7BProjectReunion%7D&result=DefaultCollection/ProjectReunion/WindowsAppSDKAggregator/GBmain//build/NuSpecs/base/build/native/Microsoft.WindowsAppSDK.Base.targets&includeFacets=false) to be `true`, when this target being imported to Sample Apps build code, for example [Notification/Push](https://github.com/microsoft/WindowsAppSDK-Samples/blob/275986bb60038280ecdeede66025f50e75fc9183/Samples/Notifications/Push/cpp-console-unpackaged/cpp-console-unpackaged.vcxproj#L228C1-L229C1), this finally causes the `WindowsAppSDK-Nuget-Native.targets` go to [**UndockedRegFreeWinRTCommon.targets**](https://microsoft.visualstudio.com/ProjectReunion/_git/WindowsAppSDK?path=/build/NuSpecs/WindowsAppSDK-Nuget-Native.targets&version=GBmain&line=165&lineEnd=166&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents) path but not [**BootstrapCommon.targets**](https://microsoft.visualstudio.com/ProjectReunion/_git/WindowsAppSDK?path=/build/NuSpecs/WindowsAppSDK-Nuget-Native.targets&version=GBmain&line=162&lineEnd=163&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents) as Aggregator does. Will this target change affect the final test result?
2. In [PR5764](https://github.com/microsoft/WindowsAppSDK/pull/5764), we removed install **Windows SDK 10.0.22000** case in `build/scripts/windows-sdk.ps1`. However, some of our sample apps require Windows SDK 10.0.22000 as target build platform. Shall we bring Windows SDK 10.0.22000 back to `build/scripts/windows-sdk.ps1` or modify the sample app's target build platform?

# Example Run for this PR (since PR build will not run Build and Test Sample Apps by default, we run it manually)
Succeed run without commit clean up:
https://microsoft.visualstudio.com/ProjectReunion/_build/results?buildId=129369163&view=results
After commit clean up and rebase main (not totally succeed because of the second point in Pending Discussion section)
https://microsoft.visualstudio.com/ProjectReunion/_build/results?buildId=129381070


build/AzurePipelinesTemplates/WindowsAppSDK-BuildAndTestSampleApps-Stage.yml
```diff
@@ -0,0 +1,142 @@
...
+    parameters:
+      IsOneBranch: ${{ parameters.IsOneBranch }}
+      JobName: "SamplesCompatTest"
+      FeatureAreas: ${{ parameters.sampleFeatureAreasList.SamplesCompatTest }}
+      BuildConfig:
+      - 'Release'
+      - 'Debug'
+      BuildPlatform:
+      - 'x64'
+      ${{ if or(eq(parameters.TestSampleAppsAfterBuildStage, 'true'), false) }}:
```
alexlamtest: Non-blocking: This if-condition can be simplified to "if eq(parameters.TestSampleAppsAfterBuildStage, 'true')" ?

build/AzurePipelinesTemplates/WindowsAppSDK-BuildAndTestSampleApps-Stage.yml
```diff
@@ -0,0 +1,142 @@
...
+        SamplesArtifactName: ${{ parameters.SamplesBuildOutputArtifactName }}_x64
+      ${{ if and(ne(parameters.TestSampleAppsAfterBuildStage, 'true'), true) }}:
```
alexlamtest: Non-blocking: This if-condition can be simplified to "if ne(parameters.TestSampleAppsAfterBuildStage, 'true')" ?

build/AzurePipelinesTemplates/WindowsAppSDK-BuildAndTestSampleApps-Stage.yml
```diff
@@ -0,0 +1,142 @@
...
+    parameters:
+      IsOneBranch: ${{ parameters.IsOneBranch }}
+      JobName: "SamplesCompatTest"
+      FeatureAreas: ${{ parameters.sampleFeatureAreasList.SamplesCompatTest }}
+      BuildConfig:
+      - 'Release'
+      - 'Debug'
+      BuildPlatform:
+      - 'arm64'
+      ${{ if or(eq(parameters.TestSampleAppsAfterBuildStage, 'true'), false) }}:
```
alexlamtest: Same comment as above.

build/AzurePipelinesTemplates/WindowsAppSDK-BuildAndTestSampleApps-Stage.yml
```diff
@@ -0,0 +1,142 @@
...
+        SamplesArtifactName: ${{ parameters.SamplesBuildOutputArtifactName }}_arm64
+      ${{ if and(ne(parameters.TestSampleAppsAfterBuildStage, 'true'), true) }}:
```
alexlamtest: Same comment as above.

build/WindowsAppSDK-Foundation-PR.yml
```diff
@@ -20,6 +20,26 @@ name: $(version)
 trigger: none
 
 parameters: # parameters are shown up in ADO UI in a build queue time
+- name: "BuildSampleApps"
+  displayName: "Run Build Sample Apps Stage"
+  type: boolean
+  default: true
```
alexlamtest: Shouldn't this default be false, given that we don't want to build sample apps in the PR pipeline?

build/WindowsAppSDK-Foundation-PR.yml
```diff
@@ -20,6 +20,26 @@ name: $(version)
...
+- name: "TestSampleAppsAfterBuildStage"
+  displayName: "Test launch of Sample apps separately in the TestSampleApps Stage"
+  type: boolean
+  default: true
```
alexlamtest: Have we verified that when this parameter is set to false, the pipeline still works as expected?

build/ProjectReunion-BuildFoundation.yml
```diff
@@ -5,6 +5,32 @@ name: $(version)
 
 trigger: none
 
+parameters:
+- name: "BuildSampleApps"
+  displayName: "Test build of sample apps"
+  type: boolean
+  default: false
```
alexlamtest, line 12: Ah, I see that you don't build sample apps by default in this scenario.

build/AzurePipelinesTemplates/WindowsAppSDK-BuildSamplesCompat-Job.yml
```diff
@@ -0,0 +1,388 @@
...
+  # Install Microsoft.WindowsAppSDK.Foundation to get all its dependencies and their version.
+  - task: NuGetCommand@2
+    inputs:
+      command: 'custom'
+      arguments: >
+        install "Microsoft.WindowsAppSDK.Foundation"
+        -Source "$(Build.SourcesDirectory)\localpackages\NugetPackages"
+        -Version "$(FoundationVersion)"
+        -OutputDirectory "$(Build.SourcesDirectory)\$(SamplesRepoName)\Samples\localpackages"
+        -FallbackSource "https://microsoft.pkgs.visualstudio.com/ProjectReunion/_packaging/Project.Reunion.nuget.internal/nuget/v3/index.json"
```
kythant, line 131: I think maybe we should just fail here and not use a FallbackSource if it can't restore from localpackages

build/WindowsAppSDK-Foundation-Nightly.yml
```diff
@@ -33,6 +33,30 @@ parameters:
...
+  default: true
+- name: "TestSampleAppsAfterBuildStage"
+  displayName: "Test launch of Sample apps separately in the TestSampleApps Stage"
+  type: boolean
+  default: true
+- name: "TestOnArm64"
+  displayName: "Enable running of sample tests on arm64 platform (Default: true)"
+  type: boolean
+  default: true
+- name: maxParallelForBuildSamplesCompatJob_x64
```
kythant: I think the number of parameters are getting too overwhelming, considering moving maxParallelForBuildSamplesCompatJob_x64 and maxParallelForBuildSamplesCompatJob_arm64 to a variable set from the pipeline instead. 

Especially if we don't usually change the values of these that often.

build/WindowsAppSDK-Foundation-Official.yml
```diff
@@ -33,6 +33,30 @@ parameters:
...
+- name: "TestSampleAppsAfterBuildStage"
+  displayName: "Test launch of Sample apps separately in the TestSampleApps Stage"
+  type: boolean
+  default: true
+- name: "TestOnArm64"
+  displayName: "Enable running of sample tests on arm64 platform (Default: true)"
+  type: boolean
+  default: true
+- name: maxParallelForBuildSamplesCompatJob_x64
+  displayName: "Maximum number of parallel runs for Samples Compat Tests on platform x64"
```
kythant: I think the number of parameters are getting too overwhelming, considering moving maxParallelForBuildSamplesCompatJob_x64 and maxParallelForBuildSamplesCompatJob_arm64 to a variable set from the pipeline instead. 

Especially if we don't usually change the values of these that often.

build/WindowsAppSDK-Foundation-PR.yml
```diff
@@ -20,6 +20,26 @@ name: $(version)
...
+- name: "TestOnArm64"
+  displayName: "Enable running of sample tests on arm64 platform (Default: true)"
+  type: boolean
+  default: true
+- name: maxParallelForBuildSamplesCompatJob_x64
```
kythant: I think the number of parameters are getting too overwhelming, considering moving maxParallelForBuildSamplesCompatJob_x64 and maxParallelForBuildSamplesCompatJob_arm64 to a variable set from the pipeline instead. 

Especially if we don't usually change the values of these that often.

test/ModifySampleAppsReferences.ps1
```diff
@@ -0,0 +1,95 @@
...
+    [string]$SampleRepoRoot = "",
+    [string]$FoundationVersion = "",
+    [string]$FoundationPackagesFolder = ""
+)
+
+# List all packages that need to be removed because they are dependencies of Microsoft.WindowsAppSDK but not of Microsoft.WindowsAppSDK.Foundation
+$packagesToRemove = @(
+    "Microsoft.WindowsAppSDK",
+    "Microsoft.Web.WebView2",
+    "Microsoft.WindowsAppSDK.AI",
```
kythant: This code won't be very future proof because we will have to keep updating it if we add a new component package package.

Consider using a wildcard

test/WindowsAppSDK.Test.SampleTests/WindowsAppSDKSampleAppTests.cs
```diff
@@ -0,0 +1,1366 @@
...
+using System.Diagnostics;
+using System.IO;
+using System.Runtime.InteropServices;
+using System.Threading;
+using System.Collections.Generic;
+
+using WEX.Logging.Interop;
+using WEX.TestExecution;
+using WEX.TestExecution.Markup;
+
```
kythant, line 18: This is supposed to be identical to the one in Aggregator correct?

What happened to the giant comment that was here?
image

TestAll.ps1
```diff
@@ -140,7 +152,7 @@ function Run-TaefTest
     $teLogFile = (Join-Path $env:Build_SourcesDirectory "BuildOutput\$Configuration\$Platform\Te.wtl")
     $teLogPathTo = (Join-Path $env:Build_SourcesDirectory "TestOutput\$Configuration\$Platform")
 
-    & $tePath $dllFile $test.Parameters /enableWttLogging /appendWttLogging /screenCaptureOnError /logFile:$teLogFile $/testMode:EtwLogger /EtwLogger:WprProfile=WDGDEPAdex /EtwLogger:SavePoint=TestFailure /EtwLogger:RecordingScope=Execution /EtwLogger:WprProfileFile=$wprProfilePath
+    & $tePath $dllFile $test.Parameters /enableWttLogging /appendWttLogging /screenCaptureOnError /logFile:$teLogFile /testMode:EtwLogger /EtwLogger:WprProfile=WDGDEPAdex /EtwLogger:SavePoint=TestFailure /EtwLogger:RecordingScope=Execution /EtwLogger:WprProfileFile=$wprProfilePath
```
alexlamtest, line 155: This looks like a bug fix, thanks!

build/ProjectReunion-BuildFoundation.yml
```diff
@@ -15,11 +41,16 @@ resources:
     - repository: WindowsAppSDKConfig
       type: git
       name: ProjectReunion/WindowsAppSDKConfig
-      ref: refs/heads/main
+      ref: refs/heads/user/haonantang/sampleApps
```
alexlamtest: I presume this change is temporary,

alexlamtest (commented):
- Regarding "default setting is we don't run Build and Test Sample Apps in PR build, but run them in Nightly and Official", it sounds like a good start. If there is demand, we can try testing sample apps on 1 x64 OS image and 1 arm64 OS image in a PR pipeline run, to keep the workload small.
- Regarding "Q1: ... Will this target change affect the final test result?", my understanding is that, although we are exercising the same sample app as in the Agg pipeline, here we would be building the sample app differently, and therefore we are also testing a different launch scenario. That might still be a legitimate scenario to exercise, if we clearly document this difference. BTW, are the SelfContainedDeployment-* sample apps affected by this issue?
- Regarding "Q2: Shall we bring Windows SDK 10.0.22000 back to build/scripts/windows-sdk.ps1 or modify the sample app's target build platform?", my vote is to modify the sample app's target build platform, because to my understanding is that everyone should shift away from "Windows SDK 10.0.22000" overtime. In the meantime, I presume "installing Windows SDK 22000 ourselves in sample app build jobs in the pipeline" can be used as a short term workaround.

Thank you very much for this great work!

build/AzurePipelinesTemplates/WindowsAppSDK-BuildAndTestSampleApps-Stage.yml
haonanttt: I missed it when cleaning up the code. Already modified

build/AzurePipelinesTemplates/WindowsAppSDK-BuildAndTestSampleApps-Stage.yml
haonanttt: I missed it when cleaning up the code. Already modified

build/AzurePipelinesTemplates/WindowsAppSDK-BuildAndTestSampleApps-Stage.yml
haonanttt: I missed it when cleaning up the code. Already modified

build/AzurePipelinesTemplates/WindowsAppSDK-BuildAndTestSampleApps-Stage.yml
haonanttt: I missed it when cleaning up the code. Already modified

build/WindowsAppSDK-Foundation-PR.yml
haonanttt: build/WindowsAppSDK-Foundation-PR.yml is the OneBranch PR pipeline: https://microsoft.visualstudio.com/ProjectReunion/_build?definitionId=99429 which almost not used, so this "true" will not actually affect the "\azp run" in github. However, to avoid confusing, I've modified this as well

build/AzurePipelinesTemplates/WindowsAppSDK-BuildSamplesCompat-Job.yml
haonanttt: The FallbackSource is still needed. Since the Foundation's dependency need to be get from it. I've tried to remove the FallbackSource and this task try to get Foundation's dependency from nuget.org, and encounter error for version not find. Link to the failed run: https://microsoft.visualstudio.com/ProjectReunion/_build/results?buildId=129567397

build/WindowsAppSDK-Foundation-Nightly.yml
haonanttt: Moved them to "variables" section

build/WindowsAppSDK-Foundation-Official.yml
haonanttt: Moved them to "variables" section

build/WindowsAppSDK-Foundation-PR.yml
haonanttt: Moved them to "variables" section

test/ModifySampleAppsReferences.ps1
haonanttt: Yes, hardcode the list here is not that future proof, but i didn't find a more proper way to handle the remove package list. The method I've tried contains:
1. Nuget install the latest Microsoft.WindowsAppSDK (similar to our nuget task to [install Foundation](https://github.com/microsoft/WindowsAppSDK/blob/35647d4d538387b9833341cfb5eac2112466e87b/build/AzurePipelinesTemplates/WindowsAppSDK-BuildSamplesCompat-Job.yml#L119C1-L127C143)) from **internal feed** --> Diff between subfolders of Microsoft.WindowsAppSDK and Microsoft.WindowsAppSDK.Foundation --> Get the removal package list.
    - Risk: If the latest version of Microsoft.WindowsAppSDK in our internal feed is broken and cannot be installed, it will fail the pipeline. [Example Run](https://microsoft.visualstudio.com/ProjectReunion/_build/results?buildId=129053525&view=results). The internal WASDK version (which is broken) at that time actually comes from a failed build
2. Nuget install the latest Microsoft.WindowsAppSDK from **public feed** --> same step as option 1
    - Risk: Our public version may have different dependency as our internal version. For example, currently we don't depend on ML package in public version. This will cause the ML package haven't been removed and bring error. [Example run](from **internal feed**)
3. Use script to download and unzip the .nupkg -->extract direct dependency from .nuspec --> for each direct dependency, recursively repeat the download, unzip, extract steps, until all dependencies are get. 
    - Risk: This method is a section of complex scripts and potentially have auth requirement when downloading the .nupkg.

Based on above results, and considering our change for the dependency is not that frequent. Perhaps hardcode might not be that unacceptable?

test/WindowsAppSDK.Test.SampleTests/WindowsAppSDKSampleAppTests.cs
haonanttt: Added this section of comment back

build/ProjectReunion-BuildFoundation.yml
haonanttt: Yes, I completed the PR in config repo and change this back to main branch

haonanttt:
> * Regarding "Q1: ... Will this target change affect the final test result?", my understanding is that, although we are exercising the same sample app as in the Agg pipeline, here we would be building the sample app differently, and therefore we are also testing a different launch scenario. That might still be a legitimate scenario to exercise, if we clearly document this difference. BTW, are the SelfContainedDeployment-* sample apps affected by this issue?

1. The "SelfContainedDeployment-* sample apps" are not affected by this issue. 2. I've found out the `WindowsAppSDKSelfContained` is set by our Base.targets. I'll reaching to Mano for seeking why we set WindowsAppSDKSelfContained to be true depending on MicrosoftWindowsAppSDKPackageDir and what is the usage for MicrosoftWindowsAppSDKPackageDir.

> * Regarding "Q2: Shall we bring Windows SDK 10.0.22000 back to build/scripts/windows-sdk.ps1 or modify the sample app's target build platform?", my vote is to modify the sample app's target build platform, because to my understanding is that everyone should shift away from "Windows SDK 10.0.22000" overtime. In the meantime, I presume "installing Windows SDK 22000 ourselves in sample app build jobs in the pipeline" can be used as a short term workaround.

This issue have been resolved by Mano's recent [Sample repo PR](https://github.com/microsoft/WindowsAppSDK-Samples/pull/508/commits)

Thank you for your insights Alex!

build/WindowsAppSDK-Foundation-PR.yml
haonanttt: Yes, if this parameter set to false, the pipeline still works well. Example run is: https://microsoft.visualstudio.com/ProjectReunion/_build/results?buildId=129701447&view=results

haonanttt:
> > * Regarding "Q1: ... Will this target change affect the final test result?", my understanding is that, although we are exercising the same sample app as in the Agg pipeline, here we would be building the sample app differently, and therefore we are also testing a different launch scenario. That might still be a legitimate scenario to exercise, if we clearly document this difference. BTW, are the SelfContainedDeployment-* sample apps affected by this issue?
> 
> 1. The "SelfContainedDeployment-* sample apps" are not affected by this issue. 2. I've found out the `WindowsAppSDKSelfContained` is set by our Base.targets. I'll reaching to Mano for seeking why we set WindowsAppSDKSelfContained to be true depending on MicrosoftWindowsAppSDKPackageDir and what is the usage for MicrosoftWindowsAppSDKPackageDir.

For this point, I've talked with Mano and realized that it is by design for us to only run in SelfContained mode if our application only use component package, as there is no framework package associated with the component package. So I added[ a comment to document the sample apps are run in SelfContained mode](https://github.com/microsoft/WindowsAppSDK/pull/5783/files#diff-14b2f2ed2897ebed232ef3e2fe2814bec72f663fa9f6e50dc4c8f89db01a1b6eR1:~:text=%23%20Note%3A%20In,generated%20until%20Aggregator.). Please help take a look whether this makes sense to you, thank you! @alexlamtest

test/ModifySampleAppsReferences.ps1
kythant: I think it's worthwhile to attempt 3. Future proofing is important. No one is going to remember to update every little thing in a random script especially this one. 

Also I am reading the rest of the script and I don't think we are actually removing the package in packagesToRemove but updating their content? Perhaps we should change the name of this variable to packagesToUpdate?

haonanttt: > I think it's worthwhile to attempt 3. Future proofing is important. No one is going to remember to update every little thing in a random script especially this one.

I've updated a script to recursively download and extract dependency package names from nuspec.

> Also I am reading the rest of the script and I don't think we are actually removing the package in packagesToRemove but updating their content? Perhaps we should change the name of this variable to packagesToUpdate?

For the packages listed in packagesToRemove, we actually remove them from the sample app's package reference. And for the nugetPackageToVersionTable, it is used to update the package's version. I've update the naming to make it more clear

haonanttt: Current example run is: https://microsoft.visualstudio.com/ProjectReunion/_build/results?buildId=129855897&view=results

build/WindowsAppSDK-Foundation-PR.yml
alexlamtest: In the Agg pipeline, setting this parameter to false will cause the sample app tests to happen right in the existing "pipeline test" stages, instead of a separate TestSampleApps Stage. In the pipeline run linked, I don't see sample app tests get run in the "pipeline test" stages. In fact, building of sample apps happened _after_ the "pipeline test" stages; therefore, the current pipeline architecture does not seem to support sample apps tests being run right in the "pipeline test" stages. I think we need to make a design decision: 1) Make the "TestSampleAppsAfterBuildStage=false" scenario work the same way as in the Agg pipeline, i.e., sample app tests get run right in the existing "pipeline test" stages, or 2) Define the "TestSampleAppsAfterBuildStage=false" scenario here as "skipping sample app tests altogether". I'm OK with both options.

build/AzurePipelinesTemplates/WindowsAppSDK-BuildAndTestSampleApps-Stage.yml
```diff
@@ -0,0 +1,145 @@
+# Note: In Foundation, the approach to test launching sample apps differs slightly from Aggregator. All sample apps in Foundation are tested in SelfContained mode.
+#       The reason for this difference is: sample apps only use component package in Foundation. There is no associated framework package, Runtime package is generated until Aggregator.
```
alexlamtest: "Runtime package is not genreated until Aggregator"?

haonanttt: Yes, the WindowsAppSDK.Runtime package is generated by our Aggregator pipeline, so there are 2 options for Sample App tests in Foundation: 1) do not use Runtime and run sample tests in SelfContain mode, 2) Include the latest Runtime package from internal feed and run sample tests still in framework dependent mode as Aggregator does. Here i choose option1

build/WindowsAppSDK-Foundation-PR.yml
haonanttt: I've read through Aggregator's code again and my understanding of the `TestSampleAppsAfterBuildStage = false` behavior in Aggregator is not like this. Since Aggregator is our internal code, I'll reach out to you offline for this.

alexlamtest: We've resolved this offline.

alexlamtest (approved):

test/GetNugetDependencies.ps1
```diff
@@ -0,0 +1,153 @@
...
+    )
+    
+    Write-Host "`nProcessing: $PackageName" -ForegroundColor Cyan
+    
+    # Build download arguments with unique output directory to avoid conflicts
+    $packageTempDir = Join-Path $TempFolder "pkg_$PackageName"
+    $downloadArgs = @("install", $PackageName, "-Source", $Source, "-OutputDirectory", $packageTempDir, "-ExcludeVersion", "-NonInteractive", "-DependencyVersion", "Ignore", "-PreRelease")
+
+    try {
+        $result = & nuget.exe @downloadArgs 2>&1
```
yeelam-gordon, line 38: Instead of just warning, we should fail here?
If not, the latest test run, I don't know what will be happening.

Please ask copilot to review any newly add powershell in this change, ensure we will fail early and reliablity on execution.

test/GetNugetDependencies.ps1
```diff
@@ -0,0 +1,153 @@
+param(
+    [Parameter(Mandatory = $true)]
+    [string]$PackageName,
+    
+    [string]$Source = "https://microsoft.pkgs.visualstudio.com/ProjectReunion/_packaging/Project.Reunion.nuget.internal/nuget/v3/index.json",
+    [string]$TempFolder = (Join-Path $env:TEMP "NuGetDeps"),
+    [string]$OutputVariableName = "dependencyList"
+)
+
```
yeelam-gordon, line 9: I think Qiutong or Boliang added
Set-StrictMode -Version 3.0
For all ps1 in the past, we should do it.

NOTE: it will break something, you need to try again and fix with copilot

test/GetNugetDependencies.ps1
```diff
@@ -0,0 +1,153 @@
...
+        $nupkgFile = Get-ChildItem -Path $packagePath -Filter "*.nupkg" | Select-Object -First 1
+        
+        if (-not $nupkgFile) {
+            Write-Warning "  No nupkg found for $PackageName"
+            return @()
+        }
+        
+        # Extract nupkg to find nuspec
+        $extractPath = Join-Path $TempFolder "extract_$PackageName"
+        if (Test-Path $extractPath) {
```
yeelam-gordon, line 56: Like this, it can be running multiple time.
Let's say the clean up failed, we should also add  -ErrorAction SilentlyContinue

test/GetNugetDependencies.ps1
yeelam-gordon: Copilot told me the following should be considered to improved:
image

test/WindowsAppSDK.Test.SampleTests/WindowsAppSDKSampleAppTests.cs
```diff
@@ -0,0 +1,1397 @@
+﻿// Copyright (c) Microsoft Corporation and Contributors.
+// Licensed under the MIT License.
+
+using System.Linq;
```
yeelam-gordon, line 4: Top Reliablity on causing rerun for the same test will be still failed. (From Copilot)
--------------------------------------------------------------------------

Highest-Priority Fixes (Do These First)
1. Guaranteed Package Cleanup (try/finally around Register → Launch → Remove)
Problem:

If a packaged app test throws after RegisterAppxPackage() but before RemoveAppxPackage(), the package remains registered. Re‑Run Risk:
Next run may see an already-registered package (different execution path, unexpected failure, or masked regression). Why First:
This is the most common “persistent state” leak that directly blocks or changes behavior on re-runs. Action:
Introduce a helper that always attempts RemoveAppxPackage() in a finally.
Optional pre-test cleanup: attempt removal first (ignore errors) to ensure deterministic starting state.
2. Reliable Process Cleanup (Fix ProcessComparer.GetHashCode + Central Final Sweep)
Problem:

ProcessComparer.GetHashCode uses reference hash, not PID; .Distinct() may misbehave.
Orphan processes may remain if detection misses them (warning path). Re‑Run Risk:
Leftover processes can:
Lock files
Consume ports
Cause single-instance sample apps to abort
Contaminate timing and environment Why Second:
Stray processes frequently cause subtle flakiness that’s hard to diagnose on pipelines. Action:
Fix GetHashCode to use Id.
Track every PID you launched (add to a ConcurrentDictionary<int, string>).
In TestCleanup and ClassCleanup, attempt to terminate any survivors.
Strengthen EnsureApplicationProcessHasExited to (a) poll, (b) always sweep by name at the end.
3. Replace Blind Sleeps With Polling for Process/Window
Problem:

Fixed Thread.Sleep(TimeSpan.FromSeconds(2)) assumes low system load.
On slower agents, detection fails → warning path → potential un-killed processes. Re‑Run Risk:
Under load or transient slowness, you accumulate false “process not found” cases that skip cleanup. Why Third:
Directly reduces probability of the first two issues even after you fix them. Action:
Introduce WaitForProcess(string name, TimeSpan timeout) and WaitForWindow(title, timeout) with 200–300 ms polling.
Use before deciding the process “doesn’t exist”.
4. Retry & Harden Package Deployment (Add-AppxPackage / Remove-AppxPackage)
Problem:

Transient deployment errors (sharing violations, 0x80073CF6, 0x80073D02) abort mid-sequence after registration or partial registration. Re‑Run Risk:
Partial/failed state blocks the next add/remove or changes exit codes. Why Fourth:
Less frequent than simple missing finally-blocks, but when it happens it is very disruptive. Action:
Wrap deployment calls in a retry helper with exponential backoff (e.g., 3 attempts: 0s, 2s, 5s).
Parse stderr for known transient codes to decide retry vs fail-fast.
5. Pre-Test (Per Method) Idempotent Sanity Sweep
Problem:

Prior aborted runs (CI cancellations, machine reuse) leave stale packages/processes. Re‑Run Risk:
Test appears to “self-heal” or fail unpredictably. Why Fifth:
This is a safety net; still useful even after 1–4 are fixed. Action:
At start of each packaged test: attempt a silent RemoveAppxPackage() (ignore failures).
Optionally: kill any lingering processes with known sample names (configurable).

yeelam-gordon: // --- Fix ProcessComparer hash code ---
private class ProcessComparer : IEqualityComparer<Process>
{
    public bool Equals(Process x, Process y)
        => (x == null && y == null) || (x != null && y != null && x.Id == y.Id);

    public int GetHashCode(Process obj)
        => obj?.Id.GetHashCode() ?? 0;
}

// --- Tracking launched processes ---
private static readonly object _procTrackLock = new object();
private static readonly HashSet<int> _launchedPids = new HashSet<int>();

private void TrackProcessIfAlive(int pid)
{
    if (pid <= 0) return;
    try
    {
        var p = Process.GetProcessById(pid);
        if (!p.HasExited)
        {
            lock (_procTrackLock) { _launchedPids.Add(pid); }
        }
    }
    catch { /* ignore */ }
}

private static void SweepTrackedProcesses()
{
    List<int> leftover;
    lock (_procTrackLock) leftover = _launchedPids.ToList();

    foreach (var pid in leftover)
    {
        try
        {
            var p = Process.GetProcessById(pid);
            if (!p.HasExited)
            {
                try
                {
                    p.Kill();
                    p.WaitForExit(3000);
                    Log.Comment($"[Cleanup] Killed leftover PID {pid}");
                }
                catch (Exception ex)
                {
                    Log.Warning($"[Cleanup] Failed to kill PID {pid}: {ex.Message}");
                }
            }
        }
        catch { /* already exited */ }
        finally
        {
            lock (_procTrackLock) { _launchedPids.Remove(pid); }
        }
    }
}

// --- Poll helpers ---
private bool WaitForProcessByName(string processName, TimeSpan timeout)
{
    var sw = Stopwatch.StartNew();
    var baseName = Path.GetFileNameWithoutExtension(processName);
    bool seen = false;
    while (sw.Elapsed < timeout)
    {
        var procs = Process.GetProcessesByName(baseName);
        if (procs.Length > 0)
        {
            foreach (var p in procs) TrackProcessIfAlive(p.Id);
            seen = true;
            // Give it a stabilizing grace period of one poll cycle
            Thread.Sleep(200);
            return true;
        }
        Thread.Sleep(200);
    }
    return seen;
}

private bool WaitForWindowTitle(string title, TimeSpan timeout, out int pid)
{
    var sw = Stopwatch.StartNew();
    pid = -1;
    while (sw.Elapsed < timeout)
    {
        var cond = UICondition.CreateFromName(title);
        if (UIObject.Root.Children.TryFind(cond, out UIObject window))
        {
            pid = window.ProcessId;
            TrackProcessIfAlive(pid);
            return true;
        }
        Thread.Sleep(200);
    }
    return false;
}

// --- Hardened EnsureApplicationProcessHasExited ---
private void EnsureApplicationProcessHasExited(int appWindowProcessId, string appProcessName)
{
    var processBaseName = Path.GetFileNameWithoutExtension(appProcessName);
    var attempts = 0;
    var max = 4;

    HashSet<int> candidatePids = new HashSet<int>();

    if (appWindowProcessId > 0) candidatePids.Add(appWindowProcessId);

    while (attempts++ < max)
    {
        // Gather processes by name
        foreach (var p in Process.GetProcessesByName(processBaseName))
        {
            candidatePids.Add(p.Id);
        }
        if (Process != null && !Process.HasExited)
        {
            candidatePids.Add(Process.Id);
        }

        if (candidatePids.Count == 0)
        {
            // Retry only if we never had a window pid and we are early
            if (appWindowProcessId == -1 && attempts < max)
            {
                Thread.Sleep(500);
                continue;
            }
            // If we expected it and never saw it, escalate (after top-level launch improvements this likely means real failure)
            if (appWindowProcessId != -1)
            {
                Log.Warning($"[Cleanup] Expected process '{appProcessName}' not found.");
            }
            return;
        }

        foreach (var pid in candidatePids.ToList())
        {
            try
            {
                var p = Process.GetProcessById(pid);
                if (!p.HasExited)
                {
                    Log.Comment($"Killing PID={pid} ({p.ProcessName})");
                    p.Kill();
                    p.WaitForExit(5000);
                }
            }
            catch (Exception ex)
            {
                Log.Comment($"Process {pid} already exited or cannot be killed: {ex.Message}");
            }
        }
        return;
    }
}

// --- Wrapper for packaged scenario with guaranteed cleanup ---
private void RunPackagedSample(
    string manifestRelativePath,
    string packageFullNameTemplate,
    string appId,
    string windowTitleOrNull,
    string processName)
{
    // Pre-emptive removal (ignore failures)
    TryRemovePackageQuiet(packageFullNameTemplate);

    bool registered = false;
    try
    {
        RegisterAppxPackage(manifestRelativePath);
        registered = true;

        LaunchAndClosePackagedApp(appId, windowTitleOrNull, processName);
    }
    finally
    {
        if (registered)
        {
            TryRemovePackageQuiet(packageFullNameTemplate);
        }
    }
}

private void TryRemovePackageQuiet(string packageFullNameTemplate)
{
    try { RemoveAppxPackage(packageFullNameTemplate); }
    catch (Exception ex) { Log.Comment($"[Pre/Post Cleanup] Ignore removal error: {ex.Message}"); }
}

// --- Example updated test using the wrapper ---
[TestMethod]
public void AppLifecycleStateNotificationsCsWpfPackaged()
{
    if (IsArchX86() || IsBuildConfigDebug())
    {
        Log.Result(WEX.Logging.Interop.TestResult.Skipped, "Not tested for x86 or Debug.");
        return;
    }
    Log.Comment($"OSVersion {Environment.OSVersion.Version}");

    if ((Environment.OSVersion.Version.Major < 10) ||
        (Environment.OSVersion.Version.Major == 10 && Environment.OSVersion.Version.Build < 19041))
    {
        Log.Result(WEX.Logging.Interop.TestResult.Skipped, "Requires Win 10 Build 19041+.");
        return;
    }
    if ((Environment.OSVersion.Version.Major == 10) &&
        (Environment.OSVersion.Version.Build == 19044 || Environment.OSVersion.Version.Build == 19045))
    {
        Log.Result(WEX.Logging.Interop.TestResult.Skipped, "Skipped on 19044/19045 until issue fixed.");
        return;
    }

    RunPackagedSample(
        manifestRelativePath: "AppLifecycle\\StateNotifications\\cs1\\cs-wpf-packaged\\CsWpfStatePkg\\bin\\[BuildArch]\\[BuildConfig]\\AppxManifest.xml",
        packageFullNameTemplate: "7718eec3-5345-4d4e-815f-97df705dc89c_1.0.0.0_[BuildArch]__s9y1p3hwd5qda",
        appId: "7718eec3-5345-4d4e-815f-97df705dc89c_s9y1p3hwd5qda!App",
        windowTitleOrNull: "C# WPF Activation",
        processName: "CsWpfState.exe");
}

// --- Central cleanup hooks ---
[TestCleanup]
public void TestCleanup()
{
    SweepTrackedProcesses();
}

[ClassCleanup]
public static void ClassCleanup()
{
    SweepTrackedProcesses();
}

yeelam-gordon: BTW, we should not have packaged app generated for our project?, probably some of above improvement are only applicable to Aggregator.

build/AzurePipelinesTemplates/WindowsAppSDK-RunTests-Steps.yml
```diff
@@ -91,6 +94,67 @@ steps:
       arguments: -NoInteractive -Offline -Verbose -CheckTAEFService -ShowSystemInfo
       workingDirectory: '$(Build.SourcesDirectory)'
 
+  - task: PowerShell@2
+    displayName: Display storage info and init variables
+    inputs:
+      targetType: 'inline'
+      script: |
+        Get-WmiObject win32_logicaldisk | Format-Table DeviceId, MediaType, @{n="Size";e={[math]::Ro53und($_.Size/1GB,2)}},@{n="FreeSpace";e={[math]::Round($_.FreeSpace/1GB,2)}}, @{n="UsedSpace";e={[math]::Round((($_.Size-$_.FreeSpace)/1GB),2)}}
```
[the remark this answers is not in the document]
haonanttt: fiexed

test/GetNugetDependencies.ps1
haonanttt: changed the catch section from showing warning to throw error

test/GetNugetDependencies.ps1
haonanttt: `Set-StrictMode -Version 3.0` added

test/GetNugetDependencies.ps1
haonanttt: `-ErrorAction SilentlyContinue` added

test/WindowsAppSDK.Test.SampleTests/WindowsAppSDKSampleAppTests.cs
haonanttt: For this issue, it not only affect Foundation but also related to Aggregator. And considering the issue is more related to packaged app, it only have minor affect to Foundation. So we will not handle this in this PR. Instead, we will use another PR for this purpose

test/WindowsAppSDK.Test.SampleTests/WindowsAppSDKSampleAppTests.cs
```diff
@@ -0,0 +1,1397 @@
...
+
+            if (Process != null)
+            {
+                Process.Dispose();
+                Process = null;
+            }
+            return;
+        }
+
+        // Invoke CreateProcess to run exeName with arguments. Wiat for completion.
```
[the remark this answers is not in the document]
haonanttt: Will not change the cs code in this PR. Will leave these points to the further PR

test/WindowsAppSDK.Test.SampleTests/WindowsAppSDKSampleAppTests.cs
```diff
@@ -0,0 +1,1397 @@
...
+        }
+
+        private static void RunPowerShellCommand(string arguments)
+        {
+            RunCommand("powershell.exe ", arguments);
+        }
+
+        private static void RemoveAppxPackage(string packageFullNameTempalte)
+        {
+            var packageFullName = packageFullNameTempalte.Replace("[BuildArch]", BuildArch);
```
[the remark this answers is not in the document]
haonanttt: Will not change the cs code in this PR. Will leave these points to the further PR

test/WindowsAppSDK.Test.SampleTests/WindowsAppSDKSampleAppTests.cs
```diff
@@ -0,0 +1,1397 @@
...
+using Microsoft.Windows.Apps.Test.Foundation;
+using Microsoft.Windows.Apps.Test.Foundation.Waiters;
+
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Runtime.InteropServices;
+using System.Threading;
+using System.Collections.Generic;
+
```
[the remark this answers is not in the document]
haonanttt: Will not change the cs code in this PR. Will leave these points to the further PR

test/WindowsAppSDK.Test.SampleTests/WindowsAppSDKSampleAppTests.cs
```diff
@@ -0,0 +1,1397 @@
...
+                return;
+            }
+
+            var exePath = GetFullFilePathFromRelativePath("AppLifecycle\\EnvironmentVariables\\cpp-win32-unpackaged\\[BuildArch]\\[BuildConfig]\\CppWinMainEnv.exe");
+            LaunchAndCloseUnpackagedApp(exePath, null, "CppWinMainEnv.exe");
+            return;
+        }
+
+        [TestMethod]
+        public void AppLifecycleEnvironmentVariablesCsWinfromsUnpackaged()
```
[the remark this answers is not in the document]
haonanttt: Will not change the cs code in this PR. Will leave these points to the further PR

yeelam-gordon (approved):

PR #5821, opened 2025-09-11
DevCheck: Add -CheckVCLibs and -InstallVCLibs options

DrusTheAxe:
New options:
* `-CheckVCLibs` - check if the `Microsoft.VCLibs*14.00*.msix` packages need to be installed
* `-InstallVCLibs` - install the `Microsoft.VCLibs*14.00*.msix` packages (if necessary)

`-CheckVCLibs` occurs by default (like other `-Check...` options)

`-FixAll` includes `InstallVCLibs`

### DIFF
--- a/tools/DevCheck/DevCheck.ps1
+++ b/tools/DevCheck/DevCheck.ps1
@@ -55,6 +55,9 @@
 .PARAMETER FixLongPath
     Enable LongPath support if necessary.
 
+.PARAMETER InstallVCLibs
+    Install VCLibs MSIX packages.
+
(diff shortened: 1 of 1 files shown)


DrusTheAxe:
> Nice!

Next up is `-InstallMSIX <Filespec>` so you can `DevCheck ... -InstallMSIX c:\foo\test\blah\*.msix` and it'll `Add-AppxPackage c:\foo\test\blah\*.msix`/ A bit more complicated than it sounds so it'll await vNext PR :-)

PR #5828, opened 2025-09-12
Add fallback for MrmGetFilePathFromName

### LINKED ISSUE #5814: WinAppSDK 1.8: MRM.cpp's MrmGetFilePathFromName now reports error when the resources.pri file doesn't exist
### Describe the bug

Depending on calling context, with WinAppSDK version 1.7, this function used to return "resources.pri" from the current exe path even if that file doesn't exist. Now the function checks for existence and returns ERROR_FILE_NOT_FOUND (https://github.com/microsoft/WindowsAppSDK/blob/main/dev/MRTCore/mrt/Core/src/MRM.cpp#L1085) (which seems to contradict the doc BTW).

This breaks Microsoft.Windows.ApplicationModel.Resources's ResourceManager https://github.com/microsoft/WindowsAppSDK/blob/main/dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/src/ResourceManager.cpp#L18

It's in a XAML island (Win32 desktop app hosts .NET + WinUI3 class library) context. There's a full repro project here: https://github.com/smourier/Win32ShowsWinUI3

Uncomment this line to reproduce: https://github.com/smourier/Win32ShowsWinUI3/blob/main/WinUI3ClassLibrary/SampleWindow.xaml.cs#L43

### Expected behavior

This should work as before
Maybe there's a workaround

### Screenshots

_No response_

### NuGet package version

Windows App SDK 1.8.0: 1.8.250907003

### Packaging type

Unpackaged

### Windows version

Windows 11 version 24H2 (26100, June 2025 Update)

### IDE

Visual Studio 2022

### Additional context

_No response_

### LINKED ISSUE #5940: MrmGetFilePathFromName violates its document
### Describe the bug

The document and comments of this method indicates that it will always succeed even if the app doesn't have a PRI file.

https://github.com/microsoft/WindowsAppSDK/blob/53d32b0589c5cbbc4ecf098344d9b27993742a22/dev/MRTCore/mrt/Core/src/MRM.cpp#L964-L967

However, it returns ERROR_FILE_NOT_FOUND:

https://github.com/microsoft/WindowsAppSDK/blob/53d32b0589c5cbbc4ecf098344d9b27993742a22/dev/MRTCore/mrt/Core/src/MRM.cpp#L1085

I have created #5828 to fix it, but no one takes care of it.

Create a WinUI3 program without PRI files. I know it's a little hard, but see:

https://github.com/microsoft/WindowsAppSDK/issues/5814

https://github.com/Alovchin91/winui3-rs/issues/5

### Expected behavior

The behavior of the method should match its document.

### Screenshots

_No response_

### NuGet package version

Windows App SDK 1.8.2: 1.8.251003001

### Packaging type

Unpackaged

### Windows version

Windows 11 version 24H2 (26100, June 2025 Update)

### IDE

Visual Studio 2022

### Additional context

_No response_

Berrysoft:
Closes #5814

Closes #5940 

According to the docs, this method should success even if no resource file is found.


dev/MRTCore/mrt/Core/src/MRM.cpp
```diff
@@ -1074,13 +1087,14 @@ STDAPI MrmGetFilePathFromName(_In_opt_ PCWSTR filename, _Outptr_ PWSTR* filePath
             PATHCCH_ALLOW_LONG_PATHS));
 
         DWORD attributes = GetFileAttributes(outputPath.get());
-        if ((attributes != INVALID_FILE_ATTRIBUTES) && !(attributes & FILE_ATTRIBUTE_DIRECTORY))
+        if ((pass == SearchPass::DefaultFallback) ||
+            ((attributes != INVALID_FILE_ATTRIBUTES) && !(attributes & FILE_ATTRIBUTE_DIRECTORY)))
         {
             // The file exists. Done.
```
riverar:
```suggestion
```

dev/MRTCore/mrt/Core/src/MRM.cpp
```diff
@@ -1074,13 +1087,14 @@ STDAPI MrmGetFilePathFromName(_In_opt_ PCWSTR filename, _Outptr_ PWSTR* filePath
             PATHCCH_ALLOW_LONG_PATHS));
 
         DWORD attributes = GetFileAttributes(outputPath.get());
-        if ((attributes != INVALID_FILE_ATTRIBUTES) && !(attributes & FILE_ATTRIBUTE_DIRECTORY))
+        if ((pass == SearchPass::DefaultFallback) ||
+            ((attributes != INVALID_FILE_ATTRIBUTES) && !(attributes & FILE_ATTRIBUTE_DIRECTORY)))
```
riverar, line 1091: I think this is going to need an audit of all the code in MUX and WAS where MRM was expected to fail right? (e.g., https://github.com/microsoft/WindowsAppSDK/blob/main/dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/src/Helper.cpp#L77-L103)

Berrysoft: There are two paths calling this method. One is the ResourceManager:

https://github.com/microsoft/WindowsAppSDK/blob/9bde0388e0177b11a012d8b475eb1edeabebcccc/dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/src/ResourceManager.cpp#L22

which is OK if the file not found, and is expected (https://github.com/microsoft/WindowsAppSDK/blob/9bde0388e0177b11a012d8b475eb1edeabebcccc/dev/MRTCore/mrt/Core/src/MRM.cpp#L966-L967). The other one is the ResourceLoader:

https://github.com/microsoft/WindowsAppSDK/blob/9bde0388e0177b11a012d8b475eb1edeabebcccc/dev/MRTCore/mrt/Microsoft.Windows.ApplicationModel.Resources/src/ResourceLoader.cpp#L14

which will fail if it cannot find the pri file.

Scottj1s: Indeed - thanks for the heads up.  Code like that will now need to test for nonexistent path as well (not just S_OK).

Berrysoft: Do I need to add tests in this PR?

Scottj1s: that would be great, @Berrysoft - thanks

Berrysoft: Well. I have to admit that the PR was created more than half a year before. I need some time to investigate how to add the tests. If it is a must, I'll do it later. But if it is not, I think I'll keep the PR as-is.

Before 1.8, all code works well with the nonexistent path. Furthermore, I have been [live-patching](https://github.com/compio-rs/winio/blob/e76fa89c513bc3c867887dbe23d208c275e72cde/winio-ui-winui/src/hook/mrm.rs#L60-L65) the MRM.dll since 1.8 to the latest 2.2. The patch always works well with all versions' WinUI. So I think this PR doesn't break the code.

riverar:
Can we slot this in for the next preview build @Scottj1s @RDMacLachlan? This is impacting reactor projects downstream.

dev/MRTCore/mrt/Core/src/MRM.cpp
Berrysoft: Hello, @Scottj1s @riverar . I read the test code in this repo and found that it already contains tests for this case:

https://github.com/microsoft/WindowsAppSDK/blob/e858676700689726e1aad07840ca6961d8489104/dev/MRTCore/mrt/Core/unittests/MrmTests.cpp#L687-L690

But I don't know why it didn't fail before this PR. I think I don't need to add other tests. They already fill the need.

Berrysoft:
Hello @Scottj1s , is it still possible to merge this PR?

dev/MRTCore/mrt/Core/src/MRM.cpp
Scottj1s: Did the audit. Both callers of `MrmGetFilePathFromName` are fine: `ResourceManager` already tolerates a missing PRI (only its `check_hresult(GetDefaultPriFile)` gate was breaking on the 1.8 `ERROR_FILE_NOT_FOUND` regression), and `ResourceLoader` throws with or without the change by design. The one internal path that keyed off the failure was `GetDefaultPriFile`'s sparse-packaged `[modulename].pri` fallback (added in #6376), which gated on `IsResourceNotFound(hr)` — since the fix makes `MrmGetFilePathFromName` return `S_OK`, that now gates on actual file existence (`CheckFile`) instead. Handled in #6603.

Scottj1s: On tests: `MrmTests::GetFilePath` and `MrtCoreUnpackagedTests.TestClass.DefaultResourceManager` already cover this — they just didn't catch the original regression because they were on the `BypassTests.json` allowlist (added in #5498, extended in #5919, after the regression). #6603 un-bypasses both and strengthens `GetFilePath`. Verified locally: MrmUnitTest 21/21, MrtCoreUnpackagedTests 27/27.

Scottj1s:
Thanks for this, @Berrysoft — I've opened #6603 to wrap it up on top of your change. The `DefaultFallback` pass added here isn't actually reachable (both terminal passes jump to `Final`), so the function returns `S_OK` with a null `*filePath`. #6603 makes `DefaultFallback` the reachable final pass returning a non-null best-effort path (legacy parity), fixes the dependent `GetDefaultPriFile` sparse-packaged fallback, and re-enables the two tests that were bypassed. Your commits are included there, so this supersedes the PR.

Berrysoft:
Thanks!

Scottj1s:
> Thanks!

Thanks for opening the issue, suggesting a fix, and waiting patiently - much appreciated.

dev/MRTCore/mrt/Core/src/MRM.cpp
riverar: Awesome work @Scottj1s, thank you. Squeaking a little more to get this landed too.

Berrysoft:
This PR is not needed now, I think? Regarding that #6603 has been merged.

riverar:
Did that fix ship in a release we can test with?

riverar:
Any updates on this? cc: @Scottj1s

Berrysoft:
I don't know if there's a release, but it seems that this PR is not needed.

riverar:
I'm confused because it seems like @Scottj1s was building on top of this PR, not replacing it.

riverar:
I see now, you're right. This PR was superseded and never cleaned up. The lack of _normal_ GitHub usage patterns by this team continues to frustrate me.

Scottj1s:
@Berrysoft , @riverar - fix has been cherry picked for August's 1.8 and 2.0 servicing releases (note: usual caveats, schedules always subject to change)

Scottj1s:
> I see now, you're right. This PR was superseded and never cleaned up. The lack of _normal_ GitHub usage patterns by this team continues to frustrate me.

fair - I considered building on top of this PR, and do appreciate your contribution, @Berrysoft.  there was enough divergence to warrant a new PR.

Berrysoft:
> fix has been cherry picked for August's 1.8 and 2.0 servicing releases

Thanks, @Scottj1s ! However, when I'm using the latest WinUI3 (2.3.1.0), `MrmGetFilePathFromName` still returns `ERROR_FILE_NOT_FOUND` (0x80070002). I detoured this method and get this return code. I watch my process through Process Explorer, and confirmed that the process is really loading the `MRM.dll` from `C:\Program Files\WindowsApps\Microsoft.WindowsAppRuntime.2_2.3.1.0_x64__8wekyb3d8bbwe\MRM.dll`. So, does winappsdk 2.3.1 really contain this fix?

riverar:
I was confused too because at one point the PR or commits were tagged 2.3.1.0. ~~But seems that's gone now, I can't find it.~~

Update: This commit is tagged 2.3.1.0. https://github.com/microsoft/WindowsAppSDK/commit/91160b078668051e58ba1e19eb699f4efc8c5b20

So now we wait for August release!

Issue #875, opened 2021-05-27
force using longPathAware compulsory for all reunion apps


ecovio1:
## Summary

# Make Developing and Developers life less painful on win32 subsystem.

reunion 1.0 is yet to hit us. force using Long path  awareness for all apps that use reunion components.
can't stuck in 1995 anymore.  though maintaining backward compatibilty (is app 32 bit or targeting windows version X ? then stuck in 255 else underlying filesystem's limit. )
win32 subsystem also has to be forward compatible for future relevancy.  it's high time the subsystem made full use of NT.

andrewleader:
Hi ecovio1, thanks for the feedback! @ptorr-msft do you have any thoughts on long path awareness?

ptorr-msft:
Thanks for the suggestion (and your interest in long paths), but as-written this is not aligned with the broad goals of Reunion because we want to allow incremental adoption of Project Reunion features in existing apps. If an existing app just wants to (e.g.) use a new WinUI control, they shouldn't have to completely rewrite their file handling code (and / or wait until the libraries they're using get updated) to deal with 32k paths. 

If you have specific suggestions around how to make it easier to embrace long paths, they would be good proposals.

(I'll also note for completeness that the entire long-path feature is off by default so there is no immediate benefit for users).

ecovio1:
> if an existing app just wants to (e.g.) use a new WinUI control, they shouldn't have to completely rewrite their file handling code (and / or wait until the libraries they're using get updated)

for that comparatively smaller change, forcing to use anything other than that will be headache, yes sure. I understand  what you are saying.  that's fine if  all apps not require it.

but but , judging on the past changes of most win32 apps since they were created, I suspect if that will be the case here. 

> If you have specific suggestions around how to make it easier to embrace long paths

should've been "for Existing apps and libraries   without buffer overrun" .

 accessing long paths using existing libraries will require abstraction and innovation.
good proposals will be made here once I find ways(if any) of accomplishing this. 

> the entire long-path feature is off by default so there is no immediate benefit for users 

a cumulative update can fix/overwrite that going downlevel to 1809 .

#
the rule of nature is :  
things to stay relevant from time to time,  things must innovate from time to time.  

if that innovation requires a change in the core, changes/innovations must must start at some point.
 it's inevitable.    
reunion was just that fresh water drop in  the win32 desert   land.   

this was the point of the proposal.

ptorr-msft:
[...]

PR #5830, opened 2025-09-12
Add longPathAware declaration to app.manifest for C# project templates

Marv51:
Since Windows 10 1607, developers can declare a flag in their app.manifest to remove the MAX_PATH limitation. This enables support for file paths longer than 260 characters. 

Modern versions of .NET support this well and most .NET libraries also support this as well due to their cross-platform design. Other platforms do not impose the same limitation.

For new C# applications in 2025 and beyond, the recommended default value for longPathAware should be true. This PR enables that setting in the C# project templates.

This change does not force WinAppSDK apps to be longPathAware. If incompatibilities arise, developers can remove the line, or set the value to false explicitly.

I have not updated the C++ templates in this PR.

----

WARNING:

### DIFF
--- a/dev/VSIX/ProjectTemplates/Desktop/CSharp/PackagedApp/BlankApp/app.manifest
+++ b/dev/VSIX/ProjectTemplates/Desktop/CSharp/PackagedApp/BlankApp/app.manifest
@@ -14,6 +14,7 @@
   <application xmlns="urn:schemas-microsoft-com:asm.v3">
     <windowsSettings>
       <dpiAwareness xmlns="http://schemas.microsoft.com/SMI/2016/WindowsSettings">PerMonitorV2</dpiAwareness>
+      <longPathAware xmlns="http://schemas.microsoft.com/SMI/2016/WindowsSettings">true</longPathAware>
     </windowsSettings>
   </application>
 </assembly>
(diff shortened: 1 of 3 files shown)


mominshaikhdevs:
* https://github.com/microsoft/WindowsAppSDK/issues/875.

* https://github.com/microsoft/winget-cli/issues/5466.

* https://github.com/rust-lang/cargo/issues/9770.

sylveon:
This is a dangerous default

Developers won't specifically be testing their apps with long paths if it's the default, so they'll think everything is fine, and then one native library they may be indirectly using does `char file[MAX_PATH]` and it blows up on either a customer's computer or as a potential security vulnerability (stack overflows are ripe for abuse).

Long paths are something that should only be enabled after full validation that your app works as expected with them.

Marv51:
Maybe a milder version of this PR would be to set this explicitly to false with a comment? What are the opinions on that?

```xml
<!-- Set to true to allow paths longer than 260 characters. May cause issues with some third-party components. -->
<longPathAware xmlns="http://schemas.microsoft.com/SMI/2016/WindowsSettings">false</longPathAware>
```

sylveon:
I would be fine with that, but just comment it out instead of an explicit false.
E.g.
```xml
<!-- Uncomment to allow paths longer than 260 characters. May cause issues with some third-party components. -->
<!-- <longPathAware xmlns="http://schemas.microsoft.com/SMI/2016/WindowsSettings">true</longPathAware> -->
```

riverar:
Note: This change does nothing until this feature is turned on machine-wide. This is off by default. (`HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\FileSystem!LongPathsEnabled=0x1`)

I agree with others here, however. This change may confuse newcomers with crashes/undefined behavior after bringing in libraries that don't properly support long paths.

mominshaikhdevs:
It has become a chicken and egg problem at this point.

> I would be fine with that, but just comment it out instead of an explicit false. E.g.
> 
> ```
> 
> 
> ```

This is the way to go forward.

riverar:
Including this feature upfront could create lots of confusion for developers who might expect it to work out of the box. (It won't work out of the box.) Developers with specialized use cases who genuinely need this functionality can figure this out through documentation or by consulting with subject matter experts.

PR #1790, opened 2021-11-13
Add license support to Deployment API and a new InstallLicenses API

DrusTheAxe:
The Deployment API `Initialize()` now installs license files embedded in the framework package along with embedded MSIX packages i.e.

```
FOREACH file IN fwkpkgdir\MSIX\*-license.xml
    InstallLicenseFile(file)
FOREACH file IN fwkpkgdir\MSIX\*.msix
    AddPackage(file)
```

We also add a new `WindowsAppRuntime_InstallLicenses()` API for use by anyone who directly installs the *.msix packages (i.e. those who don't distribute via the Store, don't use WindowsAppRuntimeInstaller.exe and don't use the Deployment API).

NOTE: The framing's in place but the implementation is stubbed. I know how but it's late so best to sleep on it before I start plowing in. What's present exposes a new entrypoint not-yet-used export with new codepath so no harm having it exist but not yet functional. Watching this space...

Updated the build pipeline to copy the implementation header into the build before compiling during build pipelines (yet w/o disrupting dev inner-loop). Needs to _get_ the Transport package containing that. @BenJKuhn I'm a little fuzzy on the details there. Feel free to update the branch.

Also changed the installer's ordering to now install licenses BEFORE packages. In worst case scenarios better to have incomplete packages than incomplete licenses as the latter is readily identified and understood as a problem, whereas tooling, understanding and even awareness of licenses' existence is a rarer thing. This change updates the installer for improved reliability and troubleshooting if something does go wrong, and aligns with the ordering in the Deployment API (for same reasons).

PR #5853, opened 2025-09-22
Refactoring and adding unit tests to deployment

guimafelipe:
This Pull Request introduces a refactoring on the `DeploymentManager.cpp` class implementation. The goal is to facilitate bug tracking and the implementation of more localized tests, making it easier to implement new changes, tracking possible regressions in inner behaviors and also on isolating the causes of existing failures.

# New modules

For this change if focused on the `Initialize` method. More specifically, on the internal `_Initialize` method.  `_Initialize` calls a internal method `Deploy(..)`, which is a 3 lines method that calls `InstallLicenses(...)` and `DeployPackages(...)`.

These two methods were the main target of the refactoring. Let's go into each of them in details next.

## Install licenses

This method is responsible for installing the licenses for the packages (Main and Singleton (and DDLM)) before installing them.
I separated it and extracted to a new namespace as the functions `WindowsAppRuntime::Deployment::Deployer::GetLicenseFiles` and `WindowsAppRuntime::Deployment::Deployer::InstallLicenses`.

### `GetLicenseFiles`

This function's responsibility is to just iterate in the filesystem and return an array with the license files to install. Can be independently tested from the actual installation of the licenses.

### `InstallLicences`
This function receives the list of licenses to install and uses the `LicenseInstaller` to execute the installation. The license installer is a stubbed file in the repository and we don't have access to its implementation. For this, I added an interface `ILicenseInstaller` that this function accepts as parameter. The `DeployementManager::InstallLicenses` implements a simple proxy/wrapper class around the real object and passes to the function. This gives us full control for mocking the installer on our tests and also a layer of separation for this namespace to not have strong dependency on the installer.

## Deploy Packages

This method is responsible for doing the actual deployment after the licenses were installed. I also divided this method in 2 functions with separated responsibilities: `WindowsAppRuntime::Deployment::Deployer::GetDeploymentPackageArguments` and `WindowsAppRuntime::Deployment::Deployer::DeployPackages`.

### `GetDeploymentPackageArguments`

This function is responsible for building the vector with the package arguments that we will use later to execute the deployments. With this refactor, this becomes a pure function with very predictable behavior and no side effects.

### ~~`DeployPackages`~~

~~This is the actual deployment. This function calls `WindowsAppRuntime::Deployment::PackageRegistrar::AddOrRegisterPackageInBreakAwayProcess` or `WindowsAppRuntime::Deployment::PackageRegistrar::AddOrRegisterPackage` depending on the arguments (in this case, if it is full trust or not). These two functions were moved to the `PackageRegistrar` namespace as they are similar and very specific on what they intent to do.~~

## Package utilities

The deployment manager's methods related to package querying were moved to a new namespace `WindowsAppRuntime::Deployment::Package`, on the files `PackageUtilities.h|cpp`. This was done in case we want to mock this behavior in tests and also for organization purposes.

# Unit tests

With this refactoring, I created the new test project `DeploymentUnitTests`. I wrote some initial tests for both deployer and package registrar namespaces. With more test scenarios also being easily implemented if needed as now the modules are more independent and the major dependencies are broken.

WARNING:


dev/Deployment/Deployer.h
```diff
@@ -0,0 +1,50 @@
...
+// Licensed under the MIT License.
+#pragma once
+
+#include <string>
+#include <filesystem>
+#include <vector>
+#include <functional>
+#include <DeploymentActivityContext.h>
+
+namespace WindowsAppRuntime::Deployment::Deployer
```
ssparach: Recommend renaming to clearer names: DeploymentHelper for the class and PackageMetaData (or PackageInformation) for the struct to better reflect their roles.

dev/Deployment/DeploymentManager.cpp
```diff
@@ -327,308 +330,67 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
...
-
-    HRESULT DeploymentManager::InstallLicenses(const std::wstring& frameworkPackageFullName)
-    {
-        ::WindowsAppRuntime::Deployment::Activity::Context::Get().Reset();
-        ::WindowsAppRuntime::Deployment::Activity::Context::Get().SetInstallStage(::WindowsAppRuntime::Deployment::Activity::DeploymentStage::GetLicensePath);
+        auto packagePath = ::WindowsAppRuntime::Deployment::Package::GetPackagePath(frameworkPackageFullName);
 
         // Build path for licenses
-        auto licensePath{ std::filesystem::path(GetPackagePath(frameworkPackageFullName)) };
+        auto licensePath{ std::filesystem::path(packagePath) };
```
ssparach, line 361: With the refactor of licensing-related logic (mostly in Deployer), why is the license path constructed here? Since the code is being refactored, it might be worth separating concerns by introducing distinct classes. For example, consider creating a dedicated LicenseManager or similar to encapsulate licensing responsibilities—to align with single-responsibility principles.

dev/Deployment/Deployer.h
```diff
@@ -0,0 +1,50 @@
...
+{
+    // Structure to hold deployment package arguments
+    struct DeploymentPackageArguments
+    {
+        const std::wstring packageIdentifier;
```
ssparach: Since PackageIdentifier is already the name of an existing struct, it might be worth considering a different name for clarity. Aligning the new name with the identifier member of the PackageIdentifier struct could help maintain consistency and avoid confusion.

dev/Deployment/PackageDeployment.h
```diff
@@ -0,0 +1,50 @@
...
+        const std::wstring& frameworkPackageFullName,
+        ::WindowsAppRuntime::Deployment::Activity::Context& initializeActivityContext,
+        const std::function<std::wstring(const std::wstring&)>& getPackagePathFunc);
+
+    // Package deployment
+    HRESULT DeployPackages(
+        const std::vector<DeploymentPackageArguments>& deploymentPackageArguments,
+        const bool forceDeployment,
+        ::WindowsAppRuntime::Deployment::Activity::Context& initializeActivity,
+        const std::function<HRESULT()>& startupNotificationsLongRunningPlatformFunc);
```
ssparach: Passing startupNotificationsLongRunningPlatformFunc as a function argument can make the interface harder to maintain. Could we instead include the header that defines this functionality and call it directly where needed? This would reduce indirection and simplify the design.

dev/Deployment/Deployer.cpp
```diff
@@ -0,0 +1,164 @@
...
+        initializeActivityContext.SetInstallStage(::WindowsAppRuntime::Deployment::Activity::DeploymentStage::InstallLicense);
+
+        // Deploy the licenses (if any)
+        for (const auto& licenseFileName : licenseFiles)
+        {
+            // Install the license file
+            auto licenseFileFullName{ licensePath };
+            licenseFileFullName /= licenseFileName;
+
+            // initializeActivityContext.Reset(); --> Why are we reseting here? It clears out all the info.
```
ssparach: Just checking—does this comment still apply, or should we clean it up during the refactor?

dev/Deployment/PackageRegistrar.cpp
```diff
@@ -0,0 +1,145 @@
...
+    // If useExistingPackageIfHigherVersion == false, Adds the current version package at the passed in path using PackageManager.
+    // If useExistingPackageIfHigherVersion == true, Registers the higher version package using the passed in path as manifest path and PackageManager.
+    // This requires the 'packageManagement' or 'runFullTrust' capabilities.
+    HRESULT AddOrRegisterPackage(
+        const std::filesystem::path& path,
+        const bool useExistingPackageIfHigherVersion,
+        const bool forceDeployment,
+        winrt::Windows::Management::Deployment::IPackageManager& packageManager,
+        ::WindowsAppRuntime::Deployment::Activity::Context& activityContext
+    )
```
ssparach: Nit: For consistency, the brace doesn’t need to be on a new line here—apply this to other areas as well.

dev/Deployment/PackageRegistrar.cpp
```diff
@@ -0,0 +1,145 @@
...
+#include <winrt/Windows.Foundation.h>
+#include <winrt/Windows.Management.Deployment.h>
+#include <filesystem>
+
+using namespace winrt;
+
+namespace WindowsAppRuntime::Deployment::PackageRegistrar
+{
+    // Helper method to process deployment operation results and extract error information
+    inline HRESULT ProcessDeploymentOperationResult(
```
ssparach: Since the logic pertains to DeploymentResult, would it make sense to move this into DeploymentResult.cpp for better alignment and cohesion? This could help keep related functionality grouped together and improve maintainability.

dev/Deployment/PackageDeployment.cpp
```diff
@@ -0,0 +1,164 @@
...
+            // initializeActivityContext.Reset(); --> Why are we reseting here? It clears out all the info.
+            initializeActivityContext.SetCurrentResourceId(licenseFileFullName);
+
+            RETURN_IF_FAILED_MSG(licenseInstaller.InstallLicenseFile(licenseFileFullName.c_str()),
+                                 "LicenseFile:%ls", licenseFileFullName.c_str());
+        }
+        return S_OK;
+    }
+
+    std::vector<DeploymentPackageArguments> GetDeploymentPackageArguments(
```
ssparach, line 11: Given that GetDeploymentPackageArgs is only invoked once and appears tightly coupled to its caller, does it need to be separated into its own method? Introducing this extra layer—along with multiple parameters like activityContext etc.—makes the flow less intuitive (especially for debugging)

ssparach (changes requested):

ssparach:
Given the scope of this refactor, I’d strongly recommend exercising the functionality manually end to end across all three scenarios as well—packaged, unpackaged, and AppContainer. It would also be valuable to add coverage for AppContainer scenarios and/or sample launch tests, even if that happens in a future PR.

DrusTheAxe:
In the description "...depending on the arguments (in this case, if it is full trust or not)."

Isn't that based on if it's **Centennial** or not?

Package management had a restriction where the caller needed to be...
1. Packaged process + AppContainer + packageManagement capability
2. Unpackaged process + IL>=Medium
3. Packaged process + IL>=Medium + NOT Centennial
4. Packaged process + IL>=Medium + Centennial + caller's Publisher == target's Publisher

Thus only packages with Publisher=Microsoft Corporation could do packagement management operations on WinAppSDK packages :-( and thus, Breakaway to avoid that constraint.

That restriction was lifted in 24H2 (Ge), but the Breakaway workaround is still very much required for downlevel.

RECOMMEND: Rephrase the description (either in painful detail, or more tersely and point to the source code for the full explanation). The latter's preferable IMO as it makes the code self-documenting - better for the long haul

dev/Deployment/Deployer.cpp
```diff
@@ -0,0 +1,164 @@
...
+#include <DeploymentActivityContext.h>
+#include <PackageDefinitions.h>
+#include <Deployer.h>
+#include <PackageUtilities.h>
+#include <PackageRegistrar.h>
+#include <functional>
+
+using namespace winrt;
+
+namespace WindowsAppRuntime::Deployment::Deployer
```
DrusTheAxe: Name's confusing with all the other Deploy* phrasing.

RECOMMEND: Better name. (Yeah, naming, hard. Still...)

SUGGESTION: Could go WIL-style "WindowsAppRuntime::Deployment::details".

Is this just licensing? Then namespace WindowsAppRuntime::Deployment::LicenseManager, ...LicenseManagement, or Licensing are good choices

dev/Deployment/Deployer.cpp
guimafelipe: I will clean this up. It was before the other PR where I cleaned this up on main branch.

dev/Deployment/Deployer.h
DrusTheAxe: *Helper always leaves a bad taste in my mouth. Maybe it's just me, just strikes me as "couldn't come up with a good name".

Could go WIL style: `namespace WindowsAppRuntime::Deployment::details`

Or break this up further into Licensing vs PackageManagement vs PackageInformation

Or....?

(Naming, hard :P)

dev/Deployment/Deployer.h
DrusTheAxe: Why are all the attributes `const`?

Normally you'd make the attributes non-const and pass around a const structure/object

dev/Deployment/Deployer.h
```diff
@@ -0,0 +1,50 @@
...
+        const std::filesystem::path packagePath;
+        const bool useExistingPackageIfHigherVersion;
```
DrusTheAxe: `const bool useExistingPackageIfHigherVersion{};`

Always initialize class/struct attributes in C++

dev/Deployment/Deployer.h
```diff
@@ -0,0 +1,50 @@
...
+        const bool isSingleton;
```
DrusTheAxe: `const bool isSingleton{};`

Always initialize class/struct attributes in C++

dev/Deployment/Deployer.h
```diff
@@ -0,0 +1,50 @@
...
+    };
+
+    // Proxy/Wrapper for license installer. Open possibility to add more methods if needed.
+    struct ILicenseInstaller
+    {
+        virtual HRESULT InstallLicenseFile(const std::wstring& licenseFilename) = 0;
```
DrusTheAxe: Why a pure virtual base class?

At a minimum, why not a NOOP base class? e.g.
```
virtual HRESULT InstallLicenseFile([[maybe_unused]] const std::wstring& licenseFilename)
{
    return S_OK;
}
```

dev/Deployment/Deployer.h
```diff
@@ -0,0 +1,50 @@
...
+    };
+
+    // Get license files from the specified path pattern
+    HRESULT GetLicenseFiles(const std::wstring& licenseFileSpec, std::vector<std::wstring>& licenseFiles);
+
+    // Install license files
+    HRESULT InstallLicenses(
+        const std::vector<std::wstring>& licenseFiles,
+        std::filesystem::path licensePath,
```
DrusTheAxe: Why not const?

Why pass-by-value and not reference?

Should be `const std::filesystem::path&` ?

dev/Deployment/Deployer.h
```diff
@@ -0,0 +1,50 @@
...
+        ::WindowsAppRuntime::Deployment::Activity::Context& initializeActivityContext,
+        const std::function<std::wstring(const std::wstring&)>& getPackagePathFunc);
+
+    // Package deployment
+    HRESULT DeployPackages(
+        const std::vector<DeploymentPackageArguments>& deploymentPackageArguments,
+        const bool forceDeployment,
+        ::WindowsAppRuntime::Deployment::Activity::Context& initializeActivity,
+        const std::function<HRESULT()>& startupNotificationsLongRunningPlatformFunc);
+}
```
DrusTheAxe: File ends w/o newline making github crank

SUGGESTION: Put cursor at end of line 50 and hit `<ENTER>`

dev/Deployment/DeploymentManager.cpp
```diff
@@ -327,308 +330,67 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
...
 
-    // Gets the package path, which is a fast and reliable way to check if the package is
-    // at least staged on the device, even without package query capabilities.
-    std::wstring DeploymentManager::GetPackagePath(std::wstring const& packageFullName)
+    HRESULT DeploymentManager::InstallLicenses(const std::wstring& frameworkPackageFullName, ::WindowsAppRuntime::Deployment::Activity::Context& initializeActivityContext)
     {
-        UINT32 pathLength{};
-        const auto rc{ GetPackagePathByFullName(packageFullName.c_str(), &pathLength, nullptr) };
-        if (rc == ERROR_NOT_FOUND)
+        class LicenseInstallerProxy : public ::WindowsAppRuntime::Deployment::Deployer::ILicenseInstaller
```
DrusTheAxe: Why a class definition inside a function?

RECOMMEND: Move to separate file. Licenseing.h, LicenseManagement.h, LicenseManager.h, etc. like the namespace suggestion

DrusTheAxe: Sorry but I'm not following. What's the reason for ILicenseInstaller and LicensingInstallerProxy? Adds complexity but I'm not seeing why it's needed or how it helps. Please enlighten me

guimafelipe:
> In the description "...depending on the arguments (in this case, if it is full trust or not)."
> 
> Isn't that based on if it's **Centennial** or not?
> 
> Package management had a restriction where the caller needed to be...
> 
> 1. Packaged process + AppContainer + packageManagement capability
> 2. Unpackaged process + IL>=Medium
> 3. Packaged process + IL>=Medium + NOT Centennial
> 4. Packaged process + IL>=Medium + Centennial + caller's Publisher == target's Publisher
> 
> Thus only packages with Publisher=Microsoft Corporation could do packagement management operations on WinAppSDK packages :-( and thus, Breakaway to avoid that constraint.
> 
> That restriction was lifted in 24H2 (Ge), but the Breakaway workaround is still very much required for downlevel.
> 
> RECOMMEND: Rephrase the description (either in painful detail, or more tersely and point to the source code for the full explanation). The latter's preferable IMO as it makes the code self-documenting - better for the long haul

This was a bit confusing to me while comparing the documentation to what was actually in the code. For example: the final lines of the [spec description](https://github.com/microsoft/WindowsAppSDK/blob/main/specs/Deployment/DeploymentAPI.md#description) says that "Alternatively, unpackaged apps that are full trust or MSIX packaged apps that have the packageManagement restricted capability can use the Deployment API to get these packages installed.".

But, on the actual code that is live in main, we have this:
```c++
// If the current application has runFullTrust capability, then Deploy the target package in a Breakaway process.
// Otherwise, call PackageManager API to deploy the target package.
// The Singleton package will always set true for forceDeployment and the running process will be terminated to update the package.
if (initializeActivity.GetIsFullTrustPackage())
{


    RETURN_IF_FAILED(AddOrRegisterPackageInBreakAwayProcess(packagePath, useExistingPackageIfHigherVersion, forceDeployment || isSingleton));
}
else
{
    RETURN_IF_FAILED(AddOrRegisterPackage(packagePath, useExistingPackageIfHigherVersion, forceDeployment || isSingleton));
}
```
We are choosing based on if it is "Full trust" or not. And we set the full trust [here](https://github.com/microsoft/WindowsAppSDK/blob/b208e39ff9cdffabedc4dab92af330a96a2cae8a/dev/Deployment/DeploymentManager.cpp#L210-L218):
```c++
auto& initializeActivityContext{ ::WindowsAppRuntime::Deployment::Activity::Context::Get() };
const bool isPackagedProcess{ AppModel::Identity::IsPackagedProcess() };
const int integrityLevel = Security::IntegrityLevel::GetIntegrityLevel();
if (isPackagedProcess && integrityLevel >= SECURITY_MANDATORY_MEDIUM_RID)
{
    initializeActivityContext.SetIsFullTrustPackage();
}


::WindowsAppRuntime::Deployment::Activity::Context::Get().SetIsFullTrustPackage();
```
This does match what I understood by the documentation, there is no mention of the capabilities that are told as needed. And notice that the line right after the if clause sets full trust for everything (which I believe it is a bug and I have a [Pull Request](https://github.com/microsoft/WindowsAppSDK/pull/5857) addressing it).

So, what I put in the description was following what is in the code behavior currently. Do you think this need to be changed?

dev/Deployment/PackageRegistrar.cpp
```diff
@@ -0,0 +1,145 @@
...
+        const winrt::Windows::Foundation::IAsyncOperationWithProgress<winrt::Windows::Management::Deployment::DeploymentResult, winrt::Windows::Management::Deployment::DeploymentProgress>& deploymentOperation,
+        ::WindowsAppRuntime::Deployment::Activity::Context& activityContext)
+    {
+        deploymentOperation.get();
+
```
DrusTheAxe: Why the blank line?

SUGGESTION: Delete line 24

dev/Deployment/PackageRegistrar.cpp
```diff
@@ -0,0 +1,145 @@
...
+            activityContext.SetDeploymentErrorInfo(
+                deploymentOperationExtendedHResult,
+                deploymentResult.ErrorText().c_str(),
+                deploymentResult.ActivityId());
+        }
+
+        // If deploymentOperationHResult indicates success, take that, ignore deploymentOperationExtendedHResult.
+        // Otherwise, return deploymentOperationExtendedHResult if there is an error in it, if not, return deploymentOperationHResult.
+        return SUCCEEDED(deploymentOperationHResult) ? deploymentOperationHResult :
+            (FAILED(deploymentOperationExtendedHResult) ? deploymentOperationExtendedHResult : deploymentOperationHResult);
```
DrusTheAxe: the logic here is

```
a=ErrorCode
b=ExtendedErrorCode
if (a=SUCCESS) return a
elseif (b=FAILED) return b
else (a=FAILED) return a
```

IOW return (a=FAILED and b=FAILED) ? b : a

SUGGESTION:
```
return (FAILED(deploymentOperationHResult) && FAILED(deploymentOperationExtendedHResult) ? return deploymentOperationExtendedHResult : deploymentOperationHResult;
```

dev/Deployment/PackageRegistrar.cpp
```diff
@@ -0,0 +1,145 @@
...
+    // Helper method to process deployment operation results and extract error information
+    inline HRESULT ProcessDeploymentOperationResult(
+        const winrt::Windows::Foundation::IAsyncOperationWithProgress<winrt::Windows::Management::Deployment::DeploymentResult, winrt::Windows::Management::Deployment::DeploymentProgress>& deploymentOperation,
+        ::WindowsAppRuntime::Deployment::Activity::Context& activityContext)
+    {
+        deploymentOperation.get();
+
+        const auto deploymentResult{ deploymentOperation.GetResults() };
+        HRESULT deploymentOperationHResult{};
+        HRESULT deploymentOperationExtendedHResult{};
```
DrusTheAxe: Those are a wordy. Common practice is to name an HRESULT variable `hr` if there's just one, or `hrFoo`, `hrBar`, `hrWhatTheHResultCameFrom`, ... if multiple

RECOMMEND: `hrError` and `hrExtendedError`

dev/Deployment/PackageRegistrar.cpp
DrusTheAxe: and ditto the `try`

RECOMMEND: Merge lines 55+56 onto the end of 54

dev/Deployment/PackageRegistrar.cpp
```diff
@@ -0,0 +1,145 @@
...
+    HRESULT AddOrRegisterPackage(
+        const std::filesystem::path& path,
+        const bool useExistingPackageIfHigherVersion,
+        const bool forceDeployment,
+        winrt::Windows::Management::Deployment::IPackageManager& packageManager,
+        ::WindowsAppRuntime::Deployment::Activity::Context& activityContext
+    )
+    try
+    {
+
```
DrusTheAxe: NIT: Remove blank line

dev/Deployment/PackageDeployment.cpp
```diff
@@ -0,0 +1,164 @@
...
+            initializeActivity.SetCurrentResourceId(package.packageIdentifier);
+            if (package.useExistingPackageIfHigherVersion)
+            {
+                initializeActivity.SetUseExistingPackageIfHigherVersion();
+            }
+
+            // If the current application has runFullTrust capability, then Deploy the target package in a Breakaway process.
+            // Otherwise, call PackageManager API to deploy the target package.
+            // The Singleton package will always set true for forceDeployment and the running process will be terminated to update the package.
+            if (initializeActivity.GetIsFullTrustPackage())
```
DrusTheAxe: This is overly broad. Breakway is needed if the caller...
1. Has package identity
2. IL>=Medium
3. RuntimeBehavior=Centennial
4. Publisher != Microsoft Corporation (target's publisher)
5. On Windows <24H2 (Ge)

RECOMMEND:
```
#include <AppModel.Identity.h>
#include <IsWindowsVersion.h>
#include <Security.IntegrityLevel.h>
...
if (!::WindowsVersion::IsWindows11_24H2OrGreater() && ::AppModel::Identity::IsPackagedProcess() && (::Security::IntegrityLevel::GetIntegrityLevel() >= SECURITY_MANDATORY_MEDIUM_RID))
{
    ...breakaway is needed...
```

Skip checking #4 (Publisher) as too rare to bother (and better to not fork behavior by publisher) and #3 (Centennial) as too hard to detect

dev/Deployment/PackageDeployment.cpp
```diff
@@ -0,0 +1,164 @@
...
+        {
+            initializeActivity.Reset();
+            initializeActivity.SetInstallStage(::WindowsAppRuntime::Deployment::Activity::DeploymentStage::AddPackage);
+            initializeActivity.SetCurrentResourceId(package.packageIdentifier);
+            if (package.useExistingPackageIfHigherVersion)
+            {
+                initializeActivity.SetUseExistingPackageIfHigherVersion();
+            }
+
+            // If the current application has runFullTrust capability, then Deploy the target package in a Breakaway process.
```
DrusTheAxe: >If the current application has runFullTrust capability

This is incorrect. `RuntimeBehavior="packagedClassicApp"` (aka Centennial) needs Breakaway, `RuntimeBehavior="win32App"` (aka Win32alacarte) doesn't.

dev/Deployment/Deployer.cpp
```diff
@@ -0,0 +1,164 @@
...
+            {
+                initializeActivity.SetUseExistingPackageIfHigherVersion();
+            }
+
+            // If the current application has runFullTrust capability, then Deploy the target package in a Breakaway process.
+            // Otherwise, call PackageManager API to deploy the target package.
+            // The Singleton package will always set true for forceDeployment and the running process will be terminated to update the package.
+            if (initializeActivity.GetIsFullTrustPackage())
+            {
+
```
DrusTheAxe: NIT: Delete blank line

dev/Deployment/Deployer.cpp
```diff
@@ -0,0 +1,164 @@
...
+                    forceDeployment || package.isSingleton,
+                    packageManager,
+                    initializeActivity
+                ));
+            }
+
+            // Always restart Push Notifications Long Running Platform when Singleton package is processed and installed.
+            if (package.isSingleton)
+            {
+                // wil callback is set up to log telemetry events for Push Notifications LRP.
```
DrusTheAxe: NIT: `// WIL callback...`

The library's name is WIL, its namespace in code is `wil`

dev/Deployment/Deployer.cpp
```diff
@@ -0,0 +1,164 @@
...
+                LOG_IF_FAILED_MSG(startupNotificationsLongRunningPlatformFunc(), "Restarting Notifications LRP failed in all 3 attempts.");
```
DrusTheAxe: NIT: `std::ignore = LOG_IF_FAILED(...);`

DrusTheAxe: Where's `3` come from?

RECOMMEND: Rephase e.g. `Failed to restart Notifications' LongRunningProcess (LRP)`

DrusTheAxe (changes requested):

dev/Deployment/DeploymentManager.cpp
guimafelipe: The reason for that is that `LicenseInstaller` is an object is passed to the function. This object comes from outside as I understood from [this PR](https://github.com/microsoft/WindowsAppSDK/pull/1790). I wanted to have control of this on tests as I suspect some of the failures we are currently having on `Initialize` comes from it. So i created the `ILicenseInstaller` interface owned by the Deployment code for whoever calls it to pass an object that satisfy this interface. This is a very OOP solution and agree that might be better options. In previous versions of this PR, I thought about just passing a function pointer for the install licenses method.

dev/Deployment/Deployer.cpp
guimafelipe: This part of the code was moved exactly as it was originally. I don't know where the 3 times decision comes from, but that is the behavior of the function as it is written internally.

dev/Deployment/PackageDeployment.cpp
guimafelipe: This comes to the reply I made to your initial comment on the PR. This was what the code was and I think it was not what the documentation was saying either. Currently, everything is running as "Full trust" too.

dev/Deployment/Deployer.h
guimafelipe: I was not accounting for that to ever happen. It forces it to receive an implementation. I think it make it clear that this is an argument that needs to be passed to the function and doesn't have a default implementation.

dev/Deployment/DeploymentManager.cpp
guimafelipe: I put the construction of the package path there because that is the entry point of all the deployment logic. In my point of view, this code does not need to be object oriented. It is a API of two functions (`GetStatus` and `Initialize`) living as static methods of the `DeploymentManager` class, which is basically a namespace at this point.

If you consider the whole API, both methods are entry points of a big procedure. We call them, they do something and then they finish the work. Most of the preparation for that work that it is needed to be done I think we can push out to the closest of the entry point as possible (for example, building the arguments), so the inner functions focus more on business logic. This way, it is easier to track the behavior of the inner functions and the we isolate the construction of the arguments they need to act closer to the entry point.

In this case, I think the single responsibility would be separating the creation of the arguments and data necessary to do the work (data instantiation) from the business logic. It makes the functions not need to reach to any external data to work or to keep any internal state that could generate side effects. We keep the functions the most testable as possible and remove from them responsibility to instantiate their own data.

edit: Another option is to have this method back to the Licensing namespace and the manager just calls the function that builds this license path. It passed through my mind but I felt it was unnecessary to add a method just to building a string (and it would not need any testing to it either). Usually I try to avoid using functions for simple tasks that will not be needed to isolate, test or reuse.

dev/Deployment/PackageDeployment.h
guimafelipe: The main problem for that is that would make it a hard dependency on generated files and really hard to test in a controlled environment. This also makes it really hard do compile as we need to drag the [PushNotificationsLongRunningPlatform-Startup.h with it an all its dependencies](https://github.com/microsoft/WindowsAppSDK/blob/b208e39ff9cdffabedc4dab92af330a96a2cae8a/dev/PushNotifications/PushNotificationsLongRunningTask.StartupTask/PushNotificationsLongRunningPlatform-Startup.h#L4-L9) to the compilation tree.

Passing a function pointer here is the easiest way to break this hard dependency. I like it because we can the function is an argument as any other and it is not an anonymous lambda. I think it gives flexibility to maintain it in the future as the hard dependency will not exist and the responsibility to pass the function bubbles up to the caller.

Another option was to implement link substitution for tests (using a different source file with the same name but created only for tests, and the linker uses it on the compilation). But I think that would be even more confusing. What do you think?

dev/Deployment/PackageDeployment.cpp
guimafelipe: I did that to separate the logic (creating the arguments based on the arguments and the external constant variables) from the IO execution (the calling the deployment of the packages to install them on the machine). 

This makes `GetDeploymentPackageArgs` testable independently of the `DeployPackages`. Its behavior depends only on its arguments and we track regressions to it independently from what happens on the actual deployment. 

`DeployPackages` is a function that we don't need to test when separated from the arguments construction. Its only task is to check an if to call either one or another function. We would rather test these internal functions instead and track their behavior.

I opt to receive multiple parameters instead of accessing global variables (like the previous `AcivityContext::Get()` because it makes more obvious that nothing external from the function will affect its behavior and everything that matters are what the functions is receiving as its parameters. The dependencies were always there, it just makes them explicit.

guimafelipe:
> Given the scope of this refactor, I’d strongly recommend exercising the functionality manually end to end across all three scenarios as well—packaged, unpackaged, and AppContainer. It would also be valuable to add coverage for AppContainer scenarios and/or sample launch tests, even if that happens in a future PR.

I completely agree that all the scenarios need to be tested and I think we should have all of it automated instead of testing it manually. I think it is a worth effort to ensure it on the goal of reducing all the current errors and crashes from the deployment area.

dev/Deployment/DeploymentManager.cpp
guimafelipe: The `Licensing` namespace defines a interface `ILicenseInstaller` with the `installLicences(...)` method. Before, it was using the license installer object directly, but this was a hard dependency on the license installer class, as I commented as a response on the comment below. The `LicenseInstallerProxy` is a implementation of this interface, and is responsibility of the caller to decide how to implement it.

This would only be called in that location, but I agree on moving it to a separated file.

dev/Deployment/PackageRegistrar.cpp
guimafelipe: That could be done, but I don't think it is a behavior that general. Initially, it was not even a separated function. I separated it in case of we wanting to test it isolated. I don't think this function would make sense out of `PackageRegistrar` namespace. I would love to hear more of your opinions about it.

dev/Deployment/PackageDeployment.h
guimafelipe: Refactored and redesigned the code to not need function pointers.

test/DeploymentUnitTests/LicensingTests.cpp
```diff
@@ -0,0 +1,383 @@
...
+
+        // Set up the mock to fail on a specific file
+        void SetupFailureOnFile(const std::wstring& filename, HRESULT errorCode)
+        {
+            m_expectedFailureMap[filename] = errorCode;
+        }
+
+        void SetShouldThrowException(bool shouldThrow)
+        {
+            m_shouldTrhowException = shouldThrow;
```
ssparach: typo: `m_shouldThrowException`

DrusTheAxe (changes requested):
Why `test/DeploymentUnitTests/` and not `test/Deployment/...whatever...` (like, the current `test/Deployment/API`)?

`test/foo` should match `dev/foo`. Under that is fair game as the feature desires

guimafelipe:
> Why `test/DeploymentUnitTests/` and not `test/Deployment/...whatever...` (like, the current `test/Deployment/API`)?
> 
> `test/foo` should match `dev/foo`. Under that is fair game as the feature desires

Moved it to the correct directory.

PR #5857, opened 2025-09-23
Fixing set full trust after if

guimafelipe:
Right after the `if` where we are setting if we should run full trust or not, we set it for every execution. This change removes the line that sets it on every case.

WARNING:


dev/Deployment/DeploymentManager.cpp
```diff
@@ -215,7 +215,6 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
             initializeActivityContext.SetIsFullTrustPackage();
```
ssparach, line 217:
```suggestion
              // Marking the package as full trust because the call originates from a MediumIL package.
            // For a packaged MediumIL app (with a non-Microsoft publisher) to deploy main and singleton packages, a breakaway helper process is required.
            initializeActivityContext.SetIsFullTrustPackage();
```

dev/Deployment/DeploymentManager.cpp
```diff
@@ -215,7 +215,6 @@ namespace winrt::Microsoft::Windows::ApplicationModel::WindowsAppRuntime::implem
...
         }
 
-        ::WindowsAppRuntime::Deployment::Activity::Context::Get().SetIsFullTrustPackage();
```
ssparach, line 218: Thanks for cleaning this up! 

@SaChinta-MSFT Just to confirm: setting the package to full trust inside the conditional should cover the MediumIL non-Microsoft publisher scenario. Was there any specific reason for the previous unconditional full-trust setting afterwards, or can we safely proceed with this removal? Thanks!

ssparach:
Could you share details on how this change was validated?
Specifically, has it been tested across different scenarios—such as packaged, unpackaged etc.—to ensure there are no regressions? Thanks!

guimafelipe:
> Could you share details on how this change was validated? Specifically, has it been tested across different scenarios—such as packaged, unpackaged etc.—to ensure there are no regressions? Thanks!

It was validated by all the Deployment tests that currently run in the different scenarios of UAP tests.

dev/Deployment/DeploymentManager.cpp
SaChinta-MSFT: It can be removed safely.

SaChinta-MSFT (approved):

PR #5863, opened 2025-09-24
Document PackageValidator API in PackageManagement.md

letao-msft:
Updating PackageManagement.md to add documentation of the new PackageValidator APIs that we are implementing.

### DIFF
--- a/specs/packagemanager/PackageManagement.md
+++ b/specs/packagemanager/PackageManagement.md
@@ -23,6 +23,7 @@ but with additional functionality, improved developer experience and performance
   - [3.10. PackageVolume Repair](#310-packagevolume-repair)
   - [3.11. Usability](#311-usability)
   - [3.12. Is\*Provisioned()](#312-312-isprovisioned)
+  - [3.13. PackageValidator](#313-packagevalidator)
 - [4. Examples](#4-examples)
   - [4.1. AddPackageAsync()](#41-addpackageasync)
   - [4.2. AddPackageByUriAsync()](#42-addpackagebyuriasync)
@@ -73,6 +74,7 @@ Additional functionality includes:
 * IsPackageRegistrationPending -- Is there an update waiting to register?
 * PackageSets -- Batch operations
 * PackageRuntimeManager -- Batch operations for use at runtime via Dynamic Dependencies
+* PackageValidator -- Validate a package has expected identity, signature, etc. before adding/staging
 * Usability -- Quality-of-Life enhancements
 
 ## 3.1. API Structure
@@ -397,6 +399,24 @@ package management APIs in Windows (e.g. Windows.Management.Deployment.PackageMa
 Is\*Provisioned\*() methods determine if the target is provisioned.
 
 These methods require administrative privileges.
+
+## 3.13. PackageValidator
+
+This API allows callers to verify that packages being processed by PackageDeploymentManager.AddPackageAsync
+and PackageDeploymentManager.StagePackageAsync match what are expected from their URI.
+
+When adding or staging a package from an external source such as HTTP URI or uncontrolled file location,
+an attacker can potentially intercept and tamper with the package data being read, causing a malicious
+package to be installed instead of the expected one.  Verifying the identity and signature of target
+packages helps ensure that such tampering has not happened.
+
+The following PackageValidators are provided and available for use directly:
+* PackageFamilyNameValidator: Validates that the package has the expected package family name.
+* PackageMinimumVersionValidator: Validates that the package has at least the expected minimum version number.
+* PackageCertificateEkuValidator: Validates that the certificate used to sign the package contains the expected Extended Key Usage (EKU) value.
+
+Custom validators can also be implemented using the IPackageValidator interface, which allows verifying
+any part of the package’s footprint data (manifest, block map, and digital signature) in any desired manner.
 
 # 4. Examples
 
@@ -761,6 +781,100 @@ PackageVersion ToVersion(uint major, uint minor, uint build, uint revision) =>
     };
 ```
 
+## 4.9. PackageValidator
+
+### 4.9.1. Using built-in PackageValidators
+
+This example shows how to use built-in PackageValidators to verify package family name, minimum version, and certificate EKU.
+
+```c++/winrt
+auto packageDeploymentManager{ winrt::Microsoft::Windows::Management::Deployment::PackageDeploymentManager::GetDefault() };
+const auto packageUri{ L"https://example.com/package.msix"};
+const winrt::hstring package{ packageUri.c_str() };
+
+IVector<IPackageValidator> validators{ winrt::single_threaded_vector<IPackageValidator>() };
+validators.Append(winrt::Microsoft::Windows::Management::Deployment::PackageFamilyNameValidator(L"ExpectedFamilyName_abcdefg123"));
+validators.Append(winrt::Microsoft::Windows::Management::Deployment::PackageMinimumVersionValidator(2, 0, 0, 0));
+validators.Append(winrt::Microsoft::Windows::Management::Deployment::PackageCertificateEkuValidator(L"1.3.6.1.4.1.311.2.1.11");
+
+winrt::Microsoft::Windows::Management::Deployment::AddPackageOptions addOptions;
+addOptions.PackageValidators.Insert(packageUri, validators);
+
+auto deploymentOperation{ packageDeploymentManager.AddPackageAsync(package, addOptions) };
+auto deploymentResult{ WaitForDeploymentOperation(deploymentOperation) };
+// deploymentOperation will fail with APPX_E_DIGEST_MISMATCH if any packageValidator rejects the package
+```
+
+### 4.9.2. Using custom PackageValidators
+
+This example shows how to implement a custom PackageValidator using IPackageValidator interface, and use it to verify a package.
+
+```idl
+// MyCustom.idl
+namespace MyCustom
+{
+    runtimeclass MyPackageValidator : [default] Microsoft.Windows.Management.Deployment.IPackageValidator
+    {
+        Boolean IsPackageValid(AppxPackagingObject packagingObject);
+    }
+}
+```
+
+```c++/winrt
+// MyCustom.cpp
+// (Assume other standard cppwinrt-generated code related to MyCustom.MyPackageValidator are present.
+// This sample only shows the custom hand-written parts.)
+namespace winrt::MyCustom::implementation
+{
+    // Implements a custom package validator that checks that the package does not delcare any capabilities.
+    struct MyCustomPackageValidator : MyCustomPackageValidatorT<MyCustomPackageValidator>
+    {
+        MyCustomPackageValidator();
+
+        bool IsPackageValid(winrt::Microsoft::Windows::Management::Deployment::AppxPackagingObject object)
+        {
+            winrt::com_ptr<IAppxPackageReader> packageReader;
+            if (FAILED(object.as<IAppxPackageReader>(packageReader.put())))
+            {
+                // object is not an msix package as expected (i.e. it is a bundle), reject it
+                return false;
+            }
+
+            winrt::com_ptr<IAppxManifestReader> manifestReader;
+            THROW_IF_FAILED(packageReader->GetManifest(manifestReader.put()));
+
+            winrt::com_ptr<IAppxManifestReader3> manifestReader3;
+            THROW_IF_FAILED(manifestReader->QueryInterface(manifestReader3.put()));
+
+            winrt::com_ptr<IAppxManifestCapabilitiesEnumerator> capabilitiesEnumerator;
+            THROW_IF_FAILED(manifestReader3->GetCapabilitiesByCapabilityClass(APPX_CAPABILITY_CLASS_ALL, capabilitiesEnumerator.put()));
+
+            BOOL hasCapabilities{};
+            THROW_IF_FAILED(capabilitiesEnumerator->GetHasCurrent(&hasCapabilities));
+            return !hasCapabilities;
+        }
+    };
+}
+```
+
+```c++/winrt
+// At the call site to PackageDeploymentManager, using the custom package validator
+auto packageDeploymentManager{ winrt::Microsoft::Windows::Management::Deployment::PackageDeploymentManager::GetDefault() };
+
+const auto packageUri{ L"https://example.com/package.msix"};
+const winrt::hstring package{ packageUri.c_str() };
+
+IVector<IPackageValidator> validators{ winrt::single_threaded_vector<IPackageValidator>() };
+validators.Append(winrt::MyCustom::MyPackageValidator());
+
+winrt::Microsoft::Windows::Management::Deployment::AddPackageOptions options;
+options.PackageValidators.Insert(packageUri, validators);
+
+auto deploymentOperation{ packageDeploymentManager.AddPackageAsync(package, options) };
+auto deploymentResult{ WaitForDeploymentOperation(deploymentOperation) };
+// deploymentOperation will fail with APPX_E_DIGEST_MISMATCH if any packageValidator rejects the package
+```
+
 # 5. Remarks
 
 ## 5.1. Platform Support
@@ -856,6 +970,41 @@ namespace Microsoft.Windows.Management.Deployment
         Ready                = 1,
         NewerAvailable       = 2,
     };
+
+    [contract(PackageDeploymentContract, 3)]
+    [default_interface]
+    runtimeclass AppxPackagingObject{
+        // This is an interop class for COM types defined in AppxPackaging.idl.
+        // The WinRT side has no methods or properties, but the object supports QueryInterface into the COM interfaces
+        // IAppxPackageReader or IAppxBundleReader, whichever is relevant for the object being read.
+    }
+
+    [contract(PackageDeploymentContract, 3)]
+    interface IPackageValidator
+    {
+        Boolean IsPackageValid(AppxPackagingObject packagingObject);
+    }
+
+    [contract(PackageDeploymentContract, 3)]
+    runtimeclass PackageFamilyNameValidator : [default] IPackageValidator
+    {
+        PackageFamilyNameValidator(String expectedPackageFamilyName);
+        Boolean IsPackageValid(AppxPackagingObject packagingObject);
+    }
+
+    [contract(PackageDeploymentContract, 3)]
+    runtimeclass PackageMinimumVersionValidator : [default] IPackageValidator
+    {
+        PackageMinimumVersionValidator(UINT major, UINT minor, UINT build, UINT revision);
+        Boolean IsPackageValid(AppxPackagingObject packagingObject);
+    }
+
+    [contract(PackageDeploymentContract, 3)]
+    runtimeclass PackageCertificateEkuValidator : [default] IPackageValidator
+    {
+        PackageCertificateEkuValidator(String expectedCertificateEku);
+        Boolean IsPackageValid(AppxPackagingObject packagingObject);
+    }
 
     /// The progress status of the deployment request.
     /// @see https://learn.microsoft.com/uwp/api/windows.management.deployment.deploymentprogress.state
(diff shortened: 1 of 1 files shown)


letao-msft:

specs/packagemanager/PackageManagement.md
```diff
@@ -761,6 +781,100 @@ PackageVersion ToVersion(uint major, uint minor, uint build, uint revision) =>
...
 
+## 4.9. PackageValidator
+
+### 4.9.1. Using built-in PackageValidators
+
+This example shows how to use built-in PackageValidators to verify package family name, minimum version, and certificate EKU.
+
+```c++/winrt
+auto packageDeploymentManager{ winrt::Microsoft::Windows::Management::Deployment::PackageDeploymentManager::GetDefault() };
+const auto packageUri{ L"https://example.com/package.msix"};
```
DrusTheAxe: RECOMMEND: `https://contoso.com/package.msix`

Contoso is Microsoft's goto fictional company

There's even a wealth of information about the fictional corporation e.g.
* https://learn.microsoft.com/en-us/microsoft-365/enterprise/contoso-overview
* https://learn.microsoft.com/en-us/microsoft-365/enterprise/contoso-case-study

specs/packagemanager/PackageManagement.md
```diff
@@ -761,6 +781,100 @@ PackageVersion ToVersion(uint major, uint minor, uint build, uint revision) =>
...
+const winrt::hstring package{ packageUri.c_str() };
```
DrusTheAxe: Why both `packageUri` and `package` variables?

`package` type is `const wchar_t*` (or maybe `const wchar_t[]`?)
`packageUri` type is `const winrt::hstring`
Why have both?

Ultimately they're used on line 801 adding the `hstring` to `PackageValidators`
and line 803 passing the `const wchar_t*` to `AddPackageAsync()`
but that takes a `winrt::hstring`. This makes a deep copy of the string when only the hstring is needed in for both calls

SUGGESTION: Replace lines 792-793 with `const winrt::hstring package{ L"https://example.com/package.msix"};`

specs/packagemanager/PackageManagement.md
```diff
@@ -761,6 +781,100 @@ PackageVersion ToVersion(uint major, uint minor, uint build, uint revision) =>
...
+
+IVector<IPackageValidator> validators{ winrt::single_threaded_vector<IPackageValidator>() };
+validators.Append(winrt::Microsoft::Windows::Management::Deployment::PackageFamilyNameValidator(L"ExpectedFamilyName_abcdefg123"));
```
DrusTheAxe: PublisherId is always 13 characters. "abcdefg123" = 10 chars

RECOMMEND: `"ExpectedFamilyName_1234567890abc"`

specs/packagemanager/PackageManagement.md
```diff
@@ -761,6 +781,100 @@ PackageVersion ToVersion(uint major, uint minor, uint build, uint revision) =>
...
+validators.Append(winrt::Microsoft::Windows::Management::Deployment::PackageMinimumVersionValidator(2, 0, 0, 0));
+validators.Append(winrt::Microsoft::Windows::Management::Deployment::PackageCertificateEkuValidator(L"1.3.6.1.4.1.311.2.1.11");
+
+winrt::Microsoft::Windows::Management::Deployment::AddPackageOptions addOptions;
+addOptions.PackageValidators.Insert(packageUri, validators);
+
+auto deploymentOperation{ packageDeploymentManager.AddPackageAsync(package, addOptions) };
+auto deploymentResult{ WaitForDeploymentOperation(deploymentOperation) };
```
DrusTheAxe: `WaitForDeploymentOperation()` doesn't exist

RECOMMEND: Replace lines 803-805 with

```
auto deploymentOperation{ packageDeploymentManager.AddPackageAsync(package, addOptions) };
auto deploymentResult{ deploymentOperation().get() };
if (deploymentResult.Status() == winrt::Microsoft::Windows::Management::Deployment::PackageDeploymentStatus::CompletedSuccess)
{
    printf(L"Success!");
}
else // deploymentResult.Status() == winrt::Microsoft::Windows::Management::Deployment::PackageDeploymentStatus::CompletedFailure)
{
    if (deploymentResult.Error() == APPX_E_DIGEST_MATCH)
    {
        printf("Digest fails validation. The retrieved package isn't a validly expected one");
    }
    else
    {
        printf("Error 0x%08X  ExtendedError 0x%08X  %ls",
               deploymentResult.Error(), deploymentResult.ExtendedError(),
               deploymentResult.ErrorMessage().c_str());
    }
}

File: specs/packagemanager/PackageManagement.md [hunk 5]
```diff
@@ -761,6 +781,100 @@ PackageVersion ToVersion(uint major, uint minor, uint build, uint revision) =>
     };
 ```
 
+## 4.9. PackageValidator
+
+### 4.9.1. Using built-in PackageValidators
+
+This example shows how to use built-in PackageValidators to verify package family name, minimum version, and certificate EKU.
+
+```c++/winrt
```
DrusTheAxe: Samples are recommended in C#

Unless there's a need to use another language. But I don't see the need here. Do you?

RECOMMEND: Use C# for examples

```c#
using Microsoft.Windows.Management.Deployment;

var pdm = PackageDeploymentManager().GetDefault();
var package = "https://contoso.com/package.msix";

var options = new AddPackageOptions();
var validators = options.Validators();
validators.Add(new PackageFamilyNameValidator("ExpectedFamilyName_1234567890abc"));
validators.Add(new PackageMinimumVersionValidator(2, 0, 0, 0));
validators.Add(new PackageCertificateEkuValidator("1.3.6.1.4.1.311.2.1.11"));

var deploymentResult = await pdm.AddPackageAsync(package, addOptions);
if (deploymentResult.Status == PackageDeploymentStatus.CompletedSuccess)
{
    Console.WriteLine("Success");
}
else // deploymentResult.Status == PackageDeploymentStatus.CompletedFailure
{
    var error = deploymentResult.Error.HResult;
    const int APPX_E_DIGEST_MISMATCH = 0x80080219;
    if (error = APPX_E_DIGEST_MISMATCH)
    {
        Console.WriteLine("The retrieved package failed our validation");
    }
    else
    {
        var extendedError = deploymentResult.ExtendedError.HResult;
        var message = deploymentResult.MessageText;
        Console.WriteLine($"Error 0x{error:X08} ExtendedError 0x{extendedError:X08} {message}");
    }
}

File: specs/packagemanager/PackageManagement.md [hunk 6]
```diff
@@ -761,6 +781,100 @@ PackageVersion ToVersion(uint major, uint minor, uint build, uint revision) =>
...
+auto packageDeploymentManager{ winrt::Microsoft::Windows::Management::Deployment::PackageDeploymentManager::GetDefault() };
+const auto packageUri{ L"https://example.com/package.msix"};
+const winrt::hstring package{ packageUri.c_str() };
+
+IVector<IPackageValidator> validators{ winrt::single_threaded_vector<IPackageValidator>() };
```
DrusTheAxe: No need. `AddPackageOptions` runtimeclass should provide a property returning the list e.g. `IVector<IPackageValidator> Validators { get; }`

See `DependencyPackageUris`, `OptionalPackageFamilyNames` and other `IVector<type>` properties in PackageManager.idl

`AddPackageOptions` implementation defines an `IVector<type> m_blah` attribute and the getter method checks if there's no vector object then create and set it, and then return the attribute. Users can access the list in the options object and change its content. They don't get to set (or replace) the collection itself.

specs/packagemanager/PackageManagement.md
```diff
@@ -761,6 +781,100 @@ PackageVersion ToVersion(uint major, uint minor, uint build, uint revision) =>
...
+namespace MyCustom
+{
+    runtimeclass MyPackageValidator : [default] Microsoft.Windows.Management.Deployment.IPackageValidator
+    {
+        Boolean IsPackageValid(AppxPackagingObject packagingObject);
+    }
+}
+```
+
+```c++/winrt
```
DrusTheAxe: Samples should be written in C# unless there's a need for other language(s)

You should be able to do this in C# with C#/WinRT, but I'm not familiar with the syntax. Plus in this case the custom validator needs to interact with the Packaging API (e.g. IAppxPackageReader) which is COM. Can this be done in C#? I suspect so, but the details probably differ significantly from C++. If so that warrants C++ and C#, to illustrate how to do it despite the significant and not well known differences.

RECOMMEND: Provide this example in C# and C++

specs/packagemanager/PackageManagement.md
```diff
@@ -857,6 +971,41 @@ namespace Microsoft.Windows.Management.Deployment
         NewerAvailable       = 2,
     };
 
+    [contract(PackageDeploymentContract, 3)]
+    [default_interface]
+    runtimeclass AppxPackagingObject{
```
DrusTheAxe: NIT: Open braces on separate line

```
runtimeclass AppxPackagingObject
{
...
}
````

Comment (DrusTheAxe) [inline specs/packagemanager/PackageManagement.md]: Oddly generic name ('Object'). I can't think of a good name atm but that's OK. This makes for a good focal point for API Review. I'm sure there'll be feedback and 'vigorous discussion' :P

File: specs/packagemanager/PackageManagement.md [hunk 9]
```diff
@@ -857,6 +971,41 @@ namespace Microsoft.Windows.Management.Deployment
...
+        // This is an interop class for COM types defined in AppxPackaging.idl.
+        // The WinRT side has no methods or properties, but the object supports QueryInterface into the COM interfaces
+        // IAppxPackageReader or IAppxBundleReader, whichever is relevant for the object being read.
```
DrusTheAxe: An `AppxPackageObject` could be an instance of a package reader or a bundle reader.

Could there be 2 different runtimeclasses here, to match the different readers?

Alternatively, this may be a case it makes more sense to use `IInspectable*` (aka `object` in C#) and not invent a new runtimeclass here at all.

If the answer's not clear I recommend leaving this comment open for API Review and we can resolve it there.

specs/packagemanager/PackageManagement.md
```diff
@@ -857,6 +971,41 @@ namespace Microsoft.Windows.Management.Deployment
...
+    }
+
+    [contract(PackageDeploymentContract, 3)]
+    interface IPackageValidator
+    {
+        Boolean IsPackageValid(AppxPackagingObject packagingObject);
```
DrusTheAxe: `IInspectable*` rather than `AppxPackagingObject`? See previous comment

specs/packagemanager/PackageManagement.md
```diff
@@ -398,6 +400,24 @@ Is\*Provisioned\*() methods determine if the target is provisioned.
 
 These methods require administrative privileges.
 
+## 3.13. PackageValidator
+
+This API allows callers to verify that packages being processed by PackageDeploymentManager.AddPackageAsync
```
DrusTheAxe: Add*() has multiple overloads e.g. `AddPackageAsync()`, `AddPackageByUriAsync()`, `AddPackageSetAsync()` and ditto for Stage*().

RECOMMEND: Use more general phrasing e.g.

```
This API enables callers to verify packages processed by add and stage requests match the caller's expected package(s). This functionality is available via `PackageDeploymentManager` methods `Add*Async()`, `Ensure*Async()` and `Stage*Async()`.
```

specs/packagemanager/PackageManagement.md
```diff
@@ -398,6 +400,24 @@ Is\*Provisioned\*() methods determine if the target is provisioned.
...
+and PackageDeploymentManager.StagePackageAsync match what are expected from their URI.
+
+When adding or staging a package from an external source such as HTTP URI or uncontrolled file location,
+an attacker can potentially intercept and tamper with the package data being read, causing a malicious
```
DrusTheAxe: CONSIDER: ...the package data being read (aka man-in-the-middle attacks), resulting in a malicious package...

specs/packagemanager/PackageManagement.md
```diff
@@ -398,6 +400,24 @@ Is\*Provisioned\*() methods determine if the target is provisioned.
...
+package to be installed instead of the expected one.  Verifying the identity and signature of target
+packages helps ensure that such tampering has not happened.
+
+The following PackageValidators are provided and available for use directly:
+* PackageFamilyNameValidator: Validates that the package has the expected package family name.
+* PackageMinimumVersionValidator: Validates that the package has at least the expected minimum version number.
+* PackageCertificateEkuValidator: Validates that the certificate used to sign the package contains the expected Extended Key Usage (EKU) value.
+
+Custom validators can also be implemented using the IPackageValidator interface, which allows verifying
```
DrusTheAxe: SUGGESTION: Phrasing tweak:

```
Custom validators can be implemented using the `IPackageValidator` interface. This can verify any part of the package’s [footprint data](https://learn.microsoft.com/windows/win32/api/appxpackaging/ne-appxpackaging-appx_bundle_footprint_file_type) (manifest, block map, and digital signature).
```

DrusTheAxe (changes requested):

specs/packagemanager/PackageManagement.md
letao-msft: Yeah, this was probably a vestige of modifying the code multiple times.  Changed.

specs/packagemanager/PackageManagement.md
letao-msft: Updated sample

specs/packagemanager/PackageManagement.md
letao-msft: switched to C#

specs/packagemanager/PackageManagement.md
letao-msft: AddPackageOptions.PackageValidators returns a IMap<String, IVector>  Caller doesn't need to create the map, but they still need to create the IVector that gets inserted into the map.

specs/packagemanager/PackageManagement.md
letao-msft: Yes the COM stuff should be doable in C# too, with some interop code.  I'll add an equivalent C# example for the custom validator case once the API is implemented and I can confirm a working sample. [Pending]

specs/packagemanager/PackageManagement.md
letao-msft: I think IInspectable works, and we don't need this runtimeclass.  Deleted this.

specs/packagemanager/PackageManagement.md
letao-msft: Updated

specs/packagemanager/PackageManagement.md
DrusTheAxe: Ahh, missed the IMap<> here.

Hmmm. I know we don't like others handing us a whole list as a property e.g.

```
runtimeclass Foo
{
    IVector<Bar> Bars;
}

Vector<Bar> b;
Foo f;
f.Bars = b;
```

The 'set'ter property is frowned upon

BAD:     `IVector<Bar> Bars;`
OK:     `IVector<Bar> Bars { get; };`

Here the Validators is good but as a Map it's values are a list provided/allocated by the caller. I don't recall seeing that pattern in API Review or if there's any concerns or guidance. I'll check.

specs/packagemanager/PackageManagement.md
```diff
@@ -398,6 +400,26 @@ Is\*Provisioned\*() methods determine if the target is provisioned.
...
+When adding or staging a package from an external source such as HTTP URI or uncontrolled file location,
+a malicious actor might perform a man-in-the-middle attack to intercept and tamper with the package data
+being read, causing a malicious package to be installed instead of the expected one.  The package might
+also be tampered at the source through supply-chain attacks.  Verifying the identity and signature of
+target packages helps ensure that such attacks have not happened.
+
+The following PackageValidators are provided and available for use directly:
+* PackageFamilyNameValidator: Validates that the package has the expected package family name.
+* PackageMinimumVersionValidator: Validates that the package has at least the expected minimum version number.
+* PackageCertificateEkuValidator: Validates that the certificate used to sign the package contains the expected Extended Key Usage (EKU) value.
```
DrusTheAxe: Markdown lines should be 100chars max

VScode has a handy smart paragraph wrap extension ([Reflow Markdown](https://marketplace.visualstudio.com/items?itemName=marvhen.reflow-markdown)) including knowing for markdown not to split a [link](url) even if it runs long.

P.S. [Markdown All in One](https://marketplace.visualstudio.com/items?itemName=yzhang.markdown-all-in-one) is the other super handy VScode extension for Markdown files

specs/packagemanager/PackageManagement.md
```diff
@@ -964,6 +1138,9 @@ namespace Microsoft.Windows.Management.Deployment
 
         Boolean IsLimitToExistingPackagesSupported { get; };    // Requires Windows >= 10.0.22621.0 (aka Win11 22H2)
         Boolean LimitToExistingPackages;
+
+        Boolean IsPackageValidatorsSupported{ get; };
+        IMap<Windows.Foundation.Uri, IVector<IPackageValidator> > PackageValidators{ get; };
```
DrusTheAxe: Got some guidance from API Review folks

Option #1 - Validators property only has a getter and it returns immutable data, and a separate function to set, in which we copy all the input into our own internal data structure (i.e. copy URIs and addref IPackageValidators into our own collection).

"Rules say that collection-typed properties should not be settable.  Nor should you store internally a reference to someone else's collection.  You don't know the threading model of those collections.  So minimally, it's `CollectionType PropName { get; }`". 

Let the developer build the map the way they want it, then hand it to you, you copy it into your own datastore, and let them observe the result with the read-only view property.

Looks like
```
runtime AddPackageOptions
{
...
void SetValidators(IMap<Uri, IVector<IValidator>>* validators);
IMapView<Uri, IVectorView<IValidator>> Validators { get; };
...
}
```
and example
```
var pdm = PackageDeploymentManager().GetDefault();
var packageUri = "https://contoso.com/package.msix";
var validators = new IList<PackageValidator>();
validators.Add(new PackageFamilyNameValidator("ExpectedFamilyName_1234567890abc"));
validators.Add(new PackageMinimumVersionValidator(new Windows.ApplicationModel.PackageVersion(2, 0, 0, 0)));
validators.Add(new PackageCertificateEkuValidator("1.3.6.1.4.1.311.2.1.11"));
var uriValidators = new Dictionary<Uri, IList<PackageValidator>>();
uriValidators[packageUri] = validators;
var options = new AddPackageOptions();
options.SetValidators(uriValidators);
var deploymentResult = await pdm.AddPackageAsync(packageUri, options);
```

Option #2 allow read/write of the contents (use IMap<X, IVector<Y>>) and snapshot the content (don't just copy the pointer and addref, use something like [How to clone a Windows Runtime map in the face of possible concurrent modification, part 1 - The Ol…](https://devblogs.microsoft.com/oldnewthing/20230719-00/?p=108462)  when  you want to use it.) In this case, same solution but replace SetValidators with a setter property
```
runtime AddPackageOptions
{
...
IMap<Uri, IVector<IValidator>> Validators { get; set; };
...
}
```
and access it carefully in our implementation to avoid issues.

specs/packagemanager/PackageManagement.md
letao-msft: Formatted

specs/packagemanager/PackageManagement.md
letao-msft: Changed to return immutable IMapView, and added a AddPackageValidator method to add one validator at a time.
The pattern of adding one at a time makes sense for validators since that's how they would be set up in real usage (see samples), and doesn't have a problem with concurrency in collections since we're adding one at a time by definition.

specs/packagemanager/PackageManagement.md
```diff
@@ -761,6 +787,136 @@ PackageVersion ToVersion(uint major, uint minor, uint build, uint revision) =>
...
+        var extendedError = deploymentResult.ExtendedError.HResult;
+        var message = deploymentResult.MessageText;
+        Console.WriteLine($"An error occurred while adding the package. Error 0x{error:X08} ExtendedError 0x{extendedError:X08} {message}");
+    }
+}
+```
+
+### 4.9.2. Using custom PackageValidation handlers
+
+This example shows how to implement a custom PackageValidation handler, and use it to verify a package.
```
DrusTheAxe, line 832: NIT: "...implement a custom PackageValidation handler for a .msix, and use it..."

The code below doesn't handle bundles and, while it could, better to keep the sample smaller and more focused.

PR #5865, opened 2025-09-24
Fixing reset activity data on deployment initialization

guimafelipe:
On the method `Initialize` from the deployment manager, the two inside methods `InstallLicences` and `DeployPackages` were resetting the activity data without setting it back, causing loss of data on our telemetry measurements.

### Install licences

Removed the `Reset()` call for every license file found, changing it to a single reset in the beginning of the method. This way, it keeps the information of the install stage but still clearing possible remaining data (like resource id).

### Deploy packages

Added a `Reset()` call in the beginning of the method so it clears previous data (for example, the resource ids that were set from install packages). This is important because if a error happens here, the telemetry data would be confusing as it would still showing a license file.

Set back the install stage after the reset for each package that is to be added.

WARNING:

### DIFF
--- a/dev/Deployment/DeploymentActivityContext.cpp
+++ b/dev/Deployment/DeploymentActivityContext.cpp
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation and Contributors.
+// Copyright (c) Microsoft Corporation and Contributors.
 // Licensed under the MIT License.
 
 #include <pch.h>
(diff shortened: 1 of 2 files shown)


ssparach:
I’ve observed instances where the failure file and line number were not consistently captured in the telemetry as expected. It may be worthwhile to investigate this as part of the telemetry improvements in this area (see `WindowsAppRuntime::Deployment::Activity::Context::SetLastFailure`).

ssparach (approved):

PR #5868, opened 2025-09-28
Changing TestAll.ps1 script to run tests locally with filtering 

guimafelipe:
When working on WinAppSDK, we want to be able to easily run tests to check if our changes are breaking anything or introducing any regressions. 

Now, the `TestAll.ps1` was changed to also work running locally and passing arguments to filter the tests and optionally build them.

WARNING:


tools/RunTests.ps1
```diff
@@ -0,0 +1,194 @@
+<#
+.SYNOPSIS
```
DrusTheAxe: What's wrong with `TestAll.ps1` (in root of the repository)?

If that lacks functionality (e.g. run explicit subset of tests) why not improve `TestAll`?

Or perhaps fork it to `Test.cmd` and `tools\Test.ps1` and change `\TestAll.cmd` to run `Test.cmd ...options for existing TestAll.cmd behavior...`

DrusTheAxe (changes requested):

tools/RunTests.ps1
```diff
@@ -0,0 +1,194 @@
...
+.EXAMPLE
+    .\RunTests.ps1 -TestDef "DeploymentTests"
+
+.EXAMPLE
+    .\RunTests.ps1 -TestDef "Deployment" -Platform "x86" -Configuration "Debug"
+#>
+
+param(
+    [Parameter(Mandatory=$true)]
+    [string]$TestDef,
```
DrusTheAxe: Can this be a filespec or pattern? e.g. `RunTests -TestDef *xyx*.testdef` to run all files found recursively matching *xyz*.testdef

tools/RunTests.ps1
```diff
@@ -0,0 +1,194 @@
...
+
+    [Parameter(Mandatory=$false)]
+    [string]$OutputFolder = "BuildOutput",
+
+    [Parameter(Mandatory=$false)]
+    [string]$Platform = "x64",
```
DrusTheAxe: This should default to the host machine's architecture

SUGGESTION: `[string]$Platform = "$($env:PROCESSOR_ARCHITECTURE)"`

tools/RunTests.ps1
```diff
@@ -0,0 +1,194 @@
...
+
+    [Parameter(Mandatory=$false)]
+    [string]$Configuration = "Release",
```
DrusTheAxe: SUGGESTION: Restrict allowed values e.g.

```
[ValidateSet("Release", "Debug", IgnoreCase=true)]
[string]$Configuration="Release"
```

tools/RunTests.ps1
```diff
@@ -0,0 +1,194 @@
...
+    [Parameter(Mandatory=$true)]
+    [string]$TestDef,
+
+    [Parameter(Mandatory=$false)]
+    [string]$OutputFolder = "BuildOutput",
+
+    [Parameter(Mandatory=$false)]
+    [string]$Platform = "x64",
+
+    [Parameter(Mandatory=$false)]
```
DrusTheAxe: Mandatory's default=false

RECOMMEND: Only specify for required parameters i.e. `[Parameter(Mandatory=$true)]`

tools/RunTests.ps1
DrusTheAxe: Restrict allowed values

SUGGESTION:
```
[ValidateSet("x64", "arm64", "anycpu", IgnoreCase=true)]
[string]$Platform=...
```

tools/RunTests.ps1
```diff
@@ -0,0 +1,194 @@
...
+.EXAMPLE
+    .\RunTests.ps1 -TestDef "Deployment" -Platform "x86" -Configuration "Debug"
+#>
+
+param(
+    [Parameter(Mandatory=$true)]
+    [string]$TestDef,
+
+    [Parameter(Mandatory=$false)]
+    [string]$OutputFolder = "BuildOutput",
```
DrusTheAxe: RECOMMEND: `$Output`

It's a path (folder) by definition. No need to redundantly name the parameter a folder

tools/RunTests.ps1
DrusTheAxe: P.S. Parameter name suggestions:
* TestDef
* Test
* Path
* File

I think `Path` most closely matches common Powershell use

DrusTheAxe (changes requested):
I have more feedback but waiting for the TestAll consolidation

tools/RunTests.ps1
guimafelipe: I used `TestAll.ps1` as base to create this script. The idea for this one is to be a tool for developers to quickly iterate on testing based on what I was doing repetitively. I did not want this to merge with what `TestAll.ps1` is doing as this later one is used on our build pipelines and requires way more stuff (log files and .wprp for example) than what is intended for a quick local test run to check if some changes broke anything.

My usual loop was to call MSBuild locally to build the tests, and then go to `BuildOutput` and run the tests. Now thinking back again, I realize it might be better to also include the test building in this script, as it makes sense we always would want to build and test together.

tools/RunTests.ps1
guimafelipe: I think we can change this to be a pattern. I first went for simplicity following the loop I had with testing locally, and usually I wanted to test everything under a `.testdef` file to check for regressions or tests breaking. But sometimes we want to test specific tests, so a parameter would be good to have and could override the parameters defined in the `.testdef` file.

tools/RunTests.ps1
DrusTheAxe: >I realize it might be better to also include the test building in this script, as it makes sense we always would want to build and test together

I don't know about that. Depends if every test run requires build. I find that's more of a mixed bag. Sometimes I run a test multiple times between changing any code

Also, don't assume dev inner loop doesn't care about log files or .wprp.

Sounds like you're not reinventing TestAll.ps1 so much as BuildAll.ps1

It would be better to enhance those, or perhaps refactor the underlying plumbing from those with lots of options for dev inner-loop use (DoBuild.ps1, DoTest.ps1) and change those to call the new scripts with options to preserve their current behavior

tools/RunTests.ps1
DrusTheAxe: -Path <path> ==> Get-ChildItem -Recurse $path

-Blah <blah> ==> For any .testdef found via -Path only run it if the blah property in the .testdef matches
there's no 'name' property in a testdef but could do this for the meaningful properties as a regex value e.g. 

`-Property:Description "Package.*Test.Add"` would only run this one case in test\packagemanagerapi\test.testdef

```
        {
            "Description": "PackageManagerTests.Add* tests for feature PackageManager (arm64 not currently enabled)",
            "Filename": "PackageManagerTests.dll",
            "Parameters": "/name=Test::PackageManager::Tests::PackageDeploymentManagerTests_Add*",
            "Architectures": ["x86", "x64"],
            "Status": "Enabled"
        },
```

as in `if ($PropertyDescription -match $DescriptionFromParsedTestDefJson)`

DrusTheAxe: BTW take a look at test\packagemanager\api\Test.testdef for an example of a .testdef with multiple Tests. One should be able to easily run a subset of Tests in a .testdef

DrusTheAxe (changes requested):

tools/RunTests.ps1
guimafelipe: I changed the script to receive more parameters and filtering based on the .testdef properties. Also, made the build step optional.

For making the `TestAll.ps1` and the `BuildAll.ps1` scripts to share code with this one, it felt it should come in a separate pull request as this one would be way too much. I would like to hear your opinions on this and if it is better to proceed with this one being a separated change after some clean up.

tools/RunTests.ps1
guimafelipe: I think the current version of the script supports these scenarios.

tools/RunTests.ps1
guimafelipe: I refactored the test code into a `Tests.psm1` module and now `TestAll.ps1` is using it

tools/RunTests.ps1
guimafelipe: Removed the `RunTests.ps1` script completely and refactored `TestAll.ps1`

PR #5871, opened 2025-09-29
Add deployment method to bug report template

lauren-ciha:
This PR updates the bug report template to include a question about the user's app deployment model. With the app deployment model, it will be easier to reproduce issues and track down root causes.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yaml
@@ -52,8 +52,10 @@ body:
       label: Deployment model
       description: Please specify the deployment model you're using.
       options:
-        - label: "Self-contained app (`<WindowsAppSDKSelfContained>true</WindowsAppSDKSelfContained>` is in the project file)"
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug-report.yaml
```diff
@@ -47,6 +47,13 @@ body:
       options:
         - "Packaged (MSIX)"
         - "Unpackaged"
+  - type: dropdown
+    attributes:
+      label: Deployment model
+      description: Please specify the deployment model you're using.
+      options:
+        - label: "Self-contained app (`<WindowsAppSDKSelfContained>true</WindowsAppSDKSelfContained>` is in the project file)"
+        - label: "Framework-dependent app (`<WindowsAppSDKSelfContained>` is not in the project file or it is set to false)"
```
RDMacLachlan, line 57: If the project is referencing a Component package "Microsoft.WindowsAppSDK.AI" the default is then self-contained without the `<selfcontained>` being added. Might need to change your descriptions

DrusTheAxe:
The description field is the auto-generated template. Please provide an actual description for the PR

PR #5883, opened 2025-10-01
Adding script to generate compile commands database

guimafelipe:
When using a text editor or IDE to work with C++, usually it is common to see a big amount of errors in the code even though everything is building correctly locally. This can be annoying, distracting and also slow down productivity.

[Clangd](https://clangd.llvm.org/) is a [Language Server Protocol](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/) for C/C++ that can help solve these problems. It removes the annoying errors (and enable real ones), enables tools like "Go to definition/Go to declaration", enables code completion, and also helps AI to do a better job in the codebase. It works on different code editors like Visual Studio, VS Code, (neo)vim and so on.

For clangd to work, it needs a `compile_commands.json` file to be generated on build, which is a compile commands database. Clangd automatically finds it in the project directory tree and links the files on it. The compile commands database is a JSON file with the commands used by `Cl.exe` for compiling each of the files in the exact way it was done in the successful build process. 

Unfortunately, MSBuild does no generate a `compile_commands.json` file natively at the current moment, which makes it necessary to use an external tool. Our build process uses the `BuildAll.ps1` script, which does various operations before actually calling MSBuild on the solution, which also makes it harder for these tools to infer the compile compile commands database based only on the project files.

A microsoft engineer, facing this same problem, developed the tool [ms2cc](https://github.com/freddiehaddad/ms2cc). This tool receives a `msbuild.log` file and the project root and generates the database.

Unfortunately, for this tool to work it needs to infer for each `Cl.exe` command, the correct absolute path to the files so it can generate the correct output. And `.vcxproj` files in our project uses relative paths, making it impossible to infer the full path from it.

So, I added a preparation script to use the context of the CL task in the binlogs to set the full paths of all the files before passing the list to ms2cc, making it work and generating the correct command compile database.

After that, install clangd on your favorite editor and enjoy all the tools and productivity.

WARNING:


dev/DeploymentAgent/DeploymentAgent.vcxproj
```diff
@@ -97,17 +97,17 @@
     </Link>
   </ItemDefinitionGroup>
   <ItemGroup>
-    <ClInclude Include="tracelogging.h" />
-    <ClInclude Include="pch.h" />
+    <ClInclude Include="$(MSBuildThisFileDirectory)tracelogging.h" />
```
DrusTheAxe: Why the `$(MSBuildThisFileDirectory)` prefix on every header and source?

VS doesn't do this when you add a file in Solution explorer (new, or point at existing).

Aren't these files relative to the project (and typically in same dir)?

DrusTheAxe:
What is a "compile commands database"?

dev/DeploymentAgent/DeploymentAgent.vcxproj
guimafelipe: For developers that work on WinAppSDK, it is common that regardless of the code editor used, there is always a lot of errors that does not help with productivity even if the project builds correctly locally. There is also no consistent Language Server Protocol support (things like "Go to Definition", "See references", etc.).

This PR is from some research and work I put to try to make this all available for WinAppSDK developers.

For this, we can use CLang Driver (https://clangd.llvm.org/), available in most of the editors, to have this support. For clangd to work, it needs  a "compile commands database", which is basically a JSON file with all the commands executed by the compiler for building each of the files in the project, with the include folders needed for each of them, so clangd can do all the correct linking between the files.

Unfortunately, MSBuild does not produce a compile commands database naturally, so we need to use an external tool for that. There are some Visual Studio extensions for that, but they not work because our build system uses powershell scripts moving lots of stuff around before even calling msbuild directly.

In my research, I found the https://github.com/freddiehaddad/ms2cc tool, from another microsoft software engineer, that was attempting to solve the same problem in their project. This tool works getting a msbuild log and converting it into a compile commands database, which is what the powershell script I added is doing.

The problem is: if we don't pass `$(MSBuildThisFileDirectory)` to CLCompile includes on our project files, the compile command will be something like:
`cl.exe ... [includess] /Fo="C:\WindowsAppSKD\obj\..." pch.cpp`

This would not be a problem is there were only one file named pch.cpp in the project. We could search for it in our file tree and know what is the full path. But there are various `pch.cpp` files in WinAppSDK, and there is no way to know only from the compilation path. We could use the object folder for that (which was also being deleted, losing track of some of the include files referenced by the tests), but there is no pattern to follow either.

By including it with `$(MSBuildThisFileDirectory)`, it will have the full path for every `.h` or `.cpp` file compiled in the msbuild logs, which will make the linking possible as there will be no ambiguity of which file is being referenced.

This way, the text editors now have full access to the LSP beneficies from clangd and there will be no more weird errors everywhere even though everything builds correctly.

WinAppSDK was already using `$(MSBuildThisFileDirectory)` in part of the includes, and broadcasted it to the rest of the project and it worked really well for generating the database. I am currently investigating if there is a way to set VS to add it automatically.

DrusTheAxe: My concern is VS is our primary build tool, and adding files to projects is commonly done via VS' Solution Explorer.

I didn't say primary _editor_. Some use VS, I (like many) rarely use VS' editor (I'm a longstanding [TSE](https://semware.com/) fan, others use VScode and other tools). But for _build_ and project + solution management VS is the primary tool.

While command line support is nice, I'm not keen on the friction of requiring `$(MSBuildThisFileDirectory)` prefixed on all files in *proj when VS doesn't do that. I don't see us banning VS, and most devs aren't going to know better (or human error will occasionally forget) to edit *.*proj files after making any edits in VS. This is just going to lead to a constant tax where command line builds regularly break as most devs expand their projects using VS, and not this additional manual editing step.

(Which, BTW, the VS team says "Do not directly edit *.*proj files." Even though many do, or need to do because the GUI lacks some GUI means to make certain edits.

This will create pains and friction for the team. That's the wrong vector - we want LESS friction for the dev team, not more.

TL;DR I'm sympathetic to letting devs use their preferred tools, but not in this way. Not benefiting some developers by imposing a permanent and recurring tax on all future development. The ROI isn't justified.

**Have you talked to the VS team about this?** That seems the better answer -- for VS to create/manage a compiler command database and any related changes needed (e.g. $(MSBuildThisFile)foo.cpp). Then more tools are supported but w/o the added dev tax to do non-VS-standard behavior.

guimafelipe: I also use Visual Studio regularly, which is also a editor. Clangd is also naturally supported in Visual Studio, which is also currently having this same problems as the other editors (showing errors where there is none, not having complete support for C/C++ LSP and so on).

I agree with you that having `$(MSBuildThisFileDirectory)` is a big friction and I am still trying to find alternatives to that. For now was the only thing that make it possible for us to correctly index all the files.

> Have you talked to the VS team about this? That seems the better answer -- for VS to create/manage a compiler command database and any related changes needed (e.g. $(MSBuildThisFile)foo.cpp). Then more tools are supported but w/o the added dev tax to do non-VS-standard behavior.

I didn't talk to them yet. I was expecting that generating a command compile database would be something that already existed by default, but it most of the tools I found related to that are external and didn't work well for our case either. I like the idea of starting a discussion about it with them.

Scottj1s: @guimafelipe I'm also concerned with having to modify every ClCompile and ClInclude.  Note that ClInclude is only an IDE amenity - it has nothing to do with the build. If you're processing the binlog to find the $task cl, you should be able to get the enclosing vcxproj location and infer the source location from that.

For completeness, another approach: cvdump on the pdb to collect the cl.exe command line and all parameters.   But my preference would be groveling the binlog for the vcxproj location.

guimafelipe: @DrusTheAxe Thank you for the feedback, I worked on an alternative via powershell scripts for that and all is working fine now.
@Scottj1s Thank you for the tip on looking for the rest of the logs for context I could use to regenerate the full paths!

BuildAll.ps1
```diff
@@ -165,6 +166,13 @@ Try {
     # PreFastSetup intentionally skips the call to MSBuild.exe below.
     if (($AzureBuildStep -eq "all") -Or ($AzureBuildStep -eq "BuildFoundation"))
     {
+        if ($WindowsAppSDKBuildPipeline -eq 1)
+        {
+            $CleanIntermediateFiles = $true
+        }
+
+        $cleanIntermediateFilesArg = if ($CleanIntermediateFiles) { "/p:WindowsAppSDKCleanIntermediateFiles=true" } else { "" }
+
```
Scottj1s, line 175: or, just:
```suggestion
        if ($WindowsAppSDKBuildPipeline -eq 1)
        {
            $cleanIntermediateFilesArg = "/p:WindowsAppSDKCleanIntermediateFiles=true"
        }
```

Scottj1s: or just this, and have pipelines supply $CleanIntermediateFiles = $true.  
```suggestion
        if ($CleanIntermediateFiles -eq $true)
        {
            $cleanIntermediateFilesArg = "/p:WindowsAppSDKCleanIntermediateFiles=true"
        }
```

Scottj1s:
A general comment - most developers don't invoke special scripts like BuildAll.ps1.  My own preference is to use the standard build tools: VS, MSBuild, dotnet - especially for inner loop.  You might consider refactoring this into a custom MSBuild target with AfterTargets="ClCompile" to hook into every C++ compilation.  Then you'd have direct access to the cl.exe args and project location and could write/append to a compile_commands.json directly.  The advantage is that you'd always have an up to date compile commands DB, regardless of your IDE or build tool.  You can implement a custom target with any arbitrary commands, including PS1 scripts, although I prefer to use custom inline C# tasks to enable debugging.

BuildAll.ps1
guimafelipe: I implemented this change in all the calls of `BuildAll.ps1`in the pipeline. Let me know what you think about it. Maybe the standard can be for the variable to be true so we don't need to change all these calls.

tools/GenerateCompilationDatabase.ps1
```diff
@@ -0,0 +1,166 @@
+Param(
```
DrusTheAxe, line 34: Missing line 1 Microsoft Copyright and MIT License

DrusTheAxe, line 34: Missing docs

```
<#
.SYNOPSIS
   blah

.DESCRIPTION
    yakkety yak

.PARAMETER foo
   The foo to yadda yadda

.EXAMPLE
    blah blah
#>
```

See top of tools\devcheck\DevCheck.ps1 for a reference

tools/GenerateCompilationDatabase.ps1
```diff
@@ -0,0 +1,166 @@
...
+    [string]$Ms2ccVersion = "1.3.0",
+    [switch]$Update
+
+)
+
+$binlogFileBase = Join-Path (Split-Path $PSScriptRoot -parent) "BuildOutput\Binlogs"
+$binlogFiles = Get-ChildItem -Path $binlogFileBase -Filter *.binlog -Recurse | Select-Object -ExpandProperty FullName
+
+Write-Host "Binlog files found:"
+$binlogFiles | ForEach-Object { Write-Host $_ }
```
DrusTheAxe: SUGGESTION: Indent the files e.g,

`$binlogFiles | ForEach-Object { Write-Host "    $_" }`

tools/GenerateCompilationDatabase.ps1
```diff
@@ -0,0 +1,166 @@
...
+
+$VCToolsInstallDir = . "${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer\vswhere.exe" -Latest -prerelease -requires Microsoft.Component.MSBuild -property InstallationPath
+write-host "VCToolsInstallDir: $VCToolsInstallDir"
```
DrusTheAxe: NIT: `Write-Host`

PS is case-insensitive but inconsistency is irksome. Typically we use MixedCase for powershell commands but regardless, please be consistent

same applies throughout

DrusTheAxe: Are you dumping variables or state/resources?

The latter is generally preferred. Doesn't matter to uses of the script if you name it `$msBuildPath`, `$msbuildexe`, `$msbuild_exe`, `$exe` or `$Fred`. But it can be relevant to the user the EXE you're going to execute

DrusTheAxe: Is this meaningful to users of the script?

Such things can be useful for debugging scenarios but not so much for common usage.

SUGGESTION: `Write-Verbose` instead of `Write-Host`. Then folks can see this added info when they want if they specify the `-Verbose` command line option

tools/GenerateCompilationDatabase.ps1
```diff
@@ -0,0 +1,166 @@
...
+
+$msBuildPath = "$VCToolsInstallDir\MSBuild\Current\Bin\msbuild.exe"
+write-host "msBuildPath: $msBuildPath"
```
DrusTheAxe: SUGGESTION: `Write-Verbose`

tools/GenerateCompilationDatabase.ps1
```diff
@@ -0,0 +1,166 @@
...
+
+Remove-Item "temp-filtered.log" -ErrorAction SilentlyContinue
+Remove-Item "temp-filtered2.log" -ErrorAction SilentlyContinue
+
+# Target "ClCompile" in file "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Microsoft\VC\v170\Microsoft.CppCommon.targets" from project "C:\WindowsAppSDK\dev\Detours\Detours.vcxproj" (target "_ClCompile" depends on it):
```
DrusTheAxe: I don't understand the comment on lines 22-23. Dead code to be removed?

tools/GenerateCompilationDatabase.ps1
```diff
@@ -0,0 +1,166 @@
...
+Write-Host "Binlog files found:"
+$binlogFiles | ForEach-Object { Write-Host $_ }
+
+$VCToolsInstallDir = . "${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer\vswhere.exe" -Latest -prerelease -requires Microsoft.Component.MSBuild -property InstallationPath
+write-host "VCToolsInstallDir: $VCToolsInstallDir"
+
+$msBuildPath = "$VCToolsInstallDir\MSBuild\Current\Bin\msbuild.exe"
+write-host "msBuildPath: $msBuildPath"
+
+Remove-Item "temp-filtered.log" -ErrorAction SilentlyContinue
```
DrusTheAxe: A few things:
1. "temp-filtered.log" will be removed from the current directory. Better to use an explicit temp location
  * 2nd best answer = `$filteredLog = Join-Path ${env:Temp} "GenerateCompilationDatabase-filtered.log"`
  * Best answer =
```
$root = Get-ProjectRoot
$temp = Join-Path $root 'temp'
$filteredLog = Join-Path $temp 'GenerateCompilationDatabase-filtered.log'
```
2. `-Force` option to help ensure the file can be removed
3. Why SilentlyContinue? If the file can't be removed won't it cause problems later when trying to write (or worse, append) to it?
   3.1. If you really want to best-effort the file deletion you may want `Ignore`. Like `SilentlyContinue` but doesn't record the error in `$Error`. But I suspect better to `-Force` and let it fail if the file can't be deleted
4. If you need to specify an error action the short form's usually preferred e.g. `... -EA:SilentlyContinue`
4. Remove-Item can emit output. Unless that's really desired you probably want `$null = Remove-Item...`

tools/GenerateCompilationDatabase.ps1
```diff
@@ -0,0 +1,166 @@
...
+Remove-Item "temp-filtered2.log" -ErrorAction SilentlyContinue
+
+# Target "ClCompile" in file "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Microsoft\VC\v170\Microsoft.CppCommon.targets" from project "C:\WindowsAppSDK\dev\Detours\Detours.vcxproj" (target "_ClCompile" depends on it):
+# Context here would be: "C:\WindowsAppSDK\dev\Detours"
+
+foreach ($binlogFile in $binlogFiles) {
```
DrusTheAxe: NIT: Open brace on new line. Same as Windows C++ and C# Coding Guidelines

```
foreach ($binlogFile in $binglogFiles)
{
    if (-Not (Test-Path $binlogFile))
    {
        ...
    }
}
```

tools/GenerateCompilationDatabase.ps1
```diff
@@ -0,0 +1,166 @@
...
+    if (-Not (Test-Path $binlogFile)) {
+        Write-Error "Binlog file not found: $binlogFile"
+        exit 1
+    }
+
+    & $msBuildPath $binlogFile /v:normal /noconlog /flp:logfile=temp.log
```
DrusTheAxe: Where's that writing temp.log? If current directory please switch to `Join-Path (Get-ProjectRoot) "temp"` for the per-repo temp location (created by DevCheck)

tools/GenerateCompilationDatabase.ps1
```diff
@@ -0,0 +1,166 @@
...
+        ForEach-Object { $_.Line } |
+        Where-Object { $_ -notmatch "Tracker.exe" } |
+        Out-File -FilePath "temp-filtered.log" -Append -Encoding utf8
+
+    Remove-Item "temp.log"
+
+    Write-Host "Processed binlog file: $binlogFile"
+}
+
+Write-Host "Filtered log file generated at: $(Get-Location)\temp-filtered.log"
```
DrusTheAxe: `Write-Verbose` ?

tools/GenerateCompilationDatabase.ps1
```diff
@@ -0,0 +1,166 @@
...
+    & $msBuildPath $binlogFile /v:normal /noconlog /flp:logfile=temp.log
+
+    Select-String -Path "temp.log" -Pattern "Target ""ClCompile"" in file","Cl.exe"  |
+        ForEach-Object { $_.Line } |
+        Where-Object { $_ -notmatch "Tracker.exe" } |
+        Out-File -FilePath "temp-filtered.log" -Append -Encoding utf8
+
+    Remove-Item "temp.log"
+
+    Write-Host "Processed binlog file: $binlogFile"
```
DrusTheAxe: `Write-Verbose` ?

tools/GenerateCompilationDatabase.ps1
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    & $msBuildPath $binlogFile /v:normal /noconlog /flp:logfile=temp.log
+
+    Select-String -Path "temp.log" -Pattern "Target ""ClCompile"" in file","Cl.exe"  |
+        ForEach-Object { $_.Line } |
+        Where-Object { $_ -notmatch "Tracker.exe" } |
+        Out-File -FilePath "temp-filtered.log" -Append -Encoding utf8
+
+    Remove-Item "temp.log"
```
DrusTheAxe: See previous Remove-Item comment

tools/GenerateCompilationDatabase.ps1
```diff
@@ -0,0 +1,166 @@
...
+
+# for each line in temp-filtered.log, if it is a "Target " line, extract the project directory and save in the contextPath variable
+# if it is a "Cl.exe" line, prepend the contextPath to every source file in that line only if it is
+# a relative path, then output the modified line to temp-filtered-2.log
+# The source files are the arguments that ends with .cpp, .c, or .h
+# They can have spaces in them. If so, they are enclosed in quotes.
+
+$lines = Get-Content "temp-filtered.log"
+
+Write-Host "Processing filtered log to adjust file paths..."
```
DrusTheAxe: `Write-Verbose` ?

tools/GenerateCompilationDatabase.ps1
```diff
@@ -0,0 +1,166 @@
...
+
+foreach ($line in $lines) {
+    if ($line -match 'Target "ClCompile" in file "([^"]+)" from project "([^"]+)"') {
+        $contextPath = Split-Path $matches[2] -parent
+        Write-Host "Context path set to: $contextPath"
+    } elseif ($line -match 'Cl\.exe (.+)$') {
```
DrusTheAxe: Are you looking for a line _containing_ `cl.exe ...anything...<end-of-line>`?
Odd to see looking for that rather than '^.*cl\.exe (.+)$' to explicitly match the whole line and not simply ends-with

Also, looking for 'cl.exe .+$' is somewhat risky as it match "c:\foo\cl.exe blahblah" as well as "c:\foo\oracl.exe blahblah" and other false positives. Unlikely, but paranoia's not a bad thing :P Better if you can specify something more explicit e.g. `"^.+\\cl\.exe (.+)$"`

tools/GenerateCompilationDatabase.ps1
```diff
@@ -0,0 +1,166 @@
...
+        $clLine = $matches[1]
+        $args = $clLine -split ' (?=(?:[^"]*"[^"]*")*[^"]*$)'
```
DrusTheAxe: Why not bake this into the regex above?

tools/GenerateCompilationDatabase.ps1
```diff
@@ -0,0 +1,166 @@
...
+
+        $modifiedArgs = @()
+        foreach ($arg in $args) {
+            if ($arg -match '^(.*\.(cpp|c|h))$' -or $arg -match '^"(.*\.(cpp|c|h))"$') {
```
DrusTheAxe: no `.hpp`?

(`.cxx` and `.hxx` are older school, winappsdk has none and we're unlikely to at this point in the industry)

tools/GenerateCompilationDatabase.ps1
```diff
@@ -0,0 +1,166 @@
...
+}
+
+Write-Host "Adjusted file paths and generated: $(Get-Location)\temp-filtered2.log"
+
+$ms2ccPath = Join-Path $PSScriptRoot "ms2cc"
+$ms2ccExe = Join-Path $ms2ccPath "ms2cc.exe"
+
+if (-Not (Test-Path $ms2ccExe)) {
+    Write-Host "Downloading ms2cc..."
+    $ms2ccUrl = "https://github.com/freddiehaddad/ms2cc/releases/download/v$Ms2ccVersion/ms2cc-$Ms2ccVersion.zip"
```
DrusTheAxe: Pulling in external code is a potential red flag, and external binaries exe more so

Perhaps add a parameter for the user to provide an explicit path to `ms2cc.exe` and another to optionally download e.g. line ~4
```
...
.PARAMETER MS2CCEXE
    Path to ms2cc.exe

.PARAMETER DownloadMS2CCEXE
    Download ms2cc.exe if necessary

.PARAMETER MS2CCEXEURL
   URL to download ms2cc ms2cc.exe (if necessary and enabled via -DownloadMS2CCEXE)
...
[string]$MS2CCEXE,
[switch]$DownloadMS2CCEXE,
[string]$MS2CCEXEURL='https://github.com/freddiehaddad/ms2cc/releases/download/v1.3.0/ms2cc-1.3.0.zip',
```

This way it can work offline using a defined exe, so even constrained and network-less environments can do so, with optional opt-in (explicit permission) to download if necessary and where to get it from (for when an explicit location's needed, 'cause, secure supply chain etc).

tools/GenerateCompilationDatabase.ps1
```diff
@@ -0,0 +1,166 @@
...
+    $zipPath = Join-Path $PSScriptRoot "ms2cc-$Ms2ccVersion.zip"
```
DrusTheAxe: Use the project's temp location, not the script's location (ie temp, not a source directory) e.g.

```
$root = Get-ProjectRoot
$temp = Join-Path $root 'temp'
$zipPath = Join-Path $temp "ms2cc-$($Ms2ccVersion).zip"
```

or if you dislike legible code :P
```
Join-Path (Join-Path (Get-ProjectRoot) 'temp') "ms2cc-$($Ms2ccVersion).zip"
```

And likewise extract to a dir under the project root's `temp`, or `.user`. 

FYI DevCheck downloads `nuget.exe` if necessary under `.user`. See `tools\devcheck\DevCheck.ps1` functions `Install-NugetExe` and `Test-NugetExe` plus parameters `-NugetExe`, `-NugetMinVersion` and `-NugetExeUpdate` for ideas and inspiration and steal shamelessly from :P

tools/GenerateCompilationDatabase.ps1
```diff
@@ -0,0 +1,166 @@
...
+
+    try {
+        Invoke-WebRequest -Uri $ms2ccUrl -OutFile $zipPath -UseBasicParsing
```
DrusTheAxe: `DevCheck` uses `curl.exe` with -L (follow redirects) and -# (show progress)

How's Invoke-WebRequest compare?

tools/GenerateCompilationDatabase.ps1
```diff
@@ -0,0 +1,166 @@
...
+
+        if (-Not (Test-Path $ms2ccPath)) {
+            New-Item -ItemType Directory -Path $ms2ccPath -Force | Out-Null
+        }
+
+        Expand-Archive -Path $zipPath -DestinationPath $ms2ccPath -Force
+        Write-Host "Extracted ms2cc to: $ms2ccPath"
+
+        Remove-Item $zipPath -Force
+        Write-Host "Cleaned up zip file"
```
DrusTheAxe: `Write-Verbose`

tools/GenerateCompilationDatabase.ps1
```diff
@@ -0,0 +1,166 @@
...
+
+$backupPath = $null
+
+# If the -Update flag is provided, we want to update an existing compile_commands.json
+# instead of overwriting it.
+if ($Update) {
+    # Save the old compile_commands.json if it exists in a new file
+    $compileCommandsPath = Join-Path (Split-Path $PSScriptRoot -parent) "compile_commands.json"
+    if (Test-Path $compileCommandsPath) {
+        $backupPath = Join-Path (Split-Path $PSScriptRoot -parent) "compile_commands_old.json"
```
DrusTheAxe: Don't write non-source to source directories (`tools` is a source directory)

Write to the project's 'temp' directory. See earlier comments re: `Join-Path $root 'temp'`

tools/GenerateCompilationDatabase.ps1
```diff
@@ -0,0 +1,166 @@
...
+        Copy-Item -Path $compileCommandsPath -Destination $backupPath
+        Write-Host "Backed up existing compile_commands.json to: $backupPath"
+    } else {
+        Write-Host "No existing compile_commands.json found to back up."
+    }
+}
+
+& $ms2ccExe -i "temp-filtered2.log" -d (Split-Path $PSScriptRoot -parent) -p
```
DrusTheAxe: See earlier comment re project's temp dir and

`Join-Path $root 'temp'`

DrusTheAxe (changes requested):

tools/GenerateCompilationDatabase.ps1
DrusTheAxe: General rule of thumb: When done, `git status` should report no new files or directories (and not because they're listed in .gitignore :P )

tools/GenerateCompilationDatabase.ps1
guimafelipe: It was a slight explanation on the type of lines I was looking to filter there with an example. I will move it to a better location.

tools/GenerateCompilationDatabase.ps1
guimafelipe: Yes, anything containing `cl.exe` works. This is just to reduce drastically the amount of lines ms2cc needs to process. If there is an `oracl.exe` there remaining, it is ok because `ms2cc` will do a better job parsing it. I could pass the entire log file to it with no problem, but I needed to merge all the log files into one so I took chance to also filter to only the important lines while doing it.

tools/GenerateCompilationDatabase.ps1
guimafelipe: Thanks for this tip. This was copied from `BuildAll.ps1` script. I will add a note to fix there too for consistency

tools/GenerateCompilationDatabase.ps1
guimafelipe: This was just for organization/reading purposes. I can move it up if it is better.

tools/GenerateCompilationDatabase.ps1
guimafelipe: This is in the case the file does not exist. The file is not expected to exist unless the script execution was stopped in the middle of the whole process. I will change it to "-Force" but I don't want it to stop in the case of the file not existing. I will change it to test the path before attempting to remove it.

tools/GenerateCompilationDatabase.ps1
guimafelipe: Invoke-WebRequest was the default I found for downloading something with powershell. Looking at the documentation, it does not seems to have the same capabilities as the ones you listed for curl.exe. Do you think it is worth changing it to use curl?

tools/GenerateCompilationDatabase.ps1
guimafelipe: This was the same pattern of `BuildAll.ps1`. I am writing the same information. I will change it to verbose only.

guimafelipe:
> A general comment - most developers don't invoke special scripts like BuildAll.ps1. My own preference is to use the standard build tools: VS, MSBuild, dotnet - especially for inner loop. You might consider refactoring this into a custom MSBuild target with AfterTargets="ClCompile" to hook into every C++ compilation. Then you'd have direct access to the cl.exe args and project location and could write/append to a compile_commands.json directly. The advantage is that you'd always have an up to date compile commands DB, regardless of your IDE or build tool. You can implement a custom target with any arbitrary commands, including PS1 scripts, although I prefer to use custom inline C# tasks to enable debugging.

I would love to have it as a MSBuild target in a future PR. We can try having this initially as a script for our team to try it out and get feedback before committing more to integrate it with the automatic build infrastructure.

In the ideal world, I would want to this to just be a switch we can toggle when calling MSBuild.

DrusTheAxe:
Powershell script as v1 stalking horse to work out the mechanics before converting to v2 MSBuild target is a good approach

> In the ideal world, I would want to this to just be a switch we can toggle when calling MSBuild.

Build logic and opt-in is the right goal

```
<PropertyGroup Condition="'$(WindowsAppSdkBuildCompilationDatabase)' == '' and "$(WINDOWSAPPSDK_BUILD_COMPILATION_DATABASE)' == '1'">
    <WindowsAppSdkBuildCompilationDatabase>true</WindowsAppSdkBuildCompilationDatabase>
</PropertyGroup>
```

Then devs who want it (regardless whether using msbuild or devenv) can
```
SET WINDOWSAPPSDK_BUILD_COMPILATION_DATABASE=1
msbuild / devenv windowsappruntime.sln
```

Set that envvar in your user profile and you get it, and devs who don't don't.

PR #5885, opened 2025-10-02
MSIX FindPackageFile*() API Spec

DrusTheAxe:
API spec for FindPackageFile*() APIs

https://task.ms/59424502

### DIFF
--- a/specs/package/Package.md
+++ b/specs/package/Package.md
@@ -120,7 +120,7 @@ The search order for a package's locations is:
 
 ### 3.1.2. Get file path for a package
 
-`Package.GetFilePath([packageFullName, ]filename[, options])` searches the location(s)
+`Package.GetFilePath(filename[packageFullName, [, options]])` searches the location(s)
 of a package for the specified file, per the algorithm:
 
 ```
@@ -165,7 +165,7 @@ package graph for the specified file, per the algorithm:
 
 ```
 FOREACH pkg IN GetPackageGraph():
-    file = Package.GetFilePath(pkg.fullname, filename)
+    file = Package.GetFilePath(filename, pkg.fullname)
     IF file != null
         RETURN file
 NEXT
@@ -233,7 +233,7 @@ using Microsoft.Windows.ApplicationModel;
 
 string GetResourcesPri(string packageFullName)
 {
-    var absoluteFilename = Package.GetFilePath(packageFullName, "resources.pri");
+    var absoluteFilename = Package.GetFilePath("resources.pri", packageFullName);
     if (absoluteFilename == null)
     {
         Console.WriteLine($"ERROR: resources.pri not found for {packageFullName}");
@@ -281,7 +281,7 @@ string GetResourcesPri(string packageFullName)
     var options = GetPackageFilePathOptions.SearchInstallPath |
                   GetPackageFilePathOptions.SearchMachineExternalPath |
                   GetPackageFilePathOptions.SearchUserExternalPath;
-    var absoluteFilename = Package.GetFilePath(packageFullName, "resources.pri", options);
+    var absoluteFilename = Package.GetFilePath("resources.pri", packageFullName, options);
     if (absoluteFilename == null)
     {
         Console.WriteLine($"ERROR: resources.pri not found for {packageFullName}");
@@ -362,7 +362,8 @@ std::wstring GetXamlWinMD()
 
 ## 4.5. Get a file path in the current process' package graph with options
 
-Locate `Microsoft.UI.Xaml.winmd` in the current process' package graph but ignore Mutable locations.
+Locate `Microsoft.UI.Xaml.winmd` in the current process' package graph but ignore Mutable locations,
+Resource packages and HostRuntime dependencies.
 
 ### 4.5.1. C# Example
 
@@ -373,7 +374,12 @@ string GetXamlWinMD()
 {
     var options = GetPackageFilePathOptions.SearchInstallPath |
                   GetPackageFilePathOptions.SearchMachineExternalPath |
-                  GetPackageFilePathOptions.SearchUserExternalPath;
+                  GetPackageFilePathOptions.SearchUserExternalPath |
+                  GetPackageFilePathOptions.SearchMainPackages |
+                  GetPackageFilePathOptions.SearchFrameworkPath |
+                  GetPackageFilePathOptions.SearchOptionalPath |
+                  GetPackageFilePathOptions.SearchStaticDependencies |
+                  GetPackageFilePathOptions.SearchDynamicDependencies;
     var absoluteFilename = PackageGraph.GetFilePath("Microsoft.UI.Xaml.winmd", options);
     if (absoluteFilename == null)
     {
@@ -390,8 +396,13 @@ string GetXamlWinMD()
 std::wstring GetXamlWinMD()
 {
     GetPackageFilePathOptions options{ GetPackageFilePathOptions_SearchInstallPath |
-                                    GetPackageFilePathOptions_SearchMachineExternalPath |
-                                    GetPackageFilePathOptions_SearchUserExternalPath };
+                                       GetPackageFilePathOptions_SearchMachineExternalPath |
+                                       GetPackageFilePathOptions_SearchUserExternalPath |
+                                       GetPackageFilePathOptions_SearchMainPackages |
+                                       GetPackageFilePathOptions_SearchFrameworkPath |
+                                       GetPackageFilePathOptions_SearchOptionalPath |
+                                       GetPackageFilePathOptions_SearchStaticDependencies |
+                                       GetPackageFilePathOptions_SearchDynamicDependencies };
     wil::unique_process_heap_string absoluteFilename;
     const HRESULT hr{ GetPackageFilePathInPackageGraph(
         L"Microsoft.UI.Xaml.winmd", options, wistd::out_param(absoluteFilename)) };
@@ -446,6 +457,27 @@ namespace Microsoft.Windows.ApplicationModel
 
         /// Include the package's User External path (if it has one) in the file search order
         SearchUserExternalPath = 0x0008,
+
+        /// Include Main packages in the file search order
+        SearchMainPackages = 0x0010,
+
+        /// Include Framework packages in the file search order
+        SearchFrameworkPackages = 0x0020,
+
+        /// Include Optional packages in the file search order
+        SearchOptionalPackages = 0x0040,
+
+        /// Include Resource packages in the file search order
+        SearchResourcePackages = 0x0080,
+
+        /// Include HostRuntime dependencies in the file search order
+        SearchHostRuntimeDependencies = 0x0100,
+
+        /// Include Static package dependencies in the file search order
+        SearchStaticDependencies = 0x0200,
+
+        /// Include Dynamic package dependencies in the file search order
+        SearchDynamicDependencies = 0x0400,
     }
 
     [contract(PackageRuntimeContract, 1)]
@@ -463,19 +495,19 @@ namespace Microsoft.Windows.ApplicationModel
         static String GetFilePath(String filename);
 
         /// Return the absolute path to the file in the package.
-        /// @param packageFullName the package.
         /// @param filename file to locate.
+        /// @param packageFullName the package.
         /// @param packageFile absolute path to the packaged file, or empty string ("") if not found.
         /// @note The package path search order is External(User or Machine) -> Mutable -> Install.
         /// @note If a package has a UserExternal location then MachineExternal location is not checked (even if the package has one).
         /// @see https://learn.microsoft.com/en-us/windows/win32/api/appmodel/nf-appmodel-getpackagepathbyfullname2
         /// @see PackageGraph.GetFilePath
         [method_name("GetFilePathInPackage")]
-        static String GetFilePath(String packageFullName, String filename);
+        static String GetFilePath(String filename, String packageFullName);
 
         /// Return the absolute path to the file in the package.
-        /// @param packageFullName the package.
         /// @param filename file to locate.
+        /// @param packageFullName the package.
         /// @param packageFile absolute path to the packaged file, or empty string ("") if not found.
         /// @param options options for the search.
         /// @note The package path search order is External(User or Machine) -> Mutable -> Install.
@@ -483,8 +515,8 @@ namespace Microsoft.Windows.ApplicationModel
         /// @see https://learn.microsoft.com/en-us/windows/win32/api/appmodel/nf-appmodel-getpackagepathbyfullname2
         /// @see PackageGraph.GetFilePath
         /// @see PackageGraph.GetFilePathOptions
-        [method_name("GetFilePathWithOptions")]
-        static String GetFilePath(String packageFullName, String filename, GetFilePathOptions options);
+        [method_name("GetFilePathInPackageWithOptions")]
+        static String GetFilePath(String filename, String packageFullName, GetFilePathOptions options);
     }
 
     [contract(PackageRuntimeContract, 1)]
@@ -542,6 +574,27 @@ typedef enum GetPackageFilePathOptions
 
     /// Include the package's User External path (if it has one) in the file search order
     GetPackageFilePathOptions_SearchUserExternalPath = 0x0008,
+
+    /// Include Main packages in the file search order
+    GetPackageFilePathOptions_SearchMainPackages = 0x0010,
+
+    /// Include Framework packages in the file search order
+    GetPackageFilePathOptions_SearchFrameworkPackages = 0x0020,
+
+    /// Include Optional packages in the file search order
+    GetPackageFilePathOptions_SearchOptionalPackages = 0x0040,
+
+    /// Include Resource packages in the file search order
+    GetPackageFilePathOptions_SearchResourcePackages = 0x0080,
+
+    /// Include HostRuntime dependencies in the file search order
+    GetPackageFilePathOptions_SearchHostRuntimeDependencies = 0x0100,
+
+    /// Include Static package dependencies in the file search order
+    GetPackageFilePathOptions_SearchStaticDependencies = 0x0200,
+
+    /// Include Dynamic package dependencies in the file search order
+    GetPackageFilePathOptions_SearchDynamicDependencies = 0x0400,
 }
 
 /// Return the absolute path to the file in the package.


specs/package/Package.md
```diff
@@ -0,0 +1,310 @@
...
+developers using MSIX to do the legwork, and doing so correctly, consistently and efficiently. These
+new APIs provide a consistent, correct and easy to use solution.
+
+### 3.1.1. Package Location Search Order
+
+The search order for a package's locations is:
+
+1. External (User or Machine)
+2. Mutable
+3. Install
```
nieubank, line 116: It would be good if this behavior could be specified. For example, I may explicitly want to skip Mutable or External location if I absolutely wanted the one shipped with the package.

DrusTheAxe: If you know it's only in 1 location then existing APIs trivially cover that -- W.AM.Package.*Path properties, GetPackagePath*(...PackagePathType_*...), etc to ge the path and assume the file exists in it, or check if you're unsure using standard mechanisms.

I was waiting on subset support to hear if someone squeaked so, congratulations, we have a winner :P Adding...

DrusTheAxe: Added FindPackageFileOptions and overloads

specs/package/Package.md
```diff
@@ -0,0 +1,505 @@
...
+// Not Found
+RETURN null
+```
+
+Overloads without the `options` parameter use the default (search everything).
+
+
+### 3.1.3. FindPackageFileInPackageGraph
+
+`FindPackageFileInPackageGraph(filename[, options])` searches the location(s) of each package in the caller's
```
jonwis: Is it necessary to provide this wrapper, vs. "a really great sample" using the primitives?

specs/package/Package.md
```diff
@@ -0,0 +1,505 @@
...
+Locate `resource.pri` in the package `Contoso_1.2.3.4_x64__1234567890abc`.
+
+### 4.1.1. C# Example
+
+```c#
+using Microsoft.Windows.ApplicationModel;
+
+string GetResourcesPri(string packageFullName)
+{
+    var absoluteFilename = Package.FindPackageFileInPackage(packageFullName, "resources.pri");
```
jonwis: That's a lot of "package".  Consider `Package.GetFilePath(packageFullName, "resources.pri")` - "Find" verbs typically return a collection, or use a pattern. This is "I know the name I want, get it." .... `Get-` methods are allowed to return null meaning "nothing matched that ID."

specs/package/Package.md
```diff
@@ -0,0 +1,505 @@
...
+    return absoluteFilename;
+}
+```
+
+### 4.1.2. C++ Example
+
+```c++
+wil::unique_process_heap_string GetResourcesPri(PCWSTR packageFullName)
+{
+    FindPackageFileOptions options{};
```
jonwis: Should we only have the "flat C" exports, and let people use cswin32 / node-native / etc to invoke the API from their environment? Seems like 2x the code currently...

specs/package/Package.md
```diff
@@ -0,0 +1,505 @@
...
+    wil::unique_process_heap_string absoluteFilename;
+    const HRESULT hr{ LOG_IF_FAILED(FindPackageFileInPackage(
+        packageFullName, L"resources.pri", options, wistd::out_param(absoluteFilename))) };
```
jonwis: Just say:

```c++
HRESULT hr { GetPackageFilePath(fullName, ...) };
if (FAILED_LOG(hr)) {
    // ...
}
```

specs/package/Package.md
```diff
@@ -0,0 +1,505 @@
...
+```c++
+wil::unique_process_heap_string GetResourcesPri(PCWSTR packageFullName)
+{
+    FindPackageFileOptions options{ FindPackageFileOptions_SearchInstallPath |
+                                    FindPackageFileOptions_SearchMachineExternalPath |
+                                    FindPackageFileOptions_SearchUserExternalPath };
+    wil::unique_process_heap_string absoluteFilename;
+    const HRESULT hr{ LOG_IF_FAILED(FindPackageFileInPackage(
+        packageFullName, L"resources.pri", options, wistd::out_param(absoluteFilename))) };
+    if (FAILED(hr))
```
jonwis: See note earlier on use of WIL

specs/package/Package.md
```diff
@@ -0,0 +1,505 @@
...
+Locate `resource.pri` in the current process' package graph.
+
+### 4.3.1. C# Example
+
+```c#
+using Microsoft.Windows.ApplicationModel;
+
+string GetXamlWinMD()
+{
+    var absoluteFilename = Package.FindPackageFileInPackageGraph("Microsoft.UI.Xaml.winmd");
```
jonwis: How is this API different from SearchPath?  Doesn't SearchPath use the current package graph too?

specs/package/Package.md
```diff
@@ -0,0 +1,505 @@
...
+
+```c++
+wil::unique_process_heap_string GetXamlWinMD()
+{
+    FindPackageFileOptions options{ FindPackageFileOptions_SearchInstallPath |
+                                    FindPackageFileOptions_SearchMachineExternalPath |
+                                    FindPackageFileOptions_SearchUserExternalPath };
+    wil::unique_process_heap_string absoluteFilename;
+    const HRESULT hr{ LOG_IF_FAILED(FindPackageFileInPackageGraph(
+        L"Microsoft.UI.Xaml.winmd", options, wistd::out_param(absoluteFilename))) };
```
jonwis: See note on WIL usage (assume that applies everywhere.)

specs/package/Package.md
```diff
@@ -0,0 +1,505 @@
...
+    }
+    return absoluteFilename;
+}
+```
+
+# 5. Remarks
+
+## 5.1. Platform Support
+
+This API is available on Windows >= 10.0.17763.0 (aka RS5).
```
jonwis, line 427: Yes, but is it functional?

specs/package/Package.md
```diff
@@ -0,0 +1,505 @@
...
+    {
+        /// Return the absolute path to the file in the package.
+        /// @param packageFullName the package, or empty string ("") to use the current process' package identity.
+        /// @param filename file to locate.
+        /// @param packageFile absolute path to the packaged file, or empty string ("") if not found.
+        /// @note The package path search order is External(User or Machine) -> Mutable -> Install.
+        /// @note If a package has a UserExternal location then MachineExternal location is not checked (even if the package has one).
+        /// @see https://learn.microsoft.com/en-us/windows/win32/api/appmodel/nf-appmodel-getpackagepathbyfullname2
+        /// @see FindPackageFileInPackageGraph
+        static String FindPackageFileInPackage(String packageFullName, String filename);
```
jonwis: So these two calls are basically the same?

```c++
auto f = Package.FindPackageFileInPackage("", "foo.winmd");
auto f = Package.FindPackageFileInPackageGraph("foo.winmd");
```

specs/package/Package.md
```diff
@@ -0,0 +1,505 @@
...
+        static String FindPackageFileInPackage(String packageFullName, String filename, FindPackageFileOptions options);
+
+        /// Return the absolute path to the file in the package graph.
+        /// @param filename file to locate.
+        /// @param packageFile absolute path to the packaged file, or empty string ("") if not found.
+        /// @note The package paths search order is External(User or Machine) -> Mutable -> Install.
+        /// @note If a package has a UserExternal location then MachineExternal location is not checked (even if the package has one).
+        /// @see https://learn.microsoft.com/en-us/windows/win32/api/appmodel/nf-appmodel-getpackagepathbyfullname2
+        /// @see FindPackageFileInPackage
+        static String FindPackageFileInPackageGraph(String filename);
```
jonwis: This is on the Package object, maybe it should be on some kind of PackageGraph object instead? The answer is for "the running instance of the package" ... I can't ask this of a non-running package? Or can I use `Package.FindPackageFileInPackage(otherPackageId, filename)` and have it tell me the path of the file _if the other package was running?_

specs/package/Package.md
```diff
@@ -0,0 +1,505 @@
...
+/// @note The package path search order is External(User or Machine) -> Mutable -> Install.
+/// @note If a package has a UserExternal location then MachineExternal location is not checked (even if the package has one).
+/// @see https://learn.microsoft.com/en-us/windows/win32/api/appmodel/nf-appmodel-getpackagepathbyfullname2
+/// @see FindPackageFileInPackageGraph()
+/// @see FindPackageFileOptions
+STDAPI FindPackageFileInPackage(
+    PCWSTR packageFullName,
+    _In_ PCWSTR filename,
+    _In_ FindPackageFileOptions options,
+    _Outptr_result_maybenull_ PWSTR* packageFile) noexcept;
```
jonwis, line 614: Missing `(`

specs/package/Package.md
```diff
@@ -0,0 +1,505 @@
...
+/// @param packageFile absolute path to the packaged file, or NULL if not found. Allocated via HeapAlloc; use HeapFree to deallocate
+/// @note The search order is External(User or Machine) -> Mutable -> Install for each package in the package graph.
+/// @note If a package has a UserExternal location then MachineExternal location is not checked (even if the package has one).
+/// @see https://learn.microsoft.com/en-us/windows/win32/api/appmodel/nf-appmodel-getpackagepathbyfullname2
+/// @see FindPackageFileInPackage()
+/// @see FindPackageFileOptions
+STDAPI FindPackageFileInPackageGraph(
+    _In_ PCWSTR filename,
+    _In_ FindPackageFileOptions options,
+    _Outptr_result_maybenull_ PWSTR* packageFile) noexcept;
```
jonwis, line 628: Missing `(`

specs/package/Package.md
```diff
@@ -0,0 +1,505 @@
...
+        /// @param filename file to locate.
+        /// @param packageFile absolute path to the packaged file, or empty string ("") if not found.
+        /// @param options options for the search.
+        /// @note The package path search order is External(User or Machine) -> Mutable -> Install.
+        /// @note If a package has a UserExternal location then MachineExternal location is not checked (even if the package has one).
+        /// @see https://learn.microsoft.com/en-us/windows/win32/api/appmodel/nf-appmodel-getpackagepathbyfullname2
+        /// @see FindPackageFileInPackageGraph
+        /// @see FindPackageFileOptions
+        [method_name("FindPackageFileInPackageWithOptions")]
+        static String FindPackageFileInPackage(String packageFullName, String filename, FindPackageFileOptions options);
```
jonwis: Naming: Consider reducing repetition:

* `Package.GetFilePath(...)`
* `Package.GetFilePathForProcess(...)`

Scottj1s:
Do any of these APIs require admin, as e.g. [PM.FindPackage](https://learn.microsoft.com/en-us/uwp/api/windows.management.deployment.packagemanager.findpackage?view=winrt-26100) does?  Would be really nice to not have that restriction (and a selling point!)

specs/package/Package.md
DrusTheAxe: Yes.

The underlying problem's cropped up multiple times recently (the most recent, 3x in 2 weeks :-( ). Expecting developers to do the multi-step recipe (and correctly) is unreasonable given empirical evidence and common sense.

This API doesn't break radical new ground, but provides a valuable quality-of-life improvement at a low cost. A worthy ROI.

specs/package/Package.md
DrusTheAxe: Lolz. Yeah, I intentionally kept the names superdescriptive awaiting review for good suggestions. I like `GetFilePath`. Thanks

Good point on Find vs Get.

specs/package/Package.md
DrusTheAxe: [SearchPath](https://learn.microsoft.com/en-us/windows/win32/api/processenv/nf-processenv-searchpathw) predates and is oblivious to package graph.

I thought of SearchPath (and the current API's spiritually inspired by it) but it doesn't fit the bill.

If `lpPath` is NULL `SearchPath` searches "the system path" (PATH envvar?) and the current directory. Or vice versa, depending on the `SafeProcessSearchMode` regkey. Either way, no package graph, just undesirable things we wouldn't want in our search.

If `lpPath` is not NULL it searches the provided path list. But I avoided leveraging SearchPathW in the implementation as it requires all paths be provided up front - we'd have to fetch all the paths to build the `lpPath` parameter and that has a small but non-zero cost.

The intent is a pair of APIs providing similar net result as...
* **Package** a la `ApplicationData.Current` vs `ApplicationDataManager.CreateForPackageFamily()` -- Get `ApplicationData` for me (current process) vs Get `ApplicationData` for a specified package
* **Package Graph** a la `GetCurrentPackagePath()` vs `GetPackagePathByFullName()` -- Get pkgdir for me (current process) vs Get pkgdir of a specified package
The latter are for a running packaged process, the former when not.

DrusTheAxe: P.S. In my past experience `SearchPath()` stopped prematurely if encountering ERROR_ACCESS_DENIED. But it's been a while. I haven't looked to see if it still does that

specs/package/Package.md
DrusTheAxe: Yes. Fully.

specs/package/Package.md
DrusTheAxe: No. If `a.exe` is a packaged process with a package graph of `[A,B,C]` then...

...`Package.FindPackageFileInPackage("", "foo.winmd");` looks in package `A`
...`Package.FindPackageFileInPackageGraph("foo.winmd");` looks in packages `A`, `B`, `C`

`Package.FindPackageFileInPackage("", "foo.winmd");` is equivalent to `Package.FindPackageFileInPackage(Windows.ApplicationModel.Package.Current.Id.FullName, "foo.winmd");`

`Package.FindPackageFileInPackageGraph("foo.winmd");` is equivalent to
```
f = `Package.FindPackageFileInPackage(Windows.ApplicationModel.Package.Current.Id.FullName, "foo.winmd");`
if (!f)
{
    foreach (Windows.ApplicationModel.Package p IN Windows.ApplicationModel.Package.Current.Dependencies)
    {
        f = `Package.FindPackageFileInPackage(p.Id.FullName, "foo.winmd");`
        if (f)
        {
             break;
        }
    }
}
```

specs/package/Package.md
DrusTheAxe: > This is on the Package object, maybe it should be on some kind of PackageGraph object instead?

That's a really good suggestion! I've been thinking of a 1st class package graph object in other cases, don't know why it didn't occur to me here.

How about this:

```
namespace Microsoft.Windows.ApplicationModel
{
    runtimeclass Package
    {
        String GetFilePath(string filename);
        String GetFilePath(string packageFullName, string filename);
        String GetFilePath(string packageFullName, string filename, GetFilePathOptions options);
    }
    runtimeclass PackageGraph
    {
        String GetFilePath(string filename);
        String GetFilePath(string packageFullName, string filename);
        String GetFilePath(string packageFullName, string filename, GetFilePathOptions options);
    }
}
```

DrusTheAxe: >The answer is for "the running instance of the package" -- I can't ask this of a non-running package?

You can
1. YES: Get(pkgfullname) uses the defined (static) package graph
2. YES: Get(current process) uses the running process' package graph
3. NO/vFUTURE: Get(process X) to use some other (not current) process' package graph

#1 works off the installed info for a package so there are only static dependencies.

#2 uses the current process' package graph which can contain static (declared) and dynamic dependencies

#3 is a process using _another process'_ package graph (containing static+dynamic dependencies). To succeed we'd need to know the actual target process e.g. foo.exe runs and calls `TryCreate/AddPackageDepedency(pkgfamilyname="X"...)` and another instance of foo.exe runs and calls `Try/Add("Y",...)` and now which process' package graph do we mean for foo.exe?

This is kinda neat but we'd need to resolve complications -- how to identify the target process, how to express that in WinRT (is a process' `HANDLE` reasonable WinRT currency today?), plus security considerations and maybe more. I haven't thought on it before (this is 1st time the concept's even come up). Not needed today but an interesting vFuture idea.

specs/package/Package.md
DrusTheAxe: See above: `GetFilePath(...)` and new `PackageGraph` runtimeclass

DrusTheAxe:
> Do any of these APIs require admin, as e.g. [PM.FindPackage](https://learn.microsoft.com/en-us/uwp/api/windows.management.deployment.packagemanager.findpackage?view=winrt-26100) does? Would be really nice to not have that restriction (and a selling point!)

No

`PM.FindPackage` requires admin privilege because it access per-machine and other-user information (not just current user).

These APIs only use current-user information thus no admin privilege needed.

specs/package/Package.md
```diff
@@ -0,0 +1,505 @@
...
+        None = 0,
+
+        /// Include the package's Install path in the file search order
+        SearchInstallPath = 1,
+
+        /// Include the package's Mutable path (if it has one) in the file search order
+        SearchMutablePath = 2,
+
+        /// Include the package's Machine External path (if it has one) in the file search order
+        SearchMachineExternalPath = 3,
```
FrankC-msft: This enum is a Flags (is being |'ed together in samples) but is being defined with underlying values as a closed set.

Should be 0x1 -> 0x8, not 1,2,3,4.

DrusTheAxe:
> How will these APIs play out when Administrator Protection is turned on?

No different than if you call existing APIs today e.g. [GetCurrentPackagePath2()](https://learn.microsoft.com/en-us/windows/win32/api/appmodel/nf-appmodel-getcurrentpackagepath2), [GetPackagePathByFullName2()](https://learn.microsoft.com/en-us/windows/win32/api/appmodel/nf-appmodel-getpackagepathbyfullname2), [GetCurrentPackageInfo2](https://learn.microsoft.com/en-us/windows/win32/api/appmodel/nf-appmodel-getcurrentpackageinfo2), [GetPackageInfo2()](https://learn.microsoft.com/en-us/windows/win32/api/appmodel/nf-appmodel-getpackageinfo2), [Windows.ApplicationModel.Package.EffectivePath](https://learn.microsoft.com/en-us/uwp/api/windows.applicationmodel.package.effectivepath?view=winrt-26100) and so on. Administrator Protection should handle these or have a good explanation why not and what to do about it.

specs/package/Package.md
DrusTheAxe: RECOMMEND: Fix values to be a bitfield and add `[Flags]` attribute to the enum

specs/package/Package.md
DrusTheAxe: Misread?

Open parenthesis line 484
Close parenthesis line 488

specs/package/Package.md
DrusTheAxe: Misread?

line 499 Open parenthesis
line 502 Close parenthesis

specs/package/Package.md
```diff
@@ -0,0 +1,505 @@
...
+}
+```
+
+### 4.1.2. C++ Example
+
+```c++
+wil::unique_process_heap_string GetResourcesPri(PCWSTR packageFullName)
+{
+    FindPackageFileOptions options{};
+    wil::unique_process_heap_string absoluteFilename;
```
dmachaj, line 208: I don't understand why this is a unique process heap string.  It is a WinRT API so the `String` type in the .idl file should make this an `HSTRING`.  Therefore I'd expect this to be a `winrt::hstring` (or equivalent).

specs/package/Package.md
```diff
@@ -0,0 +1,505 @@
...
+* Content files i.e. all other files (*.exe, *.dll, *.jpg, *.dat, ...)
+
+A package always has an Install location and may also have other locations:
+
+* Install location
+* Mutable location
+* Machine External location
+* User External location
+
+Mutable and External locations are optional and not mutually exclusive.
```
oldnewthing: Was momentarily confused since I saw "Mutable and External locations" but four bullets, so was looking for the one that *was* mutually exclusive. But none of them are. maybe "The locations aside from the Install location are optional and may be used in any combination."

specs/package/Package.md
```diff
@@ -0,0 +1,505 @@
...
+  * Package Foo has Machine External location = Q:\Foo
+  * Package Foo is registered for Bob has a User External location = X:\Bar
+  * Package Foo is registered for Mary w/no User External location
+  * THUS:
+    * Bob uses his User External location (X:\Bar) and ignores the Machine External location
+    * Mary has no User External location so she uses the Machine External location (Q:\Foo)
+
+Packages typically have Content files in one location but this isn't required. For example, some
+files can appear in any of multiple locations and some code isn't External location aware (sometimes
+ByDesign, sometimes due to bugs). Changing behavior reducing location awareness poses appcompat
```
oldnewthing: I think a word missing. "Changing behavior **by** reducing location awareness"?

specs/package/Package.md
```diff
@@ -0,0 +1,505 @@
...
+risks leading some features to need to broaden their support to search across multiple locations.
+
+The `FindPackageFile*()` APIs search the filesystem for a specified file as appropriate across the
+package(s)' location(s).
+
+The APIs are available as Win32 exports and WinRT, similar to existing APIs in `appmodel.h` and
+`namespace Windows.ApplicationModel`.
+
+**NOTE:** This functionality is available today but requires multiple steps, and understanding
+packages may have multiple locations and how to search across them. This puts a high bar on all
```
oldnewthing: Garden path sentence. Maybe "and understanding **that** packages may have..." to make less ambiguous.

specs/package/Package.md
```diff
@@ -0,0 +1,505 @@
...
+
+### 3.1.3. FindPackageFileInPackageGraph
+
+`FindPackageFileInPackageGraph(filename[, options])` searches the location(s) of each package in the caller's
+package graph for the specified file, per the algorithm:
+
+```
+FOREACH package IN GetPackageGraph():
+
+    file = FindPackageFileInPackage(package.fullname, filename)
```
oldnewthing: Presumably the packages in the package graph are enumerated in an unspecified order? Or is there a specific order produced by GetPackageGraph() which FindPackageFileInPackageGraph() honors?

specs/package/Package.md
```diff
@@ -0,0 +1,505 @@
...
+
+        /// Include the package's User External path (if it has one) in the file search order
+        SearchUserExternalPath = 4,
+    }
+
+    [contract(PackageRuntimeContract, 1)]
+    runtimeclass Package
+    {
+        /// Return the absolute path to the file in the package.
+        /// @param packageFullName the package, or empty string ("") to use the current process' package identity.
```
oldnewthing: It seems that the corresponding appmodel.h function GetPackagePathByFullName2  does not have the "empty string = current package" convention.

Should we be worried that somebody might accidentally pass null and search themselves by mistake?

```csharp
string packageToSearch = GetExtensionPackageName(); // returns empty string if extension not installed
/* oops, forgot to do "if (String.IsNullOrEmpty(packageToSearch)" */
string path = FindPackageFileInPackage(packageToSearch, "contoso-config.json");
```

Accidentally finds the calling app's own contoso-config.json instead of saying "Sorry, there is no file because there isn't even an extension package." (Either by explicit failure or by throwing an InvalidArgumentException.)

specs/package/Package.md
DrusTheAxe: Ordered

Package Graph is an ordered list of packages, with the order very (a) deterministic, (b) controllable by the developer and (c) knowable by mortals (no OldNewThing SuperDeveloper expertise needed :-)

DLL Search Order (LoadLibrary), MRT, WinRT and others (and soon FindPackageFile*() :-) use the Package Graph as an ordered list when they need a hunt sequence. You can find more details on the package graph rules and ordering here...
* [3.1.3.2. Package Graph](https://github.com/microsoft/WindowsAppSDK/blob/main/specs/dynamicdependencies/DynamicDependencies.md#3132-package-graph)
* [3.1.3.3. Add to Package Graph](https://github.com/microsoft/WindowsAppSDK/blob/main/specs/dynamicdependencies/DynamicDependencies.md#3133-add-to-package-graph)
* [7. Static Package Dependency Resolution Algorithm](https://github.com/microsoft/WindowsAppSDK/blob/main/specs/dynamicdependencies/DynamicDependencies.md#7-static-package-dependency-resolution-algorithm)

specs/package/Package.md
DrusTheAxe: This is the C++ example using the uses the Win32 export (C) API - line 199 `FindPackageFileInPackage()`, not WinRT, so it returns the output from that.

Could have returned `_Outptr_ PWSTR` and doc'd as "you need to deallocate the return value via `HeapFree()`.

RECOMMEND: Change the example function to return `std::wstring` as that's a batter C++ example, and it avoids any question over which memory deallocator to use.

specs/package/Package.md
DrusTheAxe: Per the thread below https://github.com/microsoft/WindowsAppSDK/pull/5885/files#r2404589760 the WinRT API should change to

```
namespace Microsoft.Windows.ApplicationModel
{
    runtimeclass Package
    {
        String GetFilePath(string filename);
        String GetFilePath(string packageFullName, string filename);
        String GetFilePath(string packageFullName, string filename, GetFilePathOptions options);
    }
    runtimeclass PackageGraph
    {
        String GetFilePath(string filename);
        String GetFilePath(string packageFullName, string filename);
        String GetFilePath(string packageFullName, string filename, GetFilePathOptions options);
    }
}
```

We can say all strings must be valued - any empty input strings are an error (E_INVALIDARG) and call `GetFilePath(fn)` to use current process' package identity, or `GetFilePath(pfn,fn)` to use the provided identity.

That half solves the ambiguity. Ideally you'd want 4 overloads:
```
        String GetFilePath(string filename);
        String GetFilePath(string filename, GetFilePathOptions options);
        String GetFilePath(string packageFullName, string filename);
        String GetFilePath(string packageFullName, string filename, GetFilePathOptions options);
```
but 2+3 have same arity which is a no-no for WinRT overloading. Possible solutions:

**A.** 3 overloads, to use GetFilePathOptions you must also specify the packagefullname e.g. `GetFilePath(Windows.ApplicationModel.Package.Current.Id.FullName, filename, options)`

**B.** Change method name e.g.
```
        String GetFilePath(string filename);
        String GetFilePath(string packageFullName, string filename);

        String GetFilePathWithOptions(string filename, GetFilePathOptions options);
        String GetFilePathWithOptions(string packageFullName, string filename, GetFilePathOptions options);
```

**C.** Declare pkgfullname is optional, if empty use the caller's identity

**D.** ???

Thoughts?

specs/package/Package.md
DrusTheAxe: No. Providing legos and soldering gun and 'assembly required' is difficult and/or friction for many developers to adopt. Python's "batteries include" motto is a better call here.

Most MSIX-related APIs are used via WinRT (because they only exist in WinRT, or as both Flat-C and WinRT but folks choose the latter). The intent is to always provide WinRT APIs, and also provide Flat-C APIs where necessary, or where extending or superceding APIs in appmodel.h. And if/when Flat-C APIs are provided, the WinRT APIs are built over them so no duplicate implementations ('cause, maintenance, perf, etc).

That calculus might differ if we didn't have the rich set of APIs in appmodel.h. But we do, so providing both (Flat-C and WinRT) has merit.

specs/package/Package.md
DrusTheAxe: Going with Option A as least-evil of the bunch. Will see if next round of review has anything to say.

specs/package/Package.md
```diff
@@ -0,0 +1,578 @@
...
+        /// Return the absolute path to the file in the package.
+        /// @param packageFullName the package.
+        /// @param filename file to locate.
+        /// @param packageFile absolute path to the packaged file, or empty string ("") if not found.
+        /// @note The package path search order is External(User or Machine) -> Mutable -> Install.
+        /// @note If a package has a UserExternal location then MachineExternal location is not checked (even if the package has one).
+        /// @see https://learn.microsoft.com/en-us/windows/win32/api/appmodel/nf-appmodel-getpackagepathbyfullname2
+        /// @see PackageGraph.GetFilePath
+        [method_name("GetFilePathInPackage")]
+        static String GetFilePath(String packageFullName, String filename);
```
oldnewthing: Overloading convention is that new parameters go at the end. Could we flip these?

GetFilePath("foo.txt"); // from current package
GetFilePath("foo.txt", "Awesome_bmbjhsdfk"); // from specific package
GetFilePath("foo.txt", "Awesome_bmbjhsdfk", options); // from specific package with search options

specs/package/Package.md
```diff
@@ -0,0 +1,578 @@
...
+{
+    [contractversion(1)]
+    apicontract PackageRuntimeContract{};
+
+    /// Options for GetFilePath*()
+    /// @see Package.GetFilePath
+    /// @see PackageGraph.GetFilePath
+    [contract(PackageRuntimeContract, 1)]
+    [flags]
+    enum GetFilePathOptions
```
oldnewthing, line 444: At first I was worried about "what if I want to pass options when looking in the current package?" then saw that all the options are about how to find the package [and therefore don't apply to the current package version]. Maybe call the enum "PackageSearchOptions"? That way if we have actual options for GetFilePath later (e.g. "ignore extensions", "use MRT"), we can use the name GetFilePathOptions for that.

specs/package/Package.md
```diff
@@ -0,0 +1,578 @@
...
+        /// @param packageFullName the package.
+        /// @param filename file to locate.
+        /// @param packageFile absolute path to the packaged file, or empty string ("") if not found.
+        /// @param options options for the search.
+        /// @note The package path search order is External(User or Machine) -> Mutable -> Install.
+        /// @note If a package has a UserExternal location then MachineExternal location is not checked (even if the package has one).
+        /// @see https://learn.microsoft.com/en-us/windows/win32/api/appmodel/nf-appmodel-getpackagepathbyfullname2
+        /// @see PackageGraph.GetFilePath
+        /// @see PackageGraph.GetFilePathOptions
+        [method_name("GetFilePathWithOptions")]
```
oldnewthing: GetFilePathInPackageWithOptions?

DrusTheAxe: RECOMMEND: Do that

specs/package/Package.md
DrusTheAxe: RECOMMEND: Do that

specs/package/Package.md
DrusTheAxe: They apply to the current package version.

* A package always has A and may have B, C and/or D.
* A package graph is an array of packages, each of which always has A and may have B, C and/or D.
* The options control whether we should look at a package's A, B, C and/or D.

We have 4 options today (search Install/Mutable/MachineExternal/UserExternal) and they apply to all methods.

DrusTheAxe: Hmmm. We probably should support the other filtering properties today, as there's known uses
* Filter by package type: SearchMainPackages, SearchFrameworkPackages, SearchOptionalPackages, SearchResourcePackages
* Filter by static vs dynamic: SearchStaticPackages, SearchDynamicPackages

These shine for PackageGraph as common cases include "ignore Resource packages" (ie SearchMain|SearchFramework|SearchOptional), "only search Frameworks" and "only search static (manifested) packages".

Static+Dynamic aren't meaningful for Package.Get*() as this works off definitional information (what's installed) and not runtime (what a process is touching).

These are all existing filters you can pass to Get[Current]PackageInfo[2]() e.g. [PACKAGE_FILTER_DYNAMIC](https://learn.microsoft.com/en-us/windows/win32/appxpkg/package-constants), [PACKAGE_FILTER_OPTIONAL](https://learn.microsoft.com/en-us/windows/win32/appxpkg/package-constants), ...

PackageType is low value for Package.Get*() (there's an obscure scenario). It's not _harmful_ to have one enum with all these, but should there be 2?

Or should it be a `GetFilePathOptions` runtimeclass instead of an enum, like [Package.FindRelatedPackages(...FindRelatedPackageOptions...)](https://learn.microsoft.com/uwp/api/windows.applicationmodel.package.findrelatedpackages)?

```c#
    [contract(PackageRuntimeContract, 1)]
    [flags]
    enum GetFilePathOptions
    {
        /// Default behavior
        None = 0,

        /// Include the package's Install path in the file search order
        SearchInstallPath = 0x0001,

        /// Include the package's Mutable path (if it has one) in the file search order
        SearchMutablePath = 0x0002,

        /// Include the package's Machine External path (if it has one) in the file search order
        SearchMachineExternalPath = 0x0004,

        /// Include the package's User External path (if it has one) in the file search order
        SearchUserExternalPath = 0x0008,

        /// Include Main packages in the file search order
        SearchMainPackages = 0x0010,

        /// Include Framework packages in the file search order
        SearchFrameworkPackages = 0x0020,

        /// Include Optional packages in the file search order
        SearchOptionalPackages = 0x0040,

        /// Include Resource packages in the file search order
        SearchResourcePackages = 0x0080,

        /// Include HostRuntime packages in the file search order
        SearchHostRuntimePackages = 0x0100,

        /// Include Static package dependencies in the file search order
        SearchStaticDependencies = 0x0200,

        /// Include Dynamic package dependencies in the file search order
        SearchDynamicDependencies = 0x0400,
    }
```

vs

```
    runtimeclass GetFilePathOptions
    {
        [contract(PackageRuntimeContract, 1)]
        GetFilePathOptions();

        /// Include the package's Install path in the file search order
        Boolean SearchInstallPath;

        /// Include the package's Mutable path (if it has one) in the file search order
        Boolean SearchMutablePath;

        /// Include the package's Machine External path (if it has one) in the file search order
        Boolean SearchMachineExternalPath;

        /// Include the package's User External path (if it has one) in the file search order
        Boolean SearchUserExternalPath;

        /// Include Main packages in the file search order
        Boolean SearchMainPackages;

        /// Include Framework packages in the file search order
        Boolean SearchFrameworkPackages;

        /// Include Optional packages in the file search order
        Boolean SearchOptionalPackages;

        /// Include Resource packages in the file search order
        Boolean SearchResourcePackages;

        /// Include HostRuntime packages in the file search order
        Boolean SearchHostRuntimePackages;

        /// Include Static package dependencies in the file search order
        Boolean SearchStaticDependencies;

        /// Include Dynamic package dependencies in the file search order
        Boolean SearchDynamicDependencies;
}
```

DrusTheAxe: Per offline conversation
>I prefer an enum for "inbound", a class for 'outbound". This is inbound, so enum.

RECOMMEND: Enum it is

PR #5895, opened 2025-10-08
Avoid duplicate 'WindowsAppRuntimeAutoInitializer.cpp' compilation

### LINKED ISSUE #5395: warning MSB8027: Two or more files with the name of WindowsAppRuntimeAutoInitializer.cpp will produce outputs to the same location.
I got this warning in my project. Later it is followed by this warning:
WindowsAppRuntimeAutoInitializer.obj : warning LNK4042: object specified more than once; extras ignored

The solution consist of (among others) two c++ projects: exe and dll - both are using the same version WindowsAppSDK (1.7.250401001) nuget.

I think it is caused by WindowsAppSDK nuget that has `WindowsAppSDK-Nuget-Native.AutoInitializer.targets` 
in `buildTransitive` directory. Also I think older versions of WindowsAppSDK haven't produced such warnings.

Is there a way to prevent the warnings?

mschofie:
Fixes #5395.

`build/NuSpecs/WindowsAppSDK-Nuget-Native.AutoInitializer.targets` attempts to add a `ClCompile` Item during the `WindowsAppRuntimeAutoInitializer` Target. In doing so it attempts to use the existing metadata by uttering `%(PreprocessorDefinitions)`. This syntax is fine for `ItemGroup`s _outside_ of a Target, but within a Target, it causes the (implicit) task that creates the Item to be _batched_ ([documentation](https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-batching?view=vs-2022#task-batching)) - causing an Item with the Identity of `WindowsAppRuntimeAutoInitializer.cpp` to be added for every unique set of `ClCompile.PreprocessorDefinitions`. The duplicate `ClCompile` Items cause a cascade of subsequent errors.

The fix is to add the `ClCompile` in two steps:
1. Create the `ClCompile` without specifying any metadata.
2. Modify the metadata for the ClCompile, using a condition for `'%(Identity)' == '$(WindowsAppRuntimeAutoInitializerPath)'` to force a single task batch (for the Item we're looking for).

Thanks to @chwarr for help with this.

### DIFF
--- a/build/NuSpecs/WindowsAppSDK-Nuget-Native.AutoInitializer.targets
+++ b/build/NuSpecs/WindowsAppSDK-Nuget-Native.AutoInitializer.targets
@@ -2,36 +2,42 @@
 <!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
-  <Target Name="WindowsAppRuntimeAutoInitializer">
-    <PropertyGroup>
-      <WindowsAppRuntimeAutoInitializerPath>$(MSBuildThisFileDirectory)..\..\include\WindowsAppRuntimeAutoInitializer.cpp</WindowsAppRuntimeAutoInitializerPath>
-      <WindowsAppRuntimeAutoInitializerDefines />
-      <WindowsAppRuntimeAutoInitializerDefines Condition="'$(WindowsAppSdkBootstrapInitialize)'=='true'">$(WindowsAppRuntimeAutoInitializerDefines);MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_BOOTSTRAP</WindowsAppRuntimeAutoInitializerDefines>
-      <WindowsAppRuntimeAutoInitializerDefines Condition="'$(WindowsAppSdkDeploymentManagerInitialize)'=='true'">$(WindowsAppRuntimeAutoInitializerDefines);MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_DEPLOYMENTMANAGER</WindowsAppRuntimeAutoInitializerDefines>
-      <WindowsAppRuntimeAutoInitializerDefines Condition="'$(WindowsAppSdkUndockedRegFreeWinRTInitialize)'=='true'">$(WindowsAppRuntimeAutoInitializerDefines);MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_UNDOCKEDREGFREEWINRT</WindowsAppRuntimeAutoInitializerDefines>
-      <WindowsAppRuntimeAutoInitializerDefines Condition="'$(WindowsAppSdkCompatibilityInitialize)'=='true'">$(WindowsAppRuntimeAutoInitializerDefines);MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_COMPATIBILITY</WindowsAppRuntimeAutoInitializerDefines>
-    </PropertyGroup>
+  <PropertyGroup>
+    <WindowsAppRuntimeAutoInitializerPath>$(MSBuildThisFileDirectory)..\..\include\WindowsAppRuntimeAutoInitializer.cpp</WindowsAppRuntimeAutoInitializerPath>
 
-    <!--
-    Add a ClCompile for the 'WindowsAppRuntimeAutoInitializer'. This is done in two steps;
-      1. Add the ClCompile item with no metadata, so the default metadata is applied.
-      2. Add a second ClCompile item that uses the %(Identity) metadata to force a 'task batch' with just the one item
-          we want to modify. This allows us to use "%(PreprocessorDefinitions)" (without causing more task batching) to
-          append the PreprocessorDefinitions to apply.
-    -->
-    <ItemGroup>
-      <ClCompile Include="$(WindowsAppRuntimeAutoInitializerPath)" />
-      <ClCompile Condition=" '%(Identity)' == '$(WindowsAppRuntimeAutoInitializerPath)' ">
-        <PrecompiledHeader>NotUsing</PrecompiledHeader>
-        <PreprocessorDefinitions>$(WindowsAppRuntimeAutoInitializerDefines);%(PreprocessorDefinitions)</PreprocessorDefinitions>
-      </ClCompile>
-    </ItemGroup>
-  </Target>
+    <!-- If any option is set that needs to add a preprocessor definition, enable 'WindowsAppRuntimeAutoInitializerEnabled' -->
+    <WindowsAppRuntimeAutoInitializerEnabled Condition="'$(WindowsAppSdkBootstrapInitialize)'=='true'">true</WindowsAppRuntimeAutoInitializerEnabled>
+    <WindowsAppRuntimeAutoInitializerEnabled Condition="'$(WindowsAppSdkDeploymentManagerInitialize)'=='true'">true</WindowsAppRuntimeAutoInitializerEnabled>
+    <WindowsAppRuntimeAutoInitializerEnabled Condition="'$(WindowsAppSdkUndockedRegFreeWinRTInitialize)'=='true'">true</WindowsAppRuntimeAutoInitializerEnabled>
+    <WindowsAppRuntimeAutoInitializerEnabled Condition="'$(WindowsAppSdkCompatibilityInitialize)'=='true'">true</WindowsAppRuntimeAutoInitializerEnabled>
+
+    <!-- If no options are set that require the auto-initializer, and a consumer hasn't explicitly set it, default to 'false' -->
+    <WindowsAppRuntimeAutoInitializerEnabled Condition="'$(WindowsAppRuntimeAutoInitializerEnabled)'==''">false</WindowsAppRuntimeAutoInitializerEnabled>
+  </PropertyGroup>
 
   <PropertyGroup>
     <BeforeClCompileTargets>
         $(BeforeClCompileTargets); WindowsAppRuntimeAutoInitializer;
     </BeforeClCompileTargets>
   </PropertyGroup>
+
+  <ItemDefinitionGroup>
+    <!-- Add the appropriate definition to the ClCompile ItemDefinitionGroup to apply to all compilations -->
+    <ClCompile>
+      <PreprocessorDefinitions Condition="'$(WindowsAppSdkBootstrapInitialize)'=='true'">MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_BOOTSTRAP;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions Condition="'$(WindowsAppSdkDeploymentManagerInitialize)'=='true'">MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_DEPLOYMENTMANAGER;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions Condition="'$(WindowsAppSdkUndockedRegFreeWinRTInitialize)'=='true'">MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_UNDOCKEDREGFREEWINRT;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions Condition="'$(WindowsAppSdkCompatibilityInitialize)'=='true'">MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_COMPATIBILITY;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+    </ClCompile>
+  </ItemDefinitionGroup>
+
+  <!-- If WindowsAppRuntimeAutoInitializerEnabled is true add a Target to add the WindowsAppRuntimeAutoInitializer.cpp file -->
+  <Target Name="WindowsAppRuntimeAutoInitializer" Condition=" '$(WindowsAppRuntimeAutoInitializerEnabled)'=='true' ">
+    <ItemGroup>
+      <ClCompile Include="$(WindowsAppRuntimeAutoInitializerPath)">
+        <PrecompiledHeader>NotUsing</PrecompiledHeader>
+      </ClCompile>
+    </ItemGroup>
+  </Target>
 
 </Project>


chwarr:
LGTM

Scottj1s:
@mschofie thanks for taking a stab at this.  I don't think it's a complete fix, since we had to scrap a couple earlier attempts (#5685, #5782).

My concern is that any batching at all (whether PreprocessorDefinitions, AdditionalOptions, etc) just sets a trap for subsequent dynamic source additions.  

Can we avoid batching entirely by applying macros to all sources after they've been dynamically added, and that supports Xaml generated sources, etc?
https://github.com/microsoft/WindowsAppSDK/pull/5685#issuecomment-3229606953

mschofie:
> I don't think it's a complete fix, since we had to scrap a couple earlier attempts (https://github.com/microsoft/WindowsAppSDK/pull/5685, https://github.com/microsoft/WindowsAppSDK/pull/5782).

Thanks for linking to the earlier attempts. TBH, both of those attempts appear to have the same problem as the original bug, that my approach doesn't. For example, #5782 says:

> Root Cause: When two targets in BeforeClCompileTargets both modify the same parameter (e.g., PreprocessorDefinitions), the second target to execute may cause duplicate entries to be added to the cl compile command due to parameter modification conflicts.

And that's not right - it's not that they're modifying the same ~parameter~ Item metadata, it's because they're creating multiple sets of Items when batched by `%(PreprocessorDefinitions)`. Even if those fixes worked, they'd break if a consumer had ClCompile Items with different `%(PreprocessorDefinitions)`. This PR uses the `'%(Identity)' == '$(WindowsAppRuntimeAutoInitializerPath)'` Condition to force a single Item batch.

> Can we avoid batching entirely by applying macros to all sources after they've been dynamically added, and that supports Xaml generated sources, etc?
https://github.com/microsoft/WindowsAppSDK/pull/5685#issuecomment-3229606953

I don't understand how that mitigates the batching. If there's two `ClCompile` Items with unique `PreprocessorDefinitions` metadata created during the evaluation phase, then you'd get two batches (each with one `ClCompile` Item in) if you were to access `%(PreprocessorDefinitions)` during an `ItemGroup` within a `Target` (without forcing a smaller batch first). Here's a trivial example that shows batching of 'static' Items:

```xml
<Project>
  <ItemGroup>
    <ClCompile Include="foo" PreprocessorDefinitions="FOO" />
    <ClCompile Include="bar" PreprocessorDefinitions="BAR" />
  </ItemGroup>

  <Target Name="Build">
    <Message Text="Batch: %(ClCompile.PreprocessorDefinitions) Items: %(ClCompile.Identity)" />
  </Target>
</Project>
```

If you're OK applying the preprocessor defines globally, then there's a much easier fix - simply decouple the ClCompile Item creation from setting the PreprocessorDefinitions metadata; i.e.
1. If any of the necessary properties are set; add the 'ClCompile' Item
2. Based on which are set, update the PreprocessorDefinitions in an ItemDefinitionGroup. 

Property evaluation and ItemDefinitionGroup updates would have to happen outside of the 'Target', but I'm not sure that's a problem...? This is more 'order dependent' than this current PR, but we're in a .targets file, so I think it's OK. Thoughts? So, something like (I'm coding in an edit box; apologies for typos)

```xml
  <PropertyGroup>
    <WindowsAppRuntimeAutoInitializerPath>$(MSBuildThisFileDirectory)..\..\include\WindowsAppRuntimeAutoInitializer.cpp</WindowsAppRuntimeAutoInitializerPath>
    <WindowsAppRuntimeAutoInitializerEnabled>false</WindowsAppRuntimeAutoInitializerEnabled>
    <WindowsAppRuntimeAutoInitializerEnabled Condition="'$(WindowsAppSdkBootstrapInitialize)'=='true'">true</WindowsAppRuntimeAutoInitializerEnabled>
    <WindowsAppRuntimeAutoInitializerEnabled Condition="'$(WindowsAppSdkDeploymentManagerInitialize)'=='true'">true</WindowsAppRuntimeAutoInitializerEnabled>
    <WindowsAppRuntimeAutoInitializerEnabled Condition="'$(WindowsAppSdkUndockedRegFreeWinRTInitialize)'=='true'">true</WindowsAppRuntimeAutoInitializerEnabled>
    <WindowsAppRuntimeAutoInitializerEnabled Condition="'$(WindowsAppSdkCompatibilityInitialize)'=='true'">true</WindowsAppRuntimeAutoInitializerEnabled>
  </PropertyGroup>

  <Target Name="WindowsAppRuntimeAutoInitializer" Condition=" '$(WindowsAppRuntimeAutoInitializerEnabled)'=='true' ">
    <ItemGroup>
      <ClCompile Include="$(WindowsAppRuntimeAutoInitializerPath)" />
    </ItemGroup>
  </Target>

  <ItemDefinitionGroup>
    <ClCompile>
      <PreprocessorDefinitions Condition="'$(WindowsAppSdkBootstrapInitialize)'=='true'">MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_BOOTSTRAP;%(PreprocessorDefinitions)</PreprocessorDefinitions>
      <PreprocessorDefinitions Condition="'$(WindowsAppSdkDeploymentManagerInitialize)'=='true'">MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_DEPLOYMENTMANAGER;%(PreprocessorDefinitions)</PreprocessorDefinitions>
      <PreprocessorDefinitions Condition="'$(WindowsAppSdkUndockedRegFreeWinRTInitialize)'=='true'">MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_UNDOCKEDREGFREEWINRT;%(PreprocessorDefinitions)</PreprocessorDefinitions>
      <PreprocessorDefinitions Condition="'$(WindowsAppSdkCompatibilityInitialize)'=='true'">MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_COMPATIBILITY;%(PreprocessorDefinitions)</PreprocessorDefinitions>
    </ClCompile>
  </ItemDefinitionGroup>
</Project>
```

Scottj1s:
> And that's not right - it's not that they're modifying the same parameter Item metadata, 

Agreed - that's not a correct RCA

> I don't understand how that mitigates the batching. If there's two ClCompile Items with unique PreprocessorDefinitions metadata

Right - I'm suggesting we eliminate batching via speciation of PreprocessorDefinitions (or any other %BatchingMetadata) and just apply metadata globally to all sources - pch, module, whatever.  I had [suggested](https://github.com/microsoft/WindowsAppSDK/pull/5685#issuecomment-3229606953) doing that during execution after all sources were sure to have been ClCompile included.  But I'm fine with doing it during evaluation phase, via ItemDefinitionGroup - that's very appealing.

mschofie:
Sorry, I don't understand your [suggestion](https://github.com/microsoft/WindowsAppSDK/pull/5685#issuecomment-3229606953). How would you add the PreprocessorDefines? You'd still need to utter `%(PreprocessorDefines)` in an ItemGroup in a Target, wouldn't you?

At the minute, the only two paths I see are:
1. Force a single item batch. What this PR does. 
2. Use an ItemDefinitionGroup. Which becomes a little order dependent (i.e. the `WindowsAppSdkBootstrapInitialize` (and other) properties would have to be set before `build/NuSpecs/WindowsAppSDK-Nuget-Native.AutoInitializer.targets` was evaluated).
 
I'll update this PR with the `ItemDefinitionGroup` after I've buddy tested it, and we can see which approach folks like the best.

Scottj1s:
> You'd still need to utter `%(PreprocessorDefines)` in an ItemGroup in a Target, wouldn't you?

Yes, you would - but that's not the issue per se.  Having different definitions of PreprocessorDefines is what causes the batching.  If we apply the same value to all sources, we get no batching and consequently no duplicate source compilation.

```
<Target Name="AddFoo"  BeforeTargets="ClCompile" AfterTargets="$(BeforeClCompileTargets);$(ComputeCompileInputsTargets);MakeDirsForCl" > 
  <ItemGroup>
    <ClCompile Include="Foo.Cpp">
        <AnyMetadataExceptPreprocessorDefinitions>this_is_safe</>
    </ClCompile>
    <ClCompile>
        <!-- Now, add macros to all sources, immediately before ClCompile, so no batching concerns -->
        <PreprocessorDefinitions>%PreprocessorDefinitions);FOO_DEFINES</PreprocessorDefinitions>
    </ClCompile>
  </ItemGroup>
</Target> 
```

mschofie:
> Having different definitions of PreprocessorDefines is what causes the batching.

Indeed, and I don't think we can prevent different definitions of PreprocessorDefines, because customers can easily write valid vcxproj files with different definitions of PreprocessorDefines. A customer can write the following in their .vcxproj:

```xml
<Project>
<ItemGroup>
    <ClCompile Include="CustomerSource.cpp" />
    <ClCompile Include="MoreCustomerSource.cpp" PreprocessorDefines="%(PreprocessorDefines);ENABLE_FANCY_MODE"/>
</ItemGroup>
</Project>
```

And then batching will occur in an ItemGroup (in a Target) that uses `%(PreprocessorDefines)`.

Scottj1s:
> > Having different definitions of PreprocessorDefines is what causes the batching.
> 
> Indeed, and I don't think we can prevent different definitions of PreprocessorDefines, because customers can easily write valid vcxproj files with different definitions of PreprocessorDefines. A customer can write the following in their .vcxproj:
> 
> ```
> <Project>
> <ItemGroup>
>     <ClCompile Include="CustomerSource.cpp" />
>     <ClCompile Include="MoreCustomerSource.cpp" PreprocessorDefines="%(PreprocessorDefines);ENABLE_FANCY_MODE"/>
> </ItemGroup>
> </Project>
> ```
> 
> And then batching will occur in an ItemGroup (in a Target) that uses `%(PreprocessorDefines)`.

That's true - we can't avoid that.  But we can at least avoid self-inflicted issues like this one, by applying PreprocessorDefinitions globally.  The interesting thing about this issue is that it's the result of our two targets both including ClCompile items with specific PreprocessorDefinitions: AddUndockedRegFreeWinRTCppDefines and WindowsAppRuntimeAutoInitializer.   It's actually the second target that loses, by including its source file into separate batches that the first target has already created.  So I'm fine with any solution that avoids our own (re)batching of existing batches.

mschofie:
I've updated this PR with an implementation that takes the ItemDefinitionGroup-approach. This is - IMHO - a tidier approach. The only downside is that if customers are setting any of

* WindowsAppSdkBootstrapInitialize
* WindowsAppSdkDeploymentManagerInitialize
* WindowsAppSdkUndockedRegFreeWinRTInitialize
* WindowsAppSdkCompatibilityInitialize

within a 'Target', then it won't be honored, they need to be set during evaluation. This seems like a reasonable expectation to me.

Thoughts?

build/NuSpecs/WindowsAppSDK-Nuget-Native.AutoInitializer.targets
```diff
@@ -2,15 +2,19 @@
...
+    <WindowsAppRuntimeAutoInitializerPath>$(MSBuildThisFileDirectory)..\..\include\WindowsAppRuntimeAutoInitializer.cpp</WindowsAppRuntimeAutoInitializerPath>
+    <WindowsAppRuntimeAutoInitializerEnabled>false</WindowsAppRuntimeAutoInitializerEnabled>
+    <WindowsAppRuntimeAutoInitializerEnabled Condition="'$(WindowsAppSdkBootstrapInitialize)'=='true'">true</WindowsAppRuntimeAutoInitializerEnabled>
+    <WindowsAppRuntimeAutoInitializerEnabled Condition="'$(WindowsAppSdkDeploymentManagerInitialize)'=='true'">true</WindowsAppRuntimeAutoInitializerEnabled>
+    <WindowsAppRuntimeAutoInitializerEnabled Condition="'$(WindowsAppSdkUndockedRegFreeWinRTInitialize)'=='true'">true</WindowsAppRuntimeAutoInitializerEnabled>
+    <WindowsAppRuntimeAutoInitializerEnabled Condition="'$(WindowsAppSdkCompatibilityInitialize)'=='true'">true</WindowsAppRuntimeAutoInitializerEnabled>
+  </PropertyGroup>
+
+  <!-- If WindowsAppRuntimeAutoInitializerEnabled is true add a Target to add the WindowsAppRuntimeAutoInitializer.cpp file -->
+  <Target Name="WindowsAppRuntimeAutoInitializer" Condition=" '$(WindowsAppRuntimeAutoInitializerEnabled)'=='true' ">
```
chwarr: If we're going to rely on consuming projects setting these properties before importing the .targets file, this inclusion can be an evaluation-time conditional inclusion; no target needed.

build/NuSpecs/WindowsAppSDK-Nuget-Native.AutoInitializer.targets
```diff
@@ -20,4 +24,14 @@
     </BeforeClCompileTargets>
   </PropertyGroup>
 
+  <ItemDefinitionGroup>
```
chwarr, line 24: Minor: consider putting the IDG after the PropertyGroup and before the targets so that a top-to-bottom read more closely matches the evaluation order.

build/NuSpecs/WindowsAppSDK-Nuget-Native.AutoInitializer.targets
```diff
@@ -2,15 +2,19 @@
 <!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
-  <Target Name="WindowsAppRuntimeAutoInitializer">
+  <PropertyGroup>
+    <WindowsAppRuntimeAutoInitializerPath>$(MSBuildThisFileDirectory)..\..\include\WindowsAppRuntimeAutoInitializer.cpp</WindowsAppRuntimeAutoInitializerPath>
+    <WindowsAppRuntimeAutoInitializerEnabled>false</WindowsAppRuntimeAutoInitializerEnabled>
```
chwarr:
```suggestion
    <WindowsAppRuntimeAutoInitializerEnabled Condition=" '$(WindowsAppRuntimeAutoInitializerEnabled)' == '' ">false</WindowsAppRuntimeAutoInitializerEnabled>
```

In case a consuming project has already set this to true, a conditional assignment will avoid clobbering that value.

Consider moving the false assignment after all the other conditions so that it gets set to false as a last resort if not set to anything else first.

Up to you if you want to test against '' or `!= 'true'`

build/NuSpecs/WindowsAppSDK-Nuget-Native.AutoInitializer.targets
mschofie: IIUC, the point of adding the 'ClCompile' in the Target is to 'hide it' from Solution Explorer. If the ClCompile is added outside of a Target (in a 'top level' ItemGroup) it shows in Solution Explorer.

chwarr: Looks like `<Visible>false</Visible>` metadata doesn't work anymore for cases like this. A target seems fine then.

mschofie:
I've addressed Christopher's comments - thanks! - and pushed a final version. TBH, I hadn't realized that the same pattern is used in multiple places across the build infrastructure - are folks with this PR going in without canonicalizing across the build?

Where should I document - release notes?  - the shift that the listed properties have to be set at evaluation time now?

* WindowsAppSdkBootstrapInitialize
* WindowsAppSdkDeploymentManagerInitialize
* WindowsAppSdkUndockedRegFreeWinRTInitialize
* WindowsAppSdkCompatibilityInitialize

chwarr (approved):

Scottj1s:
> I've addressed Christopher's comments - thanks! - and pushed a final version. TBH, I hadn't realized that the same pattern is used in multiple places across the build infrastructure - are folks with this PR going in without canonicalizing across the build?
> 
> Where should I document - release notes? - the shift that the listed properties have to be set at evaluation time now?
> 
> * WindowsAppSdkBootstrapInitialize
> * WindowsAppSdkDeploymentManagerInitialize
> * WindowsAppSdkUndockedRegFreeWinRTInitialize
> * WindowsAppSdkCompatibilityInitialize

We've already documented that these properties should be set in the project file (that language is imprecise, but implies outside a target):
https://learn.microsoft.com/en-us/windows/apps/package-and-deploy/project-properties#project-properties
We just happened to evaluate them late, in a target.

Scottj1s:
@mschofie Is this ready to complete? Have we confirmed it won't regress WinUI builds?

build/NuSpecs/WindowsAppSDK-Nuget-Native.AutoInitializer.targets
```diff
@@ -2,22 +2,40 @@
...
+      <PreprocessorDefinitions Condition="'$(WindowsAppSdkDeploymentManagerInitialize)'=='true'">MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_DEPLOYMENTMANAGER;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions Condition="'$(WindowsAppSdkUndockedRegFreeWinRTInitialize)'=='true'">MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_UNDOCKEDREGFREEWINRT;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions Condition="'$(WindowsAppSdkCompatibilityInitialize)'=='true'">MICROSOFT_WINDOWSAPPSDK_AUTOINITIALIZE_COMPATIBILITY;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+    </ClCompile>
+  </ItemDefinitionGroup>
+
+  <!-- If WindowsAppRuntimeAutoInitializerEnabled is true add a Target to add the WindowsAppRuntimeAutoInitializer.cpp file -->
+  <Target Name="WindowsAppRuntimeAutoInitializer" Condition=" '$(WindowsAppRuntimeAutoInitializerEnabled)'=='true' ">
+    <ItemGroup>
+      <ClCompile Include="$(WindowsAppRuntimeAutoInitializerPath)" />
```
qiutongMS: I tried to build this change and found that it might be missing `<PrecompiledHeader>NotUsing</PrecompiledHeader>` here.

qiutongMS:
> @mschofie Is this ready to complete? Have we confirmed it won't regress WinUI builds?

After adding `<PrecompiledHeader>NotUsing</PrecompiledHeader>`, I successfully built the Foundation package and tested it on WinUI. Unfortunately, the same issue still occurs on arm64ec build. (The arm64 error seems unrelated to this.)
https://microsoft.visualstudio.com/DefaultCollection/WinUI/_build/results?buildId=132757478&view=results

build/NuSpecs/WindowsAppSDK-Nuget-Native.AutoInitializer.targets
mschofie: Thanks! Sorry I missed that.

mschofie:
> > @mschofie Is this ready to complete? Have we confirmed it won't regress WinUI builds?
> 
> After adding `<PrecompiledHeader>NotUsing</PrecompiledHeader>`, I successfully built the Foundation package and tested it on WinUI. Unfortunately, the same issue still occurs on arm64ec build. (The arm64 error seems unrelated to this.) https://microsoft.visualstudio.com/DefaultCollection/WinUI/_build/results?buildId=132757478&view=results

Thanks for finding the 'PrecompiledHeader' 'NotUsing' regression - the WindowsML samples that I'm using to validate locally don't use precompiled headers, so I missed that - so sorry. Your questions:
1. _Is it ready to complete?_ I think so.
2. _Have we confirmed it won't regress WinUI builds?_ I'm not sure I really understand the question. The implementation that was settled on _will_ change behavior, but it's moving to behavior that - as per [Scott's comment](https://github.com/microsoft/WindowsAppSDK/pull/5895#issuecomment-3428128986) - is documented as the expectation. Beyond that, I've been building the WindowsML samples locally, and assuming that the CI build is covering other bases. Is there more due diligence needed? A pointer to the details would be great.

Regarding "Unfortunately, the same issue still occurs on arm64ec build." - can you share a link to the logs that show this? I'm looking through the linked build, on the "BuildMUX Release_arm64ec" job, there's five stages that error:

1. `Build WinUICppDesktopSampleApp (windows_build_container)` - This has warnings that appear to be unrelated ([link](https://microsoft.visualstudio.com/DefaultCollection/WinUI/WinUI%20Team/_build/results?buildId=132757478&view=logs&j=ff6312d8-49f8-5c13-4e07-e458a3487bdd&t=b1035083-3259-5a33-190d-918c41e9e7af&l=5057)), and an error that is an Internal Compiler Error ([link](https://microsoft.visualstudio.com/DefaultCollection/WinUI/WinUI%20Team/_build/results?buildId=132757478&view=logs&j=ff6312d8-49f8-5c13-4e07-e458a3487bdd&t=b1035083-3259-5a33-190d-918c41e9e7af&l=5079)). 

2. `Build WinUIGallery (windows_build_container)` - This has warnings that are unrelated ([link](https://microsoft.visualstudio.com/DefaultCollection/WinUI/WinUI%20Team/_build/results?buildId=132757478&view=logs&j=ff6312d8-49f8-5c13-4e07-e458a3487bdd&t=4efecb8b-3c47-51bb-f0ec-0d74cb476529&l=12329)), and an error ([link](https://microsoft.visualstudio.com/DefaultCollection/WinUI/WinUI%20Team/_build/results?buildId=132757478&view=logs&j=ff6312d8-49f8-5c13-4e07-e458a3487bdd&t=4efecb8b-3c47-51bb-f0ec-0d74cb476529&l=12415)) that's _similar_ but - from the looks of things, unrelated:
    ```
    "C:\__w\1\s\Samples\WinUIGallery\WinUIGallery\WinUIGallery.sln" (Publish target) (1:2) ->
    "C:\__w\1\s\Samples\WinUIGallery\WinUIGalleryUnitTests\WinUIGalleryUnitTests.csproj" (Publish target) (3:6) ->
    (_ComputeAppxPackagePayload target) ->
        C:\__w\1\s\packages\microsoft.windows.sdk.buildtools.msix\1.5.20250417.1\build\Microsoft.Windows.SDK.BuildTools.MSIX.Packaging.targets(1364,5): error APPX1101: Payload contains two or more files with the same destination path 'WinUIGallery.pdb'. Source files:  [C:\__w\1\s\Samples\WinUIGallery\WinUIGalleryUnitTests\WinUIGalleryUnitTests.csproj]
        C:\__w\1\s\packages\microsoft.windows.sdk.buildtools.msix\1.5.20250417.1\build\Microsoft.Windows.SDK.BuildTools.MSIX.Packaging.targets(1364,5): error APPX1101:         C:\__w\1\s\BuildOutput\obj\arm64ecfre\Samples\WinUIGallery\WinUIGalleryUnitTests\net8.0-windows10.0.22621.0\win-x64\linked\WinUIGallery.pdb [C:\__w\1\s\Samples\WinUIGallery\WinUIGalleryUnitTests\WinUIGalleryUnitTests.csproj]
        C:\__w\1\s\packages\microsoft.windows.sdk.buildtools.msix\1.5.20250417.1\build\Microsoft.Windows.SDK.BuildTools.MSIX.Packaging.targets(1364,5): error APPX1101:         C:\__w\1\s\BuildOutput\obj\arm64ecfre\Samples\WinUIGallery\WinUIGallery\WinUIGallery\net8.0-windows10.0.22621.0\win-x64\linked\WinUIGallery.pdb [C:\__w\1\s\Samples\WinUIGallery\WinUIGalleryUnitTests\WinUIGalleryUnitTests.csproj]
    ```

3. `Build NativeDX12DesktopSample (windows_build_container)` - Unrelated warnings ([link](https://microsoft.visualstudio.com/DefaultCollection/WinUI/WinUI%20Team/_build/results?buildId=132757478&view=logs&j=ff6312d8-49f8-5c13-4e07-e458a3487bdd&t=7204e64e-929c-574f-9848-8374836218c2&l=1712)) and another Internal Compiler Error ([link](https://microsoft.visualstudio.com/DefaultCollection/WinUI/WinUI%20Team/_build/results?buildId=132757478&view=logs&j=ff6312d8-49f8-5c13-4e07-e458a3487bdd&t=7204e64e-929c-574f-9848-8374836218c2&l=1724)).

4. `Build WinUICppIsland2SampleApp (windows_build_container)` - Unrelated warnings ([link](https://microsoft.visualstudio.com/DefaultCollection/WinUI/WinUI%20Team/_build/results?buildId=132757478&view=logs&j=ff6312d8-49f8-5c13-4e07-e458a3487bdd&t=81575154-4352-53b2-49ca-0fbbd9abf957&l=2358)) and another Internal Compiler Error ([link](https://microsoft.visualstudio.com/DefaultCollection/WinUI/WinUI%20Team/_build/results?buildId=132757478&view=logs&j=ff6312d8-49f8-5c13-4e07-e458a3487bdd&t=81575154-4352-53b2-49ca-0fbbd9abf957&l=2375)).

5. `Build WinUICppIslandsSampleApp (windows_build_container)` - Unrelated warnings ([link](https://microsoft.visualstudio.com/DefaultCollection/WinUI/WinUI%20Team/_build/results?buildId=132757478&view=logs&j=ff6312d8-49f8-5c13-4e07-e458a3487bdd&t=29c88d87-bd98-56ea-6a9b-d87792b4def9&l=2352)) and another Internal Compiler Error ([link](https://microsoft.visualstudio.com/DefaultCollection/WinUI/WinUI%20Team/_build/results?buildId=132757478&view=logs&j=ff6312d8-49f8-5c13-4e07-e458a3487bdd&t=29c88d87-bd98-56ea-6a9b-d87792b4def9&l=2362)).

I'm going to go through the passing stages, to look for warnings, but a link would be super handy. Thanks!

I've rebased onto the latest `main` to see if that helps with the Internal Compiler Errors.

Scottj1s:
> 2. _Have we confirmed it won't regress WinUI builds?_ I'm not sure I really understand the question. The implementation that was settled on _will_ change behavior, but it's moving to behavior that - as per [Scott's comment](https://github.com/microsoft/WindowsAppSDK/pull/5895#issuecomment-3428128986) - is documented as the expectation. Beyond that, I've been building the WindowsML samples locally, and assuming that the CI build is covering other bases. Is there more due diligence needed? A pointer to the details would be great.

This refers to the comment I made earlier about the Xaml compiler also dynamically adding generated sources.  I expect your PR should be compatible, but want to confirm.

qiutongMS:
> > > @mschofie Is this ready to complete? Have we confirmed it won't regress WinUI builds?
> > 
> > 
> > After adding `<PrecompiledHeader>NotUsing</PrecompiledHeader>`, I successfully built the Foundation package and tested it on WinUI. Unfortunately, the same issue still occurs on arm64ec build. (The arm64 error seems unrelated to this.) https://microsoft.visualstudio.com/DefaultCollection/WinUI/_build/results?buildId=132757478&view=results
> 
> Thanks for finding the 'PrecompiledHeader' 'NotUsing' regression - the WindowsML samples that I'm using to validate locally don't use precompiled headers, so I missed that - so sorry. Your questions:
> 
> 1. _Is it ready to complete?_ I think so.
> 2. _Have we confirmed it won't regress WinUI builds?_ I'm not sure I really understand the question. The implementation that was settled on _will_ change behavior, but it's moving to behavior that - as per [Scott's comment](https://github.com/microsoft/WindowsAppSDK/pull/5895#issuecomment-3428128986) - is documented as the expectation. Beyond that, I've been building the WindowsML samples locally, and assuming that the CI build is covering other bases. Is there more due diligence needed? A pointer to the details would be great.
> 
> Regarding "Unfortunately, the same issue still occurs on arm64ec build." - can you share a link to the logs that show this? I'm looking through the linked build, on the "BuildMUX Release_arm64ec" job, there's five stages that error:
> 
> 1. `Build WinUICppDesktopSampleApp (windows_build_container)` - This has warnings that appear to be unrelated ([link](https://microsoft.visualstudio.com/DefaultCollection/WinUI/WinUI%20Team/_build/results?buildId=132757478&view=logs&j=ff6312d8-49f8-5c13-4e07-e458a3487bdd&t=b1035083-3259-5a33-190d-918c41e9e7af&l=5057)), and an error that is an Internal Compiler Error ([link](https://microsoft.visualstudio.com/DefaultCollection/WinUI/WinUI%20Team/_build/results?buildId=132757478&view=logs&j=ff6312d8-49f8-5c13-4e07-e458a3487bdd&t=b1035083-3259-5a33-190d-918c41e9e7af&l=5079)).
> 2. `Build WinUIGallery (windows_build_container)` - This has warnings that are unrelated ([link](https://microsoft.visualstudio.com/DefaultCollection/WinUI/WinUI%20Team/_build/results?buildId=132757478&view=logs&j=ff6312d8-49f8-5c13-4e07-e458a3487bdd&t=4efecb8b-3c47-51bb-f0ec-0d74cb476529&l=12329)), and an error ([link](https://microsoft.visualstudio.com/DefaultCollection/WinUI/WinUI%20Team/_build/results?buildId=132757478&view=logs&j=ff6312d8-49f8-5c13-4e07-e458a3487bdd&t=4efecb8b-3c47-51bb-f0ec-0d74cb476529&l=12415)) that's _similar_ but - from the looks of things, unrelated:
>    ```
>    "C:\__w\1\s\Samples\WinUIGallery\WinUIGallery\WinUIGallery.sln" (Publish target) (1:2) ->
>    "C:\__w\1\s\Samples\WinUIGallery\WinUIGalleryUnitTests\WinUIGalleryUnitTests.csproj" (Publish target) (3:6) ->
>    (_ComputeAppxPackagePayload target) ->
>        C:\__w\1\s\packages\microsoft.windows.sdk.buildtools.msix\1.5.20250417.1\build\Microsoft.Windows.SDK.BuildTools.MSIX.Packaging.targets(1364,5): error APPX1101: Payload contains two or more files with the same destination path 'WinUIGallery.pdb'. Source files:  [C:\__w\1\s\Samples\WinUIGallery\WinUIGalleryUnitTests\WinUIGalleryUnitTests.csproj]
>        C:\__w\1\s\packages\microsoft.windows.sdk.buildtools.msix\1.5.20250417.1\build\Microsoft.Windows.SDK.BuildTools.MSIX.Packaging.targets(1364,5): error APPX1101:         C:\__w\1\s\BuildOutput\obj\arm64ecfre\Samples\WinUIGallery\WinUIGalleryUnitTests\net8.0-windows10.0.22621.0\win-x64\linked\WinUIGallery.pdb [C:\__w\1\s\Samples\WinUIGallery\WinUIGalleryUnitTests\WinUIGalleryUnitTests.csproj]
>        C:\__w\1\s\packages\microsoft.windows.sdk.buildtools.msix\1.5.20250417.1\build\Microsoft.Windows.SDK.BuildTools.MSIX.Packaging.targets(1364,5): error APPX1101:         C:\__w\1\s\BuildOutput\obj\arm64ecfre\Samples\WinUIGallery\WinUIGallery\WinUIGallery\net8.0-windows10.0.22621.0\win-x64\linked\WinUIGallery.pdb [C:\__w\1\s\Samples\WinUIGallery\WinUIGalleryUnitTests\WinUIGalleryUnitTests.csproj]
>    ```
> 3. `Build NativeDX12DesktopSample (windows_build_container)` - Unrelated warnings ([link](https://microsoft.visualstudio.com/DefaultCollection/WinUI/WinUI%20Team/_build/results?buildId=132757478&view=logs&j=ff6312d8-49f8-5c13-4e07-e458a3487bdd&t=7204e64e-929c-574f-9848-8374836218c2&l=1712)) and another Internal Compiler Error ([link](https://microsoft.visualstudio.com/DefaultCollection/WinUI/WinUI%20Team/_build/results?buildId=132757478&view=logs&j=ff6312d8-49f8-5c13-4e07-e458a3487bdd&t=7204e64e-929c-574f-9848-8374836218c2&l=1724)).
> 4. `Build WinUICppIsland2SampleApp (windows_build_container)` - Unrelated warnings ([link](https://microsoft.visualstudio.com/DefaultCollection/WinUI/WinUI%20Team/_build/results?buildId=132757478&view=logs&j=ff6312d8-49f8-5c13-4e07-e458a3487bdd&t=81575154-4352-53b2-49ca-0fbbd9abf957&l=2358)) and another Internal Compiler Error ([link](https://microsoft.visualstudio.com/DefaultCollection/WinUI/WinUI%20Team/_build/results?buildId=132757478&view=logs&j=ff6312d8-49f8-5c13-4e07-e458a3487bdd&t=81575154-4352-53b2-49ca-0fbbd9abf957&l=2375)).
> 5. `Build WinUICppIslandsSampleApp (windows_build_container)` - Unrelated warnings ([link](https://microsoft.visualstudio.com/DefaultCollection/WinUI/WinUI%20Team/_build/results?buildId=132757478&view=logs&j=ff6312d8-49f8-5c13-4e07-e458a3487bdd&t=29c88d87-bd98-56ea-6a9b-d87792b4def9&l=2352)) and another Internal Compiler Error ([link](https://microsoft.visualstudio.com/DefaultCollection/WinUI/WinUI%20Team/_build/results?buildId=132757478&view=logs&j=ff6312d8-49f8-5c13-4e07-e458a3487bdd&t=29c88d87-bd98-56ea-6a9b-d87792b4def9&l=2362)).
> 
> I'm going to go through the passing stages, to look for warnings, but a link would be super handy. Thanks!
> 
> I've rebased onto the latest `main` to see if that helps with the Internal Compiler Errors.


Actually, the errors observed on WinUI are: 
Error C1001: Internal compiler error.
Error LNK1000: Internal error during IMAGE::BuildImage.
I understand they may seem unrelated to the fix, but both the PRs Scott mentioned and other fixes I’ve tried (without submitting a PR) for this issue have triggered the same error during check-in or testing — specifically on these WinUI arm64ec sample apps: WinUICppDesktopSampleApp/NativeDX12DesktopSample/WinUICppIsland2SampleApp/WinUICppIslandsSampleApp. These errors haven’t appeared on other samples or platforms.
I haven’t yet identified the connection between these changes and the errors. I’m currently trying to build the affected WinUI samples locally to investigate what might be unique about it.

mschofie:
> Actually, the errors observed on WinUI are:
Error C1001: Internal compiler error.
Error LNK1000: Internal error during IMAGE::BuildImage.
I understand they may seem unrelated to the fix, but both the PRs Scott mentioned and other fixes I’ve tried (without submitting a PR) for this issue have triggered the same error during check-in or testing — specifically on these WinUI arm64ec sample apps: WinUICppDesktopSampleApp/NativeDX12DesktopSample/WinUICppIsland2SampleApp/WinUICppIslandsSampleApp. These errors haven’t appeared on other samples or platforms.
I haven’t yet identified the connection between these changes and the errors. I’m currently trying to build the affected WinUI samples locally to investigate what might be unique about it.

Ah, I see. I'll take a look at that today. Thanks for the context!

mschofie:
> specifically on these WinUI arm64ec sample apps: WinUICppDesktopSampleApp/NativeDX12DesktopSample/WinUICppIsland2SampleApp/WinUICppIslandsSampleApp. These errors haven’t appeared on other samples or platforms.

I'm trying to track down what I need to do to reproduce the problem. Can you share a pointer to "WinUICppDesktopSampleApp/NativeDX12DesktopSample/WinUICppIsland2SampleApp/WinUICppIslandsSampleApp", and some build instructions? Thanks!

mschofie:
OK. A status update:
1. It looks like the "WinUICppDesktopSampleApp/NativeDX12DesktopSample/WinUICppIsland2SampleApp/WinUICppIslandsSampleApp" that you're referring to is a _Microsoft internal_ project. Is that correct?
2. I've attempted to build the containing repository, but continue to hit problems. There's no documentation - that I could find - on how to build, and got as far as I could with the OneTimeSetup.cmd/Init.cmd/Build.cmd scripts.
3. The overall build fails with errors regarding the "Mock Windows AppSDK". In hoping that I'd built 'enough', I attempted to build "WinUICppIslandsSampleApp.sln", but that solution fails with errors, too.

I was simply hoping to build without and with my fix, but that isn't currently possible. Without documentation, I don't understand how I can buddy build or repro the problem.

qiutongMS:
> I was simply hoping to build without and with my fix, but that isn't currently possible. Without documentation, I don't understand how I can buddy build or repro the problem.

1.Yes, I referring to is the Microsoft internal project [WinUI/_git/microsoft-ui-xaml-lift.](https://microsoft.visualstudio.com/DefaultCollection/WinUI/_git/microsoft-ui-xaml-lift)

2.I’m also suffering from the lack of documentation to guide the build. These are the steps I tried today, which should almost follow the previously mentioned pipeline build process. I just managed to succeed as well.
```
@echo off
echo ====================================
echo ARM64EC Release Build Setup
echo ====================================

cd /d %~dp0

echo.
echo [Step 1/6] Cleaning global NuGet cache...
if exist "C:\.tools\.nuget\packages\microsoft.windowsappsdk" (
    rmdir /s /q "C:\.tools\.nuget\packages\microsoft.windowsappsdk"
    echo Cleaned C:\.tools\.nuget\packages\microsoft.windowsappsdk
)
if exist "%USERPROFILE%\.nuget\packages\microsoft.windowsappsdk" (
    rmdir /s /q "%USERPROFILE%\.nuget\packages\microsoft.windowsappsdk"
    echo Cleaned %USERPROFILE%\.nuget\packages\microsoft.windowsappsdk
)

echo.
echo [Step 2/6] Setting NUGET_PACKAGES environment variable...
set NUGET_PACKAGES=%CD%\packages
echo NUGET_PACKAGES=%NUGET_PACKAGES%

echo.
echo [Step 3/6] Initializing build environment (ARM64EC Release)...
call init.cmd arm64ecfre

echo.
echo [Step 4/6] Verifying environment variables...
echo   Platform=%Platform%
echo   Configuration=%Configuration%
echo   BUILDPLATFORM=%BUILDPLATFORM%
echo   _BuildArch=%_BuildArch%
echo   _BuildType=%_BuildType%
echo   NUGET_PACKAGES=%NUGET_PACKAGES%

echo.
echo [Step 5/6] Generating mock WindowsAppSDK package...
call pack.cmd /version 3.0.0-dev
if errorlevel 1 (
    echo WARNING: pack.cmd failed, Build.cmd will attempt to regenerate it
)

echo.
echo [Step 6/6] Starting build...
echo.
Build.cmd prodtest /version 3.0.0-dev
```

By following this script in CMD, I successfully packaged and obtained the binlog for samples. However, I think modifying WindowsAppSDK (foundation) and then testing might still be a challenging process. Based on pipeline experience, it requires first obtaining the two foundation packages and the commit with changes, then replacing the corresponding code. For now, I haven’t thought of a better alternative.

Scottj1s:
We're still getting unnecessary batching between WindowsAppRuntimeAutoInitializer.cpp and UndockedRegFreeWinRT-AutoInitializer.cpp, which are being ClCompile Included separately.  We could fold inclusion together, since all preprocessor definitions are now set globally.  But better still would be for WindowsAppRuntimeAutoInitializer.cpp to conditionally #include UndockedRegFreeWinRT-AutoInitializer.cpp directly and we could just remove WindowsAppSDK-Nuget-Native.UndockedRegFreeWinRT.targets altogether.  The initial factoring of WindowsAppRuntimeAutoInitializer.cpp was to have a single entry point for deterministically ordered auto-initialization logic.  Compiling in the undocked reg-free source would complete that.

mschofie:
> We're still getting unnecessary batching between WindowsAppRuntimeAutoInitializer.cpp and UndockedRegFreeWinRT-AutoInitializer.cpp, which are being ClCompile Included separately. We could fold inclusion together, since all preprocessor definitions are now set globally. But better still would be for WindowsAppRuntimeAutoInitializer.cpp to conditionally #include UndockedRegFreeWinRT-AutoInitializer.cpp directly and we could just remove WindowsAppSDK-Nuget-Native.UndockedRegFreeWinRT.targets altogether. The initial factoring of WindowsAppRuntimeAutoInitializer.cpp was to have a single entry point for deterministically ordered auto-initialization logic. Compiling in the undocked reg-free source would complete that.

That seems like quite a change, and `#include`'ing one .cpp from another makes me twitch a little. Perhaps I just expand this pattern to all similar use cases?

Scottj1s:
> That seems like quite a change, and `#include`'ing one .cpp from another makes me twitch a little. Perhaps I just expand this pattern to all similar use cases?

Including one cpp in another is not a concern - we do this in C++/WinRT and both sources here are under our control.  But yes, I'm also fine with folding all ClCompile Includes together.

Scottj1s:
> > That seems like quite a change, and `#include`'ing one .cpp from another makes me twitch a little. Perhaps I just expand this pattern to all similar use cases?
> 
> Including one cpp in another is not a concern - [we do this in C++/WinRT](https://github.com/microsoft/cppwinrt/blob/dcaa98c8d2abeb47ab5ce9ef1f96171e93d0ff9b/vsix/ProjectTemplates/VC/Windows%20Universal/StaticLibrary/Class.cpp#L3) and here, both sources are under our control. But yes, I'm also fine with folding all ClCompile Includes together.

PetrMinar:
Any update on this? The "duplicate 'WindowsAppRuntimeAutoInitializer.cpp' compilation" in almost every build is mega annoying.... And I still see it in WinAppSDK 2.0 Experimental 4.

qiutongMS:
> Any update on this? The "duplicate 'WindowsAppRuntimeAutoInitializer.cpp' compilation" in almost every build is mega annoying.... And I still see it in WinAppSDK 2.0 Experimental 4.

Thanks for the ping — we’re still tracking this. There’s a proposed change, but we had to pause rollout because it can cause failures in certain edge scenarios . We’re working on a safer approach.

apricopbd:
Hello! Any estimate as to when a new version of WinUI will include this fix (or a safer approach as mentioned)? Thank you!

mschofie:
#6161 is a superset of this fix - and has been merged into main! Huzzah!

PR #5903, opened 2025-10-14
Adding headers to test common header files

guimafelipe:
This pull request updates several test header files to improve consistency and ensure all necessary dependencies are explicitly included. The main focus is on adding missing system and library includes, which helps prevent build issues and makes dependencies clearer for future maintenance.

Dependency inclusion improvements:

* Added missing includes such as `<windows.h>`, `<winrt/base.h>`, `<WexTestClass.h>`, `<appmodel.h>`, `<Verify.h>`, and `<filesystem>` to multiple test headers including `TestDef.h`, `WindowsAppRuntime.Test.AppModel.h`, `WindowsAppRuntime.Test.Bootstrap.h`, `WindowsAppRuntime.Test.FileSystem.h`, `WindowsAppRuntime.Test.Package.h`, `WindowsAppRuntime.Test.TAEF.cppwinrt.h`, and `WindowsAppRuntime.Test.TAEF.h`. [[1]](diffhunk://#diff-d09924480bae94116b464963f7c1d9e156b9cbd7e1f778b73ba68c6a80c3ca0fL1-R6) [[2]](diffhunk://#diff-9245541af5d95ba2baebb88fda8ec2883f25176fe8b4e6e3c8a00e032ed3a496L1-L8) [[3]](diffhunk://#diff-f2017db45a841820e7abb3f78a2b6a6f92a5ea000a2742c416249943c7cfb7b2R7) [[4]](diffhunk://#diff-fa6555b69b354073885fab14e59ef8959a26f10c9b471d6e1a6aaf3bbe4f91fdL1-R9) [[5]](diffhunk://#diff-697e01136f05fb45edad0e81aa730980d70a655294e30a90ca687196ac79fc3fR7) [[6]](diffhunk://#diff-b4b9c441905de861b099f2ffdd06cf2a3e2f1730c6b6f33ed38421104ecfb99dL1-R10) [[7]](diffhunk://#diff-ad881a9af948cdfb9e03162761e58741e7246f79e89a45b79275c6efff491c4aL1-R10)

These changes help ensuring that each header is independent and handle its own dependencies. Without these explicit dependencies, when creating a new test file, we need to know precisely the order in which to include each header for them to work, as some headers are dependent on external includes to exist. Or use one of the `.pch` files as example, but then making it harder to maintain as the dependencies are hidden and we end up including unnecessary files.
This is also stated on [C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-contained).

WARNING:


test/inc/TestDef.h
```diff
@@ -1,6 +1,9 @@
-﻿// Copyright (c) Microsoft Corporation and Contributors.
+// Copyright (c) Microsoft Corporation and Contributors.
 // Licensed under the MIT License.
 #pragma once
+#include <string>
+#include <winrt/base.h>
+#include <windows.h>
```
DrusTheAxe: This is backwards

Headers should be included from most general to most specific -- OS, language, libraries, project.

`windows.h` *after* `winrt`? That's backwards. Likewise, `string` being a standard library header should come after `windows.h` before before C++/WinRT headers

RECOMMEND: 
```
#include <windows.h>
#include <string>
#include <winrt/base.h>
```

test/inc/WindowsAppRuntime.Test.AppModel.h
```diff
@@ -1,11 +1,11 @@
-﻿// Copyright (c) Microsoft Corporation and Contributors.
+// Copyright (c) Microsoft Corporation and Contributors.
 // Licensed under the MIT License.
 
 #ifndef __WINDOWSAPPRUNTIME_TEST_APPMODEL_H
 #define __WINDOWSAPPRUNTIME_TEST_APPMODEL_H
 
+#include <windows.h>
```
DrusTheAxe: Why?

Is this currently a problem or is this a stylistic change?

This is the most obvious example - why does this need to include `windows.h`? That's routinely the 1st include in any precompiled header, and in any source file in projects not using precompiled headers. Also, some include `<unknown.h>` rather than `windows.h` (the former includes the latter, plus other things).

I don't see the value including `windows.h` in any header file (except precomp.h/pch.h)

test/inc/WindowsAppRuntime.Test.TAEF.cppwinrt.h
```diff
@@ -1,9 +1,13 @@
-﻿// Copyright (c) Microsoft Corporation and Contributors. All rights reserved.
+// Copyright (c) Microsoft Corporation and Contributors. All rights reserved.
 // Licensed under the MIT License.
 
 #ifndef __WINDOWSAPPRUNTIME_TEST_TAEF_CPPWINRT_H
 #define __WINDOWSAPPRUNTIME_TEST_TAEF_CPPWINRT_H
 
+#include <WexTestClass.h>
+#include <winrt/base.h>
+#include <windows.h>
```
DrusTheAxe, line 7: Order's backwards. General ordering would be

* windows
* C++/WinRT
* TAEF

(assuming any of this is desirable)

test/inc/WindowsAppRuntime.Test.TAEF.h
```diff
@@ -1,9 +1,13 @@
-﻿// Copyright (c) Microsoft Corporation and Contributors. All rights reserved.
+// Copyright (c) Microsoft Corporation and Contributors. All rights reserved.
 // Licensed under the MIT License.
 
 #ifndef __WINDOWSAPPRUNTIME_TEST_TAEF_H
 #define __WINDOWSAPPRUNTIME_TEST_TAEF_H
 
+#include <WexTestClass.h>
+#include <filesystem>
+#include <windows.h>
```
DrusTheAxe, line 7: More ordering issues, and assuming any of this is apropos to start with

DrusTheAxe (changes requested):
I don't follow the reason for most of these changes. See comment for more detail

test/inc/WindowsAppRuntime.Test.AppModel.h
guimafelipe: The current problem is that for some of these headers, you can't include them alone without including these dependencies before in your file. Creating a new test that uses one of these headers requires knowing what they depend on, and it is not possible to just include the header that you actually want to use alone, requiring the include of `windows.h` before including the desire header for example.
I think it is a good practice to have headers that work as their own (by being possible to compile a `.cpp` file including only it, which is what I mean when I say a header can be compiled) and it would make it easier for others to use them within the project.

For this specific case, this header can be compile on its own because it includes `windows.h` within `appmodel.h`, but it is hidden under a conditional macro. So I included it here to make it safer.

Do you think there are any downsides on doing this change? We can close this PR if it is not desired.

ssparach (approved):

test/inc/WindowsAppRuntime.Test.AppModel.h
DrusTheAxe: There may be some merits to parts of it, but as is it's too broad a change e.g. don't add `windows.h` to any header file (e.g. cause complications if someone uses `unknown.h`). There's a hard balance here as there's also the problem where includes can intersect in...interesting ways, if not identical includes and order, and headers including headers make any removal of includes problematic as the header evolves e.g. a.h includes b.h because it need something, but later a.h is split into a.h and x.h and now b.h isn't needed in a.h...but someone who included a.h assumed b.h's were available and now you create downstream build breaks, even though your code using a.h works fine TL;DR easy to add includes , hard to remove.

I'd be more receptive to a PR making changes specifically because a new test was being stood up and it showed build errors because of some assumed includes, in which case (perhaps) add (some?) of those in these headers.

guimafelipe: > a.h includes b.h because it need something, but later a.h is split into a.h and x.h and now b.h isn't needed in a.h...but someone who included a.h assumed b.h's were available and now you create downstream build breaks

Isn't this desirable? For a simple example, we can have `a.h` including `<string>`:
```c++
// a.h
#pragma once
#include <string>
```

And then, another header includes `a.h` and uses the `<string>` that is exposed by it:

```c++
// some-header.h
#pragma once
#include "a.h"

std::wstring GetName();
```

If suddenly `a.h` removes the `#include <string>`, the files that only includes `some-header.h` and not `<string>` will break. But that break will be at compile time and will be correct in my opinion because `some-header.h` **should** be including `<string>` as it is a direct dependency. So this build break will force a change in the correct direction, stopping a usage of a indirect include that was not clear.

test/inc/WindowsAppRuntime.Test.TAEF.cppwinrt.h
```diff
@@ -1,9 +1,13 @@
-﻿// Copyright (c) Microsoft Corporation and Contributors. All rights reserved.
+// Copyright (c) Microsoft Corporation and Contributors. All rights reserved.
```
DrusTheAxe, line 1: Why do these copyright lines show up as changed? LF vs CRLF?

test/inc/WindowsAppRuntime.Test.Bootstrap.h
```diff
@@ -4,6 +4,7 @@
 #ifndef __WINDOWSAPPRUNTIME_TEST_BOOTSTRAP_H
 #define __WINDOWSAPPRUNTIME_TEST_BOOTSTRAP_H
 
+#include <windows.h>
```
DrusTheAxe, line 7: Don't include windows.h. As mentioned that's not always the right answer, and it's such a universal pervasive thing it's not helpful here and can be harmful

test/inc/WindowsAppRuntime.Test.FileSystem.h
```diff
@@ -1,10 +1,12 @@
-﻿// Copyright (c) Microsoft Corporation and Contributors.
+// Copyright (c) Microsoft Corporation and Contributors.
 // Licensed under the MIT License.
 
 #ifndef __WINDOWSAPPRUNTIME_TEST_FILESYSTEM_H
 #define __WINDOWSAPPRUNTIME_TEST_FILESYSTEM_H
 
+#include <windows.h>
 #include <filesystem>
+#include <verify.h>
```
DrusTheAxe, line 9: What's verify.h? TAEF?

test/inc/WindowsAppRuntime.Test.TAEF.h
```diff
@@ -1,9 +1,13 @@
-﻿// Copyright (c) Microsoft Corporation and Contributors. All rights reserved.
+// Copyright (c) Microsoft Corporation and Contributors. All rights reserved.
 // Licensed under the MIT License.
 
 #ifndef __WINDOWSAPPRUNTIME_TEST_TAEF_H
 #define __WINDOWSAPPRUNTIME_TEST_TAEF_H
 
+#include <windows.h>
+#include <filesystem>
+#include <WexTestClass.h>
```
DrusTheAxe: Why WexTestClass.h? I don't see any test classes here

test/inc/WindowsAppRuntime.Test.TAEF.h
```diff
@@ -1,9 +1,13 @@
-﻿// Copyright (c) Microsoft Corporation and Contributors. All rights reserved.
+// Copyright (c) Microsoft Corporation and Contributors. All rights reserved.
 // Licensed under the MIT License.
 
 #ifndef __WINDOWSAPPRUNTIME_TEST_TAEF_H
 #define __WINDOWSAPPRUNTIME_TEST_TAEF_H
 
+#include <windows.h>
+#include <filesystem>
```
DrusTheAxe, line 8: Agreed, filesystem is a good include here

test/inc/WindowsAppRuntime.Test.TAEF.cppwinrt.h
```diff
@@ -1,9 +1,13 @@
...
 // Licensed under the MIT License.
 
 #ifndef __WINDOWSAPPRUNTIME_TEST_TAEF_CPPWINRT_H
 #define __WINDOWSAPPRUNTIME_TEST_TAEF_CPPWINRT_H
 
+#include <windows.h>
+#include <winrt/base.h>
+#include <WexTestClass.h>
```
DrusTheAxe: Why WexTestClass.h? See other comment

test/inc/WindowsAppRuntime.Test.TAEF.cppwinrt.h
```diff
@@ -1,9 +1,13 @@
-﻿// Copyright (c) Microsoft Corporation and Contributors. All rights reserved.
+// Copyright (c) Microsoft Corporation and Contributors. All rights reserved.
 // Licensed under the MIT License.
 
 #ifndef __WINDOWSAPPRUNTIME_TEST_TAEF_CPPWINRT_H
 #define __WINDOWSAPPRUNTIME_TEST_TAEF_CPPWINRT_H
 
+#include <windows.h>
+#include <winrt/base.h>
```
DrusTheAxe, line 8: This leaves me uneasy for reasons I can't quite articulate at the moment (perhaps not enough coffee?)

Aha. Found it. The source of my uneasiness...

Is base.h a well defined part of C++/WinRT's API or just an implementation detail? e.g. https://learn.microsoft.com/en-us/uwp/cpp-ref-for-winrt/hstring but nowhere on that page does it say the header(s) you need to include (unlike, say, CreateFileW docs say [what header (+lib) is needed](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew#requirements)), nor is it mentioned in any example.

I know about this header, but I've pored over C++/WinRT's generated headers and implementation in more detail than most. Is `winrt/base.h` part of the official API or an implementation detail? If the former it should be mentioned in the docs, but it's not. If the latter it's probably not something we should directly reference.

test/inc/WindowsAppRuntime.Test.AppModel.h
ChrisGuzak: [this is per C++ core guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#rs-contained) and [this too](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf10-avoid-dependencies-on-implicitly-included-names)

if you can include less than `windows.h` please do

test/inc/WindowsAppRuntime.Test.FileSystem.h
guimafelipe: Yes

test/inc/WindowsAppRuntime.Test.TAEF.cppwinrt.h
guimafelipe: Yes. That is been happening in a lot of PRs, not only from me.

test/inc/WindowsAppRuntime.Test.TAEF.h
guimafelipe: The types that come from `Wex::Common` namespace comes from `WesDebug.h`. I also think it would be better to change to the specific file that declares it.

test/inc/WindowsAppRuntime.Test.TAEF.cppwinrt.h
guimafelipe: Same as previous comment, `Wex::Common` namespace and its usages. But I agree it is better to be specific here and include the specific header where the types were defined, like `WexDebug.h` or `WexString.h`.

test/inc/WindowsAppRuntime.Test.TAEF.cppwinrt.h
guimafelipe: The first mention of it I have is [here](https://learn.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/get-started#the-precompiled-header), where it shows how to include `winrt/base.h` in the pre-compiled header. So I assume it is public and intended to be included.

test/inc/WindowsAppRuntime.Test.AppModel.h
guimafelipe: The thing is that there headers uses type definitions and functions that comes from header internal from `windows.h`. And these internal headers are dependent on the macros defined in `windows.h` to function.

For example, here on `WindowsAppRuntime.Test.AppModel.h`, the type `UINT32` is defined in `basetsd.h` Which behaves well if included directly.

But on `WindowsAppRuntime.Test.TAEF.h` for example, we use the `PCWSTR` and `WCHAR` types, which are both defined in `winnt.h`. This header depends of an architecture (something like `_AMD64_`) being defined, or it errors with `#error "No Target Architecture"`. And this is defined by the `windows.h` header. The same would happen with we just include the `minwindef.h` header that internally includes `winnt.h`.

I went to the `windows.h` header here because that is what essentially being included in the `pch.h` file for the compilation of the tests alongside with `unknwn.h` in some of them (which internally includes `windows.h` on line 29).
Some of the tests include `windows.h` right after a `#define WIN32_LEAN_AND_MEAN` to exclude rerely-used stuff, which can also help to keep things small there.

Any thoughts or suggestions?
Thank you.

DrusTheAxe: @ChrisGuzak do those guidelines also apply to windows.h? I'd generally find it extremely-weird to see code #include'ing basetsd.h, minwindef.h or other parts of windows.h. And there doesn't seem a lot of merit #include'ing windows.h in headers given nearly every .h uses basic constructs from windows.h (PCWSTR, UINT32, DWORD, ...). Perhaps this is worthy of discussion at our next API Design meeting?

> Some of the tests include windows.h right after a #define WIN32_LEAN_AND_MEAN to exclude rerely-used stuff

@guimafelipe Any such definitions should occur in pch.h before any #include (pretty much the only thing I'd expect to see in pch.h between line1 //Copyright... and line2+ #include's)

TBH I never heard of `unknown.h` before working in WinAppSDK 0.x, nor would I have expected the weird build errors mixing unknown.h vs windows.h if I it didn't draw blood on multiple occasions :-( Could be a dated problem buit this isn't the sort of stuff that changes often or quickly, so I'm hesitant to slather windows.h everywhere. Not without a better understanding of the merit (like, discussion at our next API meeting).

PR #5921, opened 2025-10-17
[Build] Re-enable testing on arm64 by default for Foundation Nightly and Official pipelines

alexlamtest:
Applying changes in "Pull Request: [Build] Re-enable testing on arm64 by default for Agg Nightly and Official pipelines" against the Agg repo to the Foundation repo. The Foundation PR pipeline continue to have testing on arm64 disabled by default, but developers can enable it manually at pipeline queueing time.

--

WARNING:

### DIFF
--- a/build/AzurePipelinesTemplates/WindowsAppSDK-Test-Stage.yml
+++ b/build/AzurePipelinesTemplates/WindowsAppSDK-Test-Stage.yml
@@ -1,7 +1,7 @@
 parameters:
 - name: "TestOnArm64"
   type: boolean
-  default: false
+  default: true
 
 stages:
 - stage: Test_x86


codendone (approved):

build/AzurePipelinesTemplates/WindowsAppSDK-Test-Stage.yml
```diff
@@ -1,3 +1,8 @@
+parameters:
+- name: "TestOnArm64"
+  type: boolean
+  default: false
```
codendone: should this be `true`?

alexlamtest: false is still a good default value at least in case of a PR validation pipeline run.
But sure, happy to set it to true so that we don't need to touch this again when some day we enable arm64 testing by default again for PR validation pipeline runs.

PR #5928, opened 2025-10-20
Changing include on deployment/PackageDefinitions.h

guimafelipe:
This pull request introduces some minor improvements to the `dev/Deployment/PackageDefinitions.h` header file, removing the `pch.h` include and explicitly including `<string>` and `<map>`, both of which are used in this file.

This change is good because when we include this header on different contexts (for example, for building the unit tests), the `pch.h` file will not be [the one intended](https://github.com/microsoft/WindowsAppSDK/blob/main/dev/WindowsAppRuntime_DLL/pch.h) to be. This will be mistaken by the test's `pch.h` file which can create conflicts.

A good practice would be for header files not to depend on `pch.h` files, and only let the `.cpp` files to depend on them.

WARNING:

### DIFF
--- a/dev/Deployment/PackageDefinitions.h
+++ b/dev/Deployment/PackageDefinitions.h
@@ -1,6 +1,11 @@
-﻿// Copyright (c) Microsoft Corporation and Contributors.
+// Copyright (c) Microsoft Corporation and Contributors.
 // Licensed under the MIT License.
-#include <pch.h>
+
+#pragma once
+
+#include <map>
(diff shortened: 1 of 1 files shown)


guimafelipe:
> Nice catch!
> 
> Yeah precompiled headers should only be included by source files, never other headers.
> 
> Is this the only instance in the repo?

Only one I have seen so far. I will take some time this week to see if I find it in other parts of the repo as I've been mostly focused on deployment.

PR #5945, opened 2025-10-27
Change Nuget version format

Dreynor87:
Nuget will now follow:
M.m.p<-exp>+DateRev

WARNING:

### DIFF
--- a/build/AzurePipelinesTemplates/WindowsAppSDK-PackTransportPackage-Stage.yml
+++ b/build/AzurePipelinesTemplates/WindowsAppSDK-PackTransportPackage-Stage.yml
@@ -218,11 +218,11 @@ stages:
             $formattedTag = ''
             if(-not [String]::IsNullOrEmpty($versionTag))
             {
-              $formattedTag = '-' + $versionTag + "-" + $(versionMinDate) + $paddedRevision
+              $formattedTag = '-' + $versionTag + "+" + $(versionMinDate) + $paddedRevision
             }
             else
             {
-              $formattedTag = "-" + $(versionMinDate) + $paddedRevision
+              $formattedTag = "+" + $(versionMinDate) + $paddedRevision
             }
             Write-Host "Using Release Versioning"
             $version = $majorMinorPatchRev + $formattedTag


build/AzurePipelinesTemplates/WindowsAppSDK-PackTransportPackage-Stage.yml
```diff
@@ -219,7 +218,11 @@ stages:
             $formattedTag = ''
             if(-not [String]::IsNullOrEmpty($versionTag))
             {
-              $formattedTag = '-' + $versionTag
+              $formattedTag = '-' + $versionTag + "-" + $(versionMinDate) + $paddedRevision
```
codendone: This is not SemVer compliant.

Dreynor87: Including the Date and Revision for the release one, or in all cases? Re-reading semver.org, if I change that to a "+" does that work? I switched because I thought that M.m.p+DateRev was not ok for the vpack code.

codendone: Yes, switching to "+" should address the SemVer issue. I agree that vpack limitations also need to be considered (I'm not familiar enough there to know whether there is a problem).

I'm still curious if this is a temporary change or intended long-term. If the latter, is there a spec to review?

Dreynor87: I do intend this to be long term. I am adding the date and revision in case we need to rebuild something internal (as we have seen multiple times trying to get 2.0.0-exp2 out the door).  We dont want to have to bump everything another patch because of a problem with a single metapackage. I have not put this recent-ish design change into a spec yet.

codendone: I'm not excited about this as the long-term design. I strongly recommend a spec and reviewing that before implementing something across all components. Differences from the old 2.0 spec would be especially good to discuss.

DrusTheAxe: +stuff = build info, which is entirely informational. Whether and how this should be valued is hard to say given the lack of spec. We need spec, before doing any implementation.

RECOMMEND: Create `specs/SemanticVersion.md` and spell out how semver is defined/used (and of course follow up with spec review :-)


FYI per https://semver.org
```
<valid semver> ::= <version core>
                 | <version core> "-" <pre-release>
                 | <version core> "+" <build>
                 | <version core> "-" <pre-release> "+" <build>
```

Issue #5827, opened 2025-09-12
Order of FileTypeChoices in new pickers is not respected


Marv51:
### Describe the bug

In Windows.Storage.Pickers.FileSavePicker under FileTypeChoices it says:

> If you add multiple FileTypeChoices to the FileSavePicker, the first one added is the default file type. This default will be selected when the user opens the file picker.

https://learn.microsoft.com/en-us/uwp/api/windows.storage.pickers.filesavepicker.filetypechoices?view=winrt-26100

In the new Microsoft.Windows.Storage.Pickers.FileSavePicker there seems to be no way to influence the order of FileTypeChoices as the input order is not the displayed order. The order seems to be always alphabetically, with the alphabetically-first being selected by default.

I thought setting `DefaultFileExtension` would set the default FileTypeChoice. But it does not. It only seems to work when no FileTypeChoices are added.

### Steps to reproduce the bug

```csharp
// using Microsoft.Windows.Storage.Pickers;

var savePicker = new FileSavePicker(this.AppWindow.Id)
{
    SuggestedStartLocation = PickerLocationId.DocumentsLibrary,
    SuggestedFileName = "MyFile"
};

savePicker.DefaultFileExtension = ".ods";

savePicker.FileTypeChoices.Add("PDF", [".pdf"]);
savePicker.FileTypeChoices.Add("ABC", [".txt"]);
savePicker.FileTypeChoices.Add("Tabellenformat", [".ods"]);

await savePicker.PickSaveFileAsync();
```

### Expected behavior

Same behavior as WinRT API where the order is respected and the first is the default. Or a way to set a default FileTypeChoice manually (see also IFileDialog::SetFileTypeIndex)

### NuGet package version

Windows App SDK 1.8.0: 1.8.250907003

### Packaging type

Packaged (MSIX)

### Windows version

Windows 11 version 24H2 (26100, June 2025 Update)

### IDE

Visual Studio 2022

beeradmoore:
I dont think this ever worked (either that or my code was broken).

In DLSS Swapper [I use Win32 generated stuff](https://github.com/beeradmoore/dlss-swapper/blob/main/src/Helpers/FileSystemHelper.cs#L155-L159) and it didn't seem to work. I just tested trying to set it to `csv` and `.csv` but when I open it it is always `.json` which is the first in my list. This is the same for my open and save dialog code.

tipa:
[...]

PR #5948, opened 2025-10-28
Fix FileTypeChoices in Storage Pickers to preserve insertion order

DinahK-2SO:
**Description**

This is a fix for issue:
- #5827

The existing FileTypeChoices property was built on an unordered map, which does not preserve the user-defined order.

However, in the FileOpenPicker and FileSavePicker, the FileTypeChoices need to be displayed in the order they were inserted, rather than in a random order. 

This is important for developers' and end-users' experience because:

- Developers expect to see their file type options in the logical order they added them
- The first added option is usually the default selection
- Maintaining consistent ordering helps end users quickly locate the desired file type

Additionaly, the insertion order is respected in the legacy UWP FileSavePicker.

**Fix**

This is a backward-compatible fix. The goal of this fix is to maintain the existing Map type API contract and its good performance while ensuring that the display order of FileTypeChoices meets expectations.

This pull request refactors the implementation of the `FileTypeChoicesMap` to ensure that the insertion order of keys is preserved and provides efficient key lookups. It replaces the previous unordered map-based implementation with the implementation backed by a vector. The update also introduces custom iterator and view classes to support this ordered behavior and modifies related tests to verify the new insertion order.

-------------------------------------------------------------------

Code change details:

**Ordered map implementation and API changes:**

* Replaced the internal storage of `FileTypeChoicesMap` from an unordered map to a `std::vector` of key-value pairs (`FileTypeChoiceVector`), managed by a `shared_ptr` to preserve insertion order and support multiple views/iterators. [[1]](diffhunk://#diff-e3984d20285d293864860b889081b68d4873694f8649c2e440e359de0d9fc036L31-R83) [[2]](diffhunk://#diff-6bc0380f2bf26bb3b779e3954112b4e1a2c7ad63ff4ec2cebcbdaf0516ef2079R8-R35)
* Added custom iterator (`OrderedMapIterator`) and view (`OrderedMapView`) classes to enable ordered iteration and lookup, updating all methods (`Insert`, `Lookup`, `Remove`, `Clear`, `First`, etc.) to operate on the new ordered vector structure. [[1]](diffhunk://#diff-e3984d20285d293864860b889081b68d4873694f8649c2e440e359de0d9fc036L31-R83) [[2]](diffhunk://#diff-6bc0380f2bf26bb3b779e3954112b4e1a2c7ad63ff4ec2cebcbdaf0516ef2079L34-R222)

**API surface and helper improvements:**

* Updated method signatures to add `noexcept` where appropriate and refactored helper methods for key lookup using `std::ranges::find`. [[1]](diffhunk://#diff-e3984d20285d293864860b889081b68d4873694f8649c2e440e359de0d9fc036L20-R30) [[2]](diffhunk://#diff-6bc0380f2bf26bb3b779e3954112b4e1a2c7ad63ff4ec2cebcbdaf0516ef2079R8-R35)
* Improved memory management and performance by pre-allocating vector capacity and using shared ownership for views and iterators.

**Test updates for ordered behavior:**

* Modified picker tests to insert file type choices in a specific order and updated assertions to verify that the resulting filters maintain the insertion order, ensuring the new implementation is covered by tests. [[1]](diffhunk://#diff-c0106cdb512c855fd3f2a703625b8ea80fe7faf9dc3db1d74149691ef7b945c5L217-R222) [[2]](diffhunk://#diff-c0106cdb512c855fd3f2a703625b8ea80fe7faf9dc3db1d74149691ef7b945c5L229-R241) [[3]](diffhunk://#diff-c0106cdb512c855fd3f2a703625b8ea80fe7faf9dc3db1d74149691ef7b945c5L245-R255)

### DIFF
--- a/dev/Interop/StoragePickers/FileTypeChoicesMap.cpp
+++ b/dev/Interop/StoragePickers/FileTypeChoicesMap.cpp
@@ -116,8 +116,8 @@ namespace winrt::Microsoft::Windows::Storage::Pickers::implementation
     }
 
     // OrderedMapIterator implementation
-    OrderedMapIterator::OrderedMapIterator(std::shared_ptr<FileTypeChoiceVector const> map)
-        : m_map(std::move(map)), m_current(0)
+    OrderedMapIterator::OrderedMapIterator(std::shared_ptr<FileTypeChoiceVector const> const& map)
+        : m_map(map), m_current(0)
     {
     }
 
@@ -162,8 +162,8 @@ namespace winrt::Microsoft::Windows::Storage::Pickers::implementation
     }
 
     // OrderedMapView implementation
-    OrderedMapView::OrderedMapView(std::shared_ptr<FileTypeChoiceVector const> map)
-        : m_map(std::move(map))
+    OrderedMapView::OrderedMapView(std::shared_ptr<FileTypeChoiceVector const> const& map)
+        : m_map(map)
     {
     }
 
--- a/dev/Interop/StoragePickers/FileTypeChoicesMap.h
+++ b/dev/Interop/StoragePickers/FileTypeChoicesMap.h
@@ -49,7 +49,7 @@ namespace winrt::Microsoft::Windows::Storage::Pickers::implementation
     struct OrderedMapIterator : implements<OrderedMapIterator,
         winrt::Windows::Foundation::Collections::IIterator<winrt::Windows::Foundation::Collections::IKeyValuePair<hstring, winrt::Windows::Foundation::Collections::IVector<hstring>>>>
     {
-        OrderedMapIterator(std::shared_ptr<FileTypeChoiceVector const> map);
+        OrderedMapIterator(std::shared_ptr<FileTypeChoiceVector const> const& map);
 
         winrt::Windows::Foundation::Collections::IKeyValuePair<hstring, winrt::Windows::Foundation::Collections::IVector<hstring>> Current() const;
         bool HasCurrent() const noexcept;
@@ -66,7 +66,7 @@ namespace winrt::Microsoft::Windows::Storage::Pickers::implementation
         winrt::Windows::Foundation::Collections::IMapView<hstring, winrt::Windows::Foundation::Collections::IVector<hstring>>,
         winrt::Windows::Foundation::Collections::IIterable<winrt::Windows::Foundation::Collections::IKeyValuePair<hstring, winrt::Windows::Foundation::Collections::IVector<hstring>>>>
     {
-        OrderedMapView(std::shared_ptr<FileTypeChoiceVector const> map);
+        OrderedMapView(std::shared_ptr<FileTypeChoiceVector const> const& map);
 
         // IMapView
         winrt::Windows::Foundation::Collections::IVector<hstring> Lookup(hstring const& key) const;


dev/Interop/StoragePickers/FileTypeChoicesMap.h
```diff
@@ -28,7 +34,55 @@ namespace winrt::Microsoft::Windows::Storage::Pickers::implementation
...
     private:
-        winrt::Windows::Foundation::Collections::IMap<hstring, winrt::Windows::Foundation::Collections::IVector<hstring>> m_innerMap{
-            single_threaded_map<hstring, winrt::Windows::Foundation::Collections::IVector<hstring>>() };
+        // Use vector to maintain insertion order + hash map for O(1) lookup
+        std::vector<std::pair<hstring, winrt::Windows::Foundation::Collections::IVector<hstring>>> m_orderedMap;
+        std::unordered_map<std::wstring, size_t> m_keyToIndex; // Maps key to index in m_orderedMap
+
+        // Helper methods
+        auto FindKey(hstring const& key) const;
+        auto FindKey(hstring const& key);
```
codendone: Why not only have the const version?

dev/Interop/StoragePickers/FileTypeChoicesMap.cpp
```diff
@@ -5,11 +5,34 @@
...
     {
+        // Pre-allocate capacity for typical file type choices (usually 2-8 items)
+        m_orderedMap.reserve(8);
+        m_keyToIndex.reserve(8);
+    }
+
+    size_t FileTypeChoicesMap::FindKeyIndex(hstring const& key) const
+    {
+        auto it = m_keyToIndex.find(std::wstring(key.c_str()));
+        return (it != m_keyToIndex.end()) ? it->second : SIZE_MAX;
```
Scottj1s: minor clarification suggestion:
 constexpr size_t element_not_found{ SIZE_MAX };

dev/Interop/StoragePickers/FileTypeChoicesMap.cpp
```diff
@@ -30,42 +53,160 @@ namespace winrt::Microsoft::Windows::Storage::Pickers::implementation
...
     bool FileTypeChoicesMap::HasKey(hstring const& key) const
     {
-        return m_innerMap.HasKey(key);
+        return FindKey(key) != m_orderedMap.end();
     }
 
     winrt::Windows::Foundation::Collections::IMapView<hstring, winrt::Windows::Foundation::Collections::IVector<hstring>> FileTypeChoicesMap::GetView() const
     {
-        return m_innerMap.GetView();
+        return make<OrderedMapView>(m_orderedMap);
```
Scottj1s, line 95: If caller holds GetView return value past FileTypeChoicesMap destruction, you'll have a dangling pointer crash.  Consider std::make_shared on m_orderedMap, winrt::make_weak on FileTypeChoicesMap, etc.

dev/Interop/StoragePickers/FileTypeChoicesMap.cpp
```diff
@@ -30,42 +53,160 @@ namespace winrt::Microsoft::Windows::Storage::Pickers::implementation
...
     }
 
     void FileTypeChoicesMap::Remove(hstring const& key)
     {
-        m_innerMap.Remove(key);
+        size_t index = FindKeyIndex(key);
```
Scottj1s: Non-atomic reading and/or writing of m_orderedMap & m_keyToIndex - consider a mutex.  Same for nearly all other methods.

dev/Interop/StoragePickers/FileTypeChoicesMap.cpp
```diff
@@ -30,42 +53,160 @@ namespace winrt::Microsoft::Windows::Storage::Pickers::implementation
...
+                              winrt::Windows::Foundation::Collections::IMapView<hstring, winrt::Windows::Foundation::Collections::IVector<hstring>>& second) const
+    {
+        // For simplicity, not implementing split
+        first = nullptr;
+        second = nullptr;
+    }
+
+    winrt::Windows::Foundation::Collections::IIterator<winrt::Windows::Foundation::Collections::IKeyValuePair<hstring, winrt::Windows::Foundation::Collections::IVector<hstring>>> OrderedMapView::First() const
+    {
+        return make<OrderedMapIterator>(m_map);
```
Scottj1s, line 222: another dangling pointer concern

dev/Interop/StoragePickers/FileTypeChoicesMap.cpp
```diff
@@ -30,42 +53,160 @@ namespace winrt::Microsoft::Windows::Storage::Pickers::implementation
...
+        return static_cast<uint32_t>(m_map.size());
+    }
+
+    bool OrderedMapView::HasKey(hstring const& key) const
+    {
+        return FindKey(key) != m_map.end();
+    }
+
+    void OrderedMapView::Split(winrt::Windows::Foundation::Collections::IMapView<hstring, winrt::Windows::Foundation::Collections::IVector<hstring>>& first,
+                              winrt::Windows::Foundation::Collections::IMapView<hstring, winrt::Windows::Foundation::Collections::IVector<hstring>>& second) const
```
Scottj1s: noexcept

test/StoragePickersTests/PickerCommonTests.cpp
```diff
@@ -214,10 +214,12 @@ namespace Test::PickerCommonTests
             winrt::Microsoft::UI::WindowId windowId{};
             winrt::Microsoft::Windows::Storage::Pickers::FileOpenPicker picker(windowId);
 
-            picker.FileTypeChoices().Insert(
-                L"Documents", winrt::single_threaded_vector<winrt::hstring>({ L".txt", L".doc", L".docx" }));
             picker.FileTypeChoices().Insert(
                 L"Pictures", winrt::single_threaded_vector<winrt::hstring>({ L".png", L".jpg", L".jpeg", L".bmp" }));
+            picker.FileTypeChoices().Insert(
```
Scottj1s, line 219: add test case to exercise reuse of existing index

dev/Interop/StoragePickers/FileTypeChoicesMap.cpp
```diff
@@ -30,42 +53,160 @@ namespace winrt::Microsoft::Windows::Storage::Pickers::implementation
...
+    {
+        auto it = FindKey(key);
+        if (it != m_map.end())
+        {
+            return it->second;
+        }
+        throw winrt::hresult_out_of_bounds(L"Key not found");
+    }
+
+    uint32_t OrderedMapView::Size() const
```
Scottj1s, line 198: noexcept

dev/Interop/StoragePickers/FileTypeChoicesMap.cpp
```diff
@@ -5,11 +5,34 @@
...
 namespace winrt::Microsoft::Windows::Storage::Pickers::implementation
 {
     FileTypeChoicesMap::FileTypeChoicesMap()
     {
+        // Pre-allocate capacity for typical file type choices (usually 2-8 items)
+        m_orderedMap.reserve(8);
+        m_keyToIndex.reserve(8);
+    }
+
+    size_t FileTypeChoicesMap::FindKeyIndex(hstring const& key) const
```
Scottj1s: lots of these methods can be made noexcept

dev/Interop/StoragePickers/FileTypeChoicesMap.cpp
codendone: Or have just one std::vector with a pair of the key and its value. Lookup may theoretically be slower, but the expected count is so low that it is unlikely to matter.

Scottj1s: > Or have just one std::vector with a pair of the key and its value. Lookup may theoretically be slower, but the expected count is so low that it is unlikely to matter.

that would resolve some scenarios but not all (TOCTOU)

dev/Interop/StoragePickers/FileTypeChoicesMap.cpp
```diff
@@ -5,11 +5,34 @@
...
+    {
+        auto it = m_keyToIndex.find(std::wstring(key.c_str()));
+        return (it != m_keyToIndex.end()) ? it->second : SIZE_MAX;
+    }
+
+    auto FileTypeChoicesMap::FindKey(hstring const& key) const
```
DrusTheAxe: `auto` return type for a method is cute but harder to reader/grok. That's a long-term maintainability tax which is the wrong vector (pardon the pun) -- generally best for code to be _easily_ maintainable, not hard to maintain.

SUGGESTION: Specify actual return type

same comment

dev/Interop/StoragePickers/FileTypeChoicesMap.cpp
DrusTheAxe: Is this API an inproc or OOP WinRT object?

If inproc then why should it internally protect instances against concurrent access? The method's not static / no global variables, instances are not typically expected to be accessed concurrently, and typically if a developer does concurrently access an instance of an object from 2 threads managing that concurrent access is _their_ responsbility.

If OOP then yes, every method exposed by every interface on the runtimeclass is potentially access concurrently and needs to protect itself.

dev/Interop/StoragePickers/FileTypeChoicesMap.cpp
```diff
@@ -30,42 +53,160 @@ namespace winrt::Microsoft::Windows::Storage::Pickers::implementation
...
+    }
+
+    // OrderedMapIterator implementation
+    winrt::Windows::Foundation::Collections::IKeyValuePair<hstring, winrt::Windows::Foundation::Collections::IVector<hstring>> OrderedMapIterator::Current() const
+    {
+        if (m_current >= m_map.size())
+        {
+            throw winrt::hresult_out_of_bounds();
+        }
+        // Create a simple key-value pair
```
DrusTheAxe, line 131: NIT: Add blank line 139.5

dev/Interop/StoragePickers/FileTypeChoicesMap.cpp
```diff
@@ -30,42 +53,160 @@ namespace winrt::Microsoft::Windows::Storage::Pickers::implementation
...
+        auto tempMap = single_threaded_map<hstring, winrt::Windows::Foundation::Collections::IVector<hstring>>();
```
DrusTheAxe: NIT: `auto tempMap{ ... };`

dev/Interop/StoragePickers/FileTypeChoicesMap.cpp
```diff
@@ -30,42 +53,160 @@ namespace winrt::Microsoft::Windows::Storage::Pickers::implementation
...
+        {
+            ++m_current;
+            return m_current < m_map.size();
+        }
+        return false;
+    }
+
+    uint32_t OrderedMapIterator::GetMany(array_view<winrt::Windows::Foundation::Collections::IKeyValuePair<hstring, winrt::Windows::Foundation::Collections::IVector<hstring>>> items)
+    {
+        uint32_t copied = 0;
```
DrusTheAxe: NIT: `uint32_t copied{};`

dev/Interop/StoragePickers/FileTypeChoicesMap.cpp
```diff
@@ -30,42 +53,160 @@ namespace winrt::Microsoft::Windows::Storage::Pickers::implementation
...
+        {
+            items[copied] = Current();
+            MoveNext();
+            ++copied;
+        }
+        return copied;
+    }
+
+    // OrderedMapView implementation
+    auto OrderedMapView::FindKey(hstring const& key) const -> vector_type::const_iterator
```
DrusTheAxe: Why `auto OrderedMapView::FindKey(hstring const& key) const -> vector_type::const_iterator`
and not `vector_type::const_iterator OrderedMapView::FindKey(hstring const& key) const`
?

dev/Interop/StoragePickers/FileTypeChoicesMap.h
DrusTheAxe: +1

Usually only need `const` + non-`const` for mutating methods. Read-only behavior only needs `const` method

dev/Interop/StoragePickers/FileTypeChoicesMap.h
```diff
@@ -28,7 +34,55 @@ namespace winrt::Microsoft::Windows::Storage::Pickers::implementation
...
+        OrderedMapIterator(vector_type const& map) : m_map(map), m_current(0) {}
+
+        winrt::Windows::Foundation::Collections::IKeyValuePair<hstring, winrt::Windows::Foundation::Collections::IVector<hstring>> Current() const;
+        bool HasCurrent() const;
+        bool MoveNext();
+        uint32_t GetMany(array_view<winrt::Windows::Foundation::Collections::IKeyValuePair<hstring, winrt::Windows::Foundation::Collections::IVector<hstring>>> items);
+
+    private:
+        vector_type const& m_map;
+        size_t m_current;
```
DrusTheAxe: `size_t m_current{};`

Always initialize variables

dev/Interop/StoragePickers/FileTypeChoicesMap.h
```diff
@@ -28,7 +34,55 @@ namespace winrt::Microsoft::Windows::Storage::Pickers::implementation
...
+    };
+
+    // Custom view to maintain insertion order  
+    struct OrderedMapView : implements<OrderedMapView,
+        winrt::Windows::Foundation::Collections::IMapView<hstring, winrt::Windows::Foundation::Collections::IVector<hstring>>,
+        winrt::Windows::Foundation::Collections::IIterable<winrt::Windows::Foundation::Collections::IKeyValuePair<hstring, winrt::Windows::Foundation::Collections::IVector<hstring>>>>
+    {
+        using vector_type = std::vector<std::pair<hstring, winrt::Windows::Foundation::Collections::IVector<hstring>>>;
+
+        OrderedMapView(vector_type const& map) : m_map(map) {}
```
DrusTheAxe: Why's this method inline but not any others?

e.g. `Size()` is equally trivial or more so but not inline

SUGGEST: Change this to non-inline for consistency with the rest of the code

dev/Interop/StoragePickers/FileTypeChoicesMap.h
```diff
@@ -28,7 +34,55 @@ namespace winrt::Microsoft::Windows::Storage::Pickers::implementation
...
+        uint32_t Size() const;
+        bool HasKey(hstring const& key) const;
+        void Split(winrt::Windows::Foundation::Collections::IMapView<hstring, winrt::Windows::Foundation::Collections::IVector<hstring>>& first,
+                  winrt::Windows::Foundation::Collections::IMapView<hstring, winrt::Windows::Foundation::Collections::IVector<hstring>>& second) const;
+
+        // IIterable
+        winrt::Windows::Foundation::Collections::IIterator<winrt::Windows::Foundation::Collections::IKeyValuePair<hstring, winrt::Windows::Foundation::Collections::IVector<hstring>>> First() const;
+
+    private:
+        vector_type const& m_map;
```
DrusTheAxe: What happens when the underlying `map` object passed in to the constructor is destroyed and then this object is used?

I think Scott mentioned this earlier (dangling pointer etc)

dev/Interop/StoragePickers/FileTypeChoicesMap.cpp
```diff
@@ -5,34 +5,39 @@
...
-        size_t index = FindKeyIndex(key);
-        return (index != SIZE_MAX) ? m_orderedMap.begin() + index : m_orderedMap.end();
+        for (auto it = m_orderedMap->cbegin(); it != m_orderedMap->cend(); ++it)
+        {
+            if (it->first == key)
+            {
+                return it;
+            }
+        }
+        return m_orderedMap->cend();
```
Scottj1s: consider using the standard library
```suggestion
    return std::ranges::find(std::cbegin(m_orderedMap), std::cend(m_orderedMap), key, &FileTypeChoiceVector::value_type::first);
```

dev/Interop/StoragePickers/FileTypeChoicesMap.cpp
```diff
@@ -5,34 +5,39 @@
...
-        size_t index = FindKeyIndex(key);
-        return (index != SIZE_MAX) ? m_orderedMap.begin() + index : m_orderedMap.end();
+        for (auto it = m_orderedMap->begin(); it != m_orderedMap->end(); ++it)
+        {
+            if (it->first == key)
+            {
+                return it;
+            }
+        }
+        return m_orderedMap->end();
```
Scottj1s: and here
```suggestion
    return std::ranges::find(m_orderedMap, key, &FileTypeChoiceVector::value_type::first);
```

dev/Interop/StoragePickers/FileTypeChoicesMap.cpp
```diff
@@ -171,16 +167,36 @@ namespace winrt::Microsoft::Windows::Storage::Pickers::implementation
...
+            return FileTypeChoiceVector::const_iterator{};
+        }
+        for (auto it = m_map->cbegin(); it != m_map->cend(); ++it)
+        {
+            if (it->first == key)
+            {
+                return it;
+            }
+        }
+        return m_map->cend();
```
Scottj1s:
```suggestion
 return std::ranges::find(std::cbegin(m_map), std::cend(m_map), key, &FileTypeChoiceVector::value_type::first);
```

dev/Interop/StoragePickers/FileTypeChoicesMap.h
```diff
@@ -28,7 +37,49 @@ namespace winrt::Microsoft::Windows::Storage::Pickers::implementation
...
+        // Helper methods
+        FileTypeChoiceVector::const_iterator FindKey(hstring const& key) const noexcept;
+        FileTypeChoiceVector::iterator FindKey(hstring const& key) noexcept;
+    };
+
+    // Custom iterator to maintain insertion order
+    struct OrderedMapIterator : implements<OrderedMapIterator,
+        winrt::Windows::Foundation::Collections::IIterator<winrt::Windows::Foundation::Collections::IKeyValuePair<hstring, winrt::Windows::Foundation::Collections::IVector<hstring>>>>
+    {
+        OrderedMapIterator(std::shared_ptr<FileTypeChoiceVector const> map);
```
Scottj1s: curious about the choice to copy/pass-by-value/move.  your earlier implementation was pass-by-const-ref/copy.  either way there's a copy, but the latter is simpler and you're passing one pointer (const ref) instead of two (std::shared_ptr value).

dev/Interop/StoragePickers/FileTypeChoicesMap.h
```diff
@@ -28,7 +37,49 @@ namespace winrt::Microsoft::Windows::Storage::Pickers::implementation
...
+        std::shared_ptr<FileTypeChoiceVector const> m_map;
+        size_t m_current{};
+    };
+
+    // Custom view to maintain insertion order  
+    struct OrderedMapView : implements<OrderedMapView,
+        winrt::Windows::Foundation::Collections::IMapView<hstring, winrt::Windows::Foundation::Collections::IVector<hstring>>,
+        winrt::Windows::Foundation::Collections::IIterable<winrt::Windows::Foundation::Collections::IKeyValuePair<hstring, winrt::Windows::Foundation::Collections::IVector<hstring>>>>
+    {
+        OrderedMapView(std::shared_ptr<FileTypeChoiceVector const> map);
```
Scottj1s: same question here

dev/Interop/StoragePickers/FileTypeChoicesMap.cpp
DinahK-2SO: Thanks @Scottj1s !

Added shared_ptr

dev/Interop/StoragePickers/FileTypeChoicesMap.h
DinahK-2SO: Thanks @codendone and @DrusTheAxe for the review! We need the non-const `FindKey` to retrieve a mutable iterator that Insert and Remove can use when updating or removing entries.

dev/Interop/StoragePickers/FileTypeChoicesMap.h
DinahK-2SO: Thanks for raising this point! Updated.

dev/Interop/StoragePickers/FileTypeChoicesMap.h
DinahK-2SO: Thanks for the review! Updated to use `shared_ptr`.

dev/Interop/StoragePickers/FileTypeChoicesMap.cpp
DinahK-2SO: Thanks @Scottj1s ! I updated in code.

dev/Interop/StoragePickers/FileTypeChoicesMap.h
DinahK-2SO: Thanks @Scottj1s for flagging that. You’re right - copying the `shared_ptr` is enough to bump the ref count. 
I’ve switched OrderedMapIterator and OrderedMapView back to taking `shared_ptr const &` .

dev/Interop/StoragePickers/FileTypeChoicesMap.h
DinahK-2SO: Thanks. Also switch ed to `shared_ptr const &` here.

dev/Interop/StoragePickers/FileTypeChoicesMap.cpp
DinahK-2SO: Hi @Scottj1s , @codendone , @DrusTheAxe , Thanks everyone for the thorough feedback! 

Based on the current build layout, the storage pickers reside in Microsoft.WindowsAppRuntime.dll, and indeed we're not expecting too many pairs, so 

* I’m treating FileTypeChoicesMap as an in-proc C++/WinRT object whose per-instance synchronization remains with the caller (a.k.a no extra lock inside).

* I dropped the secondary index map in FileTypeChoicesMap and rely on the single insertion-ordered `std::vector<std::pair<>>`

PR #5950, opened 2025-10-28
Setting strict mode and error action preference to stop on remaining PS scripts

guimafelipe:
That'll set the default to choke on errors (including syntax ones). Any command you want to handle errors can use the `-ErrorAction`  option e.g:
```powershell
Add-AppxPackage $package -ErrorVariable appxerror -ErrorAction SilentlyContinue
```

Note: the only powershell script not changed was the Dev-Check configuration script, as it is just holding variables.

WARNING:

### DIFF
--- a/build/InstallVCLibs.ps1
+++ b/build/InstallVCLibs.ps1
@@ -2,7 +2,7 @@ param(
     [Parameter(Mandatory = $true)]
     [ValidateSet("x86", "x64", "arm64")]
     [string]$Platform,
-    
+
     # Expecting BuildOutput\Release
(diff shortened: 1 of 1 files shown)


lauren-ciha (commented):
Catching errors early is a great quality improvement. Are there any script errors you've seen in our pipeline runs that would be blocked by strict mode?

guimafelipe:
> Catching errors early is a great quality improvement. Are there any script errors you've seen in our pipeline runs that would be blocked by strict mode?

The `Install-VCLibs` script for example had an error that I needed to change within this PR.

